PR #3843, opened 2022-09-24
add option EXTCLK and fix reset device

standahabich:


standahabich (commented):
rename constant

standahabich (commented):
fix format

standahabich (commented):
clang valid

standahabich (commented):
add option EXTCLK and fix reset device

PR #3844, opened 2022-09-24
Fix time/automation (cron) wdt crash when time jumps ahead too much

jpeletier:

When time synchronization takes place and the clocks have drifted apart too ahead of time, a wdt crash happens due to a very long loop in `void CronTrigger::loop()`, The below loop iterates for every second of difference to trigger automations:

```cpp
    while (true) {
      this->last_check_->increment_second();
      if (*this->last_check_ >= time)
        break;

      if (this->matches(*this->last_check_))
        this->trigger();
    }
```
 
This fix detects jumps in time that are too big (in the same way a jump could happen backwards) and avoids entering a potentially too long loop.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

## Checklist:

### DIFF
--- a/esphome/components/time/automation.cpp
+++ b/esphome/components/time/automation.cpp
@@ -31,6 +31,13 @@ void CronTrigger::loop() {
       return;
     }
 
+    if (time > *this->last_check_ && time.timestamp - this->last_check_->timestamp > 900) {
+      // We went ahead in time (a lot), probably caused by time synchronization
+      ESP_LOGW(TAG, "Time has jumped ahead!");
+      this->last_check_ = time;
+      return;
+    }
+
     while (true) {
       this->last_check_->increment_second();
       if (*this->last_check_ >= time)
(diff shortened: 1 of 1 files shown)


esphome/components/time/automation.cpp
```diff
@@ -31,6 +31,13 @@ void CronTrigger::loop() {
       return;
     }
 
+    if (time > *this->last_check_ && time.timestamp - this->last_check_->timestamp > 900) {
+      // We went ahead in time (a lot), probably caused by time synchronization
+      ESP_LOGW(TAG, "Time has jumped ahead!");
+      this->last_check_ = time;
+      return;
+    }
```
GregJArnold: A little nitpicky, but I'd personally move this block to after the first block, on L29.

GregJArnold (commented):
Looks good.

esphome/components/time/automation.cpp
jpeletier: @DrunkPyromaniac you mean as an `else if` ?

GregJArnold: Yes, that way all the code for "out of whack" time is in one logical block.

jesserockz (approved):

PR #3845, opened 2022-09-24
PCA9685, fix reset device and add option EXTCLK

standahabich:

Added support for external clocks. To turn off EXTCLK it is necessary to perform SWRST which did not work.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2325

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266

```yaml
# Example config.yaml
    # Example configuration entry with external clock enable and clock source from ESP32
    pca9685:
      - id: pca9685_hub1
        external_clock_input: true

    # Individual outputs
    output:
        # generate PWM from ESP32 ledc
      - platform: ledc
        pin: 25
        id: extclk
        min_power: 1
        frequency: 40MHz

      - platform: pca9685
        pca9685_id: 'pca9685_hub1'
        channel: 0

```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs/pull/2325).


standahabich (commented):
Constant CONF_EXTCLK = "extclk" is not ordered

jesserockz (changes requested):
If its the case where frequency is no longer used, there should be a custom validation that enforces either frequency **or** ext clock to be set.

With this, the whole pre_scaler calculation and log message should be moved into `else` on [line 74](https://github.com/esphome/esphome/pull/3845/files#diff-d476747256a325bc04694f3587c53a43eacbb892b9a4073c3623eed304bfd536R74)

jesserockz (approved):

PR #3854, opened 2022-09-29
Add support for parsing the short local name in the tracker

bdraco:

Fixes missing short local names

Some devices only send a short local name, not a full one.  In Home Assistant we can't detect these without the short local name being sent since some integrations only match on local name.

related PR https://github.com/home-assistant/core/pull/79321

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
+++ b/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
@@ -644,6 +644,13 @@ void ESPBTDevice::parse_adv_(const esp_ble_gap_cb_param_t::ble_scan_result_evt_p
         this->name_ = std::string(reinterpret_cast<const char *>(record), record_length);
         break;
       }
+      case ESP_BLE_AD_TYPE_NAME_SHORT: {
+        // SHORTENED LOCAL NAME
+        // "The Shortened Local Name data type defines a shortened version of the Local Name data type. The Shortened
+        // Local Name data type shall not be used to advertise a name that is longer than the Local Name data type."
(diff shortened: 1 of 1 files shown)


esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -637,10 +637,14 @@ void ESPBTDevice::parse_adv_(const esp_ble_gap_cb_param_t::ble_scan_result_evt_p
     // (called CSS here)
 
     switch (record_type) {
+      case ESP_BLE_AD_TYPE_NAME_SHORT:
       case ESP_BLE_AD_TYPE_NAME_CMPL: {
         // CSS 1.2 LOCAL NAME
         // "The Local Name data type shall be the same as, or a shortened version of, the local name assigned to the
         // device." CSS 1: Optional in this context; shall not appear more than once in a block.
+        // SHORTENED LOCAL NAME
```
balloob, line 645: should we guard in case the complete name comes before the short one?

bdraco: I guess its possible to send both in the same packet.  That would be a bit silly

bdraco: someone will do it. I'll add a guard

jesserockz (approved):

PR #3855, opened 2022-09-29
Add Rx buffer flush to get rid of CRC errors

Darki03:

Due to persistent CRC errors even with correct reception confirmed (am I the only one in this case? ), i've investigate teleinfo.cpp code. To get rid of CRC errors I have to :
- increase "real" uart Rx buffer size in esphome yaml file (rx_buffer_size: 2048 for me, default size of 256 bytes is too small versus standard mode and its 865 bytes at 9600 bits/s)
- Add a Rx buffer flush in update() function of teleinfo.cpp in order to ensure that a unique full frame will be read

Since these modifications no more CRC errors observed over a month of continuous working.

## Types of changes

- [X ] Bugfix (non-breaking change which fixes an issue)
- [ ] Other

## Test Environment

- [X ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
# Example config.yaml

uart:
  id: uart_bus
  rx_pin:
    number: GPIO16
  #tx_pin: GPIO1
  baud_rate: 9600
  parity: EVEN
  data_bits: 7
  rx_buffer_size: 2048
  #debug:
    #direction: RX
    #dummy_receiver: false
    #after:
      #delimiter: "\n"
    #sequence:
      #- lambda: UARTDebug::log_string(direction, bytes);


teleinfo:
  id: myteleinfo
  update_interval: 10s
  historical_mode: false

## Checklist:
  - [X ] The code change is tested and works locally.

### DIFF
--- a/esphome/components/teleinfo/teleinfo.cpp
+++ b/esphome/components/teleinfo/teleinfo.cpp
@@ -70,6 +70,10 @@ void TeleInfo::setup() { state_ = OFF; }
 void TeleInfo::update() {
   if (state_ == OFF) {
     buf_index_ = 0;
+	// Flush Rx buffer at update. Ensure to start on a new Teleinfo frame
+	while (available() > 0) {
+       read();
+    }
     state_ = ON;
   }
 }


0hax:
Hi,
Thanks for contributing.
I'm using the standard configuration at 9600 too and it's running for years now without any crc issues.
So I'm wondering if this could be an issue with your circuit which generate noise on the line and which are read.
Not sure if you are french or not (I guess yes since you are using teleinfo :) ) but can you check the drawing here https://0hax.github.io/post/teleinformation/ and let me know if your circuit looks the same?
In case everything is fine, can you update the `update_interval` to 60s for example and see if crc errors disappear?

Darki03:
Hello (bonjour ;))

Thanks for your answer. It is very strange because i've spent a lot of time investigate this "issue".

My electonic interface seems correct ( opto and MOSFET) checked with Logic analyser. Furthermore, tests with "text_sensor" uart reveals only good frames !

Before modification, the CRC error occurs always on the same TAG which located just after the first 256 bytes of one frame ( this is to deterministic to be linked with a noise problem). First i have increase the buffer size which result in truncated frames, only RX buffer flush at update solves the problem ( whatever update rate used)

Maybe it is linked to the use of the hardware uart of the ESP32 or to the breadboard in use ?

I will try to test with software uart and an other ESP32 breadboard.

Thanks

0hax:
> Hello (bonjour ;))
> 
> Thanks for your answer. It is very strange because i've spent a lot of time investigate this "issue".
> 
> My electonic interface seems correct ( opto and MOSFET) checked with Logic analyser. Furthermore, tests with "text_sensor" uart reveals only good frames !
> 
> Before modification, the CRC error occurs always on the same TAG which located just after the first 256 bytes of one frame ( this is to deterministic to be linked with a noise problem). First i have increase the buffer size which result in truncated frames, only RX buffer flush at update solves the problem ( whatever update rate used)
> 
Ok interesting, what TAG was that? Also can you share the list of tags you are reading? I'm only reading something like 4 tags.
It could be that the issue happen only when reading *many* tags.

> Maybe it is linked to the use of the hardware uart of the ESP32 or to the breadboard in use ?
> 
> I will try to test with software uart and an other ESP32 breadboard.
> 
I'm afraid it will be worst with software uart. Keep using hardware uart, there is low chance to be an issue with the breadboard.

PR #3859, opened 2022-10-02
Update the ibeacon code

fhriley:

* Add min/max interval configs so the user can set the interval.
* Set the default min/max interval to 100ms per Apple's iBeacon specification.
* Add a measured power config so the user can calibrate the iBeacon per Apple's instructions.
* Add a tx power config to set the transmit power.
* Make dump_config() more useful.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2337

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
esp32_ble_beacon:
  type: iBeacon
  uuid: '00000000-0000-0000-0000-000000000001'
  major: 1
  minor: 1
  min_interval: "90ms"
  max_interval: "110ms"
  measured_power: -65
  tx_power: "9dBm"
```

## Checklist:

### DIFF
--- a/esphome/components/esp32_ble_beacon/__init__.py
+++ b/esphome/components/esp32_ble_beacon/__init__.py
@@ -1,6 +1,6 @@
 import esphome.codegen as cg
 import esphome.config_validation as cv
-from esphome.const import CONF_ID, CONF_TYPE, CONF_UUID
+from esphome.const import CONF_ID, CONF_TYPE, CONF_UUID, CONF_TX_POWER
 from esphome.core import CORE
 from esphome.components.esp32 import add_idf_sdkconfig_option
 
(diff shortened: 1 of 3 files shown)


esphome/components/esp32_ble_beacon/__init__.py
```diff
@@ -12,16 +12,41 @@
...
+            ),
+            cv.Optional(CONF_MAX_INTERVAL, default=100): cv.int_range(
+                min=20, max=10240
+            ),
+            cv.Optional(CONF_MEASURED_POWER, default=-59): cv.int_range(
+                min=-128, max=0
+            ),
+            cv.Optional(CONF_TX_POWER, default=3): cv.one_of(
+                -12, -9, -6, -3, 0, 3, 6, 9, int=True
+            ),
```
jesserockz: Please have a look at other config variables where the units are used. This makes is easier to understand when people are reading and writing their configs.

For example, an `min_interval: 100` means nothing without a unit there, it could be milliseconds, nanoseconds, seconds etc...
You can use `cv.positive_time_period_milliseconds` inside a `cv.All` to combine the int range and the unit

fhriley:
@jesserockz Changes made.

jesserockz (approved):

PR #3860, opened 2022-10-02
Add ble RSSI sensor for connected devices

RoboMagus:

Adds the option to add an RSSI sensor to `ble_client`, to poll the RSSI for connected devices. 
Currently, without this change, RSSI can only be read from devices not yet connected.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2339

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
# Example config.yaml
ble_client:
  - mac_address: ${iTag_MAC}
    id: action_iTag

sensor:
  - platform: ble_rssi
    mac_address: ${iTag_MAC}
    name: "iTag RSSI"

  - platform: ble_client
    type: rssi
    ble_client_id: action_iTag
    name: "iTag RSSI (c)"

```

## Checklist:

### DIFF
--- a/esphome/components/ble_client/ble_client.cpp
+++ b/esphome/components/ble_client/ble_client.cpp
@@ -64,6 +64,13 @@ void BLEClient::gattc_event_handler(esp_gattc_cb_event_t event, esp_gatt_if_t es
   }
 }
 
+void BLEClient::gap_event_handler(esp_gap_ble_cb_event_t event, esp_ble_gap_cb_param_t *param) {
+  BLEClientBase::gap_event_handler(event, param);
+
+  for (auto *node : this->nodes_)
+    node->gap_event_handler(event, param);
+}
+
 void BLEClient::set_state(espbt::ClientState state) {
   BLEClientBase::set_state(state);
   for (auto &node : nodes_)
--- a/esphome/components/ble_client/ble_client.h
+++ b/esphome/components/ble_client/ble_client.h
@@ -27,7 +27,8 @@ class BLEClientNode {
  public:
   virtual void gattc_event_handler(esp_gattc_cb_event_t event, esp_gatt_if_t gattc_if,
                                    esp_ble_gattc_cb_param_t *param) = 0;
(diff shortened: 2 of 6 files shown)


jesserockz (dismissed):
see below

jesserockz (changes requested):
This will break if someone tries to use this without having a top level `sensor` block in yaml.

I would suggest extending the current `ble_client` `sensor` to accept a `typed_schema` with `characteristic` and `rssi` types and the default being `characteristic` (the current type) which will allow further extensions again later on.

YAML would look like this:
```yaml
ble_client:
  mac: ...
  id: ble_client_1
  
sensor:
  - platform: ble_client
    type: rssi
    ble_client_id: ble_client_1
    name: Client 1 RSSI
```

RoboMagus:
Thanks for the feedback. I like your suggestion as it does integrate much better with the `ble_client` sensor structure used for characteristics.

However, I think making type a mandatory part of `ble_client` sensors, but having it implicitly default to `characteristic` would cause confusion. Especially when later on more sensor types could be added to the `ble_client` platform.
IMO making it explicit instead would be the better options, though that would introduce a breaking change which in general I'd like to avoid.

Is there a way to issue a warning when validating a config to inform the users about having to add `type: characteristic` to existing `ble_client` sensors? If so, that would both keep things clear and explicit and inform users about the required change.

jesserockz:
Happy for it to be explicit too. cv.typed_schema will complain with `type not specified` if not added in config.

RoboMagus:
Refactoring after the merge conflicts was a bit of a challenge, but after making the requested changes I do like this implementation a whole lot better. Thanks for the suggestion @jesserockz.

I've had a couple of esp32 boards scattered around the house running code including these changes and it works as intended.

jesserockz (approved):

PR #3864, opened 2022-10-03
fix never calling preset change trigger

### LINKED ISSUE #3615: Climate Thermostat preset_change is not triggered
### The problem

I'm creating an thermostat for my heating and try to trigger a script when the preset is changed to boost. But the `preset_change` is not triggered at all when I change the preset via Home-Assistant.

2022.9.0

Home Assistant Add-on

Home Assistant 2022.9.6

ESP32

### Board

Sonoff NSPanel

thermostat

### Example YAML snippet

```yaml
climate:
  - platform: thermostat
    id: ${id}
    name: ${name}
    sensor: ${sensor}
    visual:
      min_temperature: 20 °C
      max_temperature: 25 °C
      temperature_step: 0.1 °C
    default_target_temperature_low: 20.5
    default_mode: heat
    heat_overrun: 0.0 °C
    heat_deadband: 0.2 °C
    # min_heating_off_time: 60min
    # min_heating_run_time: 5min
    # min_idle_time: 30min
    min_heating_off_time: 5sec
    min_heating_run_time: 5sec
    min_idle_time: 5sec
    heat_action:
      - switch.turn_on: relay_1
    idle_action:
      - switch.turn_off: relay_1
    preset:
      - name: eco
        default_target_temperature_low: 20.5
      - name: home
        default_target_temperature_low: 21.5
      - name: boost
        default_target_temperature_low: 24
    preset_change:
        - logger.log: Preset has been changed! Checking boost
        - if:
            condition:
              lambda: |-
                return id(${id}).preset == CLIMATE_PRESET_BOOST;
            then:
              - logger.log: "Starting Boost"
              - script.execute: ${id}_boost_stop
            else:
              - logger.log: "Stopping Boost"
              - script.stop: ${id}_boost_stop
    # target_temperature_change_action:
    #   - script.stop: ${id}_boost_stop
```

_No response_

### Additional information

_No response_

glmnet:

Thermostat never calling change preset trigger, see fixed issue

Note to reviewer: I assume `change_preset_internal_` is only called when preset changes, seems to work ok with quick tests.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/3615

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/thermostat/thermostat_climate.cpp
+++ b/esphome/components/thermostat/thermostat_climate.cpp
@@ -974,8 +974,11 @@ void ThermostatClimate::change_preset_(climate::ClimatePreset preset) {
   auto config = this->preset_config_.find(preset);
 
   if (config != this->preset_config_.end()) {
+    ESP_LOGI(TAG, "Preset %s requested", LOG_STR_ARG(climate::climate_preset_to_string(preset)));
     if (this->change_preset_internal_(config->second)) {
-      ESP_LOGI(TAG, "Applied preset %s", LOG_STR_ARG(climate::climate_preset_to_string(preset)));
+      ESP_LOGI(TAG, "Preset %s applied", LOG_STR_ARG(climate::climate_preset_to_string(preset)));
+    } else {
+      ESP_LOGI(TAG, "No changes required to apply preset %s", LOG_STR_ARG(climate::climate_preset_to_string(preset)));
     }
     this->custom_preset.reset();
     this->preset = preset;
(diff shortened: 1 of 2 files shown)


jesserockz:
I think the issue here is that in `this->preset != preset`, `preset` is `this->preset` as there is no local variable

glmnet:
It is. But real question is if there is a need to check if this code can be reached without a real preset change.

kbx81:
I glanced at this earlier and, yeah, I do think we should retain the conditional because I don't think it makes sense to call the action if no change occurred. But that's just my two cents. haha

I think the way it is generally used is to alert some users who wish to store the (previous) set points when the preset is changed so that they may be restored later.

glmnet:
Well the check should be moved up in the call stack May be. Please sb take over the pr.

kbx81:
Alright -- should be good. I just added more proper change detection logic around the relevant parts and tested locally with both custom and built-in presets. Works great on my machines! 😁

glmnet:
Si this would not trigger a preset change if you are using two presets with the very same configuration. That is still a valid case specially when someone wants to change something on his own. Eg. A fan or duct opening etc.

kbx81:
Ah, you are correct...this is why we do PRs...and why I shouldn't try to fix things after a long exhausting day. haha
I'll push up another patch later today. 😄

kbx81:
I believe this now addresses all of the concerns/issues and should be ready to merge 😇

PR #3868, opened 2022-10-04
Add basic filtering to bluetooth_proxy

marcelveldt:

The bluetooth_proxy forwards all broadcasts/packets while we're only interested in a small part of the messages. As a result we're stressing both the network and HA core with a lot of messages.

In the future a somewhat smart filtering mechanism is thinkable where the esp can retrieve a list of devices we're interested in but this needs more discussion.

As a quick, simple but effective MVP filter we discussed filtering out packets with a very low RSSI and duplicate packets.

My tests conclude that this simple fix already brings down the amount of traffic by like 80% and reducing delays.
Marked as bug fix because without this filter problems will be reported for users running the proxy on wifi devices in crowded (many blue packets) areas.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -19,10 +20,34 @@ BluetoothProxy::BluetoothProxy() {
...
 
+  adv_data_t prev_data{};
+  if (this->prev_data_map_.find(device.address_uint64()) != this->prev_data_map_.end()) {
+    prev_data = this->prev_data_map_[device.address_uint64()];
+  }
+
+  for (auto &data : device.get_manufacturer_datas()) {
+      // ignore duplicate data
+      if (data.data == prev_data) {
+        ESP_LOGV(TAG, "Ignoring packet from %s - %s. RSSI: %d dB", device.get_name().c_str(), device.address_str().c_str(),
```
jesserockz:
```suggestion
        ESP_LOGV(TAG, "Ignoring duplicate packet from %s - %s. RSSI: %d dB", device.get_name().c_str(), device.address_str().c_str(),
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -19,10 +20,34 @@ BluetoothProxy::BluetoothProxy() {
...
+            device.get_rssi());
+        return true;
+      }
+
+      this->prev_data_map_[device.address_uint64()] = data.data;
```
bdraco: I think you'll run out of ram rather quickly with this if there are a lot of devices.

Maybe only track against the previous RSSI for a fixed number with an LRU strategy?

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -19,10 +20,34 @@ BluetoothProxy::BluetoothProxy() {
...
+    }
+  
+  ESP_LOGD(TAG, "Proxying packet from %s - %s. RSSI: %d dB", device.get_name().c_str(), device.address_str().c_str(),
```
jesserockz: Debug here just adds too much log messages for regular log checking
```suggestion
  ESP_LOGV(TAG, "Proxying packet from %s - %s. RSSI: %d dB", device.get_name().c_str(), device.address_str().c_str(),
```

jesserockz (changes requested):

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
marcelveldt: well, the rssi filter comes first so a lot is filtered out by that already. My C++ is a bit rusty... any easy way to size limit the map ? Ideally you want a FIFO like list which is limited to like 100 entries or so

marcelveldt: I have loaded this patch on 5 esp's throughout the house and I have A LOT of bluetooth traffic here.
Still going strong so not running out of memory. Good thing is that the amount of packets significantly dropped while response times from e.g. motion sensors is now better.

bdraco: There is a simple one here https://github.com/lamerman/cpp-lru-cache

I can’t say I used it though as I found it by looking for a small one with lots of stars

bdraco: This one seems to have performance comparison of a few options

https://github.com/nitnelave/lru_cache

the static one might make the most sense to avoid any memory issues

marcelveldt:
I come back to my previous findings. Response times of the motion sensors is still horrible, a real hit and miss.
Looks like half of the packets get lost and the filtering only helps a bit. At least it reduces stress on the WiFi and cpu usage on the HA server.

I'm starting to think this is related to the fact that I previously was running these QingPing sensors in Xiaomi mode and I guess in that mode an event is sent out multiple times and in QingPing mode only once. Due to the fact that the esp is only listening x amount of time for incoming packets this explains the hit and miss behavior. As far as I've debugged it is not dropping packets because its too busy as that would have been logged.

Anyways, long story short. This fix does not fix my particular issue but it does help in performance because it reduces stress (number of unusable packets) on the network and the HA core.

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
marcelveldt: Thanks for the tip on the lru_cache but I have no clue how to implement that on the dictionary-like object that I have now. Basically I'm just storing the previous value in a mapping where the MAC is the key. When a new packet comes in I want to compare the previous manufacturer_data (and maybe also the service data) and if it's the same, ignore the message. That is the same approach as bleak/dbus seems to to with a real hardware dongle.

Now only I'd like to size/number limit the mapping so it can only contain say 50 items/keys and it starts overwriting in a FIFO matter...

BTW: Still not ran into memory issues on the 5 nodes that I flashed with this patch

marcelveldt: This may be something ideal: https://github.com/nlohmann/fifo_map

bdraco: Unless I'm reading it wrong it looks like the fifo map is unbounded and there is nothing pruning the old values?

jesserockz: > There is a simple one here [lamerman/cpp-lru-cache](https://github.com/lamerman/cpp-lru-cache)

This one looks nice and simple to implement in ESPHome as a helper class (with credit of course)

marcelveldt: I've adjusted the code to the LRU cache mentioned above.

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -19,10 +20,30 @@ BluetoothProxy::BluetoothProxy() {
...
 
 bool BluetoothProxy::parse_device(const esp32_ble_tracker::ESPBTDevice &device) {
+  if (device.get_rssi() < this->min_rssi_) {
+    // ignore devices below the rssi threshold
+    ESP_LOGVV(TAG, "Ignoring packet from %s - %s. RSSI: %d dB", device.get_name().c_str(), device.address_str().c_str(),
+              device.get_rssi());
+    return true;
+  }
+
+  for (auto &data : device.get_manufacturer_datas()) {
```
bdraco, line 27: We need to handle service uuids and service data changing as well or this will break all the devices that get their information from service data

bdraco: Maybe its simpler to ignore advertisements that only contain the RSSI?

I haven't looked at the distribution of the data though so that might not be feasable

marcelveldt: How does that break as service data is just a part of the manufacturer data ?

bdraco: https://github.com/esphome/esphome/blob/fd57b21affab08a48fd03e7b1e61c522fd524f0b/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp#L49

https://github.com/esphome/esphome/blob/fd57b21affab08a48fd03e7b1e61c522fd524f0b/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp#L52

It looks separate to me but maybe I'm missing something?

marcelveldt: Yeah, I was comparing incoming data and noticed the service data just being a part (in raw bytes) of the service data but if I made the wrong assumption we can add an additional check/compare

marcelveldt: > Maybe its simpler to ignore advertisements that only contain the RSSI?
> 
> I haven't looked at the distribution of the data though so that might not be feasable

Well, that is not how it works. You just get the same data multiple times and in 8/10 cases only the RSSI changed.
So basically you just have to compare the data and if it's the same as previously, ignore. I think that is also what dbus/bleak does ?

bdraco: We do the filtering similar to how you are describing here https://github.com/home-assistant/core/blob/712f40b6b031f60c52b1c040e8494127fd7a1596/homeassistant/components/bluetooth/manager.py#L353

bleak&bluez build up the data into a dict like structure so each change gets overlayed on top

marcelveldt: Yeah so this check will be duplicate with the exception that it now does not have to be sent over the API which saves bandwidth and processing power. I'm just hunting for the cause why the current implementation works so much worse than the experimental approach with blue_monitor and the parse_data hack. That worked flawlessly. Motion sensor events got picked up instantly and (almost) never missed. So why do I miss so many events is the question. Apparently I'm not alone as others mention the same.

I will investigate some more.

jesserockz: I guess the passive monitor gateway component for esphome had a mac address filter in yaml so it was only sending what you cared about. But in the name of simplicity and user friendliness, we dont want to have a hardcoded list in yaml in ESPHome.

marcelveldt: No indeed, that is what we would want to avoid because having to administer those MAC addresses manually is a real pain. Is there any way we could log the load of the ESP somehow @jesserockz that will tell us if its just too busy forwarding packets to the API or maybe it is dropping packets already when received on the radio.

I'm aware of the limitation that the radio is shared between wifi and bluetooth but still I have a feeling that something is wrong. Door/window sensors now work 5 out of 10 times, even when close to a proxy and even multiple proxies in the same neighborhood.

marcelveldt:
@jesserockz @bdraco how do we want to move forward with this one ?
Maybe for now only implement a minimum RSSI filter and perhaps a manual MAC-adress filter ?

Another option could be that we simply require ethernet connection for the bluetooth proxy and provide no support for WiFi based devices. ?

doguitar:
Would it be possible to add a mac filter in the proxy's yaml and a button to ignore the filter temporarily? 

I really like the idea of a filter but I would definitely want a way to detect new devices similar to something like the zwave inclusion mode.

marcelveldt:
> Would it be possible to add a mac filter in the proxy's yaml and a button to ignore the filter temporarily?
> 
> I really like the idea of a filter but I would definitely want a way to detect new devices similar to something like the zwave inclusion mode.

ideally that is what's needed indeed but we do not want users to fill in Mac addresses in the config but instead work with a an automated list. Maybe as quick temporary solution a plain list of Mac addresses is possible.

WormyOne:
Just a note to add my 2 cents/experience:

- I'm using a number of ATC_ and MHO_ thermometers, and some some QingPing PIR sensors
- PIR reliability and latency is really key...
- was using ble_monitor with MAC filtering, and this worked extremely well 
- moved over to bluetooth_proxy and now experiencing same issues as reported in this thread
- I'd be fine with any of the solutions being discussed:  MAC filtering/prioritisation, proxy only on wired ethernet ESPs 

Watching this issue closely.  Incredible how useless a PIR is if you can't count on it....

thanks for efforts all

bdraco:
Only tangentially related to this PR, but part of the latency is Home Assistant side using pure python protobuf and not the cpp ones which are much slower

https://github.com/bdraco/fast_protobuf

marcelveldt:
The issue is not only that it's slower, the main issue is that it is not reliable due to missing packets. I actually ditched all qingping motion sensors for now because my family got crazy of the unreliability and I was out of ideas to solve it. I guess the only valid way is by white listing MAC addresses because that worked with the old blemonitor approach really solid.

bdraco:
We could send the list of macs we are subscribed on the hass side, but that wouldn't work with discovery, but I guess we could add a button entity in Home Assistant that would trigger the filter to be removed for a few seconds to enable discovery

marcelveldt:
That sounds like a very valid plan to me. Easy but effective. Button entity to enable discovery mode for 30 seconds or so

PeteBa:
> Button entity to enable discovery mode for 30 seconds or so

Would it make sense for the proxy to filter for very close proximity (say RSSI better than -30) during the window and for HASS to message a "bring the device close to discover" instruction ? Given the user is manually triggering the process in HA, they probably have the new device in their hand and it doesn't feel like a big ask to get them to put it near one of the proxies. That way, when it pops up in HA they know with confidence what it is without the need to refer to mac/uuids.

I recently migrated a bunch of ibeacons, mifloras and xiaomi sensors into ESPHome/HA. Rather than having to break out nRF Connect to identify, log and reconcile all the mac/uuids, I could have just moved them one-at-a-time closer to my esphome proxy. Wasn't helped that I live in a city centre full of stray neighbourhood traffic!

Sorry if straying from original topic.

bdraco:
@marcelveldt can you give dev a spin?

Jesse got the idf upgraded and I fixed a few leaks and reliability issues in the tracker.

bdraco:
I think we could replace this with an api that sets the scan filter process and maintains a allowlist (docs call it whitelist) of addresses
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/bluetooth/esp_gap_ble.html#_CPPv421esp_ble_scan_filter_t

We could than add a switch entity in hass that turned on/off the allowlist

robloh:
An allow list would be fantastic. I don’t need all the BLE LED strips to be found. After I’ve added my known devices I’d love to turn off discovery again and reduce network traffic

PR #3878, opened 2022-10-06
Add ADC2 support for ESP32

rkohrt:

This change adds support for the ADC2 to the `adc` component for ESP32, I have tested it locally on a ESp32S2, I could not test it on an ESP32, since I don't have one.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
  - platform: adc
    pin: GPIO17
    id: adc2_sensor
    attenuation: auto

  - platform: adc
    pin: GPIO10
    id: adc1_sensor
    attenuation: auto
```

## Checklist:


esphome/components/adc/adc_sensor.cpp
```diff
@@ -123,10 +135,21 @@ float ADCSensor::sample() {
 #endif
 
 #ifdef USE_ESP32
+esp_err_t get_adc_raw(const adc1_channel_t& channel1, const adc2_channel_t& channel2, ADCSensor::Channel channel_num, int& raw)
+{
+  if (channel_num == ADCSensor::Channel::channel1)
```
rkohrt: I would have liked to make this check at compile time, but mam not familiar enough with esphome to figure out a way to do that.

ssieb:
There is a reason why there's no support in esphome for ADC2.  Espressiv specifically says you can't use ADC2 with wifi on an ESP32.  I don't know if this is different on the later chips.

PR #3879, opened 2022-10-06
Fix sampling rate of [q|h]mc5883l only using seconds field of update interval

LeoVerto:

The previous implementation only used the seconds field of the update interval to calculate the sampling rate.

This would not only ignore the values specified in the minutes or milliseconds field but also result in a divison by zero when the seconds field was unset.

I've tested the code on a Wemos D1 Mini with a qmc5883l.

## Types of changes

- [ ] Other
## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266

## Checklist:

### DIFF
--- a/esphome/components/hmc5883l/sensor.py
+++ b/esphome/components/hmc5883l/sensor.py
@@ -112,6 +112,7 @@ CONFIG_SCHEMA = (
 
 
 def auto_data_rate(config):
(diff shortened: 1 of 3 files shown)


jesserockz:
Sorry I missed this PR and merged in #4031 since.

PR #3886, opened 2022-10-08
Mitsubishi Climate updates

RubyBailey:

Updating Mitsubishi Climate with the following
- Adds Receiver support
- Adds Fan Modes for 3 and 4 fan speeds as well as Quiet fan mode
- Adds supports_dry and supports_fan_only modes
- Adds Horizontal or Wide Vane default setting
- Adds Vertical Vane default setting

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2353

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266

```yaml
climate:
  - platform: mitsubishi
    name: "Heatpump"
    supports_dry: "true"
    supports_fan_ony: "true"    
    horizontal_default: "left"
    vertical_default: "down"    
    receiver_id: rcvr

remote_receiver:
  id: rcvr
  pin: 
    number: GPIO14
    inverted: True
    mode:
      input: True
      pullup: true
  tolerance: 55%

```

## Checklist:
  - [x] Tests have been updated to verify that the new code works (under `tests/` folder).


KoenVanduffel:
I ran into the same issue not being able to set the fan speeds. With great help of jesserockz i included 
```yaml
external_components:
  - source: github://pr#3886
    components: mitsubishi
```
into my yaml and managed to get the fan speed setting to work.

esphome/components/mitsubishi/climate.py
```diff
@@ -9,13 +9,81 @@
...
+    "fan4": SetFanSpeeds.MITSUBISHI_FAN_4,
+    "fan5": SetFanSpeeds.MITSUBISHI_FAN_5,
+}
+
+
+CONF_HORIZONTAL_DEFAULT = "horizontal_default"
+HorizontalDirections = mitsubishi_ns.enum("HorizontalDirections")
+HORIZONTAL_DIRECTIONS = {
+    "left": HorizontalDirections.HORIZONTAL_DIRECTION_LEFT,
+    "mleft": HorizontalDirections.HORIZONTAL_DIRECTION_MLEFT,
```
jesserockz: I would prefer if these strings were full strings and not shortened please. They are only used in config, but `middle-left` is way easier to understand compared to `mleft`

RubyBailey: I will make the changes when I get back to my computer. Thanks for the review. 
The funny thing is that I had initially looked at using full text, but then saw that the Arduino-heatpump component was using these shortened ones. I assumed shortened ones would be preferred. 😁

jesserockz: Ah I see, guess I missed that when it was merged. Would still be nicer to change to full.

RubyBailey: Agreed. I’ll change it. A heads-up that it will be a couple of months before the changes are made (in this PR and docs). Back at my computer then. (Travelling, actually in your country now 😉)

RubyBailey: Hello, Found some time to update and submit the config text as discussed above. 

Unclear why test3.1 is failing, appears to be unrelated to the component or these text changes, with the following error message:
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/8e5e7e5ab8b370d6c329ec480221332ada57f0ab'. 
Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
Warning: Back off 27.535 seconds before retry.
Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.

BTW, We enjoyed NZ, great place.

esphome/components/mitsubishi/mitsubishi.cpp
```diff
@@ -22,37 +35,129 @@ const uint16_t MITSUBISHI_HEADER_MARK = 3500;
...
+      remote_state[6] = MITSUBISHI_MODE_HEAT;
+      remote_state[8] = MITSUBISHI_MODE_A_HEAT;
+      break;
+    case climate::CLIMATE_MODE_DRY:
+      remote_state[6] = MITSUBISHI_MODE_DRY;
+      remote_state[8] = MITSUBISHI_MODE_A_DRY;
       break;
+    case climate::CLIMATE_MODE_COOL:
+      remote_state[6] = MITSUBISHI_MODE_COOL;
+      remote_state[8] = MITSUBISHI_MODE_A_COOL;
```
Unfocused, line 126: I've been hitting a bug where setting it to COOL sets it to AUTO instead - here's why: missing a `break;`
```suggestion
      remote_state[8] = MITSUBISHI_MODE_A_COOL;
      break;
```

RubyBailey: Thanks for finding this. The PR has been updated.

esphome/components/mitsubishi/climate.py
```diff
@@ -9,13 +9,81 @@
 mitsubishi_ns = cg.esphome_ns.namespace("mitsubishi")
 MitsubishiClimate = mitsubishi_ns.class_("MitsubishiClimate", climate_ir.ClimateIR)
 
-CONFIG_SCHEMA = climate_ir.CLIMATE_IR_SCHEMA.extend(
+CONF_SET_FAN_SPEEDS = "set_fan_speeds"
+CONF_FAN_LOW = "fan_low"
+CONF_FAN_MEDIUM = "fan_medium"
+CONF_FAN_HI = "fan_hi"
```
kbx81: For consistency with other components, this should be "high" not "hi".

esphome/components/mitsubishi/climate.py
```diff
@@ -9,13 +9,81 @@
...
+}
+
+CONF_VERTICAL_DEFAULT = "vertical_default"
+VerticalDirections = mitsubishi_ns.enum("VerticalDirections")
+VERTICAL_DIRECTIONS = {
+    "auto": VerticalDirections.VERTICAL_DIRECTION_AUTO,
+    "up": VerticalDirections.VERTICAL_DIRECTION_UP,
+    "mup": VerticalDirections.VERTICAL_DIRECTION_MUP,
+    "middle": VerticalDirections.VERTICAL_DIRECTION_MIDDLE,
+    "mdown": VerticalDirections.VERTICAL_DIRECTION_MDOWN,
```
kbx81: I think it would be best to keep the convention here consistent with the horizontal directions as discussed earlier -- "middle-up" and "middle-down". 😄

jesserockz (dismissed):

kbx81 (commented):
Couple suggestions...

esphome/components/mitsubishi/climate.py
```diff
@@ -9,13 +9,81 @@
...
+
+CONFIG_SCHEMA = climate_ir.CLIMATE_IR_WITH_RECEIVER_SCHEMA.extend(
     {
         cv.GenerateID(): cv.declare_id(MitsubishiClimate),
+        cv.Optional(
+            CONF_SET_FAN_SPEEDS,
+            default={
+                CONF_FAN_LOW: "fan1",
+                CONF_FAN_MEDIUM: "fan2",
+                CONF_FAN_HI: "fan3",
```
Unfocused: The two different Mitsubishi heatpumps I'm trying to control both support 4 levels of fan speed. So does another Mitsubishi model I have elsewhere that I don't need to control, along with others I remember seeing in NZ - so 4 speeds seems common.

I realise there isn't a "Max" standard fan mode, but there are 4 other modes: Middle, Focus, Diffuse, and Quiet. Could you add config options for these too, as optional? Would make sense to allow assigning all 5 possible speeds the IR remote can do (fan1 - fan5). The Toshiba platform is an example that implements some of the extra modes.

Home Assistant & ESPHome also support custom fan modes (`set_supported_custom_fan_modes()`). If you wanted extra homework, a "Max" setting would be fitting (given 4 speeds seems common), but just doing the above 4 standard modes would IMO be adequate (& be easier to do).

esphome/components/mitsubishi/climate.py
```diff
@@ -9,13 +9,81 @@
...
+        ): cv.All(
+            cv.Schema(
+                {
+                    cv.Required(CONF_FAN_LOW): cv.enum(SETFANSPEEDS),
+                    cv.Required(CONF_FAN_MEDIUM): cv.enum(SETFANSPEEDS),
+                    cv.Required(CONF_FAN_HI): cv.enum(SETFANSPEEDS),
+                }
+            ),
+        ),
+        cv.Optional(CONF_HORIZONTAL_DEFAULT, default="middle"): cv.enum(
```
Unfocused, line 53: Of the 3 different Mitsubishi models I have, none of them support horizontal swing (only vertical swing). AFAIK, only Mitsubishi's higher end models support horizontal swing (at least in NZ). No idea if any models support only horizontal swing or not. 

Could a config option be added to disable either horizontal or vertical swing?

esphome/components/mitsubishi/climate.py
Unfocused: Additionally, having "fan_" in front of these seems redunadant.

```yaml
climate:
  - platform: mitsubishi
    name: "Heatpump"
    set_fan_speeds:
      fan_low: "fan2"
      fan_medium: "fan3"
      fan_high: "fan4"
```

Versus:

```yaml
climate:
  - platform: mitsubishi
    name: "Heatpump"
    set_fan_speeds:
      low: "fan2"
      medium: "fan3"
      high: "fan4"
```

esphome/components/mitsubishi/climate.py
RubyBailey: Wanted to let you know that I saw your comments. I’m away from my computer for a few weeks and will get back to you on these afterwards.

esphome/components/mitsubishi/climate.py
kbx81: Good point -- and then you can use the constants that are already in `const.py` 😄

esphome/components/mitsubishi/climate.py
RubyBailey: I don’t see why the options need to be disabled. All the units I have tested ignore the command if they don’t have vertical or horizontal vanes or swing capability.

Unfocused: Yes, sorry, I should have been clearer. Functionally it still works. However, it's exposed to Home Assistant as being supported, so the Home Assistant UI shows the additional option for a swing mode that isn't supported/does nothing.

esphome/components/mitsubishi/climate.py
RubyBailey: Hello, I'm back. 
Not quite sure how to approach this... Let me start with this.
1) Fan speeds: mine have 3 to 5, so I decided that being able to select 3 out of those was sufficient. 
2) Fan Modes (Middle, Focus, Diffuse, and Quiet): Mine do not have those modes, though we can work on adding them if we can work out the IR codes. In my experience Mitsubishi units can get "confused" if they unexpectedly receive settings in bytes 13 to 15 that are non-compatible modes. In other words unlike the horizontal or vertical vane settings (byte 8 and 9) they somehow cannot ignore those bytes. 
3) I did take a look at the (set_supported_custom_fan_modes()), just wasn't sure how to implement them without a breaking change. I agree it would be more configurable if people could select the number of fan speeds they want to set, I just wasn't sure that if was worth the effort to implement. I felt that in most cases AUTO plus 3 selectable fan speeds should be sufficient.
Your suggestions welcome.

esphome/components/mitsubishi/climate.py
RubyBailey: I don't use HA, so I'm not familiar with the HA UI.  How does one disable something from showing up in HA?
BTW, I have units that have vertical and horizontal swing and a couple that only have horizontal swing.

Unfocused: In Home Assistant, for most things you can pick and choose each thing you want added to the UI. But some things have more complex widgets that provide a UI for a complex device type - like climate devices. Unfortunately, it's all or nothing with that - it shows what the integration reports the device supports. So for the swing mode options for my heatpumps currently, it provides me with options: Off, Both, Vertical, Horizontal.

In terms of defining that swing modes are supported, use `set_supported_swing_modes`. The Haier platform is a good example:

https://github.com/esphome/esphome/blob/aeb94e166b456f1a21ae52c3a90c66266a75772b/esphome/components/haier/climate.py#L42-L43

esphome/components/mitsubishi/climate.py
Unfocused: Welcome back!

Sorry, I hadn't meant literally send a IR code for `quiet` as the fan mode, but allow mapping it to one of the existing supported fan modes. Just like you can map `high` to `fan3` or `fan4` etc.

I had been thinking you could just map key names under `set_fan_speeds` to any of the available items the `CLIMATE_FAN_MODES` list:
https://github.com/esphome/esphome/blob/aeb94e166b456f1a21ae52c3a90c66266a75772b/esphome/components/climate/__init__.py#L68-L80

And just like the current `low`, `medium`, & `high` options, let any of those be set to one of the values in `SETFANSPEEDS` (`fan1`, `fan2`, etc).

So in my case, the config could be:

```yaml
climate:
  - platform: mitsubishi
    name: "Heatpump"
    set_fan_speeds:
      quiet: "fan1"
      low: "fan2"
      medium: "fan3"
      high: "fan4"
```


I think the above would work good & cover most use cases well enough, without needing to implement custom fan modes. 

But if you did want to support custom fan modes, you could check the key names under the `set_fan_speeds` config key against the `CLIMATE_FAN_MODES` list. If any key names don't match anything in that list, then add it via `set_supported_custom_fan_modes()`.

That would allow a config like this:

```yaml
climate:
  - platform: mitsubishi
    name: "Heatpump"
    set_fan_speeds:
      quiet: "fan1"
      low: "fan2"
      medium: "fan3"
      high: "fan4"
      walrus: "fan5"
```

`walrus` isn't in the `CLIMATE_FAN_MODES` list, so it would be added via `set_supported_custom_fan_modes()`.

sleepinggiant5:
Thank you for all your hard work here!  Two quick items: first Ive noticed that using the github://pr#3886 external component, the "off" button does not work in home assistant, but works in the base esphome Mitsubishi component.  Second, if there is any way to implement the "quiet" fan mode that would be greatly appreciated.  I'm not programming inclined but if there is any way to help would love to.  Thanks again!

RubyBailey:
Hello @sleepinggiant5,
Thanks for the comments. 
I do not use HA, so I cannot comments on that system. When you say that it works in the base component, I assume that you mean the existing esphome Mitsubishi component, not the one in this PR? Do all the other controls, except "off", work for you? 
What model of Mitsubishi(s) do you have? 
I also don't have a Mitsubishi unit that has a quiet mode. Can you provide the IR codes required for that operation? See lines 48 to 67 of mitsubishi.cpp for more information. Thanks.

esphome/components/mitsubishi/climate.py
RubyBailey: I have updated the vertical and horizontal swing modes to optional.

esphome/components/mitsubishi/climate.py
RubyBailey: updated set_fan_speeds as suggested. Thanks

esphome/components/mitsubishi/climate.py
RubyBailey: Updated "m" to middle as suggested. Thanks

esphome/components/mitsubishi/climate.py
RubyBailey: Hi, I'm working on your suggestion. Apparently there are Mitsubishi units now available that have some, if not all, the settings as configured in the  allowed in the CLIMATE_FAN_MODES list. I'd like to not use the names in the list to allow for other valid modes in the future without a breaking change. 
I haven't yet figures out how to check the CLIMATE_FAN_MODES list and add a name via set_supported_custom_fan_modes(). 
But even if that were to work the challenge I haven't figured out is how to evaluate the received IR signal based on the three default and potentially one or two custom fan speeds in the "on_receive" section of he code.  
Still working on it.

esphome/components/mitsubishi/climate.py
RubyBailey: This has now been updated. Can you please test and let me know if it works for you? Thanks

RubyBailey:
@jesserockz Sorry for the direct ping, but I cannot seem to resolve the conflicts github thinks are there. They are not present anywhere in my code. How do I proceed, Suggestions welcome. Thanks

Update: NVM, Figured it out.

X-Ryl669:
Hi, I've posted #5856 without knowing about your work.

I've 2 different units in my home (MSZ-FH35VE and MSZ-SF25VE) and I'm using HA. So if it can help, I can test this code on mine. 

I don't have a IR receiver (or at least, I haven't enabled it if there's one), I'm using a Tuya's based WIFI IR gizmo.

So far, I've used the command and logic from ToniA's HeatPumpIR code to simplify and implement my changes and I've tested them on my unit.

esphome/components/mitsubishi/mitsubishi.cpp
```diff
@@ -22,37 +35,175 @@ const uint16_t MITSUBISHI_HEADER_MARK = 3500;
...
+  } else {
+    remote_state[7] = (uint8_t) roundf(
+        clamp<float>(this->target_temperature, MITSUBISHI_TEMP_MIN, MITSUBISHI_TEMP_MAX) - MITSUBISHI_TEMP_MIN);
+  }
+
+  // Wide Vane
+  switch (this->swing_mode) {
+    case climate::CLIMATE_SWING_HORIZONTAL:
+    case climate::CLIMATE_SWING_BOTH:
+      remote_state[8] = remote_state[8] | MITSUBISHI_WIDE_VANE_SWING;  // Wide Vane Swing
```
X-Ryl669, line 154: WIDE_VANE_SWING  (0xC0) is only horizontal swing for me. So if you set BOTH here, there's no vertical swing going on. Vertical swing is 0x78, but it must be set on the remote_state[9] OR'd with the fan speed/mode. On my unit it swing only the right curtain, but I don't know what is the code for both curtains.

RubyBailey: I have MFZ-KA18NA, MSZ-A24NA, MSZ-D36NA, MSZ-GE18NA units. I used "Raw-IR-decoder-for-Arduino" to decode the IR signal. ESPHome will do the same for you using an IR Receiver Diode and the IR Receiver component. 
I did update the Vertical Vane portion of the code, see [f0ea216]  (https://github.com/esphome/esphome/pull/3886/commits/f0ea216a1875ca8edbe384a801556872a88def83) above, but somehow github isn't accepting the change. If you have any suggestions how to resolve that issue it would help me a lot.

It includes:
  // Verical Vane
  switch (this->swing_mode) {
    case climate::CLIMATE_SWING_VERTICAL:
    case climate::CLIMATE_SWING_BOTH:
      remote_state[9] = remote_state[9] | MITSUBISHI_VERTICAL_VANE_SWING | MITSUBISHI_OTHERWISE;  // Vane Swing
      break;
    case climate::CLIMATE_SWING_OFF:
    default:
      remote_state[9] = remote_state[9] | this->default_vertical_direction_ | MITSUBISHI_OTHERWISE;  // Off--> vertical default position
      break;
  }

MITSUBISHI_VERTICAL_VANE_SWING is 0x38 and MITSUBISHI_OTHERWISE adds 0x40 for 0x78.

X-Ryl669: Ok, I understand better now. I'll try to fork your branch to fit my constants in yours. 

Your implementation is a lot more complete than mine on many points, there are only few missing features:
1. In my implementation and my unit, I've a FAN only mode (which is 0x38 on the byte 6). Since you don't have it (or maybe you have, I don't know), it means that it shouldn't be accessible by default and only be enabled in the YAML file (with an additional parameter: `support_fan_mode`?)
2. In my implementation, I have also "Quiet" mode for the fan, which is value 5 (to be written in byte 9). I think it's not very good to use different FAN enum than what ESPHome's climate is supporting, so I've mapped them to `climate::CLIMATE_FAN_AUTO, climate::CLIMATE_FAN_LOW, climate::CLIMATE_FAN_MIDDLE,
             climate::CLIMATE_FAN_MEDIUM, climate::CLIMATE_FAN_HIGH, climate::CLIMATE_FAN_QUIET` in this order. The `MIDDLE` doesn't fit very well, because HA display it after the **high** item in its list. Maybe it's ok to drop one possible value, like the MIDDLE one, but the Quiet mode is very useful at night.
3. I've tried to use the *IFeel (Econocool)* mode, it doesn't work. I've understood I need to write byte 14 yet writing 0x20 here doesn't work. I don't know what's wrong. I've read somewhere that it should be 0x23 but I haven't tested it (it's hard to see if the unit is using it).
4. Similarly, the *powerful* mode doesn't work either (supposedly writing something to byte 15). I guess another thing is missing. 
5. In all cases, I've implemented them as ESPHome's climate presets, and I think that's the way to go.

Would you be ok if I send you a patch to your branch with my changes so you can verify them and apply them? 
I'll close my PR, that doesn't make any sense to keep both.

What are the issue with Github you're experiencing ?

RubyBailey: I have been able to resolve the conflicts in github so this branch should now be as up-to-date as the PR.

1. FAN only. My units don't have that mode. Byte 6 typically also effects byte 8, see comments in mitsubishi.cpp lines 79 and 83. There should be a corresponding byte 8 setting.
2. Quiet = 5, interesting. There was a comment above that suggested adding more than the 3 default fan speed modes  climate::CLIMATE_FAN_LOW, climate::CLIMATE_FAN_MEDIUM, climate::CLIMATE_FAN_HIGH and using set_supported_custom_fan_modes() for additional fan mode setting. I've been working on that, though never considered using climate::CLIMATE_FAN_MIDDLE! Haha. That's why you'll find a set_supported_custom_fan_modes({CLIMATE_FAN_MEDIUM_LOW}). I've never really liked that approach and would rather use the built-in mode if possible, but then again didn't want to break any existing installs. I got that part working but stumbled on how to return a custom value with the IR Receiver code. My challenge is that the built-in fan modes are numeric values whereas the set_supported_custom_fan_modes() are strings. FYI, mitsubishi.cpp, line 20, already has MITSUBISHI_FAN_5 = 0x05. We could use that as the QUIET mode, though I've been told that some Mitsubishi units do have 5 speeds. Not sure of that, none of mine do.. 
3. and 4. I found some units don't like byte 14 and 15. When I set one (cannot remember which one) I could no longer turn one of my units OFF. I'd suggest not enabling any of those in code unless they are thoroughly tested.

Sure, go ahead and work on this branch. Agreed that both PR's don't make sense.

X-Ryl669: I've submitted a PR on your repository for my changes. Please have a look!

RubyBailey: Thanks for all the work. I cannot figure out how do I test your changes? I can see the files you changes in my github repository, but how do I get them into my esphome environment for testing? I have esphome running in VMWare and can see my repository, but not your PR.

X-Ryl669: Oh, sorry. To test them, add this to your YAML file:
```yaml
external_components:
  - source: github://pr#5867 
    components:
      - mitsubishi
    refresh: 1s

[...]
climate:
  - platform: mitsubishi       # adjust to match your AC unit!
    name: "Clim Salon"
    set_fan_mode: "quiet_4levels"
    set_operating_mode: "auto_dry_fan_heat_cool"
    horizontal_default: "middle"
    vertical_default: "down"    
    receiver_id: receiver
```

RubyBailey: I added the external components to my YAML file and it won't compile. Been struggling to figure out what I'm missing.

current YAML file:

```
substitutions:
  devicename: testhp
  upper_devicename: TestHP
  location: Test
  mqtt_retain: "False"

esphome:
  name: $devicename
  platform: ESP8266
  board: nodemcuv2

packages:
  wifi: !include common/wifi.yaml
  mqtt: !include common/mqtt.yaml  

external_components:
  - source: github://pr#5867 
    components:
      - mitsubishi
    refresh: 1s

logger:
  level: DEBUG

remote_transmitter:
  pin: GPIO15
  carrier_duty_percent: 50%
  
climate:
  - platform: mitsubishi
    name: "Heatpump"
    retain: False
    supports_cool: False
```

All attempts have the following error when compiling:
Compiling .pioenvs/testhp/src/esphome/components/mqtt/mqtt_climate.cpp.o
In file included from src/esphome/components/mitsubishi/mitsubishi.cpp:1:
src/esphome/components/mitsubishi/mitsubishi.h: In constructor 'esphome::mitsubishi::MitsubishiClimate::MitsubishiClimate()':
src/esphome/components/mitsubishi/mitsubishi.h:55:79: error: 'CLIMATE_FAN_QUIET' is not a member of 'esphome::climate'; did you mean 'CLIMATE_FAN_HIGH'?
   55 |              climate::CLIMATE_FAN_MEDIUM, climate::CLIMATE_FAN_HIGH, climate::CLIMATE_FAN_QUIET},
      |                                                                               ^~~~~~~~~~~~~~~~~
      |                                                                               CLIMATE_FAN_HIGH
...

X-Ryl669: Strange, the file https://github.com/esphome/esphome/blob/dev/esphome/components/climate/climate_mode.h contains the definition for `CLIMATE_FAN_QUIET` so it should find it (it does find the CLIMATE_FAN_HIGH which is defined at the same place). Did you made a change to your esphome's code folder?

RubyBailey: I agree. I have installed the latest esphome version per instructions at (https://esphome.io/guides/faq.html). Still the same error. Strange though that the dashboard reports "ESPHome by Nabu Casa | Fund development | v2022.9.4 Documentation", would have expected v2023...
I'll start fresh and create a new VM tomorrow. Let you know how things go.

RubyBailey: Created a new VM, installed esphome dev per guide, still unable to compile PR. Away for the next few days will get back to this on the weekend.

RubyBailey: Still no success. I am current unable to get past the ESPHome installation, on docker without errors. On the the Docker installation the ESPHome dashboard generates an error "Request not successful (500)" and I have no idea how to to get any further. Tried a number of things noted on google searches, none were successful.

X-Ryl669: Ok, what a shame it takes so many hours for basic stuff. I hope you haven't broken your own installation.

Please post your complete YAML file and I'll build a bin file for you from my local installation and post it here so you can test it.

KoenVanduffel:
I have used your thing for quite a while now. However compiling now fails with following errors:

In file included from src/esphome/components/mitsubishi/mitsubishi.cpp:1:
src/esphome/components/mitsubishi/mitsubishi.h:78:27: error: 'comedium_lownst' has not been declared
   bool parse_state_frame_(comedium_lownst uint8_t frame[]);
                           ^~~~~~~~~~~~~~~
src/esphome/components/mitsubishi/mitsubishi.h:78:51: error: expected ',' or '...' before 'frame'
   bool parse_state_frame_(comedium_lownst uint8_t frame[]);
                                                   ^~~~~
src/esphome/components/mitsubishi/mitsubishi.cpp:233:6: error: no declaration matches 'bool esphome::mitsubishi::MitsubishiClimate::parse_state_frame_(const uint8_t*)'
 bool MitsubishiClimate::parse_state_frame_(const uint8_t frame[]) { return false; }
      ^~~~~~~~~~~~~~~~~
In file included from src/esphome/components/mitsubishi/mitsubishi.cpp:1:
src/esphome/components/mitsubishi/mitsubishi.h:78:8: note: candidate is: 'bool esphome::mitsubishi::MitsubishiClimate::parse_state_frame_(int)'
   bool parse_state_frame_(comedium_lownst uint8_t frame[]);
        ^~~~~~~~~~~~~~~~~~
src/esphome/components/mitsubishi/mitsubishi.h:49:7: note: 'class esphome::mitsubishi::MitsubishiClimate' defined here
 class MitsubishiClimate : public climate_ir::ClimateIR {
       ^~~~~~~~~~~~~~~~~
In file included from src/esphome.h:44,
                 from src/main.cpp:3:
src/esphome/components/mitsubishi/mitsubishi.h:78:27: error: 'comedium_lownst' has not been declared
   bool parse_state_frame_(comedium_lownst uint8_t frame[]);
                           ^~~~~~~~~~~~~~~
src/esphome/components/mitsubishi/mitsubishi.h:78:51: error: expected ',' or '...' before 'frame'
   bool parse_state_frame_(comedium_lownst uint8_t frame[]);
                                                   ^~~~~
*** [.pioenvs/ac-koen/src/esphome/components/mitsubishi/mitsubishi.cpp.o] Error 1
*** [.pioenvs/ac-koen/src/main.cpp.o] Error 1

X-Ryl669:
The current PR doesn't build, there was a small mistake with Find & Replace. Change `comedium_lownst` to `const` in mitsubishi.h and it'll work. But if you don't use development version of esphome directly but the dashboard, can you confirm this PR is working for you (required change to YAML):
```yaml
external_components:
  - source: github://pr#5867 
    components:
      - mitsubishi
    refresh: 1s

[...]
climate:
  - platform: mitsubishi       # adjust to match your AC unit!
    name: "Clim Salon"
    set_fan_mode: "quiet_4levels"
    set_operating_mode: "auto_dry_fan_heat_cool"
    horizontal_default: "middle"
    vertical_default: "down"    
```

KoenVanduffel:
Hi X-Ryl669,
I was using your PR5867 as alternative in the meantime. However I did not manage to set the defaults. Also copying your code I can't set the defaults.

climate.mitsubishi: [source /config/esphome/parts/airco.yaml:15]
  platform: mitsubishi
  name: Koen AC
  id: koen_ac
  sensor: koen_temperatuur
  
  [set_fan_mode] is an invalid option for [climate.mitsubishi]. Please check the indentation.
  set_fan_mode: quiet_4levels
  
  [set_operating_mode] is an invalid option for [climate.mitsubishi]. Please check the indentation.
  set_operating_mode: auto_dry_fan_heat_cool
  
  [vertical_default] is an invalid option for [climate.mitsubishi]. Did you mean [disabled_by_default]?
  vertical_default: up
  visual: 
    temperature_step: 1

No idea whether I am doing something wrong (I commented out the horizontal _default as my units do not have a horizontal swing possibility)

```# Airco definition
external_components:
  - source: github://pr#5856
#   - source: github://pr#3886
    components: 
        - mitsubishi
    refresh: 1s

remote_transmitter:
  pin: GPIO17
  carrier_duty_percent: 50%

climate:
  - platform: mitsubishi # adjust to match your AC unit!
    name: "${upper_name} AC"
    id: "${name}_ac"
    sensor: ${name}_temperatuur
    set_fan_mode: "quiet_4levels"
    set_operating_mode: "auto_dry_fan_heat_cool"
    # horizontal_default: "middle"
    vertical_default: "up"
    visual:
      temperature_step: 1
```

X-Ryl669:
I think the error line is this:
```
external_components:
  - source: github://pr#5856
```

It should read:
```
external_components:
  - source: github://pr#5867
```

(And clean the build files before rebuilding)

The reason is because I've submitted a PR for mitsubishi first (5856), then I realized @RubyBailey had written a better version (3886) from where I bootstrapped again and improved up to (5867).

KoenVanduffel:
Superb!!! this works. I was stupid and only copied the part of your code **I thought** changed... Another little never take things for granted lesson :).
Thanks for the hard work both of you!

The only thing I don't quite understand are the fanmodes. At 5 levels I get auto low medium high and middle, and quiet instead of middle when chosing that fan_mode. Looking at the code middle seems to be in between low and medium which is also what I observe.
Quiet however gets a remote state 5 which is the same as high in the other modes. It must do something different though as it is seems to be the same as middle.

I do not get middle on the simple thermostat card however when using the 4+quiet option.
image

I further have no idea what non/boost/eco are (they are on the original remote too but seem to do nothing in my units model MSZ-HR25VF)

RubyBailey:
Hello @KoenVanduffel and @X-Ryl669,
I see you’ve been busy. My apologies for not responding earlier, been busy on some more urgent issues during the past few days. 
With respect to this PR and testing X-Ryl669’s additions I have had issues testing those changes. Thanks KoenVanduffel being the inadvertent test environment. 😀 I have been unable to set up a Esphome dev VM test environment (I don’t use HA). Probably something I’ve done, I’ll figure it out eventually. 
Get back to you when I do.

X-Ryl669:
The fan mode is an issue with HA. It doesn't have true 5 levels (only 3) so I had to map them somehow to what it has, so it can be used by automations.

In HA speak, there's only  Auto/Low/Medium/High or Quiet/Middle/Focus/Diffuse (which doesn't really mean a fan power, but mainly a fan mode).

Mitsubishi on the other hand reuse the fan level 5 for either High or Quiet (depending on the model).

I've tried to map the 3 levels to HA's levels (L/M/H) by skipping level 3 (for 4 levels' models) or 4 (for 5 levels' model) depending on your model and using FAN_MIDDLE as another intermediate level. Quiet should be only available if you allow it, so it means it's not working for your configuration, I'll have a look. 

The initial PR created custom fan levels, but it means that you can't use them in automations, since there are specific for this entity.

X-Ryl669:
boost is "POWERFUL" and eco is "ECONOCOOL" if you have that option (again, it's HA own preset I've mapped). I can't test on other models, I've implemented them for my own models, I don't know if they are mapped the same on yours.  

For your reference, if you can capture the IR string train and let the receiver code decode it, you post it here so I can fix it for you.

KoenVanduffel:
Hey @RubyBailey , no need to appologize, especially not to me. I am glad people are sharing their hard work (I have a bit of a similar situation where I designed a bunch of 3d printer parts but can't really follow up due to too much work at the moment).

@X-Ryl669 at least my remotes have the eco and boost modes. However I do not notice any difference when switching these on or off (with remote nor with HA) so probably the units do not support it. They just beep as they acknowledge recieving the complete control string.

Mapping of the fan levels clearly works. In the mode with quiet it is also quiet and the in 5 level mode the high mode gives a really powerfull blast exactly as the max level on the remote does. The remote does not have the quiet setting though.
From what I feel quiet falls in between low and middle (rather similar to middle) when the setpoint is quite a bit higher than actual temperature. When the setpoint is at actual temperature the fan runs even slower than low (subjective feeling I admit) and is fully unaudible. My impression is that quiet is similar to auto but uses lower fan speeds relative to the delta between setpoint and actual. In quiet mode the highest fan speed is probably middle. In auto it can go all the way to high. Can you confirm this subjective measurement?

X-Ryl669:
> Can you confirm this subjective measurement?

I'm afraid, I can't confirm, our units behave differently. 

Both my units have a quiet mode and 4 levels (only). In my units, auto is level 0, low is 1, up to high is 4 and quiet is 5 (I know, its stupid to have the highest value set to the lowest fan speed).
Ruby's unit, IIUC, has 5 levels but no quiet mode, so auto is 0, low is 1 up to high which is 5.

Yours seems to support both which is new to me. The code I've implemented does this:
* For 4Level: Low = 1, Medium = 3, High = 4
* For 5Level: Low = 1, Middle = 2, Medium = 3, High = 5
* For 4Level + Quiet: Low = 1, Middle = 2, Medium = 3, High = 4, Quiet = 5

So, if you change the YAML from `4levels` to `5levels`, the only change is the **high** value that's mapped from to speed 5 instead of 4. 

If you change `5levels` to  `quiet_4levels`, then the speed 5 is mapped to **quiet** and **high** is mapped to speed 4.

From your testimonial, it seems that you want the latter (like me), as **high** on your remote is actually speed 4. You'll gain a new feature (**quiet**) that your remote isn't supporting but your unit does.

In all mapping, **auto** is 0 and the unit does whatever it wants (and in my experience, that's the worst way to reach the target temperature, it never reach it and oscillate around).

**middle** is always fan speed 2. I have an issue with HA for the **middle** speed as it's not displayed in the usual fan speed selector widget since it's a mode not a speed in HA's speak. That's not something I can fix however, except by removing its declaration everywhere and never be able to reach speed 2 on the unit (which is probably the most logical move). 

In my experience, whatever the speed set on the unit, when it reaches the *temperature + some margin*, the fan speed decreases until it goes to the lowest possible speed. It'll ramp up at the set speed when below the *temperature - some margin*.

Just a silly question: Are you sure you have the good remote for your unit (not a replacement one)?

RubyBailey:
To clarify, my units only have 3 and 4 speed settings, no QUIET or other "fancy" settings.
A thought; to make things easier could we just remove the 5level option? That would mean Quiet = 5 for those who have those units and High = 4 always.

X-Ryl669:
Sure, good idea. 

How is the mapping for the 3 speeds setting to IR fan speed code? Is it 123 or 135? or something else ?

RubyBailey:
The existing active code maps 123 to fan speeds. 
This PR was designed to add flexibility for the user to optionally map any speed to low, medium and high. Initially 5 speeds, but now with Quiet always being 5, then only 4 speeds.
The IR side then, if it wasn't Auto=0, any value less then or equal to the low would map to LOW, anything between low and high to MEDIUM, and anything high and above to HIGH. 
5 will always be Quiet.

This may be obvious, but just to note, everything that we are doing with these code changes and through these PR's should be a non-breaking change to the current code in ESP Home.

X-Ryl669:
In my humble opinion, I don't think setting the fan speed one by one in YAML is the right level of interface for the user.
It's too low level and require the user to know about the quirks of the devices.

I think a good level is to ask the user to describe what she has on her remote. How it's implemented underneath is up to us, not her.

So in my implementation, we have a mode to select the different possibilities (3levels, 4levels, 5levels, 4+quiet, etc...) and the code makes the decision for the user about how to map the device's level to the restricted 3 levels (+quiet) that ESPHome's support. There's no way to have all features of the device supported (unless using custom fan mode, but this breaks automation since it becomes non generic), so there'll always be a loss here.

RubyBailey:
I see your point. That makes sense to me. Agreed. I didn't mean to say that we had to do it the way I started. Your work has added a lot to the functionality. What do you suggest without breaking the existing implementations? I still think we can remove the 5levels. One less mode to implement and support.

X-Ryl669:
Ok, I've made changes that should fit all our expectations. I've also merged the dev branch into my PR so it's up to date. 

There are now 4 modes:  *3levels*, *4levels*, *5levels*, *4levels+quiet*

The mapping is straightforward:
  * For 3Level: Low = 1, Medium = 2, High = 3
  * For 4Level: Low = 1, Middle = 2, Medium = 3, High = 4
  * For 5Level: Low = 1, Middle = 2, Medium = 3, High = 4
  * For 4Level + Quiet: Low = 1, Middle = 2, Medium = 3, High = 4, Quiet = 5

In effect, *5levels* is exactly like *4levels*, so for the user the 5th fan speed is not reachable. It's still there because there are remotes with 5 fan speeds (so there is no dilemma in what to choose in YAML). 

Let me know if it's ok for you.

RubyBailey:
@X-Ryl669 Hello again, I believed that I have now resolved my VM and esphome issues, the rebuild should now be completed. How do I now test your code changes? I don't know github well. I see your (Fix and check) PR on my (mitsubishi-update) PR to esphome. I can go to your github repository and see all the changes that you have done, though they don't show up in my PR. I just don't know how to download test the changes that you've made. Suggestions?

As an aside I also see github indicating that there are other, unrelated files, that have conflict in my repo, which will not allow me to merge your PR.
esphome/components/uart/uart_component_esp32_arduino.h
esphome/components/uart/uart_component_esp_idf.h
requirements.txt

X-Ryl669:
In order to test it, use this in your ESPHome's YAML
The pull request #5867 on ESPHome is the same branch as the one on your fork, so anything that's there is here too. 

```yaml
external_components:
  - source: github://pr#5867 
    components:
      - mitsubishi
    refresh: 1s

[...]
climate:
  - platform: mitsubishi       # adjust to match your AC unit!
    name: "Clim Salon"
    set_fan_mode: "quiet_4levels"
    set_operating_mode: "auto_dry_fan_heat_cool"
    horizontal_default: "middle"
    vertical_default: "down"    
```

I've sync'd and fixed the conflicts so it should pass all CI pipeline and be mergeable in ESPHome. 

Once you've tested them and if you want to merge them in *your* PR (this page), then click on "Merge Pull Request" button on this [page](https://github.com/RubyBailey/esphome/pull/25/)

They will appear on this page, run the CI again, and hopefully they will be ready to be merged in ESPHome by a maintainer.

If you want to make changes, (and if you've merged them in your fork), on your computer, you'll have to run a "git pull" while the `mitsubishi-updates` branch is active. git will fetch all of them on your computer so you can work on the code.

Hope it helps.

RubyBailey:
Your reply helped a lot. Thanks for your patience. 
I use MQTT for controlling the heat pumps and have tested a number of yaml configurations with my heat pumps. For the most part everything has worked well.
A couple of things for follow-up:
1. I have one unit that isn't turning off consistently. Not sure why; I'll do some further testing this weekend. 
2. supports_cool: False isn't supported in the code (see https://esphome.io/components/climate/climate_ir.html)
3. I still think 5level should be removed from the configuration options unless someone can test that configuration option prior to the PR being including in esphome. I cannot test that configuration. I'm good with all the other options.
4. 3level fan modes are inconsistent. The available fan_mode commands are low, medium and high, whereas  the ir_receiver returns low, middle and medium. I think that it should return low, medium and high.

KoenVanduffel:
I will give the 5 level setting another try and see whether my units pick it up. The remote has 5 levels + auto and these are all clearly different. As i reported before my remote does not have the quiet setting but the units do use it and I like the setting, the main difference is that even when the heating is actually off the fan keeps running very slowly which improves the quality of temperature controll. In auto mode the fan stops when no heating is active causing the temperature to drop quite a bit before restarting.

X-Ryl669:
You are right for your remarks. I haven't tested the receiver much because it doesn't work for my device (I'm using CB3S module from Tuya / Libretiny not an ESP and I don't know the pinout for the receiver). While I've implemented the Power/Eco mode, I've used a ZMote system (which is a ESP01 module with both IR LED and phototransistor). Yet this system isn't stable at all, it's getting too hot in 30s or so and then fails, so my attempts to capture all possible modes are very limited.

For bibliography, here are what I was able to capture:
<details>
<summary> Various Capture</summary>

```
Fan mode only
Quiet
Vert Auto
Horz Middle

[V][mitsubishi.climate:359]: Receiving: 23,CB,26,01,00,20,38,08,30,45,00,00,00,00,10,00,00,FA
@9 = 45 = 0x01 000 101
            HV POS SPD => 5 for Quiet, V is auto, H is set


Cool
Econocool
21°C
Quiet
Vert Auto
Horz Middle
[V]	[mitsubishi.climate:359] Receiving: 23,CB,26,01,00,20,58,05,36,45,00,00,00,00,30,00,00,3D
@6 = 58
@8 = 36
@14 = 30 


Cool
Powerful
21°C
Quiet
Vert Auto
Horz Middle
[V][mitsubishi.climate:359]: Receiving: 23,CB,26,01,00,20,58,05,36,40,00,00,00,00,10,08,00,20
Diff with Econocool: @15=08

Heat
Powerful
21°C
Quiet
Vert Auto
Horz Middle
[V][mitsubishi.climate:359]: Receiving: 23,CB,26,01,00,20,48,05,30,40,00,00,00,00,10,08,00,0A

```
</details>

I've removed the *5 levels* mode (just commenting it out so it can be re-enabled easily if some proud soul wants to).

Concerning the `supports_cool` mode, it's more or less deprecated in ClimateIR, I haven't found any example implementation using it (toshiba, midea, lg, ...). Anyway, I've reworked it so it's supported, but it kind of intersect with the much more complete `OperatingMode` property.

I've also fixed the receiver mode to deduce the fan level correctly. 

Please test again and if ok, feel free to merge it.

RubyBailey:
Hello, after Christmas and New Year's travels. I have been trying to test your most recent changes and am still importing the previous code with my test yaml. I'm unsure how to test your latest code, see further below.

My current test yaml, below, still compiles the previous version of your code, e.g. not the changes that you noted above; even with a clean build. 

I added "set_fan_mode: 5levels" into the yaml below only as a test to confirm that it's not there, yet still it compiles. 

```substitutions:
  devicename: testhp
  upper_devicename: TestHP
  location: Test
  mqtt_retain: "False"

esphome:
  name: $devicename
  platform: ESP8266
  board: nodemcuv2

packages:
  wifi: !include common/wifi.yaml
  mqtt: !include common/mqtt.yaml  

external_components:
  - source: github://pr#5867 
    components:
      - mitsubishi
    refresh: 1s

logger:
  level: DEBUG

# Configure mitsubishi control
remote_transmitter:
  pin: GPIO15
  carrier_duty_percent: 50%
  
climate:
  - platform: mitsubishi
    name: "Heatpump"
    receiver_id: rcvr
    retain: False
    supports_cool: False
    set_fan_mode: 5levels

remote_receiver:
  id: rcvr
  pin:
    number: GPIO14
    inverted: True
    mode:
      input: True
      pullup: true
  tolerance: 55%


```

X-Ryl669:
Hi!

  I wish you a happy new year and all the best for you and your family. The PR that's linked above was closed before I added my changes, so I wonder it's not updating its head to the head of the fork'd branch.

If I understand ESPHome's YAML correctly, this means that you need to change:
```yaml
external_components:
  - source: github://X-Ryl669/esphome@fixAndCheck 
    components:
      - mitsubishi
    refresh: 1s
```

Hope it helps!

RubyBailey:
Hello,
I also wish you and your family a Merry Christmas and all the best in 2024. Thanks again for your help.
Your suggested link works, thanks.
I'm not sure how to make suggested changes directly in your files, so I'll take the lengthy route and put the suggestions in these comments:

climate.py

1.  `"5levels": SetFanMode.MITSUBISHI_FAN_5L,` needs to be commented out or removed, otherwise the esphome code validator will allow the mode, yet it will not compile.
2. `SETOPERATINGMODE` has 3 options (`auto_heat_cool`, `auto_dry_heat_cool`, `auto_dry_fan_heat_cool`, whereas mitsubishi.h lists five (5). These also include `auto_heat` and `auto_cool` options which the code validator current rejects. Based on your earlier comment that "the supports_cool mode, it's more or less deprecated in ClimateIR" I'm wondering if we should remove the supports_heat and supports_cool checks and use only the 5 operating modes you coded. Straight forward and elegant.
3. Further on that thought "auto" seems unnecessary in those options and could be shortened to `heat`, `cool`, `heat_cool` , `heat_cool_dry`, `heat_cool_dry_fan`. I think that would be more aligned with the IR_RECEIVER which returns the modes HEAT, COOL, HEAT_COOL and DRY (that's with my remotes; I don't have fan only mode to test)
4. I'd also suggest changing the word "operating" to "supported" in `SETOPERATINGMODE`, `set_operating_mode`, `CONF_SET_OPERATING_MODE`, `SETOPERATINGMODE` as well as in mitsubishi.h

mitsubishi.h

1. Per 3. above change "operating" to "supported" in `SetOperatingMode operating_mode_;` as well as in mitsubishi.cpp

mitubishi.cpp

1. Where `this->operating_mode_` is used to set `traits.add_supported_mode`. This would then change the .yaml file setting to `set_supported_mode:` which I think is more descriptive and better in the documentation.
2. It appears to me that `MITSUBISHI_FAN_4L` isn't checked and doesn't impact anything. I think `CLIMATE_FAN_MIDDLE` would be added with `MITSUBISHI_FAN_4L`. The comment also says defaults to 3 levels whereas in climate.py the default is `4levels`. I agree that `3levels` [low, medium, high and auto] would be the appropriate default.

```
// Default to only 3 levels in ESPHome even if most unit supports 4. The 3rd level is not used.
traits.set_supported_fan_modes(
{climate::CLIMATE_FAN_AUTO, climate::CLIMATE_FAN_LOW, climate::CLIMATE_FAN_MEDIUM, climate::CLIMATE_FAN_HIGH});
if (this->fan_mode_ == MITSUBISHI_FAN_Q4L)
traits.add_supported_fan_mode(climate::CLIMATE_FAN_QUIET);
if (/*this->fan_mode_ == MITSUBISHI_FAN_5L ||*/ this->fan_mode_ == MITSUBISHI_FAN_Q4L)
traits.add_supported_fan_mode(climate::CLIMATE_FAN_MIDDLE);  // Shouldn't be used for this but it helps
```

I should also ask how do I merge all these great changes that you've done into the open PR that I have in esphome? I still need to update the documentation section to align with all these changes.

Thanks again

X-Ryl669:
> I'm not sure how to make suggested changes directly in your files, so I'll take the lengthy route and put the suggestions in these comments:

The best way is to review [here](https://github.com/RubyBailey/esphome/pull/25) (in your ESPHome project's fork)

I'll make the change tonight and let you know there. You'll be able to merge the PR on your fork/branch, and it'll automatically update this PR once your branch is updated.

It could be a good idea to enable the workflow to run for this PR on your fork (see previous page and click to enable them) so I can check that they'll pass the CI tests here too.

RubyBailey:
Hello, I had some questions and added a couple of suggestions to the https://github.com/X-Ryl669/esphome/tree/removeSetOperatingMode fork a few days ago. Just wondering if your received those? No rush, just checking that I used the right process. Thanks

X-Ryl669:
No, I didn't get any notification for those (or I already answered). I've modified the code to account your remarks, you were correct and I think the code is clearer this way too. You can review the PR [there](https://github.com/RubyBailey/esphome/pull/26) and if you think it's ok, merge it (and you'll probably need to merge the dev branch afterward to track it).

RubyBailey:
Could it be that your latest PR was into my dev branch and not mitsubishi-updates? I had added some comments to the PR (those that you didn't get), a couple of minor things that I decided they weren't that important, deleted the comments and then merged your changes. Now your changes show up in my dev branch.  I'm now trying to determine if I can rollback those changes and somehow integrate them into to the mitsubishi-updates branch. Working on it.

RubyBailey:
Or I screwed up and merged them into dev instead of mitsubishi-updates.

RubyBailey:
I think everything is now resolved. Should be up-to-date.

X-Ryl669:
@jesserockz I think this PR is ready for review and merging. Can you take a look please ?

RubyBailey:
Hello, @X-Ryl669 and @KoenVanduffel,
I received a PR from @OlympusMonds on my fork to add "nightmode" as CLIMATE_PRESET_SLEEP to this PR. I have compiled and tested the PR on my heat pumps to ensure that there are no unexpected side-effect, but cannot test the actual mode because none of my heat pumps have that option. The one thing I found was that when CLIMATE_PRESET_SLEEP was selected that the heat pump's auditable acknowledgement (the beep) was silenced for all other commands. The commands still work. The beep would return when CLIMATE_PRESET_NONE was selected. Would you have any units that this new Preset could be tested on and, if so, would you be able to test the changes? Comments and suggestions, as always, appreciated. 

Thanks.

The PR with @OlympusMonds' changes is available for testing as an external component at:
```
external_components:
  - source: github://OlympusMonds/esphome@nightmode
    components:
      - mitsubishi
    refresh: 1s
```

P.S. Still looking forward to this PR eventually getting merged into ESPHome by the maintainers. I know they are very busy. It will just take some time and patience. :-)

KoenVanduffel:
Hi @RubyBailey i have to check whether my units support "nightmode", hope they do as the beeps do annoy me. Will test tomorrow.

OlympusMonds:
If it helps, this is the entry on Night Mode from my manual:
night_mode_manual

It stops the beeping, makes the LED less bright, and apparently makes the outdoor unit quieter.

RubyBailey:
Hello @OlympusMonds, That's interesting. My units are older ones that do not have Night Mode as an option. On my units it disables the beep sound for all commands, even when the unit is stopped or started. I didn't check if it does anything to the indicator light or the outdoor unit.
Thanks.

KoenVanduffel:
Hi, my units do not support sleep mode on their remotes, however as reported above setting the unit to sleep mode with ESPHome disables the beeper indeed (very welcome).
Whether the power output is any different from the other modes is hard to say, at least I don't notice a difference, the room heats up just as quickly as in none mode. Noise wise my outside units are on the roof and inaudible in any case (without climbing on the roof).

I do notice that in sleep mode the swing position is not followed. The unit I tested has "up" default position which is what all modes do use except sleep mode. It sets the flap in up position when idling, however as soon as the unit starts working the flow goes in down position (I think actually it is mdown but that is hard to see). Also when setting the swingmode to vertical the flap stays in mdown position (my units do not support horizontal swing).

About the supported modes my remote: it supports eco mode and none only. Eco indeed turns down the power significantly. This works both with the remote and ESPHome. Boost and sleep seem to be identical to none in terms of power output.

esphome/components/mitsubishi/climate.py
```diff
@@ -9,13 +9,69 @@
...
+        cg.add(var.set_fan_mode(config[CONF_SET_FAN_MODE]))
+
+    cg.add(var.set_supports_dry(config[CONF_SUPPORTS_DRY]))
+    cg.add(var.set_supports_fan_only(config[CONF_SUPPORTS_FAN_ONLY]))
+
+    if CONF_HORIZONTAL_DEFAULT in config:
+        cg.add(var.set_horizontal_default(config[CONF_HORIZONTAL_DEFAULT]))
+
+    if CONF_VERTICAL_DEFAULT in config:
+        cg.add(var.set_vertical_default(config[CONF_VERTICAL_DEFAULT]))
```
jesserockz: Since there are defaults, they will always be there
```suggestion
    cg.add(var.set_fan_mode(config[CONF_SET_FAN_MODE]))

    cg.add(var.set_supports_dry(config[CONF_SUPPORTS_DRY]))
    cg.add(var.set_supports_fan_only(config[CONF_SUPPORTS_FAN_ONLY]))

    cg.add(var.set_horizontal_default(config[CONF_HORIZONTAL_DEFAULT]))

    cg.add(var.set_vertical_default(config[CONF_VERTICAL_DEFAULT]))
```

esphome/components/mitsubishi/mitsubishi.cpp
```diff
@@ -81,5 +268,113 @@ void MitsubishiClimate::transmit_state() {
...
+      } else {
+        this->swing_mode = climate::CLIMATE_SWING_VERTICAL;
+      }
+      break;
+  }
+
+  ESP_LOGV(TAG, "Receiving: %02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X,%02X",
+           state_frame[0], state_frame[1], state_frame[2], state_frame[3], state_frame[4], state_frame[5],
+           state_frame[6], state_frame[7], state_frame[8], state_frame[9], state_frame[10], state_frame[11],
+           state_frame[12], state_frame[13], state_frame[14], state_frame[15], state_frame[16], state_frame[17]);
```
jesserockz: We have `format_hex_pretty` in `helpers.h` for this.

jesserockz (changes requested):

esphome/components/mitsubishi/mitsubishi.cpp
RubyBailey: I have tried a number of different approaches, searched much of the code in esphome, yet none seem to use format_hex_pretty with an array same as in this code. I have not been able to figure out how to use format_hex_pretty to format this array. Will try again tomorrow.

X-Ryl669: I think what @jesserockz was referring to is more like this:
```cpp
std::string received_fmt = format_hex(state_frame, 18);
ESP_LOGV(TAG, "Receiving: %.*s", received_fmt.length(), received_fmt.data());
```

It's not exacted equivalent in the output (no space between hex char), like this `Receiving: 020AEFEF02...` 
We can also use `format_hex_pretty` the same way to get `Receiving: 02.0A.EF.EF.02...(18)`

RubyBailey: Hello, I was trying to use ```format_hex_pretty``` but for whatever reason I just couldn't get it to display the received data in a readable format. The brain just wasn't in gear. Got it working today. Code  update. Thanks.

RubyBailey:
One last thing to do before submitting this for review (again ;-)
Working with @OlympusMonds to resolve one item in on_receive with Presets in the "nightmode" PR.

jesserockz (approved):

PR #3887, opened 2022-10-08
allow to use MQTT for discovery of IPs if mDNS is no option 

Links2004:

The changes allows the user to use MQTT as a alternative to mDNS  for discovery of IPs / ESPhome devices.

This is helpful when for example:
 - the CLI/dashboard is running not on the same network segment the the ESPhome devices and there is only access via routing
 - the CLI/dashboard is running in kubernetes
 - the host the has no mDNS support
 
## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [x] ESP8266

```yaml
# Example config.yaml
esphome:
  name: esphome-test

esp32:
  board: esp32dev
  framework:
    type: arduino

# Enable logging
logger:

# Enable Home Assistant API
api:
  encryption:
    key: "xxxxx"

ota:
  password: "yyyy"

mqtt:
  broker: mqtt.demo.com
  username: esphome
  discovery_retain: False
  discovery: False

wifi:
  ssid: !secret wifi_ssid

mdns:
  disabled: True
```

## Checklist:


Links2004:
docker image for easy testing:
ghcr.io/links2004/esphome/amd64:g8efb7594.mqtt_discover

esphome/__main__.py
```diff
@@ -287,17 +290,31 @@ def upload_program(config, args, host):
...
+
     ota_conf = config[CONF_OTA]
     remote_port = ota_conf[CONF_PORT]
     password = ota_conf.get(CONF_PASSWORD, "")
+
+    mdns_disabled = (
+        CONF_MDNS in config
+        and CONF_DISABLED in config[CONF_MDNS]
+        and config[CONF_MDNS][CONF_DISABLED] is True
+    )
```
jesserockz: Please test and confirm, but `mdns` should be auto loaded if `wifi` or `ethernet` is in the config and `CONF_DISABLED` defaults to `False` so this structure should always be in the config. 
If `wifi` or `ethernet` are not there then this would fail above on `CONF_OTA not in config` so should be safe.

```suggestion
    mdns_disabled = config[CONF_MDNS][CONF_DISABLED]
```

Links2004: you are correct, I simplify-ed the code.

and also found a other bug with the mqtt client id with haveing 2 dashboards on the same MQTT server. (which is fixed not too)
had my changes running in kubernetes and the testing on my computer.

Links2004:
Any change of some feedback or a merge?

Links2004:
Any change of some feedback or a merge?

esphome/components/mqtt/mqtt_client.cpp
```diff
@@ -14,6 +15,14 @@
 #include "lwip/dns.h"
 #include "mqtt_component.h"
 
+#include "esphome/components/wifi/wifi_component.h"
```
jesserockz: wifi is not guaranteed to be availabel, could be using ethernet.
Use the `network` component to get the ip address below.

esphome/mqtt.py
```diff
@@ -89,11 +120,75 @@ def on_disconnect(client, userdata, result_code):
...
+    def on_connect(client, userdata, flags, return_code):
+        _LOGGER.info("Send discover via MQTT broker")
+        client.publish("esphome/discover", None, retain=False)
+
+    return initialize(
+        config, [topic], on_message, on_connect, username, password, client_id
+    )
+
+
+def get_esphome_device_ip(config, username=None, password=None, client_id=None):
```
jesserockz: This really should use async/await, not just rely on the mqtt loop_forever ending, which could happen if the device never comes online?

jesserockz (changes requested):

esphome/components/mqtt/mqtt_client.cpp
Links2004: changed

esphome/mqtt.py
Links2004: added a timeout

Links2004:
@jesserockz  any change of a merge ?

jesserockz (approved):
Hi @Links2004, I have pushed some minor changes; 
- add a message to the cli prompt telling you what the choice is for
- removed `MQTT` as an upload target as its not really the target, and the code you added already uses the mqtt code to get the ip (Someone can still add `--device MQTT` to bypass the choices)
- another change that reduces the timeout to 10 seconds, as making people wait 2 mins is way too long. The device is either online or offline and will respond quickly if online.

Let me know what you think. I have tested these changes locally and I am happy with the PR in its state now.

Jesse

Links2004:
Hi @jesserockz,

thanks for taking the time to look at it.
I tested your changes in my setup and for ESP32 based devices its working,
but for ESP8266's I run in the timeout after the upload.
most likely that the ESP8266 needs more then the 10 sec to reboot and reconnect to MQTT.

```
Uploading: [============================================================] 100% Done...

INFO Waiting for result...
INFO OTA successful
INFO Successfully uploaded program.
INFO Starting looking for IP in topic esphome/discover/flexispot-desk-control
INFO Connected to MQTT broker!
INFO Send discover via MQTT broker topic: esphome/ping/flexispot-desk-control
ERROR Failed to find IP via MQTT
```

Links2004:
did some testing and 25sec works every time for me.

jesserockz (approved):
25 is fine =)

LGTM

PR #3891, opened 2022-10-08
Exposing coordinates from touchscreen binary sensor

gpambrozio:
I’ll need this for a project where I use these coordinates to draw elements on the screen and not have to repeat numbers everywhere

Exposes coordinates from touchscreen binary sensor

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

Not a full example but this is why I need this for.

Say I have this in my config.yaml:

```
image:
  - file: "images/fanblades.png"
    id: fanblades
    type: TRANSPARENT_BINARY

binary_sensor:
  - platform: touchscreen
    id: button_fan_main
    name: Fan Button
    internal: true
    x_min: 0
    x_max: 100
    y_min: 50
    y_max: 150
    on_click:
        ...
```

I then include a .h file with this code:

```
#include "esphome.h"

void drawButton(
  DisplayBuffer *it,
  TouchscreenBinarySensor *sensor,
  Image *image,
  int16_t x_offset = 0,
  int16_t y_offset = 0
) {
  int cx = sensor->get_x_min() + x_offset + (sensor->get_width() - image->get_width()) / 2;
  int cy = sensor->get_y_min() + y_offset + (sensor->get_height() - image->get_height()) / 2;
  it->image(cx, cy, image);
}
```

Then in my display's `lambda` I can do this:
```
    lambda: |-
      drawButton(&it, id(button_fan_main), id(fanblades));
``` 

## Checklist:

### DIFF
--- a/esphome/components/touchscreen/binary_sensor/touchscreen_binary_sensor.h
+++ b/esphome/components/touchscreen/binary_sensor/touchscreen_binary_sensor.h
@@ -23,6 +23,12 @@ class TouchscreenBinarySensor : public binary_sensor::BinarySensor,
     this->y_min_ = y_min;
     this->y_max_ = y_max;
   }
+  int16_t get_x_min() { return this->x_min_; }
+  int16_t get_x_max() { return this->x_max_; }
+  int16_t get_y_min() { return this->y_min_; }
+  int16_t get_y_max() { return this->y_max_; }
+  int16_t get_width() { return this->x_max_ - this->x_min_; }
+  int16_t get_height() { return this->y_max_ - this->y_min_; }
 
   void set_page(display::DisplayPage *page) { this->page_ = page; }
 


nielsnl68:
hi @gpambrozio  did you saw my work on the switchplate component? would you have time to check it out and see of it will work for you?

gpambrozio:
@nielsnl68 yes, I did and was trying it out a bit but it looks like it still can't do anything other than changing pages when you touch a button. Or maybe I don't understand how to configure it yet. But looks like the automation source is still empty so.

And for my use this change makes it way easier to use. But I look forward to be able to use switchplate in the future

nielsnl68:
> @nielsnl68 yes, I did and was trying it out a bit but it looks like it still can't do anything other than changing pages when you touch a button. Or maybe I don't understand how to configure it yet. But looks like the automation source is still empty so.

Oww good that you point this out. I found out that i forget to add to the documentation. Yes, there is a way already to "talk" between the switchplate and ESPHome and Home-Assistant using the switch's `platform: switchplate` .

```yaml
switch:
  - platform: switchplate
    widget_id: rgbLamp
    name:  RGB lightball
```

jesserockz (approved):
LGTM

PR #3903, opened 2022-10-13
Copy and adapt files from hrxl_maxsonar_wr to add support #1913

### LINKED ISSUE #1913: add support for XL-MaxSonar-WR/WRC MaxBotix sensors
Hello Community =) 

ESPHome already supports the high resolution (hr) versions of the maxbotix distance sensors, but it would be nice to also support other models. Specifically, the XL-MaxSonar-WR series with model numbers MB70XX and MB71XX as given in this datasheet: https://www.maxbotix.com/documents/XL-MaxSonar-WR_Datasheet.pdf

Got my hands on a MB7052 and using a MAX232 converter to convert RS232 to TTL. It all works, the esp8266 receives bytes without a problem, but the provided sensor platform for hrxl_maxsonar_wr will only log "Invalid data read from sensor: %s"

I found a workaround using a text sensor to read the bytes and convert them to int in HA downstream, but this solution lacks nice features like filters and throttling (which are very much desired at a publishing frequency of 6.6 Hz). 

**Additional context**

Looking at the source code, I found that the control flow requires a buffer length of 6 for values to get published. However, the lower precision versions will only send bytes with a total length of 5, e.g. 'R012\r' and not 'R0123\r'. As a result, the values are never published because the if condition never evaluates to true. 

I tried to tweak the hrxl_maxbotix_wr.cpp file, but couldn't find it. Simply replacing the value in line 17 would suffice, I guess. 
 `static const int MAX_DATA_LENGTH_BYTES = 5; // priviously 6`
Since this would also help my situation, I'd be grateful for advice on how to locate esphome source files in my HA installation.

Worschi:

Implements https://github.com/esphome/feature-requests/issues/1913: add support for XL-MaxSonar-WR/WRC MaxBotix sensors
Basically, the same code as in hrxl_maxsonar_wr component, but one constant value changed and updated namespaces.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/1913

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266

```yaml

logger:
  level: DEBUG
  baud_rate: 0 # disable logging over uartl

uart:
  id: mb_uart_bus
  tx_pin: TX
  rx_pin: RX
  baud_rate: 9600
  debug:
    after:
      delimiter: "\r"
    sequence:
      - lambda: UARTDebug::log_string(direction, bytes);

sensor:
  - platform: "xl_maxsonar_wr"
    name: "Septic Tank"
    # Tweak the filters for your application
    filters:
          window_size: 10
          send_every: 10
      - or:
        - throttle: "10sec"
        - delta: 0.02

```

## Checklist:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -94,7 +94,7 @@ esphome/components/heatpumpir/* @rob-deutsch
 esphome/components/hitachi_ac424/* @sourabhjaiswal
 esphome/components/homeassistant/* @OttoWinter
 esphome/components/honeywellabp/* @RubyBailey
-esphome/components/hrxl_maxsonar_wr/* @netmikey
+esphome/components/hrxl_maxsonar_wr/* @Worschi @netmikey
 esphome/components/hydreon_rgxx/* @functionpointer
 esphome/components/i2c/* @esphome/core
 esphome/components/i2s_audio/* @jesserockz
@@ -258,4 +258,5 @@ esphome/components/xiaomi_lywsd03mmc/* @ahpohl
 esphome/components/xiaomi_mhoc303/* @drug123
 esphome/components/xiaomi_mhoc401/* @vevsvevs
 esphome/components/xiaomi_rtcgq02lm/* @jesserockz
+esphome/components/xl_maxsonar_wr/* @netmikey
 esphome/components/xpt2046/* @nielsnl68 @numo68
--- a/esphome/components/xl_maxsonar_wr/sensor.py
+++ b/esphome/components/xl_maxsonar_wr/sensor.py
@@ -1,0 +1,29 @@
+import esphome.codegen as cg
+from esphome.components import sensor, uart
+from esphome.const import (
+    STATE_CLASS_MEASUREMENT,
+    UNIT_METER,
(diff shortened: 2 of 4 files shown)


jesserockz (changes requested):
Duplicating a whole component like this for one const change is not accepted.

The best thing to do here is add a `model` field to the existing component and change the constant based on that config variable provided by the user.

Worschi:
Alright, I see. Already thought about the redundancy...
If that's the case, I'll need to wait for somebody to implement this.

GreyMark7650:
> Alright, I see. Already thought about the redundancy... If that's the case, I'll need to wait for somebody to implement this.

That is a shame, I too just came across this as an issue.

Worschi:
Closing this, since the idea of this pull request was to add another sensor platform altogether, which is not desirable due to introduced redundancies. The way to go has been outlined by @jesserockz: 

> The best thing to do here is add a `model` field to the existing component and change the constant based on that config variable provided by the user.

PR #3905, opened 2022-10-13
Remove address type map from bluetooth proxy

jesserockz:

It seems this map was leaking memory. So I have removed it. 
This would mean its a little bit slower for active connection to connect as they need to wait for an advertisement with the type to come in live but should not be too bad.

Closes #3906

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
external_component:
  - source: github://pr#3905
    components: buetooth_proxy
```

## Checklist:


tomi19921702:
Screenshot_20221014-125125
Works great for me thanks @jesserockz when other data is needed just let me know.  You can see the data behavious better then before.

bearpawmaxim:
Hi @jesserockz! Maybe the circular buffer should be used to store the fixed amount of address=type associations ?

PR #3911, opened 2022-10-16
Add API interface to request a complete device config as JSON.

RoboMagus:

Currently there is no interface for a.o. the Dashboard to get a _complete_ device configuration. I.e. that includes secrets, packages, etc.

IMO this is the core problem for issues like esphome/issues#3697.
Exposing this API is the initial step in fixing the issue above. (Not a complete fix!)

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** ~fixes~ esphome/issues#3697.

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -932,10 +932,11 @@ class JsonConfigRequestHandler(BaseHandler):
     @bind_config
     def get(self, configuration=None):
         filename = settings.rel_path(configuration)
-        content = ""
-        if os.path.isfile(filename):
-            content = yaml_util.load_yaml(filename, clear_secrets=False)
+        if not os.path.isfile(filename):
+            self.send_error(404)
+            return
 
+        content = yaml_util.load_yaml(filename, clear_secrets=False)
         self.set_header("content-type", "application/json")
         self.write(json.dumps(content))
 
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -938,7 +938,7 @@ class JsonConfigRequestHandler(BaseHandler):
 
         try:
             content = yaml_util.load_yaml(filename, clear_secrets=False)
-            json_content = json.dumps(content, default=lambda o: f"{repr(o)}")
+            json_content = json.dumps(content, default=lambda o: {"__type": str(type(o)), "repr": repr(o)})
             self.set_header("content-type", "application/json")
             self.write(json_content)
         except Exception as err:  # pylint: disable=broad-except
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -936,9 +936,14 @@ class JsonConfigRequestHandler(BaseHandler):
             self.send_error(404)
             return
 
-        content = yaml_util.load_yaml(filename, clear_secrets=False)
-        self.set_header("content-type", "application/json")
-        self.write(json.dumps(content))
+        try:
+            content = yaml_util.load_yaml(filename, clear_secrets=False)
+            json_content = json.dumps(content, default=lambda o: f"{repr(o)}")
+            self.set_header("content-type", "application/json")
+            self.write(json_content)
+        except Exception as err:  # pylint: disable=broad-except
+            _LOGGER.warning("Error translating file %s to JSON: %s", filename, err)
+            self.send_error(500)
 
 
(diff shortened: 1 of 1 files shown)


esphome/dashboard/dashboard.py
```diff
@@ -927,6 +927,19 @@ def get(self):
         self.write(json.dumps(secret_keys))
 
 
+class JsonConfigRequestHandler(BaseHandler):
+    @authenticated
+    @bind_config
+    def get(self, configuration=None):
+        filename = settings.rel_path(configuration)
+        content = ""
+        if os.path.isfile(filename):
```
balloob: We should throw a 404 if it can’t find it.

esphome/dashboard/dashboard.py
```diff
@@ -927,6 +927,20 @@ def get(self):
...
+class JsonConfigRequestHandler(BaseHandler):
+    @authenticated
+    @bind_config
+    def get(self, configuration=None):
+        filename = settings.rel_path(configuration)
+        if not os.path.isfile(filename):
+            self.send_error(404)
+            return
+
+        content = yaml_util.load_yaml(filename, clear_secrets=False)
```
glmnet: Even though this does not validate the ESPHome config it still validates a valid yaml and hence can throw exceptions. So those should be handled here. 

I'm not sure if it will be better to try to fully validate the config here and return also the validation result. Might come handy later or at least wrap the output in a config key so "errors" or something else can be added later

RoboMagus: I've added exception handling for when the yaml file cannot be converted to json.

> I'm not sure if it will be better to try to fully validate the config here and return also the validation result. Might come handy later or at least wrap the output in a config key so "errors" or something else can be added later

Fully validating is not the purpose of this addition to the API. We'd like to be able to retrieve the contents, regardless of validity. 
There is already a websocket interface for validation, but if we'd have a usecase for a REST interface for config file validation I would still propose creating a separate one for that.

RoboMagus: Though not as likely as occurences of invalid yaml in esphome device configs, the `secret_keys` also contains no proper handling of yaml errors when loading the file: `secret_keys = list(yaml_util.load_yaml(filename, clear_secrets=False))`.

Having e.g. a `;` i.s.o. `:` as the mapping separator for one of the secrets, an exception is thrown.

Exception handling should probably also be added there.

esphome/dashboard/dashboard.py
```diff
@@ -927,6 +927,25 @@ def get(self):
...
+    @bind_config
+    def get(self, configuration=None):
+        filename = settings.rel_path(configuration)
+        if not os.path.isfile(filename):
+            self.send_error(404)
+            return
+
+        try:
+            content = yaml_util.load_yaml(filename, clear_secrets=False)
+            json_content = json.dumps(content, default=lambda o: f"{repr(o)}")
```
balloob: What is the default keyword doing here. What objects can it not serialize?

esphome/dashboard/dashboard.py
```diff
@@ -927,6 +927,25 @@ def get(self):
...
+            self.set_header("content-type", "application/json")
+            self.write(json_content)
+        except Exception as err:  # pylint: disable=broad-except
```
balloob: Do not catch generic exceptions. Instead write out all exceptions that are expected.

esphome/dashboard/dashboard.py
RoboMagus: The one's that I've found during testing `!lambda` blocks don't serialize. Unsure if there are others, but having a default and thus prevent producing JSON output seems like a nice to have fallback.

balloob: I suggest we go with a more compatible type, that can be extended in the future. 
```suggestion
            json_content = json.dumps(content, default=lambda o: {"__type": str(type(o)), "repr": repr(o)})
```

([copied from HA](https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/json.py#L55-L60))

esphome/dashboard/dashboard.py
RoboMagus: yaml loading should be the only cause for exceptions, now that JSON serialization has a fallback.

balloob (approved):

PR #3931, opened 2022-10-23
Use NAN values for TSL2591 saturation

mdonoughe:

When the TSL2591 sensor is saturated, return NAN instead of -1. -1 is very disruptive when you want to know how bright it is and trigger automations based on that. Other software like Home Assistant doesn't know -1 is a dummy value and you end up with data like this:

image

It looks like ESPHome uses NAN to indicate when a value cannot be determined.

I also added a note to the warning about decreasing integration time. Lower integration times should mean less values are accumulated and you can avoid saturation. In my case, the sensor is in a window and when the sun is to the south it shines directly on the sensor so even if I have the gain set to low the sensor saturates at 600ms integration time.

The devcontainer file was updated to use docker.io/esphome/esphome-lint:dev. Docker defaults to docker.io but Podman only does so for well known images. The docker.io version works for both.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

```yaml
# Example config.yaml
i2c:
sensor:
  - platform: tsl2591
    # This combination will easily saturate to show changed behavior
    integration_time: 600ms
    gain: maximum
    calculated_lux:
      id: i_lux
      name: "TSL2591 Lux"
```

## Checklist:


wjcarpenter:
I was not aware of the NAN convention as described here. If that's correct, then this change to the TSL2591 code looks good to me. (I can't comment one way or the other for the docker image change.)

.devcontainer/devcontainer.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ESPHome Dev",
-  "image": "esphome/esphome-lint:dev",
+  "image": "docker.io/esphome/esphome-lint:dev",
```
jesserockz: Please dont make unrelated changes in this pr
```suggestion
  "image": "esphome/esphome-lint:dev",
```

mdonoughe:
Is there anything else to do for this PR?

I don't know if there is a canonical source for information about NAN values, but it's in a few places like [Template Sensor](https://esphome.io/components/sensor/template.html?highlight=nan).

jesserockz (approved):

PR #3949, opened 2022-10-26
Set humidity accuracy decimals to 1

dirrgang:

The [technical specification o fthe sensor has a resolution of 0.1%](https://cdn-shop.adafruit.com/datasheets/Digital+humidity+and+temperature+sensor+AM2302.pdf). I feel like the sensor should send the data as accurately as possible. Reduction of the resolution can then be done via `accuracy_decimals=0` in the configuration, if one is so inclined. As it stands, the higher resolution is intransparent to the user, as the documentation states: 

> "Manually set the number of decimals to use when reporting values. *This does not impact the actual value reported to Home Assistant, it just sets the number of decimals to use when displaying it.*"

The impression I got at first was that there would be no point in increasing the decimals, as there was no obvious way of knowing that there is already a default value in play that hides one digit of resolution.

I suggest either setting the default to 1 or stating in the documentation clearly, that one character of resolution is intentionally hidden.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/dht/sensor.py
+++ b/esphome/components/dht/sensor.py
@@ -42,7 +42,7 @@ CONFIG_SCHEMA = cv.Schema(
         ),
         cv.Optional(CONF_HUMIDITY): sensor.sensor_schema(
             unit_of_measurement=UNIT_PERCENT,
-            accuracy_decimals=0,
+            accuracy_decimals=1,
             device_class=DEVICE_CLASS_HUMIDITY,
             state_class=STATE_CLASS_MEASUREMENT,
         ),


dirrgang:
I realise in hindsight that this might be chosen as a default because the DHT11 sensor's humidity accuracy has no decimal digits. I'm not familiar enough with how the templating in esphome works to make an automatic differentiation between which default is chosen depending on the sensor used. However, it still seems like this would benefit from a change. Either signify in the documentation that the DHT22 sensors can display an extra digit of accuracy, or (preferable to me, though I'm not sure how it's done) differentiate the DHT sensors and change default values accordingly.

dirrgang:
Is setting the decimal accuracy here even strictly necessary? Wouldn't it be easier to just remove the default? (I would assume the default is just "send whatever accuracy is given", and the differentiation between DHT11 and DHT22 would happen automatically.

jesserockz:
I think the default is 0 becase the DHT11 might have been the first sensor implemented which only has +- 5% accuracy, so there was no point in decimals.

I'm inclined to leave it as it is as if you have one with higher resolution, (as you said in pr description) you can change the yaml to have `accuracy_decimals: 1` to allow your specific project to have 1 decimal.

dirrgang:
> I think the default is 0 becase the DHT11 might have been the first sensor implemented which only has +- 5% accuracy, so there was no point in decimals.
> 
> I'm inclined to leave it as it is as if you have one with higher resolution, (as you said in pr description) you can change the yaml to have `accuracy_decimals: 1` to allow your specific project to have 1 decimal.

Works for me, though I suggest adding something to the documentation then?

jesserockz:
> Works for me, though I suggest adding something to the documentation then?

Yeah, Please do so if that helps =)

PR #3958, opened 2022-10-27
Fix local webserver based on esphome/esphome-webserver#17

### LINKED ISSUE #3720: web_server.local: true -> "Uncaught ReferenceError: assignment to undeclared variable url"
### The problem

When enabling `web_server.local` to include the JS/CSS assets and then going to the web interface, the page is blank and the console shows a Javascript error:

```
 Uncaught ReferenceError: assignment to undeclared variable url
    st http://box.local/:1
    <anonymous> http://box.local/:1
box.local:1:28727
    st http://box.local/:1
    <anonymous> http://box.local/:1
```

I also tested this on a different device running `2022.9.0-dev`, which worked just fine on that version, but after updating it to `2022.10.0` it showed the same issue.

2022.10.0

pip

_No response_

ESP32

### Board

nodemcu

web_server

### Example YAML snippet

```yaml
web_server:
  port: 80
  local: true
```
```


_No response_

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by RoboMagus**

Since 2022.10.x release the local use of the webserver component is no longer working, see esphome/issues#3720.
This PR contains the header based on esphome/esphome-webserver#17 and works with local use of the webserver component.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/issues#3720

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


jesserockz:
This PR shows me I need to make a better way to version the webserver code so it can be matched properly in ESPHome :thinking:

jesserockz (approved):

RoboMagus:
> This PR shows me I need to make a better way to version the webserver code so it can be matched properly in ESPHome 🤔

@jesserockz, I had been thinking of a way to realize a more hands-off integration of these repositories as well.

Here's my 2 cents on the matter:

- I'd propose having a `latest` tag on the webserver repo, that would have the generated header files attached. Using a github workflow this could easily be realized on every merge into master.

- When a new version is released on the webserver repo a workflow could be used to automatically create a PR for the esphome core repo that would update the webserver header files.

- The webserver yaml config could be extended with a version field (Both local and using the publicly hosted JS) that would pull the corresponding generated header from the webserver repo based on provided tags (eg. dev / latest / stable).
  - Depending on how the public JS is hosted, the default version selection would grab the stable JS `https://oi.esphome.io/v2/www.js`, but e.g. a dev version could be selected that would direct to `https://oi.esphome.io/dev/v2/www.js`.
  - This approach would have my personal preference, as it does away with merging repos entirely and at the same time allows for easier testing of upcomming changes to the webserver.


Thoughs?

P.s.. This is probably not the best way of dropping my suggestions, but if a new issue would have to be opened to further discuss this topic that's still possible :)

PR #3962, opened 2022-10-28
Gitpod

alexiri:

Installation in Gitpod is broken, this fixes it.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,6 +3,8 @@ ports:
   - port: 6052
     onOpen: open-preview
 tasks:
-  # yamllint disable-line rule:line-length
-  - before: pyenv local $(pyenv version | grep '^3\.' | cut -d ' ' -f 1) && script/setup
+  - init: |
+      pyenv install 3.9.0
+      pyenv local 3.9.0
+      script/setup
     command: python -m esphome dashboard config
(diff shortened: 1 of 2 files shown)


balloob:
This means no one uses it. We should delete it so we don't have to maintain it. Adding a button to the README is absurd. No one uses it and it looks like you just want to use our READMEs as advertisement space for gitpod

alexiri:
@balloob I'm sorry you got that impression. I'm not affiliated with Gitpod at all, I simply thought it was a cool service to set up a dev environment and it could make it easier for people to contribute to ESPHome.

balloob:
We very much prefer codespaces, as it's definition can also be used when doing local development and a standard is also available https://containers.dev

alexiri:
Then it's probably worthwhile to mention that in the contribution documentation, I haven't seen it anywhere. Gitpod *is* mentioned there, however, which is actually how I discovered it in the first place. :upside_down_face: 

It's probably also worth noting that Codespaces aren't available to everyone, or at least they aren't to me.

balloob:
Good point, I have opened https://github.com/esphome/esphome-docs/pull/2399

Codespaces should be more widely available soon.

PR #3972, opened 2022-10-31
Fix non-threadsafe memory allocation

bdraco:

My ESP32s don't crash on BLE operations with this change.

I'm not sure if its the right fix though so I'm leaving it up here for someone who might know more about the issue.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


bdraco:
Seems to not be an issue with the latest idf. I think it was actually just running out of ram

PR #3977, opened 2022-11-01
Close #3746: erroneous 0 measurments in pulse_meter

### LINKED ISSUE #3746: `pulse_meter` generates erroneous 0 measurements for pulse intervals <1ms
### The problem

When `pulse_meter` receives very fast pulses with an interval of <1ms then it publishes a 0 measurement, which is incorrect:

[pulse_meter/pulse_meter_sensor.cpp#L48](https://github.com/esphome/esphome/blob/dev/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L48

Problem is that the interval is truncated to ms first; if the actual interval is shorter, `pulse_meter` will publish 0. I'd suggest to simply remove it.

2022.11.0-dev

pip

ESP32

### Board

esp32dev

pulse_meter

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

jkorinth:

Removes publishing of 0 measurements in case pulse interval is <1ms.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [bug #3746](https://github.com/esphome/issues/issues/3746)

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/pulse_meter/pulse_meter_sensor.cpp
+++ b/esphome/components/pulse_meter/pulse_meter_sensor.cpp
@@ -45,11 +45,7 @@ void PulseMeterSensor::loop() {
   // We quantize our pulse widths to 1 ms to avoid unnecessary jitter
   const uint32_t pulse_width_ms = this->pulse_width_us_ / 1000;
   if (this->pulse_width_dedupe_.next(pulse_width_ms)) {
-    if (pulse_width_ms == 0) {
-      // Treat 0 pulse width as 0 pulses/min (normally because we've not
-      // detected any pulses for a while)
-      this->publish_state(0);
-    } else {
+    if (pulse_width_ms > 0) {
       // Calculate pulses/min from the pulse width in ms
       this->publish_state((60.0f * 1000.0f) / pulse_width_ms);
     }


jesserockz:
Wouldn't this mean the sensor would never be able to publish 0 after the timeout? It would just stay whatever the last value was...

cstaahl:
I don't think this is the right way to fix this. It should **not** be possible to configure the internal filter to be less than 1ms. 
I have to admit that I don't understand why it is possible in the first place. This check
https://github.com/esphome/esphome/blob/294901fbe952813362a945abf1ded99e58a362d5/esphome/components/pulse_meter/sensor.py#L40-L41
should render an error message like this: `Maximum precision is milliseconds.` if you try to set the internal filter to be less than 1ms. It works for the binary sensor but not pulse_meter.

The fact that the default value of the internal filter is 13µs just adds to my confusion.
https://github.com/esphome/esphome/blob/294901fbe952813362a945abf1ded99e58a362d5/esphome/components/pulse_meter/sensor.py#L69

@jesserockz Is there something missing in the python file that correctly set ups all of the plausibility checks? I'm sorry to ask, but all of the python code generation is close to black magic to me and I have not found any place where it is described how it all works within ESPhome. Happy to read up on it if it exists.

jesserockz:
@cstaahl The internal filter of the pulse meter uses `positive_time_period_`**microseconds** not milliseconds which is why it allows sub ms times.
This is carried over from the `pulse_counter` which is set to 13us because of hardware capabilities. There is no other reason it is set to this.

cstaahl:
My embarrassing but apparent illiteracy aside, thank you @jesserockz btw, i think there is an issue in the mismatch between the configuration possibilities and the logic in the code. I see two options:
1. Change the
```
 def validate_internal_filter(value): 
     return cv.positive_time_period_microseconds(value)
```
to 
```
 def validate_internal_filter(value): 
     return cv.positive_time_period_milliseconds(value)
```
and leave the quantization as is it is. Or

2.  Leave the configuration as it is and remove the quantization, in effect having µs precision for the pulse width.

I do not know the reasons behind implementing the quantization in the first place or what the effects of removing it would be. Maybe @stevebaxter, you know more as you implemented it in the first place?

PR #3982, opened 2022-11-02
Allow dashboard import to pull complete file from github

jesserockz:

This allows a pre-flashed device to indicate the user should start with the full config for customisation instead of having to manually copy/paste the file after adoption.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/dashboard_import/__init__.py
+++ b/esphome/components/dashboard_import/__init__.py
@@ -1,4 +1,6 @@
 from pathlib import Path
+import re
+import requests
 
 import esphome.codegen as cg
 import esphome.config_validation as cv
@@ -25,9 +27,12 @@ def validate_import_url(value):
 
 
 CONF_PACKAGE_IMPORT_URL = "package_import_url"
+CONF_IMPORT_FULL_CONFIG = "import_full_config"
+
 CONFIG_SCHEMA = cv.Schema(
     {
         cv.Required(CONF_PACKAGE_IMPORT_URL): validate_import_url,
+        cv.Optional(CONF_IMPORT_FULL_CONFIG, default=False): cv.boolean,
     }
 )
 
@@ -41,6 +46,9 @@ wifi:
 
 async def to_code(config):
     cg.add_define("USE_DASHBOARD_IMPORT")
+    url = config[CONF_PACKAGE_IMPORT_URL]
+    if config[CONF_IMPORT_FULL_CONFIG]:
+        url += "?full_config"
     cg.add(dashboard_import_ns.set_package_import_url(config[CONF_PACKAGE_IMPORT_URL]))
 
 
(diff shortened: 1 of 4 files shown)


esphome/components/dashboard_import/__init__.py
```diff
@@ -64,17 +72,36 @@ def import_config(
...
-        p.write_text(output, encoding="utf8")
+        m = re.match(
+            r"github://([a-zA-Z0-9\-]+)/([a-zA-Z0-9\-\._]+)/([a-zA-Z0-9\-_.\./]+)(?:@([a-zA-Z0-9\-_.\./]+))(?:\?([a-zA-Z0-9\-_.\./]+))?",
+            import_url,
+        )
+
+        if m is None:
+            raise ValueError
+
+        if m.group(4) and "full_config" in m.group(4):
```
balloob: Let's use url parsing. Is part of python

jesserockz: Yeah I initially started with that, but need to break up the path manually as `@main` at the end for branch is still just part of the path and needs to be moved in the middle of the url.

Another way could be to allow a full https url to the raw yaml file?

balloob: Maybe just do more basic right now and do `url.endswith("?full_config")` ?

jesserockz: I still have to reorder the components in the next line...

1, 2, **4**, 3

```py
url = f"https://raw.githubusercontent.com/{m.group(1)}/{m.group(2)}/{m.group(4)}/{m.group(3)}"
```

esphome/components/dashboard_import/__init__.py
```diff
@@ -64,17 +72,36 @@ def import_config(
...
-            },
-        }
-        output = dump(config)
-
-        if network == CONF_WIFI:
-            output += WIFI_CONFIG
-
-        p.write_text(output, encoding="utf8")
+        m = re.match(
+            r"github://([a-zA-Z0-9\-]+)/([a-zA-Z0-9\-\._]+)/([a-zA-Z0-9\-_.\./]+)(?:@([a-zA-Z0-9\-_.\./]+))(?:\?([a-zA-Z0-9\-_.\./]+))?",
```
balloob: Can you use named capture groups for easier readability.

esphome/components/dashboard_import/__init__.py
balloob: Don’t we have this logic already somewhere else where we resolve packages ?

jesserockz: Yes and No. The package code itself will convert the shortform into a full https url that `git` can then clone and cache. Whereas this code is just downloading the single file in raw text form one time.

esphome/components/dashboard_import/__init__.py
jesserockz: Learned something new =)

esphome/components/dashboard_import/__init__.py
```diff
@@ -64,17 +72,30 @@ def import_config(
...
+
+        if git_file.query and "full_config" in git_file.query:
+            url = git_file.raw_url
+            try:
+                req = requests.get(url, timeout=30)
+                req.raise_for_status()
+            except requests.exceptions.RequestException as e:
+                raise ValueError(f"Error while fetching {url}: {e}") from e
+
+            p.write_text(req.text, encoding="utf8")
```
balloob, line 85: Not for this PR, but it would be nice if _somehow_ we could override the wifi using secrets, but then I also don't want to lose the comments.

balloob (approved):

esphome/components/dashboard_import/__init__.py
jesserockz: Yeah that would be good. Something to look at down the line.

PR #3993, opened 2022-11-04
Hydreon: Use new device classes water and precipitation_intensity

functionpointer:

Use new device classes `water` and `precipitation_intensity`

## Types of changes

- [X] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** none

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:


esphome/components/hydreon_rgxx/sensor.py
```diff
@@ -70,31 +72,31 @@ def _validate(config):
             cv.Optional(CONF_ACC): sensor.sensor_schema(
                 unit_of_measurement=UNIT_MILLIMETERS,
                 accuracy_decimals=2,
-                device_class=DEVICE_CLASS_HUMIDITY,
+                device_class=DEVICE_CLASS_WATER,
```
khenderick: Is `UNIT_MILLIMETERS` supported for a device class `DEVICE_CLASS_WATER`? From https://www.home-assistant.io/integrations/sensor/#device-class it looks like water expects one of L, gal, m³, or ft³.

It most likely will work just fine as the previous `DEVICE_CLASS_HUMIDITY` did, but it feels like a device class `DEVICE_CLASS_PRECIPITATION` should exist :thinking: .

functionpointer: Yeah it isn't supported. And yes, it works fine. Just like `DEVICE_CLASS_HUMIDITY` also didn't support `UNIT_MILLIMETERS`.

`DEVICE_CLASS_PRECIPITATION` would indeed be the best. But since that doesn't exist I figured `DEVICE_CLASS_WATER` is better than `DEVICE_CLASS_HUMIDITY`.

I think the biggest downside of an unsupported unit will be incompatibility with long term statistics, which `DEVICE_CLASS_WATER` would normally allow.
A proper long term statistic specific to precipitation would be cool, but probably overkill and too niche.

khenderick: Indeed, it's certainly better. I've created https://community.home-assistant.io/t/device-class-for-precipitation/485209 to request a proper precipitation device class, as it seems the dedicated units already exist in HA: https://github.com/home-assistant/core/blob/master/homeassistant/const.py#L750. After that, it could be aligned in ESPHome as well.

jesserockz: What about `DEVICE_CLASS_PRECIPITATION_INTENSITY`?
It was added at the same time as `DEVICE_CLASS_WATER` although I am not sure of what it is actually measuring...
#3940

functionpointer: I am actually using it, look through the changes!
It expects units that measure the rate of precipitation, like mm/h.

jesserockz: https://github.com/esphome/esphome/pull/4219

khenderick (approved):

khenderick:
A brief update that I've noticed HA 2023.1 raises an error about the current version using humidity.

> Entity sensor.weather_station_rain_event (<class 'homeassistant.components.esphome.sensor.EsphomeSensor'>) is using native unit of measurement 'mm' which is not a valid unit for the device class ('humidity') it is using; Please update your configuration if your entity is manually configured, otherwise create a bug report at https://github.com/home-assistant/core/issues?q=is%3Aopen+is%3Aissue+label%3A%22integration%3A+esphome%22

This PR also solves this error.

functionpointer:
Any reason why this isn't getting merged?
As far as i can see all issues have been solved.
Just missing a review by @jesserockz?

jesserockz (approved):

PR #3564, opened 2022-06-14
Update esp32 arduino and platform versions

jesserockz:
# What does this implement/fix?

Follows on from and closes #2605

Left as a draft for testing
Relies on #3566 

## Breaking Changes:

- TLK110 Ethernet driver removed

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
external_components:
  - source: github://pr#3564
  	refresh: 0s
    components: 
      - esp32
      - esp32_camera
      - ethernet_info
      - ethernet
      - shelly_dimmer
      - wifi
      - dallas

```

## Checklist:
  - [ ] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #3998, opened 2022-11-06
Possible fix for slow pin mode on Arduino >= 2.0.3

jenscski:

The GPIO implementation changed in Arduino 2.0.3, resulting in a slow pin mode change, resulting in dallas component stop working.

The PR changes to always use idf gpio handling i ESPHome regardless of compiling using IDF or using Arduino.

https://github.com/esphome/issues/issues/3401
https://github.com/espressif/arduino-esp32/issues/7049

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/esp32/gpio.py
```diff
@@ -164,11 +162,7 @@ def validate_supports(value):
 
 
 def _choose_pin_declaration(value):
-    if CORE.using_esp_idf:
-        return cv.declare_id(IDFInternalGPIOPin)(value)
-    if CORE.using_arduino:
-        return cv.declare_id(ArduinoInternalGPIOPin)(value)
-    raise NotImplementedError
+    return cv.declare_id(IDFInternalGPIOPin)(value)
```
ssieb, line 165: Might as well remove the function entirely.

esphome/components/esp32/gpio_idf.h
```diff
@@ -1,6 +1,5 @@
 #pragma once
 
-#ifdef USE_ESP32_FRAMEWORK_ESP_IDF
 #include "esphome/core/hal.h"
 #include <driver/gpio.h>
```
ssieb, line 4: Did this compile for you?  I think when compiling with Arduino, the IDF headers aren't available (at least by default).

jesserockz:
Haha, I actually just did the same change in #3564 (https://github.com/esphome/esphome/pull/3564/commits/3f9deb12638671dbd300e60ad8ac15de899af6fd) and did not even see this, or your comment on the PR

jesserockz:
This most important thing to check is that this works with the existing arduino version which uses esp-idf v3. If not then it will have to be merged as part of #3564 only

esphome/components/esp32/gpio_idf.h
jesserockz: They are available and work, the issue here is that this file still needs `#ifdef USE_ESP32` as the failure comes from the esp8266 test

jenscski:
Closing this as it's already fixed in https://github.com/esphome/esphome/pull/3564

@jesserockz did you take a look at the different timings regarding idf and Arduino in the esp_one_wire.cpp?
https://github.com/esphome/esphome/blob/3e110681c9f92db596aa6f7385b929e914f7a41b/esphome/components/dallas/esp_one_wire.cpp#L88-L94

jesserockz:
Oh I did not, I will fix that up too. Thanks

PR #4001, opened 2022-11-07
Add Haier climate component

Yarikx:

The component adds new Haier climate component. esphome device can talk with various Haier climate devices over UART to control it. This PR is based on https://github.com/MiguelAngelLV/esphaier

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** [esphome/esphome-docs#2431](https://github.com/esphome/esphome-docs/pull/2431)

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
logger:
    baud_rate: 0 #Disable UART logging for ESP8266

uart:
    rx_pin: GPIO3
    tx_pin: GPIO1
    baud_rate: 9600

climate:
    platform: haier
    name: Haier AC
    supported_swing_mode: vertical
    update_interval: 10s

```

## Checklist:

### DIFF
--- a/esphome/components/haier/climate.py
+++ b/esphome/components/haier/climate.py
@@ -2,113 +2,40 @@ from esphome.components import climate
 import esphome.codegen as cg
 import esphome.config_validation as cv
 from esphome.components import uart
-from esphome.const import (
-    CONF_ID,
-)
+from esphome.const import CONF_ID
 
 DEPENDENCIES = ["uart"]
 
 haier_ns = cg.esphome_ns.namespace("haier")
 HaierClimate = haier_ns.class_(
-    "Haier", climate.Climate, cg.PollingComponent, uart.UARTDevice
+    "HaierClimate", climate.Climate, cg.PollingComponent, uart.UARTDevice
 )
+SwingMode = haier_ns.enum("SwingMode")
+
+CONF_SWING_MODE = "supported_swing_mode"
+SWING_MODES = {
+    "off": SwingMode.SWING_OFF,
+    "vertical": SwingMode.SWING_VERTICAL,
+    "horizontal": SwingMode.SWING_HORIZONTAL,
+    "both": SwingMode.SWING_BOTH,
+}
 
 CONFIG_SCHEMA = cv.All(
     climate.CLIMATE_SCHEMA.extend(
         {
             cv.GenerateID(): cv.declare_id(HaierClimate),
-            # cv.GenerateID(CONF_TUYA_ID): cv.use_id(Tuya),
-            # cv.Optional(CONF_SUPPORTS_HEAT, default=True): cv.boolean,
-            # cv.Optional(CONF_SUPPORTS_COOL, default=False): cv.boolean,
-            # cv.Optional(CONF_SWITCH_DATAPOINT): cv.uint8_t,
-            # cv.Optional(CONF_ACTIVE_STATE_DATAPOINT): cv.uint8_t,
-            # cv.Optional(CONF_ACTIVE_STATE_HEATING_VALUE, default=1): cv.uint8_t,
-            # cv.Optional(CONF_ACTIVE_STATE_COOLING_VALUE): cv.uint8_t,
-            # cv.Optional(CONF_HEATING_STATE_PIN): pins.gpio_input_pin_schema,
-            # cv.Optional(CONF_COOLING_STATE_PIN): pins.gpio_input_pin_schema,
-            # cv.Optional(CONF_TARGET_TEMPERATURE_DATAPOINT): cv.uint8_t,
-            # cv.Optional(CONF_CURRENT_TEMPERATURE_DATAPOINT): cv.uint8_t,
-            # cv.Optional(CONF_TEMPERATURE_MULTIPLIER): cv.positive_float,
-            # cv.Optional(CONF_CURRENT_TEMPERATURE_MULTIPLIER): cv.positive_float,
-            # cv.Optional(CONF_TARGET_TEMPERATURE_MULTIPLIER): cv.positive_float,
-            # cv.Optional(CONF_ECO_DATAPOINT): cv.uint8_t,
-            # cv.Optional(CONF_ECO_TEMPERATURE): cv.temperature,
+            cv.Optional(CONF_SWING_MODE, default="off"): cv.enum(
+                SWING_MODES, lower=True
+            ),
         }
     )
     .extend(cv.polling_component_schema("5s"))
     .extend(uart.UART_DEVICE_SCHEMA),
-    # cv.has_at_least_one_key(CONF_TARGET_TEMPERATURE_DATAPOINT, CONF_SWITCH_DATAPOINT),
-    # validate_temperature_multipliers,
-    # validate_active_state_values,
-    # cv.has_at_most_one_key(CONF_ACTIVE_STATE_DATAPOINT, CONF_HEATING_STATE_PIN),
-    # cv.has_at_most_one_key(CONF_ACTIVE_STATE_DATAPOINT, CONF_COOLING_STATE_PIN),
-    # validate_eco_values,
 )
 
 
 async def to_code(config):
-    var = cg.new_Pvariable(config[CONF_ID])
+    var = cg.new_Pvariable(config[CONF_ID], config[CONF_SWING_MODE])
     await cg.register_component(var, config)
     await climate.register_climate(var, config)
     await uart.register_uart_device(var, config)
-
-    # paren = await cg.get_variable(config[CONF_TUYA_ID])
-    # cg.add(var.set_tuya_parent(paren))
-    #
-    # cg.add(var.set_supports_heat(config[CONF_SUPPORTS_HEAT]))
-    # cg.add(var.set_supports_cool(config[CONF_SUPPORTS_COOL]))
-    # if CONF_SWITCH_DATAPOINT in config:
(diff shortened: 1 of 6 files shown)


esphome/components/haier/climate.py
```diff
@@ -0,0 +1,44 @@
...
+from esphome.components.climate import ClimateSwingMode
+from esphome.const import CONF_ID, CONF_SUPPORTED_SWING_MODES
+
+DEPENDENCIES = ["uart"]
+
+haier_ns = cg.esphome_ns.namespace("haier")
+HaierClimate = haier_ns.class_(
+    "HaierClimate", climate.Climate, cg.PollingComponent, uart.UARTDevice
+)
+SwingMode = haier_ns.enum("SwingMode")
```
oxan: Unused.
```suggestion
```

esphome/components/haier/haier.h
```diff
@@ -0,0 +1,66 @@
...
+static const uint8_t CRC = 36;
+
+static const uint8_t COMFORT_PRESET_MASK = (1 << 3);
+
+static const uint8_t MIN_VALID_TEMPERATURE = 16;
+static const uint8_t MAX_VALID_TEMPERATURE = 50;
+static const float TEMPERATURE_STEP = 0.5f;
+
+static const uint8_t POLL_REQ[13] = {255, 255, 10, 0, 0, 0, 0, 0, 1, 1, 77, 1, 90};
+static const uint8_t OFF_REQ[13] = {255, 255, 10, 0, 0, 0, 0, 0, 1, 1, 77, 3, 92};
```
oxan: We generally put these constants in the `.cpp` file, as they're not an API but more of an implementation detail.

esphome/components/haier/haier.h
```diff
@@ -0,0 +1,66 @@
...
+
+class HaierClimate : public climate::Climate, public uart::UARTDevice, public PollingComponent {
+ private:
+  uint8_t data_[37];
+  std::set<climate::ClimateSwingMode> supported_swing_modes_{};
```
oxan: We generally put private variables last, not first.

esphome/components/haier/haier.h
```diff
@@ -0,0 +1,66 @@
...
+
+ public:
+  HaierClimate() : PollingComponent(5 * 1000) {}
```
oxan: We define defaults in YAML, not in C++ code.
```suggestion
```

esphome/components/haier/haier.h
```diff
@@ -0,0 +1,66 @@
...
+
+  void setup() override {}
```
oxan: No need for an empty override.
```suggestion
```

esphome/components/haier/haier.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+namespace esphome {
+namespace haier {
+
+static const char *const TAG = "haier";
+
+void HaierClimate::dump_config() {
+  ESP_LOGCONFIG(TAG, "Haier:");
+  ESP_LOGCONFIG(TAG, "  Update interval: %u", this->get_update_interval());
+  dump_traits_(TAG);
```
oxan:
```suggestion
  this->dump_traits_(TAG);
```
In general there seems to be a `this->` prefix in some places but not in others, it's nice to be consistent (general ESPHome style is to use `this->`).

esphome/components/haier/haier.cpp
```diff
@@ -0,0 +1,286 @@
...
+    if (this->read() != 255)
+      return;
+
+    data_[0] = 255;
+    data_[1] = 255;
+
+    this->read_array(data_ + 2, sizeof(data_) - 2);
+
+    read_state_(data_, sizeof(data_));
+  }
```
oxan: You check if there's any data available, but then read `sizeof(data_)` bytes, but there might not be that many bytes available. This can block `loop()`, which is not nice as that may cause connectivity issues. 

I'd suggest something like this instead:
```cpp
  if (this->available() >= sizeof(this->data_)) {
    this->read_array(this->data_, sizeof(this->data_));
    if(this->data_[0] != 255 || this->data_[1] != 255)
      return;
    this->read_state_(this->data_, sizeof(this->data_));
  }
```
This changes the error handling though, in that it will always read `sizeof(data_)` bytes, even if the first two bytes aren't all ones. I don't know if that's the correct behaviour.

esphome/components/haier/haier.cpp
```diff
@@ -0,0 +1,286 @@
...
+  if (check != crc) {
+    ESP_LOGW(TAG, "Invalid checksum");  // TODO more info
+    return;
+  }
+
+  if (MIN_VALID_TEMPERATURE < data[TEMPERATURE] && data[TEMPERATURE] < MAX_VALID_TEMPERATURE) {
+    this->current_temperature = data[TEMPERATURE];
+  }
+
+  this->target_temperature = data[SET_TEMPERATURE] + MIN_VALID_TEMPERATURE;
```
oxan, line 99: Is it correct that `data[TEMPERATURE]` is an absolute temp, while `data[SET_TEMPERATURE]` is an offset w.r.t. the min valid temperature? Seems strange.

esphome/components/haier/haier.cpp
```diff
@@ -0,0 +1,286 @@
...
+      case climate::CLIMATE_FAN_HIGH:
+        data_[FAN_SPEED] = FAN_MAX;
+        break;
+
+      case climate::CLIMATE_FAN_ON:
+      case climate::CLIMATE_FAN_OFF:
+      case climate::CLIMATE_FAN_MIDDLE:
+      case climate::CLIMATE_FAN_FOCUS:
+      case climate::CLIMATE_FAN_DIFFUSE:
+      case climate::CLIMATE_FAN_QUIET:
```
oxan: I suppose this is just here to silence a missing-enumerator-in-switch warning, in which case this can just be a default:
```suggestion
      default:  // other modes are unsupported
```

esphome/components/haier/haier.cpp
```diff
@@ -0,0 +1,286 @@
...
+        data_[SWING] = SWING_BOTH;
+        break;
+    }
+  }
+
+  // Values for "send"
+  data_[COMMAND] = 0;
+  data_[9] = 1;
+  data_[10] = 77;
+  data_[11] = 95;
```
oxan, line 268: It looks kinda strange to have the `COMMAND` index be a constant but to use magic values for the other indices.

esphome/components/haier/haier.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+  send_data_(data_, sizeof(data_));
+}
+
+void HaierClimate::send_data_(const uint8_t *message, uint8_t size) {
+  uint8_t crc = get_checksum_(message, size);
+  this->write_array(message, size - 1);
+  this->write(crc);
+
+  dump_message_("Sent message", message, size);
```
oxan: This logs a different CRC than what is actually sent.
```suggestion
  message[size - 1] = get_checksum_(message, size);
  this->write_array(message, size);
  
  dump_message_("Sent message", message, size);
```

esphome/components/haier/haier.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+void HaierClimate::dump_message_(const char *title, const uint8_t *message, uint8_t size) {
+  ESP_LOGV(TAG, "%s:", title);
+  for (int i = 0; i < size; i++) {
+    ESP_LOGV(TAG, "  byte %02d - %d", i, message[i]);
+  }
```
oxan, line 288: It might be nicer to dump this as a single hex message instead of spamming the logs, though I can see why doing it this way might be easier for debugging. Up to you.
```suggestion
  ESP_LOGV(TAG, "%s: %s", title, format_hex_pretty(message, size).c_str());
```
(requires `#include "esphome/core/helpers.h"`)

oxan (changes requested):
Looks good in general, some small things:

esphome/components/haier/haier.cpp
Yarikx: Yes, the setpoint temperature starts from 16, so it has offset, but the current temperature is "absolute". It looks weird, but the current temperature can be outside the supported set temperature range.

esphome/components/haier/haier.cpp
Yarikx: Good point.
I believe the original reverse engineering of that protocol comes from this post https://blog.instalator.ru/archives/433 (you can google-translate it).
Those are specific parts that need to have specific values for Send command. The true meaning of those is unknown ¯\\_(ツ)_/¯
I'll leave magic numbers, but add a comment about its unknown nature.

esphome/components/haier/haier.cpp
Yarikx: I would prefer to have it displayed byte-by-byte. As the protocol is basically byte-per-value, it's much easier to understand what's going on only by looking at it.
It's LOGV, so should not create too much spam by default.

esphome/components/haier/haier.cpp
oxan: Ok, great. Just wanted to check it wasn't an oversight.

Yarikx:
Thanks @oxan for reviewing the code. I adjusted it according to your comments.  Please take a look, when you will have time.

oxan (approved):

PR #4013, opened 2022-11-10
Feature/pm1006k

AAllport:

Add support for PM 1006K

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
  - platform: pm1006
    pm_1_0:
      name: "PM1006k 1.0µm/m³"
    pm_2_5:
      name: "PM1006k 2.5µm/m³"
    pm_10_0:
      name: "PM1006k 10.0µm/m³"
    update_interval: 10s
```

## Checklist:


Habbie:
Nice! Couple of comments below.

This PR leaves 

`static const uint8_t PM1006_REQUEST[] = {0x11, 0x02, 0x0B, 0x01, 0xE1};`

unchanged. That command is from the PM1006 datasheet. The PM1006K datasheet suggests a different command. However, I do know the command from the PM1006K datasheet works on the PM1006, so I'm not surprised the PM1006 command would work on the PM1006K. However, neither datasheet promises PM1 and PM10 at these offsets in the response to -this- command. (At one time I pondered just sending the PM1006K command so the component would work with both, but I don't have a PM1006K - so I can't verify that PM1 and PM10 live at the offsets you claim [even though it seems likely!].)

Do I see correctly that a PM1006 user would just leave out pm1/pm10 from the yaml? (I'm looking forward to the docs in that regard.)

AAllport:
Thanks for the initial review @Habbie!

RE the command:
Good Question!
Changing the update interval did seem to increase the actual response rate, which would support the idea that both commands work.
Equally, I'm not 100% confident in the results I'm seeing so I've acquired a non-k variant (thanks to IKEA)

I'm guessing it could be better to put a `type` parameter in the config and switch the command.
For reference, the sheet I'm seeing puts the command at `11 01 02 EC`

RE backwards compact:
You are correct. Need to add some docs!

Habbie:
> I'm guessing it could be better to put a `type` parameter in the config and switch the command.

I do agree - things working by luck might change between revisions of the sensors (although I don't expect it).

AAllport:
Solved!

@Habbie, don't suppose you're able to assist with this?
I'm seeing the following error after 3bbc513

I've added the enum to the header, and defined it in the `sensor.py`, but then when trying to run a config, we're seeing the following error.
Any ideas?

```
  File "C:\tools\python3\Lib\site-packages\esphome\loader.py", line 162, in _lookup_module
  File "C:\tools\python3\Lib\importlib\__init__.py", line 126, in import_module
  File "C:\tools\python3\Lib\site-packages\esphome\components\pm1006\sensor.py", line 57, in <module>
    cv.Optional(CONF_TYPE): cv.enum(PM1006Type, upper=True),
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\tools\python3\Lib\site-packages\esphome\config_validation.py", line 1227, in enum
    assert isinstance(mapping, dict)
AssertionError
```

Discussed in discord, I was using `PM1006Type` as opposed to `PM1006_TYPES`

esphome/components/pm1006/sensor.py
```diff
@@ -21,17 +26,56 @@
...
+    TYPE_PM1006K: PM1006Type.PM1006_TYPE_1006K,
+}
+
+SENSORS_TO_TYPE = {
+    CONF_PM_1_0: [TYPE_PM1006K],
+    CONF_PM_2_5: [TYPE_PM1006, TYPE_PM1006K],
+    CONF_PM_10_0: [TYPE_PM1006K],
+
+}
+
```
AAllport, line 44: Should be 2 blank lines here, is fixed locally

AAllport:
I've got 2 sensors next to each other. I believe this looks like valid data!

image

Will sort out the formatting and docs tomorrow, and then I think we're ready for review?

PR #4031, opened 2022-11-14
Update_interval less that 1 second in QMC5883L integration

2mikrobi:
Prevents error that looks are described in:
https://github.com/esphome/issues/issues/1564
integration: qmc5883l

Allow use an update_interval in milliseconds (less that 1 second) in QMC5883L integration.
Until now, compilation of code with "update_interval" lower than 1s breaks with error
''' interval_hz = 1.0/interval_sec
TypeError: unsupported operand type(s) for /: 'float' and 'NoneType'''

## Types of changes

- [ ] Other

https://github.com/esphome/issues/issues/1564

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/qmc5883l/sensor.py
+++ b/esphome/components/qmc5883l/sensor.py
@@ -106,7 +106,9 @@ def auto_data_rate(config):
     interval_sec=0
     if config[CONF_UPDATE_INTERVAL].seconds != None:
       interval_sec=config[CONF_UPDATE_INTERVAL].seconds
-    interval_hz = 1.0 / (interval_sec+config[CONF_UPDATE_INTERVAL].milliseconds/1000)
+    if config[CONF_UPDATE_INTERVAL].milliseconds != None: 
+      interval_sec+=config[CONF_UPDATE_INTERVAL].milliseconds/1000
+    interval_hz = 1.0 / interval_sec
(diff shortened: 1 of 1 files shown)


esphome/components/qmc5883l/sensor.py
```diff
@@ -103,7 +103,11 @@ def validate_enum_bound(value):
 
 
 def auto_data_rate(config):
-    interval_sec = config[CONF_UPDATE_INTERVAL].seconds
+    interval_sec = 0
+    if config[CONF_UPDATE_INTERVAL].seconds is not None:
+        interval_sec = config[CONF_UPDATE_INTERVAL].seconds
+    if config[CONF_UPDATE_INTERVAL].milliseconds is not None:
+        interval_sec += config[CONF_UPDATE_INTERVAL].milliseconds / 1000
```
jesserockz: Its never None... 
Its set by line 100 above with a default of 60s

This would be better:

```suggestion
    interval_sec = config[CONF_UPDATE_INTERVAL].total_milliseconds / 1000
```

2mikrobi: Sure. It is great. Thank you.

jesserockz (approved):

PR #4049, opened 2022-11-18
Improve reliability of bluetooth active connections

bdraco:
Extend the BLE connection state machine with new states:

INIT = Connection is allocated
DISCONNECTING = Client is disconnecting
READY_TO_CONNECT = Client is discovered and ready to make a connection when the scanner is ready

This solves a few issues

- INIT ensure we never try to connect to the same address more than once due to races between the api and HA
- DISCONNECTING ensures we don't report the slot as free until we are actually disconnected since disconnects are not instant.  It also fixes a problem when Home Assistant disconnects since it previously would try to disconnect the device over and over again which could lead to a crash.
- READY_TO_CONNECT ensures don't connect to multiple devices and the same time and ensures we don't connect until the scanner is ready if we are searching for multiple addresses. The scanner is also responsible promoting a device from DISCOVERED to READY_TO_CONNECT

There were some races in the scanner:
1. When the scanner start failed it would never start again because we didn't stop the scanner and try again or release the Semaphore
2. When we held the Semaphore we would release it again right away before starting the scan which lead to a small window where we could end up in an incorrect state
3. When we held the Semaphore we would release it again right away before ending the scan which lead to a small window where we could end up in an incorrect state
4. When we were parsing services we did not hold the Semaphore so it may have been possible for the other task to alter the data (unlikely)

There were some races in the proxy:
1. We could try go connect multiple times to the same device.  This is solved with the new INIT state to ensure only one call to `get_connection_` with `true` can allocate a slot
2. Disconnect calls to the proxy could get the system in an incorrect state if called at the wrong time. There is now a guard to make sure we don't change the state if its IDLE or already DISCONNECTING which avoids a crash in the ble stack. 

Finally there was a memory leak with continuous scanning:
1. We no longer call `print_bt_device_info` if the scanning is continuous since it creates an unbounded list of previously discovered devices in `already_discovered_` which can cause an out of memory crash if there are a lot of devices in range.


esphome/components/esp32_ble_client/ble_client_base.h
```diff
@@ -42,7 +42,7 @@ class BLEClientBase : public espbt::ESPBTClient, public Component {
       memset(this->remote_bda_, 0, sizeof(this->remote_bda_));
       this->address_str_ = "";
     } else {
-      this->address_str_ = str_snprintf("%02x:%02x:%02x:%02x:%02x:%02x", 17, (uint8_t)(this->address_ >> 40) & 0xff,
+      this->address_str_ = str_snprintf("%02X:%02X:%02X:%02X:%02X:%02X", 17, (uint8_t)(this->address_ >> 40) & 0xff,
```
bdraco: This was logging in lowercase

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -332,10 +336,12 @@ void ESP32BLETracker::real_gap_event_handler_(esp_gap_ble_cb_event_t event, esp_
 }
 
 void ESP32BLETracker::gap_scan_set_param_complete_(const esp_ble_gap_cb_param_t::ble_scan_param_cmpl_evt_param &param) {
+  this->set_state(ScannerState::FAILED);
   this->scan_set_param_failed_ = param.status;
 }
 
 void ESP32BLETracker::gap_scan_start_complete_(const esp_ble_gap_cb_param_t::ble_scan_start_cmpl_evt_param &param) {
+  this->set_state(ScannerState::FAILED);
```
bdraco: If we got to this state , scanning would never get restarted on the next try

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -332,10 +336,12 @@ void ESP32BLETracker::real_gap_event_handler_(esp_gap_ble_cb_event_t event, esp_
 }
 
 void ESP32BLETracker::gap_scan_set_param_complete_(const esp_ble_gap_cb_param_t::ble_scan_param_cmpl_evt_param &param) {
+  this->set_state(ScannerState::FAILED);
```
bdraco: If we got to this state , scanning would never get restarted on the next try

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -120,9 +122,24 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
...
+        // Utilize the address type cache
+        connection->set_state(espbt::ClientState::DISCOVERED); 
+        connection->remote_addr_type_ = this->address_type_map_[msg.address];      
+        connection->remote_bda_[0] = (msg.address >> 40) & 0xFF;
+        connection->remote_bda_[1] = (msg.address >> 32) & 0xFF;
+        connection->remote_bda_[2] = (msg.address >> 24) & 0xFF;
+        connection->remote_bda_[3] = (msg.address >> 16) & 0xFF;
+        connection->remote_bda_[4] = (msg.address >> 8) & 0xFF;
+        connection->remote_bda_[5] = (msg.address >> 0) & 0xFF;
+        esp_ble_gap_stop_scanning(); // connection will fail if we are still scanning
```
bdraco: Connections fail most of the time if the scanner is running

esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -47,11 +47,28 @@ bool BLEClientBase::parse_device(const espbt::ESPBTDevice &device) {
   this->remote_bda_[4] = (addr >> 8) & 0xFF;
   this->remote_bda_[5] = (addr >> 0) & 0xFF;
   this->remote_addr_type_ = device.get_address_type();
+
+  // Connection will fail if we are still scanning
+  esp_ble_gap_stop_scanning();
```
bdraco: Connections fail most of the time if the scanner is running

esphome/components/bluetooth_proxy/bluetooth_proxy.h
```diff
@@ -45,12 +45,15 @@ class BluetoothProxy : public esp32_ble_tracker::ESPBTDeviceListener, public Com
...
         free++;
+        ESP_LOGV("bluetooth_proxy", "[%d] Free connection", connection->get_connection_index());           
+      } else {
+        ESP_LOGV("bluetooth_proxy", "[%d] Used connection by [%s]", connection->get_connection_index(),
+                 connection->address_str().c_str());
       }
     }
     return free;
   }
   int get_bluetooth_connections_limit() { return this->connections_.size(); }
```
bdraco:
```suggestion
  int get_bluetooth_connections_limit() { return this->connections_.size(); }

```

esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -10,6 +10,7 @@ namespace esp32_ble_client {
 
 static const char *const TAG = "esp32_ble_client";
 
+
```
bdraco:
```suggestion
```

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -124,7 +126,9 @@ void ESP32BLETracker::loop() {
       for (auto *client : this->clients_) {
         if (client->parse_device(device)) {
           found = true;
+          ESP_LOGW(TAG, "Found device, checking if we should stop scanner");
```
bdraco:
```suggestion
```

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -124,7 +126,9 @@ void ESP32BLETracker::loop() {
...
           if (client->state() == ClientState::DISCOVERED) {
+            ESP_LOGW(TAG, "Found device, stop scanning");
```
bdraco:
```suggestion
```

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -93,11 +93,13 @@ void ESP32BLETracker::loop() {
       connecting = true;
   }
 
+
+
```
bdraco:
```suggestion
```

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -799,6 +803,8 @@ void ESP32BLETracker::dump_config() {
 }
 
 void ESP32BLETracker::print_bt_device_info(const ESPBTDevice &device) {
+  /* Removing this function significantly decreases crashes */
+  return;
```
bdraco: FIXME

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -290,8 +290,8 @@ void ESP32BLETracker::end_of_scan_() {
   }
 
   ESP_LOGD(TAG, "End of scan.");
-  this->scanner_idle_ = true;
-  this->already_discovered_.clear();
+  this->set_state(ScannerState::IDLE);
+  // this->already_discovered_.clear();
```
bdraco: FIXME

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -266,8 +266,8 @@ void ESP32BLETracker::start_scan_(bool first) {
     for (auto *listener : this->listeners_)
       listener->on_scan_end();
   }
-  this->already_discovered_.clear();
-  this->scanner_idle_ = false;
+  // this->already_discovered_.clear();
```
bdraco: FIXME

bdraco:
scan start can fail right after disconnect https://github.com/espressif/esp-idf/issues/5105

```
[10:35:47][I][esp32_ble_client:073]: [0] [DB:02:80:1F:AD:1D] Disconnecting.
[10:35:47][D][esp32_ble_tracker:264]: Starting scan...
[10:35:47][E][esp32_ble_tracker:158]: Scan start failed: 1
[10:35:49]lld_pdu_get_tx_flush_nb HCI packet count mismatch (0, 1)
[10:35:49][V][esp32_ble_client:096]: [0] [DB:02:80:1F:AD:1D] gattc_event_handler: event=41 gattc_if=4
[10:35:49][V][esp32_ble_client:143]: [0] [DB:02:80:1F:AD:1D] ESP_GATTC_DISCONNECT_EVT, reason 62
```

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -83,86 +87,108 @@ void ESP32BLETracker::loop() {
...
-            esp_ble_gap_stop_scanning();
-#ifdef USE_ARDUINO
-            constexpr TickType_t block_time = 10L / portTICK_PERIOD_MS;
-#else
-            constexpr TickType_t block_time = 0L;  // PR #3594
-#endif
-            if (xSemaphoreTake(this->scan_end_lock_, block_time)) {
-              xSemaphoreGive(this->scan_end_lock_);
+    if (this->scan_result_index_ &&  // if it looks like we have a scan result we will take the lock
+        xSemaphoreTake(this->scan_result_lock_, 5L / portTICK_PERIOD_MS)) {
```
bdraco, line 113: This used to release the semaphore before it was done reading the buffer.  That seemed like the buffer could be changed by the gap event though so I changed it to not release the lock until after its done parsing

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -256,8 +282,9 @@ bool ESP32BLETracker::ble_setup() {
 }
 
 void ESP32BLETracker::start_scan_(bool first) {
-  if (!xSemaphoreTake(this->scan_end_lock_, 0L)) {
-    ESP_LOGW(TAG, "Cannot start scan!");
+  // The lock must be held when calling this function.
+  if (xSemaphoreTake(this->scan_end_lock_, 0L)) {
+    ESP_LOGE(TAG, "start_scan called without holding scan_end_lock_");
```
bdraco, line 285: I reversed this so you have to hold the lock to call this so we don't have lock/unlock/lock which was a race condition where the scanner state could get flipped incorrectly between lock and call

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -284,8 +311,9 @@ void ESP32BLETracker::start_scan_(bool first) {
 }
 
 void ESP32BLETracker::end_of_scan_() {
-  if (!xSemaphoreTake(this->scan_end_lock_, 0L)) {
-    ESP_LOGW(TAG, "Cannot clean up end of scan!");
+  // The lock must be held when calling this function.
+  if (xSemaphoreTake(this->scan_end_lock_, 0L)) {
+    ESP_LOGE(TAG, "end_of_scan_ called without holding the scan_end_lock_");
```
bdraco, line 314: I reversed this so you have to hold the lock to call this so we don't have lock/unlock/lock which was a race condition where the scanner state could get flipped incorrectly between lock and call

bdraco:
This is ready for review.  I'm happy to break it up a couple PRs if it makes it easier.

bdraco:
Going to rebase this on top of #3564

bdraco:
```
/Users/bdraco/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp.o: in function `esphome::ISRInternalGPIOPin::digital_read()':
/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp:87: multiple definition of `esphome::ISRInternalGPIOPin::digital_read()'; .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp.o:/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp:114: first defined here
/Users/bdraco/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp.o: in function `esphome::ISRInternalGPIOPin::digital_write(bool)':
/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp:91: multiple definition of `esphome::ISRInternalGPIOPin::digital_write(bool)'; .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp.o:/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp:118: first defined here
/Users/bdraco/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp.o: in function `esphome::ISRInternalGPIOPin::clear_interrupt()':
/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp:95: multiple definition of `esphome::ISRInternalGPIOPin::clear_interrupt()'; .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp.o:/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp:122: first defined here
/Users/bdraco/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp.o: in function `esphome::ISRInternalGPIOPin::pin_mode(esphome::gpio::Flags)':
/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio_arduino.cpp:107: multiple definition of `esphome::ISRInternalGPIOPin::pin_mode(esphome::gpio::Flags)'; .pioenvs/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp.o:/Users/bdraco/trackers/.esphome/build/tracker_proxy_a/src/esphome/components/esp32/gpio.cpp:125: first defined here
```

bdraco:
Had to cleanup the old build dir

bdraco:
Managed to get the scan to fail to start when testing https://github.com/home-assistant/core/pull/82543

```
[13:42:58][D][esp32_ble_tracker:285]: Starting scan...
[13:42:58][D][esp32_ble_tracker:174]: Pausing scan to make connection...
[13:42:58][I][esp32_ble_client:059]: [0] [FC:68:A0:E7:C8:D1] 0x01 Attempting BLE connection
[13:42:58][E][esp32_ble_tracker:162]: Scan start failed: 1
[13:42:59]lld_pdu_get_tx_flush_nb HCI packet count mismatch (0, 1)
[13:42:59][V][esp32_ble_client:096]: [0] [FC:68:A0:E7:C8:D1] gattc_event_handler: event=41 gattc_if=4
[13:42:59][V][esp32_ble_client:096]: [1] [] gattc_event_handler: event=41 gattc_if=5
```

Verified it was able to recover now and it didn't leave the scanner off which meant all the connections would fail in the future.

bdraco:
Looks like there is still some type of race as we can try to start scanning twice

```
[13:57:18][D][esp32_ble_tracker:285]: Starting scan...
[13:57:18][D][esp32_ble_tracker:285]: Starting scan...
[13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
[13:57:18][D][esp32_ble_tracker:285]: Starting scan...
[13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
[13:57:18][D][esp32_ble_tracker:285]: Starting scan...
[13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
[13:57:18][D][esp32_ble_tracker:285]: Starting scan...
[13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
[13:57:18][D][esp32_ble_tracker:285]: Starting scan...
[13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
[13:57:18][D][esp32_ble_tracker:285]: Starting scan...
```

bdraco:
> Looks like there is still some type of race as we can try to start scanning twice
> 
> ```
> [13:57:18][D][esp32_ble_tracker:285]: Starting scan...
> [13:57:18][D][esp32_ble_tracker:285]: Starting scan...
> [13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
> [13:57:18][D][esp32_ble_tracker:285]: Starting scan...
> [13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
> [13:57:18][D][esp32_ble_tracker:285]: Starting scan...
> [13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
> [13:57:18][D][esp32_ble_tracker:285]: Starting scan...
> [13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
> [13:57:18][D][esp32_ble_tracker:285]: Starting scan...
> [13:57:18][E][esp32_ble_tracker:162]: Scan start failed: 1
> [13:57:18][D][esp32_ble_tracker:285]: Starting scan...
> ```

Fixed in d2c2ab36ad8bae5892ee6c1847c362cfc5291fb5

codyc1515:
Hey. Let me know when you think this is ready to test and I will take a look. Some instructions around how to build this branch would be appreciated in advance (thank you!). I am using Docker.

bdraco:
> Hey. Let me know when you think this is ready to test and I will take a look. Some instructions around how to build this branch would be appreciated in advance (thank you!). I am using Docker.

https://github.com/esphome/issues/issues/3761#issuecomment-1327942546

virtualm2000:
Running this for few hours now and everything looks good and stable. Even flooding the proxies by restarting HA cause no issues. I have 20 various Eve devices and 4 proxies.

Thanks!

esphome/components/esp32_ble_client/ble_characteristic.cpp
```diff
@@ -42,7 +42,7 @@ void BLECharacteristic::parse_descriptors() {
     desc->handle = result.handle;
     desc->characteristic = this;
     this->descriptors.push_back(desc);
-    ESP_LOGV(TAG, "[%d] [%s]    descriptor %s, handle 0x%x", this->service->client->get_connection_index(),
+    ESP_LOGI(TAG, "[%d] [%s]    descriptor %s, handle 0x%x", this->service->client->get_connection_index(),
```
bdraco: These do a bit of memory churn to log. Might be better to log them all at V as it actually impacts Connect time

esphome/components/bluetooth_proxy/bluetooth_connection.cpp
```diff
@@ -21,6 +21,7 @@ bool BluetoothConnection::gattc_event_handler(esp_gattc_cb_event_t event, esp_ga
   switch (event) {
     case ESP_GATTC_DISCONNECT_EVT: {
       api::global_api_server->send_bluetooth_device_connection(this->address_, false, 0, param->disconnect.reason);
+      ESP_LOGI(TAG, "[%d] [%s] Disconnected, freeing slot.", this->connection_index_, this->address_str_.c_str());
```
bdraco: Can be broken out to new pr

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -9,11 +9,37 @@
 namespace esphome {
 namespace bluetooth_proxy {
 
+static const uint8_t MAX_CACHE_SIZE = 36;
```
bdraco: The cache change can be separated into a new pr

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -52,6 +78,20 @@ void BluetoothProxy::dump_config() {
   ESP_LOGCONFIG(TAG, "  Active: %s", YESNO(this->active_));
 }
 
+int BluetoothProxy::get_bluetooth_connections_free() {
```
bdraco: The connection logging change can be a new pr

esphome/components/esp32_ble_client/ble_client_base.h
```diff
@@ -43,7 +45,7 @@ class BLEClientBase : public espbt::ESPBTClient, public Component {
...
```
bdraco: This could move into its own pr

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -777,6 +804,8 @@ void ESPBTDevice::parse_adv_(const esp_ble_gap_cb_param_t::ble_scan_result_evt_p
         this->service_datas_.push_back(data);
         break;
       }
+      case ESP_BLE_AD_TYPE_INT_RANGE:
```
bdraco: This is to suppress useless logging. It could be its own pr

esphome/components/esp32_ble_client/ble_characteristic.cpp
bdraco:
```suggestion
    ESP_LOGV(TAG, "[%d] [%s]    descriptor %s, handle 0x%x", this->service->client->get_connection_index(),
```

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
bdraco: https://github.com/esphome/esphome/pull/4096

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -777,6 +804,8 @@ void ESPBTDevice::parse_adv_(const esp_ble_gap_cb_param_t::ble_scan_result_evt_p
...
+        break;
```
bdraco:
```suggestion
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
bdraco: Since its linked to the state machine changes it really can't

esphome/components/esp32_ble_client/ble_client_base.h
bdraco: https://github.com/esphome/esphome/pull/4097

bdraco:
```suggestion
      this->address_str_ = str_snprintf("%02x:%02x:%02x:%02x:%02x:%02x", 17, (uint8_t)(this->address_ >> 40) & 0xff,
```

esphome/components/bluetooth_proxy/bluetooth_connection.cpp
bdraco: https://github.com/esphome/esphome/pull/4098

bdraco:
```suggestion
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -52,6 +78,20 @@ void BluetoothProxy::dump_config() {
...
+      free++;
+      ESP_LOGI(TAG, "[%d] Free connection", connection->get_connection_index());
+    } else {
+      ESP_LOGI(TAG, "[%d] Used connection by [%s]", connection->get_connection_index(),
+               connection->address_str().c_str());
+    }
+  }
+  return free;
+}
+
```
bdraco:
```suggestion
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
bdraco: https://github.com/esphome/esphome/pull/4098

esphome/components/bluetooth_proxy/bluetooth_proxy.h
```diff
@@ -41,15 +41,7 @@ class BluetoothProxy : public esp32_ble_tracker::ESPBTDeviceListener, public Com
...
-  int get_bluetooth_connections_free() {
-    int free = 0;
-    for (auto *connection : this->connections_) {
-      if (connection->address_ == 0) {
-        free++;
-      }
-    }
-    return free;
-  }
+  int get_bluetooth_connections_free();
```
bdraco:
```suggestion
  int get_bluetooth_connections_free() {
    int free = 0;
    for (auto *connection : this->connections_) {
      if (connection->address_ == 0) {
        free++;
      }
    }
    return free;
  }
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
bdraco:
```suggestion
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -9,11 +9,37 @@
...
+    auto top_time = this->times_queue_.top();
+    if (this->address_time_map_.find(top_time) != this->address_time_map_.end()) {
+      uint64_t expire_address = this->address_time_map_[top_time];
+      this->address_type_map_.erase(expire_address);
+      this->address_time_map_.erase(expire_address);
+      this->time_address_map_.erase(top_time);
+    }
+    this->times_queue_.pop();
+  }
+
```
bdraco:
```suggestion
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -120,9 +151,21 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
...
+      if (this->address_type_map_.find(msg.address) != this->address_type_map_.end()) {
+        // Utilize the address type cache
+        connection->found_device(msg.address, this->address_type_map_[msg.address]);
+        ESP_LOGI(TAG, "[%d] [%s] Using connect cache", connection->get_connection_index(),
+                 connection->address_str().c_str());
+      } else {
+        ESP_LOGI(TAG, "[%d] [%s] Searching to connect", connection->get_connection_index(),
+                 connection->address_str().c_str());
+        connection->set_state(espbt::ClientState::SEARCHING);
+      }
```
bdraco:
```suggestion
```

esphome/components/bluetooth_proxy/bluetooth_proxy.h
```diff
@@ -60,6 +60,10 @@ class BluetoothProxy : public esp32_ble_tracker::ESPBTDeviceListener, public Com
 
   BluetoothConnection *get_connection_(uint64_t address, bool reserve);
 
+  std::map<uint64_t, esp_ble_addr_type_t> address_type_map_;
+  std::map<uint64_t, uint64_t> address_time_map_;
+  std::map<uint64_t, uint64_t> time_address_map_;
+  std::priority_queue<uint64_t, std::vector<uint64_t>, std::greater<uint64_t>> times_queue_;
```
bdraco:
```suggestion
```

esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -23,35 +23,40 @@ void BLEClientBase::setup() {
...
+  this->remote_bda_[0] = (addr >> 40) & 0xFF;
+  this->remote_bda_[1] = (addr >> 32) & 0xFF;
+  this->remote_bda_[2] = (addr >> 24) & 0xFF;
+  this->remote_bda_[3] = (addr >> 16) & 0xFF;
+  this->remote_bda_[4] = (addr >> 8) & 0xFF;
+  this->remote_bda_[5] = (addr >> 0) & 0xFF;
+  this->remote_addr_type_ = address_type;
+  this->set_state(espbt::ClientState::DISCOVERED);
+}
+
```
bdraco:
```suggestion
```

esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -23,35 +23,40 @@ void BLEClientBase::setup() {
...
-
-  auto addr = device.address_uint64();
-  this->remote_bda_[0] = (addr >> 40) & 0xFF;
-  this->remote_bda_[1] = (addr >> 32) & 0xFF;
-  this->remote_bda_[2] = (addr >> 24) & 0xFF;
-  this->remote_bda_[3] = (addr >> 16) & 0xFF;
-  this->remote_bda_[4] = (addr >> 8) & 0xFF;
-  this->remote_bda_[5] = (addr >> 0) & 0xFF;
-  this->remote_addr_type_ = device.get_address_type();
+  this->found_device(device.address_uint64(), device.get_address_type());
```
bdraco:
```suggestion
  this->set_state(espbt::ClientState::DISCOVERED);

  auto addr = device.address_uint64();
  this->remote_bda_[0] = (addr >> 40) & 0xFF;
  this->remote_bda_[1] = (addr >> 32) & 0xFF;
  this->remote_bda_[2] = (addr >> 24) & 0xFF;
  this->remote_bda_[3] = (addr >> 16) & 0xFF;
  this->remote_bda_[4] = (addr >> 8) & 0xFF;
  this->remote_bda_[5] = (addr >> 0) & 0xFF;
  this->remote_addr_type_ = device.get_address_type();
```

esphome/components/esp32_ble_client/ble_client_base.h
```diff
@@ -28,6 +28,8 @@ class BLEClientBase : public espbt::ESPBTClient, public Component {
   float get_setup_priority() const override;
 
   bool parse_device(const espbt::ESPBTDevice &device) override;
+  void found_device(uint64_t addr, esp_ble_addr_type_t address_type);
+
```
bdraco:
```suggestion
```

bdraco:
There are some remaining stability issues: 

- [x] There seems to be a crash related to `this->already_discovered` with `scan_continuous_` (maybe leaking)... https://github.com/esphome/issues/issues/3797#issuecomment-1320231104 but there are likely more but I think the main reason I'm getting crashes is busy mDNS and turning off mdns doesn't seem to turn it off at a low level so its still allocating ram  `[13:09:39][V][esp-idf:000]: E (400914) MDNS_Networking: Cannot allocate memory (line: 123, free heap: 672 bytes)` even with mdns off

- [x] https://github.com/esphome/esphome/pull/3905 fixed the leak but it also made connections slower so it times out on these devices because they don't advertise frequently so we cannot connect in the allotted 20s window. We can adjust the timeouts in 
https://github.com/Bluetooth-Devices/bleak-retry-connector/pull/68 but without the cache we are going to still hit them 

- [x] If the scan fails it will never find the device because the only states in the scan code are idle and not idle.  It doesn't know how to recover from a failed scan start.  If it failed we need to release the lock so it can try to start again.  I also fixed the lock/release/lock to call `start_scan_` and `stop_scan_` while holding the lock instead to avoid multiple lock/unlocks each loop.

- [x] If the scanner is running while we are connecting the connection seems to always? fail.

- [x] disconnects don't seem instant, we may need to wait for the disconnected event before trying again or it fails https://github.com/esphome/aioesphomeapi/compare/main...bdraco:aioesphomeapi:disconnect_wait?expand=1. https://github.com/espressif/esp-idf/issues/5105. https://github.com/esphome/aioesphomeapi/pull/310

- [x] Since ble disconnects aren’t instant we need a disconnecting state when the hass connection drops to avoid trying to disconnect over and over  (fixed by adding a disconnecting state)
Fixes pages and pages of this when hass connection drops:
```
[13:04:53][I][esp32_ble_client:079]: [1] [F2:49:E8:15:26:3C] Disconnecting.
[13:04:53][I][esp32_ble_client:079]: [2] [DB:02:80:1F:AD:1D] Disconnecting.
[13:04:53][I][esp32_ble_client:079]: [0] [C4:7C:8D:6D:5A:99] Disconnecting.
[13:04:53][I][esp32_ble_client:079]: [1] [F2:49:E8:15:26:3C] Disconnecting.
[13:04:53][I][esp32_ble_client:079]: [2] [DB:02:80:1F:AD:1D] Disconnecting.
[13:04:53][I][esp32_ble_client:079]: [0] [C4:7C:8D:6D:5A:99] Disconnecting.
[13:04:53][I][esp32_ble_client:079]: [1] [F2:49:E8:15:26:3C] Disconnecting.
[13:04:53][I][esp32_ble_client:079]: [2] [DB:02:80:1F:AD:1D] Disconnecting.
```
- [x] We need to defer connecting if any of the other connections is currently in a discovering state since you can have one waiting for an advertisment and than another one starts connecting and the first one will never get it

- [x] Large service lists from HomeKit devices make the esp run out of ram.   https://github.com/esphome/esphome/pull/4063. This got 2 working at a time.  3 is still iffy, but I can't think of another way to reduce the memory pressure without getting rid of BLEService, BLEDescriptor, BLECharacteristic and use the native ESP ones

- [x] There is a race where a connection fails and we try again before the connection is cleaned up which resulted in the state being set back to SEARCHING because the connect call is called again.  We need to make sure that it can only be set to SEARCHING when its being allocated https://github.com/esphome/esphome/pull/4049/commits/593632dff30c95478b6aebe09d344503b24e6619

- [x] There appears to be another case were the services for different client devices can get mixed up if they are connecting at the same time but this might just be out of ram -- this was caused by the duplicate connections which can no longer happen after https://github.com/esphome/esphome/pull/4049/commits/593632dff30c95478b6aebe09d344503b24e6619

This was fixed by the IDF upgrade:

- [x] sometimes services are missing.  This is a race somewhere. `2022-11-19 20:49:49.105 ERROR (MainThread) [switchbot.devices.device]  (D1:FB:3B:74:C8:90): characteristic missing: cba20003-224d-11e6-9fb8-0002a5d5c51b; Stopping trying; RSSI: -48`

These are likely problems on the hass side (or lib):

- [x] If we get an exception in the callback and from the lib we end up with an unretrieved future https://github.com/bdraco/home-assistant/commit/e250824e60db6461e35dd4d4dfceef745c661bc1. https://github.com/home-assistant/core/pull/82537

There is still a bit of memory pressure for large services list but it doesn't seem to cause crashes anymore

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -120,9 +125,11 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
...
         return;
       }
-      ESP_LOGV(TAG, "[%d] [%s] Searching to connect", connection->get_connection_index(),
-               connection->address_str().c_str());
-      connection->set_state(espbt::ClientState::SEARCHING);
+      if (connection->state() != espbt::ClientState::INIT) {
+        ESP_LOGW(TAG, "[%d] [%s] Connection already in progress", connection->get_connection_index(),
+                 connection->address_str().c_str());
+        return;
+      }
```
bdraco:
```suggestion
      ESP_LOGV(TAG, "[%d] [%s] Searching to connect", connection->get_connection_index(),
               connection->address_str().c_str());
      connection->set_state(espbt::ClientState::SEARCHING);
      if (connection->state() != espbt::ClientState::INIT) {
        ESP_LOGW(TAG, "[%d] [%s] Connection already in progress", connection->get_connection_index(),
                 connection->address_str().c_str());
        return;
      }
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -123,6 +128,11 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
       ESP_LOGV(TAG, "[%d] [%s] Searching to connect", connection->get_connection_index(),
                connection->address_str().c_str());
       connection->set_state(espbt::ClientState::SEARCHING);
+      if (connection->state() != espbt::ClientState::INIT) {
```
jesserockz: The line above this sets the state to SEARCHING, so this will always be true...

bdraco: It should make a bit more sense coupled with https://github.com/esphome/esphome/pull/4099 as I'll need to rebase this one as it will conflict with that one.

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -123,6 +128,11 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
       ESP_LOGV(TAG, "[%d] [%s] Searching to connect", connection->get_connection_index(),
                connection->address_str().c_str());
       connection->set_state(espbt::ClientState::SEARCHING);
```
bdraco:
```suggestion
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -123,6 +128,11 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
...
+      if (connection->state() != espbt::ClientState::INIT) {
+        ESP_LOGW(TAG, "[%d] [%s] Connection already in progress", connection->get_connection_index(),
+                 connection->address_str().c_str());
+        return;
+      }
```
bdraco, line 164:
```suggestion
      }
      connection->set_state(espbt::ClientState::SEARCHING);
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
bdraco: I reordered it so it make sense standalone in this PR

jesserockz: I see, I was gonna go through the list of PRs in order in https://github.com/esphome/issues/issues/3761#issuecomment-1324176431 =)

bdraco:
Found another race and fixed in the last push.

If we try to connect, fail, and than try again, the previous attempt could have succeeded while the hass client was resetting to try again and the esp could be in a connected/established state so we need to handle that case and send back the connected state

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -83,86 +87,105 @@ void ESP32BLETracker::loop() {
...
-      ESPBTDevice device;
-      device.parse_scan_rst(this->scan_result_buffer_[i]);
-
-      bool found = false;
-      for (auto *listener : this->listeners_) {
-        if (listener->parse_device(device))
-          found = true;
+  if (!this->scanner_idle_) {
+    if (!connecting && xSemaphoreTake(this->scan_end_lock_, 0L)) {
+      if (!ready_to_connect && this->scan_continuous_) {
```
bdraco: Also added a check to guard start scan with `! ready_to_connect` to ensure we don't start and than stop it in the same loop

bdraco:
Doing some more testing this morning


The Bluetooth stack seems to have a bug where if you start and stop the scanner in the same loop it gets confused.  We should parse the ble events first before deciding to start/stop the scanner to avoid that

bdraco:
Ran this with [`bdraco:connect_without_resolving_services`](https://github.com/bdraco/esphome/tree/connect_without_resolving_services?rgh-link-date=2022-11-27T23%3A08%3A35Z) merged in overnight and its the first time the esp stayed stable overnight with all the homekit devices paired.   I'm pushing everything through one esp proxy to stress test it

bdraco:
Last fix was for when we timeout if we were just about ready to connect but all the slots were full we wouldn't free the slot

bdraco:
With this + [`bdraco:connect_without_resolving_services`](https://github.com/bdraco/esphome/tree/connect_without_resolving_services?rgh-link-date=2022-11-27T23%3A08%3A35Z) I'm able to keep the remote connection alive now with VERBOSE logging while loading the device and filling up all the slots

jesserockz:
Tested this on top of latest dev and working nicely

jesserockz (approved):

PR #4067, opened 2022-11-22
Avoid compiling user services code if none are configured

bdraco:

Avoid compiling user services code if none are configured

It looks like we can ifdef some of this out if there are no services configured

I'm not 100% sure this isn't used for something else though

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/api/api_pb2_service.cpp
+++ b/esphome/components/api/api_pb2_service.cpp
@@ -607,12 +607,14 @@ bool APIServerConnectionBase::read_message(uint32_t msg_size, uint32_t msg_type,
       break;
     }
     case 42: {
+#ifdef USE_API_USER_SERVICES
       ExecuteServiceRequest msg;
       msg.decode(msg_data, msg_size);
 #ifdef HAS_PROTO_MESSAGE_DUMP
       ESP_LOGVV(TAG, "on_execute_service_request: %s", msg.dump().c_str());
 #endif
       this->on_execute_service_request(msg);
+#endif
       break;
     }
     case 45: {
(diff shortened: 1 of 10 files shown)


esphome/components/api/api_pb2_service.cpp
```diff
@@ -607,12 +607,14 @@ bool APIServerConnectionBase::read_message(uint32_t msg_size, uint32_t msg_type,
       break;
     }
     case 42: {
+#ifdef USE_API_USER_SERVICES
```
jesserockz, line 610: These files are generated and cannot be manually ifdef'd like this...
See the first line of the file =)

esphome/components/api/api_pb2_service.h
```diff
@@ -93,7 +93,9 @@ class APIServerConnectionBase : public ProtoService {
   bool send_get_time_response(const GetTimeResponse &msg);
   virtual void on_get_time_response(const GetTimeResponse &value){};
   bool send_list_entities_services_response(const ListEntitiesServicesResponse &msg);
+#ifdef USE_API_USER_SERVICES
```
jesserockz, line 96: Same here, cant manually change this file

esphome/components/api/custom_api_device.h
```diff
@@ -6,6 +6,7 @@
 
 namespace esphome {
 namespace api {
+#ifdef USE_API_USER_SERVICES
```
jesserockz, line 9: This is a class that a user can use in a custom component, we have no way of knowing if its used at config validation / code generation time

jesserockz (changes requested):
Because of custom user services via included headers, I dont think this is a safe operation to do...

bdraco:
This one is probably not worth pursuing.  We need more runtime heap memory and this isn’t going to make much difference

PR #4078, opened 2022-11-23
Added expose_entities option to mqtt:

grillp:

This allows you to configure the mqtt client so that entities are not exposed to mqtt, meaning they do not send status updates through mqtt, nor can they receive actions through mqtt.

This is for the situation where:
- You use the `api:` component to integrate with HA
- You don't use mqtt discovery to connect to HA
- Your configuration wants to use the `mqtt_client` to interact with other components (via MQTT subscriptions and publishing through lambdas)
- You don't want your Esphome entities to publish their state as they are not used by HA (HA uses the api:)

The current state is that all components, always publish state through MQTT when this may not be desired.

This PR adds an optional boolean configuration option `expose_entities`, which stops this behavior when set to `false`

The option defaults to `true` to ensure the current behavior.

## Types of changes

- [ ] Other

Not an issue but an open feature request: https://github.com/esphome/feature-requests/issues/375

None that I could find

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/2471

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

To turn off publishing of component states:
```yaml
mqtt:
  broker: mqtt.mydomain.com
  id: mqtt_client
  publish_component_state: false
```
To turn on publishing of component states (default), leave off the `publish_component_state` config option:
```yaml
mqtt:
  broker: mqtt.mydomain.com
  id: mqtt_client
```
or define it explicitly:
```yaml
mqtt:
  broker: mqtt.mydomain.com
  id: mqtt_client
  publish_component_state: true
```

## Checklist:


grillp:
Just some notes on the implementation.

I don't like the way I dive into the `CORE` object to check on configuration options in the various `to_code()` methods as it breaks encapsulation.

I did first try to replace the `cv.OnlyWith(CONF_MQTT_ID, "mqtt")` with a new config validation class ` OnlyWithConfig(CONF_MQTT_ID, ["mqtt", CONF_PUBLISH_COMPONENT_STATE]
]):` which would check `CORE.config` at the time the configuration files are validated.

But unfortunately, `CORE.config` is not defined yet at the time the config files are being validated. 🤦 

So that's why I went back to diving into `CORE.config` during code generation.

Is there a nicer way to check other configuration options during code generation?

PR #4063, opened 2022-11-22
Make parse_characteristics and parse_descriptors lazy to reduce memory pressure

bdraco:
# What does this implement/fix?

Reduces memory pressure.

On reconnect we can usually avoid parsing out many of these which reduces the blocking time and memory. The end goal is to be able to get to this commit where we can release the char memory after sending the service list: https://github.com/bdraco/esphome/commit/581d4de2185839749078c9d99efa88e1df9053a0 but it will take a few more merges to be able to do that.

This change was enough to stop the esp from running out of ram when talking to HomeKit devices with large services list.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4092, opened 2022-11-26
Fix registering for gatt notify when characteristic only support indicate

bdraco:

This fixes yale locks and homekit devices never getting notify

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/bluetooth_proxy/bluetooth_connection.cpp
```diff
@@ -253,6 +253,12 @@ esp_err_t BluetoothConnection::notify_characteristic(uint16_t handle, bool enabl
     return ESP_GATT_INVALID_HANDLE;
   }
 
+  if (!(characteristic->properties & (ESP_GATT_CHAR_PROP_BIT_NOTIFY | ESP_GATT_CHAR_PROP_BIT_INDICATE))) {
```
bdraco: This check could be done in home assistant instead which would allow us to avoid parsing the char on reconnects if we do all the other possible checks in ha as well. That should relieve some memory pressure on the device and avoid the crash with the smoke alarms

bdraco: These checks are really for implementer error so we should move them all to ha to reduce cpu and 
Memory pressure on the esp chip.  I’ll work on a pr for ha and then remove this.

it’s still good to merge as is if I don’t get the ha side done first as I’ll scrape out all these checks and move them to ha at once since it’s not going to make much difference unless we can actually avoid parsing the char

bdraco: https://github.com/esphome/esphome/pull/4063 has to happen first before we can actually reduce the memory pressure more by shifting more to ha.

bdraco: https://github.com/home-assistant/core/pull/82807

esphome/components/bluetooth_proxy/bluetooth_connection.cpp
```diff
@@ -253,6 +253,12 @@ esp_err_t BluetoothConnection::notify_characteristic(uint16_t handle, bool enabl
...
+    ESP_LOGW(TAG, "[%d] [%s] Cannot notify GATT characteristic, does not support notify or indicate.",
+             this->connection_index_, this->address_str_.c_str());
+    return ESP_GATT_REQ_NOT_SUPPORTED;
+  }
+
```
bdraco:
```suggestion
```

esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -168,7 +168,12 @@ bool BLEClientBase::gattc_event_handler(esp_gattc_cb_event_t event, esp_gatt_if_
                  this->address_str_.c_str(), param->reg_for_notify.handle, descr->uuid.to_string().c_str());
         break;
       }
-      uint16_t notify_en = 1;
+      /*
+        1 = notify
+        2 = indicate
+      */
+      uint16_t notify_en = descr->characteristic->properties & ESP_GATT_CHAR_PROP_BIT_NOTIFY ? 1 : 2;
```
bdraco, line 175: This is the same logic bluez uses under the hood

jesserockz (approved):

PR #4105, opened 2022-11-28
Add graphical display menu

MrMDavidson:

Adds a new component `graphical_display_menu` / `GraphicalDisplayMenu` which implements `DisplayMenuComponent`. Provides a mechanism for `DisplayBuffer` based displays (einks, etc) to render menus. Code currently supports non-fixed-height menu options.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2572

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Popup Example `config.yaml`:

```yaml
# Example config.yaml
display:
  - platform: ...
    id: my_display

font:
  - file: ...
    id: my_font
    size: 16

graphical_display_menu:
  id: my_menu
  display: my_display
  # Optional: For displays that use `update_interval: never` this provides a trigger when redraw would have fired
  on_redraw:
    then:
      component.update: my_display
  font: my_font
  active: true
  items:
    - type: label
      text: Normal label
    - type: menu
      text: Child Menu
      items:
        - type: label
          text: A child menu item
        - type: back
          text: Back    
    
```

## Adavanced Drawing Mode Example `config.yaml`

```yaml
# Example config.yaml
display:
  - platform: ...
    id: my_display
    pages:
      - id: menu_page
        lambda: |-
          const auto display_width = it.get_width();
          const auto display_height = it.get_height();

          it.print(0, 0, id(my_font), COLOR_ON, TextAlign::TOP_LEFT, "Other drawing example");

          const auto font_height = 20;
          it.menu(0, font_height, id(my_menu), display_width , display_height - font_height);

font:
  - file: ...
    id: my_font
    size: 16

graphical_display_menu:
  id: my_menu
  # Optional: You'll want this to trigger the drawing of the lambda
  on_redraw:
    then:
      component.update: my_display
  font: my_font
  active: true
  items:
    - type: label
      text: Normal label
    - type: menu
      text: Child Menu
      items:
        - type: label
          text: A child menu item
        - type: back
          text: Back    
    
```

## Checklist:


MrMDavidson:
I'd be curious about your thoughts on this @numo68. One thought I had in starting this implementation is that the rendering/layout of menu items might be better suited to the menu items themselves rather than the menu. I'd then imagine basically having a specialised hierarchy of menu items (eg. `GraphicalDisplayMenuItemBase` vs `LcdDisplayMenuItemBase`) that provides the basic drawing but someone could come along and add `MyVeryParticularGraphicalDisplayMenuItem`. For dynamically sized items the menu item itself would also know if the potential size had been changed or not so could invalidate a cached measurement of the item rather than having to calculate the size every draw loop.

nielsnl68:
Hi @MrMDavidson,

Did you know i am working on a GUI (switchplate) version for esphome. 

Do you think we could intergrade your menu into my switchplate code at https://github.com/nielsnl68/switchplate

numo68:
Hi @MrMDavidson ,

nice to see someone stepping up and extending this for graphical displays, thanks :)

> One thought I had in starting this implementation is that the rendering/layout of menu items might be better suited to the menu items themselves rather than the menu.

My idea was to have the menu item manage the data and states/transitions only, and the menu having the responsibility to render it. I still think that it is a clean way to separate the concerns.

I can well imagine the menu rendering engine attaching some data to the menu items for caching purposes or similar (transparent to the item), or the menu item setting some kind of a 'dirty' flag indicating that the value to display changed so the positions/dimensions might need to be recalculated. Maintaining a parallel hierarchy for both menu and menu items depending on the type of display to use seem harder to maintain to me.

I am however not very good at the best practices on how to implement GUIs so I guess I let others comment on what approach is optimal.

landonr:
excited to see the progress @mrmdavidson! I've been working on a lcd menu for esphome on my project here https://github.com/landonr/esphome-remote and I want to make it more useful for general esphome users as a component. I've been looking at https://github.com/Spirik/GEM for inspiration. 
- my ui supports menu hierarchy, checkboxes, sliders, and has a scrollbar. 
- the menu items in my project have a state and a type so the ui can draw checkboxes, or icons
- my project has a ton of custom code for managing media players that I need to split up from the ui
- it would be nice if the menu code was separate for the display rendering so people could use tft_espi or lvgl if they want to
- how do you plan on viewing sensor states inside of a child menu? can you update your example

nagyrobi:
@MrMDavidson the menu component works exceptionally well on Character LCD displays and having the option to display the same logic on graphical displays is very welcome. 

Will this be full-screen only, or will this be able to be shown among other items on the screen, eg. in a rectangle to constrain it within a space, and/or on a separate page?

MrMDavidson:
> Will this be full-screen only, or will this be able to be shown among other items on the screen, eg. in a rectangle to constrain it within a space

@nagyrobi: It's all positioned dynamically so adding a bounding box to constrain it isn't too hard. 

> on a separate page?

Currently it creates its own `DisplayPage` on setup to take ownership of to draw the menu. From memory I don't think I'd be able to draw over the top of an existing page that is already defined as I need to access the `display_writer_t`. It might still be worth while allowing users to pass in a specific `DisplayPage` so they can skip it easier for navigation purposes, or whatever

MrMDavidson:
> My idea was to have the menu item manage the data and states/transitions only, and the menu having the responsibility to render it. I still think that it is a clean way to separate the concerns.
>
> I can well imagine the menu rendering engine attaching some data to the menu items for caching purposes or similar (transparent to the item), or the menu item setting some kind of a 'dirty' flag indicating that the value to display changed so the positions/dimensions might need to be recalculated.

@numo68 I think the current approach works well as a sort of flyweight pattern and if there's no new menus. The problem I see is that if someone wants to add a new menu type they have to add a new `MenuItemType`, create the `MenuItem` derivative, and update the `DisplayMenuComponent`. If you flip it over so the `MenuItem` draws itself `draw_menu()` basically just becomes;

```
int current_y = 0;

for (size_t i = 0; i < this->displayed_item_->items_size(); i++) {
  auto *item = this->displayed_item->get_item(i);
  int x = 0;
  int y = current_y;
  int width = 0;
  int height = 0;
  item->draw(x, y, &width, &height);
  current_y += height;
}
```

And any new styles of menu just require a new implementation of `MenuItem`. We could then have `MenuItem` contain basic state, `LcdMenuItem` derive from it and basically take what's in the current `draw_item` so the current menu items don't need to be changed. And then `GraphicalDisplayMenuItem` from `MenuItem`. 

> Maintaining a parallel hierarchy for both menu and menu items depending on the type of display to use seem harder to maintain to me.

The parallel hierarchy should only be an issue for the python code in `menu_item_to_code`.

> I am however not very good at the best practices on how to implement GUIs so I guess I let others comment on what approach is optimal.

I might branch from this if I get a chance and have a quick hack at an example so we can see it in practice.

MrMDavidson:
> excited to see the progress @MrMDavidson! I've been working on a lcd menu for esphome on my project here https://github.com/landonr/esphome-remote and I want to make it more useful for general esphome users as a component. I've been looking at https://github.com/Spirik/GEM for inspiration.
> 
> * my ui supports menu hierarchy, checkboxes, sliders, and has a scrollbar.
> * the menu items in my project have a state and a type so the ui can draw checkboxes, or icons
> * my project has a ton of custom code for managing media players that I need to split up from the ui
> * how do you plan on viewing sensor states inside of a child menu? can you update your example

@landonr Have a look at https://esphome.io/components/display_menu/index.html - this is just an implementation of this design for non-character displays. As a result it handles labels, sub menus, switches, numbers, etc. A switch would render something like `Room Lights: [On]` using

```
graphical_display_menu:
  ...
  items:
    type: switch
    text: 'Room Lights'
    on_text: 'On'
    off_text: 'Off'
    switch: room_light_switch

switch:
  - platform: ...
    id: room_light_switch
```

> * it would be nice if the menu code was separate for the display rendering so people could use tft_espi or lvgl if they want to

In ESPHome a lot of display implementations implement an abstraction `DisplayBuffer` this `graphical_display_menu` component relies exclusively on this abstraction for drawing. You pass it an existing display which implements `DisplayBuffer` and it tells the display what it wants to draw. I'm testing it using a `waveshare_epaper` component for instance;

```yaml
display:
  - platform: waveshare_epaper
    id: my_menu_display

graphical_display_menu:
  id: graphical_menu
  display: my_menu_display # This is the id from above and will do all the interaction with the epaper for me
```

You can see some of the implementations of this abstraction at the bottom of https://esphome.io/components/display/index.html

MrMDavidson:
> Hi @MrMDavidson,
> 
> Did you know i am working on a GUI (switchplate) version for esphome.
> 
> Do you think we could intergrade your menu into my switchplate code at https://github.com/nielsnl68/switchplate

@nielsnl68 Ah, no, I wasn't aware. I was actually independently designing some menu drawing code similar to what's in this PR for a device in my home. Noticed in the esphome 2022.11 release notes the `lcd_menu`, took a quick look, and realised I could tweak my code to fit into that hierarchy and make use of all the hard work @numo68 had put in.

You're welcome to integrate this into your project - but this implementation makes use of the common base of `lcd_menu` - `DisplayMenuComponent` - so there'd be a lot of porting.

nielsnl68:
> @nielsnl68 Ah, no, I wasn't aware. I was actually independently designing some menu drawing code similar to what's in this PR for a device in my home. Noticed in the esphome 2022.11 release notes the `lcd_menu`, took a quick look, and realised I could tweak my code to fit into that hierarchy and make use of all the hard work @numo68 had put in.

I agree on that, he made a very nice component and i am not intended to change that. My idea is more of adding a place holder in my `switchplate` component where the lcd_menu needs to be shown. And make it able to handle screen touches etc.

nagyrobi:
Yeah, have the menu shown as just a part of the graphical screen
menu
and fill the rest of the space with labels, buttons etc the usual way.

nielsnl68:
Hi @MrMDavidson and @numo68,

Some more idea's and questions after looking at the code.

> the menu item setting some kind of a 'dirty' flag indicating that the value to display changed so the positions/dimensions might need to be recalculated.
I use the same mechanisme in my `switchplate` component there is it now named `INVALIDATE_`.  I like the term **dirty** more so i will change it to that as well.

@MrMDavidson, in your code you added the `display_updater` property as well and i was wondering do you have other ideas in mind then just update the display that is set by `display_buffer_`? 
When not, maybe it could be just a boolean? 

The the component name is atm very long. may i suggest a new name: `gfx_menu`?

And as last, i am not sure about your choose to create a `display_page` automatically, could you make that optional. so the end user could decide on which page it wants to show this menu?

I look forward to see your progress.

MrMDavidson:
@nielsnl68 

> @MrMDavidson, in your code you added the `display_updater` property as well and i was wondering do you have other ideas in mind then just update the display that is set by `display_buffer_`? When not, maybe it could be just a boolean?

The intent of this is if you have a display that doesn't automatically redraw (einks set to `update_interval: never`, for instance) it requires an external mechanism for updating. The `DisplayBuffer` itself doesn't have a public API for this. However calling `update` on the `PollingComponent` tends to do this. From my quick investigations because esphome doesn't compile with run time type I couldn't cast-and-null-check the provided `DisplayBuffer` to a `PollingComponent` to do it that way, so need to have it passed in as an actual `PollingComponent` . You'll note whilst the example config passes the same id in to both `display_buffer` and `display_updater` that `display_buffer_` is a `DisplayBuffer` vs `display_updater_` being a `PollingComponent` - I'm not super happy with this aspect of the implementation and am open to ideas.

> 
> The the component name is atm very long. may i suggest a new name: `gfx_menu`?

I'm not sure it applies to component names but the guidelines for contributing to esphome state "It’s preferred to use long variable/function names over short and non-descriptive ones." so was following this here. (My original prototype called this a `display_buffer_menu` but that felt needlessly obtuse to the user of the component)

> And as last, i am not sure about your choose to create a `display_page` automatically, could you make that optional. so the end user could decide on which page it wants to show this menu?

Looking into it - I don't believe this is possible in a straightforward way. A `DisplayPage` doesn't expose any way to update the `display_writer_t`. So it'd require a configuration something like... where the user has to remember to call drawing the menu if required.

```yaml
display:
  - platform: ...
    id: my_display
    pages:
      - id: page_that_might_have_a_menu_sometimes
        lambda: |-
          // Normal drawing here
          it.printf(...);
          it.circle(...);

          id(display_menu).draw_menu_if_active(it);


graphical_display_menu:
  id: display_menu
  page: page_that_might_have_a_menu_sometimes
```

Open to ways this may be achievable, though.

MattW2:
@MrMDavidson , this was a very well timed PR. I saw the recently added Display Menu component and built a bench prototype with the plans of making an ESPHome temperature controller with ssd1306 OLED display and 5 way joystick for local control.  I had several ssd1306 displays so used that thinking it would just work. Only after building and starting to design the menu did I realize Display Menu does not work with non LCD character displays and found your PR. I've forked your code to be able to adapt it to Home Assistant 2022.11.5, I needed to change `aioesphomeapi==12.2.0` otherwise this has been working without issue. 

The core of my menu is several "type: number" selections to input temperature and vary other control parameters. Upon adding the "type: number" I received a continuous string of errors
```
[W][display:176]: Encountered character without representation in font: '['
[W][display:176]: Encountered character without representation in font: ']'
```
Turns out '[' and ']' are used to bracket the value in the number selection. 

In your example code you may want to include the added glyphs over the standard list to save others from having to figure this out all over again. 
```
font:
  - file: "gfonts://Roboto"
    id: roboto
    size: 10
    glyphs:
      ['&', '@', '!', ',', '.', '"', '%', '(', ')', '+', '-', '_', ':', '°', '0',
       '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E',
       'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S',
       'T', 'U', 'V', 'W', 'X', 'Y', 'Z', ' ', 'a', 'b', 'c', 'd', 'e', 'f',
       'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't',
       'u', 'v', 'w', 'x', 'y', 'z','å', 'ä', 'ö', '/', '[', ']']
```
Thanks for the really great addition to ESPHome!

MrMDavidson:
> @MrMDavidson , this was a very well timed PR. I saw the recently added Display Menu component and built a bench prototype with the plans of making an ESPHome temperature controller with ssd1306 OLED display and 5 way joystick for local control. 

That sounds very similar to what I'm building, @MattW2 only I'm using a waveshare eink! I'd basically written very similar menu code when `lcd_menu` was released.

> I've forked your code to be able to adapt it to Home Assistant 2022.11.5, I needed to change `aioesphomeapi==12.2.0` 

I presume a rebase from dev should include this - I'll try and do so today.

> otherwise this has been working without issue.

That's great to hear. Are you using the component permanently or dynamically dismissing the menu and using the display for other purposes? That's something that probably needs more testing as I was focused on layout and basic stuff.

> The core of my menu is several "type: number" selections to input temperature and vary other control parameters. Upon adding the "type: number" I received a continuous string of errors
> 
> ```
> [W][display:176]: Encountered character without representation in font: '['
> [W][display:176]: Encountered character without representation in font: ']'
> ```
> 
> Turns out '[' and ']' are used to bracket the value in the number selection.
> 
> In your example code you may want to include the added glyphs over the standard list to save others from having to figure this out all over again.
> 
> ```
> font:
>   - file: "gfonts://Roboto"
>     id: roboto
>     size: 10
>     glyphs:
>       ['&', '@', '!', ',', '.', '"', '%', '(', ')', '+', '-', '_', ':', '°', '0',
>        '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E',
>        'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S',
>        'T', 'U', 'V', 'W', 'X', 'Y', 'Z', ' ', 'a', 'b', 'c', 'd', 'e', 'f',
>        'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't',
>        'u', 'v', 'w', 'x', 'y', 'z','å', 'ä', 'ö', '/', '[', ']']
> ```

Ah - good call! I wonder if I should switch to ( and ) instead, which are part of the default glyph set? The other option could be to allow this to be specified by the user. If your font includes them you can actually use emojis which adds some nice flair (eg. I've been using 🌡️, 🔥, ❄️, etc for my thermostat controller). But in this case someone may want to use ▶️ and ◀️ to show values or selections.

> Thanks for the really great addition to ESPHome!

Thank @numo68 for the base menu component and whoever added the `DisplayBuffer` abstraction for displays. The implementation to date only took about day or so!

MattW2:
> Ah - good call! I wonder if I should switch to ( and ) instead, which are part of the default glyph set? The other option could be to allow this to be specified by the user. If your font includes them you can actually use emojis which adds some nice flair (eg. I've been using 🌡️, 🔥, ❄️, etc for my thermostat controller). But in this case someone may want to use ▶️ and ◀️ to show values or selections.

Personally I prefer [ and ] over ( and ) but having the character to be used to bracket the numerical values be a parameter does seem like a clean solution to this. 

I would be curious to see example YAML: code with mixed characters and emojies. What font are you using for this?

landonr:
- it would be nice to be able to set the selection color for something other than white. I think global for the whole menu would be fine, but per menu item could be cool
- after selecting a switch to edit it would be nice if the display updated to somehow show it was editing. to me it wasn't immediately clear why I was stuck in that menu

MrMDavidson:
> > Ah - good call! I wonder if I should switch to ( and ) instead, which are part of the default glyph set? The other option could be to allow this to be specified by the user. If your font includes them you can actually use emojis which adds some nice flair (eg. I've been using 🌡️, 🔥, ❄️, etc for my thermostat controller). But in this case someone may want to use ▶️ and ◀️ to show values or selections.
> 
> Personally I prefer [ and ] over ( and ) but having the character to be used to bracket the numerical values be a parameter does seem like a clean solution to this.

I've added a new option for custom value generation. By default it'll use `(value)` so it's in the default glyph set for fonts. Have a play with it with;

```yaml
graphical_display_menu:
  ...
  menu_item_value: !lambda |-
    std::string label = " @";
    label.append(it->get_value_text());
    label.append("@");
    
    return label;
```

For example to surround the value (eg. Of the sensor) with @ symbols. Note: This is untested as my physical device is boot looping and I haven't had a chance to set up a new one.

> I would be curious to see example YAML: code with mixed characters and emojies. What font are you using for this?

Some UI focused desktop fonts will have Emojis and regular characters in the mix. Once you've found one you can just put them in the glyphs for the font;

```yaml
font:
  - file: "fonts/my_font.ttf"
    glyphs: "abcdefg... 🌡💧🔥"
```

MrMDavidson:
> * it would be nice to be able to set the selection color for something other than white. I think global for the whole menu would be fine, but per menu item could be cool

I've added a new optional `foreground_color:` and `background_color:` option. By default these will be `COLOR_ON` and `COLOR_OFF` as per previously. But you could do;

```yaml
color:
  - id: my_menu_foreground
    red: 100%
    green: 0%
    blue: 0%

graphical_display_menu:
  ...
  foreground_color: my_menu_foreground
```

> * after selecting a switch to edit it would be nice if the display updated to somehow show it was editing. to me it wasn't immediately clear why I was stuck in that menu

Are you using `immediate_edit: true` out of curiosity? Is it more obvious if its `false` ? Otherwise `lcd_menu` uses different characters for select vs edit. I like inverted display for selection (at least on einks it looks nice). One option might be if on edit it transfers the inverted section to just the value component of the menu item?

landonr:
> I've added a new optional foreground_color: and background_color: option. By default these will be COLOR_ON and COLOR_OFF as per previously. But you could do;
im getting a crash now with or without foreground_color being provided
```
WARNING Found stack trace! Trying to decode it
WARNING Decoded 0x4008eb74: invoke_abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp32/panic.c:715
WARNING Decoded 0x4008eded: abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp32/panic.c:715
WARNING Decoded 0x4016fdf3: __cxxabiv1::__terminate(void (*)()) at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:112
WARNING Decoded 0x4016fe3a: std::terminate() at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:112
WARNING Decoded 0x4016f503: __cxa_throw at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_throw.cc:87
WARNING Decoded 0x4017099f: std::__throw_logic_error(char const*) at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/src/c++11/functexcept.cc:135 (discriminator 1)
WARNING Decoded 0x400e885f: void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*>(char const*, char const*, std::forward_iterator_tag) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/shared_ptr_base.h:488
 (inlined by) void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct_aux<char const*>(char const*, char const*, std::__false_type) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/basic_string.h:195
 (inlined by) void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*>(char const*, char const*) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/basic_string.h:214
 (inlined by) std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(char const*, std::allocator<char> const&) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/basic_string.h:457
WARNING Decoded 0x400ebab1: esphome::TemplatableValue<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, esphome::display_menu_base::MenuItem const*>::TemplatableValue<decltype(nullptr), 0>(decltype(nullptr)) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/shared_ptr_base.h:488
 (inlined by) setup() at /Users/landon/dev/esphome-remote/.esphome/build/menutestfire/src/esphome/components/graphical_display_menu/graphical_display_menu.h:22
WARNING Decoded 0x400f91d6: loopTask(void*) at /Users/landon/.platformio/packages/framework-arduinoespressif32/cores/esp32/main.cpp:18
WARNING Decoded 0x40090852: vPortTaskWrapper at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/freertos/port.c:355 (discriminator 1)
```

> Are you using immediate_edit: true out of curiosity? Is it more obvious if its false ? Otherwise lcd_menu uses different characters for select vs edit. I like inverted display for selection (at least on einks it looks nice). One option might be if on edit it transfers the inverted section to just the value component of the menu item?

i was using `false`. when its `true` its pretty clear because it switches immediately and i can navigate away. In my project i set the text color to the foreground color when making edits, so the bar disappears while editing

MrMDavidson:
> > I've added a new optional foreground_color: and background_color: option. By default these will be COLOR_ON and COLOR_OFF as per previously. But you could do;
> > im getting a crash now with or without foreground_color being provided
> 
> ```
> WARNING Found stack trace! Trying to decode it
> WARNING Decoded 0x4008eb74: invoke_abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp32/panic.c:715
> WARNING Decoded 0x4008eded: abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp32/panic.c:715
> WARNING Decoded 0x4016fdf3: __cxxabiv1::__terminate(void (*)()) at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:112
> WARNING Decoded 0x4016fe3a: std::terminate() at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:112
> WARNING Decoded 0x4016f503: __cxa_throw at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_throw.cc:87
> WARNING Decoded 0x4017099f: std::__throw_logic_error(char const*) at /builds/idf/crosstool-NG/.build/src/gcc-5.2.0/libstdc++-v3/src/c++11/functexcept.cc:135 (discriminator 1)
> WARNING Decoded 0x400e885f: void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*>(char const*, char const*, std::forward_iterator_tag) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/shared_ptr_base.h:488
>  (inlined by) void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct_aux<char const*>(char const*, char const*, std::__false_type) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/basic_string.h:195
>  (inlined by) void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*>(char const*, char const*) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/basic_string.h:214
>  (inlined by) std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(char const*, std::allocator<char> const&) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/basic_string.h:457
> WARNING Decoded 0x400ebab1: esphome::TemplatableValue<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, esphome::display_menu_base::MenuItem const*>::TemplatableValue<decltype(nullptr), 0>(decltype(nullptr)) at /Users/landon/.platformio/packages/toolchain-xtensa32/xtensa-esp32-elf/include/c++/5.2.0/bits/shared_ptr_base.h:488
>  (inlined by) setup() at /Users/landon/dev/esphome-remote/.esphome/build/menutestfire/src/esphome/components/graphical_display_menu/graphical_display_menu.h:22
> WARNING Decoded 0x400f91d6: loopTask(void*) at /Users/landon/.platformio/packages/framework-arduinoespressif32/cores/esp32/main.cpp:18
> WARNING Decoded 0x40090852: vPortTaskWrapper at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/freertos/port.c:355 (discriminator 1)
> ```

Can I get you to do a "... > Clean Build Files" and try again? If that doesn't work try specifying the `menu_item_value` like in my earlier comment.

> > Are you using immediate_edit: true out of curiosity? Is it more obvious if its false ? Otherwise lcd_menu uses different characters for select vs edit. I like inverted display for selection (at least on einks it looks nice). One option might be if on edit it transfers the inverted section to just the value component of the menu item?
> 
> i was using `false`. when its `true` its pretty clear because it switches immediately and i can navigate away. In my project i set the text color to the foreground color when making edits, so the bar disappears while editing

Are you able to take some photos of this behaviour for me?

landonr:
> Can I get you to do a "... > Clean Build Files" and try again? If that doesn't work try specifying the menu_item_value like in my earlier comment.

i had to go back to commit `6113de0d` for the crash to disappear. here's my config 

```
display:
  - platform: ili9341
    model: M5STACK
    cs_pin: 14
    dc_pin: 27
    reset_pin: 33
    id: my_display
    update_interval: never

graphical_display_menu:
  id: my_menu
  display: my_display
  display_updater: my_display
  font: medium_font
  active: true
  menu_item_value: !lambda |-
    std::string label = " @";
    label.append(it->get_value_text());
    label.append("@");
    return label;
  items:
    - type: label
      text: Normal label
    - type: menu
      text: Child Menu
      items:
        - type: label
          text: A child menu item
        - type: back
          text: Back    
```

MrMDavidson:
> > Can I get you to do a "... > Clean Build Files" and try again? If that doesn't work try specifying the menu_item_value like in my earlier comment.
> 
> i had to go back to commit `6113de0d` for the crash to disappear. here's my config
> 
> ```
> display:
>   - platform: ili9341
>     model: M5STACK
>     cs_pin: 14
>     dc_pin: 27
>     reset_pin: 33
>     id: my_display
>     update_interval: never
> 
> graphical_display_menu:
>   id: my_menu
>   display: my_display
>   display_updater: my_display
>   font: medium_font
>   active: true
>   menu_item_value: !lambda |-
>     std::string label = " @";
>     label.append(it->get_value_text());
>     label.append("@");
>     return label;
>   items:
>     - type: label
>       text: Normal label
>     - type: menu
>       text: Child Menu
>       items:
>         - type: label
>           text: A child menu item
>         - type: back
>           text: Back    
> ```

Finally had some time to investigate this; I think I've fixed your issue @landonr can you rebuild with latest and let me know?

> after selecting a switch to edit it would be nice if the display updated to somehow show it was editing. to me it wasn't immediately clear why I was stuck in that menu

I see what you mean by this now. I'll have a little play in the next few days and see what I come up with. Open to suggestions.

MrMDavidson:
> > Can I get you to do a "... > Clean Build Files" and try again? If that doesn't work try specifying the menu_item_value like in my earlier comment.
> 
> i had to go back to commit `6113de0d` for the crash to disappear. here's my config
> 
> ```
> display:
>   - platform: ili9341
>     model: M5STACK
>     cs_pin: 14
>     dc_pin: 27
>     reset_pin: 33
>     id: my_display
>     update_interval: never
> 
> graphical_display_menu:
>   id: my_menu
>   display: my_display
>   display_updater: my_display
>   font: medium_font
>   active: true
>   menu_item_value: !lambda |-
>     std::string label = " @";
>     label.append(it->get_value_text());
>     label.append("@");
>     return label;
>   items:
>     - type: label
>       text: Normal label
>     - type: menu
>       text: Child Menu
>       items:
>         - type: label
>           text: A child menu item
>         - type: back
>           text: Back    
> ```

Finally had a moment to investigate this. It should be fixed now @landonr 

> after selecting a switch to edit it would be nice if the display updated to somehow show it was editing. to me it wasn't immediately clear why I was stuck in that menu

I see what you mean by this. I'll try and have a look into some alternatives later this week. Open to suggestions on how to visually separate it here.

landonr:
>Finally had a moment to investigate this. It should be fixed now

Thanks works great! @MrMDavidson 

I looked through the display_menu_base code. `_editing` is protected but it could get moved to public so we can do something like this 
```
graphical_display_menu:
  id: my_menu
  display: my_display
  # Optional: For displays that use `update_interval: never` will trigger an `update()` after rendering
  display_updater: my_display
  font: medium_font
  active: true
  foreground_color: my_red
  background_color: my_white
  menu_item_value: !lambda |-
    std::string label = " [";
    if(it->editing()) {
      label.append("*");
    }
    label.append(it->get_value_text());
    label.append("]");
    return label;
```

MrMDavidson:
I noticed that, at least for me, this was double drawing and wasting a lot of time. There's two changes here that may impact people, so if anyone has these configurations, I'd be keen to hear if this breaks things for them;
1. The background of a menu items is now only explicitly drawn when the item is selected (Improved drawing from ~180ms -> 40ms for my menu). If you have a non-eink screen, this might leave _artifacts_? Not sure, let me know.
2. If a `display_updater_` is provided `draw_menu` now just calls `display_updater_->update()` which will then have the drawing lambda fire to call `draw_menu_internal`. Again, this works great for my e-ink and prevents a duplicate of the drawing (eg. 40ms instead of 80ms per iteration). If you're using a non-eink screen/not using a `display_updater` I'd be keen to hear how this works for you.

foxsam21:
I wanted to try this PR with an oled display, I am not sure what the minimal configuration needs to be. I tried compiling but I get an error: 
```
In file included from src/esphome/components/graphical_display_menu/graphical_display_menu.cpp:1:
src/esphome/components/graphical_display_menu/graphical_display_menu.h:43:8: error: 'void esphome::graphical_display_menu::GraphicalDisplayMenu::on_before_show()' marked 'override', but does not override
   void on_before_show() override;
        ^~~~~~~~~~~~~~
src/esphome/components/graphical_display_menu/graphical_display_menu.h:44:8: error: 'void esphome::graphical_display_menu::GraphicalDisplayMenu::on_before_hide()' marked 'override', but does not override
   void on_before_hide() override;
        ^~~~~~~~~~~~~~
```
And here is the relevant part of the config I tried:
```yaml
graphical_display_menu:
  id: my_menu
  display: my_display
  # Optional: For displays that use `update_interval: never` will trigger an `update()` after rendering
  display_updater: my_display
  font: helvetica_12
  active: true
  items:
    - type: label
      text: Normal label
    - type: menu
      text: Child Menu
      items:
        - type: label
          text: A child menu item
        - type: back
          text: Back    
```

MrMDavidson:
> I wanted to try this PR with an oled display, I am not sure what the minimal configuration needs to be. I tried compiling but I get an error:
> 
> ```
> In file included from src/esphome/components/graphical_display_menu/graphical_display_menu.cpp:1:
> src/esphome/components/graphical_display_menu/graphical_display_menu.h:43:8: error: 'void esphome::graphical_display_menu::GraphicalDisplayMenu::on_before_show()' marked 'override', but does not override
>    void on_before_show() override;
>         ^~~~~~~~~~~~~~
> src/esphome/components/graphical_display_menu/graphical_display_menu.h:44:8: error: 'void esphome::graphical_display_menu::GraphicalDisplayMenu::on_before_hide()' marked 'override', but does not override
>    void on_before_hide() override;
>         ^~~~~~~~~~~~~~
> ```
> 
> And here is the relevant part of the config I tried:
> 
> ```yaml
> graphical_display_menu:
>   id: my_menu
>   display: my_display
>   # Optional: For displays that use `update_interval: never` will trigger an `update()` after rendering
>   display_updater: my_display
>   font: helvetica_12
>   active: true
>   items:
>     - type: label
>       text: Normal label
>     - type: menu
>       text: Child Menu
>       items:
>         - type: label
>           text: A child menu item
>         - type: back
>           text: Back    
> ```

Ah! Ever since [c939c55](https://github.com/esphome/esphome/pull/4105/commits/c939c5570cc7d8d6e42fa58fdfce2a4b53150509) you'll also need to have the base `display_menu_base` included in your `external_components` block;

```yaml
external_components:
  - source: github://pr#4105
    components: [ display_menu_base, graphical_display_menu ]
```

Should work @foxsam21

numo68 (approved):
Hi, looks good for me but I don't have any hardware to actually test it on currently so this was only a code review. I also don't have the rights to approve things for the merging itself, you'll need to find someone who does for it.

There is one automated test failing, not sure what that message means.

Thanks for working on this :)

MrMDavidson:
> Hi, looks good for me but I don't have any hardware to actually test it on currently so this was only a code review. I also don't have the rights to approve things for the merging itself, you'll need to find someone who does for it.

Thank you!

> There is one automated test failing, not sure what that message means.

Yeah, I'm a bit stumped on the fix for this. It doesn't seem like the gsl header library is included in the build so I'm not sure what the fix is here.

MrMDavidson:
> > Finally had a moment to investigate this. It should be fixed now
> 
> Thanks works great! @MrMDavidson
> 
> I looked through the display_menu_base code. `_editing` is protected but it could get moved to public so we can do something like this
> 
> ```
> graphical_display_menu:
>   id: my_menu
>   display: my_display
>   # Optional: For displays that use `update_interval: never` will trigger an `update()` after rendering
>   display_updater: my_display
>   font: medium_font
>   active: true
>   foreground_color: my_red
>   background_color: my_white
>   menu_item_value: !lambda |-
>     std::string label = " [";
>     if(it->editing()) {
>       label.append("*");
>     }
>     label.append(it->get_value_text());
>     label.append("]");
>     return label;
> ```

@landonr  It's a bit more complicated than that as editing refers to the whole menu. With the latest version you could tweak your `menu_item_value` to the below to get your desired effect

```yaml
   menu_item_value: !lambda |-
     std::string label = " [";
     if (it->is_item_selected && it->is_menu_editing) {
       label.append("*");
     }
     label.append(it->menu->get_value_text());
     label.append("]");
     return label;
 ```

By default it'll now render a menu item with values as something like "My menu item (selected value)" normally and "My menu item >selected value<" when the item is being edited.

**Comment by MrMDavidson** [comment]
For anyone that has been using this with an eink or other screen that required a forced call to update to render your old;
```yaml
graphical_display_menu:
  # ....
  display_updater: my_display
```

Should be converted to the new `on_redraw` trigger;
```yaml
graphical_display_menu:
  # ...
  on_redraw:
    then:
      component.update: my_display
```

nielsnl68:
@MrMDavidson  i like this idea, when will this event executed? I hope only when there is a change?

MrMDavidson:
> @MrMDavidson i like this idea, when will this event executed? I hope only when there is a change?

Yeah, it hooks into the `update()` override. So should only be in response to the menu needing to be redrawn (eg. User has changed the selection, etc).

This method also allows more control. Instead of blindly updating the display you might throttle updates, for instance. Or you might want to log when they occur. Or get log output on it;

```yaml
on_redraw:
  then:
    - logger.log: 'About to redraw'
    - component.update: my_display
    - logger.log: 'redraw fired'
```

MrMDavidson:
Hey @jesserockz, sorry for the ping, but can I get some assistance here? I'm a bit unsure how to fix the clang issue being warned about  (`gsl::owner`) - and I can find no usages of this anywhere in the codebase.

jesserockz:
I'm away from computer, but I think you shouldn't use `new DisplayPage` but create it with a unique_ptr/shared_ptr if passed around. ( Have not looked at the code in detail yet

MrMDavidson:
> I'm away from computer, but I think you shouldn't use `new DisplayPage` but create it with a unique_ptr/shared_ptr if passed around. ( Have not looked at the code in detail yet

Brilliant, that worked, thanks! Build is all green now 🎉

bzzbzzz:
Do I get this error because I'm using rp2040: board:rpipicow which is not supported, or because I messed up somewhere in the yaml configuration?

> ERROR Error while reading config: Invalid YAML syntax:
> 
> while parsing a block mapping
>   in "/config/esphome/rp-pcw3-oled1.yaml", line 59, column 3:
>       id: my_menu
>       ^
> expected <block end>, but found '<block sequence start>'
>   in "/config/esphome/rp-pcw3-oled1.yaml", line 69, column 5:
>         - type: menu
>         ^

MrMDavidson:
> Do I get this error because I'm using rp2040: board:rpipicow which is not supported, or because I messed up somewhere in the yaml configuration?
> 
> > ERROR Error while reading config: Invalid YAML syntax:
> > while parsing a block mapping
> > in "/config/esphome/rp-pcw3-oled1.yaml", line 59, column 3:
> > id: my_menu
> > ^
> > expected , but found ''
> > in "/config/esphome/rp-pcw3-oled1.yaml", line 69, column 5:
> > - type: menu
> > ^

Looks like you've got some issues in your yaml configuration. The board type shouldn't matter.

MrMDavidson:
Hi @jesserockz, this has been approved by @numo68, but they don't have merge permissions. Is it possible to get someone to have a 2nd look at this and get it merged?

jesserockz:
I have only quickly looked over the code and read the comments, does this have to hook in and create a new DisplayPage? Could it not be rendered inside an existing page?

eg:
```yaml

display:
  ...
  pages:
    - id: my_menu
      lambda:
        id(my_graphical_display_menu).draw(it);
```

MrMDavidson:
> I have only quickly looked over the code and read the comments, does this have to hook in and create a new DisplayPage? Could it not be rendered inside an existing page?
> 
> eg:
> 
> ```yaml
> display:
>   ...
>   pages:
>     - id: my_menu
>       lambda:
>         id(my_graphical_display_menu).draw(it);
> ```

I did look into this. It is possible but then the menu could only be displayed on the provided page (or you'd have to repeat the call on every page). The core menu the graphical menu derives from has the concept of showing/hiding which you'd usually trigger from an external event (eg. button press). To me it makes more sense for the menu to take ownership of the display as needed (it'll toggle back to the prior screen when you hide it again).

|              | Pros                                        | Cons                                                                        |
|--------------|---------------------------------------------|-----------------------------------------------------------------------------|
| **Screen Draws** | Could resize to use a portion of the screen | Needs to be explicitly drawn on each page - less useful for a "pop up menu" |
|              | Could share screen with graphs, etc         |                                                                             |
| **Menu Draws**   | Applies to any page                         | Uses entire screen, cannot draw with other elements (graphs, etc)           |
|              | Automatically sizes to screen               |                                                                             |

I think it comes down to how we see this being used (in conjunction with the hardware). Is this for pop up configuration menus? Or a general display component like a graph? Thoughts?

jesserockz:
Well ultimately it would be up to the user, but I would think you would still dedicate a page to it, but this allows you to customise the page a little bit, borders, title header, date/time etc.

You can still trigger the change to that page with a button press if the display is not already showing that page.

nagyrobi:
> The core menu the graphical menu derives from has the concept of showing/hiding which you'd usually trigger from an external event (eg. button press).

Yes, that's how it works on the LCD display.

Wouldn't it be possible to implement both possibilities with a configurable option? Because a graphical display has so many more possibilities, one might not want to hide it, but instead show it on a _settings_ page, eventually resized only to a part of the screen. 

I think the bigger the screen resolution, the more likely people would want that. And the smaller the resolution, they'd prefer the main hide/show approach on top of all the pages.

MrMDavidson:
Okay, so I can see a couple of options here;
1. We make providing the display to the menu optional. The existing code path is documented as "Pop up mode" (or something). If a user doesn't provide the display then it's "Advanced mode" (or something) and the user draws it like a normal graphical component (`id(my_menu).draw(it, xPos, yPos, width, height)` sort of thing) and they can provide their own placement, headers, borders, etc.
2. We remove the existing mode and document how you'd achieve it ("On press, switch to your settings page, which has a drawing lambda like so.. `id(my_menu).draw(it, 0, 0, width, height)`) and the expectation is it always draws like a normal component
3. Refactor the existing `DisplayPage` ownership/creation/etc out and move it into a `PopupMenu` specialisation. This specialisation would require a a display to be passed to it and would handle page switching, etc, like the current implementation does. The base `GraphicalDisplayMenu` would simply draw when told to.

1 or 2 is the simplest. I'm not personally a huge fan of documentation that requires a user to do something for (what I think) is a common use case - so my inclination is towards 3. Thoughts?

nagyrobi:
1)

jesserockz:
It would be nice to do 1 where the user can have a choice. Personally I think people would use it as a full time menu as well, not just a popup.

nielsnl68:
I'm also a big fan of option **1**.

MrMDavidson:
I've got a first cut of Option 1 implemented; drawing all seems to be fine, handles limited drawing bounds, etc. Next step is to ensure the input related hooks continue to work in the "advanced drawing mode". Will endeavour to get that done over the next few days.

MrMDavidson:
@jesserockz The failure on `test6.yaml` looks like a known bug? I note in the file it's got a comment about waiting for https://github.com/platformio/platform-raspberrypi/pull/36 

Other than this, I think this is ready for review. I've updated the PR description with both drawing modes and will update the corresponding docs branch once people are happy.

I'm not sure if there's a proper way to pass in defines from a yaml? I've followed the convention of having a define (`USE_GRAPHICAL_DISPLAY_MENU` as per `USE_QR_CODE`, etc) and have just been commenting it out for testing.

Edit: For anyone devs that fine this comment about the define with ESPHome the solution is to add it to the code generator in your `to_code` in `__init__.py`. Eg.
```python
async def to_code(config):
    # Normal config setup code...

    cg.add_define("USE_GRAPHICAL_DISPLAY_MENU")
```

MrMDavidson:
As of https://github.com/esphome/esphome/pull/4105/commits/16893a451b1a225449e2c8b551dc5d5a3bf3cbe0 you'll need to add `display`, `graph`, and `font` to your `external_components` if you're testing this;

```yaml
external_components:
  - source: github://pr#4105
    components: [ display_menu_base, graphical_display_menu, display, graph, font ]
```

MrMDavidson:
Hey @jesserockz this has had the latest dev merged into it and updated inline with the `Display`/`DisplayBuffer`, etc, refactor, and is ready for review.

LaneaLucy:
i have some minor problems with popup mode:

1. as you can see on [Discord](https://discord.com/channels/429907082951524364/1139986719740014673/1141061515898789950) the text is overdrawn, like the code thinks the display is bigger then it is. display is SH1106.
2. the hide option should in popup mode switch to the last page or some other page. or at least it should be mentioned that you need to call hide action AND page change action.

with advanced mode i only get "error: 'class esphome::display::Display' has no member named 'menu'" even with "components: [ display_menu_base, graphical_display_menu, display, graph, font ]"

Thx

EDIT:
works with esp8266:
  board: nodemcuv2

MrMDavidson:
> i have some minor problems with popup mode:
> 
> 1. as you can see on [Discord](https://discord.com/channels/429907082951524364/1139986719740014673/1141061515898789950) the text is overdrawn, like the code thinks the display is bigger then it is. display is SH1106.

Sorry, I don't use Discord, can I get you to include the relevant information here? Picture + config would be helpful.

> 2. the hide option should in popup mode switch to the last page or some other page. or at least it should be mentioned that you need to call hide action AND page change action.

That should be the behaviour. If it isn't I've introduced a regression when I added advanced mode. I'll try and validate over the next few days.

> 
> with advanced mode i only get "error: 'class esphome::display::Display' has no member named 'menu'" even with "components: [ display_menu_base, graphical_display_menu, display, graph, font ]"
> 
> Thx
> 
> EDIT: works with esp8266: board: nodemcuv2

To confirm; you got this working? I have a feeling there may a "Clean Build Files" would fix this if not.

LaneaLucy:
> > i have some minor problems with popup mode:
> > 
> > 1. as you can see on [Discord](https://discord.com/channels/429907082951524364/1139986719740014673/1141061515898789950) the text is overdrawn, like the code thinks the display is bigger then it is. display is SH1106.
> 
> Sorry, I don't use Discord, can I get you to include the relevant information here? Picture + config would be helpful.
> 




Config is just plain.

> > 2. the hide option should in popup mode switch to the last page or some other page. or at least it should be mentioned that you need to call hide action AND page change action.
> 
> That should be the behaviour. If it isn't I've introduced a regression when I added advanced mode. I'll try and validate over the next few days.
> 
It isn't right now for me. Workaround I use is to do set page action together with hide action


> > with advanced mode i only get "error: 'class esphome::display::Display' has no member named 'menu'" even with "components: [ display_menu_base, graphical_display_menu, display, graph, font ]"
> > Thx
> > EDIT: works with esp8266: board: nodemcuv2
> 
> To confirm; you got this working? I have a feeling there may a "Clean Build Files" would fix this if not.

Popup mode is working indeed on the nodemcu which is a esp 8266, but with the overdrawing you can see in the pictures. Advanced mode not...

LaneaLucy:
> I have a feeling there may a "Clean Build Files" would fix this if not.

nope

error: 'class esphome::display::Display' has no member named 'menu'

MrMDavidson:
> > I have a feeling there may a "Clean Build Files" would fix this if not.
> 
> nope
> 
> error: 'class esphome::display::Display' has no member named 'menu'

Ahhh! I think I know what this is. I'm not sure if there's a way to pass DEFINEs to the build process and this is all behind a `#ifdef`. I think you'll need to jump into a local copy of the source and commend out the `#ifdef USE_GRAPHICAL_DISPLAY_MENU` / `#endif` blocks.

gaaf:
Hi,

I'm trying to get the menu to popup on the bottom half of the display. If I set the menu to active in the yaml, the menu will show on the bottom half due to the display draw lambda. But after a `leave` and then an `enter`, the menu covers the whole screen. Same if menu starts as inactive initially and gets activated by a `display_menu.show`.

How can I get the menu tot popup only on the bottom half of the display, while my drawing routine can still update the upper half?

I seem to miss an action to just activate the menu, without it overtaking the display (something like `display_menu.activate: oled_menu`).

```yaml
display:
  - platform: ssd1351_spi
    id: oled
    pages:
      - id: main_page
        lambda: |-
          static const int line_height = 14;
          static const int display_width = it.get_width();
          static const int display_height = it.get_height();

          it.print(6, 0*line_height, id(my_font), c_yellow, "Main page");
          it.line(0, line_height+1, display_width, line_height+1, c_red);

          if (id(oled_menu).is_active()) {
            it.menu(0, (int)(display_height / 2), id(oled_menu), display_width , display_height);
          } else {
            // draw rest of screen if no menu
          }
```
```yaml
graphical_display_menu:
  id: oled_menu
  font: my_font
  display: oled
  active: true
  on_redraw:
    then:
      component.update: oled
  mode: rotary
  items:
    -
```
```yaml
binary_sensor:
  - platform: gpio
    id: rotary_button
    internal: true
    pin:
      number: 34
      inverted: true
    filters:
      - delayed_on_off: 5ms
    on_click:
      min_length: 100ms
      then:
        - if:
            condition:
              display_menu.is_active: oled_menu
            then:
              - display_menu.enter: oled_menu
            else:
              - display_menu.show_main: oled_menu
```

MrMDavidson:
> Hi,
> 
> I'm trying to get the menu to popup on the bottom half of the display. If I set the menu to active in the yaml, the menu will show on the bottom half due to the display draw lambda. But after a `leave` and then an `enter`, the menu covers the whole screen. Same if menu starts as inactive initially and gets activated by a `display_menu.show`.

Unfortunately I'm away from my PC for the next couple of weeks so will be unable to look into this issue until then.

gaaf:
> i have some minor problems with popup mode:
> 
>     1.  the text is overdrawn, like the code thinks the display is bigger then it is. display is SH1106.

A fix for the out-of-bounds draw is here: Check https://github.com/MrMDavidson/esphome/pull/125

gaaf:
> > I'm trying to get the menu to popup on the bottom half of the display. If I set the menu to active in the yaml, the menu will show on the bottom half due to the display draw lambda. But after a `leave` and then an `enter`, the menu covers the whole screen. Same if menu starts as inactive initially and gets activated by a `display_menu.show`.
> 
> Unfortunately I'm away from my PC for the next couple of weeks so will be unable to look into this issue until then.

I found the cause. Still had a `display:` configured in the menu's config. After removing that,it works as expected.

beyondthemind91an:
Hi! How's this project going? I'm really excited about this, finally having a menu 🙂
I'm still quite new to GitHub, so I'm not sure how everything works. But, is it just a review that needs to be done for it to be accessible for use?

MrMDavidson:
> Hi! How's this project going? I'm really excited about this, finally having a menu 🙂 I'm still quite new to GitHub, so I'm not sure how everything works. But, is it just a review that needs to be done for it to be accessible for use?

Nope - you can use it as is now (provided you're okay with alpha code) as an external component. You can find details of doing this in my earlier comment - https://github.com/esphome/esphome/pull/4105#issuecomment-1638046897 . Once the PR gets reviewed and accepted it'll go into a future release. 

Interim docs can be found in the PR for the docs project at https://github.com/esphome/esphome-docs/pull/2572

Finally, until ESPHome 2023.12 is released you will need to be using the ESPHome Dev plugin for HomeAssistant (you can run it along side the release version)

beyondthemind91an:
> > Hi! How's this project going? I'm really excited about this, finally having a menu 🙂 I'm still quite new to GitHub, so I'm not sure how everything works. But, is it just a review that needs to be done for it to be accessible for use?
> 
> Nope - you can use it as is now (provided you're okay with alpha code) as an external component. You can find details of doing this in my earlier comment - [#4105 (comment)](https://github.com/esphome/esphome/pull/4105#issuecomment-1638046897) . Once the PR gets reviewed and accepted it'll go into a future release.
> 
> Interim docs can be found in the PR for the docs project at [esphome/esphome-docs#2572](https://github.com/esphome/esphome-docs/pull/2572)
> 
> Finally, until ESPHome 2023.12 is released you will need to be using the ESPHome Dev plugin for HomeAssistant (you can run it along side the release version)

Thank you! I got it to work, finally, on my SSD1306. I first had some trouble with getting the fonts to work, but that was related to my HA setup and not graphical display. I will also try it on a ST7789, if that's supported.

Looking forward to the official release!

MrMDavidson:
> Thank you! I got it to work, finally, on my SSD1306. I first had some trouble with getting the fonts to work, but that was related to my HA setup and not graphical display. I will also try it on a ST7789, if that's supported.

Yup - because the underlying abstraction is the one that all of the displays implement this will work for any of the graphical style displays. I personally use it with the Waveshare eink screens where it works pretty well!
 
> Looking forward to the official release!

Me too! If you encounter any issues please drop a note in here.

esphome/components/graphical_display_menu/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    if CONF_MENU_ITEM_VALUE in config:
+        if isinstance(config[CONF_MENU_ITEM_VALUE], core.Lambda):
+            template_ = await cg.templatable(
+                config[CONF_MENU_ITEM_VALUE],
+                [(MenuItemValueArgumentsConstPtr, "it")],
+                cg.std_string,
+            )
+            cg.add(var.set_menu_item_value(template_))
+        else:
+            cg.add(var.set_menu_item_value(config[CONF_MENU_ITEM_VALUE]))
```
jesserockz:
```suggestion
    if (menu_item_value_config := config.get(CONF_MENU_ITEM_VALUE, None)) is not None:
        if isinstance(menu_item_value_config, core.Lambda):
            template_ = await cg.templatable(
                menu_item_value_config,
                [(MenuItemValueArgumentsConstPtr, "it")],
                cg.std_string,
            )
            cg.add(var.set_menu_item_value(template_))
        else:
            cg.add(var.set_menu_item_value(menu_item_value_config))
```

esphome/components/graphical_display_menu/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+    if CONF_FOREGROUND_COLOR in config:
+        foreground_color = await cg.get_variable(config[CONF_FOREGROUND_COLOR])
+        cg.add(var.set_foreground_color(foreground_color))
```
jesserockz:
```suggestion
    if foreground_color_config := config.get(CONF_FOREGROUND_COLOR):
        foreground_color = await cg.get_variable(foreground_color_config)
        cg.add(var.set_foreground_color(foreground_color))
```

esphome/components/graphical_display_menu/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+    if CONF_BACKGROUND_COLOR in config:
+        background_color = await cg.get_variable(config[CONF_BACKGROUND_COLOR])
+        cg.add(var.set_background_color(background_color))
```
jesserockz:
```suggestion
    if background_color_config := config.get(CONF_BACKGROUND_COLOR):
        background_color = await cg.get_variable(background_color_config)
        cg.add(var.set_background_color(background_color))
```

esphome/components/graphical_display_menu/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+
+    if CONF_DISPLAY in config:
+        drawing_display = await cg.get_variable(config[CONF_DISPLAY])
+        cg.add(var.set_display(drawing_display))
```
jesserockz:
```suggestion
    if display_config := config.get(CONF_DISPLAY):
        drawing_display = await cg.get_variable(display_config)
        cg.add(var.set_display(drawing_display))
```

jesserockz (changes requested):
Sorry for the delay in reviewing. 
Just a few minor python nitpicks, but otherwise good to go.

CODEOWNERS
```diff
@@ -116,6 +116,7 @@ esphome/components/gp8403/* @jesserockz
 esphome/components/gpio/* @esphome/core
 esphome/components/gps/* @coogle
 esphome/components/graph/* @synco
+esphome/components/graphical_display_menu/* @MrMDavidson
```
MrMDavidson, line 119: @jesserockz Thoughts on this? Should I set it to someone else?

jesserockz: Why would it be set to someone else? This is your PR...

jesserockz (approved):

CODEOWNERS
MrMDavidson: image

Mentioned it because the linter complains as I don't have write access to the repo. Not sure if that's an actual issue or not.

PR #4110, opened 2022-11-28
Reduce memory needed to send the services list

bdraco:

Instead of parsing the characteristics and descriptors, adding them to the api response, and than releasing the memory, we avoid the parse by fetching them from `esp_ble_gattc_get_all_char`/`esp_ble_gattc_get_all_descr` which reduces the number of copies of descriptors in memory when sending the services list which is where our high water mark for memory use currently is.

With the v2 connect code linked in https://github.com/esphome/aioesphomeapi/pull/317 we can also do it for `services` which will finally eliminate the two copies of gatt services in memory!

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


bdraco:
Really no good reason to not do the char list now as well since it will reduce the memory pressure even if we have to parse it later, and sending the services list is currently the memory high water mark so any thing we can do here is going to make everything more stable

jesserockz (approved):

PR #4111, opened 2022-11-28
Combine the scanner failure check so we only stop the scanner once

bdraco:
# What does this implement/fix?

If the scanner failed to start and the set scan param failed we could try to stop the scanner twice to get it back in a good state. We now check both at once to ensure we only make one call to esp_ble_gap_stop_scanning per loop.

Move the stop after the start check to ensure we wait to the next loop to start it to give it time to settle.

Fixes the lock not being released if we reject starting the scan.

Fix loop of
```
[10:18:04][E][esp32_ble_tracker:154]: Scan start failed: 1
[10:18:04][D][esp32_ble_tracker:289]: Starting scan...
[10:18:04][E][esp32_ble_tracker:148]: Scan set param failed: 1
[10:18:04][E][esp32_ble_tracker:154]: Scan start failed: 1
[10:18:04][D][esp32_ble_tracker:289]: Starting scan...
[10:18:04][E][esp32_ble_tracker:148]: Scan set param failed: 1
[10:18:04][E][esp32_ble_tracker:154]: Scan start failed: 1
...
```

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [ ] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4109, opened 2022-11-28
Avoid parsing characteristics and descriptors to reduce memory pressure

bdraco:
# What does this implement/fix?

Replaces calls that used BLECharacteristic with esp_ble_gattc*

The final goal is to only use the internal char/services/desc cache since we currently have two copies of it in memory which is getting us to the OOM state (the internal esp gatt services list and the BLEService/BLECharacteristic/etc structures). #4102 is the other piece to get us there to eliminate the need to parse characteristic

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [ ] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4113, opened 2022-11-28
Active BLE connection v3

bdraco:
Needs 

- https://github.com/esphome/esphome/pull/4115
- https://github.com/esphome/esphome/pull/4102
- https://github.com/esphome/esphome/pull/4109
- https://github.com/esphome/esphome/pull/4110
- https://github.com/esphome/esphome/pull/4111
- https://github.com/esphome/esphome/pull/4112

related https://github.com/esphome/aioesphomeapi/pull/317 https://github.com/home-assistant/core/pull/82815

This change versions the connect api with version 3 of connect with or without a cache.

v3 allows the esp to wipe the services list and gatt cache as soon as it sends it to the
client to solve the esp running out of memory.

This bumps the proxy version to `3` so the client knows the device supports
the v3 connect scheme which makes it responsible for flipping the bit on the
gatt descriptor to enable notifications since the esp will have wiped it services
list to save ram.

The client is than responsible for sending everything as a handle only.

If the connection is requested with a cache the esp will skip resolving services and requesting mtu as this can speed up connect times by 70-80% and solves the esp OOM.
IMG_0169 (1)

In a future PR we can reduce the memory pressure some more by avoiding the service parsing and pull the services from `esp_ble_gattc_get_service` when they are using the v2 connection https://github.com/esphome/esphome/pull/4109#issue-1466985797

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -264,16 +296,16 @@ float BLEClientBase::parse_char_value(uint8_t *value, uint16_t length) {
     case 0xD:  // int12.
     case 0xE:  // int16.
       if (length > 2) {
-        return (float) ((int16_t)(value[1] << 8) + (int16_t) value[2]);
+        return (float) ((int16_t) (value[1] << 8) + (int16_t) value[2]);
```
bdraco:
```suggestion
        return (float) ((int16_t)(value[1] << 8) + (int16_t) value[2]);
```

esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -264,16 +296,16 @@ float BLEClientBase::parse_char_value(uint8_t *value, uint16_t length) {
...
       }
     case 0xF:  // int24.
       if (length > 3) {
-        return (float) ((int32_t)(value[1] << 16) + (int32_t)(value[2] << 8) + (int32_t)(value[3]));
+        return (float) ((int32_t) (value[1] << 16) + (int32_t) (value[2] << 8) + (int32_t) (value[3]));
```
bdraco:
```suggestion
        return (float) ((int32_t)(value[1] << 16) + (int32_t)(value[2] << 8) + (int32_t)(value[3]));
```

esphome/components/esp32_ble_client/ble_client_base.cpp
```diff
@@ -264,16 +296,16 @@ float BLEClientBase::parse_char_value(uint8_t *value, uint16_t length) {
...
       }
     case 0x10:  // int32.
       if (length > 4) {
-        return (float) ((int32_t)(value[1] << 24) + (int32_t)(value[2] << 16) + (int32_t)(value[3] << 8) +
-                        (int32_t)(value[4]));
+        return (float) ((int32_t) (value[1] << 24) + (int32_t) (value[2] << 16) + (int32_t) (value[3] << 8) +
+                        (int32_t) (value[4]));
```
bdraco:
```suggestion
        return (float) ((int32_t)(value[1] << 24) + (int32_t)(value[2] << 16) + (int32_t)(value[3] << 8) +
                        (int32_t)(value[4]));
```

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -157,14 +217,50 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
...
+                 connection->get_connection_index(), connection->address_str().c_str());
+        return;
+      } else if (connection->state() == espbt::ClientState::CONNECTING) {
+        ESP_LOGW(TAG, "[%d] [%s] Connection request ignored, already connecting", connection->get_connection_index(),
+                 connection->address_str().c_str());
+        return;
+      } else if (connection->state() == espbt::ClientState::DISCONNECTING) {
+        ESP_LOGW(TAG, "[%d] [%s] Connection request ignored, device is disconnecting",
+                 connection->get_connection_index(), connection->address_str().c_str());
+        return;
```
bdraco, line 214: The logging is probably a little much here because I'm still trying to track down an issue where we request disconnect and it never happens.  The `ESP_GATTC_DISCONNECT_EVT` just doesn't seem to get delivered sometimes but since there is no way for the state to be set to `DISCONNECTING` without `esp_ble_gattc_close` returning ESP_OK it may be a bug in the idf we can't fix.

bdraco:
Aligned the connection version and the `bluetooth_proxy_version` per Jesse's feedback 👍

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -80,6 +80,10 @@ void BluetoothProxy::loop() {
     if (connection->send_service_ == connection->services_.size()) {
       connection->send_service_ = -1;
       api::global_api_server->send_bluetooth_gatt_services_done(connection->get_address());
+      if (connection->connection_type_ == espbt::ConnectionType::V3_WITH_CACHE ||
+          connection->connection_type_ == espbt::ConnectionType::V3_WITHOUT_CACHE) {
+        connection->release_services();
+      }
```
bdraco, line 86: In a followup we can avoid parsing the services in the first place as its still pretty obvious where the uncached connections are on the memory graph, but its not leaking now 👍 

Screenshot 2022-11-28 at 8 22 22 PM

esphome/components/api/api_pb2.cpp
```diff
@@ -6,6 +6,7 @@
 namespace esphome {
 namespace api {
 
+#ifdef HAS_PROTO_MESSAGE_DUMP
```
jesserockz, line 9: This line will get dropped next regenerate, will need to add it to the generate script if we want it in there
https://github.com/esphome/esphome/pull/4119

jesserockz: No need to remove here, I will regenerate once all merged

jesserockz (approved):

PR #4131, opened 2022-12-01
Fix saving light state for restore modes LIGHT_RESTORE_AND_OFF/ON

myhomeiot:
For restore modes LIGHT_RESTORE_AND_OFF/ON if light is change the state (on/off) this always counted as change and will be written to the flash. This PR fix it by set saving light state according restore mode.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/light/light_state.cpp
+++ b/esphome/components/light/light_state.cpp
@@ -276,7 +276,15 @@ void LightState::set_immediately_(const LightColorValues &target, bool set_remot
 void LightState::save_remote_values_() {
   LightStateRTCState saved;
   saved.color_mode = this->remote_values.get_color_mode();
-  saved.state = this->remote_values.is_on();
+  switch (this->restore_mode_) {
+    case LIGHT_RESTORE_AND_OFF:
+    case LIGHT_RESTORE_AND_ON:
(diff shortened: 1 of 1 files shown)


bkaufx:
Was this actually a problem in some situation?  The light component is supposed to handle this in the setup function by replacing the state after loading from flash.  https://github.com/esphome/esphome/blob/dev/esphome/components/light/light_state.cpp#L71-L76

I think it makes more sense to always have the current light state stored accurately in flash and then handle any logic required for proper restoration at the time of restoration.

myhomeiot:
Everything works ok, but for restore modes LIGHT_RESTORE_AND_OFF/ON it's will do unnecessary write to flash for almost every state change of the light (on/off), where written state is useless for restore modes LIGHT_RESTORE_AND_OFF/ON.

By default [`flash_write_interval`](https://esphome.io/components/esphome.html) is 1 minute.

jesserockz (approved):

jesserockz:
This makes sense as the preference code will see the actual raw data being saved has not changed unless you change brightness/colour etc when using these modes.

PR #4132, opened 2022-12-01
Implement a slow sigma-delta modulation based output

Cat-Ion:

A sigma-delta modulation-based output component that toggles a binary output at fixed intervals.

An example is at https://en.wikipedia.org/wiki/Delta-sigma_modulation#/media/File:Pulse-density_modulation_1_period.gif 

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2500

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
output:
  - id: sddac
    platform: sigma_delta_output
    update_interval: 60s
    output: relay_1
```

## Checklist:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -204,6 +204,7 @@ esphome/components/sgp4x/* @SenexCrenshaw @martgras
 esphome/components/shelly_dimmer/* @edge90 @rnauber
 esphome/components/sht4x/* @sjtrny
 esphome/components/shutdown/* @esphome/core @jsuanet
+esphome/components/sigma_delta_output/* @Cat-Ion
 esphome/components/sim800l/* @glmnet
 esphome/components/sm2135/* @BoukeHaarsma23
 esphome/components/sml/* @alengwenus
--- a/esphome/components/sigma_delta_output/__init__.py
+++ b/esphome/components/sigma_delta_output/__init__.py
@@ -1,0 +1,1 @@
+CODEOWNERS = ["@Cat-Ion"]
--- a/esphome/components/sigma_delta_output/output.py
+++ b/esphome/components/sigma_delta_output/output.py
@@ -5,7 +5,6 @@ import esphome.codegen as cg
 from esphome.const import (
     CONF_ID,
     CONF_PIN,
-    CONF_OUTPUT,
     CONF_TURN_ON_ACTION,
     CONF_TURN_OFF_ACTION,
 )
@@ -24,7 +23,6 @@ CONFIG_SCHEMA = cv.All(
     output.FLOAT_OUTPUT_SCHEMA.extend(cv.polling_component_schema("60s")).extend(
         {
             cv.Required(CONF_ID): cv.declare_id(SigmaDeltaOutput),
-            cv.Optional(CONF_OUTPUT): cv.use_id(output.BinaryOutput),
             cv.Optional(CONF_PIN): pins.gpio_output_pin_schema,
             cv.Optional(CONF_STATE_CHANGE_ACTION): automation.validate_automation(
                 single=True
@@ -66,7 +64,3 @@ async def to_code(config):
         await automation.build_automation(
             var.get_turn_off_trigger(), [], config[CONF_TURN_OFF_ACTION]
         )
-
-    if CONF_OUTPUT in config:
-        output_component = await cg.get_variable(config[CONF_OUTPUT])
-        cg.add(var.set_output(output_component))
--- a/esphome/components/sigma_delta_output/output.py
+++ b/esphome/components/sigma_delta_output/output.py
@@ -1,0 +1,35 @@
+from esphome import pins, automation
+from esphome.components import output
+import esphome.config_validation as cv
+import esphome.codegen as cg
+from esphome.const import (
+    CONF_UPDATE_INTERVAL,
+    CONF_ID,
+    CONF_OUTPUT,
+)
+
+DEPENDENCIES = []
+
+
+sigma_delta_output_ns = cg.esphome_ns.namespace("sigma_delta_output")
+SigmaDeltaOutput = sigma_delta_output_ns.class_(
+    "SigmaDeltaOutput", output.FloatOutput, cg.PollingComponent
+)
+
+CONFIG_SCHEMA = cv.All(
+    output.FLOAT_OUTPUT_SCHEMA.extend(cv.polling_component_schema("60s")).extend(
+        {
+            cv.Required(CONF_ID): cv.declare_id(SigmaDeltaOutput),
+            cv.Required(CONF_OUTPUT): cv.use_id(output.BinaryOutput),
+        }
+    ),
+)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await output.register_output(var, config)
+
+    output_component = await cg.get_variable(config[CONF_OUTPUT])
+    cg.add(var.set_output(output_component))
--- a/esphome/components/sigma_delta_output/output.py
+++ b/esphome/components/sigma_delta_output/output.py
@@ -1,9 +1,13 @@
+from esphome import automation, pins
 from esphome.components import output
 import esphome.config_validation as cv
 import esphome.codegen as cg
 from esphome.const import (
     CONF_ID,
+    CONF_PIN,
     CONF_OUTPUT,
+    CONF_TURN_ON_ACTION,
+    CONF_TURN_OFF_ACTION,
 )
 
 DEPENDENCIES = []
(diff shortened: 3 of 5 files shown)


nagyrobi:
Wow, nice!
What kind of `output` does it support? Reading the docs it leads to the impression of acting on a binary output like a pin or switch, but from the config looks like it's still a float output.
Also can the period be set?

Cat-Ion:
> Wow, nice! What kind of `output` does it support? Reading the docs it leads to the impression of acting on a binary output like a pin or switch, but from the config looks like it's still a float output. Also can the period be set?

It acts as a FloatOutput, but connects to a BinaryOutput -- if I got the `cv.Required(CONF_OUTPUT): cv.use_id(output.BinaryOutput),` right and it does what I think it does?

The period is set with `update_interval`.

nagyrobi:
Can you please add `on_turn_on` and `on_turn_off` actions, so it can be used as a plug-in replacement for `slow_pwm`, which is able to ditectly control any needed component?

Cat-Ion:
Sure, I'll add all the other `slow_pwm` stuff while I'm at it.

nagyrobi:
Trying to test on a HA Addon, running into this error at updating the node:

```
INFO Generating C++ source...
  File "/usr/local/bin/esphome", line 33, in <module>
    sys.exit(load_entry_point('esphome', 'console_scripts', 'esphome')())
  File "/esphome/esphome/__main__.py", line 961, in main
    return run_esphome(sys.argv)
  File "/esphome/esphome/__main__.py", line 948, in run_esphome
    rc = POST_CONFIG_ACTIONS[args.command](args, config)
  File "/esphome/esphome/__main__.py", line 396, in command_run
    exit_code = write_cpp(config)
  File "/esphome/esphome/__main__.py", line 176, in write_cpp
    generate_cpp_contents(config)
  File "/esphome/esphome/__main__.py", line 188, in generate_cpp_contents
    CORE.flush_tasks()
  File "/esphome/esphome/core/__init__.py", line 616, in flush_tasks
    self.event_loop.flush_tasks()
  File "/esphome/esphome/coroutine.py", line 246, in flush_tasks
    next(task.iterator)
  File "/esphome/esphome/__main__.py", line 168, in wrapped
    await coro(conf)
  File "/config/esphome/.esphome/external_components/b6989db6/esphome/components/sigma_delta_output/output.py", line 70, in to_code
    output_component = await cg.get_variable(config[CONF_OUTPUT])
KeyError: 'output'
```

Config:

```
substitutions:
  pwm_period: 210s

external_components:
  - source: github://pr#4132
    components: [ sigma_delta_output ]
    refresh: 1min

output:
- platform: sigma_delta_output
  id: heat_01_pwm_actor
  update_interval: ${pwm_period}
  turn_on_action:
    - switch.turn_on: heat_01_sw
  turn_off_action:
    - switch.turn_off: heat_01_sw
```

Cat-Ion:
Thanks, I'd forgotten to check if the now optional output is actually set. Should be fixed now :)

nagyrobi:
Ok, running now for a test drive.

nagyrobi:
Slow PWM:
image

Sigma-Delta:
image

I don't think it's correct because I didn't wait for the PID to stabilize but the tendency seems good.
What I like about the Sigma-Delta is that pulse widths are more consistent, putting less stress on the heating system.

Do I understand correctly that the pulses can be multiples of `update_interval`? Because looking at the graph, it seems to me that the ON states are not less than exactly 3:30 (210s) as in the config above:
image

In `slow_pwm` the `period` parameters allow specifying the time the valve needs to be fully open or close. With a PID controller and `slow_pwm` by applying a 50% duty cycle on this period, we can keep the valve close to half open (with small oscillation).

How can we achieve a similar logic with `update_interval` in this Sigma-Delta? My first guess would be that I should use a much smaller amount of time than the whole period of closing and opening, because the valve will constantly fully open and close instead of staying close to 50%...? But how small, and how will that be related to the valve opening/closing time?

Cat-Ion:
Slow PWM lets you set the period time - the sum of how long your output is active and then inactive. With the sigma delta output you control length of each pulse. The output is either on or off for `update_interval` before being recalculated.

So where you don't get to control the shortest toggling time with a slow PWM (at 210s period and 0.01 output you still only get 2.1 seconds), here you can just set `update_interval` to 30s and will know that your output's never going to be toggled faster than once per 30 seconds.

nagyrobi:
Hmm this needs to be explained better in the docs so people clearley understand the difference. But now I see that I may achieve some results with my valve period/2 or period/3 without too much stress on the heater.

Cat-Ion:
I did my best to explain it in the docs, but feel free to request changes to that. I don't know if there's a better way to show it than with the pictures.

patrickcollins12:
This is really great! I've been working on pid climate recently. From a users perspective this world be best as an option within slow_pwm rather than a separate component? What do you think?

If it is to be kept as a separate component I think the docs need to articulate better when to use slow_pwm and when to use this.

nagyrobi:
I tried to update the docs page to increase the understandability. 

> the docs need to articulate better when to use slow_pwm and when to use this.

This page already references `slow_pwm` more than necessary. As soon as it will merged, I plan to polish the `slow_pwm` docs to mention `sigma_delta_output` as an alternative for the cases where the requirement to provide a `period` is not obvious.

TLDR:
My thoughts (based on my uderstanding so far):
- `slow_pwm`  should be recommended for the cases where `period` can be matched with functional elements of the system, e.g. an electric thermal actuator valve head opening/closing time. Say, if such a valve has an opening time of 4 minutes, a Slow PWM with a level of 50% will switch it on and off every 2 minutes, resulting in the valve being kept half open. Or when a pump has a maximum operation cycle of 3 minutes when an obligatory pause has to be done.
- `sigma_delta_output` should be recommended for the cases where there's no timeframe to be constrained in, thus not only the classic _duty cycle_ concept can apply. E.g. where implementing a density of pulses to show an output level is possible, agnostic from any kind of periodicity.

@Cat-Ion please correct me if I'm wrong. Also, for any other use cases you could recommend, please help us writing down.

Cat-Ion:
> As soon as it will merged, I plan to polish the `slow_pwm` docs to mention `sigma_delta_output` as an alternative for the cases where the requirement to provide a `period` is not obvious.

We could adjust the `slow_pwm` docs in my documentation branch, if you want? It makes sense to me to have both changes in one go.

> TLDR: My thoughts (based on my uderstanding so far):
> 
> * `slow_pwm`  should be recommended for the cases where `period` can be matched with functional elements of the system, e.g. an electric thermal actuator valve head opening/closing time. Say, if such a valve has an opening time of 4 minutes, a Slow PWM with a level of 50% will switch it on and off every 2 minutes, resulting in the valve being kept half open.

What do you mean by opening time? I'm not sure I understand what the head is doing that requires a four minute period. For slow things like a thermal actuator I would probably use my component to keep the number of changes lower. If it needs 10 seconds to change its state, anything below 20 seconds won't do very much, so having a state last for at least a minute or so would probably be better, no?

> Or when a pump has a maximum operation cycle of 3 minutes when an obligatory pause has to be done.

Yes, this is a great use case for `slow_pwm` - setting a maximum duty cycle is not straightforward with `sigma_delta_output`, while here you can just limit the state to e.g. 0.75.

> * `sigma_delta_output` should be recommended for the cases where there's no timeframe to be constrained in, thus not only the classic _duty cycle_ concept can apply. E.g. where implementing a density of pulses to show an output level is possible, agnostic from any kind of periodicity.

I would use `sigma_delta_output` in most cases where there is no maximum duty cycle limitation, but definitely recommend it where changing between on and off is "expensive" or takes some non-negligible time and you want to limit the toggling rate. So not a duty cycle limitation, but a frequency limitation?

nagyrobi:
> What do you mean by opening time?

The time the thermal actuator head needs to open the valve from fully closed to fully open. There are models that take that long - for underfloor heating manifolds, for example.

Cat-Ion:
> > What do you mean by opening time?
> 
> The time the thermal actuator head needs to open the valve from fully closed to fully open. There are models that take that long - for underfloor heating manifolds, for example.

So to fully open and then fully close you need 8 minutes?  That sounds like an excellent example for something where I would use `sigma_delta_output`, no? With `slow_pwm` you can always get a duty cycle that's too fast, like anything below 13% if you use a period of an hour. With a `sigma_delta_output` you could set the update interval to 4 minutes and would know that the valve has fully opened or closed between updates.

nagyrobi:
> We could adjust the `slow_pwm` docs in my documentation branch, if you want? It makes sense to me to have both changes in one go.

Go ahead!

jesserockz (approved):

PR #3004, opened 2022-01-05
slow_pwm: allow to restart a cycle on state change

Chupaka:
# What does this implement/fix? 

I'm trying to leverage slow_pwm to control my underfloor heating actuators (to periodically heat up the floor not to let it to freeze too much when cooking in the room - thermometer goes high up at that moment, for example), but because the period is long (>20 minutes), I'd like to disable the actuator for the last several minutes of those 20 minutes, not at quite random (compared to the switching time) moments of time.

So, here's the option to reset a cycle start time when a new value is written to slow_pwm :)

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#1774

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266

## Example entry for `config.yaml`:

```yaml
output:
  - platform: slow_pwm
    pin: GPIO5
    id: my_slow_pwm
    period: 15s
    restart_cycle_on_state_change: false
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).
  
If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4158, opened 2022-12-07
Expose lambda action to reset a cycle in `slow_pwm`

johndbritton:

In https://github.com/esphome/esphome/pull/3004 the `restart_cycle_on_state_change` configuration option was added to `slow_pwm`. This extracts the code that restarts the cycle to a method and makes that method `public`.

I have a `slow_pwm` that has two control modes:
  1. Climate PID
  2. Manual

I don't want to restart the cycle on state changes when the Climate PID is controlling the `slow_pwm` output but I do want to reset the cycle when when the manual controls are manipulated.

## Types of changes

- [x] Other - refactor

I don't think this is applicable, as I think it will be auto generated.

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
- lambda: "id(my_slow_pwm).reset_cycle();"
```

## Checklist:

@Chupaka It looks like you added the original reset cycle functionality, what do you think of this?

### DIFF
--- a/esphome/components/slow_pwm/slow_pwm_output.cpp
+++ b/esphome/components/slow_pwm/slow_pwm_output.cpp
@@ -70,7 +70,7 @@ void SlowPWMOutput::dump_config() {
 void SlowPWMOutput::write_state(float state) {
   this->state_ = state;
   if (this->restart_cycle_on_state_change_)
-    this->period_start_time_ = millis();
+    this->restart_cycle();
 }
 
 }  // namespace slow_pwm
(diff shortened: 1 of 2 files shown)


jesserockz (approved):

nagyrobi:
@johndbritton Is this modification reflected in the docs?

johndbritton:
@nagyrobi I think it will get added to the doxygen docs automatically.

I'm not really familiar with the conventions around how to document lambda stuff as it seems to be documented differently module by module.

nagyrobi:
It should be added manually as a note (tip). Otherwise people reading the docs website won't know it even exists, and may increase support load by asking questions individually.

Target audience is not sharp C++ programmers...

PR #3509, opened 2022-05-28
Support for LibreTiny platform (RTL8710, BK7231 & other modules)

kuba2k2:
# What does this implement/fix?

This is not a single chip, but rather a universal PlatformIO framework that I'll use to add support for cheap IoT chips found in modules manufactured by Tuya, Inc. These chips are normally incompatible with ESP32/ESP8266 firmwares, and have very limited vendor and community support, so people tend to swap these for ESP-12 modules, which is also not always possible. This is often very troublesome and wasteful.

My project is made to be as compatible with ESP32 Arduino core as possible. This decision was made especially with ESPHome in mind, so that porting it would require minimal changes to its components. The APIs are made to resemble ESP32 APIs closely.

**If you're reading this and have any non-ESP IoT chips - you can be helpful by contacting me in this discussion** (or anywhere else).

### Porting progress
- [x] Boots :clap: 
- [x] WiFi, mDNS, preferences backend
- [x] OTA
- [x] GPIO, PWM (ledc)
- [x] ADC
- [ ] I2C
- [ ] SPI

As I'm totally new to ESPHome's codebase, any reviews of the work done so far will be appreciated.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3159

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] LibreTiny?

## Example entry for `config.yaml`:

```yaml
bk72xx:
  board: cb2s
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4168, opened 2022-12-09
Add API for dashboard to get boards list by platform

kuba2k2:

This PR changes the newly-added dashboard API for board retrieval (I'm really glad you added this!). A new endpoint is added, to allow retrieval of boards by platform. Furthermore, the structure is a bit different - it contains section titles in the response (i.e. "ESP32", or "ESP8266").

I created [a PR in `dashboard` as well](https://github.com/esphome/dashboard/pull/382). It refactors the wizard dialog a bit, making the user first choose "a platform", and, in the next step, "a board" that belongs to this platform. 
There's a mapping with default boards for each platform. If a platform (say, ESP32) has a default board (`esp32dev`), the next dialog page will not be shown - **thus, the current UI flow stays the same for all ESP platforms** after my changes.

These two PRs will greatly simplify integrating the dashboard with [my LibreTuya pull request](https://github.com/esphome/esphome/pull/3509) making the experience more seamless. The dialog for choosing a board looks like this:
obraz

Consider this PR a (hopefully) non-breaking, non-functional change, that just makes the API and UI more extendable.

Note that both PRs would have to be merged simultaneously, as the dashboard can't work without the wizard API changes.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] tested with dashboard in dev mode

## Checklist:

### DIFF
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -728,18 +728,6 @@ def _get_platform_boards(platform, title=None):
 
 class BoardsRequestHandler(BaseHandler):
     @authenticated
-    def get(self):
-        boards = {}
-        boards.update(_get_platform_boards("esp32"))
-        boards.update(_get_platform_boards("esp8266"))
-        boards.update(_get_platform_boards("rp2040", "Raspberry Pi"))
-
-        self.set_header("content-type", "application/json")
-        self.write(json.dumps(boards))
-
-
-class BoardsPlatformRequestHandler(BaseHandler):
-    @authenticated
     def get(self, platform: str):
         boards = _get_platform_boards(platform)
 
@@ -1117,8 +1105,7 @@ def make_app(debug=get_bool_env(ENV_DEV)):
             (f"{rel}json-config", JsonConfigRequestHandler),
             (f"{rel}rename", EsphomeRenameHandler),
             (f"{rel}prometheus-sd", PrometheusServiceDiscoveryHandler),
-            (f"{rel}boards", BoardsRequestHandler),
-            (f"{rel}boards/(.*)", BoardsPlatformRequestHandler),
+            (f"{rel}boards/(.*)", BoardsRequestHandler),
         ],
         **app_settings,
     )
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -693,20 +693,43 @@ class PrometheusServiceDiscoveryHandler(BaseHandler):
         self.write(json.dumps(sd))
 
 
+def _get_platform_boards(platform, title=None):
+    from esphome.components.esp32.boards import BOARDS as ESP32_BOARDS
+    from esphome.components.esp8266.boards import BOARDS as ESP8266_BOARDS
+    from esphome.components.rp2040.boards import BOARDS as RP2040_BOARDS
+
+    boards = {
+        "esp32": ESP32_BOARDS,
+        "esp8266": ESP8266_BOARDS,
+        "rp2040": RP2040_BOARDS,
+    }
+    return {
+        platform: {
+            "title": title or platform.upper(),
+            "items": {
+                key: val[const.KEY_NAME] for key, val in boards[platform].items()
+            },
+        },
+    }
+
+
 class BoardsRequestHandler(BaseHandler):
     @authenticated
     def get(self):
-        from esphome.components.esp32.boards import BOARDS as ESP32_BOARDS
-        from esphome.components.esp8266.boards import BOARDS as ESP8266_BOARDS
-        from esphome.components.rp2040.boards import BOARDS as RP2040_BOARDS
+        boards = {}
+        boards.update(_get_platform_boards("esp32"))
+        boards.update(_get_platform_boards("esp8266"))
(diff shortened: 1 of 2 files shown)


dj-fiorex:
Very good work on this! i want it!

esphome/dashboard/dashboard.py
```diff
@@ -1081,7 +1100,7 @@ def set_extra_headers(self, path):
             (f"{rel}json-config", JsonConfigRequestHandler),
             (f"{rel}rename", EsphomeRenameHandler),
             (f"{rel}prometheus-sd", PrometheusServiceDiscoveryHandler),
-            (f"{rel}boards", BoardsRequestHandler),
+            (f"{rel}boards/(.*)", BoardsRequestHandler),
```
balloob: `.*` is too accepting. We only accept a certain set of platforms and that's only [a-z-]

esphome/dashboard/dashboard.py
```diff
@@ -693,20 +693,39 @@ def get(self):
...
+
+    boards_items = {
+        key: val[const.KEY_NAME]
+        for key, val in platform_boards.items()
+        if variant is None or val[const.KEY_VARIANT] == variant
+    }
+    boards_items = sorted(boards_items.items(), key=lambda item: item[1])
+    return {
+        platform: {
+            "items": dict(boards_items),
```
balloob: this is already a dict.
```suggestion
            "items": boards_items,
```

esphome/dashboard/dashboard.py
```diff
@@ -693,20 +693,39 @@ def get(self):
         self.write(json.dumps(sd))
 
 
+def _get_platform_boards(platform, title=None):
```
balloob: why extract to a function?

esphome/dashboard/dashboard.py
balloob: Why wrap everything in an `"items"` ?

esphome/dashboard/dashboard.py
kuba2k2: This was extracted to a function when the `/boards` endpoint still existed. You're right it should be inlined now.

esphome/dashboard/dashboard.py
kuba2k2: To allow defining board groups (like the vendor grouping, in LibreTuya for example). I know it's not currently used, but sooner or later I'll need this functionality.

> this is already a dict.

No, `boards_items` is a list of tuples. I'll refactor this further to make variable names less ambiguous.

esphome/dashboard/dashboard.py
```diff
@@ -695,20 +695,40 @@ def get(self):
...
+            "esp32s2": ESP32_BOARDS,
+            "esp32s3": ESP32_BOARDS,
+            "esp32c3": ESP32_BOARDS,
+            "esp8266": ESP8266_BOARDS,
+            "rp2040": RP2040_BOARDS,
+        }
+        # filter all ESP32 variants by requested platform
+        for variant, boards in platform_to_boards.items():
+            if not variant.startswith("esp32"):
+                continue
```
balloob: We should only do this for the variant that's chosen. 
```suggestion
        if platform.startswith("esp32"):
            boards = platform_to_boards[variant]
```

esphome/dashboard/dashboard.py
```diff
@@ -695,20 +695,40 @@ def get(self):
...
+                k: v
+                for k, v in boards.items()
+                if v[const.KEY_VARIANT] == variant.upper()
+            }
+
+        # map to a {board_name: board_title} dict
+        platform_boards = {
+            key: val[const.KEY_NAME]
+            for key, val in platform_to_boards[platform].items()
         }
```
balloob:
```suggestion
        if platform.startswith("esp32"):
            boards = {
                k: v
                for k, v in platform_to_boards[variant].items()
                if v[const.KEY_VARIANT] == variant.upper()
            }
        else:
            boards = platform_to_boards[variant]

        # map to a {board_name: board_title} dict
        platform_boards = {
            key: val[const.KEY_NAME]
            for key, val in boards.items()
        }
```

esphome/dashboard/dashboard.py
```diff
@@ -695,20 +695,37 @@ def get(self):
...
+            "esp32s3": ESP32_BOARDS,
+            "esp32c3": ESP32_BOARDS,
+            "esp8266": ESP8266_BOARDS,
+            "rp2040": RP2040_BOARDS,
         }
+        # filter all ESP32 variants by requested platform
+        if platform.startswith("esp32"):
+            boards = {
+                k: v
+                for k, v in platform_to_boards[platform].items()
```
balloob:
```suggestion
            "esp32": ESP32_BOARDS,
            "esp8266": ESP8266_BOARDS,
            "rp2040": RP2040_BOARDS,
        }
        # filter all ESP32 variants by requested platform
        if platform.startswith("esp32"):
            boards = {
                k: v
                for k, v in platform_to_boards["esp32"].items()
```

balloob (approved):

PR #4171, opened 2022-12-11
Speed up bluetooth proxy connections when using esp-idf

bdraco:

We now enable `CONFIG_BT_GATTC_CACHE_NVS_FLASH` by default unless `cache_services` is set to `False` when using `esp-idf` since it allows us to avoid resolving the services on every connection and store them on flash instead. This significantly speeds up connecting to a device once we have already seen it.

This gets us closer to NIMBLE performance since it also uses NVS https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/bluetooth/nimble/index.html#programming-sequence

Users with many BLE devices must reflash via serial to take advantage of the larger NVS storage.

The goal is to avoid hitting this path as its what takes an extra 2-4s to get the connection working
https://github.com/espressif/esp-idf/blob/454aeb3a48ac2b92cfa9d8b6a01d1b53179ec50a/components/bt/host/bluedroid/bta/gatt/bta_gattc_act.c#L697

Since its using NVS and we only write when we see a new device or clear the cache erases should be very infrequent as each esp tends to connect to the same devices over and over.  Assuming a worst case of wiping the cache every 24 hours for 3 devices that should be 26280 write operations per year, and assuming the esp can do 100k writes over the flash we should kill the flash in 1948 years.  Even if the esp can only do 10k writes over the flash it would take 194 years to kill the flash.  If we have very large services list it might take half the time to kill the flash since we end up with a larger data chunk being erased so the absolute worse case should be about 97 years.

As this will result in some additional wear on the flash its only enabled when bluetooth proxies are enabled, and can be disabled with the `cache_services` option

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable): https://github.com/esphome/esphome-docs/pull/2520

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
bluetooth_proxy:
  active: true
  cache_services: true
```

## Checklist:

### DIFF
--- a/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
+++ b/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
@@ -244,22 +244,6 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest 
                  connection->address_str().c_str());
         return;
       }
-      if (msg.request_type == api::enums::BLUETOOTH_DEVICE_REQUEST_TYPE_CONNECT_V3_WITH_CACHE) {
-        connection->set_connection_type(espbt::ConnectionType::V3_WITH_CACHE);
-        ESP_LOGI(TAG, "[%d] [%s] Connecting v3 with cache", connection->get_connection_index(),
-                 connection->address_str().c_str());
-      } else if (msg.request_type == api::enums::BLUETOOTH_DEVICE_REQUEST_TYPE_CONNECT_V3_WITHOUT_CACHE) {
-        // No cache requested so clear the cache
-        esp_ble_gattc_cache_clean(this->remote_bda_);
-        connection->set_connection_type(espbt::ConnectionType::V3_WITHOUT_CACHE);
-        ESP_LOGI(TAG, "[%d] [%s] Connecting v3 without cache", connection->get_connection_index(),
-                 connection->address_str().c_str());
-      } else {
-        // No cache requested so clear the cache
-        esp_ble_gattc_cache_clean(this->remote_bda_);
-        connection->set_connection_type(espbt::ConnectionType::V1);
-        ESP_LOGI(TAG, "[%d] [%s] Connecting v1", connection->get_connection_index(), connection->address_str().c_str());
-      }
       if (msg.has_address_type) {
         connection->remote_bda_[0] = (msg.address >> 40) & 0xFF;
         connection->remote_bda_[1] = (msg.address >> 32) & 0xFF;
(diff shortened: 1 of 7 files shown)


esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -268,6 +256,22 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest
...
       }
+      if (msg.request_type == api::enums::BLUETOOTH_DEVICE_REQUEST_TYPE_CONNECT_V3_WITH_CACHE) {
+        connection->set_connection_type(espbt::ConnectionType::V3_WITH_CACHE);
+        ESP_LOGI(TAG, "[%d] [%s] Connecting v3 with cache", connection->get_connection_index(),
+                 connection->address_str().c_str());
+      } else if (msg.request_type == api::enums::BLUETOOTH_DEVICE_REQUEST_TYPE_CONNECT_V3_WITHOUT_CACHE) {
+        // No cache requested so clear the cache
+        if (msg.has_address_type) {
+          esp_ble_gattc_cache_clean(connection->remote_bda_);
+        }
```
bdraco: There is no point in clearing this here since we are going to request them anyways which will superceed the cache

esphome/components/esp32/__init__.py
```diff
@@ -398,11 +398,12 @@ async def to_code(config):
 
 IDF_PARTITIONS_CSV = """\
 # Name,   Type, SubType, Offset,   Size, Flags
-nvs,      data, nvs,     ,        0x4000,
+nvs,      data, nvs,     ,        0x4000
 otadata,  data, ota,     ,        0x2000,
 phy_init, data, phy,     ,        0x1000,
 app0,     app,  ota_0,   ,      0x1C0000,
 app1,     app,  ota_1,   ,      0x1C0000,
+storage,  data, nvs,     ,       0x69000,
```
bdraco: I guess we are stuck with the default `nvs`

```
host/bluedroid/bta/gatt/bta_gattc_co.c:            if ((status = nvs_open(fname, NVS_READWRITE, &cache_env->cache_addr[*index].cache_fp)) == ESP_OK) {
host/bluedroid/bta/gatt/bta_gattc_co.c:    if ((err_code = nvs_open(cache_addr, NVS_READWRITE, &fp)) == ESP_OK) {
host/bluedroid/bta/gatt/bta_gattc_co.c:        if ((err_code = nvs_open(cache_addr, NVS_READWRITE , fp)) == ESP_OK) {
```

bdraco:
Testing is going well.

This will need docs for the config option

nagyrobi:
> Assuming a worst case of wiping the cache every 24 hours for 3 devices that should be 26280 write operations per year, and assuming the esp can do 100k writes over the flash we should kill the flash in 1948 years.

What am I missing?
100000 total writes supported / 26280 writes per year = 3.8 years

bdraco:
100000 writes of the entire flash.  We are only writing very tiny objects so its only a small slice of that

bdraco:
The esp has wear leveling built-in but realistically the services almost never change unless there is a firmware update so unlikely to be an issue.  Also its per device so its likely the same 3-4 devices we are connecting to over and over so it likely to cache it once and never have to write it again.

https://stackoverflow.com/questions/63780850/esp32-best-way-to-store-data-frequently

https://www.esp32.com/viewtopic.php?p=29525#p29525

nagyrobi:
Can't PSRAM be also used for this task? For boards having it.

bdraco:
> Can't PSRAM be also used for this task?

The only way to avoid this code path (which is what makes all the connections slow) 

https://github.com/espressif/esp-idf/blob/454aeb3a48ac2b92cfa9d8b6a01d1b53179ec50a/components/bt/host/bluedroid/bta/gatt/bta_gattc_act.c#L697 

is to use NVS: `#if (GATTC_CACHE_NVS == TRUE)`

bdraco:
Still working great. Memory is stable

I have IP devices that take longer to respond than Bluetooth now 😂

bdraco:
Moving this to wider production today

jesserockz (approved):

jesserockz:
Also tested a simple OTA to make sure ESPHome still runs even on the old partition tables.

bdraco:
Finished testing. Did a serial reflash on all my production systems.  Everything is working solid with 51 total devices

PR #4193, opened 2022-12-15
revert change in deep sleep for ESP32C3

pistifonok:

Revert cast remove in deep sleep for ESP32C3. Since the gpio_num_t cast remove the pin-wakeup is broken.

Draft, needs testing.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (commented):
This would be casting to the enum based on the pin number as value, then the bit shift will just cast back to the value, so its not going to actually do anything different...??

pistifonok:
> This would be casting to the enum based on the pin number as value, then the bit shift will just cast back to the value, so its not going to actually do anything different...??

Hi, i dont know what is the difference but just tested again and when i remove the "external_components:" from the config, it stops working. Tried just now 3 time back and forth.

"ESPHome version 2023.3.1 compiled on Mar 23 2023, 23:55:24"

Best regards
Pistifonok

PR #4200, opened 2022-12-17
Added Kamstrup Multical 40x component

cfeenstra1024:

Adds support for reading a Kamstrup multical 40x meter. This meter is used with district heating in for example 
The Netherlands. It uses an optical interface connected to an ESP UART.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** n/a

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3513

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
uart:
  baud_rate: 1200
  stop_bits: 2
  tx_pin: GPIO15
  rx_pin: GPIO13

sensor:
  - platform: kamstrup_kmp
    heat_energy:
      name: 'Heat Energy'
    power:
      name: 'Heat Power'
    temp_diff:
      name: 'Heat Temperature Difference'
    flow:
      name: 'Heat Flow'
    update_interval: 60s

```

## Checklist:
  - [x] Tests have been added to verify that the new code works (under `tests/` folder). => in test3.yaml


esphome/components/kamstrup_mc40x/sensor.py
```diff
@@ -0,0 +1,100 @@
...
+CODEOWNERS = ["@cfeenstra1024"]
+DEPENDENCIES = ["uart"]
+
+kamstrup_mc40x_ns = cg.esphome_ns.namespace("kamstrup_mc40x")
+KamstrupMC40xComponent = kamstrup_mc40x_ns.class_(
+    "KamstrupMC40xComponent", cg.PollingComponent, uart.UARTDevice
+)
+
+CONF_HEAT_ENERGY = "heat_energy"
+# CONF_POWER = "power"
```
jesserockz:
```suggestion
```

esphome/components/kamstrup_mc40x/kamstrup_mc40x.cpp
```diff
@@ -0,0 +1,292 @@
...
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace kamstrup_mc40x {
+
+static const char *const TAG = "KamstrupMC40x";
+
+void KamstrupMC40xComponent::setup() {}
+
+void KamstrupMC40xComponent::dump_config() {
```
jesserockz: You should also `LOG_SENSOR` for each child sensor here

esphome/components/kamstrup_mc40x/kamstrup_mc40x.cpp
```diff
@@ -0,0 +1,292 @@
+#include "kamstrup_mc40x.h"
+
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace kamstrup_mc40x {
+
+static const char *const TAG = "KamstrupMC40x";
```
jesserockz:
```suggestion
static const char *const TAG = "kamstrup_mc40x";
```

jesserockz (changes requested):
Units of measurement cannot be "dynamic" They are sent once when Home Assistant connects and will never be sent again until it disconnects and reconnects again, so it is not guaranteed to be correct.

You should define sensible defaults in the `sensor_schema` and any user is able to override those in their own YAML file when they add this if their meter does not match.

cfeenstra1024:
@jesserockz Thanks for your code review and feedback.
I've applied the suggested changes, and also did 2 more things (sorry about that!):

- Renamed component from 'kamstrup_mc40x' to 'kamstrup_kmp'. It turns out this protocol (KMP) is used by more Kamstrup devices than just the MC40x devices. So the new name better covers the functionality.
- Added 5 custom sensors. The idea behind this is that the user can use these to obtain measurements from the device that are not explicitly implemented by this component. If the user happen to know the command code, the custom sensors can be used to read the measurement.

I have fully re-tested the component with the actual hardware.
Also the documentation in the linked PR is updated accordingly.

cfeenstra1024:
Just a comment to keep the PR alive. 
Would be nice if this can be merged ;)

cfeenstra1024:
Just a comment to keep the PR alive again.
Would be very nice if this can be merged ;)

jeroen85:
I am running this integration currently on my Multical 403. Works as expected! Thanks @cfeenstra1024

esphome/components/kamstrup_kmp/sensor.py
```diff
@@ -0,0 +1,158 @@
...
+                device_class=DEVICE_CLASS_EMPTY,
+                state_class=STATE_CLASS_MEASUREMENT,
+                unit_of_measurement=UNIT_EMPTY,
+            ).extend({cv.Required(CONF_COMMAND): cv.hex_uint16_t}),
+            cv.Optional(CONF_CUSTOM_5): sensor.sensor_schema(
+                accuracy_decimals=1,
+                device_class=DEVICE_CLASS_EMPTY,
+                state_class=STATE_CLASS_MEASUREMENT,
+                unit_of_measurement=UNIT_EMPTY,
+            ).extend({cv.Required(CONF_COMMAND): cv.hex_uint16_t}),
```
jesserockz: Something like this would allow the user to specify a list of any number of custom sensors. It just has to be handled in `to_code` and in the cpp to have `std::vector<sensor::Sensor *> custom_sensors` to handle them.

```suggestion
            cv.Optional(CONF_CUSTOM): cv.ensure_list(
                sensor.sensor_schema(
                    accuracy_decimals=1,
                    device_class=DEVICE_CLASS_EMPTY,
                    state_class=STATE_CLASS_MEASUREMENT,
                    unit_of_measurement=UNIT_EMPTY,
                ).extend({cv.Required(CONF_COMMAND): cv.hex_uint16_t})
            ),
```

jesserockz (changes requested):

esphome/components/kamstrup_kmp/sensor.py
cfeenstra1024: Thanks for mentioning this. This is a much neater solution.
Applied the suggested changes and did a full test with the actual hardware.
Doc PR is also updated.

esphome/components/kamstrup_kmp/sensor.py
```diff
@@ -0,0 +1,132 @@
...
+            continue
+        conf = config[key]
+        sens = await sensor.new_sensor(conf)
+        cg.add(getattr(var, f"set_{key}_sensor")(sens))
+
+    # Custom sensors
+    if CONF_CUSTOM in config:
+        for i, conf in enumerate(config[CONF_CUSTOM]):
+            sens = await sensor.new_sensor(conf)
+            cg.add(getattr(var, "set_custom_sensor")(i, sens, conf[CONF_COMMAND]))
```
jesserockz:
```suggestion
        for conf in config[CONF_CUSTOM]:
            sens = await sensor.new_sensor(conf)
            cg.add(var.add_custom_sensor(sens, conf[CONF_COMMAND]))
```

esphome/components/kamstrup_kmp/kamstrup_kmp.h
```diff
@@ -0,0 +1,128 @@
...
+  void set_flow_sensor(sensor::Sensor *sensor) { flow_sensor_ = sensor; }
+  void set_volume_sensor(sensor::Sensor *sensor) { volume_sensor_ = sensor; }
+  void setup() override;
+  void dump_config() override;
+  float get_setup_priority() const override;
+  void update() override;
+  void set_custom_sensor(int i, sensor::Sensor *sensor, uint16_t command) {
+    this->custom_sensors_.insert(this->custom_sensors_.begin() + i, sensor);
+    this->custom_commands_.insert(this->custom_commands_.begin() + i, command);
+  }
```
jesserockz:
```suggestion
  void add_custom_sensor(sensor::Sensor *sensor, uint16_t command) {
    this->custom_sensors_.push_back(sensor);
    this->custom_commands_.push_back(command);
  }
```

esphome/components/kamstrup_kmp/kamstrup_kmp.h
```diff
@@ -0,0 +1,128 @@
...
+    "kCal",  "Mcal", "Gcal", "varh",  "kvarh",   "Mvarh",   "Gvarh",    "VAh",      "kVAh",     "MVAh",
+    "GVAh",  "kW",   "kW",   "MW",    "GW",      "kvar",    "kvar",     "Mvar",     "Gvar",     "VA",
+    "kVA",   "MVA",  "GVA",  "V",     "A",       "kV",      "kA",       "C",        "K",        "l",
+    "m3",    "l/h",  "m3/h", "m3xC",  "ton",     "ton/h",   "h",        "hh:mm:ss", "yy:mm:dd", "yyyy:mm:dd",
+    "mm:dd", "",     "bar",  "RTC",   "ASCII",   "m3 x 10", "ton x 10", "GJ x 10",  "minutes",  "Bitfield",
+    "s",     "ms",   "days", "RTC-Q", "Datetime"};
+
+class KamstrupKMPComponent : public PollingComponent, public uart::UARTDevice {
+ public:
+  void set_heat_energy_sensor(sensor::Sensor *sensor) { heat_energy_sensor_ = sensor; }
```
jesserockz: Please make sure to prefix all member variables and function calls with `this->`
```suggestion
  void set_heat_energy_sensor(sensor::Sensor *sensor) { this->heat_energy_sensor_ = sensor; }
```

jesserockz (changes requested):
The custom sensor stuff doesnt need the index, and can be a little cleaner =)

esphome/components/kamstrup_kmp/kamstrup_kmp.h
cfeenstra1024: applied suggestion

esphome/components/kamstrup_kmp/sensor.py
cfeenstra1024: applied suggestion

esphome/components/kamstrup_kmp/kamstrup_kmp.h
cfeenstra1024: Added all missing `this->`. 
Fully tested again on the actual hardware.

esphome/components/kamstrup_kmp/sensor.py
```diff
@@ -0,0 +1,132 @@
...
+            cv.Optional(CONF_FLOW): sensor.sensor_schema(
+                accuracy_decimals=1,
+                device_class=DEVICE_CLASS_VOLUME,
+                state_class=STATE_CLASS_MEASUREMENT,
+                unit_of_measurement=UNIT_LITRE_PER_HOUR,
+            ),
+            cv.Optional(CONF_VOLUME): sensor.sensor_schema(
+                accuracy_decimals=1,
+                device_class=DEVICE_CLASS_VOLUME,
+                state_class=STATE_CLASS_MEASUREMENT,
```
jeroen85: I get a warning for this Volume state class in HA

`Entity sensor.meterkast_heat_volume (<class 'homeassistant.components.esphome.sensor.EsphomeSensor'>) is using state class 'measurement' which is impossible considering device class ('volume') it is using; expected None or one of 'total', 'total_increasing'; Please update your configuration if your entity is manually configured,`

cfeenstra1024: Hi @jeroen85,
I've tested this again, but I don't get this warning in the HA log. I'm using HA 2023.11.2. You can override the `device_class` or `state_class` with the optional config parameters.  Refer to https://esphome.io/components/sensor/index.html for the details.

cfeenstra1024: OK, I've checked the Kamstrup documentation, and the 'Volume' value is an accumulated or consumed volume. Therefore, I do agree that state_class 'total' would be correct. I've updated the source. Thanks for mentioning.

jesserockz: It should be `total` if it can go up and down, like the volume of a tank, or `total_increasing` if it is only going up (with the possibility of resetting to 0) such as "consumed" volume from pipes.

cfeenstra1024: Changed to `total_increasing`

esphome/components/kamstrup_kmp/kamstrup_kmp.cpp
```diff
@@ -0,0 +1,315 @@
+#include "kamstrup_kmp.h"
+
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace kamstrup_kmp {
+
+static const char *const TAG = "kamstrup_kmp";
+
+void KamstrupKMPComponent::setup() {}
```
jesserockz: No need for an empty override of setup
```suggestion
```

esphome/components/kamstrup_kmp/kamstrup_kmp.cpp
```diff
@@ -0,0 +1,315 @@
...
+void KamstrupKMPComponent::dump_config() {
+  ESP_LOGCONFIG(TAG, "kamstrup_kmp:");
+  if (this->is_failed()) {
+    ESP_LOGE(TAG, "Communication with Kamstrup meter failed!");
+  }
+  LOG_UPDATE_INTERVAL(this);
+
+  if (this->heat_energy_sensor_ != nullptr) {
+    LOG_SENSOR("  ", "Heat Energy", this->heat_energy_sensor_);
+  }
```
jesserockz: There is a nullptr check inside this macro already so its not needed here
```suggestion
  LOG_SENSOR("  ", "Heat Energy", this->heat_energy_sensor_);
```

esphome/components/kamstrup_kmp/kamstrup_kmp.cpp
```diff
@@ -0,0 +1,315 @@
...
+        buffer[buffer_len++] = (uint8_t) data;
+        if (data == 0x0D) {
+          break;
+        }
+      } else {
+        ESP_LOGE(TAG, "Error while reading from UART");
+      }
+    } else {
+      delay(1);
+      timeout--;
```
jesserockz, line 157: If there are 8+ sensors in this component, this can end up timing out for over 2 seconds. If this code is required still, then each sensor read should be put into a queue that is read one at a time in `loop` instead of all at once in the `update` call.

jesserockz (changes requested):

cfeenstra1024:
Applied requested changes, and tested on the actual hardware. 
Also created a fresh (up-to-date) PR for the documentation.

jeroen85:
Tested on my hardware (Multical 403), works OK and as expected

jeroen85:
@jesserockz Can this be merged please?

wimjanse:
Please merge, seems to be working ok

jesserockz (approved):

PR #4203, opened 2022-12-18
Add dfrobot_sen0395 mmwave radar component

niklasweber:

Add component for dfrobot mmwave radar which allows easy configuration via settings action and allows to read state via uart.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** -

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2554

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
external_components:
  - source: github://pr#4203
    components: [ dfrobot_mmwave_radar ]

uart:
  - id: uart_mmwave
    tx_pin: 14
    rx_pin: 27
    baud_rate: 115200

dfrobot_mmwave_radar:
    - id: mmwave
      uart_id: uart_mmwave
    # - id: mmwave_2

binary_sensor:
  - platform: dfrobot_mmwave_radar
    name: Mmwave Detected UART

  - platform: gpio
    name: Mmwave Detected GPIO
    device_class: motion
    pin:
      number: 13
      mode: INPUT_PULLDOWN

button:
  - platform: template
    name: "Update Settings"
    on_press:
      # - dfrobot_mmwave_radar.start
      # - dfrobot_mmwave_radar.stop
      # - dfrobot_mmwave_radar.reset
      - dfrobot_mmwave_radar.settings:
          # id: mmwave
          factory_reset: true
          detection_segments:
            - [0cm, 3m]
          output_latency:
            delay_after_detect: 0s
            delay_after_disappear: 0s
          start_after_power_on: true
          turn_on_led: false
          presence_via_uart: true
          sensitivity: 7
```

## Checklist:


EverythingSmartHome:
Hey @niklasweber I was going to give this a test out to give some feedback but I can't seem to get it to run - have tried using both:

```
external_components:
  - source:
      type: local
      path: my_components
```

And:

```
external_components:
  - source: github://pr#4203
    components: [ dfrobot_mmwave_radar ]
```

But both give the same error when compiling:
```
Failed config

binary_sensor.dfrobot_mmwave_radar: [source /config/esphome/everything-presence2.yaml:110]
  
  Component binary_sensor.dfrobot_mmwave_radar requires component dfrobot_mmwave_radar.
  platform: dfrobot_mmwave_radar
  name: Mmwave Detected via UART
```

I'm not sure if this is something your still working on and is expected, or more likely I'm doing something wrong - I'd be happy to help test if I can get it installed!

niklasweber:
Hi @EverythingSmartHome, thanks for testing my component! There are still some small fixed I need to make like adding codeowners and tests. But in general the code should work. You can find my documentation PR [here](https://github.com/esphome/esphome-docs/pull/2554). You can checkout the documentation preview [here](https://deploy-preview-2554--esphome.netlify.app/components/binary_sensor/dfrobot_mmwave). 

Looks like you are missing the ``dfrobot_mmwave_radar:`` entry. In my PR description above I added an example config which you can use for orientation. Mainly you need the ``uart:`` entry, the ``dfrobot_mmwave_radar:`` entry and a ``binary_sensor`` with ``platform: dfrobot_mmwave_radar``. The integration is only necessary if you want to read the presence state via UART or change settings. If you only read the state via a GPIO binary_sensor, you don't need the mmwave radar component. Hope this helps. Thanks again for taking a look at it and any questions are welcome. I hope my documentation makes it a bit clearer.

EverythingSmartHome:
@niklasweber Told you I was doing something stupid! :D

I did have that in there initially and then for some reason I removed it without thinking - thanks it works now, will give it a test. Good job on implementing the delay on as well as delay off functions and the detection segments, those will be very useful.

Do you think it would be possible to have UI controls for the settings with a UI button that applies the settings in the boxes?

niklasweber:
@EverythingSmartHome That's good to hear!
I thought about making it possible to configure the sensor from the UI. But my main concern is that you might break the sensors flash or eeprom when changing the settings a lot (like adding the settings to some automation or excessively changing the settings manually). 

Changing settings requires the saveCfg command to be run which seems to write the config to some non-volatile memory. I am speculating you could break the sensor memory when writing to it too often. I am very interested in other opinions on that topic. 

I am not aware of an option to automatically create entities like switches / numbers / etc. without specifying them in the yaml config. So currently the only way to change settings from the home assistant GUI would be to create a [template switch](https://esphome.io/components/switch/template.html) / [template number](https://esphome.io/components/number/template.html) / etc. and call the ``dfrobot_mmwave_radar.settings`` action from the corresponding ``turn_on_action:`` , ``turn_off_action:``, ``on_value`` etc. For example:

```yaml
switch:
  - platform: template
    name: "Sensor LED"
    lambda: |-
      ...
    turn_on_action:
      - dfrobot_mmwave_radar.settings:
          turn_on_led: true
    turn_off_action:
      - dfrobot_mmwave_radar.settings:
          turn_on_led: false
```
I didn't test this yet. If it works, I might add it to the documentation. I will also check out if it is possible to automatically create these entries when defining ``dfrobot_mmwave_radar:`` without manually creating switches and numbers for each setting. Has anybody come across an integration which automatically creates entities?

Nonetheless I still think changing settings from the gui is a bad idea or at least should be used with caution.

EverythingSmartHome:
I'm unsure on the longevity of the flash that you raise concerns about so I agree it might be best to be cautious regarding flashing too many times, but I was wondering if it would be possible to have a "Save settings" button that will only commit the settings once pressed. Changing the settings in the UI wouldn't actually save them until you press the save settings button - hopefully I explained that properly.

The other reason for it is that currently with the manual config, as soon as you change any setting for distance or sensitivity, it immediately starts saving to the sensor and then the user tries to change another setting before it has finished saving the first one and sometimes changes aren't always saved. With the "Save settings" button, all settings would be sent at once so this would resolve that issue too.

Perhaps its not currently possible to do this with ESPHome yet - I'm going to try the suggestion above however and will feedback!

EverythingSmartHome:
This was my initial idea was something kinda like this:

```
select:
  - platform: template
    name: "mmWave LED"
    id: mmwave_led
    optimistic: true
    options:
      - "on"
      - "off"
    initial_option: "on"

button:
  - platform: template
    name: Save Settings
    on_press:
      - dfrobot_mmwave_radar.settings:
          turn_on_led: !lambda |-
            if (id(mmwave_led).state.c_str() == "on") {
              return true;
            } else {
              return false;
            }
```

Basically select the options in the UI that you want to set for the sensor, then only when you hit the save settings button would it save.

However the code above tells me that the "turn_on_led" option is not templatable and so I can't try it. Is that something that we could make possible?

niklasweber:
Hi @EverythingSmartHome,

I experimented a bit and tried to automatically create entities to change the settings. There's probably a way, but I wasn't able to automatically create a select entity if ``select:`` is not specified in the config, as the code for the select component is not copied to the build directory if not specified.

Anyways, I asked in the #devs discord channel and I was advised not to automatically create these components as the user loses control over the entities. I like your idea of creating select entities in combination with a save button. Also because a select entity allows to have an option like "No change". I played around with Home Assistant Helpers and created a draft:

settings_draft

(For apply changes I would use a button instead. I couldn't create a helper button.)

The user could set only the option he or she would want to change and keep the other options on "No change" or not specify them in the yaml config if they are never used. When apply changes is pressed, only the options which have something other than "No change" are configured and saved.

Since I was advised not to automatically create these components/entities the user should create them manually. In order to keep the yaml config as clean as possible I would move the logic to the code instead of using lambdas.
Something like this:

```yaml
select:
  - platform: dfrobot_mmwave_radar
    name: "Factory Reset"
    type: set_factory_reset
  - platform: dfrobot_mmwave_radar
    name: "Update detection segments"
    type: update_detection_segments
  - platform: dfrobot_mmwave_radar
    name: "Update output delays"
    type: update_output_delays
  - platform: dfrobot_mmwave_radar
    name: "Update sensitivity"
    type: update_sensitivity
  - platform: dfrobot_mmwave_radar
    name: "LED"
    type: set_led
  - platform: dfrobot_mmwave_radar
    name: "Start After Power On"
    type: set_start_after_power_on
  - platform: dfrobot_mmwave_radar
    name: "Presence via UART"
    type: set_presence_via_uart

number:
  - platform: dfrobot_mmwave_radar
    name: "Delay after detect"
    type: delay_after_detect
  - platform: dfrobot_mmwave_radar
    name: "Delay after disappear"
    type: delay_after_disappear

button:
  - platform: dfrobot_mmwave_radar
    name: "Apply changes"
    type: apply_changes
```

I think using an input text is not supported by esphome, I would use 8 numbers instead and a select to choose something like "No change", "One segment", "Two segments", "Three segments", "Four segments".

I still need to check if I can implement it like this, but it would make the yaml config a bit slimmer and easier.
What do you think?

EverythingSmartHome:
Hey @niklasweber 

The code side of it is a little over my head, but the end result looks good! I might suggest that the detection segments should be done in YAML rather than front-end - the only reason it that it's a little more complex and has the potential for the user to enter the wrong value which might result in the sensor not working properly?

I would also perhaps change the delay on and delay off to sliders as this means its less likely for the user to enter an unsupported value from my experience - I used number boxes like you have above for EP1 but I noted some users were manually typing in values that aren't supported by the sensor. A slider keeps it on fixed values (unless they manually go in and modify the state). Just a suggestion to consider! 

I might also suggest changing the latency values to seconds rather than ms, since the output latency you will generally want to have as a higher value like 15 seconds for example.

For reference, here is the entities for EP1:

image

Finally, I note in the code for this PR that you use the UART command for setting the latency as

```
outputLatency -1 120 400
```

However I have found that command to only work for setting the off latency, for some reason the sensor completely ignores whatever value you give it for the on latency. I tested this and verified it on multiple different DFRobot sensors I have (as you can imagine, I have quite a few!)

A better command for this that is not documented is the ```setLatency``` command, which sets the on and off latency but is done in seconds rather than ms, which saves some calculation. Here is the snippet from EP1 config that I use to set that, and it works very well:

```
set_action:
      - switch.turn_off: mmwave_sensor
      - delay: 1s
      - uart.write: !lambda |-
          std::string mss = "setLatency " + to_string(id(mmwave_on_latency).state) + " " + to_string(id(mmwave_off_latency).state);
          return std::vector<unsigned char>(mss.begin(), mss.end());
      - delay: 1s
      - uart.write: "saveCfg 0x45670123 0xCDEF89AB 0x956128C6 0xDF54AC89"
      - delay: 1s
      - switch.turn_on: mmwave_sensor
```

If you have a sensor you can test with, you can connect directly to UART on it and verify this behaviour with the ```getLatency``` command - you should observe that the outputLatency command will change the off latency value but ignores the on latency value, where as the setLatency command will do both correctly.

If you want to chat more about it then give me a shout on Discord, would be really good to get this into ESPHome with all the functions available!

EverythingSmartHome:
Hey @niklasweber , any progress on this?

niklasweber:
> Hey @niklasweber , any progress on this?

Hi @EverythingSmartHome, unfortunately not. I was struggling quite a bit getting it easy to configure and also to get rid of errors in the GitHub pipeline. I will reduce my scope and only focus on removing errors and enabling templates, so that a configuration like the one you suggested can be written. I am currently on holiday, so I can't work on this. I will continue after my holiday.

functionpointer:
I have been using this integration and i have to say it works great!

Can't wait to have this merged properly :)

nithino:
Any progress in this?

niklasweber:
> Any progress in this? 

Yes, I added new switches of the dfrobot_mmwave_radar platform to make it easy to add configuration controls to home assistant. Not all settings are implemented yet. Currently you can only change settings using the settings action (which now also supports lambdas).

jesserockz:
I started looking at this PR and realised how much I want to change a whole lot of the code to clean it up. 

So currently I have it checked out locally to make quite a few changes...haha

Stay tuned
jesse

niklasweber:
> I started looking at this PR and realised how much I want to change a whole lot of the code to clean it up. 
> 
> 
> 
> So currently I have it checked out locally to make quite a few changes...haha
> 
> 
> 
> Stay tuned
> 
> jesse

Oh great! Thanks for that! Can you summarise what you are planning to change?

jesserockz:
> Oh great! Thanks for that! Can you summarise what you are planning to change?

Mostly moving the switch/sensor code away from the main h/cpp into sub-folders like other components do.
At the same time, modifying some of the codegen to be a bit cleaner. 

I have made a start with `switch` but its not in a place to push a commit yet. Possibly over this weekend I will.

jesserockz:
@niklasweber I have pushed changes that touch basically everything :sweat_smile: 

The most noticeable is the rename of the component to `dfrobot_sen0395` to better reflect this chip.

I have been testing it on the EP1 and it all seems to be working well.

niklasweber:
Thank you @jesserockz  for the cleanup! It looks a lot better now.
All sensor settings can be changed using the settings action.
For ease of use, switches for the PowerCommand, LedModeCommand, UartOutputCommand and SensorCfgStartCommand are implemented (and cleaned up).
I would still like to add components for the remaining commands:

FactoryResetCommand -> Button (or switch) component
SensitivityCommand -> Number component
OutputLatencyCommand -> 2x number component
DetRangeCfgCommand -> 2,4,6 or 8x number component

Also I want to make sure the component states can be saved.

I'll try to find some time for the implementation.
Otherwise the component should be complete.

jesserockz:
If you want this PR merged sooner rather than later, I am happy to merge it as is, and you can make another one for buttons and number etc?

niklasweber:
That would be awesome.  I will add the rest later.

jesserockz:
@niklasweber Could you please update the docs PR to match the changes?

kbx81 (dismissed):

esphome/components/dfrobot_sen0395/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+)
+
+CONF_DFROBOT_SEN0395_ID = "dfrobot_sen0395_id"
+
+CONF_DELAY_AFTER_DETECT = "delay_after_detect"
+CONF_DELAY_AFTER_DISAPPEAR = "delay_after_disappear"
+CONF_DETECTION_SEGMENTS = "detection_segments"
+CONF_OUTPUT_LATENCY = "output_latency"
+CONF_START_AFTER_POWER_ON = "start_after_power_on"
+CONF_TURN_ON_LED = "turn_on_led"
```
kbx81: These don't seem to be used anywhere 🤔

kbx81: Same for `CONF_PRESENCE_VIA_UART` on the line below

esphome/components/dfrobot_sen0395/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+
+dfrobot_sen0395_ns = cg.esphome_ns.namespace("dfrobot_sen0395")
+DfrobotSen0395Component = dfrobot_sen0395_ns.class_(
+    "DfrobotSen0395Component", cg.Component
+)
+
+# Actions
+DfrobotSen0395PowerAction = dfrobot_sen0395_ns.class_(
+    "DfrobotSen0395PowerAction", automation.Action
+)
```
kbx81: This action doesn't seem to have a corresponding `automation.register_action`

esphome/components/dfrobot_sen0395/__init__.py
kbx81: Ah, I see, these were moved over to the `switch` component 😅

jesserockz: Ah yes I removed them from the settings action

jesserockz (approved):

PR #4204, opened 2022-12-19
Handle deprecated sdkconfig option

### LINKED ISSUE #104: [BUG] Lamp not booting up after flashing with latest ESPHome
**Describe the bug**
After flashing latest esphome 2022.12 lamp fails to boot or not connectable. Lamp is not accessible in ha anymore, ip can not be reached.

validation get's me this:
WARNING GPIO12 is a Strapping PIN and should be avoided.
Attaching external pullup/down resistors to strapping pins can cause unexpected failures.
See https://esphome.io/guides/faq.html#why-am-i-getting-a-warning-about-strapping-pins
WARNING GPIO4 is a Strapping PIN and should be avoided.
Attaching external pullup/down resistors to strapping pins can cause unexpected failures.
See https://esphome.io/guides/faq.html#why-am-i-getting-a-warning-about-strapping-pins

**To Reproduce**
1. updated from 2022.11.5 to 2022.12.0

**Expected behavior**
Lamp would work normal

**Additional context**
none

Please investigate if it's a problem with new esphome.

mmakaay:

With ESPHome 2022.12.0 came a new ESP-IDF framework version, in which the sdkconfig option `CONFIG_ESP32_PHY_CALIBRATION_AND_DATA_STORAGE` was deprecated in favour of `CONFIG_ESP_PHY_CALIBRATION_AND_DATA_STORAGE`.

This option is used by the esp32 `ignore_efuse_mac_crc` option to circumvent a piece of code of the framework in which the MAC address of the device is read from efuse. The reason to circumvent it, is that the function that handles the reading operation panics when the CRC (also from ufuse, so immutable) mismatches the MAC address from efuse. Some devices have a wrong CRC, which is why this option was implemented.

The problem with using the deprecated sdkconfig option, is that the panicing code has become active again, resulting in a boot loop after flashing with ESPHome 2022.12.0.

Solution from the PR: use the correct sdkconfig option, based on the framwork version against which the code is compiled.

I tested the change with my Xiaomi Bedside Lamp 2 component.
* Flashing 2022.11.0 to 2022.12.0 -> device goes into boot loop after flashing
* Flashing 2022.11.0 to PR commit -> device works after flashing

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/mmakaay/esphome-xiaomi_bslamp2/issues/104

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/esp32/__init__.py
+++ b/esphome/components/esp32/__init__.py
@@ -356,9 +356,14 @@ async def to_code(config):
 
         if conf[CONF_ADVANCED][CONF_IGNORE_EFUSE_MAC_CRC]:
             cg.add_define("USE_ESP32_IGNORE_EFUSE_MAC_CRC")
-            add_idf_sdkconfig_option(
-                "CONFIG_ESP32_PHY_CALIBRATION_AND_DATA_STORAGE", False
-            )
+            if framework_ver.major >= 4:
+                add_idf_sdkconfig_option(
+                    "CONFIG_ESP_PHY_CALIBRATION_AND_DATA_STORAGE", False
+                )
+            else:
+                add_idf_sdkconfig_option(
+                    "CONFIG_ESP32_PHY_CALIBRATION_AND_DATA_STORAGE", False
+                )
 
         cg.add_define(
             "USE_ESP_IDF_VERSION_CODE",
--- a/esphome/components/esp32/__init__.py
+++ b/esphome/components/esp32/__init__.py
@@ -356,7 +356,7 @@ async def to_code(config):
 
         if conf[CONF_ADVANCED][CONF_IGNORE_EFUSE_MAC_CRC]:
             cg.add_define("USE_ESP32_IGNORE_EFUSE_MAC_CRC")
-            if framework_ver.major >= 4:
+            if (framework_ver.major, framework_ver.minor) >= (4, 4):
                 add_idf_sdkconfig_option(
                     "CONFIG_ESP_PHY_CALIBRATION_AND_DATA_STORAGE", False
                 )


ssieb:
Is there any point in testing the version?  If someone deliberately uses an older version, then that's on them.  Also, it's a very small number of devices involved.

jesserockz:
Esphome with idf has only supported major version 4 or above...

mmakaay:
> Is there any point in testing the version? If someone deliberately uses an older version, then that's on them. Also, it's a very small number of devices involved.

I think that checking the version is the cleanest thing to do here. ESPHome allows users to configure the framework version to use, so it doesn't feel out of bounds to me to do so. It might be a small use case, but without this version test, people that downgrade the framework version might end up with bricked devices. 

In fact, the current hot fix for not letting people brick their lamps, is that I published a config package that forces the use of the old framework. This prevents ESPHome 2022.12.0 from breaking stuff.
When the next version of ESPHome would support only the new sdkconfig option, and people would then flash their new firmware using it, the lamp would again be bricked.

With the version check in place, I can keep the hot fix for now and release a version of my firmware that depends on it.
Once ESPHome has the PR in it, I can create a new version of the firmware, and giver that one a config package that uses the latest version of the framework.
The lock-step release that I have to do, will likely result in some lamps using the old and some lamps using the new framework version, while being compiled with the latest and greatest version of ESPHome.

I'm quickly typing this, because I have to drive home in a few minutes, but I hope I got the point across 😄

jesserockz (approved):

PR #4208, opened 2022-12-19
Handle zero padding anywhere in the combined adv data

### LINKED ISSUE #3913: Bluetooth proxy doesn't work (well) with 2022.12
### The problem

I've updated this morning to 2022.12 and noticed that all of my bluetooth devices failed to show up in home assistant. I only use bluetooth_proxy over 4 esp32 relay boards.

I've tried a few things like changing from arduino to esp-idf and back, I also flashed each through serial (using edge) to make sure the partition table would be done but no change.

I don't really have any more info unfortunately, I didn't see any errors anywhere. Home Assistant just does not see the announces of my switchbot curtains or airthings devices, and keep retrying their setup over and over.

I do have a switchbot contact sensor that did not fail to setup, but it also never sent any new values. That may be normal.
That's the only device I have that does not depend on `active: true`, so it's possible that the issue is only with active devices. I don't have other passive devices here to make sure, sadly.

I downgraded to 2022.11 and everything is back to normal, so it must be related to the bluetooth_proxy change in 2022.12 I imagine

2022.12

Docker

_No response_

ESP32

### Board

doit-devkit-v1 and olimex's esp32-poe, seem to affect both

bluetooth_proxy

### Example YAML snippet

```yaml
bluetooth_proxy:
  active: true
```
```


```txt
Nothing in the logs
```

### Additional information

I have a lot of Switchbot Curtains, and one Airthings Wave+.
The curtains and airthings on the side of the house I updated on all failed to appear in HA after the update.

I kept the other side of the house on 2022.11, at the same time, and the curtains on that side kept working, so it's not a HA issue.
Downgrading all the boards back to 2022.11 fixed everything.

bdraco:

#4162 fixed the case were there was padding between the `adv_data` and the `scan_rsp` data. It looks like we can get `adv_data` with zero padding anywhere.

fixes esphome/issues#3913

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
+++ b/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
@@ -706,11 +706,8 @@ void ESPBTDevice::parse_adv_(const esp_ble_gap_cb_param_t::ble_scan_result_evt_p
   while (offset + 2 < len) {
     const uint8_t field_length = payload[offset++];  // First byte is length of adv record
     if (field_length == 0) {
-      if (offset < param.adv_data_len && param.scan_rsp_len > 0) {  // Zero padded advertisement data
-        offset = param.adv_data_len;
-        continue;
-      }
-      break;
+      offset++;  // Possible zero padded advertisement data
+      continue;
     }
 
     // first byte of adv record is adv record type
--- a/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
+++ b/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
@@ -696,9 +696,6 @@ void ESPBTDevice::parse_scan_rst(const esp_ble_gap_cb_param_t::ble_scan_result_e
   }
 
   ESP_LOGVV(TAG, "Adv data: %s", format_hex_pretty(param.ble_adv, param.adv_data_len + param.scan_rsp_len).c_str());
-  ESP_LOGVV(TAG, "Adv data length: %d", param.adv_data_len);
-  ESP_LOGVV(TAG, "Scan rsp length: %d", param.scan_rsp_len);
-
 #endif
 }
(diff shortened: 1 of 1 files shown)


esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -706,11 +706,8 @@ void ESPBTDevice::parse_adv_(const esp_ble_gap_cb_param_t::ble_scan_result_evt_p
   while (offset + 2 < len) {
     const uint8_t field_length = payload[offset++];  // First byte is length of adv record
     if (field_length == 0) {
-      if (offset < param.adv_data_len && param.scan_rsp_len > 0) {  // Zero padded advertisement data
-        offset = param.adv_data_len;
```
bdraco, line 710: Maybe this was just an off by one bug and this should have been `param.adv_data_len - 1`

bdraco: If this PR fixes the issue we can try that next

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -696,6 +696,9 @@ void ESPBTDevice::parse_scan_rst(const esp_ble_gap_cb_param_t::ble_scan_result_e
   }
 
   ESP_LOGVV(TAG, "Adv data: %s", format_hex_pretty(param.ble_adv, param.adv_data_len + param.scan_rsp_len).c_str());
+  ESP_LOGVV(TAG, "Adv data length: %d", param.adv_data_len);
+  ESP_LOGVV(TAG, "Scan rsp length: %d", param.scan_rsp_len);
+
```
bdraco:
```suggestion
```

jesserockz (approved):

PR #4199, opened 2022-12-16
Fix broken PULSE_METER

cstaahl:
# What does this implement/fix?
Pulse_meter from 2022.10.0 loses track of its internal state. As a result it misses pulses that should have been counted. This PR addresses this by 
- Changing state to keep track of filtered state instead of state of last interrupt
- Adding a pending state
- Changes to edge tracking to increase robustness

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/3730

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
sensor:
  - platform: pulse_meter
    pin:
      number: 4
      mode:
        input: true
        pullup: true
    name: 'Current electricity consumption (pulse_meter)'
    unit_of_measurement: "kW"
    id: electric_power_pulse_meter
    device_class: "power"
    internal_filter_mode: "PULSE"
    internal_filter: 100ms
    accuracy_decimals: 2
    timeout: 180s
    filters:
      - multiply: 0.08
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4231, opened 2022-12-28
Refactor `pulse_meter` to better handle higher frequencies

TrentHouliston:

Fixed the pulse_meter being unable to detect frequencies higher than 1kHz.
The main reason was the old code quantized the times to milliseconds, however, it was also less stable than it could be as it only looked at the last pulse width in each loop.
The new code is able to keep a count of how many pulses occurred between the loops and use them to get a more stable output.

I also refactored the pulse_meter component to simplify it and make it safer with the ISR having only a single set/get pointer pair that is swapped to ensure a consistent state.
This also resulted in much simpler code with fewer moving parts so hopefully, it will be easier to maintain in the future.

I did notice that people were having trouble with the pulse width filtering in a previous version of this code too (https://github.com/esphome/issues/issues/3730 fixed with https://github.com/esphome/esphome/pull/4199).
Since this is a significant refactor if someone who has access to other systems that use the pulse width filtering that can test this new code works that would be appreciated.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266 (Shelly 1PM)
- [ ] RP2040

(On a Shelly 1PM)
```yaml
  - platform: pulse_meter
    pin: GPIO05
    unit_of_measurement: "W"
    state_class: measurement
    name: "Target Power"
    accuracy_decimals: 1
    timeout: 30s
    filters:
      - multiply: 0.03448473917
      - or:
        - delta: 1
        - throttle: 10s
    
    total:
      name: "Target Total Daily Energy"
      device_class: energy
      state_class: total_increasing
      accuracy_decimals: 7
      unit_of_measurement: kWh
      filters:
        - multiply: 5.74745653e-7
        - throttle: 10s
        - heartbeat: 10s
```

## Checklist:


bkaufx:
The pulse_* components are pretty confusing.  How is this different from pulse_counter?

stevebaxter:
@bkaufx - pulse_counter doesn't work well for slow pulse rates (e.g. utility meters), pulse_meter does. The difference is that pulse_counter counts the pulses in a time period, if there are very few pulses (or no pulses) it gives inaccurate results. pulse_meter measures the time between pulses to give accurate results no matter what the pulse rate is.

It's great to see all these enhancements to pulse_meter, I guess in an ideal world it would eventually replace pulse_counter.

bkaufx:
Thanks for the explanation.  It makes more sense now that I looked again with your explanation in mind and realized you are publishing the value in the loop function and not update.  I was thinking you were counting pulses between two update intervals like pulse_counter.

That's always bugged me about pulse_counter.  You can get two different results for a given pulse frequency depending on how the update period hits relative to the pulse period.  I'd love to see it replaced by this.

TrentHouliston:
@bkaufx With the way that it's calculating in pulse_meter even if it were in an update call it would still be accurate (just lower update rate).
I do think replacing the pulse_counter with this would be great long term, the only thing would be finding a way to use the counter on the ESP32 hardware.

I originally came to this component because I was (indirectly) using the pulse counter through the `hlw8012` component and I wouldn't get updates when the power switched on for an update cycle, and if I decreased the update frequency I'd instead get a bunch of garbage.

This one worked much better, however, the device I'm monitoring can output frequencies as slow as basically 0 up to 1-2kHz and up that high this one would have problems with the interrupts colliding with the main code, and the times were short enough that it was somewhat unstable. By the time I finished, I had accidentally refactored the whole thing :P

I intend to do a few follow up PRs after this one (if it's successful) adding a little bit of extra functionality that I need

cstaahl:
I really like many of the features of this refactoring. I haven't had time to review the code in detail yet, but I am slightly worried about regression on some issues that have been fixed in the past. I will take my time in the coming days to look closer at this.

@TrentHouliston one feature that I understand is important to you is the removal of the quantization to 1 ms. I deliberately didn't touch this one in https://github.com/esphome/esphome/pull/4199 as I didn't know why it was there in the first place and I didn't want to add potential errors in that bug fix.

I have been running this [branch](https://github.com/cstaahl/esphome/tree/no_pulse_quantization) locally for a couple of weeks to try to find out if there any potential issues or pitfalls with removing the quantization and I haven't observed any issues yet. I was planning on adding a PR for a cleaned up version of this branch once I was sure that it worked reliably.  Feel free to try if that branch works just as well for your use case as the code in this PR.

I am suspecting that you based your work on the code before https://github.com/esphome/esphome/pull/4199 and that you worked partially in parallel (some of the issues you mentioned are fixed in that PR) with that fix.  I can't decide just yet if building on **this** PR and potentially making additions for some of the issues that may arise (review the comments in the code in https://github.com/esphome/esphome/pull/4199 for hints to what they are) **or** adding some of the very good features in this PR on top of https://github.com/esphome/esphome/pull/4199.

When I review the code in detail I will comment more on some of the issues I worry about regression with and make references at the appropriate places in the code.

TrentHouliston:
The 1ms quantisation is definitely an issue for me since the pulses that I get sometimes exceed 1000/s and my pulses would just cap out. Without the quantisation, I did notice that at the higher frequencies it got a little more unstable. That's the reason that I changed the code so that it would count how many pulses happened between two times (rather than just measuring the pulse width of the last time). That way I get the average pulse width between loop updates when the frequency of the pulses is greater than the frequency of the loop which is more stable.

I may have missed some issues that were fixed in #4199 as I was first testing with whatever is on main ESPHome, however, I did base this code on after that PR.

My plan for addressing #4199 was because I was entirely filtering and counting in the ISR and then sending the filtered updates through to the main loop. It shouldn't ever get out of sync since it only gets information through the get/set pair that the ISR sends through which contains the total count of pulses that happened between the runs of loop.

I've tried to go through all the other issues related to pulse_meter to make sure this code addresses them, however, I may have still missed some.
I'll be having another look over the next 24 hours or so to double-check.

I'm also worried about regression though since I don't have access to any devices that would exhibit the buggy behaviour.

I'll give your other branch a try and see how it goes too

cstaahl:
I just gave this PR a try. I have two pulse_meter sensors configured:

```
sensor:  
  - platform: pulse_meter
    pin:
      number: 4
      mode:
        input: true
        pullup: true
    name: 'Current electricity consumption (pulse_meter)'
    unit_of_measurement: "kW"
    id: electric_power_pulse_meter
    device_class: "power"
    internal_filter_mode: "PULSE"
    internal_filter: 100ms
    accuracy_decimals: 2
    timeout: 180s
    filters:
      - multiply: 0.08
    total:
      name: "electricity_pulse_meter_total"
      unit_of_measurement: "kWh"
      device_class: "energy"
      accuracy_decimals: 3
      filters:
        - multiply: 0.0013333
  
  - platform: pulse_meter
    pin:
      number: 5
      mode:
        input: true
        pullup: true
    name: 'Current gas consumption (pulse_meter)'
    unit_of_measurement: "m³/h"
    id: gas_flow_pulse_meter
    device_class: "gas"
    internal_filter_mode: "PULSE"
    internal_filter: 100ms
    accuracy_decimals: 2
    timeout: 120s
    filters:
      - multiply: 0.6
    total:
      name: "gas_pulse_meter_total"
      unit_of_measurement: "m³"
      device_class: "energy"
      accuracy_decimals: 3
      filters:
        - multiply: 0.01
```

There seems to be an issue with the timeout.

```
05:49:19][D][pulse_meter:079]: No pulse detected for 120s, assuming 0 pulses/min
[05:49:19][D][sensor:126]: 'Current gas consumption (pulse_meter)': Sending state 0.00000 m³/h with 2 decimals of accuracy
[05:49:19][D][pulse_meter:079]: No pulse detected for 120s, assuming 0 pulses/min
[05:49:19][D][sensor:126]: 'Current gas consumption (pulse_meter)': Sending state 0.00000 m³/h with 2 decimals of accuracy
[05:49:19][D][pulse_meter:079]: No pulse detected for 120s, assuming 0 pulses/min
[05:49:19][D][sensor:126]: 'Current gas consumption (pulse_meter)': Sending state 0.00000 m³/h with 2 decimals of accuracy
[05:49:19][D][pulse_meter:079]: No pulse detected for 120s, assuming 0 pulses/min
[05:49:19][D][sensor:126]: 'Current gas consumption (pulse_meter)': Sending state 0.00000 m³/h with 2 decimals of accuracy
[05:49:19][D][pulse_meter:079]: No pulse detected for 120s, assuming 0 pulses/min
[05:49:19][D][sensor:126]: 'Current gas consumption (pulse_meter)': Sending state 0.00000 m³/h with 2 decimals of accuracy
```

I'll leave it running with my electricity meter only (it doesn't ever encounter the timeout) to see if there is any other issue.

Regarding the issues from the past, the short version is (I'll provide more details when I have more time at hand): 
- We cannot rely on the read pin value in the ISR. There is so much fast jitter (atleast on my esp8266 when provoked and evidently this happens with attached sensors as well) that we will read a high value even with falling edges or the other way around.
- Any handling of data and calculation done outside the ISR will be interrupted and values changed by an ISR

TrentHouliston:
Ahh yes, sorry. Timeouts were working properly but I did one final refactor to try and make it clearer to read and accidentally broke them. I didn't notice when I retested because I didn't get any timeouts. I've fixed it now (was just a missing flag).

I tried your branch, and I got similar effects to when I first removed the quantisation which is to say better than with it still in. However, when the frequency gets higher the output gets a little more jittery than what I ended up with. I'm pretty sure it's because I'm averaging all the pulses between loops rather than taking the last one that happened before the loop.

As for your two comments:

1. I did try to put some code in to handle the fast jitter on the pins, if you have any suggestions as to how to handle it better I'm happy to put them in. That will be a problem with any version of the code though.
2. This code is pretty bulletproof when it comes to ISRs now, the transfer of data between the ISR and main loop is done with a single pointer write (the `set = get` line) and since it's all on one simple pointer set it should be impossible to have an ISR  interrupt at a point that would leave it in an inconsistent state

cstaahl:
I have had a closer look at the code and I think we should build on it. However there are changes necessary to the logic of the `pulse_intr` method.

Before I get into the code, a few words on the intended behavior. In pulse_mode the pulse_meter should behave equivalent to the binary_sensor with the `delayed_on_off` filter. 

Side note: For a future PR, harmonizing the behavior between the pulse_meter / pulse_counter and the binary_sensor is something I think we should do in order to avoid confusion. That would involve a breaking change though. Unfortunately I was not aware of the binary_sensor and its configuration options when I first added the pulse_mode.

I dug out my notes on evil pulses, edges and disturbances and digitized it. For reference I've pasted illustrations of some key edge cases at the end of the comment as an image.

One of them may cause an issue, the others I think you have covered with this approach.

Screenshot 2023-01-01 20 21 23

A​ small disturbance in a pulse that is sufficiently ​long for it to have two remaining pulses that are longer than `filter_us` will cause the rising edge in the middle of the pulse to be registered as last_edge_candidate_us_​

I tried out similar approaches but I could not get it to work for all the edge cases without ending up in something very complex that still didn't work. 

I ended up with one variable tracking the filtered state and a second one indicating a pending state change. I think that this is a more robust approach.

Ultimately I think combining your approach with the `get_` and `set_`  pointers to access the state variables, counting the pulses when they occur faster than the loop interval with the pulse ISR logic from my last PR would be the best solution.  

In addition, adding processing of the pulses to the loop method would be a good thing as you don't have to wait for a matching falling edge for a pulse to be registered.

My notes on some tricky edge cases:
pulse_meter

TrentHouliston:
I think that the one case you flagged should work fine with the code here.
In the code here `last_edge_candidate_us_` would be the "unfiltered" edges, while the value that gets put into the set pointer `last_detected_edge_us_` is the filtered state that loop bases its calculations off.
- The (incorrect) rising edge would get set as `last_edge_candidate_us_`
- The falling edge would not count the candidate as the edge since it's not of `filter_us_` length
- The following (correct) rising edge would then be stored as `last_edge_candidate_us_`
- Then the next falling edge would be of the correct length and put that correct `last_edge_candidate_us_` into `set->last_detected_edge_us_`

What we really need though is unit tests. Is there a way to write unit tests in esphome? I wasn't able to find one for components.

If we wanted to replicate the behaviour of `delayed_on_off` we would need to do some extra work. In that setup, you'd need both a pulse width filter and a second value for how long the sensor should be in one state to be considered in that state (the `delayed_on_off` filter).
In the cases, you've shown where the pulse went low/high with both a rising/falling edge detected the current code would consider it to be the beginning/end of a pulse and treat it accordingly. You need that extra filter value to ignore the noise of a specific length.

I think all the cases where an edge is missed because of high frequencies should work fine since it's not just the current pin value, but also the previous value that's used to detect the edge.
That said, I'm currently treating that noise such that I reset the candidate if it's not a full pulse yet which now that I look at it is different to what you did.
https://github.com/esphome/esphome/blob/1d16c752d2109b534a6a49f2c3a3d094919a5e58/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L115-L121

I worked out a way to do processing in the main loop to trigger that the pulse is valid if it were on for a long time, however it makes it trickier with the ISR so I left it out initially.
The code logic is much simpler when the ISR does its own thing and then passes information back to the main loop. ISR->Loop is easy, Loop->ISR is harder.
The best option I could think of for that is that if the main loop notices that a pulse should have occurred based on what the ISR has sent it then it counts the pulse, but flags that it has "preprocessed" a pulse. So that later when the ISR processes that same pulse the main loop should ignore it. Otherwise, if the main loop tries to process variables that the ISR is using it becomes dangerous as the ISR could interrupt and change them at any time either missing or double counting a pulse.

cstaahl:
I'm sorry, I mixed up the cases yesterday after having made some changes to the overview. The one I flagged earlier would be treated correctly. However, this one would not
Screenshot 2023-01-02 09 46 50
With the first falling edge of the disturbance we will have `sensor->last_pin_val_ = true` and both if-clauses will evaluate to true
https://github.com/esphome/esphome/blob/1d16c752d2109b534a6a49f2c3a3d094919a5e58/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L108-L114 will evaluate to true, `sensor->set_->last_detected_edge_us_` will be update and the count incremented.

At the next rising edge, only µs later, this block will evaluate to true
https://github.com/esphome/esphome/blob/1d16c752d2109b534a6a49f2c3a3d094919a5e58/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L104-L106 and the `sensor->last_edge_candidate_us_` will be updated.
Finally at the falling edge (at the end of the middle pulse) `sensor->set_->last_detected_edge_us_` will be update and the count incremented.

I agree about unit testing. There is a folder in the code, https://github.com/esphome/esphome/tree/dev/tests/unit_tests, but I don't know if the framework is operational or how to use it.

> If we wanted to replicate the behaviour of delayed_on_off we would need to do some extra work. In that setup, you'd need both a pulse width filter and a second value for how long the sensor should be in one state to be considered in that state (the delayed_on_off filter).

I disagree. The second filter you mentioned would suffice as it implicitly also limits the pulse width (or rater pulse interval). If you put a minimum length on the pulse length, or time spent in a state, then the pulse width can at minimum be 2 times that time period.
The `EDGE` mode addresses the time between two rising edges, so the pulse width or interval, and the `PULSE` mode addresses the length of the pulses. Currently they are mutually exclusive, but your comment made me think about it. Why not provide the possibility to combine them?
```
internal_filter:
  pulse_interval: 200 ms # (former EDGE mode)
  delayed_on_off: 50 ms # (former PULSE mode)
```
or even
```
internal_filter:
  pulse_interval: 200 ms
  delayed_on: 50 ms
  delayed_off: 60 ms
```
Admittedly the current naming of EDGE and PULSE is not that intuitive and can be improved.

>That said, I'm currently treating that noise such that I reset the candidate if it's not a full pulse yet which now that I look at it is different to what you did.

I guess you refer to these lines:

https://github.com/cstaahl/esphome/blob/e53ef94af8073b1337cf8238be52281f7b73d407/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L130-L134

Resetting the candidate is correct if we want comparable behavior to the binary_sensor.

I understand the complexity with the processing in the loop function and am fine with leaving that out for a start. In that case however we may need a check for when the timeout hits. Otherwise we risk discarding an uncounted pulse if we hit the timeout in a high state.

TrentHouliston:
I understand what you mean with the delayed_on_off now for the pulse filter. The current filter value is the delay value which implicitly defines a pulse width.

I've updated the code so that an edge is only counted if it remains in that state for filter length time. That should give the behaviour you're after.
Except for maybe case 4 where it would take the larger 2nd pulse rather than the smaller first pulse although this would match what delayed_on_off would do.

cstaahl:
>I've updated the code so that an edge is only counted if it remains in that state for filter length time. That should give the behaviour you're after.

This updated code will not handle the "phantom" interrupts correctly.
Screenshot 2023-01-03 06 34 05
Consider the first registered interrupt at the end of the middle pulse. This will be a rising edge while we are in a high pulse (because the first falling edge is missed for some reason).

We will skip this entire clause
https://github.com/esphome/esphome/blob/62ffe7e16ce16468170e85ad343d1674e750c4bb/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L103-L104
and go straight to
https://github.com/esphome/esphome/blob/62ffe7e16ce16468170e85ad343d1674e750c4bb/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L120-L121

In the next interrupt, which is the falling edge, we should actually count the high pulse.
This will be true
https://github.com/esphome/esphome/blob/62ffe7e16ce16468170e85ad343d1674e750c4bb/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L103-L104
However this will not (because we just updated `sensor->last_intr_ `)
https://github.com/esphome/esphome/blob/62ffe7e16ce16468170e85ad343d1674e750c4bb/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L105-L106

Now `sensor->in_pulse_ ` is out of sync and for the next rising edge (the one of the third pulse) this clause will evaluate to true
https://github.com/esphome/esphome/blob/62ffe7e16ce16468170e85ad343d1674e750c4bb/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L112-L113

Now we will increment at the end of low pulses instead at the end of high pulses until another one of these quick interrupts in succession happens and puts the `sensor->in_pulse`_  back in sync.

These are the same issues I struggled with when I made the changes in https://github.com/esphome/esphome/pull/4199 ;)

We are missing a clause like this one
```
    // Ignore false edges that may be caused by bouncing and exit the ISR ASAP
    if (pin_val == sensor->sensor_is_high_) {
      sensor->pending_state_change_ = NONE;
      return;
```
The one I had, resulted in a behavior which is not consistent with delayed_on_off. But if we combine it with what you had in a previous commit:
```
  // We were in a pulse and the pin is still high, a dropout must have happened faster than we could measure
  // If we hadn't yet reached our filter time, reset the candidate to now
  // However if we already reached our filter time ignore it as noise
  else if (sensor->last_pin_val_ && pin_val) {
    if (now - sensor->last_edge_candidate_us_ < sensor->filter_us_) {
      sensor->last_edge_candidate_us_ = now;
    }
```
I think we have something that will work well!

These interrupts are really tricky and the binary_sensor does have it a bit easier because it is polling sensor.
Did you have the chance to look at the unit testing framework available? I'd be happy to write some unit tests with the knowledge I gathered so far. However I don't have any experience of setting up unit tests for a microcontroller.

TrentHouliston:
As far as I can tell the unit testing framework tests the codegen etc, but there isn't anything that tests the components themselves. I also haven't ever set up unit tests for a microcontroller.

This part of the code is in the if statement that checks for when last/current are the same so that the phantom edge would just be ignored entirely (including the last interrupt timing)
https://github.com/esphome/esphome/blob/62ffe7e16ce16468170e85ad343d1674e750c4bb/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L120-L121
So it should be like that pulse doesn't exist at all.

That said with those phantom pulses we might want to reset the last interrupt timing if we haven't passed the filter timeout. I'll go add that in.

TrentHouliston:
Feel free to look again @cstaahl when you have a chance

cstaahl:
The ISR for PULSE looks very good now. I don't have physical access to any of my ESP:s until next week. I will make some tests then where and provoke this phantom interrupts to happen to make sure it works.

One thing I noticed when I was checking the EDGE ISR due to https://github.com/esphome/issues/issues/3602 (there seems to be an issue that is unrelated to your PR) is that it increments the count regardless of pin_value.
https://github.com/esphome/esphome/blob/d1d2fdd79bb7b24cc0f7c631d2261ccf2358c2a8/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L85-L95

This part needs to be added 
```
   // We only look at rising edges  
  const bool pin_val = sensor->isr_pin_.digital_read();
    if (!pin_val) {
      return;
    }
```

TrentHouliston:
I added this in the new code that should take care of the double counting.
Instead of having the interrupt trigger on any change like it previously did it only triggers on rising edges https://github.com/esphome/esphome/blob/d1d2fdd79bb7b24cc0f7c631d2261ccf2358c2a8/esphome/components/pulse_meter/pulse_meter_sensor.cpp#L23

I'm hesitant to add that check for the pin value as for my use case I'm using edge mode and the device outputs pulses that are 38µs long. I'm worried that if I check the pin value in the ISR I might start missing pulses. I'd rather see if changing the ISR mode to rising edge fixes the issue first before checking the pin value.

cstaahl:
Yeah, I missed that you changed how you attach the interrupt. This should then work the same way as before. I worry that the EDGE mode will work only for ideal signals, but that may have its use cases. Do you have any means of verifying that it's counting the pulses correctly in EDGE mode? I suppose you haven't observed any signs to the contrary.

I suspect that the issue https://github.com/esphome/issues/issues/3602 is caused by the "phantom" interrupts at the end of the pulse. This causes some/many/all falling edges (depending on the bounciness) to also have at least one rising edge.

TrentHouliston:
Since my pulses are about 38µs long, I just set my filter length to about 50µs which will kill any extra edges at the other end. Besides that since it's chip to chip communication it's likely as clean as you'll get.

I would say that if you reliably have a pulse that's 10ms or longer the PULSE filter is what you should use. And use EDGE filter for cases where you have very short pulse widths.
It might even make sense to make PULSE the default given that I think that most people are using this for pulses of 100ms wide or more.

If someone were to use the EDGE filter mode I think it's important to mention that the filter value should be longer than your pulse width

cstaahl (approved):
I've made some final checks using hardware and make-shift unit tests. The unit tests (https://github.com/cstaahl/esphome/tree/pulse_meter_unit_tests) are still very much work in progress, but verify that there won't be a regression on some of the issues previously observed with the pulse_meter.

This PR is a great improvement on the pulse_meter sensor. Thank you very much @TrentHouliston

TrentHouliston:
@jesserockz whats the next step for this?

esphome/components/pulse_meter/pulse_meter_sensor.cpp
```diff
@@ -1,74 +1,75 @@
...
 static const char *const TAG = "pulse_meter";
 
+void PulseMeterSensor::set_pin(InternalGPIOPin *pin) { this->pin_ = pin; }
+void PulseMeterSensor::set_filter_us(uint32_t filter) { this->filter_us_ = filter; }
+void PulseMeterSensor::set_timeout_us(uint32_t timeout) { this->timeout_us_ = timeout; }
+void PulseMeterSensor::set_total_sensor(sensor::Sensor *sensor) { this->total_sensor_ = sensor; }
+void PulseMeterSensor::set_filter_mode(InternalFilterMode mode) { this->filter_mode_ = mode; }
+
+void PulseMeterSensor::set_total_pulses(uint32_t pulses) { this->total_pulses_ = pulses; }
+
```
jesserockz: No need to move simple setters into the c++ file

jesserockz (changes requested):
Apart from the below comment, looks fine to me

esphome/components/pulse_meter/pulse_meter_sensor.cpp
TrentHouliston: Should I just have made them into inline functions instead?
Having them in the header not inline would be a potential ODR violation.

I'll go make them inline functions in the header

jesserockz: `inline` is not required inside a `class`

TrentHouliston: I'll remove it

jesserockz (approved):

ssieb:
@TrentHouliston see https://github.com/esphome/issues/issues/4941

PR #4234, opened 2022-12-28
Change 150 microsecond delay to 175 microseconds to fix a low recurre…

hwstar:
…nce of a display initialization error on a 1602 Buy Display part number ERM1602SYG-6 3.3V display operating in 4 bit parallel mode.

Change 150 microsecond delay to 175 microseconds to fix a low recurrence of a display initialization error on a 1602 Buy Display part number ERM1602SYG-6 3.3V display operating in 4 bit parallel mode.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/lcd_base/lcd_display.cpp
+++ b/esphome/components/lcd_base/lcd_display.cpp
@@ -55,6 +55,31 @@ void LCDDisplay::setup() {
     this->write_n_bits(0x03, 4);
     delay(5);
     this->write_n_bits(0x03, 4);
+
+    // Change 150 microsecond delay to 175 microseconds to fix
+    // a low recurrence of a display initialization error on a
+    // 1602 Buy Display part number ERM1602SYG-6
+    // 3.3V display operating in 4 bit parallel mode
+
+    // After about 1 in 20 power ups or resets the display
+    // will not initialize properly using 150 microseconds.
+    // (I am aware of the boot loop counter forcing entry
+    // into safe mode after 10 short reset attempts, so I did not count that)
+    // I tested a range of values from 125 microseconds
+    // to 300 microseconds and determined that
+    // 125 microseconds caused the display to not
+    // initialize with a 100% failure rate.
+    // 155 microseconds worked 100% of the time,
+    // and I settled on 175 microseconds.
+
+    // Now the strange thing is; is that there
+    // is no timing requirement mentioned in the datasheet
+    // for that particular phase of the initialization for my LCD part,
+    // yet there is a 150 microsecond delay in the original lcd_base.cpp file.
+    // So I either missed something in the datasheet,
+    // or there is an undocumented requirement for
+    // a delay at that point during initialization.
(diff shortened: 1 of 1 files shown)


esphome/components/lcd_base/lcd_display.cpp
```diff
@@ -55,13 +55,13 @@ void LCDDisplay::setup() {
...
     this->write_n_bits(0x03, 4);
-    delayMicroseconds(150);
+    delayMicroseconds(175);
     this->write_n_bits(0x02, 4);
   } else {
     this->command_(LCD_DISPLAY_COMMAND_FUNCTION_SET | display_function);
     delay(5);  // 4.1ms
     this->command_(LCD_DISPLAY_COMMAND_FUNCTION_SET | display_function);
-    delayMicroseconds(150);
+    delayMicroseconds(175);
```
dcousens, line 89: probably a comment as to why this magic number, might be helpful

hwstar:
After about 1 in 20 power ups or resets  the display will not initialize properly using 150 microseconds. (I am aware of the boot loop counter forcing entry into safe mode after 10 short reset attempts, so I did not count that) I tested  a range of values from 125 microseconds to 300 microseconds and determined that 125 microseconds caused the display to not initialize with a 100% failure rate. 155 microseconds worked 100% of the time, and I settled on 175 microseconds.

Now the strange thing is; is that there is no timing requirement mentioned in the datasheet for that particular phase of the initialization for my LCD part, yet there is a 150 microsecond delay in the original lcd_base.cpp file. So I either missed something in the datasheet, or there is an undocumented requirement for a delay at that point during initialization.

esphome/components/lcd_base/lcd_display.cpp
hwstar: After about 1 in 20 power ups or resets the display will not initialize properly using 150 microseconds. (I am aware of the boot loop counter forcing entry into safe mode after 10 short reset attempts, so I did not count that) I tested a range of values from 125 microseconds to 300 microseconds and determined that 125 microseconds caused the display to not initialize with a 100% failure rate. 155 microseconds worked 100% of the time, and I settled on 175 microseconds.

Now the strange thing is; is that there is no timing requirement mentioned in the datasheet for that particular phase of the initialization for my LCD part, yet there is a 150 microsecond delay in the original lcd_base.cpp file. So I either missed something in the datasheet, or there is an undocumented requirement for a delay at that point during initialization.

dcousens: thanks for the explanation @hwstar,  you could probably put that complete comment into the code? :yellow_heart:

hwstar: Discussion added as comments.

PR #4244, opened 2022-12-30
increase size filter range

max246:

Increase size of the filter value from uint8 to unit32 to allow a large number than 255

## Types of changes

- [x ] Bugfix (non-breaking change which fixes an issue)
- [ ] Other

**Related issue or feature (if applicable):** fixes <[link to issue](https://github.com/esphome/issues/issues/3537)>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
remote_receiver:
  pin: D0
  dump:
    - rc_switch
  tolerance: 50%
  filter: 350us
  idle: 4ms
  buffer_size: 2kb
```

## Checklist:

### DIFF
--- a/esphome/components/remote_receiver/remote_receiver.h
+++ b/esphome/components/remote_receiver/remote_receiver.h
@@ -20,7 +20,7 @@ struct RemoteReceiverComponentStore {
   uint32_t buffer_read_at{0};
   bool overflow{false};
   uint32_t buffer_size{1000};
-  uint8_t filter_us{10};
+  uint32_t filter_us{10};
   ISRInternalGPIOPin pin;
 };
 #endif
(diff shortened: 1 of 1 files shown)


max246:
Can it be merge?

jesserockz (changes requested):
The config validation side only allows up to 255 as a value, also the code that uses the filter in RMT is `uint8_t filter_ticks_thresh`, so not sure this will work as intended?

max246:
> The config validation side only allows up to 255 as a value, also the code that uses the filter in RMT is `uint8_t filter_ticks_thresh`, so not sure this will work as intended?

Right that means this requests can be done if RMT wont update their drivers.

Should I just cancel the PR or is there another way you can think of it can be done?

jesserockz:
Yeah unless there is a different method in rmt, we cant achieve this in ESPHome right now.

Best to close the PR for now. I dont have time to look into another solution right now, but may visit in the future. Happy to look again if you find something else.

PR #4249, opened 2023-01-01
Add always trigger stop

### LINKED ISSUE #2045: Time-based cover: Allow triggering the stop action even if has_end_stop_sensor is true
I've a cover with 2 relays: up and down that are managed by a ESP8266.
It has a classic up and down button close to the window for manual operation (for better WAF). 
If I trigger to close the cover with HA, it'll toggle to `CLOSED` state (as expected). However, when someone use the physical button to open it, HA isn't aware of it and nor is ESPHome. 
So the next time I use the "close" action in HA, nothing happens and the cover doesn't close (unless I use `has_end_stop_sensor` variable).
If I'm using the `has_end_stop_sensor` the cover will close (as expected), but the relay remained powered after the clover is closed because the stop action isn't called. So if someone press the physical "open" button, it'll have both phases on the motor which isn't good (on my motor, it triggers the programming mode where it sets the new end/stop position, but on many system it's forbidden, it'll trip the fuse).

So ideally, I'd like to be able to run the stop action (that's releasing the relay) after any action is performed (open or closed) **and** use the `has_end_stop_sensor` option since HA can't really know the position of the cover and only get an estimate of it.

If I don't use the `cover` type but a plain button, I can set a time based button to open or close the cover, but then HA doesn't know it's a cover, and there's no tracking of the cover state. So in that case, I can't set the cover midway. If I don't add a delay, then I have to use a switch and it requires manipulating the interface twice without forgetting to release the relay.

**Additional context**
The code that would need to change is in `time_based_cover.cpp`, around line 34, that should read:
```cpp
  if (this->is_at_target_()) {
    if (this->has_built_in_endstop_ && !this->always_trigger_stop_ &&
        (this->target_position_ == COVER_OPEN || this->target_position_ == COVER_CLOSED)) {
      // Don't trigger stop, let the cover stop by itself.
      this->current_operation = COVER_OPERATION_IDLE;
    } else {
      this->start_direction_(COVER_OPERATION_IDLE);
    }
```
And a new configuration option `always_trigger_stop` should be added to the class.

X-Ryl669:

This should fix [#2045](https://github.com/esphome/feature-requests/issues/2045)

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/feature-requests#2045

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2571

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
- platform: time_based
  name: "Kitchen"
  id: cover_1
  open_action:
    - switch.turn_on: relay_kitchen_up
  open_duration: 20s
  close_action:
    - switch.turn_on: relay_kitchen_down
  close_duration: 20s
  stop_action:
    - switch.turn_off: relay_kitchen_up
    - switch.turn_off: relay_kitchen_down
  manual_control: true
```

## Checklist:

### DIFF
--- a/esphome/components/time_based/time_based_cover.cpp
+++ b/esphome/components/time_based/time_based_cover.cpp
@@ -79,6 +79,14 @@ void TimeBasedCover::control(const CoverCall &call) {
     auto pos = *call.get_position();
     if (pos == this->position) {
       // already at target
+      if (this->manual_control_ && (pos == COVER_OPEN || pos == COVER_CLOSED)) {
+        // for covers with manual control switch, we can't rely on the computed position, so if
+        // the command triggered again, we'll assume it's in the opposite direction anyway.
+        auto op = pos == COVER_CLOSED ? COVER_OPERATION_CLOSING : COVER_OPERATION_OPENING;
+        this->position = pos == COVER_CLOSED ? 1.0f : 0.0f;
+        this->target_position_ = pos;
+        this->start_direction_(op);
+      }
       // for covers with built in end stop, we should send the command again
       if (this->has_built_in_endstop_ && (pos == COVER_OPEN || pos == COVER_CLOSED)) {
         auto op = pos == COVER_CLOSED ? COVER_OPERATION_CLOSING : COVER_OPERATION_OPENING;
--- a/esphome/components/time_based/time_based_cover.cpp
+++ b/esphome/components/time_based/time_based_cover.cpp
@@ -83,7 +83,7 @@ void TimeBasedCover::control(const CoverCall &call) {
         // for covers with manual control switch, we can't rely on the computed position, so if
         // the command triggered again, we'll assume it's in the opposite direction anyway.
         auto op = pos == COVER_CLOSED ? COVER_OPERATION_CLOSING : COVER_OPERATION_OPENING;
-        this->position = pos == COVER_CLOSED ? 1.0f : 0.0f;
+        this->position = pos == COVER_CLOSED ? COVER_OPEN : COVER_CLOSED;
         this->target_position_ = pos;
         this->start_direction_(op);
       }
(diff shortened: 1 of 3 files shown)


X-Ryl669:
Ready for merge.

nagyrobi:
In what aspect does this relate to https://github.com/esphome/esphome/pull/2045 ?

X-Ryl669:
Not this one. [This one](https://github.com/esphome/feature-requests/issues/2045).
If I understand correctly, feature requests are in their own repository now.

X-Ryl669:
Fixed the link too here.

esphome/components/time_based/time_based_cover.cpp
```diff
@@ -79,6 +79,14 @@ void TimeBasedCover::control(const CoverCall &call) {
     auto pos = *call.get_position();
     if (pos == this->position) {
       // already at target
+      if (this->manual_control_ && (pos == COVER_OPEN || pos == COVER_CLOSED)) {
+        // for covers with manual control switch, we can't rely on the computed position, so if
+        // the command triggered again, we'll assume it's in the opposite direction anyway.
+        auto op = pos == COVER_CLOSED ? COVER_OPERATION_CLOSING : COVER_OPERATION_OPENING;
+        this->position = pos == COVER_CLOSED ? 1.0f : 0.0f;
```
glmnet: can this line better be:
```suggestion
        this->position = pos == COVER_CLOSED ? COVER_OPEN : COVER_CLOSED;
```

X-Ryl669: Right! Thanks for spotting it.

jesserockz (approved):

PR #4250, opened 2023-01-01
Fix graph limits for negative values

### LINKED ISSUE #3918: graph scaling does not handle negative values
### The problem

The automatic scaling of the graph does not handle negative values correctly. If the minimum value is negative, the graph scale is rounded to the next higher grid step value. The drawn graph trace will then leave the graph area. See attached photo.

graph-scaling-issue

The problem seems to be in the calculation of the minimum y grid value in the following line:
https://github.com/esphome/esphome/blob/9bf7c977751d3b4e5cef123372a07f43f35f597b/esphome/components/graph/graph.cpp#L125
Rounding towards 0 occurs, but rounding towards -INF (negative infinity) is required.
Correction could be something like the following line, but I don't have the capability to set up a dev environment and test the change at the moment:
int yn = int(ymin / y_per_div) - (ymin < 0 && fmodf(ymax, y_per_div) != 0 ? 1 : 0);

Also, it appears that there are some undocumented configuration options of the graph component.

2022.9.4 (and all others versions since the addition of graph)

pip

_No response_

ESP32

### Board

LilyGO T5_V2.3_2.13, configured as board esp32dev

graph

### Example YAML snippet

```yaml
graph:
  - id: aut_graph
    duration: 3h
    width: 122
    height: 85
    border: false
    x_grid: 1h
    y_grid: 10
    traces:
      - sensor: kst
      - sensor: aut

display:
  - platform: waveshare_epaper
    id: epaper
    cs_pin: 5
    dc_pin: 17
    busy_pin: 4
    reset_pin: 16
    model: 2.13in-ttgo-b73
    #rotation: 90
    full_update_every: 100
    update_interval: never
    lambda: |-
      it.graph(0, 0, id(aut_graph));
```

```txt
[23:52:04][V][graph:157]: Updating graph. ymin 0.000000, ymax 10.000000
[23:52:05][V][graph:035]: Updating trace with value: -6.530000
```

### Additional information

_No response_

unhold:
Round down min value instead of truncating using.
Use ceilf() to round up to max value.
The size of an example build decreased by 184 bytes Flash and 8 bytes RAM.

This fixes graph limit calculation with negative values.
The previous truncation (rounding towards 0) led to wrong min value calculation and graph lines leaving the graph area.
You can also see a description and picture in the issue: 

## Types of changes

- [ ] Other

fixes [issue 3918](https://github.com/esphome/issues/issues/3918)

not applicable

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:
  - [n/a] Tests have been added to verify that the new code works (under `tests/` folder).

  - [n/a] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

### DIFF
--- a/esphome/components/graph/graph.cpp
+++ b/esphome/components/graph/graph.cpp
@@ -122,8 +122,8 @@ void Graph::draw(DisplayBuffer *buff, uint16_t x_offset, uint16_t y_offset, Colo
   }
 
   // Adjust limits to nice y_per_div boundaries
-  int yn = int(ymin / y_per_div);
-  int ym = int(ymax / y_per_div) + int(1 * (fmodf(ymax, y_per_div) != 0));
+  int yn = floorf(ymin / y_per_div);
+  int ym = ceilf(ymax / y_per_div);
   ymin = yn * y_per_div;
   ymax = ym * y_per_div;
   yrange = ymax - ymin;


unhold:
I'm seeing some unexpected behavior with certain values, better continue testing.

PR #4261, opened 2023-01-05
Generate main.h to access globals externally

djwlindenaar:
Generate main.h which holds all global variables (things with pointers) so you could use id(...) in an external cpp file (which is supported by the includes directive)

## Types of changes

- [ ] Other

[esphome/issues#4004](https://github.com/esphome/issues/issues/4004)

Will write documentation when this pull request is agreed upon :)
## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
esphome:
  name: "${device_name}"
  includes:
    - test.h
    - test.cpp
  on_boot:
    priority: 200
    then: 
      - lambda: !lambda test();
```
test.cpp
```c
#include "main.h"
#include "test.h"

void test()
{
  ESP_LOGD("test", "%f", id(input_fan_speed).state);
}
```
test.h
```c
#ifndef _test_h_
#define _test_h_

extern void test();

#endif
```
## Checklist:

### DIFF
--- a/esphome/core/__init__.py
+++ b/esphome/core/__init__.py
@@ -781,12 +781,10 @@ class EsphomeCore:
             text = str(statement(exp))
             text = text.rstrip()
             if "*" in text:
-                exports_code.append('extern '+text)
+                exports_code.append("extern " + text)
             elif text.startswith("using"):
                 exports_code.append(text)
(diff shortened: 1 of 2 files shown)


ssieb:
I don't think this should be added.

PR #4262, opened 2023-01-05
Dashboard args for HTTPS

JohnR-LoD:

Adds arguments to the dashboard command to allow passing SSL certificate pem files to serve the dashboard over HTTPS without the need for proxies.

The `--cert` argument takes the full, not relative path to the full chain pem file,
The `--key` argument takes the full, not relative path to the private key pem file, 

These can also be passes as environmental variables `ESPHOME_DASHBOARD_CERT_PATH` and `ESPHOME_DASHBOARD_KEY_PATH` respectively.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2575

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

**Note**: I could not find existing tests for the dashboard to expand upon and starting them from scratch is beyond my ability

### DIFF
--- a/esphome/__main__.py
+++ b/esphome/__main__.py
@@ -792,6 +792,18 @@ def parse_args(argv):
     parser_dashboard.add_argument(
         "--socket", help="Make the dashboard serve under a unix socket", type=str
     )
+    parser_dashboard.add_argument(
+        "--cert",
+        help="The optional path to the full chain pem file to use for HTTPS",
+        type=str,
+        default="",
+    )
+    parser_dashboard.add_argument(
+        "--key",
+        help="The path to the private key pem file to use for HTTPS",
+        type=str,
+        default="",
+    )
 
     parser_vscode = subparsers.add_parser("vscode")
     parser_vscode.add_argument("configuration", help="Your YAML configuration file.")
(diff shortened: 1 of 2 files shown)


JohnR-LoD:
Sorry, this one took a few tries to get right.

May still want to change the script to exclude the file names themselves. It is currently setup to add the default filenames created by certbot to the path passed in the arguments.

jesserockz:
While we appreciate the work you have done, we don't agree that "applications" should be responsible for SSL. It ends up causing issues in the future and reverse proxies are the correct place for SSL termination.

JohnR-LoD:
Is that the position of Nabu Casa at large? If so, are there plans to depreciate that capability in other projects like Home Assistant?

ssieb:
I didn't even know HA has that capability.  I use a reverse proxy for HA anyway.

balloob:
We are not going to remove configuring SSL directly from HA. However it's a big pain, as configuring SSL directly in HA means that HA cannot serve files like TTS, video snapshots or media to media players, as they reject invalid certs. Reverse proxies are the correct place for SSL termination.

PR #4269, opened 2023-01-07
Fix duplicate HttpRequestResponseTrigger definition and use int32_t instead of int

gmbuell:


gmbuell:
This fixes https://github.com/esphome/issues/issues/4010

PR #4292, opened 2023-01-14
Fix off-by-one bedjet fan speed (#3873)

### LINKED ISSUE #3873: BedJet fan speed is off by 5%
### The problem

This comes from discussion in #3807 .

Essentially, there seems to be an "off by one" error with the speed range that the bedjet component reports to HA and the value that HA sends back to the bedjet component when the user selects 100% fan.

Using the [`mushroom/fan` card](https://github.com/piitaya/lovelace-mushroom/blob/main/docs/cards/fan.md), I moved the slider all the way to 100% and then saw this in the ESPHome logs:

```
[10:20:37][W][bedjet:056]: Invalid fan speed index 20, expecting 0-19.
```

Additionally, the speed reported to the user via the HA ui is 5% off from what the bedjet remote shows. The remote is always 5% ahead.

----

Here's the autoconfig payload that's sent to HA:

Topic: `homeassistant/fan/bj_controller-8<...>0/bedjet_fan/config`

Payload:

```json
{
    "pct_cmd_t": "bj_controller/fan/bedjet_fan/speed_level/command",
    "pct_stat_t": "bj_controller/fan/bedjet_fan/speed_level/state",
    "spd_rng_max": 20,
    "name": "BedJet Fan",
    "stat_t": "bj_controller/fan/bedjet_fan/state",
    "cmd_t": "bj_controller/fan/bedjet_fan/command",
    "avty_t": "bj_controller/status",
    "uniq_id": "ESPfanbedjet_fan",
    "dev": {
        "ids": "3<...>0",
        "name": "bj_controller-8<...>0",
        "sw": "esphome v2022.11.4 Dec  4 2022, 10:18:41",
        "mdl": "mhetesp32minikit",
        "mf": "espressif"
    }
}
```

Looking [at the docs](https://www.home-assistant.io/integrations/fan.mqtt/#speed_range_min) for `speed_range_min` attribute: 

```
The minimum of numeric output range (off not included, so speed_range_min - 1 represents 0 %). The number of speeds within the speed_range / 100 will determine the percentage_step.
```

Docs say that the default value is `1` which seems to make sense. If you sent a value of `0`, HA might think that sending a speed value of `-1` was how you set the fan to the lowest possible speed ... which seems wrong :/.

> The fan index sounds like a bug, and your explanation makes sense. Feel free to open a ticket or PR for that and I'll take a look. Should be simple enough. I think the root of the issue is that HA wants to allow setting 0%, but that should actually be "off". I don't think we have a way to tell HA that we want to start at min=5% ("fan speed index 0 == 5%"), so we probably need to update the fan component to have a range of 0-20 and subtract one when sending the speed to BedJet, with a special case for 0==off.

Based on what I'm seeing in the HA docs, I think you are right. A small tweak to the mqtt payload so the value for `payload_off` is set to `0` instead of the default `OFF` should be sufficient.

2022.11.4

pip

_No response_

ESP32-IDF

### Board

_No response_

bedjet

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

jhansche:

Fan speed excludes 0 (reserved for off), and the speed will be `clamp`ed between 1..max. Current max is set to 20, which means we've been passing HA/ESPHome fan speed values of 1-20 through as-is into the BedJet, but BedJet expects to get 0-19.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/issues#3873

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/bedjet/fan/bedjet_fan.cpp
+++ b/esphome/components/bedjet/fan/bedjet_fan.cpp
@@ -37,9 +37,13 @@ void BedJetFan::control(const fan::FanCall &call) {
 
   // ignore speed changes if not on or turning on
   if (this->state && call.get_speed().has_value()) {
-    this->speed = *call.get_speed();
-    this->parent_->set_fan_index(this->speed);
-    did_change = true;
+    auto speed = *call.get_speed();
+    if (speed >= 1) {
+      this->speed = speed;
+      // Fan.speed is 1-20, but Bedjet expects 0-19, so subtract 1
+      this->parent_->set_fan_index(this->speed - 1);
+      did_change = true;
+    }
   }
 
   if (did_change) {
@@ -57,8 +61,9 @@ void BedJetFan::on_status(const BedjetStatusPacket *data) {
     did_change = true;
   }
 
-  if (data->fan_step != this->speed) {
-    this->speed = data->fan_step;
+  // BedjetStatusPacket.fan_step is in range 0-19, but Fan.speed wants 1-20.
+  if (data->fan_step + 1 != this->speed) {
+    this->speed = data->fan_step + 1;
     did_change = true;
   }
 


jhansche:
Tested locally, comparing the Fan component side by side with the Climate component, and with the remote.
* Fan % tracks correctly, all 3 agree: Fan, Climate (preset fan speed), and BedJet remote's display.
* Fan at 5% – 100% each work correctly, and track each other correctly.
* Fan at 0% turns off the unit without resetting or clearing the Fan speed. Turning it back on returns to the previous fan speed.
* Tested various combinations of ON/OFF via Fan switch, ON/OFF by Climate mode, implicit OFF by Fan 0%, implicit ON by Fan speed only

jesserockz (approved):

PR #4307, opened 2023-01-18
Fix ethernet clk_mode for GPIO0_OUT

jorticus:

The ethernet component has a bug preventing `clk_mode: GPIO0_OUT` from working.

This is due to the fact that the IDF enum for clk_mode_t has the same value for `EMAC_CLK_IN_GPIO` and `EMAC_APPL_CLK_OUT_GPIO` (both are equal to 0), and so the logic that sets rmii.clock_mode sets the incorrect value - we would expect mode to be set to EMAC_CLK_OUT, but instead it is set to EMAC_CLK_EXT_IN.

Fix is to separate the two enums and select the correct mode+gpio in the python layer.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
ethernet:
  type: RTL8201
  mdc_pin: GPIO23
  mdio_pin: GPIO18
  clk_mode: GPIO0_OUT
  phy_addr: 0
```

## Checklist:


jorticus:
Apologies, I mixed up my base branch and confused the bot. Things have been fixed now, but let me know if you'd rather me re-submit.

jesserockz (approved):

PR #4309, opened 2023-01-18
simple commit to allow virtual GUI methods

nielsnl68:

Just making all GUI function virtual so my Graphic component can have replace them with Arduino_GFX methods.

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

None changes in yaml file.

## Checklist:

  - [X] No documentation changes needed.


esphome/components/display/display_buffer.h
```diff
@@ -360,14 +371,14 @@ class DisplayBuffer {
...
-  void set_rotation(DisplayRotation rotation);
+  virtual void set_rotation(DisplayRotation rotation);
 
   // Internal method to set display auto clearing.
   void set_auto_clear(bool auto_clear_enabled) { this->auto_clear_enabled_ = auto_clear_enabled; }
 
   virtual int get_height_internal() = 0;
   virtual int get_width_internal() = 0;
-  DisplayRotation get_rotation() const { return this->rotation_; }
+  virtual DisplayRotation get_rotation() const { return this->rotation_; }
```
jesserockz: Why would you need to override this simple getter and setter?

nielsnl68: The graphics controller (Arduino_GFX) has his own way of handling the rotation. It uses just 0, 1 (=90), 2 (=180), 3 (=270) as options. to allow back calculation i need to override the `get_rotation()` as well.

jesserockz (approved):

esphome/components/display/display_buffer.h
```diff
@@ -19,6 +19,8 @@
 #include "esphome/components/qr_code/qr_code.h"
 #endif
 
+#define ENABLE_VIRTUAL_DISPLAYBUFFER
```
jesserockz: What is this for?

nielsnl68: > What is this for?

This allows me to check, when a user uses the Arduino_GFX graphics driver, to see which version of DisplayBuffer is being used and gives an error when the `ENABLE_VIRTUAL_DISPLAYBUFFER` is not set. 

And can be removed just before it go's in production.

jesserockz: I don't like arbitrary defines being added like this, just to be removed later on.
If its going to be a compile error in the other code, you might as well not have this and the compilation will still fail when trying to use the other code as it will not be able to override the non-virtual methods.

nielsnl68: Clear, i removed it.

jesserockz (dismissed):

tests/test1.yaml
jesserockz, line 1: Why delete this?

jesserockz (commented):
@nielsnl68 Do you still want this merged in?

nielsnl68:
I'm closing this PR because i do not think it is worth now there is so much work going on making the esphome display faster etc.

PR #4330, opened 2023-01-21
Add internal_temperature component

### LINKED ISSUE #744: ESP32 internal temperature sensor
Hi!

First of all, thanks for this awesome project, I was genuinely surprised how easy and straightforward using this is! What an awesome DIY experience!

I would love to be able to get the internal CPU temperature sensor value from an ESP32. I want to use it to track it's performance and load under different circumstances.

According to this (and some other) posts, it should be possible to get the value, because there is a function for it defined in an other lib: https://www.esp32.com/viewtopic.php?p=9286&sid=41486dfeb69cd567659dc070219b6dfe#p9286
But I'm not sure if ESPHome is using it and the library is linked. When I tried to use it in a sensor lambda, the linker could not find the implementation for it.

I have an ESP32 with a camera and I want to test different image qualities and framerates to find the optimal setup and also figure out what is the bottleneck (network traffic or CPU thermal throttling, etc.)

**Additional context**

Mat931:

A sensor for the internal CPU temperature (ESP32, RP2040)

Note: The original ESP32 returns many invalid temperature readings. (raw value 128, calculated temperature would be 53.3°C) This component ignores the invalid readings. (see https://github.com/arendst/Tasmota/discussions/13026)

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/744

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2611

## Test Environment

Please help me test this on the ESP32S2, ESP32S3, ESP32C3 and RP2040. I only have ESP32 devices!

|Device|Framework|Compiles|Tested|
|---|---|---|---|
|ESP32|Arduino|yes|yes|
|ESP32|ESP-IDF|yes|yes|
|ESP32S2|Arduino|yes|no|
|ESP32S2|ESP-IDF|yes|no|
|ESP32S3|Arduino|yes|no|
|ESP32S3|ESP-IDF|yes|no|
|ESP32C3|Arduino|yes|yes|
|ESP32C3|ESP-IDF|yes|yes|
|RP2040|Arduino|yes|no|

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [X] RP2040

```yaml
external_components:
  - source: github://pr#4330
    components: [internal_temperature]
    refresh: always

sensor:
  - platform: internal_temperature
    name: "Internal Temperature"
```

## Checklist:


SeByDocKy:
I tested your PR. In the log I got the temperature but nothing is sent to HA...

nagyrobi:
Not sure if "cpu" is the correct term to use in this case. These are microcontrollers, not microprocessors...
Perhaps something like "internal_temperature" would be a better match.

Mat931:
@SeByDocKy Thanks for testing! That's interesting, the same thing happened to me when installing it for the first time on an ESP32 with no other components.

- What microcontroller are you using?
- Can you try rebooting the ESP / RP2040 or reuploading the firmware?
- Did you specify a name for the sensor?

@nagyrobi Yeah, internal_temperature would fit better. I'll rename the component tomorrow.

SeByDocKy:
> @SeByDocKy Thanks for testing! That's interesting, the same thing happened to me when installing it for the first time on an ESP32 with no other components.
> 
> * What microcontroller are you using?
> * Can you try rebooting the ESP / RP2040 or reuploading the firmware?
> * Did you specify a name for the sensor?
> 
> @nagyrobi Yeah, internal_temperature would fit better. I'll rename the component tomorrow.

Strange now the entity is available.... Maybe after a reboot


i) ESP32 WROOM32
```
esp32:
  board: esp32dev
  framework:
    type: arduino
    version: 2.0.6   #2.0.5
    platform_version: 5.3.0 #5.2.0
    # type: esp-idf
    
external_components:
  - source: github://pr#4330
    components: [cpu_temperature]
    refresh: 0s
```



iii) 

```
- platform: cpu_temperature
    name: ${prefix}_cpu_temperature
    update_interval: 10s
```


Often in the logs, got ignoring temperature as depicted below, but I guess it's normal.

image


Btw, I don't why during 1h, the new entity didn't update (I tried to delete and add again the ESPhome node without success)

Mat931:
@SeByDocKy I think the problem was that the sensor didn't have a value yet when connecting to HA. Now the temperature is set to NAN / Unknown if the first measurement after booting the ESP is invalid. Hope that fixes the issue :)

SeByDocKy:
Work fine for me :)

eavyon:
I can confirm it works on board lolin_c3_mini > esp32c3
Doesnt work on board nodemcu-32s(az delivery wroom32) > esp32 gives only invalid values
Both arduino framework

Mat931:
@eavyon
Thanks for testing!
On the ESP32 there can be many invalid values in a row, that doesn't mean it isn't working at all. Sometimes rebooting the ESP helps. If there's still no valid measurement after multiple hours it's probably broken.
Can you tell me the chip version from esptool.py or from the logs when uploading via serial?

Mine (confirmed working) are:

- ESP32-D0WDQ6 (revision v0.0)
- ESP32-D0WDQ6 (revision v1.0)
- ESP32-D0WD (revision v1.0)
- ESP32-U4WDH-S (revision v3.0) (Shelly Plus 1PM)

eavyon:
@Mat931 I checked again after restarting and it is now reporting last few days. 
Revision ESP32-D0WDQ6 (revision 1)
image

Mat931:
@eavyon Nice, thanks! Would be great if you could also test it with esp-idf on your ESP32C3 :)

eavyon:
@Mat931 I flashed with esp-idf and can confirm it also works normally last few hours on ESP32C3.

esphome/components/internal_temperature/internal_temperature.cpp
```diff
@@ -0,0 +1,65 @@
...
+#endif
+uint8_t temprature_sens_read();
+#ifdef __cplusplus
+}
+#endif
+#else  // CONFIG_IDF_TARGET_ESP32
+#include "driver/temp_sensor.h"
+#endif  // CONFIG_IDF_TARGET_ESP32
+#endif  // USE_ESP32
+
```
oxan: No need for the `#ifdef __cplusplus` as this is a `.cpp` file and will thus always compile as C++, and we prefer our own `USE_ESP32_VARIANT_*` macros over those from the IDF (also change this below please). I also added a comment to note why we define the prototype here manually.
```suggestion
#ifdef USE_ESP32
#ifdef USE_ESP32_VARIANT_ESP32
// there is no official API available on the original ESP32
extern "C" {
uint8_t temprature_sens_read();
}
#else  // USE_ESP32_VARIANT_ESP32
#include "driver/temp_sensor.h"
#endif  // USE_ESP32_VARIANT_ESP32
#endif  // USE_ESP32
```

esphome/components/internal_temperature/internal_temperature.cpp
```diff
@@ -0,0 +1,65 @@
...
+  temp_sensor_stop();
+  success = (result == ESP_OK);
+#endif  // CONFIG_IDF_TARGET_ESP32
+#endif  // USE_ESP32
+#if USE_RP2040
+  temperature = analogReadTemp();
+  success = (temperature != 0.0f);
+#endif  // USE_RP2040
+  if (success && std::isfinite(temperature)) {
+    ESP_LOGD(TAG, "Got temperature: %.1f°C", temperature);
```
oxan:
```suggestion
```
This log statement is superfluous as `this->publish_state` will already log the published value.

esphome/components/internal_temperature/internal_temperature.h
```diff
@@ -0,0 +1,19 @@
...
+namespace esphome {
+namespace internal_temperature {
+
+class InternalTemperatureSensor : public sensor::Sensor, public PollingComponent {
+ public:
+  void dump_config() override;
+
+  void update() override;
+
+  std::string unique_id() override;
```
oxan: This override isn't necessary nor desired for new components.

esphome/components/internal_temperature/internal_temperature.cpp
```diff
@@ -0,0 +1,65 @@
+#include "internal_temperature.h"
+#include "esphome/core/log.h"
+#include "esphome/core/hal.h"
```
oxan: Doesn't seem to be used:
```suggestion
```

oxan (changes requested):

oxan (approved):

PR #4338, opened 2023-01-23
Ledc fix

### LINKED ISSUE #4050: LEDC bit_depth calculation is not working properly on Arduino framework in ESPHome 2022.12.3 and dev
### The problem

I am using the esp32-c3-devkitm-1 board (or a board that's close enough to that one).
Important note: I used the ledc component from the dev branch

The LEDC component is working fine under the IDF framework.  With my configuration,
 ```
 - platform: ledc
    id: fan_speed
    pin: 
      number: GPIO04
    power_supply: fan_power
    channel: 0
    frequency: 25000 Hz
    inverted: true
    min_power: 0%
    max_power: 100%
```
I get a bit_depth resolution of 11 bits.

When I run the same configuration but using the arduino framework, it stops working.
After some investigations, it appears that:

- the computed bit depth for my frequency (25kHz) is still 11 bits but it's not accepted by the call to ledcSetup (returns 0 and not the actual frequency when successful)
- with a bit-depth of 10, it's working nicely.
- Since Arduino 2.0.2, the actual clock being used has changed (see https://github.com/espressif/arduino-esp32/blob/2.0.2/cores/esp32/esp32-hal-ledc.c), as it uses the XTAL clock if it's available, the AUTO CLK if not.  The ESP32C3-mini I use has the crystal for sure.
- Forcing the IDF esphome code to use LEDC_USE_XTAL_CLK instead of LEDC_AUTO_CLK gives the very same issue (only 10-bits can be used instead the computed 11-bits).

Using Arduino 2.0.1 (the version before the default clock change for LEDC in Arduino), ESPHome ledc component is actually working with 11-bits bit_depth.  The change in ledc implementation in arduino 2.0.2+ is then breaking the current and dev code of the ledc component in ESPHome.

2022.12.3, dev

pip

_No response_

ESP32

### Board

esp32-c3-devkitm-1

ledc

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by xvil**

This PR fixes the bit-depth calculation of LEDC when using the arduino framework.
It also add some fallback mechanism when the initially computed bit-depth is not compatible with the user-defined frequency of the PWM (both for IDF and Arduino)

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [link to issue](https://github.com/esphome/issues/issues/4050)

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
esphome:
  name: esp-ledc-tst

esp32:
  board: esp32-c3-devkitm-1
  framework:
    type: arduino
      
# Enable logging
logger:
  baud_rate: 115200
  level: verbose
  logs:
    ledc: verbose

# PWM output
output:
  - platform: ledc
    id: fan_pwm
    pin: 
      number: GPIO04
    frequency: 25000 Hz
    inverted: true
    min_power: 0%
    max_power: 100%

# Used to manually set the fan speed
fan:
  - platform: speed
    output: fan_pwm
    name: "Fan Speed"
    id: fan_speed
```

## Checklist:


jesserockz (changes requested):
I know the existing code doesn't clear the warnings, but if we are making changes, might as well clear the error if it successfully sets the frequency

xvil:
I have implemented the error status reset.
I have added the fallback of the bit_depth calculation for set_frequency when using the ESP_IDF framework.
The common code has refactored into a dedicated function to avoid duplication as much as possible.

The changes have been tested locally with a sample yaml, built and deployed on my ESP32C3 board.

jesserockz (approved):

PR #4340, opened 2023-01-23
Added local min and local max filters

xvil:

As the values out of a sensor evolve over time, it could be useful, in some automation using ESPHome, to know what is the last local minimum or the last local minimum.

The last local minimum is defined as the most recent minimum discovered when values started to increase again.
It can only be discovered when a new value becomes higher or equal to the last one and the trend was decreasing before.

Following the same idea, the last local maximum is defined as the most recent maximum discovered when values started to decrease again.
It can only be discovered when a new value becomes lower or equal to the last one and the trend was increasing before.

This PR adds those two new filters:
* Local minimum
* Local maximum

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
  - platform: copy
    source_id: in_temp
    id: local_min_in_sensor
    name: "Local minimum for Input Temperature"
    filters:
    - local_min

  - platform: copy
    source_id: in_temp
    id: local_max_in_sensor
    name: "Local maximum for Input Temperature"
    filters:
    - local_max
```

## Checklist:


jesserockz (approved):

jesserockz:
@xvil Thanks for this.
Are you planning to add these to the docs so they can be merged in?

PR #4343, opened 2023-01-24
Add next_url to improv serial component config

jesserockz:

This allows project creators to specify their own Next URL for client side tools like ESP Web Tools to link the user through to. This could be something like an online manual, which can also have query params to alter the manual contents such as an IP address which will allow for dynamic changes in the manual specific to the device.

Valid substitutions are `esphome_version`, `ip_address`, `device_name`. See below for an example of how to use them in YAML.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2620

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
improv_serial:
  next_url: https://example.com/new-device?ip={{ip_address}}&name={{device_name}}&version={{esphome_version}}

```

## Checklist:

### DIFF
--- a/esphome/components/improv_base/__init__.py
+++ b/esphome/components/improv_base/__init__.py
@@ -1,0 +1,42 @@
+import re
+
+import esphome.config_validation as cv
+import esphome.codegen as cg
+
+from esphome.const import __version__
+
+CODEOWNERS = ["@esphome/core"]
+
+CONF_NEXT_URL = "next_url"
+
+VALID_SUBSTITUTIONS = ["esphome_version", "ip_address", "device_name"]
+
+
+def validate_next_url(value):
+    value = cv.url(value)
+    test = r"{{(?!" + r"\b|".join(VALID_SUBSTITUTIONS) + r"\b)(\w+)}}"
+    result = re.search(test, value)
+    if result:
+        raise cv.Invalid(
+            f"Invalid substitution(s) ({', '.join(result.groups())}) in next_url. Valid substitutions are: {', '.join(VALID_SUBSTITUTIONS)}"
+        )
+    return value
+
+
+IMPROV_SCHEMA = cv.Schema(
+    {
+        cv.Optional(CONF_NEXT_URL): validate_next_url,
+    }
+)
+
+
+def _process_next_url(url: str):
+    if "{{esphome_version}}" in url:
+        url = url.replace("{{esphome_version}}", __version__)
+    return url
+
+
+async def setup_improv_core(var, config):
+    if CONF_NEXT_URL in config:
+        cg.add(var.set_next_url(_process_next_url(config[CONF_NEXT_URL])))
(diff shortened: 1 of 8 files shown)


esphome/components/improv_base/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+def _process_next_url(url: str):
+    if "{{esphome_version}}" in url:
+        url = url.replace("{{esphome_version}}", __version__)
+    return url
+
+
+async def setup_improv_core(var, config):
+    if CONF_NEXT_URL in config:
+        cg.add(var.set_next_url(_process_next_url(config[CONF_NEXT_URL])))
+    cg.add_library("esphome/Improv", "1.2.3")
```
balloob, line 42: Is that a real lib nr?

esphome/components/improv_base/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+IMPROV_SCHEMA = cv.Schema(
+    {
+        cv.Optional(CONF_NEXT_URL): validate_next_url,
+    }
+)
+
+
+def _process_next_url(url: str):
```
balloob, line 33: Shouldn't you be able to statically replace device name too ?

balloob: Ah no, mac address suffix.

esphome/components/improv_serial/improv_serial_component.cpp
```diff
@@ -100,6 +100,9 @@ std::vector<uint8_t> ImprovSerialComponent::build_rpc_settings_response_(improv:
   std::string webserver_url = "http://" + ip.str() + ":" + to_string(USE_WEBSERVER_PORT);
   urls.push_back(webserver_url);
 #endif
+  if (!this->next_url_.empty()) {
```
balloob: can we put it before the web server.

balloob (approved):

esphome/components/improv_base/__init__.py
jesserockz: Yeah, it was moved from https://github.com/esphome/esphome/pull/4343/files#diff-acc0bd628c6c90fba1c8fc07f4e06065aad3d8afc559979f695414c10e0b3a69L40

esphome/components/improv_serial/improv_serial_component.cpp
jesserockz: Sure

jesserockz: [`18abf9a` (#4343)](https://github.com/esphome/esphome/pull/4343/commits/18abf9a784752c1861e4d8ea804cf633eb18570b)

PR #4357, opened 2023-01-27
Fix pressure compensation in SCD4X

CarlosGS:

When setting `ambient_pressure_compensation` with a static value (i.e. not using `ambient_pressure_compensation_source`), the `set_` function was being executed while the sensor was not initialized. So the value was being overridden with 0hPa.

I've separated the `if(initialized_)` check and placed it earlier in the function so `ambient_pressure_` is set correctly when called before setup.

Also, in the dynamic pressure correction I've removed a division that discarded +-10hPa changes. Since pressure is already treated as an integer in hPa (i.e. 990hPa or 1012hPa, without decimals) the I2C updates aren't really that frequent to need rounding down to +-10hPa steps. It should also minimize the small noise introduced when pressure oscillates near round values i.e. 999->990hPa, 1000->1000hPa.

The breaking change is switching from Bar to mBar units (there was a redundant /1000 *1000 operation) to unify with other sensors such as the SCD30 and compensation sensors as BME280/680 etc, that already work in mBar/hPa.
I've added examples in the documentation to offer two compensation options (via sensor and via HASS service).

## Types of changes

- [ ] Other

**Related documentation PR:** esphome/esphome-docs#2626

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/scd4x/scd4x.cpp
+++ b/esphome/components/scd4x/scd4x.cpp
@@ -59,7 +59,7 @@ void SCD4XComponent::setup() {
 
       // If pressure compensation available use it
       // else use altitude
-      if (ambient_pressure_compensation_) {
+      /*if (ambient_pressure_compensation_) {
         if (!this->update_ambient_pressure_compensation_(ambient_pressure_)) {
           ESP_LOGE(TAG, "Error setting ambient pressure compensation.");
           this->error_code_ = MEASUREMENT_INIT_FAILED;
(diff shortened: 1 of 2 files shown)


martgras (approved):
Looks good to me.  Switching to mbar also makes sense.

CarlosGS:
Thanks @jesserockz! Please merge the documentation PR too when possible https://github.com/esphome/esphome-docs/pull/2626 :)

PR #4000, opened 2022-11-07
Initial stab at importing idf components

jesserockz:
# What does this implement/fix?

This allows idf-based esp32 builds to reference idf components.

Todo: Allow adding arbitrary components added via config entries, maybe:
```yaml
esp32:
  framework:
    type: esp-idf
    components:
      - name: esp32-camera:
        source: https://github.com/espressif/esp32-camera
        refresh: 1day
```

Thoughts/suggestions?

Here is an example change for `esp32_camera`:
```diff
diff --git a/esphome/components/esp32_camera/__init__.py b/esphome/components/esp32_camera/__init__.py
index 753b6ed9d..786884bc6 100644
--- a/esphome/components/esp32_camera/__init__.py
+++ b/esphome/components/esp32_camera/__init__.py
@@ -16,7 +16,7 @@ from esphome.const import (
     CONF_TRIGGER_ID,
 )
 from esphome.core import CORE
-from esphome.components.esp32 import add_idf_sdkconfig_option
+from esphome.components.esp32 import add_idf_sdkconfig_option, add_idf_component
 from esphome.cpp_helpers import setup_entity
 
 DEPENDENCIES = ["esp32"]
@@ -263,8 +263,11 @@ async def to_code(config):
     cg.add_define("USE_ESP32_CAMERA")
 
     if CORE.using_esp_idf:
-        cg.add_library("espressif/esp32-camera", "1.0.0")
+        add_idf_component(
+            "esp32-camera", "https://github.com/espressif/esp32-camera.git", "v2.0.2"
+        )
         add_idf_sdkconfig_option("CONFIG_RTCIO_SUPPORT_RTC_GPIO_DESC", True)
+        add_idf_sdkconfig_option("CONFIG_SCCB_CLK_FREQ", 100000)
 
     for conf in config.get(CONF_ON_STREAM_START, []):
         trigger = cg.new_Pvariable(conf[CONF_TRIGGER_ID], var)
```

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
[...]

PR #4364, opened 2023-01-30
Initial attempt at supporting ESP-IDF 5.0.0

kbx81:

This is just a batch of changes I found necessary to get a successful build with IDF 5.0.

It's...a mess...but we have to start somewhere. My hope is that others will find at least some of this work helpful as we move towards full, proper support of the updated framework.

In general, this is very much a WIP and I don't expect we'll merge this specific PR...but, who knows. 😄

Version 6.0 of the platform also seems to complain about `printf` (and siblings) all over the place. I'm guessing this is some compiler setting but I don't know how and didn't take the time to look up how to fix that...so for now I have a commit with that nonsense. You'll see it complain in any other component(s) where those functions are used.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

Note that, for now, this **requires** importing the mDNS component via @jesserockz's #4000 PR; but, with this, I was able to get a successful build. That said, mDNS **will not actually work** because we have more work to do to get the hostname set correctly. You'll probably notice warnings about this if/when you build.

That said, _if you run this on your ESP, you can still do updates OTA, but you'll need to use the device's IP as it won't respond to any hostname._

```yaml
# Example config.yaml
esphome:
  platformio_options:
    build_flags:
      - -DCONFIG_FREERTOS_ENABLE_BACKWARD_COMPATIBILITY

esp32:
  board: featheresp32
  framework:
    type: esp-idf
    version: "5.0.0"
    platform_version: "6.0.0"
    components:
      - name: esp-protocols
        source: github://espressif/esp-protocols

mdns:
  disabled: true
```

**NOTE:** I also ran into an issue building as described in [this thread](https://github.com/espressif/esp-idf/issues/8489); read the comments for the resolution (basically just run `install.sh` followed by `. export.sh`).

## Checklist:
  - [x] The code change is tested and works locally...mostly, kind of.

### DIFF
--- a/esphome/components/mdns/__init__.py
+++ b/esphome/components/mdns/__init__.py
@@ -4,10 +4,13 @@ from esphome.const import (
     CONF_PROTOCOL,
     CONF_SERVICES,
     CONF_SERVICE,
+    KEY_CORE,
+    KEY_FRAMEWORK_VERSION,
 )
 import esphome.codegen as cg
 import esphome.config_validation as cv
 from esphome.core import CORE, coroutine_with_priority
+from esphome.components.esp32 import add_idf_component
 
 CODEOWNERS = ["@esphome/core"]
 DEPENDENCIES = ["network"]
@@ -78,6 +81,16 @@ async def to_code(config):
             cg.add_library("ESP8266mDNS", None)
         elif CORE.is_rp2040:
             cg.add_library("LEAmDNS", None)
+
+    if CORE.using_esp_idf and CORE.data[KEY_CORE][KEY_FRAMEWORK_VERSION] >= cv.Version(
+        5, 0, 0
+    ):
+        add_idf_component(
+            "mdns",
+            "https://github.com/espressif/esp-protocols.git",
+            "mdns-v1.0.9",
+            "components/mdns",
+        )
 
     if config[CONF_DISABLED]:
         return
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,6 +24,7 @@ RUN \
         python3-setuptools=52.0.0-4 \
         python3-pil=8.1.2+dfsg-0.3+deb11u1 \
         python3-cryptography=3.3.2-1 \
+        python3-venv=3.9.2-3 \
         iputils-ping=3:20210202-1 \
         git=1:2.30.2-1 \
         curl=7.74.0-1.3+deb11u7 \
--- a/esphome/components/esp32/core.cpp
+++ b/esphome/components/esp32/core.cpp
@@ -7,6 +7,7 @@
 #include <freertos/task.h>
 #include <esp_idf_version.h>
 #include <esp_task_wdt.h>
+#include <esp_timer.h>
 #include <soc/rtc.h>
 
 #if ESP_IDF_VERSION_MAJOR >= 4
--- a/esphome/components/ota/ota_backend_esp_idf.cpp
+++ b/esphome/components/ota/ota_backend_esp_idf.cpp
@@ -8,6 +8,10 @@
 #include <esp_ota_ops.h>
 #include "esphome/components/md5/md5.h"
 
+#if ESP_IDF_VERSION_MAJOR >= 5
+#include <spi_flash_mmap.h>
+#endif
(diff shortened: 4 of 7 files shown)


oxan:
Some discussion on the discord: https://discord.com/channels/429907082951524364/1076282397903048725

stintel:
Tested the changes I just pushed on ESP32-C3-DevKitM-1 and WEMOS LOLIN D32 PRO using the following version combinations:

* esp-idf 4.4.4 + platform-espressif 5.3.0
* esp-idf 5.0.1 + platform-espressif 6.0.0

The example config can be reduced to this:
```yaml
---
esphome:
  name: idf501

esp32:
  board: featheresp32
  framework:
    type: esp-idf
    version: "5.0.1"
    platform_version: "6.0.0"

```

FYI: platform-espressif32 >= 6.0.1 doesn't work due to https://github.com/platformio/platform-espressif32/commit/8507dd8e7fc6d3fb56ec5372fc24a3044226a058. It now requires python-venv. I could fix that, but might be getting out of scope for this PR. Please let me know what people prefer.

esphome/components/mdns/__init__.py
```diff
@@ -79,6 +84,14 @@ async def to_code(config):
         elif CORE.is_rp2040:
             cg.add_library("LEAmDNS", None)
 
+    if CORE.using_esp_idf and (framework_ver.major, framework_ver.minor) >= (5, 0):
```
oxan: Minor nit; but you can compare `cv.Version` instances directly:
```suggestion
    if CORE.using_esp_idf and framework_ver >= cv.Version(5, 0, 0):
```
(or maybe even drop the `framework_ver` variable).

esphome/core/config.py
```diff
@@ -383,6 +386,9 @@ async def to_code(config):
     if CORE.using_arduino:
         CORE.add_job(add_arduino_global_workaround)
 
+    if CORE.using_esp_idf and (framework_ver.major, framework_ver.minor) >= (5, 0):
```
oxan: same here.

oxan (approved):
LGTM.

stintel:
Retested the changes on ESP32-C3-DevKitM-1 and WEMOS LOLIN D32 PRO using the following version combinations:

    esp-idf 4.4.4 + platform-espressif 5.3.0
    esp-idf 5.0.1 + platform-espressif 6.0.0

This time also with IPv6 enabled and disabled, due to the changes I made to resolve conflicts after d42f35de5d54. @jesserockz can you please verify I've correctly resolved these conflicts?

I also noticed that switching esp-idf versions requires wiping esp-idf's python dependencies installed by platformio. This is exactly what they fixed in https://github.com/platformio/platform-espressif32/commit/8507dd8e7fc6d3fb56ec5372fc24a3044226a058, so I've added a commit to install python3-venv.

Afterwards retested with platform-espressif32 6.0.1 and 6.1.0.

New example config to avoid those python dependency issues:
```yaml
---
esphome:
  name: idf501

esp32:
  board: featheresp32
  framework:
    type: esp-idf
    version: "5.0.1"
    platform_version: "6.1.0"
```

jesserockz (approved):
Changes look fine. Will give them a quick test locally during the week if not today.

esphome/core/config.py
```diff
@@ -383,6 +384,11 @@ async def to_code(config):
     if CORE.using_arduino:
         CORE.add_job(add_arduino_global_workaround)
 
+    if CORE.using_esp_idf and CORE.data[KEY_CORE][KEY_FRAMEWORK_VERSION] >= cv.Version(
+        5, 0, 0
+    ):
+        cg.add_build_flag("-Wno-format")
```
jesserockz: Is this still required after https://github.com/esphome/esphome/pull/4671 ?

stintel: No. Dropped the commit.

stintel:
Anything still blocking this PR from being merged?

jesserockz (approved):

PR #4365, opened 2023-01-30
Added password checking on fingerprint scanning

nitobosch:

Adds more security through the password. Until now, the password is only checked on setup at system boot, with that change, the password would be checked also at fingerprint scanning.

## Types of changes

- [X] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/fingerprint_grow/fingerprint_grow.cpp
+++ b/esphome/components/fingerprint_grow/fingerprint_grow.cpp
@@ -90,7 +90,7 @@ void FingerprintGrowComponent::finish_enrollment(uint8_t result) {
 void FingerprintGrowComponent::scan_and_match_() {
   if (this->check_password_()) {
     if (this->sensing_pin_ != nullptr) {
-	  ESP_LOGD(TAG, "Scan and match");
+      ESP_LOGD(TAG, "Scan and match");
     } else {
   	  ESP_LOGV(TAG, "Scan and match");
     }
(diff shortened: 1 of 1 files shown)


OnFreund:
What is the potential attack that this is protecting against?

nitobosch:
> What is the potential attack that this is protecting against?

Password is only checked on setup, so if the system has already booted with an authorized sensor and then it's replaced by an un-authorized sensor without rebooting the system. The matched fingers scanned through the "new" sensor will be sended as OK, although the "new" sensor doesn´t have the correct password.

OnFreund:
The password check isn't verifying the sensor in any way...

nitobosch:
> The password check isn't verifying the sensor in any way...

When you set a _**new_password**_ (https://esphome.io/components/fingerprint_grow.html#setting-a-new-password) it's set on the fingerprint reader, so then you need to indicate de correct _**password**_ in your .yaml in order to comunicate with it. So, if you include the password in your .yaml it'll be checked at setup that the reader has that password set.

OnFreund:
Yes, it's verifying the client, not the sensor. The client is sending the password to the sensor over plain text.

nitobosch:
Ok, but the client shouldn't work with a sensor that doesn't have that password set.

nitobosch:
The fingers are stored on the reader, so if a reader that doesn´t have the same password that the client is connected to it, it would send matched fingers for unexpected fingerprints. Because the expected fingerprints are stored in the reader that has the password set. It's very insecure, because the sensor is more likely to be accessible by anyone and it could be changed relatively easily.

OnFreund:
But now you've given an attacker an easy way to obtain the password (without your changes they have to trigger a restart of the ESP somehow, now they just need to trigger a scan), and use it to download all of the data off of your sensor.
You are right that the GROW protocol is inherently insecure, but what you're suggesting doesn't address it but rather adds another attack vector. The right way to protect against the attack you're describing is through limits on attempts, and by not relying solely on the fingerprint.

loongyh:
Couple of clarifications:
1. The password is set on the reader - it's required by the reader on every startup to unlock access. Once the correct password is verified by the reader, it'll stay unlocked till the reader resets by power cycle or command.
2. As @OnFreund stated, it's used to secure the data in the reader such that an attacker can't just plug it into his device and steal your fingerprint data. Of course this only works if the attacker unplugs it - the attacker can still access the unlocked reader by simply plugging in the RX/TX without disconnecting power.
3. That being said, password is a 32-bit unsigned integer (32 bits of entropy). It has been bruteforced in 5mins, [see this thread here](https://community.home-assistant.io/t/cannot-communicate-with-grow-r503-fingerprint-sensor-after-setting-password/432628/5).
4. TLDR - it's a $10 reader. Don't use this to safeguard your gold bars.

jesserockz:
Honestly I don't see how this increases any security at all. Someone can snoop the uart line and get the password in "plain-text" program their own fingerprint sensor with the same password, then replace your sensor with theirs and ESPHome will accept any fingerprint they programmed in the past...

ssieb:
> Honestly I don't see how this increases any security at all.

I agree and said the same thing in discord when this was discussed.

loongyh:
Actually, a better implemention may involve a rolling password -

1. on boot, esphome logs in with the current password and then immediately sets a new random password and persists it as the current password (we have to figure a way to persist this on runtime).
2. on next boot, it logs in with the new password and again immediately sets a new random password...

Probably a moot effort since the password is brute-forceable in a short time. Also if there's a way for the attacker to trigger a reboot, the new password can be sniffed out too.

But if all the attacker wants is to unlock the door, the fingerprint reader itself is not even required. Just send the relevant GrowResponse OK payload when ESPHome sends the Search GrowCommand to trigger a fingerprint match. In practicality, nobody's gonna bother with doing this, probably easier to just smash the door as is the case of [xkcd 538](https://xkcd.com/538/).

PR #4366, opened 2023-01-30
Add support for ATM90E26

danieltwagner:

This branch adds support for the ATM90E26 energy meter. It's similar in functionality to the ATM90E32 which is already supported, but is single-phase only (1P2W).

Sadly, much of the way the functionality is accessed is substantially different from the -32 version of the chip, down to the way registers are accessed. I therefore re-implemented this, rather than extending the ATM90E32 component.

This has been working fine for a bit over a week for me but I'm looking for additional feedback before writing up any documentation.

## Types of changes

- [ ] Other

https://github.com/esphome/esphome-docs/pull/3002

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
substitutions:
  disp_name: House
  update_time: 10s
  current_cal: '30027'  # default: 31251
  voltage_cal: '17526'  # default: 26400 - Application node says this should be < 32768, maybe for some internal computation? -- this one is outputting 1/4 the line voltage, we must therefore multiply computed outputs.
  metering_cal: '7481'  # default: 7481 - Calibrate this to match your meter based on the CF1 (CFx) pulse
  plconst_cal: '889913' # default: 1429876, compute as 838860800 * (gain_pga * <sampled voltage at 1Amp current> * <sampled voltage at reference voltage> / (<pulse constant (e.g. 3200 pulses/kWh)> * <reference voltage, e.g. 230V>))
  # 838860800 * (1 * 6 * 271 / (1666*4 * 230))

# Energy Monitor
spi:
  clk_pin: 18
  miso_pin: 19
  mosi_pin: 23

# frequently check for updates while editing this file or compiling esphome
external_components:
  - source: github://danieltwagner/esphome@atm90e26
    components: [ atm90e26 ]
    refresh: 30s

sensor:
  - platform: atm90e26
    cs_pin: 5
    voltage:
      name: ${disp_name} Volts
      accuracy_decimals: 1
      filters:
        - multiply: 4
    current:
      name: ${disp_name} Amps
      id: "ct1Amps"
# The max value for current that the meter can output is 65.535. If you expect to measure current over 65A, 
# divide the gain_ct by 2 (120A CT) or 4 (200A CT) and multiply the current and power values by 2 or 4 by uncommenting the filter below
#      filters:
#        - multiply: 2
    power:
      name: ${disp_name} Watts
      accuracy_decimals: 1
      id: "ct1Watts"
      filters:
        - multiply: 4
    reactive_power:
      name: ${disp_name} Reactive Power
    power_factor:
      name: ${disp_name} Power Factor
      accuracy_decimals: 2
      id: "ct1PowerFactor"
    forward_active_energy:
      name: ${disp_name} Forward Active Energy
    reverse_active_energy:
      name: ${disp_name} Reverse Active Energy
    frequency:
      name: ${disp_name} Freq
    line_frequency: 50Hz
    pl_const: ${plconst_cal}
    meter_constant: '1666.66'
    gain_metering: ${metering_cal}
    gain_voltage: ${voltage_cal}
    gain_ct: ${current_cal}
    gain_pga: 1X
    update_interval: ${update_time}

  - platform: total_daily_energy
    name: ${disp_name} Total kWh
    power_id: ct1Watts
    filters:
      - multiply: 0.001
    unit_of_measurement: kWh

time:
  - platform: sntp
    id: sntp_time

```

## Checklist:

### DIFF
--- a/esphome/components/atm90e26/atm90e26.cpp
+++ b/esphome/components/atm90e26/atm90e26.cpp
@@ -45,8 +45,8 @@ void ATM90E26Component::setup() {
   this->spi_setup();
 
   uint16_t mmode = 0x422;  // default values for everything but L/N line current gains
-  mmode |= gain_pga_ << 13;
-  mmode |= n_line_gain_ << 11;
+  mmode |= (gain_pga_ & 0x7) << 13;
+  mmode |= (n_line_gain_ & 0x3) << 11;
 
   this->write16_(ATM90E26_REGISTER_SOFTRESET, 0x789A);  // Perform soft reset
   this->write16_(ATM90E26_REGISTER_FUNCEN,
--- a/esphome/components/atm90e26/atm90e26.cpp
+++ b/esphome/components/atm90e26/atm90e26.cpp
@@ -178,7 +178,7 @@ void ATM90E26Component::write16_(uint8_t a_register, uint16_t val) {
 
 float ATM90E26Component::get_line_current_() {
   uint16_t current = this->read16_(ATM90E26_REGISTER_IRMS);
-  return (float) current / 1000;
+  return current / 1000.0f;
 }
 
 float ATM90E26Component::get_line_voltage_() {
--- a/esphome/components/atm90e26/atm90e26.cpp
+++ b/esphome/components/atm90e26/atm90e26.cpp
@@ -44,12 +44,13 @@ void ATM90E26Component::setup() {
   ESP_LOGCONFIG(TAG, "Setting up ATM90E26 Component...");
   this->spi_setup();
 
-  uint16_t mmode = 0x422; // default values for everything but L/N line current gains
+  uint16_t mmode = 0x422;  // default values for everything but L/N line current gains
   mmode |= l_line_gain_ << 13;
   mmode |= n_line_gain_ << 11;
 
-  this->write16_(ATM90E26_REGISTER_SOFTRESET, 0x789A);    // Perform soft reset
-  this->write16_(ATM90E26_REGISTER_FUNCEN, 0x0030);       // Voltage sag irq=1, report on warnout pin=1, energy dir change irq=0
+  this->write16_(ATM90E26_REGISTER_SOFTRESET, 0x789A);  // Perform soft reset
+  this->write16_(ATM90E26_REGISTER_FUNCEN,
+                 0x0030);  // Voltage sag irq=1, report on warnout pin=1, energy dir change irq=0
   uint16_t read = this->read16_(ATM90E26_REGISTER_LASTDATA);
   if (read != 0x0030) {
     ESP_LOGW(TAG, "Could not initialize ATM90E26 IC, check SPI settings: %d", read);
@@ -60,13 +61,13 @@ void ATM90E26Component::setup() {
   this->write16_(ATM90E26_REGISTER_SAGTH, 0x17DD);  // Voltage sag threshhold 0x1F2F
 
   // Set metering calibration values
-  this->write16_(ATM90E26_REGISTER_CALSTART, 0x5678);       // CAL Metering calibration startup command
+  this->write16_(ATM90E26_REGISTER_CALSTART, 0x5678);  // CAL Metering calibration startup command
 
   // Configure
-  this->write16_(ATM90E26_REGISTER_MMODE, mmode);           // Metering Mode Configuration (see above)
+  this->write16_(ATM90E26_REGISTER_MMODE, mmode);  // Metering Mode Configuration (see above)
 
-  this->write16_(ATM90E26_REGISTER_PLCONSTH, (pl_const_ >> 16));  // PL Constant MSB
-  this->write16_(ATM90E26_REGISTER_PLCONSTL, pl_const_ & 0xFFFF); // PL Constant LSB
+  this->write16_(ATM90E26_REGISTER_PLCONSTH, (pl_const_ >> 16));   // PL Constant MSB
+  this->write16_(ATM90E26_REGISTER_PLCONSTL, pl_const_ & 0xFFFF);  // PL Constant LSB
 
   // TODO: Calibrate this to be 1 pulse per Wh
   this->write16_(ATM90E26_REGISTER_LGAIN, metering_gain_);  // L Line Calibration Gain (active power metering)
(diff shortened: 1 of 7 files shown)


danieltwagner:
Updated with CODEOWNERS but am still looking for feedback before writing docs. If all is well I'm happy to create a PR in `esphome-docs` to get this merged :)

esphome/components/atm90e26/atm90e26.cpp
```diff
@@ -0,0 +1,234 @@
...
+  this->write_byte(a_register & 0x7F);
+  delayMicroseconds(4);
+  this->write_byte((val >> 8) & 0xFF);
+  this->write_byte(val & 0xFF);
+  this->disable();
+}
+
+float ATM90E26Component::get_line_current_() {
+  uint16_t current = this->read16_(ATM90E26_REGISTER_IRMS);
+  return (float) current / 1000;
```
jesserockz: Try changing these to divide by a float instead of casting them first.
```suggestion
  return current / 1000.0f;
```

jesserockz (commented):
Generally looks good, just one thing on casting to float to divide.

Please add to a test yaml file and create the docs =)

esphome/components/atm90e26/atm90e26.cpp
```diff
@@ -0,0 +1,234 @@
...
+}
+
+float ATM90E26Component::get_power_factor_() {
+  uint16_t val = this->read16_(ATM90E26_REGISTER_POWERF);  // signed
+  if (val & 0x8000) {
+    return (float) -(val & 0x7FF) / 1000;
+  } else {
+    return (float) val / 1000;
+  }
+}
```
jesserockz: Don't forget to change these to divide by a float instead of casting first.

jesserockz (changes requested):
Please change the floats and add a docs PR then this is good to go.

esphome/components/atm90e26/atm90e26.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+float ATM90E26Component::get_forward_active_energy_() {
+  uint16_t val = this->read16_(ATM90E26_REGISTER_APENERGY);
+  if ((UINT32_MAX - this->cumulative_forward_active_energy_) > val) {
+    this->cumulative_forward_active_energy_ += val;
+  } else {
+    this->cumulative_forward_active_energy_ = val;
+  }
+  return ((float) this->cumulative_forward_active_energy_ / 1000);
```
itssimon:
```suggestion
  return (float) this->cumulative_forward_active_energy_;
```

Is the division by 1000 here to report the energy in kWh? It should be Wh as configured in `sensor.py`. The ATM90E32 component also reports this in Wh.

esphome/components/atm90e26/atm90e26.cpp
```diff
@@ -0,0 +1,234 @@
...
+}
+
+float ATM90E26Component::get_reverse_active_energy_() {
+  uint16_t val = this->read16_(ATM90E26_REGISTER_ANENERGY);
+  if (UINT32_MAX - this->cumulative_reverse_active_energy_ > val) {
+    this->cumulative_reverse_active_energy_ += val;
+  } else {
+    this->cumulative_reverse_active_energy_ = val;
+  }
+  return ((float) this->cumulative_reverse_active_energy_ / 1000);
```
itssimon:
```suggestion
  return (float) this->cumulative_reverse_active_energy_;
```

esphome/components/atm90e26/atm90e26.cpp
danieltwagner: Great question, I've definitely not looked at those closely enough.
"Theory of energy registers" https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-46002-SE-M90E26-Datasheet.pdf says this register contains tenths of pulses. The meaning of a pulse is in turn determined by the `PL_Constant` and the metering gain. However, I understand at the end of the day each pulse ought to be 1 Wh so I'll make that change.

danieltwagner:
Apologies for dropping this, I'll try and wrap it up this weekend.

esphome/components/atm90e26/atm90e26.cpp
danieltwagner: Okay after reviewing docs again it turns out this is the same type of pulse as your meter would issue (though mine is ancient and just has a revolving disk, still the same concept) and can be calibrated using the constants mentioned above. Adjusted code accordingly by introducing a new config option with hopefully common default of 3200 pulses/kWh.

danieltwagner:
I've uploaded the documentation in https://github.com/esphome/esphome-docs/pull/3002 and updated the description of this PR to link it. I've also made a couple of small adjustments. I believe this should now address all previous comments and should be ready to be merged.

esphome/components/atm90e26/sensor.py
```diff
@@ -0,0 +1,155 @@
...
+                unit_of_measurement=UNIT_HERTZ,
+                icon=ICON_CURRENT_AC,
+                accuracy_decimals=1,
+                state_class=STATE_CLASS_MEASUREMENT,
+            ),
+            cv.Required(CONF_LINE_FREQUENCY): cv.enum(LINE_FREQS, upper=True),
+            cv.Optional(CONF_METER_CONSTANT, default=3200.0): cv.positive_float,
+            cv.Optional(CONF_PL_CONST, default=1429876): cv.uint32_t,
+            cv.Optional(CONF_GAIN_METERING, default=7481): cv.uint16_t,
+            cv.Optional(CONF_GAIN_VOLTAGE, default=26400): cv.uint16_t,
```
itssimon: Voltage gain should also have range validation. And possibly the metering gain as well?

```suggestion
            cv.Optional(CONF_GAIN_VOLTAGE, default=26400): cv.int_range(min=0, max=32767),
```

esphome/components/atm90e26/sensor.py
```diff
@@ -0,0 +1,155 @@
...
+                state_class=STATE_CLASS_TOTAL_INCREASING,
+            ),
+            cv.Optional(CONF_FREQUENCY): sensor.sensor_schema(
+                unit_of_measurement=UNIT_HERTZ,
+                icon=ICON_CURRENT_AC,
+                accuracy_decimals=1,
+                state_class=STATE_CLASS_MEASUREMENT,
+            ),
+            cv.Required(CONF_LINE_FREQUENCY): cv.enum(LINE_FREQS, upper=True),
+            cv.Optional(CONF_METER_CONSTANT, default=3200.0): cv.positive_float,
```
itssimon: Here in Australia the meter constant is typically 1000. Maybe it's worth calling out in the documentation that this differs in different parts of the world.

esphome/components/atm90e26/sensor.py
```diff
@@ -0,0 +1,155 @@
...
+            cv.Optional(CONF_PL_CONST, default=1429876): cv.uint32_t,
+            cv.Optional(CONF_GAIN_METERING, default=7481): cv.uint16_t,
```
itssimon, line 109: Would you be able to provide a bit more details about the procedure to calibrate the metering gain, and which sensors are impacted by this?

esphome/components/atm90e26/sensor.py
```diff
@@ -0,0 +1,155 @@
...
+CONF_GAIN_METERING = "gain_metering"
+CONF_GAIN_VOLTAGE = "gain_voltage"
+CONF_GAIN_CT = "gain_ct"
+LINE_FREQS = {
+    "50HZ": 50,
+    "60HZ": 60,
+}
+PGA_GAINS = {
+    "1X": 0x4,
+    "4X": 0x0,
```
itssimon, line 44: Have you tested setting pga_gain to anything but 1X? When I set this to 4X the current sensor value read 0.01, while it was around 3A when set to 1X. Is it worth double checking that these numbers are correct?

itssimon (changes requested):

esphome/components/atm90e26/sensor.py
jesserockz: I would make this a required field with no default and document common cases perhaps.

esphome/components/atm90e26/sensor.py
danieltwagner: That info is in the docs; do you think it needs expanding upon? I provided examples there that I hoped would help but am open to any suggestions!

esphome/components/atm90e26/sensor.py
danieltwagner: I haven't tested other values, but if you search for `MMode` in the [datasheet](https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-46002-SE-M90E26-Datasheet.pdf) those are as laid out here. I'm not currently near the device but could try validating on my end next week...

esphome/components/atm90e26/sensor.py
danieltwagner: I suppose that's fair... I've made it required and updated the docs accordingly. Note, however, that the meaning of what a "pulse" is is also entirely up to the user (and is defined by the `meter_gain`). I've referenced it in the docs, too, and the calibration procedure also mentions how these values work together.

I don't think we should make all fields that could have an impact on measurement accuracy as required, as the chip does come with defaults and I think mimicking these here makes sense. Would you agree?

esphome/components/atm90e26/sensor.py
danieltwagner: As far as I can see both voltage and metering gain (ugain and lgain) can take a full uint16 (up to 65k), so this should already be correct. Unless you're saying that using `uint16_t` doesn't enforce 65535 as maximum value?

itssimon: Right, all good then. You might want to clarify that in the docs though, which currently imply that 32768 is the maximum value:

> This would imply a gain_voltage value of 230 / 86.6 * 26400 = 70115. However, the chip’s application note says this value must be below 32768.

esphome/components/atm90e26/sensor.py
itssimon: I agree that mimicking the defaults of the chip is reasonable. Would just suggest a little note in the docs, which I see you've already done, so all good from me!

esphome/components/atm90e26/sensor.py
itssimon: These numbers are indeed correct, however there is an error in the bitwise left-shift logic to calculate MMode. I'll leave comment there instead.

esphome/components/atm90e26/atm90e26.cpp
```diff
@@ -0,0 +1,235 @@
...
+  this->status_clear_warning();
+}
+
+void ATM90E26Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up ATM90E26 Component...");
+  this->spi_setup();
+
+  uint16_t mmode = 0x422;  // default values for everything but L/N line current gains
+  mmode |= l_line_gain_ << 13;
+  mmode |= n_line_gain_ << 11;
```
itssimon, line 49: This is currently overwriting the right-most bit of L line gain.

```suggestion
  mmode |= n_line_gain_ << 10;
```

esphome/components/atm90e26/sensor.py
itssimon: I've double checked the `MMode` calculation and it seems correct indeed according to the datasheet.
(sorry for previous deleted comments, made a mistake there)

esphome/components/atm90e26/sensor.py
danieltwagner: I'm sorry, you were right. Looking with fresh eyes: ugain (volt_gain) should be limited to 15 bits (as documented) while IgainL (ct_gain) shouldn't be. I had swapped them about in the python sensor. I think I confused myself by having the datasheet naming, easier-to-read yaml naming, and then an additional internal name. I've removed the additional internal level of naming to be consistent with the yaml names to make things more obvious. Thanks for the close read.

esphome/components/atm90e26/sensor.py
danieltwagner: I modified it to explicitly mask the ints, not that that should have any influence, but also it makes it more explicit what is happening so ought to help the reader... I then changed 1X to 4X and my current went from 1.18A to 4.72A, so it seems to be working for me.

danieltwagner:
Rebased it against current dev just to make sure. Please lmk your thoughts on the gratuitous bitmasks:
```
  uint16_t mmode = 0x422;  // default values for everything but L/N line current gains
  mmode |= (gain_pga_ & 0x7) << 13;
  mmode |= (n_line_gain_ & 0x3) << 11;
  ```
Happy to remove them again but I think it might help the reader?

jesserockz (approved):

PR #4379, opened 2023-02-02
add MQTT preset support for Climate components

jmichiel:

The current preset state and changing presets through MQTT was still lacking.
ClimateTraits::get_supports_presets() was also updated to be in line with ClimateTraits::get_supports_fan_modes() behaviour.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation:** esphome/esphome-docs#2641

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

Tested with a dummy custom climate: discovery is correct, state changes are published and commands forwarded.

## Checklist:


esphome/components/mqtt/mqtt_climate.cpp
```diff
@@ -66,12 +66,36 @@ void MQTTClimateComponent::send_discovery(JsonObject root, mqtt::SendDiscoveryCo
   // temperature units are always coerced to Celsius internally
   root[MQTT_TEMPERATURE_UNIT] = "C";
 
-  if (traits.supports_preset(CLIMATE_PRESET_AWAY)) {
-    // away_mode_command_topic
-    root[MQTT_AWAY_MODE_COMMAND_TOPIC] = this->get_away_command_topic();
-    // away_mode_state_topic
-    root[MQTT_AWAY_MODE_STATE_TOPIC] = this->get_away_state_topic();
+  if (traits.get_supports_presets() || !traits.get_supported_custom_presets().empty()) {
```
jesserockz, line 69: You added this check inside `get_supports_presets()`
```suggestion
  if (traits.get_supports_presets()) {
```

jesserockz (changes requested):

esphome/components/climate/climate_traits.h
```diff
@@ -107,7 +107,7 @@ class ClimateTraits {
   void add_supported_preset(ClimatePreset preset) { supported_presets_.insert(preset); }
   void add_supported_custom_preset(const std::string &preset) { supported_custom_presets_.insert(preset); }
   bool supports_preset(ClimatePreset preset) const { return supported_presets_.count(preset); }
-  bool get_supports_presets() const { return !supported_presets_.empty(); }
+  bool get_supports_presets() const { return !supported_presets_.empty() || !supported_custom_presets_.empty(); }
```
jesserockz: Just had a quick browse through the code and this change might break existing usages where it is also checking for custom in a different conditional

jesserockz (changes requested):

esphome/components/climate/climate_traits.h
jmichiel: I feared as much and perhaps that's why it wasn't aligned with get_supports_fan_modes?

jmichiel: I reverted the changes

jesserockz (approved):

PR #4391, opened 2023-02-06
Update the delta filter to be able to take a percentage change

TrentHouliston:

Adds functionality to the delta filter so that it is able to check if a new value is a percentage of the old value different if you pass a percentage value.
It maintains the old functionality if you don't use a percentage as your configuration.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2647

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

Example of a percentage filter
```yaml
# Example config.yaml
sensor:
  - platform: ...
    filters:
      - delta: 1%
```

## Checklist:


esphome/components/sensor/__init__.py
```diff
@@ -480,9 +480,21 @@ async def lambda_filter_to_code(config, filter_id):
...
+def validate_delta(config):
+    try:
+        return (cv.positive_float(config), False)
+    except cv.Invalid:
+        pass
+    try:
+        return (cv.percentage(config), True)
+    except cv.Invalid:
+        pass
+    raise cv.Invalid("Delta filter requires a positive number or percentage value.")
```
jesserockz, line 478: This will not work as expected as `0.5` satisfies both `cv.positive_float` (0.5f) and `cv.percentage` (50%)

jesserockz (changes requested):

esphome/components/sensor/__init__.py
TrentHouliston: But `0.5` should satisfy postive_float first and return as an absolute delta value so it will never even try the percentage.

Only if you put a `%` symbol on the end of the value will it fail to work in `positive_float` and throw and then try as a percentage.

jesserockz (approved):

PR #4414, opened 2023-02-11
Gracefully reject vacuum map upload requests

Rjevski:

This implements basic handling of the `0x28` command which is used by robot vacuums to upload map data.

The [Tuya documentation](https://developer.tuya.com/en/docs/iot/tuya-cloud-universal-serial-port-access-protocol?id=K9hhi0xxtn9cb#title-29-(Optional)%20Map%20streaming%20for%20robot%20vacuum) suggests this is an optional (probably paid) feature and the serial protocol has provisions to be able to reject such requests if the service is not enabled.

This PR implements responses to this command with the `0x01` response code (which means the service is not enabled). This is a more graceful way of rejecting it rather than simply ignoring the command and raising an error.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

To test, you'll need to enable the TuyaMCU integration:

```yaml
# Example config.yaml

# disable default logger to free up UART
logger: null

uart:
  id: myuart
  debug:
  # configure your UART here

tuya:
  uart_id: myuart
```

Once done, send `55 aa 03 28 00 00 2a` (bytes in hex) over the serial port associated with TuyaMCU.

You should receive back `55 aa 00 28 00 01 00 28` and an informational message printed in the ESPHome log.

## Checklist:

### DIFF
--- a/esphome/components/tuya/tuya.cpp
+++ b/esphome/components/tuya/tuya.cpp
@@ -246,7 +246,7 @@ void Tuya::handle_command_(uint8_t command, uint8_t version, const uint8_t *buff
     case TuyaCommandType::VACUUM_MAP_UPLOAD:
       this->send_command_(
           TuyaCommand{.cmd = TuyaCommandType::VACUUM_MAP_UPLOAD, .payload = std::vector<uint8_t>{0x01}});
-      ESP_LOGI(TAG, "Vacuum map upload requested, responding that it is not enabled.");
+      ESP_LOGW(TAG, "Vacuum map upload requested, responding that it is not enabled.");
       break;
     default:
       ESP_LOGE(TAG, "Invalid command (0x%02X) received", command);
--- a/esphome/components/tuya/tuya.cpp
+++ b/esphome/components/tuya/tuya.cpp
@@ -243,6 +243,10 @@ void Tuya::handle_command_(uint8_t command, uint8_t version, const uint8_t *buff
       ESP_LOGE(TAG, "LOCAL_TIME_QUERY is not handled");
 #endif
       break;
+    case TuyaCommandType::VACUUM_MAP_UPLOAD:
+      this->send_command_(TuyaCommand{.cmd = TuyaCommandType::VACUUM_MAP_UPLOAD, .payload = std::vector<uint8_t>{0x01}});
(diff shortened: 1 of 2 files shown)


esphome/components/tuya/tuya.cpp
```diff
@@ -243,6 +243,11 @@ void Tuya::handle_command_(uint8_t command, uint8_t version, const uint8_t *buff
       ESP_LOGE(TAG, "LOCAL_TIME_QUERY is not handled");
 #endif
       break;
+    case TuyaCommandType::VACUUM_MAP_UPLOAD:
+      this->send_command_(
+          TuyaCommand{.cmd = TuyaCommandType::VACUUM_MAP_UPLOAD, .payload = std::vector<uint8_t>{0x01}});
+      ESP_LOGI(TAG, "Vacuum map upload requested, responding that it is not enabled.");
```
Rjevski: Should this log a warning instead of info? Not sure what the ESPHome convention is around this. FYI the original log message without this PR would be an error about the command being unknown.

ssieb: Yes, a warning would be good.  I assume that if you respond, the device won't keep sending the query.

esphome/components/tuya/tuya.cpp
```diff
@@ -243,6 +243,11 @@ void Tuya::handle_command_(uint8_t command, uint8_t version, const uint8_t *buff
       ESP_LOGE(TAG, "LOCAL_TIME_QUERY is not handled");
 #endif
       break;
+    case TuyaCommandType::VACUUM_MAP_UPLOAD:
+      this->send_command_(
+          TuyaCommand{.cmd = TuyaCommandType::VACUUM_MAP_UPLOAD, .payload = std::vector<uint8_t>{0x01}});
+      ESP_LOGW(TAG, "Vacuum map upload requested, responding that it is not enabled.");
```
jesserockz, line 253: This probably doesnt need to be logged as a warning here
```suggestion
      ESP_LOGV(TAG, "Vacuum map upload requested, responding that it is not enabled.");
```

jesserockz: Nevermind, just noticed multiple other commands do similar.

jesserockz (approved):

PR #4416, opened 2023-02-12
Add support for MAX31850 1-wire thermocouple amp

lylavoie:

* Updated the Dallas driver to support the MAX31850 device.
* Overloads the DS1825 device id, MAX31850 device is enabled through yaml override.
* Throw ESS_LOGE errors for devices faults (short or open junction).

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2663

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

  - Refer to PR in esphome-docs


Guncotton:
Empirically validated function on ESP8266, two MAX31850 sensors on a NODEMCUv2 monitoring a fireplace. Thank you for the added functionality.

esphome/components/dallas/sensor.py
```diff
@@ -27,6 +28,7 @@
             cv.Optional(CONF_ADDRESS): cv.hex_int,
             cv.Optional(CONF_INDEX): cv.positive_int,
             cv.Optional(CONF_RESOLUTION, default=12): cv.int_range(min=9, max=12),
+            cv.Optional(CONF_MAX31850, default=False): cv.boolean,
```
jesserockz, line 31: I don't know how many others there might be, but I would prefer a `model` enum approach here that defaults to `DS18XXX`

jesserockz (changes requested):

esphome/components/dallas/sensor.py
lylavoie: I can update to that approach, the other DS chips can be "auto," since the model ID in the onewire address / serial number isn't being overloaded yet.  Would you be ok with a model enum that include auto, along with the specific models, i.e. it can force something for a specific sensor?

PR #4442, opened 2023-02-15
Dev mqtt device info

hlyi:

Add MQTT device info (manufacturer, model, name) configuration for discovery message. It addressed hard coded values in code, for example, manufacturer is hardcoded to `espressif`, which is no longer true for rpi2040.

## Types of changes

- [x] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2672

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


hlyi:
`Run script/clang-tidy for ESP32 Arduino 2/4` flagged an error. However, I think the code is correct. It uses newly defined macros in generated `esphome/core/defines.h`. How to resolve the issue?

oxan:
> `Run script/clang-tidy for ESP32 Arduino 2/4` flagged an error. However, I think the code is correct. It uses newly defined macros in generated `esphome/core/defines.h`. How to resolve the issue?

One problem is that `mqtt_component.cpp` doesn't actually include `esphome/core/defines.h`; and the other is that CI doesn't use the generated macros, but the macros from the [dummy defines.h](https://github.com/esphome/esphome/blob/dev/esphome/core/defines.h) that is in the repository.

However, using defines for things like this isn't consistent with ESPHome's architecture: if all config settings were added as defines, we'd have thousands of them, and it'd become an unmanagable mess. Instead, this information should be stored as a member of `MQTTClientComponent` (or `MQTTDiscoveryInfo` is probably even better), and the code generation should generate a method call to set that field.

hlyi:
Thanks for explaining the failure and provide your inputs, @oxan! If `define.h` was manually maintained, what you said make perfect sense. Since `define.h` is generated by the flow, just like other intermediate files, the size/management is lesser a concern IMHO.  The device info is constant, even creating a method, it could be called once at startup. To me, for embedded device like ESP, flash and SRAM are premium sources. Extra code space, SRAM for variable/stack, and additional runtime hit may not be a good tradeoff for the use case, especially if there were hundreds/thousands similar cases in `define.h` .

jesserockz (changes requested):
~~As a side note for this PR, you shouldnt actually be using `defines` for this information. Defines should only be used where things are required to be used across multiple components.~~
Didnt see the above comment before writing this.


But also, You should use similar to the API:
```c++
#if defined(USE_ESP8266) || defined(USE_ESP32)
  resp.manufacturer = "Espressif";
#elif defined(USE_RP2040)
  resp.manufacturer = "Raspberry Pi";
#endif
  resp.model = ESPHOME_BOARD;
```

For custom manufacturer and model, what we do is use the `ESPHOME_PROJECT_NAME` and split it on HA side (when using API
```python
    if device_info.project_name:
        project_name = device_info.project_name.split(".")
        manufacturer = project_name[0]
        model = project_name[1]
        hw_version = device_info.project_version
```

I dont see a point in adding yet another way to define the same information


The `device_info[MQTT_DEVICE_NAME] = node_name;` could be changed to fetch the new `friendly_name` from `App.get_friendly_name()` with a fallback to `node_name`

hlyi:
@jesserockz Thanks for the suggestion! Esphome API has its own integration in HA, so splitting ESPHOME_PROJECT_NAME on HA work. However, HA MQTT integration has the standard method to specify device information ( via discovery message) that we need to follow. 

I think you had a good point that  Native API and MQTT should extract the information from the same configuration. How could  `mqtt/__init__.py` get the value of ESPHOME_PROJECT_NAME ( or equivalent python variable)?

jesserockz:
The define is already there as `ESPHOME_PROJECT_NAME` if the user has specified it, so the mqtt cpp code can check for that with `#ifdef ESPHOME_PROJECT_NAME`

The user can set it with
```yaml
esphome:
  project:
    name: jesserockz.my_sensor
    version: 1.0
```

hlyi:
@jesserockz Based upon your suggestion, now manufacturer and model are derived from CORE.config.project.name. still need your help to bypass the error from  'run script/clang-tidy for ESP32 Arduino 2/4'.

BTW, I could not find App.get_friendly_name() function in 2022.12.8 release.

jesserockz:
> BTW, I could not find App.get_friendly_name() function in 2022.12.8 release.

Thats because it was released in 2023.2.0
Components should only be developed against the dev branch anyway, not a stable release as there could be API changes already.

jesserockz:
> @jesserockz Based upon your suggestion, now manufacturer and model are derived from CORE.config.project.name. still need your help to bypass the error from 'run script/clang-tidy for ESP32 Arduino 2/4'.

You are still doing it wrong. You dont need to make any modifications in python for this. Just check if the `ESPHOME_PROJECT_NAME` macro is defined and use that directly in cpp code to set the fields on the json object. If `ESPHOME_PROJECT_NAME` is not defined, the code basically should not change from what it is now.

hlyi:
> You are still doing it wrong. You dont need to make any modifications in python for this. Just check if the `ESPHOME_PROJECT_NAME` macro is defined and use that directly in cpp code to set the fields on the json object. If `ESPHOME_PROJECT_NAME` is not defined, the code basically should not change from what it is now.

As mentioned here, I deliberately avoid extracting those constants  in ESPHOME cpp code, due to limited flash/SRAM and runtime hit for embedded system. As you mentioned earlier, native API doesn't process it either and let HA integration to do it. For MQTT, HA already had a standard way to specify device info, it is not a good idea to change it. So, I moved the processing to build time, instead of run time.

oxan:
> If `define.h` was manually maintained, what you said make perfect sense. Since `define.h` is generated by the flow, just like other intermediate files, the size/management is lesser a concern IMHO. 

The problem is not just the size of the file, it's also that preprocessor macros aren't namespaced and aren't typed, which increases the odds of conflicts and reduces the ability to statically detect problem. If hypothetically we would want to move more things to compile-time, we would go with a `constexpr`-based solution, not the preprocessor.

> The device info is constant, even creating a method, it could be called once at startup. To me, for embedded device like ESP, flash and SRAM are premium sources. Extra code space, SRAM for variable/stack, and additional runtime hit may not be a good tradeoff for the use case, especially if there were hundreds/thousands similar cases in `define.h` .

While you're theoretically right, in practice this isn't a problem, as the ESP cores are fairly big for embedded standards. It's a whole different world from the 8-bit Arduinos where squeezing every byte and cycle could actually be useful (but even there it's usually not necessary). The disadvantages in maintainability outweigh the advantages of saving a few bytes. Especially because the rest of ESPHome doesn't follow this pattern, and consistency in a codebase is much more important than saving a couple bytes here.

Besides, at least on ESP8266 but if I remember correctly also on ESP32, the `.rodata` section is still loaded into memory, so unless you do special tricks the string will end up in memory anyway (and if you preprocess it instead of using it as-is, it can't be deduplicated and will be loaded into memory twice!).

PR #4456, opened 2023-02-17
Replace name and friendly name in full adopted configs

jesserockz:

This fixes an issue where setting `name_add_mac_suffix: true` and `import_full_config: true` together results in a local config after adoption without the known mac suffix.

This does direct string replacements on the hardcoded name and friendly name to update them with the mac suffix value.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/dashboard_import/__init__.py
```diff
@@ -82,56 +81,48 @@ def import_config(
...
+        try:
+            req = requests.get(url, timeout=30)
+            req.raise_for_status()
+        except requests.exceptions.RequestException as e:
+            raise ValueError(f"Error while fetching {url}: {e}") from e
+
+        contents = req.text
+
+        # remove mac suffix (-######) to get original name
+        original_name = name[:-7]
```
balloob: shouldn't you parse it first to make sure it had a Mac suffix? Or is that mandatory for import url?

jesserockz: Good point, its not mandatory

nielsnl68:
What is the status of this PR?

jesserockz:
I still need to finish this PR off

pszafer:
@jesserockz do you plan to work on that in nearby future? I'm not pushing, just asking, because if no, then I'd like to make some PR to esphome creators docs, that `import_full_config: true` doesn't allow `creators` device to be successfully imported by end user. I' ve spent whole weekend trying to get it working until I started to look in source code ;)

jesserockz:
@pszafer Thanks for pushing me. I have made some more updates to the PR that does the replacement a bit cleaner.

PR #4461, opened 2023-02-18
Add binary sensors for pvvx_mithermometer flags

QRPp:

Binary sensors for pvvx_mithermometer flags field (the five bits currently provided).

(Also AUTO_LOAD esp32_ble_tracker instead of just depending on it, simpler usage that way.)

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
sensor:
  - platform: pvvx_mithermometer
    mac_address: 11:22:33:44:55:66
    rds_input:
      name: hall_motion
      device_class: motion
      filters:
        - delayed_on: 15s
        - delayed_off: 300s
```

## Checklist:

If the PR in principle looks good, can create and send another one for the docs too, of course.

### DIFF
--- a/esphome/components/pvvx_mithermometer/sensor.py
+++ b/esphome/components/pvvx_mithermometer/sensor.py
@@ -1,6 +1,6 @@
 import esphome.codegen as cg
 import esphome.config_validation as cv
-from esphome.components import sensor, esp32_ble_tracker
+from esphome.components import binary_sensor, sensor, esp32_ble_tracker
 from esphome.const import (
     CONF_BATTERY_LEVEL,
     CONF_BATTERY_VOLTAGE,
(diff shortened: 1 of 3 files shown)


esphome/components/pvvx_mithermometer/sensor.py
```diff
@@ -24,13 +24,19 @@
 
 CODEOWNERS = ["@pasiz"]
 
-DEPENDENCIES = ["esp32_ble_tracker"]
+AUTO_LOAD = ["binary_sensor", "esp32_ble_tracker"]
```
jesserockz, line 27: Some things are not auto loaded for a reason =)

Adding `binary_sensor` here will load in a lot of code that people not using these flags don't need.
At this point, it should be split out into a hub/component structure if there are multiple types.

jesserockz (changes requested):

esphome/components/pvvx_mithermometer/sensor.py
pasiz: Exact. Most users aren't using binary sensor, just want temperature and humidity.

PR #4462, opened 2023-02-18
Simplify sensor schema generation

balloob:

The sensor schema generation was doing a lot of unnecessary work. Cleaned it up

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/sensor/__init__.py
+++ b/esphome/components/sensor/__init__.py
@@ -272,48 +272,21 @@ def sensor_schema(
     state_class: str = _UNDEF,
     entity_category: str = _UNDEF,
 ) -> cv.Schema:
-    schema = SENSOR_SCHEMA
-    if class_ is not _UNDEF:
-        schema = schema.extend({cv.GenerateID(): cv.declare_id(class_)})
-    if unit_of_measurement is not _UNDEF:
-        schema = schema.extend(
-            {
-                cv.Optional(
-                    CONF_UNIT_OF_MEASUREMENT, default=unit_of_measurement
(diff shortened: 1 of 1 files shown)


nielsnl68 (approved):
This is a very cool idea.  
We should document this idea somewhere to use in other components as well

balloob:
Let's consider that if I get it to work ;-)

oxan (approved):
An idea I've had is to swap the definitions of `sensor_schema()` and `SENSOR_SCHEMA` around, so that the function has the canonical definition of the schema in it and the variable is just the result of the function without any defaults set. That'd reduce the duplication even further. Something in the style of this example:

```python
def sensor_schema(
    class_: MockObjClass = cv.UNDEFINED,
    unit_of_measurement: str = cv.UNDEFINED,
    icon: str = cv.UNDEFINED,
) -> cv.Schema:
    return cv.ENTITY_BASE_SCHEMA.extend({
        cv.GenerateID(): cv.declare_id(class_),
        cv.Optional(CONF_UNIT_OF_MEASUREMENT, default=unit_of_measurement): validate_unit_of_measurement,
        cv.Optional(CONF_ICON, default=icon): validate_icon,
    })

SENSOR_SCHEMA = sensor_schema(Sensor)
```

In any case, this PR is already an improvement, so I'm fine with merging this and leaving further improvements for the future.

jesserockz (approved):

PR #4464, opened 2023-02-18
Feat: Add GREE climateir component

PiuPiuson:

Added support for GREE air conditioners on climateir. Only transmit supported at the moment; will try to implement receive when schedule allows.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2685

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
climate:
  - platform: gree
    name: "living_room_ac"
    id: climate_ac
    model: yan
    transmitter_id: ir_transmit
    sensor: temp
```

## Checklist:


nagyrobi:
Finally! Gonna test this today. `heatpumpir` also suppprts Gree protocols, but in reality they don't work (at leat not with my three units here).
Do you plan to add the Gree _I-Feel_ functionality? (send periodically the current room temperature back to the unit via IR for more precise temp control).

PiuPiuson:
Glad to see someone found this useful!
Sadly I don't have an i-feel compatible unit so no way to test :/ 

Also if you check the fork on my profile I've got some rudimentary RX working (temp, mode, status) so feel free to pull and give that a try as well

nagyrobi:
I have 3 Airwell awsi-hdd012-n11 devices. They are commanded by IR protocol GREE (as discovered by Tasmota's IRHVAC implementation). I'm using them with Tasmota and [tasmota_irhvac](https://github.com/hristo-atanasov/Tasmota-IRHVAC) HA custom component but I really want to get rid of that and switch to ESPHome - and your component is the right way to do it!

So I've tested it, and with this component model `generic` doesn't work, but `yan`, `yaa`, `yac` all seem to work.

I also use `esp8266_restore_from_flash: true` setting to keep the last state of the devices saved when the IR blaster is rebooted or power-cycled. Seemed like `yaa` model behaved faulty in the regard that after power-cycling the node the climate was reset to mode `auto` when adjusting temperature. Does your component support saving the states with `esp8266_restore_from_flash`?

In mode `auto` target temp stays stuck to 23 degrees on the unit's display, despite setting it to any value on the UI. Changing to mode `heat` allows me to select any temperature (I'm using the units for heating now).

My units can only swing vertical, so options _horizontal_ and _both_ could be removed in this case. My suggestion would be to have an override in the config, like 
```
    supports_swing_vertical: false
    supports_swing_horizontal: false
```
where both options could be `true` by default if not specified, and the user could set one of them or both to false if his device doesn't support it so it would be removed from the supported features thus not showing in the UI.

Some binary things which are missing:
- There's an extra "Turbo" button which increases fan speed quite some more
- These units have on their remote a "Light" button, where I can turn off or on the LED display on the unit. This is important in bedrooms where a strong lighting display is disturbing during the night. 
- Also there's "Beep" to disable beeping of the unit when receiving commands
- There's also a "Quiet" button which makes the unit operating less loud
- Also there's an "Eco" button to reduce energy consumption
- There's also a "Clean" setting to reset the clean filter warning light

For the "Turbo", I'd suggest to simply add it to the fan speeds list if possible, a little trickery would be needed where the real speed is `max` + the binary variable in the IR codes corresponding to Turbo set to 1. 
```
    supports_turbo_focus: true
```
As for Turbo naming, perhaps `CLIMATE_FAN_FOCUS` would be suitable like [somebody tried here](https://github.com/Hagakurje/ESPHome_Gree_AC/blob/7aeeb394cfacf5075aa4dbcbca07fb16949c0525/gree_ir.h#L130). If not, then just like a binary switch as below.

For the rest of the binary settings, it would be nice if we had optional settings pointing to template switches in the config:
```
    light_switch_id: sw_light
    beep_switch_id: sw_beep
    quiet_switch_id: sw_quiet
    eco_switch_id: sw_eco
    clean_switch_id: sw_clean
```
so one could simply add ESPHome template switches  with the IDs set as above. Or of course the component could implement them on its own, but not sure if it's possible. These could be optional, meaning if not specified, the feature would be missing, just like now.
(all the above works with Tasmota's IRHVAC)

tests/test1.yaml
```diff
@@ -3203,25 +3206,25 @@ lcd_menu:
       lambda: 'ESP_LOGI("lcd_menu", "root leave");'
   items:
     - type: back
-      text: 'Back'
```
jesserockz: Please revert these formatting changes to unrelated code.

nagyrobi:
At least a "Light" switch would be nice to be implemented...

jesserockz (dismissed):

jesserockz (dismissed):

PR #3917, opened 2022-10-18
add usb hid keyboard emulation

tomaszduda23:
# What does this implement/fix?
Emulating a USB keyboard.

It needs:
- https://github.com/esphome/esphome/pull/7520
- https://github.com/esphome/esphome/pull/7513

Optional:
- https://github.com/esphome/esphome/pull/7210

**Quick description and explanation of changes**
It adds dependency to https://github.com/adafruit/Adafruit_TinyUSB_Arduino.
## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

## TODO
- [x] add esphome/components/web_server/server_index_v2_keyboard.h
- [x] framework-arduinoespressif32 uses tinyusb behind
- [x] part of implementation is duplicated in this PR already
- [ ] check esp32-idf https://github.com/espressif/esp-idf/blob/636ff35b52f10e1a804a3760a5bd94e68f4b1b71/examples/peripherals/usb/device/tusb_hid/main/tusb_hid_example_main.c
- [ ] check rpi pico https://github.com/raspberrypi/pico-examples/tree/master/usb/device/dev_hid_composite
- [x] check how to add new device type. This is needed for web server.

**Related issue or feature (if applicable):** fixes <link to issue>
https://github.com/esphome/feature-requests/issues/1918
**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x]  ESP32-S2
- [x]  ESP32-S3

## Example entry for `config.yaml`:

```yaml
external_components:
  - source: github://pr#3917
    components: [hid, keyboard, usb_device, web_server, component_iterator, api]

    refresh: always

esphome:
  name: esp-white-control
  board: lolin_s2_mini
  platform: ESP32

web_server:
  local: true

button:
  - platform: template
    name: "Press ShiftLeft + B"
    on_press:
      then:
        - keyboard.down:
            id: kbd
            keys: 
              - ShiftLeft
              - 5
        - delay: 100ms
        - keyboard.set:
            id: kbd
  - platform: template
    name: "Press K"
    on_press:
      then:
        - keyboard.down:
            id: kbd
            keys: KeyK
[...]

PR #4470, opened 2023-02-19
remove dependencies to componentes from core

tomaszduda23:

1. remove dependencies to componentes from core. Right now components depend on core and core depends on components.
2. remove duplicate code
3. add unit tests
4. This allows to add new component type to web_server and api with external_components

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/core/application.h
```diff
@@ -72,59 +33,32 @@ class Application {
     this->compilation_time_ = compilation_time;
   }
 
-#ifdef USE_BINARY_SENSOR
-  void register_binary_sensor(binary_sensor::BinarySensor *binary_sensor) {
-    this->binary_sensors_.push_back(binary_sensor);
+  template<class Entity> const std::vector<Entity *> get_entity() {
```
jesserockz: Shouldn't this be plural?

```suggestion
  template<class Entity> const std::vector<Entity *> get_entities() {
```

esphome/core/application.h
```diff
@@ -72,59 +33,32 @@ class Application {
...
-  void register_binary_sensor(binary_sensor::BinarySensor *binary_sensor) {
-    this->binary_sensors_.push_back(binary_sensor);
+  template<class Entity> const std::vector<Entity *> get_entities() {
+    static_assert(std::is_base_of<EntityBase, Entity>::value, "Only EntityBase subclasses can be used");
+    resize_(Entity::ENTITY_TYPE);
+    std::vector<Entity *> result;
+    for (auto obj : this->entities_[Entity::ENTITY_TYPE]) {
+      result.push_back(static_cast<Entity *>(obj));
+    }
+    return result;
```
oxan: Returning a vector by const value doesn't make much sense, especially since we specifically create a copy. The old code returned the vector by const-reference, can't we do the same with `this->entities_[Entity::ENTITY_TYPE]`?

I also dislike the fact that a member is modified in a getter, ideally this'd be a const member.

esphome/core/controller.h
```diff
@@ -1,87 +1,43 @@
...
+
+  template<typename Func> void add_on_state_callback(Func &&fn) {
+    using Entity = typename std::remove_pointer<typename std::tuple_element<0, arguments_t<Func>>::type>::type;
+    static_assert(std::is_base_of<EntityBase, Entity>::value, "Only EntityBase subclasses can be used");
+    for (auto &obj : App.get_entities<Entity>()) {
+      if (include_internal_ || !obj->is_internal()) {
+        obj->add_on_state_callback(AutoLambda<callback_t<Func>>::make_lambda(fn, obj));
+      }
+    }
+  }
```
oxan, line 33: I think it'd be easier and cleaner to add a state callback that doesn't have the state argument to all entities. Half of them already don't pass the state, and you can easily access the state once you got the object anyway.

esphome/core/controller.h
```diff
@@ -1,87 +1,43 @@
...
+
+  template<typename Func> void add_new_remote_values_callback(Func &&fn) {
+    using Entity = typename std::remove_pointer<typename std::tuple_element<0, arguments_t<Func>>::type>::type;
+    static_assert(std::is_base_of<EntityBase, Entity>::value, "Only EntityBase subclasses can be used");
+    for (auto &obj : App.get_entities<Entity>()) {
+      if (include_internal_ || !obj->is_internal()) {
+        obj->add_new_remote_values_callback(AutoLambda<callback_t<Func>>::make_lambda(fn, obj));
+      }
+    }
+  }
```
oxan: Let's not merge the complexity of the light component into core, and instead add a `add_on_state_callback` as alias for `add_new_remote_values_callback` in the `LightState` class.

esphome/core/controller.h
```diff
@@ -1,87 +1,43 @@
...
+
+ protected:
+  bool include_internal_{false};
 };
```
oxan, line 34: I'm not sure it still makes sense to have `Controller` as a parent class once there's just `add_on_state_callback` left.

tests/gtest/application_test.cpp
```diff
@@ -0,0 +1,65 @@
+
```
oxan: I like having tests, and I'm not opposed to googletest, but having tests that aren't run regularly is a recipe for bitrot in my experience. Can we add this to CI? 

Maybe it's best to split out test infrastructure to a separate PR actually, though I can see how that would complicate things (GitHub still lacks proper support for stacked PRs :disappointed:) @jesserockz what do you think?

esphome/core/component_iterator.h
```diff
@@ -21,106 +22,42 @@ class ComponentIterator {
...
+    using Entity = typename std::remove_pointer<typename std::tuple_element<0, arguments_t<Func>>::type>::type;
+    static_assert(std::is_base_of<EntityBase, Entity>::value, "Only EntityBase subclasses can be used");
+    // it is done that way to allow adding custom components without changing core
+    if (callbacks_.size() < Entity::ENTITY_TYPE + 1) {
+      callbacks_.resize(Entity::ENTITY_TYPE + 1);
+    }
+    auto cb = [fn](EntityBase *entity) -> bool { return fn(static_cast<Entity *>(entity)); };
+    callbacks_[Entity::ENTITY_TYPE] = cb;
+  }
+
```
oxan, line 45: Having both the `on_*` event-style methods and `add_*` callback-style methods in the same API is weird. I quite like the `on_*` event-style methods here, is there any way we can preserve that?

oxan (commented):
I'm not sure I quite understand what you want to achieve with this. Adding a new entity type seems quite useless without changing api/mqtt/web_server/etc as well, so an external component can't really provide any functionality through a new entity type.

Additionally, I assume the idea is that new entity types set their `ENTITY_TYPE` constant to `EntityType::MAX`? That means that you can only use external components that define one entity type, which seems a bit restrictive.

However, I actually like this PR, just for the code deduplication (and because it'll make my life easier in a project I'm working on :wink: ). I've left a few comments on the design below.

tomaszduda23:
> I'm not sure I quite understand what you want to achieve with this. Adding a new entity type seems quite useless without changing api/mqtt/web_server/etc as well, so an external component can't really provide any functionality through a new entity type.

I would like to use my other PR https://github.com/esphome/esphome/pull/3917 as external_components but since it had to touch core I'm not able to. If this get merge I can replace web_server without touching the core.

And removing duplicated code is always good.

> 
> Additionally, I assume the idea is that new entity types set their `ENTITY_TYPE` constant to `EntityType::MAX`? That means that you can only use external components that define one entity type, which seems a bit restrictive.

EntityType::MAX is not used in the code. vector is resized each time when you register new entity type so you can just add custom like  EntityType::MAX, EntityType::MAX+1, EntityType::MAX+2.

esphome/core/controller.h
tomaszduda23: That make sense but I would rather to put it as separate PR. I'm not sure what strategy you have here. If this PR gets too big it might be difficult to merge.

tests/gtest/application_test.cpp
tomaszduda23: I was going to add it to CI. I was thinking about 2 options:
1. As test in platformio.ini <- I've not take a look into this yet but it may be difficult since some of files won't compile on x86.
2. Just add a script which call meson setup and meson test

esphome/core/application.h
tomaszduda23: > Returning a vector by const value doesn't make much sense, especially since we specifically create a copy. The old code returned the vector by const-reference, can't we do the same with `this->entities_[Entity::ENTITY_TYPE]`?

I did so to avoid casting on caller side. Anyway it looks strange.

tests/gtest/application_test.cpp
jesserockz: Moving the test stuff into a new PR is the best option, until this PR is merged, the second PR will have the same changes, but once this is merged it can be rebased and made smaller to look at again.

tomaszduda23: https://github.com/esphome/esphome/pull/4473

esphome/core/controller.h
tomaszduda23: I just took a look. This would require to touch several components. Making it compile seems to be easy. It will be difficult to make sure that everything works as before though.

tomaszduda23: Callbacks are unified in web_server and api_server so I consider this done since I don't want to touch all the components.

oxan:
> I would like to use my other PR #3917 as external_components but since it had to touch core I'm not able to. If this get merge I can replace web_server without touching the core.

From a quick look I don't think `keyboard` actually makes much sense as an entity, especially since it doesn't actually integrate with the MQTT/HA API like all other entities do. Providing a better way to extend the webserver seems like the better solution there.

> And removing duplicated code is always good.

That I can agree with.

> EntityType::MAX is not used in the code. vector is resized each time when you register new entity type so you can just add custom like EntityType::MAX, EntityType::MAX+1, EntityType::MAX+2.

Yes, I understand that, but it still requires cooperation between the external components you use to avoid them both using e.g. `EntityType::MAX+1`.

I've been thinking a bit about this over the past week, and I think there's a better solution than using a vector-of-vectors: a tuple. This has the advantage that entity types are identified by their actual type instead of an integer from the `EntityType` enumerator, so you can't ever have conflicts, and it does not incur one additional pointer indirection on every entity access. Something like this:

```cpp
template<typename... Ts>
class EntityRegistry
{
public:
  template<typename T>
  const std::vector<T *>& get_entities() {
    return std::get<std::vector<T *>, std::vector<Ts *>...>(entities_);
  }
  template<typename T>
  T* get_entity_by_key(uint32_t key) {
    for (auto *obj : get_entities<T>())
      if (obj->get_object_id_hash() == key)
        return obj;
    return nullptr;
  }

private:
  std::tuple<std::vector<Ts *>...> entities_{};
};
```

Now, this requires all entity types to be passed as template parameters, but we can do that at the codegen phase. One problem I haven't worked out is that I'd really like `Application` not to become a template type, but we can probably work something out there.

oxan:
> One problem I haven't worked out is that I'd really like `Application` not to become a template type, but we can probably work something out there.

I guess what we can do is add an autogenerated header `esphome/core/globals.h` (like `defines.h` and `version.h`), have it declare an `extern EntityRegistry<...> EntityRegistryInstance`. Then that header can be included from `application.h`, the `Application` methods can call methods on that instance, and we can define the `EntityRegistryInstance` object in the generated `main.cpp`. It's all a bit convoluted, but it encapsulates the actual implementation from the API (so it can always be changed later), and it doesn't add indirections.

I'm open for better ideas, though.

tomaszduda23:
> From a quick look I don't think `keyboard` actually makes much sense as an entity, especially since it doesn't actually integrate with the MQTT/HA API like all other entities do. Providing a better way to extend the webserver seems like the better solution there.
> 
You still need to list keyboard up to show it on the web. Making entity was pretty natural.
I was thinking about using HA API as well since it could be controlled from HA. Keyboard requires API to be pretty interactive so not sure if it would work (restAPI is too slow so I used websocket).

I like tuple idea. Generating global.h would work. You just need something like
```
namespace sensor{
class Sensor;
}
using entities_t = std::tuple<std::vector<sensor::Sensor*>>;
```
Than the rest would be implemented as before.

tomaszduda23:
I made simple update using std::touple. It requires c++17 to compile. If you fine with that approach I can make it compile for c++11. Then add generator for global.h.

oxan:
> I like tuple idea. Generating global.h would work. You just need something like

Yes, I like the `using` approach more than the convoluted mess I described in https://github.com/esphome/esphome/pull/4470#issuecomment-1455192840. Probably should name the header `entities.h` or something then, though.

oxan:
> I made simple update using std::touple. It requires c++17 to compile. If you fine with that approach I can make it compile for c++11. Then add generator for global.h.

Yes, I like this approach! I'm not sure even the `EntityRegistry` type is needed, we can just put `std::tuple<entities_t> entities_` in `Application.h` directly.

tomaszduda23:
> Yes, I like this approach! I'm not sure even the `EntityRegistry` type is needed, we can just put `std::tuple<entities_t> entities_` in `Application.h` directly.

`using entities_t = std::tuple<binary_sensor::BinarySensor *>;` has different type since it is used in two places so it seems that EntityRegistry is needed.

It can be review again.

tomaszduda23:
@oxan would you mind to review again?

PR #4482, opened 2023-02-24
Add check for completed upload size to prevent incomplete updates and potential bootloops

nuclearcat:
In certain situations, the update process may terminate prematurely due to connection resets or other bugs, resulting in an incomplete update. This can lead to a bootloop, which would require manual recovery and be particularly inconvenient if the device is installed in a hard-to-reach location.

To avoid this issue, a simple check for completed upload size can be added to ensure that the update is fully uploaded before proceeding.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/web_server_base/web_server_base.cpp
```diff
@@ -78,6 +78,11 @@ void OTARequestHandler::handleUpload(AsyncWebServerRequest *request, const Strin
   }
 
   if (final) {
+    // incomplete update! probably tcp reset, bad wifi, etc
+    if (request->contentLength() != this->ota_read_length_) {
+      ESP_LOGI(TAG, "OTA update incomplete, terminating");
+      return;
```
jesserockz, line 88: This should also call abort to clear out anything
```suggestion
      Update.abort();
      return;
```

jesserockz (changes requested):

jesserockz (approved):

jesserockz (changes requested):
Just actually tested this on ESP32 and its failing everytime, the content length sent from the browser actually includes more than just the firmware bytes...

```
[07:39:42][I][web_server_base:083]: Size: 1006778 != 1006576
[07:39:42][I][web_server_base:084]: OTA update incomplete, terminating
```


image


```shell
$ ls -al firmware.bin
-rw-r--r-- 1 jesse users 1006576 Mar  9 07:38 firmware.bin
```

nuclearcat:
It seems there is no way to get real data length from ESPAsyncWebServer. It exists in private _parsedLength, but we can't access it.
Also i am having problems to reproduce OTA corruption issue reliably, it happens only in one place with unreliable wifi network, and sensor is installed in very hard to reach place, so it is difficult to debug it.

Best i got:
```
timeout -s KILL 3 curl -v --limit-rate 100k -F 'upload=@acmainroom.bin' http://acmainroom.local/update
```
But it doesn't set "final", so OTA just ends in incomplete state, and if i restart it start from beginning.

So i have to investigate this corruption issue more, will test it better next time, sorry for disturbing.

PR #4485, opened 2023-02-25
Move Font glyphs to SPI RAM.

Fabian-Schmidt:

Move the Font glyph data to SPI RAM (if possible). Given that Display devices are expected to have SPI RAM more infrequent data can be moved to the SPI RAM.

In my scenario with 3 Fonts this saves 3kb of heap size and some fragmentation.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -256,7 +256,7 @@ void DisplayBuffer::print(int x, int y, Font *font, Color color, TextAlign align
     if (glyph_n < 0) {
       // Unknown char, skip
       ESP_LOGW(TAG, "Encountered character without representation in font: '%c'", text[i]);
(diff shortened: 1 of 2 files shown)


nielsnl68:
What happens when you do not have SPI RAM (btw is that the same as PSRAM)?

jesserockz (approved):

jesserockz:
> What happens when you do not have SPI RAM (btw is that the same as PSRAM)?

The ExternalRamAllocator will use regular ram if there is no PSRAM, it has flags to not fallback to normal ram, but this is not set in this case.

Yes SPI RAM and PSRAM are the same thing

PR #4486, opened 2023-02-25
Use PSRam for BLE scan results.

Fabian-Schmidt:

Given that the Bluetooth stack is consuming so much memory. I try to reduce the memory consumption a bit by moving the BLE scan results to SPI RAM (PSRAM).  It saves 3kb.

Also if PSRAM is used more scan results can be cached,

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
+++ b/esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
@@ -55,9 +55,9 @@ void ESP32BLETracker::setup() {
   }
   ExternalRAMAllocator<esp_ble_gap_cb_param_t::ble_scan_result_evt_param> allocator(
       ExternalRAMAllocator<esp_ble_gap_cb_param_t::ble_scan_result_evt_param>::ALLOW_FAILURE);
-  this->scan_result_buffer_ = allocator.allocate(this->scan_result_buffer_size_);
+  this->scan_result_buffer_ = allocator.allocate(ESP32BLETracker::SCAN_RESULT_BUFFER_SIZE);
   memset(this->scan_result_buffer_, 0,
-         sizeof(esp_ble_gap_cb_param_t::ble_scan_result_evt_param) * this->scan_result_buffer_size_);
+         sizeof(esp_ble_gap_cb_param_t::ble_scan_result_evt_param) * ESP32BLETracker::SCAN_RESULT_BUFFER_SIZE);
 
   global_esp32_ble_tracker = this;
   this->scan_result_lock_ = xSemaphoreCreateMutex();
@@ -112,7 +112,7 @@ void ESP32BLETracker::loop() {
         xSemaphoreTake(this->scan_result_lock_, 5L / portTICK_PERIOD_MS)) {
       uint32_t index = this->scan_result_index_;
       if (index) {
-        if (index >= this->scan_result_buffer_size_) {
+        if (index >= ESP32BLETracker::SCAN_RESULT_BUFFER_SIZE) {
           ESP_LOGW(TAG, "Too many BLE events to process. Some devices may not show up.");
         }
         for (size_t i = 0; i < index; i++) {
@@ -327,7 +327,7 @@ void ESP32BLETracker::gap_scan_stop_complete_(const esp_ble_gap_cb_param_t::ble_
 void ESP32BLETracker::gap_scan_result_(const esp_ble_gap_cb_param_t::ble_scan_result_evt_param &param) {
   if (param.search_evt == ESP_GAP_SEARCH_INQ_RES_EVT) {
     if (xSemaphoreTake(this->scan_result_lock_, 0L)) {
-      if (this->scan_result_index_ < this->scan_result_buffer_size_) {
+      if (this->scan_result_index_ < ESP32BLETracker::SCAN_RESULT_BUFFER_SIZE) {
         this->scan_result_buffer_[this->scan_result_index_++] = param;
       }
       xSemaphoreGive(this->scan_result_lock_);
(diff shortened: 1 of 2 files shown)


esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
```diff
@@ -231,7 +231,12 @@ class ESP32BLETracker : public Component, public GAPEventHandler, public GATTcEv
   SemaphoreHandle_t scan_result_lock_;
   SemaphoreHandle_t scan_end_lock_;
   size_t scan_result_index_{0};
-  esp_ble_gap_cb_param_t::ble_scan_result_evt_param scan_result_buffer_[16];
+#if BOARD_HAS_PSRAM
```
jesserockz: This flag is only added when using arduino as a framework. (We probably could just add it for both arduino and esp-idf tbh)

Fabian-Schmidt: Using the generic `CONFIG_SPIRAM` flag instead?
It is used by Arduino and ESP-IDF.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -53,6 +53,11 @@ void ESP32BLETracker::setup() {
     ESP_LOGE(TAG, "BLE Tracker was marked failed by ESP32BLE");
     return;
   }
+  ExternalRAMAllocator<esp_ble_gap_cb_param_t::ble_scan_result_evt_param> allocator(
+      ExternalRAMAllocator<esp_ble_gap_cb_param_t::ble_scan_result_evt_param>::ALLOW_FAILURE);
+  this->scan_result_buffer_ = allocator.allocate(ESP32BLETracker::SCAN_RESULT_BUFFER_SIZE);
+  memset(this->scan_result_buffer_, 0,
+         sizeof(esp_ble_gap_cb_param_t::ble_scan_result_evt_param) * ESP32BLETracker::SCAN_RESULT_BUFFER_SIZE);
```
oxan: This will blow up if the allocation fails. Either the `ALLOW_FAILURE` flag should be removed from the allocator, or a check to see if `this->scan_result_buffer_` is not nullptr should be added before writing to it.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
```diff
@@ -231,7 +231,12 @@ class ESP32BLETracker : public Component, public GAPEventHandler, public GATTcEv
   SemaphoreHandle_t scan_result_lock_;
   SemaphoreHandle_t scan_end_lock_;
   size_t scan_result_index_{0};
-  esp_ble_gap_cb_param_t::ble_scan_result_evt_param scan_result_buffer_[16];
+#if CONFIG_SPIRAM
+  const static u_int8_t SCAN_RESULT_BUFFER_SIZE = 32;
+#else
+  const static u_int8_t SCAN_RESULT_BUFFER_SIZE = 16;
+#endif  // CONFIG_SPIRAM
+  esp_ble_gap_cb_param_t::ble_scan_result_evt_param *scan_result_buffer_;
```
oxan, line 239: I think converting `scan_result_buffer_` to an `std::vector` is a cleaner solution, though I'm not sure how involved that'd be.

oxan (changes requested):

esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
Fabian-Schmidt: The existing solution with C style array is working fine. No need to 'fix' it.

Also `vector` can (and do) cause a lot of memory fragmentation and this is already an issue especially when using Bluetooth, Wifi and Display, memory is already tight. My opinion is that less usage of `vector` or only with `reserve` and `shrink_to_fit` operations are needed.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
Fabian-Schmidt: Adjusted the call and changed `ALLOW_FAILURE` to `NONE`.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
oxan: Using scissors to cut your lawn also works, but that doesn't mean it's the best solution ;)

Kidding aside, in general we prefer not to do manual memory management, as it's much easier to make mistakes when doing that as opposed to using STL containers. In fact, there's a CI check that prohibits use of malloc/free - I should add the `allocate()` function to that. If memory fragmentation is a problem, we can indeed use `reserve`, or use a pointer to an `std::array` instance instead.

In any case, given that a C-style array is already used here (and the BLE-code is already quite messy), I don't really oppose this change and will leave it up to @jesserockz.

jesserockz: I think its fine as it is and we can revisit changing the type later with some more testing if needed.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
jesserockz: Changing this to NONE is going to `abort` when it cannot allocate anywhere, which is almost certainly going to cause a boot loop until it gets into safe mode.

It would be quite rare to not be able to allocate the memory, but I think the ALLOW_FAILURE with a nullptr check afterwards is better as it can then mark the component as failed like other components do now.

Fabian-Schmidt: Okay. I changed it to mark component as failed instead of abort.

jesserockz (approved):

PR #4490, opened 2023-02-26
Add a LG-UART component

kquinsland:

This implements UART based control over LG brand TVs and Monitors.
Most LG screens have some RS232 based control input that can be used to set/poll certain things.
Unfortunately, the number of "things" that can be set/polled is different for each display/model/firmware version.

E.G.: some displays respond to "turn on/off" commands but do not respond to "are you on or off?" commands.

I haven't written any serious C++ in more than a decade so I was mostly following the architecture of the `bedjet` component; there is a `hub` that does all of the work and then a 'child' `switch` or `number` or `sensor` component is responsible for transforming the value of the parameter into the appropriate metaphor.

E.G.: the packet format for "are you on or off?" and "what is the current volume level" is the same so the `hub` handles most of the comms and just lets each child component know when an update has come in for the particular param. When it makes sense, the user can elect to have the reply represented by a `switch` or a `number` .

## Types of changes

- [ ] Other

I have a WIP branch that I will turn into a PR after the code here is reviewed... my C++ is **rusty** so I'll wait for a "this code is good enough" or a "this code needs to be re-factored to ..." before i put more time into the docs.

If you require a PR for the docs before this can be reviewed, let me know and i'll open the docs PR and update this post.

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

This is a direct copy/paste from the test file that I've been using.
It's the exact same hardware that I used here: https://github.com/kquinsland/lg-m43mu79-esp-home-bridge

Although the LG device that I am using / testing against has changed.

```yaml

# See: https://esphome.io/components/uart.html
uart:
  id: uart_01
  tx_pin: GPIO16
  rx_pin: GPIO17
  # LG monitor wants 9600/8N1... pretty standard :)
  baud_rate: 9600

lg_uart:
  id: lg01
  screen_number: 01
  uart_id: uart_01

switch:
  - name: "Power"
    platform: lg_uart
    icon: "mdi:power"
    device_class: "switch"
    lg_uart_id: lg01
    # ka for power
    command: "ka"
    update_interval: 5s
    # Assume that the screen is on at boot.

number:
  - name: "Volume level"
    icon: "mdi:volume-source"
    platform: lg_uart
    lg_uart_id: lg01
    # docs indicate kf for audio level
    command: "kf"
    # And that values ranges from 0x00 to 0x65
    decode_base: 16
    mode: slider
    # For testing; Like with the audio mute setting, this seems to have a ~ 10s delay on it.
    update_interval: 15s

sensor:
  - name: "active input (raw)"
    internal: true
    id: sens_lg_active_inpt_raw
    platform: lg_uart
    lg_uart_id: lg01
    # xb is for setting input
    command: "xb"
    # And that active inputs are numbers 00 through 99
    decode_base: 10
    # For testing; Like with the audio mute setting, this seems to have a ~ 10s delay on it.
    update_interval: 10s
```

## Checklist:


esphome/components/lg_uart/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+import esphome.codegen as cg
+import esphome.config_validation as cv
+from esphome.components import uart
+from esphome.const import CONF_UART_ID, CONF_ID
+
+CODEOWNERS = ["@kquinsland"]
+DEPENDENCIES = ["uart"]
+MULTI_CONF = False
```
jesserockz: Why `False`? (Which is the default btw and dont need to add it is this is the case...)

esphome/components/lg_uart/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Hub needs the ID of the UART and set ID
+CONF_SCREEN_NUMBER = "screen_number"
+
+# Clients
+CONF_LG_UART_ID = "lg_uart_id"
+CONF_LG_UART_CMD = "lg_uart_cmd"
+
+CONF_DECODE_BASE = "decode_base"
+CONF_DECODE_BASE_TYPES = [10, 16]
```
jesserockz: The `CONF_` prefix is for the config strings, not for values
```suggestion
DECODE_BASE_TYPES = [10, 16]
```

esphome/components/lg_uart/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+        # Docs indicate that ID of 0 is broadcast / all screens.
+        # By default, each comes from the factory set as 1.
+        cv.Optional(CONF_SCREEN_NUMBER, default=1): cv.int_range(min=0, max=99),
+    }
+)
+
+# Every component that uses us as a platform will need this
+LG_UART_CLIENT_SCHEMA = cv.Schema(
+    {
+        cv.Required(CONF_LG_UART_ID): cv.use_id(LGUartHub),
```
jesserockz: This allows automatic linking if there is only a single `lg_uart` component (The user doesnt have to add the id in yaml)
```suggestion
        cv.GenerateID(CONF_LG_UART_ID): cv.use_id(LGUartHub),
```

esphome/components/lg_uart/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+CODEOWNERS = ["@kquinsland"]
+DEPENDENCIES = ["uart"]
+MULTI_CONF = False
+
+# Hub needs the ID of the UART and set ID
+CONF_SCREEN_NUMBER = "screen_number"
+
+# Clients
+CONF_LG_UART_ID = "lg_uart_id"
+CONF_LG_UART_CMD = "lg_uart_cmd"
```
jesserockz: Please just use `CONF_COMMAND` from `const.py`. No need to prefix this with `lg_uart`

```suggestion
```

esphome/components/lg_uart/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+        # We need to know which uart device to use
+        cv.Required(CONF_UART_ID): cv.use_id(uart.CONF_ID),
+        # Docs indicate that ID of 0 is broadcast / all screens.
+        # By default, each comes from the factory set as 1.
+        cv.Optional(CONF_SCREEN_NUMBER, default=1): cv.int_range(min=0, max=99),
+    }
+)
+
+# Every component that uses us as a platform will need this
+LG_UART_CLIENT_SCHEMA = cv.Schema(
```
jesserockz, line 45: If all the children require `CONF_LG_UART_CMD` (to be renamed) why is that not just in this schema?

esphome/components/lg_uart/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+lg_uart_ns = cg.esphome_ns.namespace("lg_uart")
+LGUartHub = lg_uart_ns.class_("LGUartHub", cg.PollingComponent, uart.UARTDevice)
+
+# the root component
+CONFIG_SCHEMA = cv.COMPONENT_SCHEMA.extend(
+    {
+        cv.GenerateID(): cv.declare_id(LGUartHub),
+        # We need to know which uart device to use
+        cv.Required(CONF_UART_ID): cv.use_id(uart.CONF_ID),
```
jesserockz: This component should be extending `uart.uart_device_schema`, not declaring the uart id here.
```suggestion
```

esphome/components/lg_uart/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # ID of the UART component
+    await uart.register_uart_device(var, config)
+
+    # Screen ID / number
+    cg.add(var.set_screen_number(config[CONF_SCREEN_NUMBER]))
+    return var
+
+
+def validate_uart_cmd_len(config):
```
jesserockz: Once the CONF_COMMAND is moved into the shared schema, this should just be a 	cv.Length` validation on the config entry directly (combine string_strict and length with `cv.All`)

jesserockz (changes requested):
Couldnt get past the first file without a whole bunch of changes to be made so submitting those for now. =)

esphome/components/lg_uart/__init__.py
kquinsland: That was a remnant from testing / trying to figure out how that's used internally.

esphome/components/lg_uart/__init__.py
kquinsland: The command is distinct per feature. `ka` for power, `ke` for audio ... etc so it's a property of the switch/number ... etc and not of the 'hub'.

**edit** I mis=understood your question; that's left over from an early implementation.

kquinsland:
Gentle poke to @jesserockz. I've addressed the code duplication in the python and done some re-factor on the C++:

- `vec<uint8>` rather than char array for the packets
- some common code moved into the child class

PR #4493, opened 2023-02-27
Feat: add support for hex color in color component

### LINKED ISSUE #2138: Add support for hex in color component
I'd like to have an option to set the color of a color component using the hex code (`#rrggbb`) instead of red green and blue values.

It's just a simplification, when colors are copied from a palette, so they don't have to be converted manually. Also, it would make the yaml a bit shorter when there are many colors.

**Additional context**

I'd be happy to implement this myself in case a PR is welcome :)

dorianim:

Fixes https://github.com/esphome/feature-requests/issues/2138

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/feature-requests#2138

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
color:
  - id: test
    hex: '112233'
```

## Checklist:

  - [X] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs). https://github.com/esphome/esphome-docs/pull/2731


nielsnl68 (approved):
I do not know of I would at the `def hex_color()` in the *config_validation.py* but i like it.

dorianim:
@nielsnl68 I tried to stick to the scheme and express what it is. Maybe color_code would be better?
How would you name it?

nielsnl68:
> @nielsnl68 I tried to stick to the scheme and express what it is. Maybe color_code would be better? How would you name it?

I did not mean that the function name was wrong. I tried to say that i would not put the function in *config_validation.py* file myself. It is not wrong or anything.

dorianim:
Oh, I see. Yes, you're probably right about that. The color code validator shouldn't be needed anywhere else. I'll move it.

esphome/components/color/__init__.py
```diff
@@ -1,7 +1,10 @@
+import voluptuous as vol
+
 from esphome import config_validation as cv
 from esphome import codegen as cg
 from esphome.const import CONF_BLUE, CONF_GREEN, CONF_ID, CONF_RED, CONF_WHITE
 
+Invalid = vol.Invalid
```
oxan: Please drop both these changes, and use `cv.Invalid` directly instead.

esphome/components/color/__init__.py
```diff
@@ -10,6 +13,20 @@
 CONF_GREEN_INT = "green_int"
 CONF_BLUE_INT = "blue_int"
 CONF_WHITE_INT = "white_int"
+CONF_HEX = "hex"
+
+
+def hex_color(value):
+    if not value.startswith("#"):
+        raise Invalid("Color must start with #")
```
oxan: I'm not sure we want to do this: `#` also indicates a comment in YAML, so it'll require escaping, and since we're only accepting hex for this option, we don't need it to resolve ambiguity. 

@jesserockz what do you think?

esphome/components/color/__init__.py
```diff
@@ -10,6 +13,20 @@
...
+    value = value.lstrip("#")
+    if len(value) != 6:
+        raise Invalid("Color must have six digits")
+    try:
+        return (int(value[0:2], 16), int(value[2:3], 16), int(value[3:5], 16))
```
oxan:
```suggestion
        return (int(value[0:2], 16), int(value[2:4], 16), int(value[4:6], 16))
```

esphome/components/color/__init__.py
```diff
@@ -22,11 +39,12 @@
         cv.Exclusive(CONF_BLUE_INT, "blue"): cv.uint8_t,
         cv.Exclusive(CONF_WHITE, "white"): cv.percentage,
         cv.Exclusive(CONF_WHITE_INT, "white"): cv.uint8_t,
+        cv.Exclusive(CONF_HEX, "hex"): hex_color,
```
oxan: `Exclusive` ensures not more than one item from the same group (indicated by the second parameter) is present, so this doesn't make hex incompatible with red/green/blue.

oxan (changes requested):

esphome/components/color/__init__.py
dorianim: That's a good point.

esphome/components/color/__init__.py
dorianim: How should it be done instead?

oxan: I'm not quite sure tbh.

jesserockz:
I pushed a commit that should fix it. I don't think we can de-duplicate the `.extend(cv.COMPONENT_SCHEMA)` but that's fine.

dorianim:
Thanks! Just tried to do the same thing :)
image

jesserockz (approved):
Thanks. Please also add a PR to the docs repo targeting `next`

dorianim:
@jesserockz done! https://github.com/esphome/esphome-docs/pull/2731 :)

PR #4499, opened 2023-03-02
Update IRremoteESP8266 2.8.4 to support ESP32-C3

phoenixxie0:
Update IRremoteESP8266 2.8.4 to support ESP32-C3

Update IRremoteESP8266 2.8.4 to support ESP32-C3

## Types of changes

- [ x ] Bugfix (non-breaking change which fixes an issue)
- [ ] Other

**Related issue or feature (if applicable):** fixes <[link to issue](https://github.com/esphome/issues/issues/4208)>

## Test Environment

- [ x ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/heatpumpir/climate.py
+++ b/esphome/components/heatpumpir/climate.py
@@ -118,4 +118,4 @@ def to_code(config):
     cg.add_library("tonia/HeatpumpIR", "1.0.20")
 
     if CORE.is_esp8266 or CORE.is_esp32:
-        cg.add_library("crankyoldgit/IRremoteESP8266", "2.7.12")
+        cg.add_library("crankyoldgit/IRremoteESP8266", "2.8.4")
--- a/platformio.ini
+++ b/platformio.ini
@@ -92,7 +92,7 @@ lib_deps =
     ESP8266HTTPClient                     ; http_request (Arduino built-in)
     ESP8266mDNS                           ; mdns (Arduino built-in)
     DNSServer                             ; captive_portal (Arduino built-in)
-    crankyoldgit/IRremoteESP8266@2.7.12   ; heatpumpir
+    crankyoldgit/IRremoteESP8266@2.8.4   ; heatpumpir
 build_flags =
     ${common:arduino.build_flags}
     -Wno-nonnull-compare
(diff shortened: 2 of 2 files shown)


phoenixxie0:
I have do a PR to heatpumpir, to use IRremoteESP8266@2.8.4.

rob-deutsch:
Sorry for my delay here @phoenixxie0 .

We're running into a bit of a technicality here:

- **Current situation** - We use tonia/HeatpumpIR 1.0.20, which specifies crankyoldgit/IRremoteESP8266 ~2.7.12 in library.json. esphome uses crankyoldgit/IRremoteESP8266 2.7.12 which is of course consistent.
- **In the PR** - It uses tonia/HeatpumpIR 1.0.20, but it also uses crankyoldgit/IRremoteESP8266 2.8.4 which is *not* consistent with the ~2.7.12 specified.

There are two options to deal with this:
1. **Accept this PR:** I'm not a fan of this because second-guessing dependencies gives us maintenance tasks we don't want to take on. This package is provided as a nicety for esphome users, but we don't want to be maintaning it.
2. **Submit a PR to upstream:** tonia/HeatpumpIR could release a new version (e.g. 1.0.21) which has a dependancy of crankyoldgit/IRremoteESP8266 ^2.7.12 or ~2.8.4. This integration would then point to 1.0.21.

I have a strong preference for the 2nd option. This keeps the maintenance burden for the `heatpumpir` integration minimal. We should only rely on "official" versions of tonia/HeatpumpIR.

Also, if we went for the 1st option we'd be increasing the burden for the tonia/HeatpumpIR maintainer, because we'll push out code where we second guess the official releases. For example, I don't want to be adding issues to their project like "we get strange behaviour X when we use a mishmash of dependencies."

We should, of course, actively contribute and help the maintainer of tonia/HeatpumpIR, if we're benefiting from the improvements.

phoenixxie0:
@rob-deutsch 
I'd like to preference for the 2nd option. But the [tonia/HeatpumpIR](https://github.com/ToniA/arduino-heatpumpir) does not have any release after 1.0.9. I have already do a [PR](https://github.com/ToniA/arduino-heatpumpir/commit/0aef3cdcdd1e09b4b70eb54f7fa5e6cb69c92a01) for it, and it has accept. But He does not release a new version. I will try to do more about it.

PR #4500, opened 2023-03-02
SHT4x: change heater timing to minimize interference with measurements

### LINKED ISSUE #4263: SHT4x: index out of bounds + heater affects measurements
### The problem

There are 2 issues with SHT4x heater implementation currently:

1. One is simple to fix. From the code below you can see that `heater_command_` is actually a command not and index, so accessing `MEASURECOMMANDS[heater_command_]` would result in out of bounds error.

```cpp
static const uint8_t MEASURECOMMANDS[] = {0xFD, 0xF6, 0xE0};

void SHT4XComponent::start_heater_() {
  uint8_t cmd[] = {MEASURECOMMANDS[this->heater_command_]};

  ESP_LOGD(TAG, "Heater turning on");
  this->write(cmd, 1);
}

void SHT4XComponent::setup() {
  ESP_LOGCONFIG(TAG, "Setting up sht4x...");

  if (this->duty_cycle_ > 0.0) {
    uint32_t heater_interval = (uint32_t)(this->heater_time_ / this->duty_cycle_);
    ESP_LOGD(TAG, "Heater interval: %i", heater_interval);

    if (this->heater_power_ == SHT4X_HEATERPOWER_HIGH) {
      if (this->heater_time_ == SHT4X_HEATERTIME_LONG) {
        this->heater_command_ = 0x39;
      } else {
        this->heater_command_ = 0x32;
      }
    } else if (this->heater_power_ == SHT4X_HEATERPOWER_MED) {
      if (this->heater_time_ == SHT4X_HEATERTIME_LONG) {
        this->heater_command_ = 0x2F;
      } else {
        this->heater_command_ = 0x24;
      }
    } else {
      if (this->heater_time_ == SHT4X_HEATERTIME_LONG) {
        this->heater_command_ = 0x1E;
      } else {
        this->heater_command_ = 0x15;
      }
    }
    ESP_LOGD(TAG, "Heater command: %x", this->heater_command_);

    this->set_interval(heater_interval, std::bind(&SHT4XComponent::start_heater_, this));
  }
}
```

2. Another one is more difficult. As heater naturally causes temperature increase, if update_interval and heater_interval are not synced to each other in a way that guarantees maximal distance between heater turning off and next measurement, then temperature measurements will be affected. You can see it in the plot, when I turned on heater at around 22:00. I used `heater_max_duty: 0.01` and `heater_time: Short (110ms)`, which resulted in turning heater on every 11 seconds for 110ms. As update_interval (e.g. 60s) is independent of heater_interval, some of them will inevitable co-occur in time. 

image

TBH I'm not sure what would be the best way how to solve the latter issue in general case as it depends on update_interval/max_duty/heater_time/heater_power.

2023.2.0

pip

_No response_

ESP32

### Board

_No response_

sht4x

### Example YAML snippet

```yaml
sensor:
  - platform: sht4x
    temperature:
      name: "SHT41 Temperature"
      id: sht41_temp
    humidity:
      name: "SHT41 Humidity"
      id: sht41_humid
    precision: High
    heater_max_duty: 0.01
    heater_power: High
    heater_time: Short
```

_No response_

### Additional information

_No response_

stas-sl:

As described in https://github.com/esphome/issues/issues/4263, if heater is used it could affect temperature measurements. In this PR I'm implementing alternative approach by running heater at the end of `update()` method, after measurements have been already taken. Thus we'll have max possible cool off time before next measurement. Instead of `heater_max_duty` config option, I'm introducing `heater_period` which means how often heater should be run: 0 - don't use heater, 1 - run on every update, 2 - every second update and so on. 

If someone has a better solution let me know.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4263

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** If PR will be approved, I'll update docs.

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
sensor:
  - platform: sht4x
    id: sht41
    temperature:
      name: "SHT4x Temperature"
    humidity:
      name: "SHT4x Humidity"
    precision: High
    heater_power: High
    heater_time: Long
    heater_period: 1
```

## Checklist:


pilotak:
@stas-sl could you please resolve the conflicts so it can be merged?

stas-sl:
@pilotak done!

pilotak:
Perfect, hopefully the admins will pick this up shortly as it works much better with your fix

stas-sl:
Interesting, so you have already tested it? I wonder what effect did you see?

stas-sl:
I'm asking because with previous implementation heater was never turned on, because of a mistake in implementation. But after fixing it, so that heater worked as expected, I noticed that it interferes with measurements. Therefore I decided to change approach as I described here and in the referenced issue. So ideally, it should not affect measurements, so it would be not very easy to notice if there is any effect.

pilotak:
with you patch  https://github.com/esphome/issues/issues/4263#issuecomment-1450863394 it dropped the temperature by 1 degree which was the real temperature. So i believe this patch is now included in this PR. Anyway i will test this specific PR hopefully in the evening and will let you know

stas-sl:
Yes, this PR does pretty much the same as that piece of code in the comment.

However my understanding is that heater is supposed to improve *humidity* sensor measurements in *some* environments as it is described in these documents: [Using the Integrated Heater of SHT4x for Heat-Assisted Polymer
Decontamination](https://sensirion.com/media/documents/FEE9F039/62459F54/Application_Note_Heater_Decontamination_SHT4xX.pdf) and [Using the Integrated Heater of SHT4x in High-Humidity
Environments](https://sensirion.com/media/documents/A88858C9/629626D4/Application_Note_Creep_Mitigation_SHT4x.pdf). Particularly, high humidity and VOC contaminated environments. But as currently I'm using it indoors, where in winter I have <40% humidity, and I believe the air is not that much contaminated, I don't see significant difference. 

Considering temp measurements, I do see slight increase +0.3-0.4°, when heater is on (high power, 1s duration, every 60 sec). It is not as bad as I posted in the issue, but there is still a little bit of interference, which probably could be avoided if using less power or extending update_interval.

So, TBH, after implementing it and testing, I'm not currently using it for indoor measurements.

PR #4509, opened 2023-03-04
Reserve vector memory for core components before usage.

Fabian-Schmidt:

Reserve memory for all vectors in core components (application & scheduler) before first usage. This avoids vector resize and lowers memory fragmentation.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


oxan:
I don't mind the idea of reserving memory for the vectors in advance, but passing the expected size as a gazillion parameters is quite ugly imo. Additionally, doing it in `pre_setup()` causes the memory to be reserved prior to the OTA safe mode check, which makes it unrecoverable if too much memory is reserved here (okay, it's just for a bunch of pointers, so it's quite unlikely we'll run out of memory here... but still, we should do as little as possible prior to the safe mode check).

How about adding `reserve_binary_sensors_space(size_t space)` methods (or something like that; I'm not sold on the name), and inserting calls to them after the OTA safe mode check?

Fabian-Schmidt:
> I don't mind the idea of reserving memory for the vectors in advance, but passing the expected size as a gazillion parameters is quite ugly imo. Additionally, doing it in `pre_setup()` causes the memory to be reserved prior to the OTA safe mode check, which makes it unrecoverable if too much memory is reserved here (okay, it's just for a bunch of pointers, so it's quite unlikely we'll run out of memory here... but still, we should do as little as possible prior to the safe mode check).
> 
> How about adding `reserve_binary_sensors_space(size_t space)` methods (or something like that; I'm not sold on the name), and inserting calls to them after the OTA safe mode check?

I agree it is ugly to have this many parameter.
To keep it all within the one `to_code` method of the core I cannot see how to easily move a method call to an after ota point in time.

oxan:
Having two `to_code` coroutines with different priorities is fine with me.

Fabian-Schmidt:
> Having two `to_code` coroutines with different priorities is fine with me.

How is this possible? Each component can only have one `to_code` method.

oxan:
> How is this possible? Each component can only have one `to_code` method.

You can add another job with a different priority to the core in `to_code`, similar to what's done with `_add_automations`.

oxan:
Though I've submitted #4529 which should do allow that in a nicer way.

PR #4511, opened 2023-03-05
climate: add on_control callbacks

### LINKED ISSUE #2136: Climate components: separate callback on configuration changes?

I would like to have a callback for (climate) components that changes only when control parameters, not world state, changes.  `on_config` or such.

I am currently writing (yet another) thermostat using esphome.  I've got a character LCD with controllable backlight and a local touch sensor, and I would like the following backlight policy: turn on either when HA changes some part of the Climate configuration or the user touches a button, and turn off 10 seconds after the last reason to be on.

Almost all of that is possible right now, but attempting to use the `on_state` trigger to watch for HA changes *also* catches every single sample from the temperature sensor, which means the backlight is constantly flashing. :(

**Additional context**

nwf:

This lets downstream components respond to climate configuration changes, which take place through ClimateCall objects, without also being notified every time the state changes, which happens every time the input sensor announces a new value.

FIXES https://github.com/esphome/feature-requests/issues/2136

## Types of changes

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2720

## Test Environment

- [x] ESP8266

```yaml
# Example config.yaml

```

## Checklist:

I have not tested it on a real device, but it does compile and I think I have plumbed it right.  I'll get there.


jesserockz (changes requested):
Changes look good, but I don't like `on_message`. Perhaps `on_config` or `on_control`?

nwf:
OK.  I wasn't sure how you'd feel about expanding the set of trigger names.  I had originally suggested `on_config` but I think I like `on_control` better; thank you for the suggestion.  I'll go flip everything to that.

oxan (approved):

jesserockz (approved):

PR #4518, opened 2023-03-05
Support for custom mac address

jfparis:
This commit adds support for custom mac addresses previously implanted with "espefuse.py burn_custom_mac"

## What does this implement/fix?

ESP32 allows user to burn a custom MAC address via espefuse.py
https://docs.espressif.com/projects/esptool/en/latest/esp32/espefuse/burn-custom-mac-cmd.html
This is especially useful when a batch of chips comes out with identical MAC or a blank (00:..:00) MAC
This PR support such MAC

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:


esphome/core/helpers.cpp
```diff
@@ -443,7 +443,8 @@ void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parame
   // without doing the CRC check.
   esp_efuse_read_field_blob(ESP_EFUSE_MAC_FACTORY, mac, 48);
 #else
-  esp_efuse_mac_get_default(mac);
+  if (esp_efuse_mac_get_custom(mac) == ESP_OK) {}
+  else esp_efuse_mac_get_default(mac);
```
ssieb:
```suggestion
  if (esp_efuse_mac_get_custom(mac) != ESP_OK)
    esp_efuse_mac_get_default(mac);
```

jfparis: Thank you. Fixed that

jfparis:
Hello. Is there anything I can do / change to help this into the main branch?

esphome/core/helpers.cpp
```diff
@@ -443,7 +443,8 @@ void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parame
   // without doing the CRC check.
   esp_efuse_read_field_blob(ESP_EFUSE_MAC_FACTORY, mac, 48);
 #else
-  esp_efuse_mac_get_default(mac);
+  if (esp_efuse_mac_get_custom(mac) != ESP_OK)
```
mmakaay: Wouldn't it be more useful to  implement this as follows?

```c++
#if defined(USE_ESP32)
  if (esp_efuse_mac_get_custom(mac) != ESP_OK) {
#if defined(USE_ESP32_IGNORE_EFUSE_MAC_CRC)
    // On some devices, the MAC address that is burnt into EFuse does not
    // match the CRC that goes along with it. For those devices, this
    // work-around reads and uses the MAC address as-is from EFuse,
    // without doing the CRC check.
    esp_efuse_read_field_blob(ESP_EFUSE_MAC_FACTORY, mac, 48);
#else
    esp_efuse_mac_get_default(mac);
#endif
  }
#elif defined(USE_ESP8266)
```

That would bring custom MAC support to all devices, not only those for which `USE_ESP32_IGNORE_EFUSE_MAC_CRC` is not defined.

jfparis: This makes sense. Thank you for reviewing

PR #4521, opened 2023-03-06
Upgraded Haier climate component implementation

paveldn:

This is a new implementation of the Haier AC component based on my own development: https://github.com/paveldn/haier-esphome
Current Haier component implementation only works with one kind of AC that is using the SmartAir2 application, dependent on the Arduino framework. My implementation supports both protocol versions and is dependent only on UART, sensor and climate components of ESP home so can be used for any platform and board that support these components. Also, my implementation in communication protocol fixed a bug with encoding FF bytes with FF 55. This bug is present in the current implementation.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2809

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

uart:
  baud_rate: 9600
  tx_pin: 19
  rx_pin: 18
  id: uart_id

climate:
  - platform: haier
    id: haier_ac_climate
    protocol: hOn
    name: Haier AC 
    uart_id: uart_id
    wifi_signal: true
    beeper: true
    outdoor_temperature:
      name: Haier AC outdoor temperature
    visual:
      min_temperature: 16 °C
      max_temperature: 30 °C
      temperature_step: 1 °C
    supported_modes:
    - 'OFF'
    - AUTO
    - COOL
    - HEAT
    - DRY
    - FAN_ONLY
    supported_swing_modes:
    - 'OFF'
    - VERTICAL
    - HORIZONTAL
    - BOTH

switch:
  - platform: template
    id: haier_ac_display_switch
    name: Haier AC display
    icon: mdi:led-on
    entity_category: config
    restore_state: yes
    lambda: |-
      return id(haier_ac_climate).get_display_state();
    turn_on_action:
      climate.haier.display_on: haier_ac_climate
    turn_off_action:
      climate.haier.display_off: haier_ac_climate
  - platform: template
    id: haier_ac_beeper_switch
    entity_category: config
    name: Haier AC beeper
    icon: mdi:volume-high
    restore_state: yes
    lambda: |-
      return id(haier_ac_climate).get_beeper_state();
    turn_on_action:
      climate.haier.beeper_on: haier_ac_climate
    turn_off_action:
      climate.haier.beeper_off: haier_ac_climate

select:
  - platform: template
    id: haier_ac_vertical_direction
    name: Haier AC airflow vertical
    entity_category: config
    icon: mdi:arrow-expand-vertical
    optimistic: true
    restore_value: true
    initial_option: Center 
    options:
      - Up
      - Center
      - Down
    on_value:
      then:
        - climate.haier.set_vertical_airflow:
           id: haier_ac_climate
           vertical_airflow: !lambda |-
             return esphome::haier::AirflowVerticalDirection(i);
  - platform: template
    id: haier_ac_horizontal_direction
    name: Haier AC airflow horizontal
    entity_category: config
    icon: mdi:arrow-expand-horizontal
    optimistic: true
    restore_value: true
    initial_option: Center 
    options:
      - Left
      - Center
      - Right
    on_value:
      then:
        - climate.haier.set_horizontal_airflow:
           id: haier_ac_climate
           horizontal_airflow: !lambda |-
             return esphome::haier::AirflowHorizontalDirection(i);

```

## Checklist:


jesserockz:
@Yarikx I would like your opinion and testing on this please as you wrote the original which I only just merged to release next week in 2023.3.0.

So if possible I would like to have this in the beta before then

hcaldicott:
I would love to see this version implemented into ESP Home as it is far more feature complete and supports both Haier protocols. @paveldn's code is a complete re-implementation of the reverse engineering work documented on the HASS forums that @Yarikx merged into the ESPHome codebase. The current code that @Yarikx implemented won't support my device, but with @paveldn's code merged I am happy to assist with beta testing.

paveldn:
@jesserockz  Is there anything I can do to help with reviewing and testing my implementation?

hcaldicott:
@jesserockz @Yarikx any way we can progress this? Keen to see @paveldn's updates merged into ESPHome.

Yarikx:
Great job, @paveldn on this component. 
Your version is definitely much more robust than the current implementation and I would be happy to update the component. 

As it replaces the whole component with totally new code, I don't think my review is essential as a code owner (also my C++ is not my main language). Though I suggest pinging somebody from the core team. You can try reaching them in esphome discord directly.

One thing I would still suggest changing is the configuration format and making sure it will be backward compatible with the current version. As the current Haier component is already in master, any change to the configuration will be breaking change, for instance, a new required field `protocol` will need to be added.

paveldn:
Thanks, @Yarikx. I already made the smartair2 default protocol for backward compatibility as you suggested and tested it with reference config from help - it worked OK. 
Probably I will first wait for @jesserockz  to give feedback about your comment. Contacting the core team will be my plan "B" 😄

jesserockz (commented):
I started to write a review with a bunch of comments, then realised it would just be faster to modify the code locally as they were all connected.

Please verify my changes which should simplify the config validation and code generation functions.

Also need to remove the extra constructors that are not being used. I will do another review of the rest soon. 

Also please remove the component from all test yaml files except 1 (You can choose which one).

paveldn:
Hi @jesserockz,

First of all thanks for reviewing my code and for your suggestions. I implemented some small adjustments on top of your changes and made it buildable again by fixing constructors for cpp files. Thanks a lot, your version of Python code is simpler and more readable. I also removed extra tests as you suggested, and tested the code with my AC and it works OK. Thanks again!

esphome/components/haier/climate.py
```diff
@@ -1,43 +1,364 @@
...
+        _LOGGER.info("Detected log level for Haier protocol: %s", _level)
+        if _level not in logger.LOG_LEVEL_SEVERITY:
+            raise cv.Invalid("Unknown log level for Haier protocol")
+        _severity = logger.LOG_LEVEL_SEVERITY.index(_level)
+        cg.add_build_flag(f"-DHAIER_LOG_LEVEL={_severity}")
+    else:
+        _LOGGER.info(
+            "No logger component found, logging for Haier protocol is disabled"
+        )
+        cg.add_build_flag("-DHAIER_LOG_LEVEL=0")
```
jesserockz, line 330: What is the point of this code and setting the level to match ESPHome level? If you log at the right levels in code `ESP_LOGD/V/I` then this should not matter

esphome/components/haier/climate.py
```diff
@@ -1,43 +1,364 @@
...
 async def to_code(config):
+    cg.add(haier_ns.init_haier_protocol_logging())
     var = cg.new_Pvariable(config[CONF_ID])
     await cg.register_component(var, config)
-    await climate.register_climate(var, config)
     await uart.register_uart_device(var, config)
+    await climate.register_climate(var, config)
+
+    if (CONF_WIFI_SIGNAL in config) and (config[CONF_WIFI_SIGNAL]):
+        cg.add_build_flag("-DHAIER_REPORT_WIFI_SIGNAL")
```
jesserockz: We generally dont set build flags like this. If you want to exclude some code, you can check `USE_WIFI` within this component

esphome/components/haier/haier_base.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include "esphome/components/climate/climate.h"
+#include "esphome/components/uart/uart.h"
+#include "haier_base.h"
+
+using namespace esphome::climate;
+using namespace esphome::uart;
+
+#ifndef ESPHOME_LOG_LEVEL
+#warning "No ESPHOME_LOG_LEVEL defined!"
+#endif
```
jesserockz: A component should not throw a random warning about this.

paveldn:
> What is the point of this code and setting the level to match ESPHome level? If you log at the right levels in code `ESP_LOGD/V/I` then this should not matter

I am using my own library to handle the transport level of the Haier protocol. This library is not dependent on ESPhome. For example, I also use it in Windows and Linux for tests and protocol simulation. I have a special mechanism to pass logs from the library to ESPhome but if in ESPHome this logs level is not active in library strings anyway prepared (also buffer allocated etc.) to avoid this I am using this code. It is just the most efficient way to avoid wasting resources.

> We generally dont set build flags like this. If you want to exclude some code, you can check `USE_WIFI` within this component

I am using this to avoid dependency on the WiFi component in case ESP with ethernet is used. Didn't know about `USE_WIFI` will use it. Thanks.

> A component should not throw a random warning about this.

OK Will remove them.

esphome/components/haier/haier_base.cpp
```diff
@@ -0,0 +1,311 @@
...
+#if (HAIER_LOG_LEVEL > 4)
+    ESP_LOGV(TAG, "Phase transition: %s => %s", phase_to_string_(this->protocol_phase_), phase_to_string_(phase));
+#else
+    ESP_LOGV(TAG, "Phase transition: %d => %d", (int) this->protocol_phase_, (int) phase);
+#endif
+    this->protocol_phase_ = phase;
+  }
+}
+
+bool HaierClimateBase::check_timout_(std::chrono::steady_clock::time_point now,
```
jesserockz: typo?
```suggestion
bool HaierClimateBase::check_timeout_(std::chrono::steady_clock::time_point now,
```

platformio.ini
```diff
@@ -65,6 +65,7 @@ lib_deps =
     rweather/Crypto@0.4.0                                 ; dsmr
     dudanov/MideaUART@1.1.8                               ; midea
     tonia/HeatpumpIR@1.0.20                               ; heatpumpir
+    pavlodn/HaierProtocol@0.9.18                          ; haier
```
jesserockz: This should only be added once, if it can be used on any platform it should go above in the `[common]` `lib_deps`

jesserockz (changes requested):

platformio.ini
paveldn: Done. Thanks

esphome/components/haier/haier_base.cpp
```diff
@@ -0,0 +1,311 @@
...
+#include "esphome/components/uart/uart.h"
+#include "haier_base.h"
+
+using namespace esphome::climate;
+using namespace esphome::uart;
+
+namespace esphome {
+namespace haier {
+
+const char TAG[] = "haier.climate";
```
jesserockz:
```suggestion
static const char *const TAG = "haier.climate";
```

esphome/components/haier/hon_climate.cpp
```diff
@@ -0,0 +1,857 @@
...
+#include "hon_climate.h"
+#include "hon_packet.h"
+
+using namespace esphome::climate;
+using namespace esphome::uart;
+
+namespace esphome {
+namespace haier {
+
+const char TAG[] = "haier.climate";
```
jesserockz:
```suggestion
static const char *const TAG = "haier.climate";
```

esphome/components/haier/smartair2_climate.cpp
```diff
@@ -0,0 +1,457 @@
...
+#include "smartair2_climate.h"
+#include "smartair2_packet.h"
+
+using namespace esphome::climate;
+using namespace esphome::uart;
+
+namespace esphome {
+namespace haier {
+
+const char TAG[] = "haier.climate";
```
jesserockz:
```suggestion
static const char *const TAG = "haier.climate";
```

jesserockz (changes requested):

jesserockz (approved):

PR #4527, opened 2023-03-07
Add support for new clones of mpu6050 responding with 0x70 address

jakehdk:

Adds support for new clones of mpu6050 responding with 0x70 address.
I personally have 5 samples of these, so there is a good chance that others also will receive these going forward (Aliexpress) 

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

## Checklist:


oxan:
While this PR is fine (aside from the formatting), I'm wondering whether there's any good reason why we should hard fail instead of only warn when the ID register doesn't match?

jesserockz:
> While this PR is fine (aside from the formatting), I'm wondering whether there's any good reason why we should hard fail instead of only warn when the ID register doesn't match?

Nobody will even see the warning...haha
Its good to know which ones do work with this code I guess...

But I do see your point there

jakehdk:
Do you want me to do the suggested format changes before merging?

from:
if (!this->read_byte(MPU6050_REGISTER_WHO_AM_I, &who_am_i) || (who_am_i != 0x68 && who_am_i != 0x70 && who_am_i != 0x98)) {

to:
 if (!this->read_byte(MPU6050_REGISTER_WHO_AM_I, &who_am_i) ||
     (who_am_i != 0x68 && who_am_i != 0x70 && who_am_i != 0x98)) {

I do not really think it looks nicer/cleaner, but I am not a "real developer" 😄

oxan:
> Do you want me to do the suggested format changes before merging?

Yes, please. I can see why you think it's not necessarily better; but having a consistent style and autoformatter is worth more than having the best style (which is usually a matter of personal taste anyway), and we have to draw the limit on line length somewhere.

jakehdk:
wow - looks like I have made a mess trying to change author name.
Sorry - I will close this a try again.

PR #4530, opened 2023-03-07
Upgrade clang-tidy to v13

oxan:

Upgrade clang-tidy to v13, and fix or suppress a few new warnings. 

LLVM upstream is already at v15, but that version isn't available in the repositories used for our containers yet. There'll be a new Debian release in a few months, so don't bother with working around that.

Draft PR for now as I haven't actually tested the changes yet, mostly interested to see what CI says.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** n/a

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/binary_sensor/binary_sensor.h
```diff
@@ -12,12 +12,14 @@ namespace esphome {
...
-      ESP_LOGCONFIG(TAG, "%s  Device Class: '%s'", prefix, (obj)->get_device_class().c_str()); \
+  do { \
+    if ((obj) != nullptr) { \
+      ESP_LOGCONFIG(TAG, "%s%s '%s'", prefix, LOG_STR_LITERAL(type), (obj)->get_name().c_str()); \
+      if (!(obj)->get_device_class().empty()) { \
+        ESP_LOGCONFIG(TAG, "%s  Device Class: '%s'", prefix, (obj)->get_device_class().c_str()); \
+      } \
     } \
-  }
+  } while (0)
```
jesserockz, line 22: Not sure I understand the purpose of wrapping in a do while loop that will only execute once...

jesserockz: Ah nevermind. I see why this can be useful if used in a non-bracketed conditional

jesserockz (approved):
Looks fine to me from a review standpoint

esphome/components/binary_sensor/binary_sensor.h
oxan: Yeah, it's a common trick to make macros behave like functions in all situations. I added them here because the new clang-tidy warned about it.

oxan:
So the reason this is still a draft is that while wrapping the macros in do/while, I realized that I actually quite disliked them being macros and that they probably should be functions instead (`dump_config()` isn't called often enough for the function call overhead to matter, and it probably saves code space anyway). I started working on that but life got in the way. You can merge this as-if if you want, otherwise I'll submit that PR sometime and then drop the conflicting changes from here once it's merged.

PR #4535, opened 2023-03-08
Upgrade clang-format to v13

oxan:

Upgrade clang-format to v13, and apply some changed formatting (which mostly seems to be actually honoring the `SpaceAfterCStyleCast` setting).

@nielsnl68 This should solve the problem you mentioned [here](https://github.com/esphome/esphome/discussions/4436#discussioncomment-5159284), it was caused by a version mismatch between the clang-format your VScode used and what we used in CI.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


nielsnl68:
> @nielsnl68 This should solve the problem you mentioned [here](https://github.com/esphome/esphome/discussions/4436#discussioncomment-5159284), it was caused by a version mismatch between the clang-format your VScode used and what we used in CI.

Perfect, i will comeback when i found other differences, when i see them.

nielsnl68:
BTW. Not sure of i should put it here, There is on `clang` requirement that i really dislike and that is the need to have the following:
```cPP
    if (1 == 1)
      return;
```
To me, this could be allow to introduce bugs when someone add a new line between the `if` statement and the `return;`  or could be wrongly read when de code writer misses the decency to do proper indents etc. 
there for imho the above statement should always be:
```cpp
    if (1 == 1)  return;
```
or 
```cPP
    if (1 == 1) {
      return;
    }
```
But that is just me. Not sure why clang does want this to have the `return;` on the new line.

oxan:
> BTW. Not sure of i should put it here, There is on `clang` requirement that i really dislike and that is the need to have the following:

To be honest, I prefer the way it currently is. Putting it all on one line makes it too easy to miss the return, while the variant with the braces is overly verbose. In my experience it doesn't happen often that you insert something in between and create a bug; clang-format also automatically adjusts the indents so it should be evident that the return is no longer conditional.

That said, you're free to always add braces in your own code, I don't think any of the clang tools remove them.

jesserockz:
While I like the single line return personally, for a project this size and so many people looking at it, it makes sense to drop it to the next line.

Haha the `SpaceAfterCStyleCast` is something I always have to manually adjust locally too

PR #4549, opened 2023-03-12
Added response for Tuya RSSI command

catalin2402:
Implemented wifi rssi response for tuya MCU

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/network/util.cpp
+++ b/esphome/components/network/util.cpp
@@ -49,15 +49,5 @@ std::string get_use_address() {
 #endif
   return "";
 }
-
-uint8_t get_wifi_rssi() {
-#ifdef USE_WIFI
-  if (wifi::global_wifi_component != nullptr)
-    return wifi::global_wifi_component->wifi_rssi();
-#endif
-
-  return 0;
-}
-
 }  // namespace network
 }  // namespace esphome
--- a/esphome/components/network/util.cpp
+++ b/esphome/components/network/util.cpp
@@ -49,5 +49,6 @@ std::string get_use_address() {
 #endif
   return "";
 }
+
 }  // namespace network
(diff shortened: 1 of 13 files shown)


esphome/components/network/util.cpp
```diff
@@ -50,5 +50,14 @@ std::string get_use_address() {
...
 
+uint8_t get_wifi_rssi() {
+#ifdef USE_WIFI
+  if (wifi::global_wifi_component != nullptr)
+    return wifi::global_wifi_component->wifi_rssi();
+#endif
+
+  return 0;
+}
+
```
jesserockz: Not sure I like the wifi rssi being added in the network utils when it will never apply to ethernet...

Probably best to just have this code directly inside the `case TuyaCommandType::WIFI_RSSI`

jesserockz (changes requested):

esphome/components/network/util.cpp
catalin2402: Got it, thanks for the review, I will move the code from network to tuya component :)
In my original thought, I wanted to move the code from the WiFiSignalSensor sensor to use the same network utils, and maybe other components who would need to access the same method can use the utils instead

jesserockz: Well technically the code is already common for the different wifi backends with `wifi::global_wifi_component->wifi_rssi()` and adding another wrapper around that is just extra code. I dont think it needs to be moved anywhere else.

jesserockz (approved):

PR #4569, opened 2023-03-15
EntityBase: Move ObjectId to Flash 

Fabian-Schmidt:

This is a memory improvement.
The `EntityBase` field `object_id_` can be pre-calculated in python and then stay in flash instead of being copied to memory as `std::string`. The calculation in python follows the same rules as the c++ code to ensure the resulting object id is the same. In python it is much easier to also validate two ids are unique (improvement idea).

Each `std::string` instance (even empty once) occupy at least 24 bytes (15 byte capacity + length + flags) of memory. By using char pointer instead of std::string they stay in flash.

The `object_id_` field has zero direct uses. Therefore no benefit of using an `std::string` here. In case any existing code is referencing this field it will break as the field was renamed.

In my projects I get easily more than 25 instances of `EntityBase`. In my testing I get ~1kb of more free memory by performing this change.

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (approved):

jesserockz:
Lol, moving them into flash made the flash size too large for test3. 
I actually need to split that file up a bit, not much you can do in this PR

Fabian-Schmidt:
While re checking all EntityBase client classes I found that Sprinkler was not using the common method `setup_entity` and therefore must be adjusted or else it would have no Object Id being set.

kbx81 (approved):
I _think_ the sprinkler change looks good, but @jesserockz knows Python better than me so it's probably best if he were to double-check. 😄

Looks like you also have a merge conflict to resolve. 😇

jesserockz:
This is because Sprinkler is not an Entity. An Entity is only something that can be represented in HA - Sensor/Switch/Cover/Climate/etc

jesserockz:
Oh I see Sprinkler extends EntityBase... 
Is there a reason for this @kbx81 as it doesnt really make sense unless it can be sent directly in the API to HA...

kbx81:
@jesserockz a reason...is copy/paste a reason? 😆

This was the first time I created a new component, so I'm not surprised to come across architectural errors such as this. I agree, knowing what I know now, it doesn't really make sense to inherit from EntityBase since sprinkler itself is more or less a controller that automates other entities -- it doesn't represent an entity itself.

Totally fine by me if we trim some fat here. 😇

Fabian-Schmidt:
Thanks @kbx81 . I removed the `EntityBase` as base class for sprinkler and moved the name locally. Looks like more was not used by Sprinkler from the base class.

jesserockz:
It might pay to move the sprinkler changes to another PR to merge first as there are more required like adding CONF_NAME to the schema

jesserockz:
https://github.com/esphome/esphome/pull/4606

jesserockz (approved):
Thanks for updating after my `name` changes

PR #4574, opened 2023-03-17
Wrap ipv6 code a bit more

jesserockz:

IPv6 support was being compiled in even if `enable_ipv6: false` was set due to `CONFIG_LWIP_IPV6` being set to true/y/1 by default. This changes the sdk option to set to false correctly.

In doing this I found some IPv6 code was not being wrapped with `#if LWIP_IPV6` correctly.

I have also moved the `htons` port conversion to be inside the socket addr function to make the "api" simpler and just accept the port number as is.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/socket/socket.h
+++ b/esphome/components/socket/socket.h
@@ -44,7 +44,10 @@ std::unique_ptr<Socket> socket(int domain, int type, int protocol);
 /// Create a socket in the newest available IP domain (IPv6 or IPv4) of the given type and protocol.
 std::unique_ptr<Socket> socket_ip(int type, int protocol);
 
-/// Set a sockaddr to the any address for the IP version used by socket_ip().
+/// Set a sockaddr to the specified address and port for the IP version used by socket_ip().
+socklen_t set_sockaddr(struct sockaddr *addr, socklen_t addrlen, const char *ip_address, uint16_t port);
+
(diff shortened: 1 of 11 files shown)


esphome/components/socket/socket.h
```diff
@@ -44,7 +44,10 @@ std::unique_ptr<Socket> socket(int domain, int type, int protocol);
 /// Create a socket in the newest available IP domain (IPv6 or IPv4) of the given type and protocol.
 std::unique_ptr<Socket> socket_ip(int type, int protocol);
 
-/// Set a sockaddr to the any address for the IP version used by socket_ip().
+/// Set a sockaddr to the specified address and port for the IP version used by socket_ip().
+socklen_t set_sockaddr(struct sockaddr *addr, socklen_t addrlen, const char *ip_address, uint16_t port);
```
oxan, line 48: @jesserockz I don't think this function (at least in its current implementation) makes any sense. The `ip_address` parameter should be an IPv4 or IPv6 address, depending on the socket type returned by `socket_ip()`, but there's no way for the user to know which type that is. Usually if you have a string containing an IP address, that's the starting point and determines whether you want to create an IPv4 or IPv6 socket, not the other way around.

PR #4577, opened 2023-03-17
Introduce support for platformio `build_cache_dir`

### LINKED ISSUE #2171: Reuse compiled source between nodes
I've been managing a few identical devices (apart from their identity) for a little while now, and I've been following along when it's been compiling and I see the same code being compiled again and again. This makes sense of course, because it's simpler to see each device as a completely separate unit.

I imagine that in my and many other instances though, at least parts of these files will be identical. It is my impression that the device config is used to substitute values into the source (`cpp`) files, which is then compiled. This could be f.ex. the wifi configuration, which would (at minimum) contain an SSID to connect to. This SSID is the same though, on all my devices. So I'm thinking it should be possible to reuse the compiled output (the `.o` file).

If I'm right in assuming the `.cpp` files are modified with the values they need (from the device config), it should be possible to do something like:

1. Substitute in the values needed in the `.cpp` file
2. Hash the `.cpp` file, f.ex. using `md5` or something less intensive like [`xxhash`](https://github.com/Cyan4973/xxHash)
3. Check if a file exists at `/somedir/output/HASH.o`
4. If the file exists, copy it in to the output, where the output would normally go. We know this output is identical, to what we would have produced, as the source input is identical
5. If the file does not exist, compile the `.cpp`, and copy the output to `/somedir/output/HASH.o`
6. Proceed as normal, perfom linking as normal, build a firmware as normal - this should work, as all compiled output will still be in the place it usually is.

The temporary/cache location can be cleared at will - if it's empty, the sources are compiled again and placed there. Likewise, if there are access timestamps, you can remove the least recently used files in order to keep a rolling set of used files.

N/A

**Additional context**
N/A

LordMike:
Adds option support for setting the `build_cache_dir` in platformio. This setting helps speed up builds of the same device currently, as shown in [this comment](https://github.com/esphome/feature-requests/issues/2171#issuecomment-1473686421). In the future, hopefully [this issue](https://github.com/platformio/platformio-core/issues/4574) is accepted and resolved, it could also help speed up builds of identical devices, by sharing the build output between them.

I'm by no means a python developer, so I've been following the format of the `build_path` config value. I don't really know where these values come from, but I hope the `esphome addon` can set the `build_cache_path` to a value like `/esphome_cache/`.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2171

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
# I am unsure how this will look. 
build_cache_path: /esphome_cache/
```

## Checklist:


LordMike:
Note: An entirely different approach to this, is for the `esphome addon` to set the environment variable `PLATFORMIO_BUILD_CACHE_DIR`

This has less impact here, but won't help users not using the addon.

jesserockz:
I just tested this locally and it had **zero** effect on a duplicated yaml file with only the name changed

LordMike:
In one of the issues in Platform IO I found that the full paths of the compiled items are actually used to calculate the hashes used for caching. So as it is now, pio will not be able to use the cache between environments. 

But. What this pr can do is two things: in the future when pio changes to support this, esphome is ready, and secondly it ensures the cache can be used within one device. 

So if an update (as seen in the hass add on) comes along, it is my experience that I will often have to recompile everything, but with the cache it should be less. 

I believe the recompiling is due to some modification time stamps on files, but I haven’t checked that part. 

Edit: the issue in pio is: https://github.com/platformio/platformio-core/issues/4574

ssieb:
I don't know about the addon, but when using the cli, the build cache works.  A recompile only compiles the files required for what's changed, which often is only `main.cpp`.

jesserockz:
ESPHome will always recompile everything after an ESPHome version update by design due to issues in the past with changed/removed files.

LordMike:
Ooh.. I think I saw this hack somewhere, a function in pio was being replaced in esphome.. It had something related to hashing of sources and timestamps.

Edit: The hack: https://github.com/esphome/esphome/blob/b184b01600c9cb20f8d130dbac270809ac706c24/esphome/platformio_api.py#L18-L41

This just makes this PR useless right now, but essential once PIO (hopefully) fixes caching such that it can be used between devices. 

It may even be that you could remove that hack, enable this PR, and then on builds clear out the output folder before compiling. The underlying SCons caches items based on their contents, while the caching esphome is currently relying on - if I understand correctly - is that the output objects (`main.cpp.o`) modification timestamp is newer than the source file (`main.cpp`)..

As I saw SCons (underlying system for PIO), it not only looks at the contents of the thing being compiled, but also all referenced sources. So this feature could be relied on, and maybe the past issues would disappear by themselves :).

PR #4582, opened 2023-03-18
Add a simple 'skip_initial' filter

sybrenstuvel:
This filter simply skips the first `N` values, then passes everything as-is. This is quite useful when you know the first few sensor readings should be ignored.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes my unreported issue, briefly discussed in Discord.

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2760

## Test Environment

I tested on these, but it should be independent on the exact hardware used.

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
sensor:
  - platform: sgp30
    id: mysensor_sgp30
    eco2:
      id: mysensor_sgp30_co2
      name: "eCO₂"
      accuracy_decimals: 0
      filters:
        - skip_initial: 41
```

## Checklist:

### DIFF
--- a/esphome/components/sensor/__init__.py
+++ b/esphome/components/sensor/__init__.py
@@ -194,7 +194,7 @@ SensorPublishAction = sensor_ns.class_("SensorPublishAction", automation.Action)
 Filter = sensor_ns.class_("Filter")
 QuantileFilter = sensor_ns.class_("QuantileFilter", Filter)
 MedianFilter = sensor_ns.class_("MedianFilter", Filter)
-SkipFilter = sensor_ns.class_("SkipFilter", Filter)
+SkipInitialFilter = sensor_ns.class_("SkipInitialFilter", Filter)
 MinFilter = sensor_ns.class_("MinFilter", Filter)
 MaxFilter = sensor_ns.class_("MaxFilter", Filter)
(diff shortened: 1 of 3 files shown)


jesserockz (commented):
Why make it more complex with a sub schema requiring `send_first_at` instead of just a filter named something like `skip_first` / `skip_initial` that takes only an integer as the value?

nagyrobi:
I think usage of `send_first_at` term is a bit confusing in this context.

sybrenstuvel:
> Why make it more complex with a sub schema requiring `send_first_at` instead of just a filter named something like `skip_first` / `skip_initial` that takes only an integer as the value?

That's mostly unfamiliarity with the ESPHome Python code to construct the schemas. I just copied code from the median filter, which is the filter I used to solve my problem (but couldn't because of the `send_first_at <= send_every` requirement). And for consistency with the other filters I'm familiar with.

I have no strong feelings about how this filter is exposed in the YAML file, I'll update this PR for the feedback.

jesserockz (approved):
Looks good. Thanks for making the changes

PR #4584, opened 2023-03-18
Added support for M5CoreInk display gdew0154m09

tobiasoort:

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** n/a

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2762

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
spi:
  clk_pin: GPIO18
  mosi_pin: GPIO23

display:
  - platform: gdew0154m09
    id: my_display
    reset_pin: GPIO0
    cs_pin: GPIO9
    dc_pin: GPIO15
    busy_pin: GPIO4
    rotation: 0
    update_interval: 60s

```

## Checklist:


esphome/components/gdew0154m09/gdew0154m09.cpp
```diff
@@ -0,0 +1,278 @@
...
+    for (int count = 0; count < pixsize; count++) {
+      data_(0xff); // set every pixel to ff
+    }
+  }
+  return 0;
+}
+
+int GDEW0154M09::clear(int mode) {
+  int pixsize = this->get_buffer_length_();
+  if (mode == 0) {
```
nielsnl68: Hi, i would suggest to use `static const` here instead a numeriek , so ever one can later understand what this mode does.

esphome/components/gdew0154m09/gdew0154m09.cpp
```diff
@@ -0,0 +1,278 @@
...
+  }
+  return 0;
+}
+
+size_t GDEW0154M09::get_buffer_length_() {
+  return size_t(this->get_width_internal()) * size_t(this->get_height_internal()) / 8u;
+}
+
+void HOT GDEW0154M09::display_() {
+  // this->clear_dsram_();
```
nielsnl68, line 209: Please let you inspire here by the *ili9xxx* component to allow partially screen updates.

esphome/components/gdew0154m09/gdew0154m09.cpp
```diff
@@ -0,0 +1,278 @@
...
+    delay(10);
+    App.feed_wdt();
+  }
+  return true;
+}
+
+int GDEW0154M09::clear_dsram_() {
+  int pixsize = this->get_buffer_length_();
+  for (int i = 0; i < 2; i++) {
+    command_(0x10); // data transmission start DTM1
```
nielsnl68: To allow a better understanding of the commands you use it is a better practice to setup a list of `static const` with all used commands and replace the numeric commands in you code.

nielsnl68 (changes requested):
Hi i added some suggestions to make your component more readable and understanding for others.
I hope you they are useful to you.

esphome/components/gdew0154m09/gdew0154m09.cpp
tobiasoort: Agree. This was a 1:1 port from the M5Stack example code; thanks for the feedback because that will definitely make it clearer.

esphome/components/gdew0154m09/gdew0154m09.cpp
tobiasoort: Partial mode is total barf on this display at the moment. I'd prefer to get non-partial polished and finish before diving into the complexities of partial updates.

esphome/components/gdew0154m09/gdew0154m09.cpp
tobiasoort: This has diminishing value as the specs are very very uncomprehensible and only a few make sense, and some are even undocumented yet used by the M5Stack example code (i'm looking at you, `0xaa`). I started to document the commands but after starting could already stop by lack of good docs from the vendor.

esphome/components/gdew0154m09/gdew0154m09.cpp
nielsnl68: Top, good luck :)

esphome/components/gdew0154m09/gdew0154m09.cpp
nielsnl68: Best choose,

esphome/components/gdew0154m09/gdew0154m09.cpp
nielsnl68: Ach, I know this issue as well, i found a lot of undocumented Command code's while i was upgrade the ili9xxx and i still missing some of them.

tobiasoort: done; implemented all commands and some data messages where handy.

esphome/components/gdew0154m09/display.py
```diff
@@ -0,0 +1,56 @@
...
+)
+
+
+CONFIG_SCHEMA = cv.All(
+    display.FULL_DISPLAY_SCHEMA.extend(
+        {
+            cv.GenerateID(): cv.declare_id(GDEW0154M09),
+            cv.Required(CONF_DC_PIN): pins.gpio_output_pin_schema,
+            cv.Required(CONF_RESET_PIN): pins.gpio_output_pin_schema,
+            cv.Required(CONF_CS_PIN): pins.gpio_output_pin_schema,
```
jesserockz: This is provided by the `spi.spi_device_schema()` already
```suggestion
```

esphome/components/gdew0154m09/display.py
```diff
@@ -0,0 +1,56 @@
...
+            cv.Optional(CONF_BUSY_PIN): pins.gpio_input_pin_schema,
+        }
+    )
+    .extend(cv.polling_component_schema("1s"))
+    .extend(spi.spi_device_schema()),
```
jesserockz: To change the CS pin to required
```suggestion
    .extend(spi.spi_device_schema(True)),
```

esphome/components/gdew0154m09/display.py
```diff
@@ -0,0 +1,56 @@
...
+    await cg.register_component(var, config)
+    await display.register_display(var, config)
+    await spi.register_spi_device(var, config)
+
+    dc = await cg.gpio_pin_expression(config[CONF_DC_PIN])
+    cg.add(var.set_dc_pin(dc))
+    reset = await cg.gpio_pin_expression(config[CONF_RESET_PIN])
+    cg.add(var.set_reset_pin(reset))
+    busy = await cg.gpio_pin_expression(config[CONF_BUSY_PIN])
+    cg.add(var.set_busy_pin(busy))
```
jesserockz: This pin is set as optional so you have to check its set
```suggestion
    if CONF_BUSY_PIN in config:
        busy = await cg.gpio_pin_expression(config[CONF_BUSY_PIN])
        cg.add(var.set_busy_pin(busy))
```

esphome/components/gdew0154m09/gdew0154m09_defines.h
```diff
@@ -0,0 +1,27 @@
...
+
+static const int CMD_DTM1_DATA_START_TRANS = 0x10;
+static const int CMD_DTM2_DATA_START_TRANS2 = 0x13;
+static const int CMD_DISPLAY_REFRESH = 0x12;
+static const int CMD_PSR_PANEL_SETTING = 0x00;
+static const int CMD_UNDOCUMENTED_0x4D = 0x4D;  //  NOLINT
+static const int CMD_UNDOCUMENTED_0xAA = 0xaa;  //  NOLINT
+static const int CMD_UNDOCUMENTED_0xE9 = 0xe9;  //  NOLINT
+static const int CMD_UNDOCUMENTED_0xB6 = 0xb6;  //  NOLINT
+static const int CMD_UNDOCUMENTED_0xF3 = 0xf3;  //  NOLINT
```
jesserockz: Why are these `NOLINT`ed?

esphome/components/gdew0154m09/gdew0154m09_defines.h
jesserockz, line 1: These should all be `uint8_t` type to save on flash space

esphome/components/gdew0154m09/gdew0154m09.h
```diff
@@ -0,0 +1,114 @@
...
+    CMD_TRES_RESOLUTION_SETTING, 3, 0xc8, 0x00, 0xc8,
+    CMD_TCON_TCONSETTING, 1, 0x00,
+    CMD_CDI_VCOM_DATA_INTERVAL, 1, 0xd7,
+    CMD_PWS_POWER_SAVING, 1, 0x00,
+    CMD_PON_POWER_ON, 0
+};
+// clang-format on
+
+// LUTs provided for future reference - currently unused.
+const unsigned char LUT_VCOM_D_C1[] PROGMEM = {
```
jesserockz: Please use `uint8_t` instead of `unsigned char`
```suggestion
const uint8_t LUT_VCOM_D_C1[] PROGMEM = {
```

esphome/components/gdew0154m09/gdew0154m09.cpp
```diff
@@ -0,0 +1,274 @@
...
+}
+
+void HOT GDEW0154M09::display_() {
+  // this->clear_dsram_();
+  this->set_draw_addr_(0, 0, GDEW0154M09_WIDTH, GDEW0154M09_HEIGHT);
+  this->draw_buff_(this->lastbuff_, this->buffer_, this->get_buffer_length_());
+}
+
+void GDEW0154M09::draw_buff_(uint8_t *lastbuff, uint8_t *buff, size_t size) {
+  command_(CMD_DTM1_DATA_START_TRANS);
```
jesserockz: This is just one random line, please change to consistently use `this->`
```suggestion
  this->command_(CMD_DTM1_DATA_START_TRANS);
```

jesserockz (changes requested):

esphome/components/gdew0154m09/gdew0154m09_defines.h
tobiasoort: because clang-format wants to change `0x4D` into `0x4_D` since it thinks letters and numbers should always be separate.

tobiasoort: example output from `clang-tidy for ESP32 Arduino`:

```
-static const int CMD_UNDOCUMENTED_0x4D = 0x4D;
-static const int CMD_UNDOCUMENTED_0xAA = 0xaa;
-static const int CMD_UNDOCUMENTED_0xE9 = 0xe9;
-static const int CMD_UNDOCUMENTED_0xB6 = 0xb6;
-static const int CMD_UNDOCUMENTED_0xF3 = 0xf3;
+static const int CMD_UNDOCUMENTED_0X4_D = 0x4D;
+static const int CMD_UNDOCUMENTED_0X_AA = 0xaa;
+static const int CMD_UNDOCUMENTED_0X_E9 = 0xe9;
+static const int CMD_UNDOCUMENTED_0X_B6 = 0xb6;
+static const int CMD_UNDOCUMENTED_0X_F3 = 0xf3;
```

jesserockz: haha, thats fine then. Was just curious

tobiasoort:
@jesserockz all requested changes have been done.

esphome/components/gdew0154m09/gdew0154m09.cpp
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  const uint32_t start = millis();
+  while (this->busy_pin_->digital_read()) {
+    uint32_t waittime = millis() - start;
+    if (waittime > timeout) {
+      ESP_LOGE(TAG, "Timeout while displaying image! was %dms", waittime);
+      return false;
+    }
+    delay(10);
```
kbx81, line 122: You're just checking the state of a GPIO pin, so I would argue the `delay()` is not necessary here -- let it roll out of the loop as soon as the line clears. 🙂

esphome/components/gdew0154m09/gdew0154m09.cpp
```diff
@@ -0,0 +1,273 @@
...
+    this->command_(CMD_DISPLAY_REFRESH);
+    this->wait_until_idle_();
+  }
+
+  return 0;
+}
+
+int GDEW0154M09::write_init_list_(const unsigned char *list) {
+  int list_limit = list[0];
+  unsigned char *start_ptr = ((unsigned char *) list + 1);
```
kbx81, line 193: For consistency with the rest of the platform, please use `uint8_t` instead of `unsigned char` -- also, instead of just `int`, I tend to push people towards `int32_t` or  `uint32_t`, as well. 😄

kbx81 (changes requested):
Couple nit-picks, but otherwise LGTM 😄

tobiasoort:
I'm I changed this PR to draft. After tinkering with another E-paper I noticed two things:

- It seems a lot of e-paper support (for various devices of various vendors) is accumulated in the `waveshare_epaper` platform. I'm going to try to integrate this display in there as well
- After testing my code extensively (for months) in hardware, I noticed undesirable wear on the e-ink display. I probably didn't write the proper power-down procedures for this display. I have to fix this before we go to main.

esphome/components/gdew0154m09/gdew0154m09.cpp
tobiasoort: I would think that would be less power-efficient, but I don't know how delay is handled on-cpu (if its just a busy-wait then it makes no difference). I stole this from other display components with similar setups which also included a similar `delay()` call. I'm happy to learn :)

kbx81: As far as I know, the only way it would be more power-efficient is if the CPU sleeps during the delay; I could be incorrect myself, but I don't believe that's the case with this environment.

tobiasoort:
Closing this pullrequest - will port the (improved) code into the `waveshare_epaper` component.

PR #4593, opened 2023-03-20
Fix EzoCommandType enum

### LINKED ISSUE #4321: [ezo] Bug in EzoCommandType enum
### The problem

I'm having issues triggering the ezo callbacks (specifically "on_calibration"). Looking at the code, I identified the following, which seems a mistake:

https://github.com/esphome/esphome/blob/a31fb3c987b7011b7ccb7735f8210f2feaff1d90/esphome/components/ezo/ezo.h#L18

According to C++ rules, EZO_CALIBRATION gets implicitly assigned the value "4" (previous enum value +1), which is the same of the explicitly assigned EZO_SLEEP.

I believe the code ends up never reading the response from an issued EZO_CALIBRATION command, because looking at this line:
https://github.com/esphome/esphome/blob/a31fb3c987b7011b7ccb7735f8210f2feaff1d90/esphome/components/ezo/ezo.cpp#L67
the command gets confused with EZO_SLEEP, and it is removed from the queue before the reply can be parsed.

The fix should be simple, by updating the enum definition such that every item has a unique value.

However, I'm not submitting a PR as I don't have a test environment. Available if I can support in any other way.

2023.2.4

Home Assistant Add-on

_No response_

ESP32

### Board

nodemcu-32s

ezo

### Example YAML snippet

```yaml
button: # Used to trigger sending the I2C message
  - platform: template
    name: "Retrieve pH calibration status"
    on_press:
      - lambda: |-
          id(ezo_ph).get_calibration();

text_sensor: # Used to see the outcome
  - platform: template
    name: "pH calibration status"
    id: ph_calibration_status

sensor:
  - platform: ezo
    id: ezo_ph
    address: 99
    name: "Swimming pool pH"
    unit_of_measurement: "pH"
    accuracy_decimals: 2
    update_interval: 10s
    on_calibration: # Note this callback never gets triggered
      - text_sensor.template.publish:
          id: ph_calibration_status
          state: "test state"
```

_No response_

### Additional information

_No response_

alfredopironti:
Assign explicit value to EZO_CALIBRATION, and rescale all subsequent values.

The current enum definition misses explicit assignment to the EZO_CALIBRATION value. This makes EZO_CALIBRATION and EZO_SLEEP end up with the same value, which causes malfunction in the code. Also, the indexing between EzoCommandType enum and [EZO_COMMAND_TYPE_STRINGS[]](https://github.com/esphome/esphome/blob/cd57469e06aaf223c518c72b17c07e80722982fc/esphome/components/ezo/ezo.cpp#L8) is currently broken.

This patch assigns an explicit value to EZO_CALIBRATION, and rescales all subsequent values, to make the enum work as intended.

## Types of changes

- [X ] Bugfix (non-breaking change which fixes an issue)
- [ ] Other

**Related issue or feature (if applicable):** fixes [#4321](https://github.com/esphome/issues/issues/4321)

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:
  - [ X] The code change is tested and works locally.

### DIFF
--- a/esphome/components/ezo/ezo.h
+++ b/esphome/components/ezo/ezo.h
@@ -12,14 +12,14 @@ static const char *const TAG = "ezo.sensor";
 
 enum EzoCommandType : uint8_t {
   EZO_READ = 0,
-  EZO_LED = 1,
-  EZO_DEVICE_INFORMATION = 2,
-  EZO_SLOPE = 3,
-  EZO_CALIBRATION = 4,
-  EZO_SLEEP = 5,
-  EZO_I2C = 6,
-  EZO_T = 7,
-  EZO_CUSTOM = 8
+  EZO_LED,
+  EZO_DEVICE_INFORMATION,
+  EZO_SLOPE,
+  EZO_CALIBRATION,
+  EZO_SLEEP,
+  EZO_I2C,
+  EZO_T,
+  EZO_CUSTOM
 };
 
 enum EzoCalibrationType : uint8_t { EZO_CAL_LOW = 0, EZO_CAL_MID = 1, EZO_CAL_HIGH = 2 };
--- a/esphome/components/ezo/ezo.h
+++ b/esphome/components/ezo/ezo.h
@@ -15,11 +15,11 @@ enum EzoCommandType : uint8_t {
   EZO_LED = 1,
   EZO_DEVICE_INFORMATION = 2,
   EZO_SLOPE = 3,
-  EZO_CALIBRATION,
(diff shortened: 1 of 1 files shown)


jesserockz (changes requested):
Lets just remove the values entirely so it doesn't get missed in the future.

jesserockz (approved):

ssieb (approved):

PR #4594, opened 2023-03-20
`EntityBase` Name can stay in flash.

Fabian-Schmidt:

This is a memory improvement. The `EntityBase` string `name_` does not need to be copied to memory, instead it can stay in Flash. Each std::string instance (even empty once) occupy at least 24 bytes (15 byte capacity + length + flags) of memory.
By using `StringRef` instead of std::string they stay in flash. 
`StringRef` is adopted from nghttp2 and has high compatibility with std::string to minimize code changes, while occupying 8 instead of (at least) 24 bytes per instance.

This PR removes the (mostly) unused constructor of `EntityBase`. It removes both uses of the constructor (Sprinkler and Partition) and all child's calling this constructor.
The exclusive only way to set the name is via the method `set_name`.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


Fabian-Schmidt:
Sorry @kbx81 to have bothered you with this.  I removed sprinkler from this PR.

kbx81:
No worries -- it prompted some other discussion which led to other optimizations. Nothing wrong with that! 😄

jesserockz (approved):
Looks good to me.

PR #4044, opened 2022-11-18
small DisplayBuffer images and font update

nielsnl68:
# What does this implement/fix?

Changed the way transparant images are saved. 
- TRANSPANT_BINARY: are now working the same way as BINARY images.
- TRANSPARENT_IMAGE: 🆕 replaces the the old TRANSPANT_BINARY.

Renaming the font `bottom` to `height` and allow the get the baseline and height.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4600, opened 2023-03-20
Add transparency support to all image types

### LINKED ISSUE #4226: Waveshare E-paper Images rendered as black squares
### The problem

Since version 2023.2.3 (not sure it started there, because I only update my firmware when necessary), images are rendered as black squares. These are PNG with a transparency layer. Before they would render correctly.

So the image is loaded as such: 

```
image:
  - file: "images/restafval2.png"
    id: Lrestafvalpic
    type: TRANSPARENT_BINARY
```
Then it is called upon like such in the display section:
```
it.image(255, 450, id(Lrestafvalpic));
```
This generates a black square.
This is the actual image used: 
restafval2
This is the result:
IMG_1486

2023.2.3

Docker

2023.2.3

ESP32

### Board

Firebeetle32

display/it.image

### Example YAML snippet

```yaml
image:
  - file: "images/restafval2.png"
    id: Lrestafvalpic
    type: TRANSPARENT_BINARY

display:
  - platform: waveshare_epaper
    cs_pin: 22
    dc_pin: 21
    busy_pin: 5
    reset_pin: 13
    model: 7.50in-hd-b
    update_interval: never
    reset_duration: 2ms
    rotation: 90°
    id: epaper_display
    lambda: |-
      it.image(255, 450, id(Lrestafvalpic));
```

```txt
No errors in the log.
```

### Additional information

_No response_

guillempages:
Add support for transparency for all image and animation types.
Also add support for 32 bit images, including an alpha channel. If the alpha channel is 0, the pixel will not be drawn, so that underlying pixels are still shown.

A new configuration option has been introduced for images `use_transparency`, which when set to true will convert the alpha channel of the original image to the color "(0,0,1)" (i.e. very dark blue), and when the renderer finds this color it will skip the pixel.

TRANSPARENT_BINARY and RGBA image types have `use_transparency=true` by default and cannot be set to false; all other types have it `false` by default.

## Types of changes

- [ ] Other

**Breaking change**
This removes the recently introduced (but still undocumented) "TRANSPARENT_IMAGE" type. The same can be achieved by using any other image type and the new `use_transparency: true` option.

**Related issue or feature (if applicable):** Fixes esphome/issues#4226

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2770

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

Has been tested with a Lilygo T-Display-S3 TFT, with a Lilygo T5 4.7 E-Paper and with an Inkplate 10 E-Paper

```yaml
image:
 - file: "example.png"
   id: my_transparent_image
   type: RGBA
   # use_transparency: True # Implicit for RGBA images
 - file: "another_example.png"
   id: my_transparent_small_image
   type: RGB565
   use_transparency: true
 - file: "grayscale_example.png"
   id: my_transparent_grayscale
   type: GRAYSCALE
   use_transparency: true

animation:
  - file: "animation.gif"
    id: my_animation
    type: BINARY
    use_transparency: true
```

## Checklist:


esphome/components/image/__init__.py
```diff
@@ -94,7 +110,7 @@ async def to_code(config):
     elif config[CONF_TYPE] == "RGB565":
         image = image.convert("RGB")
         pixels = list(image.getdata())
-        data = [0 for _ in range(height * width * 3)]
+        data = [0 for _ in range(height * width * 2)]
```
guillempages, line 158: This is not really related to this PR, but saw that while implementing it.
The RGB565 was using 3 bytes per pixel instead of the just needed 2, so the allocated buffer was too big.

guillempages: If desired, I could put that into a separate PR

nielsnl68:
Hi, @guillempages i also did some work on the images in https://github.com/esphome/esphome/pull/4398.
Can we work together to make one PR?

guillempages:
@nielsnl68 I should probably have waited two days until you got yours merged :-D

Jokes aside, sure, we can try to combine our efforts. I was just trying to support 32 bit images (and treat the alpha channel as all-or-nothing), and also tried to keep the amount of changes minimal.

It's pretty late now here (CET), so I'll call it a day, but we can try to sync tomorrow. Which timezone are you in? You can try pinging me on discord at say 21:00 CET (same nick)

nielsnl68:
> @nielsnl68 I should probably have waited two days until you got yours merged :-D

Then you would have had a long wait. My PR is not ready to be merge yet. I have still some issues to fix.

I'm from the Netherlands ;) 
And same here you can also try to ping me on Discord as well.

esphome/components/image/__init__.py
```diff
@@ -22,13 +22,16 @@
 ImageType = display.display_ns.enum("ImageType")
 IMAGE_TYPE = {
     "BINARY": ImageType.IMAGE_TYPE_BINARY,
+    "TRANSPARENT_BINARY": ImageType.IMAGE_TYPE_BINARY,
     "GRAYSCALE": ImageType.IMAGE_TYPE_GRAYSCALE,
-    "RGB24": ImageType.IMAGE_TYPE_RGB24,
-    "TRANSPARENT_BINARY": ImageType.IMAGE_TYPE_TRANSPARENT_BINARY,
     "RGB565": ImageType.IMAGE_TYPE_RGB565,
-    "TRANSPARENT_IMAGE": ImageType.IMAGE_TYPE_TRANSPARENT_BINARY,
+    "TRANSPARENT_IMAGE": ImageType.IMAGE_TYPE_RGB565,
```
guillempages: TRANSPARENT_IMAGE was just added recently (#4044, merged Jan 11th); it might be worth marking this as a breaking change, and removing the TRANSPARENT_IMAGE type completely, since it now feels out-of-place. Also, this image type was never documented anyway.
Opinions?

esphome/components/image/__init__.py
guillempages: I think the original scope of this PR has increased so much, that having this small fix in it is OK.
If the core reviewers disagree, I'd be happy to do the extra PR, but closing this comment for the moment.

esphome/components/image/__init__.py
nielsnl68: I agree on this.

I introduced the `TRANSPARANT_IMAGE` because the old `TRANSPARENT_BINARY` was a hybrid between RGBA and binary in the conversion method.

In the https://github.com/esphome/esphome/pull/4398 that you used for this PR as agreed on. Both Transparant types are ment for temporary placeholders, so the changes older yaml code would stay running.

esphome/components/image/__init__.py
```diff
@@ -64,72 +71,124 @@ async def to_code(config):
...
 
+    is_transparent_type = config[CONF_TYPE] in [
+        "TRANSPARENT_BINARY",
+        "TRANSPARENT_IMAGE",
+        "RGBA",
+    ]
+    if config.get(CONF_USE_TRANSPARENCY, None) is False and is_transparent_type:
+        # TODO: Would be nice to also print the line where the error happened
+        raise core.EsphomeError(
+            f'Image "{config[CONF_ID]}": Image type {config[CONF_TYPE]} must always be transparent.'
```
nielsnl68: I do not see the need for this error message. Just set the `CONF_USE_TRANSPARENCY` flag to true automatically.

esphome/components/image/__init__.py
```diff
@@ -64,72 +71,124 @@ async def to_code(config):
...
+                    if not a:
+                        continue
+                elif image.getpixel((x, y)):
                     continue
                 pos = x + y * width8
                 data[pos // 8] |= 0x80 >> (pos % 8)
+    else:
+        # TODO: Would be nice to also print the line where the error happened
+        raise core.EsphomeError(
+            f"Image f{config[CONF_ID]} has not supported type {config[CONF_TYPE]}."
```
nielsnl68:
```suggestion
            f"Image f{config[CONF_ID]} has not a supported type {config[CONF_TYPE]}."
```

esphome/components/image/__init__.py
```diff
@@ -64,72 +71,124 @@ async def to_code(config):
...
-                " the resize parameter."
+                'The image "%s" you requested is very big. Please consider'
+                " using the resize parameter.",
+                path,
             )
 
+    is_transparent_type = config[CONF_TYPE] in [
+        "TRANSPARENT_BINARY",
+        "TRANSPARENT_IMAGE",
+        "RGBA",
```
nielsnl68: Not sure of this is really needed. We have that `CONF_USE_TRANSPARENCY` is that not enough?

nielsnl68 (changes requested):
looks to be almost okay. 

I added some suggestions to modify below. Thanks for taking over my earlier PR.

esphome/components/image/__init__.py
guillempages: I would keep the TRANSPARENT_BINARY, since I expect it to be used in a lot of legacy code, and remove the TRANSPARENT_IMAGE one, unless @jesserockz advises otherwise

esphome/components/image/__init__.py
guillempages: TRANSPARENT_BINARY is used in a lot of legacy code, and RGBA is transparent by definition (A is the alpha channel). This forces the transparency for those "transparent-by-definition" types, so that you can just use `type: RGBA` or `type: TRANSPARENT_BINARY` and avoid adding the `use_transparency` option

esphome/components/image/__init__.py
guillempages: I think it is nicer to make the user aware of mistakes in the configuration, than just override his "wishes" if they make no sense. @jesserockz Do you have a policy regarding this?

esphome/components/image/__init__.py
guillempages: How about?
```suggestion
            f"Image f{config[CONF_ID]} has an unsupported type: {config[CONF_TYPE]}."
```

esphome/components/display/display_buffer.cpp
```diff
@@ -303,43 +304,35 @@ void DisplayBuffer::vprintf_(int x, int y, Font *font, Color color, TextAlign al
...
-      for (int img_x = 0; img_x < image->get_width(); img_x++) {
-        for (int img_y = 0; img_y < image->get_height(); img_y++) {
-          if (image->get_pixel(img_x, img_y))
-            this->draw_pixel_at(x + img_x, y + img_y, color_on);
-        }
+  Color color;
+  bool transparent = image->has_transparency();
+  for (int img_x = 0; img_x < image->get_width(); img_x++) {
+    for (int img_y = 0; img_y < image->get_height(); img_y++) {
+      switch (image->get_type()) {
```
guillempages: After doing some benchmarking, having a separate for loop within each switch case is about 10% faster than having a single loop with a switch within:

Original code (without transparency implementation):
    2.142 2.135 2.131 2.122 2.121 2.121 s

With transparency (switch-within-single-for):
    2.334 2.332 2.329 2.322 2.321 2.319 s

With transparency, using a lambda to avoid the switch (single for):
    2.334 2.332 2.329 2.322 2.321 2.319 s

With transparency using multiple-for-within-switch:
    2.186 2.185 2.184 2.177 2.176 2.172 s

esphome/components/image/__init__.py
nielsnl68: Much better :D

esphome/components/image/__init__.py
nielsnl68: Maybe i read this snapshot wrong, when will this exception being raised? 
Is it only when the user set the CONF_USE_TRANSPARENCY to false and choose image type with transparency?
When so, then this exception is legitimate.

guillempages: Exactly, this exception is only thrown when the user explicitly sets one of the transparent types, and `use_transparency` to false.

esphome/components/display/display_buffer.cpp
guillempages: The final implementation results are pretty similar to the original code (when transparency is not used). Of course the timings might be different depending on the chosen image.

2.136 2.135 2.134 2.125 2.125 2.125 s

nielsnl68 (approved):

esphome/components/animation/__init__.py
```diff
@@ -50,33 +55,51 @@ async def to_code(config):
...
 
+    is_transparent_type = config[CONF_TYPE] in [
+        "TRANSPARENT_BINARY",
+        "RGBA",
+    ]
+    if config.get(CONF_USE_TRANSPARENCY, None) is False and is_transparent_type:
+        # TODO: Would be nice to also print the line where the error happened
+        raise core.EsphomeError(
+            f'Animation "{config[CONF_ID]}": Image type {config[CONF_TYPE]} must always be transparent.'
+        )
```
jesserockz: This should be a custom validation, not something checked in `to_code`

guillempages: makes sense; I'll try to find out how to do that later today

guillempages: done. And it does feel much better :-)

Unfocused:
Just looked to see if this supported chroma-key transparency, and saw that's what you're doing internally. So could I request that colour be configurable? 

Especially with the resurengce of retro computing (hot pink #FF00FF was commonly used, rather than green), there's still a fair amount of image resources available that use it. And all those low pixel resources can work great on LED matrixes especially.

As a bonus, having it configurable would let people avoid issues with any image already using the default chroma-key colour for parts of the image that isn't meant to be transarent.

.gitattributes
```diff
@@ -1,2 +1,3 @@
 # Normalize line endings to LF in the repository
 * text eol=lf
+*.png binary
```
Unfocused, line 3: This shouldn't be needed - Git is pretty good at auto-detecting binary files

guillempages: "should" is a very bad word in engineering 😉
Actually I needed to add this explicitly; otherwise the newly added file for the tests was not being detected right and the test was failing.

So I agree it "shouldn't" be needed, but unfortunately it is.

guillempages:
> Just looked to see if this supported chroma-key transparency, and saw that's what you're doing internally. So could I request that colour be configurable?
> 
> Especially with the resurengce of retro computing (hot pink #FF00FF was commonly used, rather than green), there's still a fair amount of image resources available that use it. And all those low pixel resources can work great on LED matrixes especially.
> 
> As a bonus, having it configurable would let people avoid issues with any image already using the default chroma-key colour for parts of the image that isn't meant to be transarent.

I'm not exactly sure what you are proposing here. The code is not actually using a color in the original image as a transparent color, but it is expecting Pillow to create an alpha-channel image from the original one. Then, the information on the alpha channel is used to decide whether a pixel is transparent or not. The actual color (very dark blue) used to store the transparency values in non-alpha-mode images is just an internal implementation, and was chosen because normally it would be "almost" black, so not really noticeable. I would expect retrocomputing images to be still parsed by pillow, and generate a "normal" RGBA representation from it, so that the final shown image would still work.

Do you have an actual example of an image that is not being displayed correctly with this implementation, or are we talking "theoretical problem" here? ;-)

Initially I also thought about making this configurable, but while implementing found out that it would complicate both configuration and code for no practical benefit. (And before you say that just having a variable color, instead of the hard-coded (0,0,1) is not much more complex: not in terms of written code, but in terms of generated code; this code is used in a hot loop, and I already spent quite some time optimizing it as much as I could; even switching the order of the conditions to improve the "common case" performance)

esphome/components/display/display_buffer.cpp
```diff
@@ -643,14 +674,22 @@ Color Image::get_rgb565_pixel(int x, int y) const {
   auto r = (rgb565 & 0xF800) >> 11;
   auto g = (rgb565 & 0x07E0) >> 5;
   auto b = rgb565 & 0x001F;
-  return Color((r << 3) | (r >> 2), (g << 2) | (g >> 4), (b << 3) | (b >> 2));
+  Color color = Color((r << 3) | (r >> 2), (g << 2) | (g >> 4), (b << 3) | (b >> 2));
+  if (rgb565 == 1 && transparent_) {
```
Unfocused: Ugh, this hurt to read. I think this would be much clearer if you specificed the (0,0,1) colour in hex or binary or something, anything other than decimal `1`. Ditto in the Python code.

```suggestion
  if (rgb565 == 0x0001 && transparent_) {
```

Unfocused:
Ah, my apologies, I've misunderstood. But I've read through the code more closely now. I missed that you're altering the colours of the pixels if it finds a pixel in the original image of the colour `(0,0,1)`. That wasn't mentioned in the description, so I had assumed it was leaving them unmodified - which would have meant any images with pixels of that colour would have parts drawn as transparent that weren't meant to be transparent.

My first comment on this was perhaps pooorly worded. Using this to for configurable chroma-keyed transparency for images without an alpha channel would be a nice thing to have. But my (wrong) assumption above would have been a bug that would have effectively make some images unusable.

Being able to configure the chroma-key colour would have been an alternative solution to that issue. But I understand why you've chosen to alter the original pixel colour instead.

Although, given that the bit depth of the colours is significantly reduced for `RGB565`, the colour change will be more noticable. I think that needs to be documented. However, the effect could be signicantly reduced if you change the chroma-key to `(0,1,0)` for the `RGB565` format (using green instead of blue). Since you'd be taking a colour value away from 6-bit color (green channel, 64 levels of colour) instead of 5-bit colour (blue channel, only 32 levels of colour).

guillempages:
> Although, given that the bit depth of the colours is significantly reduced for RGB565, the colour change will be more noticable. I think that needs to be documented. However, the effect could be signicantly reduced if you change the chroma-key to (0,1,0) for the RGB565 format (using green instead of blue). Since you'd be taking a colour value away from 6-bit color (green channel, 64 levels of colour) instead of 5-bit colour (blue channel, only 32 levels of colour).

I've tested it (on a single display, maybe not representative) by painting some rectangles with the darkest RGB565 blue and rgb565 green and I don't think it really makes a difference. AFAIK, the human eye is less sensitive to blue than to green; that's why I chose blue.

That said, It is true that the green channel has one bit more, so I changed the code to use green instead of blue for RGB565 images for the transparent color.
This might cause maintainability issues later on though, since different bit depths use different colors for storing the transparent color. (i.e. inconsistency). Let's see what other reviewers think about that.

jesserockz (approved):

PR #4604, opened 2023-03-21
Limit range on filter time period for remote_receiver

RoboMagus:

Adds range validation to the `CONF_FILTER` option in `remote_receiver` component. As in firmware this value is truncated to 8 bits, this should be better reflected during configuration. Without the upper bound validation for the range the value entered by the user may result in integer rolover with only a compiler warning that may not provide enough information to unknowing users. 

It's quite reasonable the user may wish to filter out spikes larger than 255us and without feedback during configuration would not be aware of this limitation.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
remote_receiver:
  pin: D1
  dump: all
  tolerance: 50%
  filter: 260us # <-- This would happily compile, only with a compiler warning of w.r.t integer range. But results in an effective value of 5us.
  idle: 4ms

```

## Checklist:

### DIFF
--- a/esphome/components/remote_receiver/__init__.py
+++ b/esphome/components/remote_receiver/__init__.py
@@ -12,7 +12,7 @@ from esphome.const import (
     CONF_TOLERANCE,
     CONF_MEMORY_BLOCKS,
 )
-from esphome.core import CORE
+from esphome.core import CORE, TimePeriod
 
 AUTO_LOAD = ["remote_base"]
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Nice, thanks!

PR #4614, opened 2023-03-23
Adds i2c timeout config

### LINKED ISSUE #4340: I2C timeout too short for PN532
### The problem

Connecting a PN532 NFC/RFID board via I2C to ESP32 (esp-idf) does not work. Sometimes it gets enumerated (0x24 address), but mostly not. Anyway, the PN532 component initialization never succeeds. On the very verbose logs, the bus scan returns "timeout" on 0x24, and timeouts later in initialization, even if some early communication succeeds. Reducing the bus frequency to 10kHz helps a little, but does not solve the problem. 

With a logic analyzer, I noticed PN532 does "clock stretching", up to at least 440µs. That, mentioned in the component datasheet, and within I2C specs, is longer than the default timeout on ESP32 (100µs), with esp-idf at least. It is possible that this issue affects other I2C peripherals as well. The I2C component does not provide a way to change that timeout. 

I was able to change the I2C component code to include a configuration option for changing the timeout on esp-idf to 1000µs, which solves the problem in my setup. I will make a pull-request linked to this issue.

2023.3.0

Home Assistant Add-on

2023.03.1

ESP32-IDF

### Board

W32-ETH-01

i2c

### Example YAML snippet

```yaml
i2c:
  scl: GPIO32
  sda: GPIO33
  scan: True
  frequency: 50kHz

pn532_i2c:
  update_interval: 1s
```

```txt
[VV][i2c.idf:173]: 0x23 TX 
[VV][i2c.idf:210]: TX to 23 failed: not acked
[VV][i2c.idf:173]: 0x24 TX 
[VV][i2c.idf:213]: TX to 24 failed: timeout
[VV][i2c.idf:173]: 0x25 TX 
[VV][i2c.idf:210]: TX to 25 failed: not acked
...
[C][pn532:018]: Setting up PN532...
[VV][i2c.idf:173]: 0x24 TX 0000FF02FED4022A00
[V][pn532:246]: Reading ACK...
[VV][i2c.idf:149]: 0x24 RX 01
[VV][i2c.idf:149]: 0x24 RX 010000FF00FF00
[V][pn532:257]: ACK valid: YES
[V][pn532_i2c:043]: Reading response
[VV][i2c.idf:149]: 0x24 RX 01
[VV][i2c.idf:149]: 0x24 RX 010000FF06FAD5
[V][pn532:262]: Sending NACK for retransmit
[VV][i2c.idf:173]: 0x24 TX 0000FFFF0000
[VV][i2c.idf:213]: TX to 24 failed: timeout
[V][pn532_i2c:049]: Reading response of length 5
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[VV][i2c.idf:149]: 0x24 RX 00
[V][pn532_i2c:032]: Timed out waiting for readiness from PN532!
[D][pn532_i2c:051]: No response data
[E][pn532:029]: Error getting version
[E][component:113]: Component pn532 was marked as failed.
```

### Additional information

Changing the timeout to 1000µs (probably less) solves the issue.

tracestep:

PN532 in I2C mode (and probably other I2C peripherals) use clock stretching to be able to keep up (see https://github.com/esp8266/Arduino/issues/1541 and https://www.reddit.com/r/esp32/comments/hgwwr4/has_anyone_managed_to_get_i2c_working_with_the/). However, the default bus timeout on esp-idf is 100µs, which is too low (according to my observations, PN532 would require at least 500µs). There is no option to change timeout on the I2C component. This PR implements that.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4340

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2778

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
i2c:
  scl: GPIO32
  sda: GPIO33
  scan: True
  frequency: 50kHz
  timeout: 1ms
```

## Checklist:


esphome/components/i2c/__init__.py
```diff
@@ -75,6 +80,7 @@ async def to_code(config):
         cg.add(var.set_scl_pullup_enabled(config[CONF_SCL_PULLUP_ENABLED]))
 
     cg.add(var.set_frequency(int(config[CONF_FREQUENCY])))
+    cg.add(var.set_timeout(int(config[CONF_TIMEOUT].total_microseconds)))
```
jesserockz: Since this is only_with_esp_idf, this will crash on any other framework

jesserockz (changes requested):

esphome/components/i2c/__init__.py
tracestep: I'm sorry, I am not entirely familiar with the component configuration code, or ESPHome / Github testing workflow, by that matter. I was under the impression that by using `cv.Optional` and `only_with_esp_idf` the code would work if the `timeout` option was not used.

I corrected that by using `cv.SplitDefault` and testing for `CORE.using_esp_idf` before adding the `set_timeout` function. I contemplated adding the timeout option for the Arduino framework using `TwoWire::setTimeOut`, but the default I²C timeout there is 50ms, while the maximum timeout on esp-idf is around 13ms, so I don't know how to reconcile that without a breaking change. Since nobody else seems to be having problems, I suppose I²C timeouts are not an issue there.

esphome/components/i2c/__init__.py
```diff
@@ -80,8 +83,9 @@ async def to_code(config):
         cg.add(var.set_scl_pullup_enabled(config[CONF_SCL_PULLUP_ENABLED]))
 
     cg.add(var.set_frequency(int(config[CONF_FREQUENCY])))
-    cg.add(var.set_timeout(int(config[CONF_TIMEOUT].total_microseconds)))
     cg.add(var.set_scan(config[CONF_SCAN]))
+    if CORE.using_esp_idf:
```
jesserockz: The simplest here is to do this which opens it up if we allow the timeout to be set on arduino later without changing this line again.
```suggestion
    if CONF_TIMEOUT in config:
```

jesserockz (changes requested):

esphome/components/i2c/__init__.py
```diff
@@ -131,6 +146,20 @@ def final_validate_device_schema(
...
+            msg=f"Component {name} requires a minimum timeout of {min_timeout} for the I2C bus",
+        )
+
+    if max_timeout is not None:
+        hub_schema[cv.Required(CONF_TIMEOUT)] = cv.Range(
+            max=cv.time_period(max_timeout),
+            max_included=True,
+            msg=f"Component {name} cannot be used with a timeout of over {max_timeout} for the I2C bus",
+        )
+
```
jesserockz: I know this is not used anywhere yet, but the way this works will force a component to be esp-idf only.

jesserockz (changes requested):
The actual timeout changes look good.

But either need to remove the final validate changes, or update them to ignore min/max if not using idf

esphome/components/i2c/__init__.py
tracestep: Is there a way to set a different range in the schema depending on the framework? If not, what would the implications of validating max/min in the `final_validate_device_scheme` function against hard-coded values instead of ones defined on the schema?

The reason I did not implement `timeout` for the Arduino framework is that they have very different limits (1µs to 13107µs for esp-idf vs 1ms to >>50ms for Arduino)... if it is acceptable to check only in the validate function, it would be just a matter of testing the framework there.

In any case, leaving things as they are, would it be ok just to wrap lines 149 to 161 in a `if CORE.using_esp_idf: ` statement?

jesserockz: > In any case, leaving things as they are, would it be ok just to wrap lines 149 to 161 in a `if CORE.using_esp_idf:`  statement?

Yes wrap them. 


But just because a component defined min and max, doesnt mean they need to change to `Required`, they can stay as `SplitDefault`, its only the validation minimum/maximum that needs to change.

Also, if min **and** max are supplied to this function, only `max` will be enforced as it overrides the min one.

tracestep: I did it and now ci-custom check throws an error:
```
Unable to load component mcp9600.sensor:
  File "/home/runner/work/esphome/esphome/esphome/loader.py", line 165, in _lookup_module
Error:     module = importlib.import_module(f"esphome.components.{domain}")
  File "/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/importlib/__init__.py", line 127, in import_module
Error:     return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
Error:   File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
Error:   File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
Error:   File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/runner/work/esphome/esphome/esphome/components/mcp9600/sensor.py", line 60, in <module>
Error:     FINAL_VALIDATE_SCHEMA = i2c.final_validate_device_schema(
  File "/home/runner/work/esphome/esphome/esphome/components/i2c/__init__.py", line [14](https://github.com/esphome/esphome/actions/runs/4567168973/jobs/8060632518?pr=4614#step:10:15)9, in final_validate_device_schema
Error:     if CORE.using_esp_idf:
  File "/home/runner/work/esphome/esphome/esphome/core/__init__.py", line 6[15](https://github.com/esphome/esphome/actions/runs/4567168973/jobs/8060632518?pr=4614#step:10:16), in using_esp_idf
Error:     return self.target_framework == "esp-idf"
  File "/home/runner/work/esphome/esphome/esphome/core/__init__.py", line 607, in target_framework
Error:     return self.data[KEY_CORE][KEY_TARGET_FRAMEWORK]
KeyError: 'core'
CODEOWNERS file is not up to date.
Please run `script/build_codeowners.py`
Error: Process completed with exit code 1.
```
I see that `mcp9600` component has a blank `__init__.py`... maybe not importing `CORE`? It does impose validation of the I²C frequency.

tracestep: Well, I went ahead and implemented the timeout option for the Arduino framework as well, by moving the range checking to the C++ code. I could not find definitive information on the default timeout for ESP8266 or RP2040, so I made it such that the correspondent `set_timeout` function is only called if `timeout` option is present in the yaml code, thus leaving the defaults as they were and guaranteeing that any component that uses I²C will continue to work without setting the timeout option.

esphome/components/i2c/i2c_bus_arduino.cpp
```diff
@@ -41,6 +41,18 @@ void ArduinoI2CBus::setup() {
...
 #endif
+  if (timeout_ > 0) {  // if timeout specified in yaml
+#if defined(USE_ESP32)
+    // https://github.com/espressif/arduino-esp32/blob/master/libraries/Wire/src/Wire.cpp
+    wire_->setTimeOut(timeout_ / 1000);  // unit: ms
+#elif defined(USE_ESP8266)
+    // https://github.com/esp8266/Arduino/blob/master/libraries/Wire/Wire.h
+    wire_->setClockStretchLimit(timeout_);  // unit: us
+#elif defined(USE_RP2040)
+    // https://github.com/arduino/ArduinoCore-mbed/blob/master/libraries/Wire/Wire.h
```
jesserockz: ESPHome uses https://github.com/earlephilhower/arduino-pico/blob/master/libraries/Wire/src/Wire.h for Arduino, which implements `Stream` and uses

https://github.com/earlephilhower/ArduinoCore-API/blob/e37df85425e0ac020bfad226d927f9b00d2e0fb7/api/Stream.h#L68-L69
```c++
  void setTimeout(unsigned long timeout);  // sets maximum milliseconds to wait for stream data, default is 1 second
  unsigned long getTimeout(void) { return _timeout; }
```
and
https://github.com/earlephilhower/arduino-pico/blob/0963611fc701db920274112dde00f65e2997f763/libraries/Wire/src/Wire.cpp#L239
```c++
_buffLen = i2c_read_blocking_until(_i2c, address, _buff, quantity, !stopBit, make_timeout_time_ms(_timeout));

Comment (tracestep) [reply esphome/components/i2c/i2c_bus_arduino.cpp]: I have read conflicting reports on the use of `Stream::setTimeout` for this purpose (coping with clock stretching), but yeah, guess it can't hurt. New commit published.

**Comment by Szewcson** [comment]
I'm trying add support for GDK101 radiation sensor and encountered problem with i2c timeout. Workaround which works for me is adding delay of ~10ms before `uint8_t status = wire_->endTransmission(stop);` so i think this should be the same as increasing timeout. Unfortunately when I applied this PR and set maximal timeout (~13ms) I still getting a timeout error. I even tried setting maximal timeout and adding additional 5ms delay. I tried on efr32 both on arduino and esp-idf, and on efr8266 . On efr8266 don't know why I got no ack instead of timeout, but other behavior is the same. I also created PR for popper logging of timeout error https://github.com/esphome/esphome/pull/4697

**Comment by tracestep** [comment]
@Szewcson, if you are using `wire`, you are probably on the Arduino framework, not `esp-idf`. The maximum timeout of 13ms is only for the `esp-idf` framework. Arduino uses a default of 50ms for the timeout, which can be increased further. In my testings, these timeouts apply to clock stretching (see datasheet for your sensor to see by how much it can stretch the clock) within or right after a read / write. Maybe the reason for the timeouts is something else, probably related to the specific library or times between a command and a read.

**Comment by Szewcson** [comment]
Tried it on both platforms, and also on esp8266. This quick hack with delay before `Wire.endTransmission()` works, of course, only for arduino ( I don't have time to play with esp-idf - only checked that problem also occurs there). Unfortunately, [documentation ](url=https://www.radonshop.com/ftlab-gdk101-gamma-radiation-sensor-module-arduino)for that sensor module is far from being good. There is a sample code (for arduino) and not much more. The code looks really basic. 

```
void Gamma_Mod_Read(int cmd){
  /* Begin Write Sequence */
  Wire.beginTransmission(addr);
  Wire.write(cmd);
  Wire.endTransmission();
  /* End Write Sequence */
  delay(10);
  /* Begin Read Sequence */
  Wire.requestFrom(addr, 2);
  byte i = 0;
  while(Wire.available())
  {
    buffer[i] = Wire.read();
    i++;
  }
  /* End Read Sequence */
  /* View Results */
  Print_Result(cmd);
}
```

Initialization is just `Wire.begin();`

I previously ran it on Tasmota without any quirks to the driver, but they driver is more complicated. In the sample code, there is additional delay after `Wire.endTransmission()`, but in our driver delay before that function was a dirty workaround. If I have a time, and borrow a logic analyzer, I will try to look at the transmission. But first, I will try to use 50ms+ timeout on arduino.

File: esphome/components/i2c/__init__.py [hunk 5]
```diff
@@ -131,6 +140,20 @@ def final_validate_device_schema(
...
+            min_included=True,
+            msg=f"Component {name} requires a minimum timeout of {min_timeout} for the I2C bus",
+        )
+
+    if max_timeout is not None:
+        hub_schema[cv.Required(CONF_TIMEOUT)] = cv.Range(
+            max=cv.time_period(max_timeout),
+            max_included=True,
+            msg=f"Component {name} cannot be used with a timeout of over {max_timeout} for the I2C bus",
+        )
```
jesserockz: If you specify both min and max in the validate call, only max will take effect as it gets overriden in the second if.

jesserockz (changes requested):

esphome/components/i2c/__init__.py
tracestep: I apologize for the extended delay, but I was away and could not get back to the project for a while. I confess I am not familiar at all with the python side of ESPHome, and had simply used the frequency validation above those lines (already in the codebase) as a template. After getting to understand the code a bit, I saw they were also buggy.

I guess there are probably more elegant solutions, but I believe my latest commit solves the issue.

zagi988:
C'mon, please fix this... Either give us timeout or don't hardcode boot priorities so we can load i2c after some GPIO that would turn on the pull up line.

jesserockz (dismissed):

jesserockz (dismissed):

esphome/components/i2c/i2c_bus_arduino.cpp
```diff
@@ -52,6 +52,18 @@ void ArduinoI2CBus::set_pins_and_clock_() {
...
+  if (timeout_ > 0) {  // if timeout specified in yaml
+#if defined(USE_ESP32)
+    // https://github.com/espressif/arduino-esp32/blob/master/libraries/Wire/src/Wire.cpp
+    wire_->setTimeOut(timeout_ / 1000);  // unit: ms
+#elif defined(USE_ESP8266)
+    // https://github.com/esp8266/Arduino/blob/master/libraries/Wire/Wire.h
+    wire_->setClockStretchLimit(timeout_);  // unit: us
+#elif defined(USE_RP2040)
+    // https://github.com/earlephilhower/ArduinoCore-API/blob/e37df85425e0ac020bfad226d927f9b00d2e0fb7/api/Stream.h
+    wire_->SetTimeout(timeout_ / 1000);  // unit: ms
```
jesserockz:
```suggestion
    wire_->setTimeout(timeout_ / 1000);  // unit: ms
```

jesserockz (dismissed):

jesserockz (approved):

PR #4621, opened 2023-03-26
Add ability to clear the gatt cache

bdraco:

With BlueZ we can fully clear the cache when something goes wrong with the services, however since this is also a cache on the ESP32 we need to be able to clear the on device cache as well for the proxies since if something goes wrong with the service resolution it can cache the bad resolution on NVS forever.

Our current client implementation is limited to clearing the memory cache in Home Assistant https://github.com/home-assistant/core/blob/89355e087952417a6824507fd3b197f9d0520e19/homeassistant/components/esphome/bluetooth/client.py#L512

related issue https://github.com/esphome/issues/issues/4156

https://github.com/esphome/aioesphomeapi/pull/410

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/api/api_connection.cpp
+++ b/esphome/components/api/api_connection.cpp
@@ -953,7 +953,9 @@ DeviceInfoResponse APIConnection::device_info(const DeviceInfoRequest &msg) {
   resp.webserver_port = USE_WEBSERVER_PORT;
 #endif
 #ifdef USE_BLUETOOTH_PROXY
-  resp.bluetooth_proxy_version = bluetooth_proxy::global_bluetooth_proxy->has_active() ? 5 : 1;
+  resp.bluetooth_proxy_version = bluetooth_proxy::global_bluetooth_proxy->has_active()
+                                     ? bluetooth_proxy::ACTIVE_CONNECTIONS_VERSION
+                                     : bluetooth_proxy::PASSIVE_ONLY_VERSION;
 #endif
   return resp;
 }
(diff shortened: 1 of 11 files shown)


esphome/components/api/api_connection.cpp
```diff
@@ -953,7 +953,7 @@ DeviceInfoResponse APIConnection::device_info(const DeviceInfoRequest &msg) {
   resp.webserver_port = USE_WEBSERVER_PORT;
 #endif
 #ifdef USE_BLUETOOTH_PROXY
-  resp.bluetooth_proxy_version = bluetooth_proxy::global_bluetooth_proxy->has_active() ? 4 : 1;
+  resp.bluetooth_proxy_version = bluetooth_proxy::global_bluetooth_proxy->has_active() ? 5 : 1;
```
jesserockz: Maybe we should move these "magic" versions into a const inside the `bluetooth_proxy` namespace?

esphome/components/api/api_frame_helper.cpp
```diff
@@ -74,7 +74,7 @@ const char *api_error_to_str(APIError err) {
 
 #define HELPER_LOG(msg, ...) ESP_LOGVV(TAG, "%s: " msg, info_.c_str(), ##__VA_ARGS__)
 // uncomment to log raw packets
-//#define HELPER_LOG_PACKETS
+// #define HELPER_LOG_PACKETS
```
bdraco:
```suggestion
//#define HELPER_LOG_PACKETS
```

esphome/components/api/api_connection.cpp
bdraco: 157dcdcca2eada261c8b7d6f72320cab611904bb

jesserockz (approved):

PR #4636, opened 2023-03-30
Binary map bugfixes

kahrendt:

The binary sensor map component with SUM type outputs NAN if all the binary_sensory channels are off on boot. If one channel turns on and off, the sensor then outputs 0.0. This PR has the sensor output 0.0 on boot if all the channels are off when using the SUM type.
(This gives a different result for this particular case, but this PR seems to be the intended behavior. I am unsure if I should mark it as a breaking change because of this.)

The component has an implicit max of 64 binary_sensor channels, as a uint64_t stores each binary state in a bit. Therefore, the sensor.py configuration validation limits the number of binary_sensor channels to 64.

The current component logs at the debug level whenever the sensor updates. ESPHome already logs the state of the sensor whenever it is updated, so this is redundant. I removed two log output commands for regular sensor updates and changed one to log at the verbose level to describe why the sensor outputs NAN when using the GROUP type.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2795

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
switch:
  - platform: template
    name: "switch 1"
    id: switch_1
    optimistic: true
binary_sensor:
  - platform: template
    id: binary_sensor_1
    name: "binary sensor 1"
    lambda: |-
      return id(switch_1).state;
sensor:
  - platform: binary_sensor_map
    name: 'sum'
    type: SUM
    channels:
      - binary_sensor: binary_sensor_1
        value: 2
# the 'sum' sensor originally sent NAN on boot. If you toggle the switch on and off, then it would send 0.0. This PR sends 0.0 on boot.
```

## Checklist:

### DIFF
--- a/esphome/components/binary_sensor_map/binary_sensor_map.cpp
+++ b/esphome/components/binary_sensor_map/binary_sensor_map.cpp
@@ -30,7 +30,7 @@ void BinarySensorMap::process_group_() {
     if (bs.binary_sensor->state) {
       num_active_sensors++;
       total_current_value += bs.sensor_value;
-      mask |= 1 << i;
+      mask |= 1ULL << i;
     }
   }
   // check if the sensor map was touched
@@ -58,7 +58,7 @@ void BinarySensorMap::process_sum_() {
     auto bs = this->channels_[i];
     if (bs.binary_sensor->state) {
       total_current_value += bs.sensor_value;
-      mask |= 1 << i;
+      mask |= 1ULL << i;
     }
   }
 
--- a/esphome/components/binary_sensor_map/binary_sensor_map.cpp
+++ b/esphome/components/binary_sensor_map/binary_sensor_map.cpp
@@ -38,12 +38,11 @@ void BinarySensorMap::process_group_() {
     // did the bit_mask change or is it a new sensor touch
     if (this->last_mask_ != mask) {
       float publish_value = total_current_value / num_active_sensors;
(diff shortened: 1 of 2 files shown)


esphome/components/binary_sensor_map/binary_sensor_map.cpp
```diff
@@ -52,28 +51,22 @@ void BinarySensorMap::process_group_() {
...
-  // check all binary_sensors for its state. when active add its value to total_current_value.
-  // create a bitmask for the binary_sensor status on all channels
+  // - check all binary_sensor states
+  // - if active, add its value to total_current_value
+  // - creates a bitmask for the binary_sensor status on all channels
   for (size_t i = 0; i < this->channels_.size(); i++) {
     auto bs = this->channels_[i];
     if (bs.binary_sensor->state) {
       total_current_value += bs.sensor_value;
       mask |= 1 << i;
```
anatoly-savchenkov: I know you are not touching this code but there is a long standing bug here (and in process_group_()), code should be 
`mask |= 1ULL << i;`. Could you please address this problem as well.

jesserockz (approved):

PR #4641, opened 2023-03-31
SHT21 from Sensirion needs longer delay

martmaiste:
SHT21 from Sensirion does not work with delay 50 but works with 80. HTU21 works with both.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

## Checklist:


jesserockz (changes requested):
This should be changed to use `set_timeout` to run blocks of code asynchronously instead of blocking the loop

PR #4648, opened 2023-04-03
Add push to talk voice assistant

jesserockz:

Moving with the Year of the Voice in Home Assistant, this PR adds a few "linked" components.

- `i2s_audio` `microphone` for getting the audio data into ESPHome.
- `voice_assistant` asks Home Assistant to start a UDP server and then sends the received microphone data via UDP to that server.
- The changes to `media_player` are to accommodate the co-existence of the microphone and speaker on the same i2s pins, but only 1 can take control at a time.

## Related PRs
- https://github.com/esphome/aioesphomeapi/pull/412
- https://github.com/home-assistant/core/pull/90691

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2813

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

This example works on an m5stack atom echo
```yaml
i2s_audio:
  i2s_lrclk_pin: GPIO33
  i2s_bclk_pin: GPIO19

microphone:
  - platform: i2s_audio
    id: echo_microphone
    i2s_din_pin: GPIO23

voice_assistant:
  microphone: echo_microphone

binary_sensor:
  - platform: gpio
    pin:
      number: GPIO39
      inverted: true
    name: ${friendly_name} Button
    id: echo_button
    on_press:
      - voice_assistant.start:
    on_release:
      - voice_assistant.stop:

media_player:
  - platform: i2s_audio
    name: None
    id: echo_audio
    dac_type: external
    i2s_dout_pin: GPIO22
    mode: mono

```

## Checklist:


herostrat:
Wow, massive respect for this. Awesome work!

SeByDocKy:
WHoaahhh .....

esphome/components/i2s_audio/media_player/__init__.py
```diff
@@ -7,20 +7,18 @@
 from esphome.const import CONF_ID, CONF_MODE
 from esphome.core import CORE
 
+from .. import i2s_audio_ns
```
nielsnl68:
```suggestion
from .. import i2s_audio_ns, CONF_I2S_DOUT_PIN, CONF_I2S_BCLK_PIN, CONF_I2S_LRCLK_PIN 
```

nielsnl68 (changes requested):
I hope you will breakout the `I2S_Audio` part from the `media_player` in the future as well, like you do with the `microphone` and `voice_assistant`. That way we could update the RTTTL component etc. to output using I2S as well or create an other component that can use the audion out component.

something like:
```yaml
speaker:
  platform: i2s_audio
  id: my_speaker
  ....

media_player:
  output: my_speaker
  ...
```
so we can do something like
```yaml
rtttl:
  output: my_speaker
  ....
```
etc.

nagyrobi:
Whooa!

One suggestion though, maybe the naming and the approach could be a bit more universal. Instead of `voice_assistant` maybe `push_to_talk` could be more appropriate, as there could be other use cases of the UDP audio stream, not only voice assistants.

For example we could build a simple two-way intercom doorbell, where the core part is the same, it could just be a different UDP port with some other service forwarding the stream to the mobile app, moreover, could be in the future another ESPHome node listening with a speaker...

jesserockz:
I actually named it `push_to_talk` originally, but then realised the `push_to_talk` part really was just the binary sensor triggers telling the `voice_assistant` to start, so that is done in yaml, and we may look at hot-word detection in the future if possible, so that would be another thing that triggers the start of `voice_assistant`

I do agree with you though partly, but its easy to set up another component that has its own udp setup to send the data to a different server like an intercom/direct p2p audio etc. This component is specifically set up to transmit the data to HA via the random port it receives.

The voice_assistant part is quite small itself and any component can latch onto a `microphone` and request the data "stream" and do whatever it wants.

nagyrobi:
@jesserockz please don't forget about the `annouce` flag coming from HA when TTS is being played. This was implemented to differentiate between normal media playback (eg. music) vs announcements or signalizations, which have a higher priority. Eg. if speakers are powered off, `annouce` can turn them on. If it's playing some music, that volume can be dimmed for wile the announcement sounds. When announcement playback finishes, the flag (=binary sensor) should go false, allowing to power the speakers back off or restore music original volume.

Also see https://github.com/esphome/feature-requests/issues/1771

PR #4652, opened 2023-04-03
Throttle PN532 updates after bus read fail (fix esphome/issues#4389)

tracestep:

If bus read is failed (e.g.: module failed, connection problems, noise), updating PN532 component will block code execution for 1s on every update (due to esp-idf bug - see [https://github.com/espressif/esp-idf/issues/4999](https://github.com/espressif/esp-idf/issues/4999)). This PR works around that by throttling updates if that happens, which limits problems with other devices on the bus and other components.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes <[esphome/issues#4389](https://github.com/esphome/issues/issues/4389)>

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
i2c:
  scl: GPIO32
  sda: GPIO33
  scan: True
  frequency: 100kHz
  timeout: 2ms

pn532_i2c:
  update_interval: 250ms
  id: rfid
  on_tag:
    then:
      - binary_sensor.template.publish:
          id: tag_present
          state: ON
      - delay: 1s
      - binary_sensor.template.publish:
          id: tag_present
          state: OFF
```

## Checklist:


jesserockz (changes requested):
Not sure I like this, particularly because it just wont do anything for a whole minute with no logs to tell the user why. The one log with `throttling updates` would easily be missed.

Where does the 60 seconds even come from? Is that recommended by NXP?

tracestep:
One minute is actually just an arbitrary interval I felt could warrant a 1s pause in code execution. Ideally one would make a configuration option to choose whether or not to throttle and by how much, assuming the underlying issue (blocking for a whole second) can't be solved, but I'm not able to get into that for now. If not that, I don't know how to alert the user apart from the warning-level log (EDIT: by the way, the warning message in the log would repeat every minute).

I have a branch on my fork which implements asynchronous reads on PN532 (see https://github.com/esphome/issues/issues/4382 and https://github.com/tracestep/esphome/tree/pn532-async) which fixes that issue, but I could not find a way to do something similar in this one.

I did this because I am not 100% confident on the reliability of my PN532 board, and would not want the rest of the tasks (which are time-sensitive in my case) to completely break if it suddenly died or became unresponsive because of a power glitch or something else (I poll the RFID reader every 250 ms for quicker response on tags). But feel free to leave this in the back burner. Maybe mark the change as `breaking`?

tracestep:
I had not taken into account the possibility of a longer `update_interval_`... would something like the code below be acceptable?

```C++
void PN532::update() {
  if (!this->updates_enabled_)
    return;

  for (auto *obj : this->binary_sensors_)
    obj->on_scan_end();

  if (!this->status_has_warning()) {
    this->retries_ = 0;
    this->throttle_ = std::max((uint32_t)2e3, this->update_interval_);
  } else {
    if (millis() - this->last_update_ < this->throttle_)
      return;
    switch (this->retries_) {
      case 0:
        this->retries_++;
        this->throttle_ = std::max((uint32_t)5e3, this->update_interval_);
        break;
      case 1:
        this->retries_++;
        this->throttle_ = std::max((uint32_t)10e3, this->update_interval_);
        break;
      default:
        this->throttle_ = std::max((uint32_t)60e3, this->update_interval_);
    }
  }
  this->last_update_ = millis();
  if (!this->write_command_({
          PN532_COMMAND_INLISTPASSIVETARGET,
          0x01,  // max 1 card
          0x00,  // baud rate ISO14443A (106 kbit/s)
      })) {
    ESP_LOGW(TAG, "Requesting tag read failed. Throttling updates, next in %d s", this->throttle_ / 1000);
    this->status_set_warning();
    return;
  }
  this->status_clear_warning();
  this->requested_read_ = true;
}
```

HeMan (approved):

PR #4654, opened 2023-04-04
Add support for mdi images

guillempages:

This PR add support for specifying Material Design Icons directly in the YAML configuration, without the need of manually downloading them and converting them to PNG.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2837

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
image:
  - file: mdi:alert-outline
    resize: 80x80
    id: alert
```

## Checklist:

### DIFF
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -2,9 +2,7 @@ import logging
 
 from pathlib import Path
 import time
-from reportlab.graphics import renderPM
 import requests
-from svglib.svglib import svg2rlg
 
 from esphome import core
 from esphome.components import display, font
@@ -46,6 +44,33 @@ MDI_DOWNLOAD_TIMEOUT = 10  # seconds
 Image_ = display.display_ns.class_("Image")
 
 
+def validate_svglib_installed(value):
+    """Validate that svglib and reportlab are installed"""
+    try:
+        import reportlab
+        import svglib
+    except ImportError as err:
+        raise cv.Invalid(
+            "Please install the svglib and reportlab python packages to use this feature. "
+            "(pip install reportlab svglib)"
+        ) from err
+
+    if reportlab.__version__[0] < "3":
+        raise cv.Invalid(
+            "Please update your reportlab installation to at least 3.0.x. "
+            "(pip install -U reportlab)"
+        )
+    # svglib package does not offer the __version__ attribute
+    # Nevertheless to make linters happy, the package needs to be used here.
+    if svglib.__name__ != "svglib":
+        raise cv.Invalid(
+            "Please update your svglib installation to at least 1.0.x. "
+            "(pip install -U svglib)"
+        )
+
+    return value
+
+
 def validate_cross_dependencies(config):
     """
     Validate fields whose possible values depend on other fields.
@@ -80,7 +105,9 @@ IMAGE_SCHEMA = cv.Schema(
         {
             cv.Required(CONF_ID): cv.declare_id(Image_),
             cv.Exclusive(CONF_FILE, "input"): cv.file_,
-            cv.Exclusive(CONF_MDI, "input"): cv.string,
+            cv.Exclusive(CONF_MDI, "input"): cv.All(
+                cv.string, validate_svglib_installed
+            ),
             cv.Optional(CONF_RESIZE): cv.dimensions,
             # Not setting default here on purpose; the default depends on the source type
             # (file or mdi), and will be set in the "validate_cross_dependencies" validator.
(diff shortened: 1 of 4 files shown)


esphome/components/image/__init__.py
```diff
@@ -50,86 +59,180 @@
...
+        ]:
+            _LOGGER.warning(
+                'MDI image "%s" has type %s. This wastes program memory; please change it to (TRANSPARENT_)BINARY',
+                config[CONF_ID],
+                config[CONF_TYPE],
+            )
+        from pathlib import Path
+        import requests
+        from svglib.svglib import svg2rlg
+        from reportlab.graphics import renderPM
```
guillempages: Is this the right place to put the imports (like the PIL import in line 60)?
Or would it be better to put them at the top of the file?

esphome/components/image/__init__.py
```diff
@@ -22,21 +22,30 @@
...
 
+
 IMAGE_SCHEMA = cv.Schema(
     {
         cv.Required(CONF_ID): cv.declare_id(Image_),
-        cv.Required(CONF_FILE): cv.file_,
+        cv.Exclusive(CONF_FILE, "input"): cv.file_,
+        cv.Exclusive(CONF_MDI, "input"): cv.string,
         cv.Optional(CONF_RESIZE): cv.dimensions,
         cv.Optional(CONF_TYPE, default="BINARY"): cv.enum(IMAGE_TYPE, upper=True),
```
guillempages: For MDI images, "TRANSPARENT_BINARY" would be a better default; need to find a way of having sensible defaults depending on the rest of the parameters.

esphome/components/image/__init__.py
```diff
@@ -50,86 +59,180 @@
...
+                config[CONF_TYPE],
+            )
+        from pathlib import Path
+        import requests
+        from svglib.svglib import svg2rlg
+        from reportlab.graphics import renderPM
+
+        mdi = config[CONF_MDI]
+        url = f"https://raw.githubusercontent.com/Templarian/MaterialDesign/master/svg/{mdi}.svg"
+        _LOGGER.debug("Downloading %s MDI image from %s", mdi, url)
```
guillempages: Should this be "_LOGGER.info" instead of debug? Probably not, since in case of error the URL is printed anyway.

guillempages: Changed to info, since the image is now only downloaded once every 24 hours.

esphome/components/image/__init__.py
guillempages: Moved them to the top of the file

esphome/components/image/__init__.py
guillempages: Decided to force the type to be TRANSPARENT_BINARY for MDI images; everything else does not make any sense IMHO

guillempages:
The PR is made on top of #4600; once that one is merged I can rebase to make the modification smaller / the review easier

requirements_optional.txt
```diff
@@ -1,2 +1,4 @@
 pillow>4.0.0
+svglib>=1.5.0
+reportlab>=3.5.0
```
Unfocused: Will need to not just document this, but also add a runtime check & warning if they're missing (instead of unconditionally importing them).

At the moment, if these aren't installed, the `image`, `animation`, and `online_image` components effectively disappear - ESPHome exits with an error that those components can't be found if a device config tries to use them (not the most helpful message, admittedly). `animation` and `online_image` are affected despite not using these new libraries because they both import from the `image` component, which of course breaks them too.

See here for how the `font` component handles needing Pillow:
https://github.com/esphome/esphome/blob/4b664b6f09a6eef16e380a953973eeda59ea2f19/esphome/components/font/__init__.py#L62-L77

Ideally they'd be checked & imported only if they're actually going to be used - if someone never uses MDI, this shouldn't get in their way.

guillempages: I actually asked @jesserockz what's the official policy with imports, and the answer was "imports at the beginning unless there is a reason not to", so that's why I put that at the beginning.

That said, I agree that it's not nice to crash silently if those are not installed; especially when not used. So I'll revisit that.

esphome/components/image/__init__.py
```diff
@@ -78,8 +85,10 @@ def validate_cross_dependencies(config):
     have "use_transparency" set to True.
     Also set the default value for those kind of dependent fields.
     """
+    is_mdi = CONF_FILE in config and config[CONF_FILE][CONF_SOURCE] == SOURCE_MDI
```
guillempages, line 102: FYI, the online runtime images commit I am working on, also uses this method, but does not have `CONF_FILE`, it has `CONF_URL` instead. That's why I'm explicitly checking for `CONF_FILE in config` even though `CONF_FILE` is Required.

Unfocused: Really like this change to use `file: "mdi:whatever"` instead. 

I was debating whether to suggest something like that, since using `mdi` as the key name meant tightly coupling icon support to explicitly only MDI, which didn't feel future-proof.

jesserockz (approved):

PR #4671, opened 2023-04-09
fix compilation with latest esp-idf

Mic92:

When bumping esp-idf to 6.1.0 to fix some other issues, it became apparent that esphome uses wrong modifier for printf when printing u32/i32 integer.
This not completely fixes the build for it and I will make follow up pull requests for the other changes. This can be merged independently as it can also improve compatibility with other targets.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/api/api_connection.cpp
+++ b/esphome/components/api/api_connection.cpp
@@ -1,5 +1,6 @@
 #include "api_connection.h"
 #include <cerrno>
+#include <cinttypes>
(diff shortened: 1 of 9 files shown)


jesserockz (approved):
Thanks

PR #4675, opened 2023-04-11
Add dashboard API to get firmware binaries

kuba2k2:

Dashboard API improvements for https://github.com/esphome/dashboard/pull/469. See that PR for description of changes.

## Types of changes

- [x] Other

## Checklist:


jesserockz (changes requested):
I dont think the same endpoint should be used for getting the list of available files per platform, it should have its own endpoint

kuba2k2:
Okay, should be better now. I don't even know why did I put that in the same endpoint. I'll update the dashboard in a moment.

kuba2k2:
I updated esphome and dashboard to current upstream branch, and updated the list endpoint.

Also, I added a Popen()-based process spawning method, so that the dashboard can be tested on Windows. Nothing changes for Linux/Docker users, it only adds Windows compatibility (which would raise a Python exception prior to that).

Everything was tested locally for logs, building and downloading on ESP32.

esphome/__main__.py
jesserockz, line 1: Please make these changes in their own PR

esphome/components/esp32/__init__.py
```diff
@@ -80,6 +80,23 @@ def get_esp32_variant(core_obj=None):
     return (core_obj or CORE).data[KEY_ESP32][KEY_VARIANT]
 
 
+def get_download_types(storage_json=None):
```
jesserockz: This should be required (or it will crash the below code)
```suggestion
def get_download_types(storage_json):
```

esphome/components/esp8266/__init__.py
```diff
@@ -50,6 +50,17 @@ def set_core_data(config):
     return config
 
 
+def get_download_types(storage_json=None):
```
jesserockz:
```suggestion
def get_download_types(storage_json):
```

esphome/components/rp2040/__init__.py
```diff
@@ -42,6 +42,17 @@ def set_core_data(config):
     return config
 
 
+def get_download_types(storage_json=None):
```
jesserockz:
```suggestion
def get_download_types(storage_json):
```

esphome/dashboard/dashboard.py
```diff
@@ -481,40 +522,61 @@ def post(self):
...
+        compressed = self.get_argument("compressed", "0") == "1"
 
-        else:
+        storage_path = ext_storage_path(settings.config_dir, configuration)
+        storage_json = StorageJSON.load(storage_path)
+        if storage_json is None:
+            self.send_error(404)
+            return
+
+        file_name = self.get_argument("file", "firmware.bin")
```
jesserockz: Please also check the `type` as a fallback so we can deprecate that param over time

esphome/dashboard/dashboard.py
```diff
@@ -202,7 +203,11 @@ class EsphomeCommandWebSocket(tornado.websocket.WebSocketHandler):
     def __init__(self, application, request, **kwargs):
         super().__init__(application, request, **kwargs)
         self._proc = None
+        self._queue = None
         self._is_closed = False
+        # Windows doesn't support non-blocking pipes,
+        # use Popen() with a reading thread instead
+        self._use_popen = os.name == "nt"
```
jesserockz: Please move any dashboard fixes for windows into their own PR

jesserockz (changes requested):

esphome/dashboard/dashboard.py
kuba2k2: Should I also worry about not specifying `type=` at all (and falling back to default `firmware.bin`)? That will not work for RP2040, as `firmware.uf2` is used instead. The type/file parameter should be mandatory, and AFAIK has been passed by the dashboard already at all times.

jesserockz: Making them mandatory is fine. The download endpoint has been documented to be used externally recently for a tasmota ota update to ESPHome so I don't want to break that.

jesserockz (dismissed):

jesserockz (approved):

PR #4690, opened 2023-04-15
Custom mac addresses

jfparis:
This commit adds support for custom mac addresses previously implanted with "espefuse.py burn_custom_mac"

## What does this implement/fix?

ESP32 allows user to burn a custom MAC address via espefuse.py
https://docs.espressif.com/projects/esptool/en/latest/esp32/espefuse/burn-custom-mac-cmd.html
This is especially useful when a batch of chips comes out with identical MAC or a blank (00:..:00) MAC
This PR support such MAC

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/wifi/wifi_component_esp32_arduino.cpp
+++ b/esphome/components/wifi/wifi_component_esp32_arduino.cpp
@@ -32,6 +32,9 @@ bool WiFiComponent::wifi_mode_(optional<bool> sta, optional<bool> ap) {
   bool current_ap = current_mode & 0b10;
   bool enable_sta = sta.value_or(current_sta);
   bool enable_ap = ap.value_or(current_ap);
+  uint8_t mac[6];
+  get_mac_address_raw(mac);
+  set_mac_address(mac);
   if (current_sta == enable_sta && current_ap == enable_ap)
     return true;
 
--- a/esphome/core/helpers.cpp
+++ b/esphome/core/helpers.cpp
@@ -443,8 +443,7 @@ void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parame
   // without doing the CRC check.
   esp_efuse_read_field_blob(ESP_EFUSE_MAC_FACTORY, mac, 48);
 #else
-  if (esp_efuse_mac_get_custom(mac) == ESP_OK) {
-  } else
+  if (esp_efuse_mac_get_custom(mac) != ESP_OK)
     esp_efuse_mac_get_default(mac);
 #endif
 #elif defined(USE_ESP8266)
--- a/esphome/core/helpers.cpp
+++ b/esphome/core/helpers.cpp
@@ -436,16 +436,17 @@ bool HighFrequencyLoopRequester::is_high_frequency() { return num_requests > 0; 
 
 void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parameter)
 #if defined(USE_ESP32)
+  if (esp_efuse_mac_get_custom(mac) != ESP_OK) {
 #if defined(USE_ESP32_IGNORE_EFUSE_MAC_CRC)
-  // On some devices, the MAC address that is burnt into EFuse does not
-  // match the CRC that goes along with it. For those devices, this
-  // work-around reads and uses the MAC address as-is from EFuse,
-  // without doing the CRC check.
-  esp_efuse_read_field_blob(ESP_EFUSE_MAC_FACTORY, mac, 48);
+    // On some devices, the MAC address that is burnt into EFuse does not
+    // match the CRC that goes along with it. For those devices, this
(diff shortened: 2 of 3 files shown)


mmakaay (approved):
LGTM

I tested this code successfully on my Xiaomi bedside lamp without the last patch.
I will re-test with the latest code too.

**Update**: tested the code on my lamp and it I've seen no problems with it.

mmakaay:
I did some more testing, now with a NodeMCU board. It uses the standard MAC address for both ESP-IDF and Arduino. When I burn a custom MAC onto the EFUSE, then this custom MAC is used. So functionally all looks good.

There's one thing that might need some attention. Without a custom MAC burnt into the EFUSE, many messages are logged about the failing read of the custom MAC (to the serial console). These look like:
```
esp-idf:000]: E (64) system_api: Base MAC address from BLK3 of EFUSE version error, version = 0
```
While the errors are not harmful (version = 0 simply indicates that there is no custom MAC available), these errors might throw off users that are for example investigating WiFi issues. It would be  better if these errors would no be shown repeatedly. Especially since they will be logged for the default configuration of an ESP32 device.

Possible options that come to mind here:
- Use a static variable inside `get_mac_address_raw()` to keep track if the first call to `esp_efuse_mac_get_custom()` returned an error. If it did return an error, then the next call to  `get_mac_address_raw()` should not try calling it again.
- Create an advanced configuration option (at the same level as `advanced: ignore_efuse_mac_crc`) that tells the code that it should try to read a custom MAC address from the device. When a user enables it without having burnt a custom MAC to EFUSE, the errors are a good indicator that something fishy is going on with the custom MAC.

mmakaay:
Added the [breaking change] label, to match the breaking change checkbox in the PR message. I was about to suggest that this would not be considered breaking, but I thought a bit about it and the unexpected behaviour could be that a device changes its MAC address after flashing (when a custom address was burnt to EFUSE before).

This might be a valid reason to go for the second option that I proposed in my previous post: create a config option to indicate that a custom MAC address must be read from EFUSE. When doing that, this PR would not be breaking anymore and at the same time it would suppress the unwanted error logging.

jfparis:
I can easily do option 1 (Actually I have it running and was about to commit/ push)
Option 2 is a bit more tricky. Unlike the CRC toggle for `ignore_efuse_mac_crc`, this is something that needs to be applied to both to the arduino and the espressif framework

esphome/core/helpers.cpp
```diff
@@ -436,15 +436,20 @@ bool HighFrequencyLoopRequester::is_high_frequency() { return num_requests > 0;
 
 void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parameter)
 #if defined(USE_ESP32)
+  static bool custom_mac_not_set =
+      false;  // avoid repeated attempt to obtain custom macs to avoid spurious error messages in the log
+  if (!custom_mac_not_set && esp_efuse_mac_get_custom(mac) != ESP_OK) {
+    custom_mac_not_set = true;
```
mmakaay: For its purpose, the boolean name might better be  something like `suppress_custom_mac_address_check`, since that is closer to describing what effect the boolean has. With the current name, the boolean is lying at first when a custom mac  is set, while the boolean claims that it is not. Might be confusing for other coders (of even you :-) reading through this in the future.

mmakaay (changes requested):
The spurious logging fix is okay.
Just a small issue regarding the naming of the helper boolean.

esphome/core/helpers.cpp
jfparis: Thank you for the review and comments. This has been fixed

mmakaay:
I'm happy with the current iteration of the code.
Thanks for the modifications 👍

esphome/components/wifi/wifi_component_esp32_arduino.cpp
```diff
@@ -32,6 +32,9 @@ bool WiFiComponent::wifi_mode_(optional<bool> sta, optional<bool> ap) {
   bool current_ap = current_mode & 0b10;
   bool enable_sta = sta.value_or(current_sta);
   bool enable_ap = ap.value_or(current_ap);
+  uint8_t mac[6];
+  get_mac_address_raw(mac);
+  set_mac_address(mac);
```
jesserockz, line 37: Is it really necessary to get and set the mac every time the wifi mode is changed?

jesserockz (commented):
Most devices work perfectly fine without this code, Could we not still have the efuse setting of the mac behind an advanced flag, similar to how it is now for idf with the ignore?

jfparis:
They "work" but we ignore the mac address in the efuse which should probably be taken into account if it was set.

I will give this one last try and make it an optional feature (it will be for the espresif framework only).

PR #4691, opened 2023-04-15
Web_server: prevent bootloop from css_include/js_include files being too large, also add option to set new index.html page for local: true

snetsplit:

This fixes a bricking boot loop when js or css includes are too large. This pull also adds setting a custom local index html page with  a new option like:

```
 web_server:
  local: true
  port: 80
  local_page_include: "./index.html"
```

The behavior of the current release is the index html page is stored as a gzipped byte array statically, if config local: true. Presumably for speed and to save flash space. 

CSS, and JS includes, however, are stored as strings in RAM. The web_server loads at boot, and if either of these strings are too large, runs out of memory and crashes the esp system. This causes a boot loop.

This pull fixes that by instead reading CSS or JS includes, and then convert them to a gzipped byte array just like the site index page. Then cg.progmem_array is used to add them to c++ code as static PROGMEM constants where they can be accessed.  This saves flash memory and RAM memory and prevents the web_server component from crashing.

Further since this happens at compile time, it was simple to add an option for a custom local page include.

I'm unsure which file to update in the tests folder, but I will be very happy to if given guidance. Like wise this my first pull request and I hope everything is good, but please let me know if any changes need to happen. Thank you!

Edit: added tests to test3.yaml, and test7.yaml. Also in the test/components folder, added a web_server folder with test files. Hope this is what's being looked for.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes `[web_server bricks (boot loop) device if js_include is above 26KB(ish)](https://github.com/esphome/issues/issues/4303)`

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```
web_server:
  port: 80
  local: true
  custom_local_page: "./index.html"
  css_include: ./v2/www.css"
  js_include: "./v2/www.js"
  css_url: ""
  js_url: ""
```
## Checklist:


esphome/components/web_server/__init__.py
```diff
@@ -17,7 +18,27 @@
...
+
+    globals()["CONF_CSS_INCLUDE_ARRAY"] = CONF_CSS_INCLUDE_ARRAY
+    globals()["CONF_JS_INCLUDE_ARRAY"] = CONF_JS_INCLUDE_ARRAY
+    globals()["CONF_LOCAL_PAGE_INCLUDE"] = CONF_LOCAL_PAGE_INCLUDE
+    globals()["CONF_LOCAL_PAGE_INCLUDE_ARRAY"] = CONF_LOCAL_PAGE_INCLUDE_ARRAY
+except ImportError:
+    globals()["CONF_JS_INCLUDE_ARRAY"] = "js_include_array"
+    globals()["CONF_CSS_INCLUDE_ARRAY"] = "css_include_array"
+    globals()["CONF_LOCAL_PAGE_INCLUDE"] = "local_page_include"
+    globals()["CONF_LOCAL_PAGE_INCLUDE_ARRAY"] = "local_page_include_array"
```
jesserockz: There is really no reason to add these fields to the common const.py file as they wont be used by other components.

snetsplit: Thanks for the guidance! I've removed them const.py. Was trying to follow the existing patterns and wasn't sure if they needed to be in there for some reason.

jesserockz: If you think a CONST will be reused, then for sure add to const.py, otherwise its fine in the component code. It can always be moved later.

PR #4693, opened 2023-04-15
Allow partially looping animations

guillempages:
Add the possibility of specifying a "loop" in an animation; where the
requested frames (start - end) will be repeateadly shown for "count" times.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2847

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
animation:
  - id: fire
    file: "fire.gif"
    loop:
      start_frame: 4
      end_frame: 7
      repeat: 5
# When calling "next_frame()" on the animation, frames 4 to 7 will be repeated 5 times before moving on to frame 8.
```

## Checklist:

### DIFF
--- a/esphome/components/animation/__init__.py
+++ b/esphome/components/animation/__init__.py
@@ -6,13 +6,24 @@ import esphome.components.image as espImage
 from esphome.components.image import CONF_USE_TRANSPARENCY
 import esphome.config_validation as cv
 import esphome.codegen as cg
-from esphome.const import CONF_FILE, CONF_ID, CONF_RAW_DATA_ID, CONF_RESIZE, CONF_TYPE
+from esphome.const import (
+    CONF_FILE,
+    CONF_ID,
+    CONF_RAW_DATA_ID,
+    CONF_REPEAT,
+    CONF_RESIZE,
+    CONF_TYPE,
+)
 from esphome.core import CORE, HexInt
 
 _LOGGER = logging.getLogger(__name__)
 
 DEPENDENCIES = ["display"]
 MULTI_CONF = True
+
+CONF_LOOP = "loop"
+CONF_START_FRAME = "start_frame"
+CONF_END_FRAME = "end_frame"
 
 Animation_ = display.display_ns.class_("Animation", espImage.Image_)
 
@@ -48,6 +59,13 @@ ANIMATION_SCHEMA = cv.Schema(
             # Not setting default here on purpose; the default depends on the image type,
             # and thus will be set in the "validate_cross_dependencies" validator.
             cv.Optional(CONF_USE_TRANSPARENCY): cv.boolean,
+            cv.Optional(CONF_LOOP): cv.All(
+                {
+                    cv.Optional(CONF_START_FRAME, default=0): cv.positive_int,
+                    cv.Optional(CONF_END_FRAME): cv.positive_int,
+                    cv.Optional(CONF_REPEAT): cv.positive_int,
+                }
+            ),
             cv.GenerateID(CONF_RAW_DATA_ID): cv.declare_id(cg.uint8),
         },
         validate_cross_dependencies,
@@ -227,3 +245,8 @@ async def to_code(config):
         espImage.IMAGE_TYPE[config[CONF_TYPE]],
     )
     cg.add(var.set_transparency(transparent))
+    if CONF_LOOP in config:
+        start = config[CONF_LOOP][CONF_START_FRAME]
+        end = config[CONF_LOOP].get(CONF_END_FRAME, frames)
+        count = config[CONF_LOOP].get(CONF_REPEAT, -1)
+        cg.add(var.set_loop(start, end, count))
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -773,12 +773,31 @@ Color Animation::get_grayscale_pixel(int x, int y) const {
   return Color(gray, gray, gray, alpha);
 }
 Animation::Animation(const uint8_t *data_start, int width, int height, uint32_t animation_frame_count, ImageType type)
-    : Image(data_start, width, height, type), current_frame_(0), animation_frame_count_(animation_frame_count) {}
-int Animation::get_animation_frame_count() const { return this->animation_frame_count_; }
+    : Image(data_start, width, height, type),
+      current_frame_(0),
+      animation_frame_count_(animation_frame_count),
+      loop_start_frame_(0),
+      loop_end_frame_(animation_frame_count_),
+      loop_count_(0),
+      loop_current_iteration_(1) {}
+void Animation::set_loop(uint32_t start_frame, uint32_t end_frame, int count) {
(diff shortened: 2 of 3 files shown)


guillempages:
This PR is based on #4600 to avoid merge conflicts later on; once that one is merged, a simple rebase to "dev" should make this one much smaller.

esphome/components/animation/__init__.py
```diff
@@ -48,6 +59,13 @@ def validate_cross_dependencies(config):
             # Not setting default here on purpose; the default depends on the image type,
             # and thus will be set in the "validate_cross_dependencies" validator.
             cv.Optional(CONF_USE_TRANSPARENCY): cv.boolean,
+            cv.Optional(CONF_LOOP): cv.All(
+                {
+                    cv.Optional(CONF_START_FRAME, default=0): cv.int_,
+                    cv.Optional(CONF_END_FRAME): cv.Any(None, cv.int_),
+                    cv.Optional(CONF_REPEAT): cv.Any(None, cv.int_),
```
kroimon: Do you need `None` here at all if they are optional anyway?

esphome/components/display/display_buffer.cpp
```diff
@@ -773,12 +773,36 @@ Color Animation::get_grayscale_pixel(int x, int y) const {
...
 void Animation::next_frame() {
   this->current_frame_++;
+  if (loop_count_ && this->current_frame_ <= loop_end_frame_) {
+    if (this->loop_current_iteration_ < loop_count_ || loop_count_ < 0) {
+      if (this->current_frame_ >= loop_end_frame_) {
+        this->current_frame_ = loop_start_frame_;
+        this->loop_current_iteration_++;
+      }
+    }
+  }
```
kroimon: Your innermost `if` is effectively a `==` as you already checked for `<=` in the outermost `if`.
You can merge that into one or two `if`s:

```suggestion
  if (loop_count_ && this->current_frame_ == loop_end_frame_
      && (this->loop_current_iteration_ < loop_count_ || loop_count_ < 0)) {
    this->current_frame_ = loop_start_frame_;
    this->loop_current_iteration_++;
  }
```

kroimon (changes requested):

esphome/components/animation/__init__.py
kbx81: ^^^ second this -- I just tried locally and this seems to work whether the parameter is specified or not:

```suggestion
                    cv.Optional(CONF_END_FRAME): cv.int_,
                    cv.Optional(CONF_REPEAT): cv.int_,
```

guillempages: I followed the upstream voluptuous documentation, or some stack overflow code to detect whether a value was set or not, but you are right; that is superfluous. It can be done much more easily.

esphome/components/display/display_buffer.cpp
guillempages: Thanks for pointing this out. I spent so much time trying to make the loop work at all, that I did not see the redundancy anymore.

kroimon: Happens to all of us 😄 That's exactly what code reviews are for.

esphome/components/animation/__init__.py
kbx81: I should've mentioned this the first time, but `cv.positive_int` would probably be even better for these (start frame, end frame, repeat -- I'm assuming you'd never want to allow a value less than zero). 😄

esphome/components/animation/__init__.py
```diff
@@ -48,6 +59,13 @@ def validate_cross_dependencies(config):
             # Not setting default here on purpose; the default depends on the image type,
             # and thus will be set in the "validate_cross_dependencies" validator.
             cv.Optional(CONF_USE_TRANSPARENCY): cv.boolean,
+            cv.Optional(CONF_LOOP): cv.All(
+                {
+                    cv.Optional(CONF_START_FRAME, default=0): cv.int_,
+                    cv.Optional(CONF_END_FRAME): cv.int_,
+                    cv.Optional(CONF_REPEAT): cv.int_,
```
kbx81:
```suggestion
                    cv.Optional(CONF_START_FRAME, default=0): cv.positive_int,
                     cv.Optional(CONF_END_FRAME): cv.positive_int,
                     cv.Optional(CONF_REPEAT): cv.positive_int,
```

kbx81 (commented):
One more quick suggestion for validation -- I should've mentioned this before, sorry! I hope it makes sense. If not, feel free to discard.

esphome/components/animation/__init__.py
kbx81: Left a suggestion in case it's easier to just merge in the change that way. Sorry for not catching this the first time!

esphome/components/animation/__init__.py
guillempages: I'm using "-1" as the default for the repeat in the c++ part of the code, but you are right, the configuration itself should be positive only; default is handled by "None" here. 
Thanks!

guillempages:
@kbx81 I've implemented your suggestion, and while I was at it, I also changed the types in the c++ part whenever possible, to also use unsigned types (and thus avoid some unnecessary checks for >0 )

esphome/components/display/display_buffer.cpp
```diff
@@ -773,12 +773,29 @@ Color Animation::get_grayscale_pixel(int x, int y) const {
...
+      loop_start_frame_(0),
+      loop_end_frame_(animation_frame_count_),
+      loop_count_(0),
+      loop_current_iteration_(1) {}
+uint32_t Animation::get_animation_frame_count() const { return this->animation_frame_count_; }
 int Animation::get_current_frame() const { return this->current_frame_; }
+void Animation::set_loop(uint32_t start_frame, uint32_t end_frame, int count) {
+  loop_start_frame_ = std::min(start_frame, animation_frame_count_);
+  loop_end_frame_ = std::min(end_frame, animation_frame_count_);
+  loop_count_ = count;
```
kroimon, line 786: I just noticed that you should probably reset the current loop iteration when updating the parameters, for example via a lambda.
```suggestion
  if (loop_count_ != count) {
    this->loop_current_iteration_ = 1;
  }
  loop_count_ = count;
```

guillempages: The lambda thought is a valid one; I hadn't thought about that, I considered this setter method as "part of the constructor".

Nevertheless I would do that without the if; when the parameters are reset in a lambda, the iteration count should also restart.

Also, weren't you the one (rightly) warning me about useless loops? In your suggestion there is a useless assignment: if the ˋifˋ is not entered, there is no point in setting the ˋloop_countˋ to ˋcountˋ; they are already equal ;-)

kbx81:
I'm good with this -- any other feedback/changes? 😄

guillempages:
Not from my side 😉

PR #4701, opened 2023-04-17
[ili9xxx] Allow config of spi data rate.

Fabian-Schmidt:

Allow config of spi data rate for ili9xxx display.
For example WT32-SC01 can be run with 80MHz for faster processing. A full redraw can be reduced from ~215ms to ~180ms by increasing the data rate to 80MHz.
Without setting the data_rate the current default rate of 40MHz is used.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2874

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

- platform: ili9xxx
    model: ST7796
    cs_pin: 15
    dc_pin: 21
    reset_pin: 22
    data_rate: 80Mhz

```

## Checklist:


esphome/components/spi/__init__.py
```diff
@@ -16,6 +16,22 @@
...
+    20e6: SPIDataRate.DATA_RATE_20MHZ,
+    10e6: SPIDataRate.DATA_RATE_10MHZ,
+    5e6: SPIDataRate.DATA_RATE_5MHZ,
+    2e6: SPIDataRate.DATA_RATE_2MHZ,
+    1e6: SPIDataRate.DATA_RATE_1MHZ,
+    2e5: SPIDataRate.DATA_RATE_200KHZ,
+    75e3: SPIDataRate.DATA_RATE_75KHZ,
+    1e3: SPIDataRate.DATA_RATE_1KHZ,
+}
+
```
jesserockz:
```suggestion
SPI_DATA_RATE_OPTIONS = {
    80e6: SPIDataRate.DATA_RATE_80MHZ,
    40e6: SPIDataRate.DATA_RATE_40MHZ,
    20e6: SPIDataRate.DATA_RATE_20MHZ,
    10e6: SPIDataRate.DATA_RATE_10MHZ,
    5e6: SPIDataRate.DATA_RATE_5MHZ,
    2e6: SPIDataRate.DATA_RATE_2MHZ,
    1e6: SPIDataRate.DATA_RATE_1MHZ,
    2e5: SPIDataRate.DATA_RATE_200KHZ,
    75e3: SPIDataRate.DATA_RATE_75KHZ,
    1e3: SPIDataRate.DATA_RATE_1KHZ,
}
SPI_DATA_RATE_SCHEMA = cv.All(cv.frequency, cv.enum(SPI_DATA_RATE_OPTIONS))

```

esphome/components/ili9xxx/display.py
```diff
@@ -95,6 +96,7 @@ def _validate(config):
             cv.Optional(CONF_COLOR_PALETTE_IMAGES, default=[]): cv.ensure_list(
                 cv.file_
             ),
+            cv.Optional(CONF_DATA_RATE, default="40MHz"): spi.SPIDataRate_SCHEMA,
```
jesserockz:
```suggestion
            cv.Optional(CONF_DATA_RATE, default="40MHz"): spi.SPI_DATA_RATE_SCHEMA,
```

esphome/components/ili9xxx/display.py
```diff
@@ -173,3 +175,6 @@ def load_image(filename):
     if rhs is not None:
         prog_arr = cg.progmem_array(config[CONF_RAW_DATA_ID], rhs)
         cg.add(var.set_palette(prog_arr))
+
+    spi_data_rate = spi.SPIDataRate_OPTIONS[config[CONF_DATA_RATE]]
```
jesserockz:
```suggestion
    spi_data_rate = config[CONF_DATA_RATE]
```

jesserockz (dismissed):
The below changes should work and make a little simpler, please test though

Fabian-Schmidt:
Thanks updated code based on feedback.

esphome/components/ili9xxx/display.py
```diff
@@ -95,6 +96,7 @@ def _validate(config):
             cv.Optional(CONF_COLOR_PALETTE_IMAGES, default=[]): cv.ensure_list(
                 cv.file_
             ),
+            cv.Optional(CONF_DATA_RATE, default="40MHz"): spi.SPI_DATA_RATE_SCHEMA,
```
kbx81, line 102: I would strongly suggest using a safer default such as 10 MHz -- most people are connecting these with dupont wires and 40 MHz is likely to cause issues in this case. Speeds greater than 10 MHz are likely to require a custom PCB which is definitely not what most people are using. 😇

```suggestion
            cv.Optional(CONF_DATA_RATE, default="10MHz"): spi.SPI_DATA_RATE_SCHEMA,
```

kbx81 (changes requested):
Quick comment about the default speed -- that aside, LGTM 😄

esphome/components/ili9xxx/display.py
kbx81: ...related -- it would probably be worthwhile to add a note to the docs about this. I realize most people just want the fastest speed possible; however, there are electrical/physical limitations that software can't fix and people are likely to open issues based on this if it's not called out. 😇

nielsnl68: I disagree, the default was always 40MHz. Why should we change that to 10MHz now, just because we can set the value lower.

kroimon:
The use of the define makes the unlikely but theoretically possible use of two displays with different data rates impossible, right?
That should then probably be added to the docs as a hint.

nielsnl68:
> The use of the define makes the unlikely but theoretically possible use of two displays with different data rates impossible, right? That should then probably be added to the docs as a hint.
From my understanding is this no problem, the SPI bus switched speed based on the connected device.

esphome/components/ili9xxx/display.py
kbx81: I didn't realize the original default was 40 MHz -- but I still think this is a risky value to use by default given how many folks who use the platform in general tend to wire projects up. Frequencies higher than about ten MHz tend to be affected negatively by the extra capacitance that breadboards/perfboards/dev boards introduce; we have seen this with the Ethernet component, for example.

I only made the suggestion because I think we should use a safer default that should work for everyone as opposed to a default that may perform better but cause problems for folks with less-than-perfect wiring...regardless of what it was going before. Still, I'm ok with the more aggressive setting, but it might be worth adding a note to the docs to try slowing it down if you experience issues with the display.

nielsnl68: I understand what you are saying, and i would have agreed with it when this was a new component. Sadly it is not and we should not down talk the people that are using this component already and only want to have a faster display and not slower.

Also, in the last year or so that i have worked on the ili9xxx component and before i have never heard or seen issues that where related to the fast SPI communication.

So i belief for sure we can let it be set to 40MHz and when someone has indeed issues like you for shadow than we can now advice to set the `data rate` to a slower frequency.

kbx81: Sounds good, works for me! 😄

Fabian-Schmidt:
> The use of the define makes the unlikely but theoretically possible use of two displays with different data rates impossible, right? That should then probably be added to the docs as a hint.

Correct. Having two displays with different SPI speed should result in the last configured speed be used by all ili9xxx displays and a compile warning be generated.
If it is desired that multiple displays can be used with different SPI speed then the PR will be much larger / complex.

jesserockz:
I guess to make it per display we would need to template the ILI9xxx class with the speed. 
But honestly, is there anyone that uses multiple displays with a single ESP?

Fabian-Schmidt:
> I guess to make it per display we would need to template the ILI9xxx class with the speed. But honestly, is there anyone that uses multiple displays with a single ESP?

I have added a note to the docs about this behaviour.

kbx81 (approved):

kbx81:
@Fabian-Schmidt if you'll rebase this, I think we can merge. 🙃

kbx81 (approved):

esphome/components/ili9xxx/ili9xxx_display.h
```diff
@@ -15,10 +15,14 @@ enum ILI9XXXColorMode {
...
 
+#ifndef ILI9XXXDisplay_DATA_RATE
+#define ILI9XXXDisplay_DATA_RATE spi::DATA_RATE_40MHZ
+#endif  // ILI9XXXDisplay_DATA_RATE
+
 class ILI9XXXDisplay : public PollingComponent,
                        public display::DisplayBuffer,
                        public spi::SPIDevice<spi::BIT_ORDER_MSB_FIRST, spi::CLOCK_POLARITY_LOW,
-                                             spi::CLOCK_PHASE_LEADING, spi::DATA_RATE_40MHZ> {
+                                             spi::CLOCK_PHASE_LEADING, ILI9XXXDisplay_DATA_RATE> {
```
nielsnl68, line 25: Why the use of a `#DEFINE` for this, why not use a direct approach?

kbx81: True -- this could easily be an `enum`. Easy fix/patch, though 😇

Fabian-Schmidt: This is a C++ template class inheritance. So I must specify a solid type here or make `ILI9XXXDisplay` a template to propagate the parameter to the child classes and make all 20 child classes also templates so I can specify them from python. Given this would have been a larger change and the only benefit is that two displays could be connected with different SPI data rate. I opted for the easier solution.

PR #4704, opened 2023-04-17
[ili9xxx] Add background color to simplify partial refresh

Fabian-Schmidt:

Add new optional setting `background_color` for simpler partial refresh solution.

If `background_color` is set then during the `print` operation any pixel not used by the font will be set to background color. If you are writing exactly the same pixels the display driver (like ili9xxx) logic can / will ignore them. When writing a new value it will remove old pixels (avoid ghosting).

Without this change you must remember the last printed value, compare them and blank only the area where the value is printed.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2852

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml for WT32-SC01 with partial refresh.
    color:
      - id: bg_color
        red: 0%
        green: 0%
        blue: 0%
      - id: font_color
        red: 100%
        green: 3%
        blue: 5%

    display:
      - platform: ili9xxx
        model: ST7796
        cs_pin: 15
        dc_pin: 21
        reset_pin: 22
        auto_clear_enabled: false // Must be set to false
        background_color: bg_color // New setting
        lambda: |-
          static bool first_draw {true};
          if(first_draw) {
            first_draw = false;
            // On first set screen blank.
            it.fill(Color::BLACK);
          }
          // As the first two lines don't change they will not be redrawn. I don't need to remember last printed value.
          it.print(0, 0, id(font_name), id(font_color), TextAlign::TOP_LEFT, "Heading");
          it.printf(0, 40, id(font_name), id(font_color), TextAlign::TOP_LEFT, "%.2f", 123.0f);

          static uint32_t test;
          test++;
          if (test % 2 == 0) {
            it.printf(0, 240, id(font_name), id(font_color), TextAlign::TOP_LEFT, "%u     ", test); // additional spaces as numbers have different width.
          } else {
            it.printf(0, 240, id(font_name), id(font_color), TextAlign::TOP_LEFT, "%u g ", test);
          }
```

## Checklist:

### DIFF
--- a/esphome/components/display/__init__.py
+++ b/esphome/components/display/__init__.py
@@ -2,6 +2,7 @@ import esphome.codegen as cg
 import esphome.config_validation as cv
 from esphome import core, automation
 from esphome.automation import maybe_simple_id
+from esphome.components import color
 from esphome.const import (
     CONF_AUTO_CLEAR_ENABLED,
     CONF_ID,
@@ -37,6 +38,7 @@ DisplayOnPageChangeTrigger = display_ns.class_(
 )
 
 CONF_ON_PAGE_CHANGE = "on_page_change"
+CONF_BACKGROUND_COLOR = "background_color"
 
 DISPLAY_ROTATIONS = {
     0: display_ns.DISPLAY_ROTATION_0_DEGREES,
@@ -56,6 +58,7 @@ def validate_rotation(value):
 BASIC_DISPLAY_SCHEMA = cv.Schema(
     {
         cv.Optional(CONF_LAMBDA): cv.lambda_,
+        cv.Optional(CONF_BACKGROUND_COLOR): cv.use_id(color),
     }
 )
 
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -137,6 +137,7 @@ int DisplayBuffer::get_height() {
   }
 }
 void DisplayBuffer::set_rotation(DisplayRotation rotation) { this->rotation_ = rotation; }
+void DisplayBuffer::set_background_color(Color val) { this->background_color_ = val; }
 void HOT DisplayBuffer::draw_pixel_at(int x, int y, Color color) {
   if (!this->get_clipping().inside(x, y))
     return;  // NOLINT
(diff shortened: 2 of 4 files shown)


Unfocused:
I don't think this solves the problem. Plus I think you might have misunderstood how partial updates work.

> Without this change you must remember the last printed value, compare them and blank only the area where the value is printed.

With this PR, you still need to do this. Your comment in the example code actually explains why:

```cpp
it.printf(0, 240, id(font_name), id(font_color), TextAlign::TOP_LEFT, "%u     ", test); // additional spaces as numbers have different width.
```

Adding those spaces at the end of the text is just *guessing* that the space needed to draw that text will be large enough to cover the previously drawn text. Due to variable-width characters, the only way to know how much space will be needed to draw any text is to use `it.get_text_bounds()`. Also due to variable-width characters, it's unlikely that adding space characters will result in it exactly matching a specific size. It'll either be too small and leave ghosted pixels, or be too large and draw over pixels it shouldn't be drawing over.

Here's an example: Assume you're using a some font at a size where the 'a' character takes up 13 pixels horizontally, 'b' takes up 12 pixels, 'c' takes up 11 pixels, and a space takes up 9 pixels. And you have an area of the screen that you want to draw some text to - it's 27 pixels wide, and has a border drawn around it. If you draw the text "ab", it would draw into an area that's 25 pixels horizontally - it fits into the space on the screen you set aside for this text, no problem. Now say you want to replace the text "ab" with the text "c". Drawing "c" will take up only 11 pixels - you'll have 14 pixels ghosted from the previous text. If you append one space character, it'll take up 20 pixels - leaving 5 pixels ghosted. If you append two spaces, it'll take up 29 pixels - that's drawn over all of the old "ab" text. But it's also drawn over 4 pixels wide of an area that shouldn't be drawn over, inadvertently drawing over part of the border you had around that area (and everything else in that area).

> Without this change you must remember the last printed value, compare them and blank only the area where the value is printed.

Thankfully, you don't have to do it this way anymore - since the new [screen clipping functions](https://esphome.io/components/display/index.html#screen-clipping) were added in ESPHome 2023.2.0, it's now both easier and safer. Now you can solve it this way:

1. Determine the largest area you want some text to take up
2. Use `it.start_clipping()` to clip to that area
3. Use `it.fill()` to overwrite all pixels in that area & remove previous contents
4. Draw the text
5. Use `it.end_clipping()` to reset the clipping area

I think you might have also misunderstood how partial updates work. Where you say:
> As the first two lines don't change they will not be redrawn

The display driver may or may not send those pixels to the screen (depends on the display driver), but the pixels will always still need to be calculated regardless. So that text will always be redrawn to the display buffer. If you want that text to only be drawn once then you need to only do it when `first_draw == true`

Your example code adapted to using clipping & with the first two lines of text only drawing once, would be:
```yaml
color:
  - id: bg_color
    red: 0%
    green: 0%
    blue: 0%
  - id: font_color
    red: 100%
    green: 3%
    blue: 5%

display:
  - platform: ili9xxx
    model: ST7796
    cs_pin: 15
    dc_pin: 21
    reset_pin: 22
    auto_clear_enabled: false
    lambda: |-
      static bool first_draw {true};
      if(first_draw) {
        first_draw = false;
        // On first set screen blank. 
        // Using clear() instead of fill()
        it.clear();
        it.print(0, 0, id(font_name), id(font_color), TextAlign::TOP_LEFT, "Heading");
        it.printf(0, 40, id(font_name), id(font_color), TextAlign::TOP_LEFT, "%.2f", 123.0f);
      }

      static uint32_t test;
      test++;
      it.start_clipping(0, 240, 100, 50);
      it.fill(id(bg_color));
      if (test % 2 == 0) {
        it.printf(0, 0, id(font_name), id(font_color), TextAlign::TOP_LEFT, "%u", test);
      } else {
        it.printf(0, 0, id(font_name), id(font_color), TextAlign::TOP_LEFT, "%u g", test);
      }
      it.end_clipping();
```


The drawing functions could be made to automatically do steps 2, 3, 4, and 5 - which I think is in the same spirit of what you intended here. They'd just need to be able to take a new parameter to specify the clipping dimensions.

However... there are currently 3 different text drawing functions (`print()`, `printf()`, and `strftime()`). Each of those functions currently has 4 variations each that accept different parameters, to allow some parameters to be optional (like the `align` parameter). That's a total of 12 text drawing functions. If a clipping parameter were to be added, it would *also* have to be optional (otherwise it would break compatibility) - which would mean adding one more variation to each of the existing functions, and making a total of 24 text drawing functions. IMO, that would make the API too messy.

Fabian-Schmidt:
Thanks for the detailed answer.

The issue using your approach is for the `ili9xxx` display component. It only uses a single frame buffer. It calculates the area to update as soon as a pixel color is changed. So by using `it.fill()` it has identified that the pixel value has changed and a redraw is necessary for this pixel area. Even if afterwards the same pixel value is set again. It has already marked the area as redraw is necessary.

I completely agree that it is not an ideal solution.
But I was looking for a simple one. A solution were the `lambda` does not get too long or complicated.

Using `start_clipping` for every single print is cumbersome. Would it not be nicer to have an additional parameter for print with max width / length? (Internally it might use the clipping feature)

Unfocused:
> So by using `it.fill()` it has identified that the pixel value has changed and a redraw is necessary for this pixel area. Even if afterwards the same pixel value is set again. It has already marked the area as redraw is necessary.

Ah, I see. I don't think any of the other display drivers mark areas as dirty like that. I do think this needs a generic solution that would work with all `DisplayBuffer`-based display drivers, rather than solving this issue differently for each display driver (IMO they already behave too differently in some cases).

I was sure I saw some PR that refactored display buffers so each display driver didn't need to implement this itself (each in different ways), but I can't find it now.


> Using `start_clipping` for every single print is cumbersome. Would it not be nicer to have an additional parameter for print with max width / length? (Internally it might use the clipping feature)

Agreed. I described that possibility at the end of my comment, along with the downside. I wish the API allowed extending it easier (it's so much easier to do this cleanly in dynamic languages). Chaining calls like how the Light component does things with `LightCall` and `perform()`, or passing in an options object, etc. 

Maybe that's a conversation that needs to happen somewhere more visible. Especially as it seems there's already different people working on replacing the drawing API in completely different ways (which is going to be even more of a mess if display drivers keep solving the same issues in different ways).

As for clipping, I did wonder if a RAII-style scoped guard would help, so at the very least you'd never have to manually call `end_clipping()`. Something like:

```cpp
it.clear();
{
  DisplayClippingGuard clipper(10, 10, 50, 100);
  it.printf(...);
}
```

As soon as `clipper` goes out of scope, it's destructor gets called, which would call `end_clipping()`. So it would be equivilent to:

```cpp
it.clear();
{
  it.start_clipping(10, 10, 50, 100); // DisplayClippingGuard constructor
  it.printf(...);
  it.end_clipping(); // DisplayClippingGuard destructor
}
```

ESPHome uses this type of thing internally (eg, `WarnIfComponentBlockingGuard`), but I don't think there's currently anything like it used in lambdas. It would help, but I'd also worry it would just confuse people not used to C++.

Fabian-Schmidt:
I will reevaluate this PR as you are correct not every display driver must solve the problem differently.

Part of the issue might also that the `DisplayBuffer` only offers a single FrameBuffer so a partial refresh is always depending on the implementation of the display driver.

As for cleaner implementation of printing maybe something like:

```C++
it.clipped(10, 10, 50, 100).printf(...);
```

Makes it clear that the clipped only applies to one statement and avoids a lot of methods.

PR #4706, opened 2023-04-18
Add ethernet powerdown (fixes esphome/issues#4420)

tracestep:

Power down ethernet PHY on ethernet enabled ESP32 boards to save power on deep sleep.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes <[esphome/issues#4420](https://github.com/esphome/issues/issues/4420)>

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
ethernet:
  id: eth
  type: LAN8720
  mdc_pin: GPIO23
  mdio_pin: GPIO18
  clk_mode: GPIO0_IN
  phy_addr: 1
  power_pin: GPIO16

deep_sleep:
  id: sleeper
  touch_wakeup: false
  esp32_ext1_wakeup:
    pins: [GPIO39, GPIO35, GPIO14]
    mode: ANY_HIGH

button:
  - id: button_sleep
    platform: template
    name: Sleep
    icon: mdi:sleep
    on_press:
      then:
        - deep_sleep.enter:
            id: sleeper
```

## Checklist:


esphome/components/ethernet/ethernet_component.h
```diff
@@ -82,6 +84,7 @@ class EthernetComponent : public Component {
   uint32_t connect_begin_;
   esp_netif_t *eth_netif_{nullptr};
   esp_eth_handle_t eth_handle_;
+  esp_eth_phy_t *phy_;
```
jesserockz:
```suggestion
  esp_eth_phy_t *phy_{nullptr};
```

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -356,6 +358,21 @@ std::string EthernetComponent::get_use_address() const {
 
 void EthernetComponent::set_use_address(const std::string &use_address) { this->use_address_ = use_address; }
 
+bool EthernetComponent::powerdown() {
```
jesserockz, line 360: This should just be in `on_shutdown` instead of making another function for it.
This should never be called during normal runtime anyway as it will mess things up.
```suggestion
void EthernetComponent::on_shutdown() {
```

esphome/components/ethernet/ethernet_component.h
```diff
@@ -58,6 +59,7 @@ class EthernetComponent : public Component {
   network::IPAddress get_ip_address();
   std::string get_use_address() const;
   void set_use_address(const std::string &use_address);
+  bool powerdown();
```
jesserockz, line 62:
```suggestion
```

esphome/components/ethernet/ethernet_component.h
```diff
@@ -45,6 +45,7 @@ class EthernetComponent : public Component {
   void dump_config() override;
   float get_setup_priority() const override;
   bool can_proceed() override;
+  void on_shutdown() override { powerdown(); }
```
jesserockz, line 48:
```suggestion
  void on_shutdown() override;
```

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -356,6 +358,21 @@ std::string EthernetComponent::get_use_address() const {
...
+    ESP_LOGE(TAG, "Ethernet PHY not assigned");
+    return false;
+  }
+  global_eth_component->connected_ = false;
+  global_eth_component->started_ = false;
+  if (phy_->pwrctl(phy_, false) != ESP_OK) {
+    ESP_LOGE(TAG, "Error powering down ethernet PHY");
+    return false;
+  }
+  return true;
```
jesserockz:
```suggestion
  if (this->phy_ == nullptr) {
    ESP_LOGE(TAG, "Ethernet PHY not assigned");
    return;
  }
  this->connected_ = false;
  this->started_ = false;
  if (this->phy_->pwrctl(this->phy_, false) != ESP_OK) {
    ESP_LOGE(TAG, "Error powering down ethernet PHY");
  }
```

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -84,7 +86,7 @@ void EthernetComponent::setup() {
     }
   }
 
-  esp_eth_config_t eth_config = ETH_DEFAULT_CONFIG(mac, phy);
+  esp_eth_config_t eth_config = ETH_DEFAULT_CONFIG(mac, phy_);
```
jesserockz:
```suggestion
  esp_eth_config_t eth_config = ETH_DEFAULT_CONFIG(mac, this->phy_);
```

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -52,30 +54,30 @@ void EthernetComponent::setup() {
 
   esp_eth_mac_t *mac = esp_eth_mac_new_esp32(&mac_config);
 
-  esp_eth_phy_t *phy;
+  phy_ = nullptr;
```
jesserockz:
```suggestion
```

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -52,30 +54,30 @@ void EthernetComponent::setup() {
...
   switch (this->type_) {
     case ETHERNET_TYPE_LAN8720: {
-      phy = esp_eth_phy_new_lan87xx(&phy_config);
+      phy_ = esp_eth_phy_new_lan87xx(&phy_config);
```
jesserockz: Same for all of these below too
```suggestion
      this->phy_ = esp_eth_phy_new_lan87xx(&phy_config);
```

jesserockz (changes requested):

esphome/components/ethernet/ethernet_component.h
tracestep: I missed that, I had followed the original code that did not initialize `phy` as a local variable, but I can see how important that is for a class member,

esphome/components/ethernet/ethernet_component.cpp
tracestep: I did it this way because there might eventually be different / additional things to do on shutdown events before or after the powerdown method (which really should only send the command to the PHY controller), like do something about the network stack. I my use case there is a need to issue `powerdown()` in some situations to save a bit of power without going into full deep sleep; the user code can reboot the processor, go into deep sleep, or call setup() again if needed. I really do not see any harm in having a separate event handler `on_shutdown()`  and a potentially user-callable / inheritable method `powerdown()`. People who go messing around the C++ API should be able to assess the consequences of calling `powerdown()`.

tracestep: There also might be a need to check whether the PHY really went into powerdown or some error happened, in which case we could try again, alert the user or whatever.

EDIT: I just checked, and there are no ill-effects in calling `powerdown()` twice (once when you can test if it was successful and again when it gets called by `on_shutdown()`.

jesserockz: Ok, this can stay as is then

jesserockz (approved):

PR #4555, opened 2023-03-13
Add option flip_x

rafal83:
# What does this implement/fix?

Add option to flip display on x axis

## Types of changes

- [*] New feature (non-breaking change which adds functionality)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2746

## Test Environment

- [ ] ESP32

## Example entry for `config.yaml`:
```yaml
# Example config.yaml
display:
  - platform: max7219digit
    cs_pin: GPIO19
    rotate_chip: 180
    reverse_enable: true
    flip_x: true
    update_interval: 100ms
    num_chips: 9
    lambda: |-
      it.scroll(false);
      it.strftime(0, 0, id(digit_font), "%H:%M:%S %d-%m", id(hass_time).now());
```

## Checklist:
  - [*] The code change is tested and works locally.

If user exposed functionality or configuration variables are added/changed:
  - [*] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4720, opened 2023-04-20
Revert "Add option flip_x (#4555)"

### LINKED ISSUE #4432: [MAX7219digit] flip_x commit breaks reversed display
### The problem

The last [commit ](https://github.com/esphome/esphome/pull/4555) from @rafal83 breaks rotated and reversed displays. My clock is installed upside down so i used to rotate & flip it and it worked fine:
```
    rotate_chip: 180
    reverse_enable: true
```
But with that commit the display is completely broken(see details)

2023.04

Home Assistant Add-on

2023.4.3

ESP32

### Board

_No response_

MAX7219digit

### Example YAML snippet

```yaml
display:
  - platform: max7219digit
    cs_pin: GPIO25
    num_chips: 4
    intensity: 0
    rotate_chip: 180
    reverse_enable: false
    flip_x: false
    lambda: |-
      it.print(0, 5, id(digit_font), TextAlign::CENTER_LEFT, "1234567890");

```
<details>

image

</details>

==========================================================================

    reverse_enable: false
    flip_x: true

<details>

image

</details>

==========================================================================

    reverse_enable: true
    flip_x: true

With reverse enable the text is even duplicated from both directions
<details>

image

</details>

==========================================================================

    reverse_enable: true
    flip_x: false

<details>

image

</details>

_No response_

### Additional information

_No response_

voed:
This reverts commit a31fb3c. 

## Types of changes

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4432 and https://github.com/esphome/issues/issues/4437

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation :** https://github.com/esphome/esphome-docs/pull/2863

## Test Environment

- [x] ESP32

## Checklist:

### DIFF
--- a/esphome/components/max7219digit/display.py
+++ b/esphome/components/max7219digit/display.py
@@ -7,7 +7,6 @@ CODEOWNERS = ["@rspaargaren"]
 DEPENDENCIES = ["spi"]
 
 CONF_ROTATE_CHIP = "rotate_chip"
-CONF_FLIP_X = "flip_x"
 CONF_SCROLL_SPEED = "scroll_speed"
 CONF_SCROLL_DWELL = "scroll_dwell"
 CONF_SCROLL_DELAY = "scroll_delay"
@@ -68,7 +67,6 @@ CONFIG_SCHEMA = (
                 CONF_SCROLL_DWELL, default="1000ms"
             ): cv.positive_time_period_milliseconds,
             cv.Optional(CONF_REVERSE_ENABLE, default=False): cv.boolean,
-            cv.Optional(CONF_FLIP_X, default=False): cv.boolean,
         }
     )
     .extend(cv.polling_component_schema("500ms"))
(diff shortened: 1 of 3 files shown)


jesserockz:
I don't understand how deleting this feature is making any difference for you. If its set to `false` in config, it is basically the exact same code as if the feature was never there...

voed:
> I don't understand how deleting this feature is making any difference for you. If its set to `false` in config, it is basically the exact same code as if the feature was never there...

See related issues for the detailed explanation. It is not.

voed:
Also i do not really understand what it should do. Isn't it the same as 'reverse' feature we already have?

ssieb:
There was a bug in the python code, fix is in https://github.com/esphome/esphome/pull/4727
This feature flips each block individually, not the whole display like reverse does.

PR #4724, opened 2023-04-21
I2c scan recovery reset fix

gcopeland:

This fixes i2c's bus recovery logic, especially when combined with scan, to avoid WDT reboot.

Current i2c bus recovery logic has an arbitrary (based on comment) 500ms delay when attempting to recover the i2c bus when the clock line is held low. This both violates the esphome mandate of no delays longer than 10ms, but also triggers the WDT, ensuring recovery is never possible.

This bug was discovered because of an issue with a wiring harness which resulted in the i2c bus clock line held low. While this does prevent recovery of the bus, it also prevents continued operation of the device without the associated i2c bus as it simply boot loops because of the WDT.

This fix does several things. One, it uses the related cycles count as a basis for retry timing. Two, attempts many more retries for recovery. Three, arbitrary reduces the net retry duration below that of the esphome delay mandate. Four, resets the WDT while within the clock recovery logic. 

Technically the reset is not required (untested) as is, however, the pre-existing comment says the 500ms is arbitrary, is rather large, and I do not know if larger delays are required for universal compatibility with i2c devices which may be slow to respond. Which means either the cycle count and/or the delay can be changed well above the 10ms esphome mandate if required, to universally support all i2c devices which may hold low the bus longer than the current values (I defer to i2c expert here). Regardless, removal of the WDT logic is on the table. As is modification to the recovery delay/cycle count.

## Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)
- [ ] Other

## Test Environment

- [x] ESP32
- [] ESP32 IDF
- [x] ESP8266 *compiled
- [ ] RP2040

Tested against ESP32, arduino.
Compiled against ESP8266, arduino.
Code also modifies IDF variant. Not compiled. Not tested.

i2c:
  # For debugging, set scan to true. For production
  # set scan to 'false'.
  scan: true
  frequency: 10kHz

## Checklist:

Tested against ESP32, arduino.
Compiled against ESP8266, arduino.
Code also modifies IDF variant. Not compiled. Not tested.

Testing on ESP32, with the clock line pulled low confirms the follow resolution:
1. board no longer reboots because of WDT - allowing progression of scan and recovery attempts
2. bus scan now correctly continues and fails, logging the failures of address detection
3. the i2c bus fails and associated devices on the bus are ignored
4. esphome continues running on the node with the associated failed bus/devices.

### DIFF
--- a/esphome/components/i2c/i2c_bus_arduino.cpp
+++ b/esphome/components/i2c/i2c_bus_arduino.cpp
@@ -240,7 +240,7 @@ void ArduinoI2CBus::recover_() {
       esp_task_wdt_reset();
       delayMicroseconds(half_period_usec * 2);
 #elif  defined(USE_ESP8266)
-      EspClass::ESP.wdtFeed();
+      EspClass::wdtFeed();
       delayMicroseconds(half_period_usec * 2);
 #elif
       delay(25);
--- a/esphome/components/i2c/i2c_bus_arduino.cpp
+++ b/esphome/components/i2c/i2c_bus_arduino.cpp
@@ -239,7 +239,7 @@ void ArduinoI2CBus::recover_() {
 #if defined(USE_ESP32)
       esp_task_wdt_reset();
       delayMicroseconds(half_period_usec * 2);
-#elif  defined(USE_ESP8266)
+#elif defined(USE_ESP8266)
       EspClass::wdtFeed();
       delayMicroseconds(half_period_usec * 2);
 #elif
--- a/esphome/components/i2c/i2c_bus_arduino.cpp
+++ b/esphome/components/i2c/i2c_bus_arduino.cpp
@@ -234,7 +234,7 @@ void ArduinoI2CBus::recover_() {
     // No point in trying to recover the bus by forcing a uC reset. Bus
     // should recover in a few ms or less else not likely to recovery at
     // all.
-    auto wait = 200;
+    auto wait = 250;
     while (wait-- && digitalRead(scl_pin_) == LOW) {  // NOLINT
 #if defined(USE_ESP32) || defined(USE_ESP8266)
       esp_task_wdt_reset();
--- a/esphome/components/i2c/i2c_bus_arduino.cpp
+++ b/esphome/components/i2c/i2c_bus_arduino.cpp
@@ -5,9 +5,7 @@
 #include "esphome/core/helpers.h"
 #include <Arduino.h>
 #include <cstring>
-#if defined(USE_ESP32)
-#include <esp_task_wdt.h>
-#endif
+#include "esphome/core/application.h";
 
 namespace esphome {
 namespace i2c {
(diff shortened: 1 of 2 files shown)


esphome/components/i2c/i2c_bus_arduino.cpp
```diff
@@ -227,10 +230,21 @@ void ArduinoI2CBus::recover_() {
...
     while (wait-- && digitalRead(scl_pin_) == LOW) {  // NOLINT
+#if defined(USE_ESP32)
+      esp_task_wdt_reset();
+      delayMicroseconds(half_period_usec * 2);
+#elif defined(USE_ESP8266)
+      EspClass::wdtFeed();
+      delayMicroseconds(half_period_usec * 2);
+#elif
       delay(25);
+#endif
```
jesserockz: `App.feed_wdt()` exists as a layer so component code doesnt need to mostly check the chip.

```suggestion
      App.feed_wdt();
      delayMicroseconds(half_period_usec * 2);
```

esphome/components/i2c/i2c_bus_arduino.cpp
```diff
@@ -5,6 +5,9 @@
 #include "esphome/core/helpers.h"
 #include <Arduino.h>
 #include <cstring>
+#if defined(USE_ESP32)
+#include <esp_task_wdt.h>
+#endif
```
jesserockz:
```suggestion
#include "esphome/core/application.h";
```

esphome/components/i2c/i2c_bus_esp_idf.cpp
```diff
@@ -273,10 +274,14 @@ void IDFI2CBus::recover_() {
...
-    // time. We'll stick to 500ms here.
-    auto wait = 20;
+    // time. We yield and reset the WDT, so as to avoid triggering reset.
+    // No point in trying to recover the bus by forcing a uC reset. Bus
+    // should recover in a few ms or less else not likely to recovery at
+    // all.
+    auto wait = 250;
     while (wait-- && gpio_get_level(scl_pin) == 0) {
-      delay(25);
+      esp_task_wdt_reset();
```
jesserockz: I know this is idf only, but the function actually does a little bit more.
```suggestion
      App.feed_wdt();
```

esphome/components/i2c/i2c_bus_esp_idf.cpp
```diff
@@ -6,6 +6,7 @@
 #include "esphome/core/helpers.h"
 #include <cstring>
 #include <cinttypes>
+#include <esp_task_wdt.h>
```
jesserockz:
```suggestion
#include "esphome/core/application.h";
```

jesserockz (changes requested):

jesserockz (approved):

jesserockz:
Thanks for making the changes

PR #4726, opened 2023-04-22
Allow substitutions to be valid names

jgoguen:

When a config snippet has a substitution used for a name field, allow validation to treat the substitution token as valid.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] VS Code

```yaml
# Example config.yaml
# Open with VS Code with ESPHome.esphome-vscode 
# installed and set for local validation
esphome:
  name: ${name}
  build_path: build/${name}
  platform: ESP8266
```

## Checklist:

### DIFF
--- a/esphome/components/substitutions/__init__.py
+++ b/esphome/components/substitutions/__init__.py
@@ -1,18 +1,13 @@
 import logging
-import re
 
 import esphome.config_validation as cv
 from esphome import core
-from esphome.const import CONF_SUBSTITUTIONS
+from esphome.const import CONF_SUBSTITUTIONS, VALID_SUBSTITUTIONS_CHARACTERS
 from esphome.yaml_util import ESPHomeDataBase, make_data_base
 from esphome.config_helpers import merge_config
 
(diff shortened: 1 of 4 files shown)


ssieb:
Is this just to fix something in vscode?  Because that example config has no problem now.

jgoguen:
> Is this just to fix something in vscode? Because that example config has no problem now.

Yes, this is for VS Code, and anything else that may call `esphome.config_validation.valid_name()` or `esphome.config_validation.validate_id_name()` before substitutions are processed.

jgoguen:
@ssieb is this something ESPHome will consider?

jesserockz:
The extension will call the validation with a flag you can check with `if CORE.vscode:`

jgoguen:
> The extension will call the validation with a flag you can check with `if CORE.vscode:`

Updated code to only allow substitution placeholders when run with vscode

jesserockz (approved):

PR #4729, opened 2023-04-23
Make network::IPAddress ip address agnostic.

HeMan:

This makes ESPHome more ip address agnostic. It changes `network::IPAddress` to use `ip_addr_t` instead of `std::array<uint8_t, 4>` and removes lots of explicit and implicit casts to and from `unit32_t` and uses the built in Python library `ipaddress` for validating addresses.
It breaks (uncommon) way of representing ipaddresses as a list in yaml:

```yaml
# This still works
wifi:
  ssid: MyHomeNetwork
  password: VerySafePassword

  # Optional manual IP
  manual_ip:
    static_ip: 192.168.0.123
    gateway: 192.168.0.1
    subnet: 255.255.255.0
```
This doesn't work
```yaml
wifi:
  ssid: MyHomeNetwork
  password: VerySafePassword

  # Optional manual IP
  manual_ip:
  # list as multiple lines
    static_ip:
      - 192
      - 168
      - 0
      - 123
    # list on one line
    gateway: [192, 168, 0, 1]
    subnet: 255.255.255.0
```
## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/captive_portal/captive_portal.cpp
+++ b/esphome/components/captive_portal/captive_portal.cpp
@@ -48,7 +48,7 @@ void CaptivePortal::start() {
   this->dns_server_ = make_unique<DNSServer>();
   this->dns_server_->setErrorReplyCode(DNSReplyCode::NoError);
   network::IPAddress ip = wifi::global_wifi_component->wifi_soft_ap_ip();
-  this->dns_server_->start(53, "*", (uint32_t) ip);
+  this->dns_server_->start(53, "*", IPAddress(ip));
 #endif
 
   this->base_->get_server()->onNotFound([this](AsyncWebServerRequest *req) {
--- a/esphome/components/e131/e131_packet.cpp
+++ b/esphome/components/e131/e131_packet.cpp
@@ -67,8 +67,8 @@ bool E131Component::join_igmp_groups_() {
     if (!universe.second)
       continue;
 
-    ip4_addr_t multicast_addr = {static_cast<uint32_t>(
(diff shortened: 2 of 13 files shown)


HeMan:
This now works on all hardware I've tried and passes all esphome tests. It still needs some work.

esphome/config_validation.py
```diff
@@ -979,21 +979,8 @@ def ssid(value):
...
-    else:
-        raise Invalid("IPv4 address must consist of either string or integer list")
-    if len(parts) != 4:
-        raise Invalid("IPv4 address must consist of four point-separated integers")
-    parts_ = list(map(int, parts))
-    if not all(0 <= x < 256 for x in parts_):
-        raise Invalid("IPv4 address parts must be in range from 0 to 255")
-    return IPAddress(*parts_)
+def ipaddress(value):
+    return IPAddress(value)
```
jesserockz: This loses all of the nice validation errors and removes the ability for a list (I doubt its used much, but it would be a breaking-change nonetheless)

It really should be `cv.Any(cv.ipv4, cv.ipv6)` (Obviously you would need to create `def ipv6(value)`)

jesserockz (changes requested):

HeMan:
I've decided to change scope a bit. The discussion in this PR are no longer relevant, I will thus close this PR and open a new, with correct description.

PR #4731, opened 2023-04-23
Add support for ESP32-{S2,S3,C3} to debug component

### LINKED ISSUE #3433: Debug Component doesn't work on ESP32-S2 and ESP32-S3
### The problem

Compilation errors when trying to use the debug component on ESP32-S2 and ESP32-S3 devices.

2022.7.0-dev

Docker

N/A

ESP32

### Board

um_feathers3

debug

### Example YAML snippet

```yaml
esphome:
  name: $device_name

esp32:
  board: um_feathers3
  variant: esp32s3
  framework:
    type: arduino
    version: 2.0.3
    platform_version: 5.0.0

...

debug:
  update_interval: 5s

sensor:
  - platform: debug
    free:
      name: "${device_name}: Heap Free"
      entity_category: "diagnostic"
    # fragmentation:
    #   name: "Heap Fragmentation"
    block:
      name: "${device_name}: Heap Max Block"
      entity_category: "diagnostic"
    loop_time:
      name: "${device_name}: Loop Time"
      entity_category: "diagnostic"
```

```txt
Compilation error message:

      In file included from /config/.esphome/platformio/packages/framework-arduinoespressif32/tools/sdk/esp32s3/include/esp_rom/include/esp32/rom/rtc.h:18,
                       from src/esphome/components/debug/debug_component.cpp:15:
      /config/.esphome/platformio/packages/framework-arduinoespressif32/tools/sdk/esp32s3/include/esp_rom/include/esp32/rom/ets_sys.h:29:2: error: #error "This header should only be included when building for ESP32"
       #error "This header should only be included when building for ESP32"
        ^~~~~
```

### Additional information

Solution: debug_component.cpp needs updating to include chip-specific rtc.h file (i.e. esp32s2/rom/rtc.h, esp32s3/rom/rtc.h).
Additionally, reset_reason and wakeup_reason switch statements need updating with new values.

jayme-github:

Add support for ESP32-{S2,S3,C3} to debug component.

## Types of changes

Not sure if it should be classified as feature instead.

- [ ] Other

**Related issue or feature (if applicable):** fixes [ esphome/issues/3433
](https://github.com/esphome/issues/issues/3433)
## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:


nielsnl68:
Hi, can you make sure the test issues are resolved? Then we can do a review.

nielsnl68 (approved):
Looks good to me.

nielsnl68:
With the new options you added, should there be no new entries for the `sensor` and `text_sensor` platform components.

jayme-github:
> With the new options you added, should there be no new entries for the `sensor` and `text_sensor` platform components.

I'm not sure what you are referring to. I'm not really experienced with ESP things and/or this codebase so I might be missing something obvious. My understanding was that this only supports proper reset reasons (text sensor values) for those particular ESP32 variants.

kahrendt (commented):
I have tested this on an S3, and it works well! The only issue is that it states the chip is unknown. In the switch at line [120](https://github.com/esphome/esphome/pull/4731/files#diff-9a34821728cbc6e3297a26de0565002449519327fbb1622ca76fecc7e8649540R120), you should add the new cases (see [espressif's documentation](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/misc_system_api.html#_CPPv416esp_chip_model_t))

For example, I used this to get the proper chip model for my S3 device:
```
  switch (info.model) {
    case CHIP_ESP32:
      model = "ESP32";
      break;
    case CHIP_ESP32S3:
      model = "ESP32-S3";
      break;
    default:
      model = "UNKNOWN";
  }
```

jayme-github:
Good catch, thanks! I've added the models there as well.

jesserockz (approved):

esphome/components/debug/debug_component.cpp
```diff
@@ -113,6 +121,18 @@ void DebugComponent::dump_config() {
...
+      break;
+    case CHIP_ESP32S2:
+      model = "ESP32-S2";
+      break;
+    case CHIP_ESP32S3:
+      model = "ESP32-S3";
+      break;
+    case CHIP_ESP32H2:
+      model = "ESP32-H2";
+      break;
```
jesserockz: Actually one change to make here would be to wrap each of these in `#if defined(USE_ESP32_VARIANT_ESPxx)` like the includes as we know the chip cant be the others when compiling.

jesserockz (changes requested):

esphome/components/debug/debug_component.cpp
jayme-github: I agree. Sorry that it took me so long to come around again.

jesserockz (approved):

PR #4750, opened 2023-04-27
Enable IPv6 for ESP32 arduino

HeMan:

Initial ESP32 Arduino platform IPv6 support. This should take it to about the same level as the ESP IDF IPv6 support.

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
esphome:
  name: ipv6

esp32:
  board: esp32dev
  framework:
    type: arduino

wifi:
  ssid: wifissid
  password: !secret wifipwd

network:
  enable_ipv6: true

```

## Checklist:

### DIFF
--- a/esphome/components/network/__init__.py
+++ b/esphome/components/network/__init__.py
@@ -1,3 +1,4 @@
+from esphome.core import CORE
 import esphome.codegen as cg
 import esphome.config_validation as cv
 from esphome.components.esp32 import add_idf_sdkconfig_option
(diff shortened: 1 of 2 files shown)


esphome/components/network/__init__.py
```diff
@@ -14,16 +13,14 @@
...
 
 async def to_code(config):
-    if CONF_ENABLE_IPV6 in config:
-        add_idf_sdkconfig_option("CONFIG_LWIP_IPV6", config[CONF_ENABLE_IPV6])
-        add_idf_sdkconfig_option(
-            "CONFIG_LWIP_IPV6_AUTOCONFIG", config[CONF_ENABLE_IPV6]
-        )
+    if CONF_ENABLE_IPV6 in config and config[CONF_ENABLE_IPV6]:
+        cg.add_build_flag("-DCONFIG_LWIP_IPV6")
+        cg.add_build_flag("-DCONFIG_LWIP_IPV6_AUTOCONFIG")
```
jesserockz: When using IDF, these sdkconfig options default to true (at least the first one does), which is why I was explicitly setting them to the value set in config.

Would be good to continue to set the sdkconfig flags if using idf.

jesserockz (changes requested):

esphome/components/network/__init__.py
HeMan: Does my fix look good?

HeMan:
Replaced by #4865

PR #4758, opened 2023-04-29
play_folder bugfix and addition of play_mp3

llluis:

There's a bug in the current implementation of play_folder action.
Conditions are swapped for play_folder and play_large_folder.

Also, I added a new command to the esphome implementation supported by dfplayer: play_mp3

Both modifications allow users to easily manage files in the SD card without the need to use a global filenumber on disk needed by the standard play function. Files can be played referencing the name in the folder.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2879

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
on_...:
  then:
    - dfplayer.play_folder:
        folder: 2
        file: 1
```
```yaml
on_...:
  then:
    - dfplayer.play_mp3:
        file: 23
```

## Checklist:

### DIFF
--- a/esphome/components/dfplayer/dfplayer.cpp
+++ b/esphome/components/dfplayer/dfplayer.cpp
@@ -7,10 +7,10 @@ namespace dfplayer {
 static const char *const TAG = "dfplayer";
 
 void DFPlayer::play_folder(uint16_t folder, uint16_t file) {
-  if (folder < 100 && file < 256) {
+  if (folder <= 10 && file <= 1000) {
     this->ack_set_is_playing_ = true;
     this->send_cmd_(0x0F, (uint8_t) folder, (uint8_t) file);
-  } else if (folder <= 10 && file <= 1000) {
+  } else if (folder < 100 && file < 256) {
     this->ack_set_is_playing_ = true;
     this->send_cmd_(0x14, (((uint16_t) folder) << 12) | file);
   } else {
--- a/esphome/components/dfplayer/__init__.py
+++ b/esphome/components/dfplayer/__init__.py
@@ -40,6 +40,7 @@ DEVICE = {
 
 NextAction = dfplayer_ns.class_("NextAction", automation.Action)
 PreviousAction = dfplayer_ns.class_("PreviousAction", automation.Action)
+PlayMp3Action = dfplayer_ns.class_("PlayMp3Action", automation.Action)
 PlayFileAction = dfplayer_ns.class_("PlayFileAction", automation.Action)
 PlayFolderAction = dfplayer_ns.class_("PlayFolderAction", automation.Action)
 SetVolumeAction = dfplayer_ns.class_("SetVolumeAction", automation.Action)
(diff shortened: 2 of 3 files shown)


jesserockz (approved):

esphome/components/dfplayer/dfplayer.cpp
```diff
@@ -7,10 +7,10 @@ namespace dfplayer {
...
 void DFPlayer::play_folder(uint16_t folder, uint16_t file) {
-  if (folder < 100 && file < 256) {
+  if (folder <= 10 && file <= 1000) {
     this->ack_set_is_playing_ = true;
     this->send_cmd_(0x0F, (uint8_t) folder, (uint8_t) file);
-  } else if (folder <= 10 && file <= 1000) {
+  } else if (folder < 100 && file < 256) {
     this->ack_set_is_playing_ = true;
     this->send_cmd_(0x14, (((uint16_t) folder) << 12) | file);
   } else {
```
glmnet, line 16: This doesn't seem right, if file is checked to be <= 1000 then it shouldn't later be casted to uint8_t (8 bits max is 255)
Do you have a link where these functions (`0x0F 0x14`) are documented?

llluis: I have the page from the manufacturer: https://wiki.dfrobot.com/DFPlayer_Mini_SKU_DFR0299

image

The ``play_large_folder`` can only be found in their "official" library: https://github.com/DFRobot/DFRobotDFPlayerMini

https://github.com/DFRobot/DFRobotDFPlayerMini/blob/e15e9fcf8cfc91be356944e92b45a560bb926689/DFRobotDFPlayerMini.cpp#L344

https://github.com/DFRobot/DFRobotDFPlayerMini/blob/e15e9fcf8cfc91be356944e92b45a560bb926689/DFRobotDFPlayerMini.cpp#L368

Implementation in esphome added the condition to use the 2 functions automatically. The official lib exposes both without constraints.

I didn't touch the cast, just corrected the conditions.

glmnet: Yes, I wrote that code.
The thing is that you are checking file to be <= 1000 and then casting to 8 bit, so this would result in data loss, e.g. if file = 0, 256, 512, etc, they all result in 0.

I understand the command `0x0F` would allow folder and file to be 1 byte each, i.e. from 0 to 255. so folder and file can go from 0 to 255.
In the other hand, command `0x14` is shifting folder 12 bits so 4 bits for folder and 12 bits for file, so folder can go from 0 to 15 and file from 0 to 4095.

Unfortunately, documentation is quite disappointing. I might have not tested all the edge cases when I did this back then, nor I'm testing now. Just a heads up.

llluis: Thanks much for the feedback @glmnet (and for the implementation).
I'm actively using the lib and dfplayer in a personal project and will try to further test it. Agree on the lack of documentation.

My problem is that the play_large_folder is a looping command and I needed only one file play. 
Didn't want to mess with global file numbers in the card.

glmnet: What do you mean by loop command?

If the commands behave differently then maybe they should be two different actions or an argument in the action.

PR #4760, opened 2023-04-29
Add Micronova component

Jorre05:

New component to support pellet stoves with a Micronova board.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2890

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example config.yaml
```yaml
micronova:
  enable_rx_pin: 4
  update_interval: 20s

text_sensor:
  - platform: micronova
    stove_state:
      name: Stove status

number:
  - platform: micronova
    thermostat_temperature:
      name: Thermostat temperature
      step: 1
    power_level:
      name: Thermostat temperature
sensor:
  - platform: micronova
    room_temperature:
      name: Room temperature
    fumes_temperature:
      name: Fumes temperature
    water_temperature:
      name: Water temperature
    water_pressure:
      name: Water pressure
    stove_power:
      name: ${long_name} Power
    fan_speed:
      fan_rpm_offset: 240
      name: Fan RPM
    memory_address_sensor:
      memory_location: 0x20
      memory_address: 0x7d
      name: Custom Address sensor

switch:
   - platform: micronova
     stove_switch:
      name: Stove on/off switch

button:
  - platform: micronova
    custom_button:
      name: Custom button
      memory_location: 0xA0
      memory_address: 0x7D
      memory_data: 0x0F

```

## Checklist:


Jorre05:
Any pointers on how to implement the script/ci-suggest-changes suggestions in an automated fashion? Also, the "script/clang-tidy for ESP32 Arduino 2/4" output baffles me.

grahambrown11:
> Hey there @grahambrown11, mind taking a look at this pull request as it has been labeled with an integration (`alarm_control_panel`) you are listed as a [code owner](https://github.com/esphome/esphome/blob/dev/CODEOWNERS#L22) for? Thanks! (message by CodeOwnersMention)

What the heck... this seems you cherry picked commits instead of rebasing/merging from dev...

Jorre05:
> What the heck... this seems you cherry picked commits instead of rebasing/merging from dev...

Humm, the conflict was on test1.yaml, but it looks like I messed things up during a rebase. Don't really know how to fix this now...

grahambrown11:
> Humm, the conflict was on test1.yaml, but it looks like I messed things up during a rebase. Don't really know how to fix this now...

Sorry I just realised the tone of `What the heck...` is quite harsh 😳 
I was just surprised as to why I was getting tagged on an unrelated PR...
I'm also not 100% sure how to recover here... I would just copy my code out and start a fresh PR, maybe there are some git wizards out there who know better.

Sorry again if I came across negatively 😄

Jorre05:
> Sorry again if I came across negatively 😄

No hard feelings here. I've redone de merge/rebase. That seems to solve it, I think. But now the CI checks seem to fail, but it looks like a general problem, not specific to my code. I see the same errors in other recent pull requests.

Jorre05:
@jesserockz I overheard a discussion on Discord. As I understand, my code assumes that the top level types (sensor, button, switch and text_sensor) are present but does not check that. I never bumped into that problem because my devices all have those included. Any pointers to some docs on how to properly code this to be in line with the coding standards?

esphome/components/micronova/__init__.py
```diff
@@ -17,6 +17,7 @@
 CODEOWNERS = ["@jorre05"]
 
 DEPENDENCIES = ["uart"]
+AUTO_LOAD = ["sensor", "text_sensor", "button", "switch"]
```
jesserockz: Autoloading all the types is not an accepted solution. (Existing components are in the process of being migrated away from using this one at a time when time permits)

jesserockz (changes requested):

esphome/components/micronova/__init__.py
jesserockz: You should be either making an interface class to talk between the parent and children and/or using `#ifdef USE_SENSOR` etc to remove the code that is not available to use.

Jorre05: Thanks for the review, I'll look into that.

Jorre05: @jesserockz 

The challenge in this component is that the new value of a text_sensor influences a switch status. Also the switch's behavior depends on the current status of the text_sensor (e.g. you can't turn off the stove if it's in 'cooling down' mode). And the buttons, they need the current value of another sensor to work correct.

I can't seem to find an existing, up-to-date, component that implements this. From what I found, I believe I can use the existing teleinfo component as an example for the interface implementation you suggest. Correct?

Also, are you suggesting I split the configuration like this?
```yaml
micronova:
  enable_rx_pin: 7
  update_interval: 10s

sensor:
  platform: micronova
  room_temperature:
    memory_address: 0x01
    name: Ambient temperature
    id: ambient_temperature
  thermostat_temperature:
    name: Stove thermostat temperature

button:
  platform: micronova
  but_temp_up:
    memory_address: 0x21
    name: Increase thermostat

switch:
  platform: micronova
  stove_switch:
    memory_location: 0x80
    name: Switch stove on/off

```

Jorre05: @jesserockz I reworked the component. Can you review?

Jorre05:
Problems with IDF framework, converting back to draft.

esphome/components/micronova/button/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    CONF_MEMORY_LOCATION,
+    CONF_MEMORY_ADDRESS,
+    MICRONOVA_LISTENER_SCHEMA,
+    micronova_ns,
+)
+
+MicroNovaButton = micronova_ns.class_("MicroNovaButton", button.Button, cg.Component)
+
+CONF_BUT_TEMP_UP = "but_temp_up"
+CONF_BUT_TEMP_DOWN = "but_temp_down"
```
jesserockz: Config items should be well named, and since they are already buttons, they don't need the `but` prefix...
```suggestion
CONF_TEMPERATURE_UP = "temperature_up"
CONF_TEMPERATURE_DOWN = "temperature_down"
```

esphome/components/micronova/sensor/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+        ).extend(MICRONOVA_LISTENER_SCHEMA),
+        cv.Optional(CONF_STOVE_POWER): sensor.sensor_schema(
+            MicroNovaSensor,
+            state_class=STATE_CLASS_MEASUREMENT,
+            accuracy_decimals=0,
+        ).extend(MICRONOVA_LISTENER_SCHEMA),
+        cv.Optional(CONF_FAN_SPEED): sensor.sensor_schema(
+            MicroNovaSensor,
+            state_class=STATE_CLASS_MEASUREMENT,
+            unit_of_measurement=UNIT_RPM,
```
jesserockz: This already exists in const.py
```suggestion
            unit_of_measurement=UNIT_REVOLUTIONS_PER_MINUTE,
```

esphome/components/micronova/switch/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    MicroNova,
+    MicroNovaFunctions,
+    CONF_MICRONOVA_ID,
+    CONF_MEMORY_LOCATION,
+    CONF_MEMORY_ADDRESS,
+    MICRONOVA_LISTENER_SCHEMA,
+    micronova_ns,
+)
+
+CONF_STOVE_SWITCH = "stove_switch"
```
jesserockz: This doesnt need a suffix, its already under `switch`
```suggestion
CONF_STOVE = "stove"
```

jesserockz (changes requested):

esphome/components/micronova/sensor/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+            unit_of_measurement=UNIT_CELSIUS,
+            device_class=DEVICE_CLASS_TEMPERATURE,
+            state_class=STATE_CLASS_MEASUREMENT,
+            accuracy_decimals=1,
+        ).extend(MICRONOVA_LISTENER_SCHEMA),
+        cv.Optional(CONF_STOVE_POWER): sensor.sensor_schema(
+            MicroNovaSensor,
+            state_class=STATE_CLASS_MEASUREMENT,
+            accuracy_decimals=0,
+        ).extend(MICRONOVA_LISTENER_SCHEMA),
```
jesserockz: Should this sensor have device_class POWER and UoM Watts?

Jorre05: No. The stove output power is a number ranging from 0 tot 6 in my case. 6 meaning 'full throttle', 0 is off. device_class POWER could be considered.

esphome/components/micronova/micronova.cpp
```diff
@@ -0,0 +1,113 @@
...
+  uint16_t checksum = 0;
+
+  write_data[0] = location;
+  write_data[1] = address;
+  write_data[2] = data;
+
+  checksum = ((uint16_t) write_data[0] + (uint16_t) write_data[1] + (uint16_t) write_data[2]) & 0xFF;
+  write_data[3] = checksum;
+
+  ESP_LOGD(TAG, "Write 4 bytes [%02X,%02X,%02X,%02X]", write_data[0], write_data[1], write_data[2], write_data[3]);
```
jesserockz:
```suggestion
  ESP_LOGV(TAG, "Write 4 bytes [%02X,%02X,%02X,%02X]", write_data[0], write_data[1], write_data[2], write_data[3]);
```

esphome/components/micronova/micronova.cpp
```diff
@@ -0,0 +1,113 @@
...
+  this->flush();
+  this->enable_rx_pin_->digital_write(false);
+  // Give the stove some time to reply
+  delay(STOVE_REPLY_DELAY);  // NOLINT
+
+  this->read_byte(&reply_data[0]);
+  this->read_byte(&reply_data[1]);
+
+  this->enable_rx_pin_->digital_write(true);
+  ESP_LOGD(TAG, "First 2 bytes from [%02X:%02X] [%02X,%02X]", write_data[0], write_data[1], reply_data[0],
```
jesserockz:
```suggestion
  ESP_LOGV(TAG, "First 2 bytes from [%02X:%02X] [%02X,%02X]", write_data[0], write_data[1], reply_data[0],
```

esphome/components/micronova/micronova.cpp
```diff
@@ -0,0 +1,113 @@
...
+    ESP_LOGW(TAG, "Reading excess byte 0x%02X", trash_rx);
+  }
+
+  this->enable_rx_pin_->digital_write(true);
+  this->write_byte(addr);
+  this->write_byte(reg);
+  this->flush();
+  this->enable_rx_pin_->digital_write(false);
+  // Give the stove some time to reply
+  delay(STOVE_REPLY_DELAY);  // NOLINT
```
jesserockz: This 60ms delay is going to cause this component to log warnings every time. See https://github.com/esphome/issues/issues/4717

If possible, the code should be made asyncronous if it has to wait a long time before continuing to execute

esphome/components/micronova/micronova.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  this->read_byte(&data[0]);
+  this->read_byte(&data[1]);
+
+  this->enable_rx_pin_->digital_write(true);
+  ESP_LOGD(TAG, "First 2 bytes from [0x%02X:0x%02X] [0x%02X,0x%02X] dec: [%d,%d]", addr, reg, data[0], data[1], data[0],
```
jesserockz:
```suggestion
  ESP_LOGV(TAG, "First 2 bytes from [0x%02X:0x%02X] [0x%02X,0x%02X] dec: [%d,%d]", addr, reg, data[0], data[1], data[0],
```

esphome/components/micronova/sensor/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+                cg.add(sens.set_function(MicroNovaFunctions.STOVE_FUNCTION_FAN_SPEED))
+                cg.add(sens.set_fan_speed_offset(conf[CONF_FAN_RPM_OFFSET]))
+            if key == CONF_MEMORY_ADDRESS_SENSOR:
+                cg.add(sens.set_memory_location(conf.get(CONF_MEMORY_LOCATION, 0x00)))
+                cg.add(sens.set_memory_address(conf.get(CONF_MEMORY_ADDRESS, 0x00)))
+                cg.add(
+                    sens.set_function(
+                        MicroNovaFunctions.STOVE_FUNCTION_MEMORY_ADDRESS_SENSOR
+                    )
+                )
```
jesserockz: Honestly, if there is an `if` block for each type here, theres not much point in using a `for` loop over the keys. You might as well do it the same way as every other component:

```py
if room_temperature_config := config.get(CONF_ROOM_TEMPERATURE):
  room_temperature_sensor = ...
  
if thermostat_temperature_config := config.get(...):
  ...
```

jesserockz (changes requested):

esphome/components/micronova/micronova.cpp
Jorre05: Done

esphome/components/micronova/micronova.cpp
Jorre05: done

esphome/components/micronova/micronova.cpp
Jorre05: done

esphome/components/micronova/sensor/__init__.py
Jorre05: Done, on all __init__'s

esphome/components/micronova/micronova.cpp
Jorre05: This would be a hard nut to crack. The uart to the stove is a single line serial. You either transmit or receive. Moreover the protocol style request/reply based. One can easily mess things up by interrupting the stoves reply, a reply that came late for example. I'll give this some thought.

Jorre05: Aha, there is a yield() function. I'll explore that.

Jorre05: Ok, I've removed the delay() and split the code to process the request/reply asynchronously. But still, reading two bytes at baud 1200 (higher baud rate is not possible) takes about 50ms. I don't think I can make it faster than this. But the processing time in loop() went from 160ms to 50ms.

esphome/components/micronova/micronova.h
```diff
@@ -0,0 +1,164 @@
...
+  uint8_t memory_location_ = 0;
+  uint8_t memory_address_ = 0;
+};
+
+class MicroNovaSensorListener : public MicroNovaBaseListener {
+ public:
+  MicroNovaSensorListener() {}
+  MicroNovaSensorListener(MicroNova *m) : MicroNovaBaseListener(m) {}
+  virtual void request_value_from_stove();
+  virtual void process_value_from_stove(int value_from_stove);
```
jesserockz: Should these be changed to enforce implementation in the final class?
```suggestion
  virtual void request_value_from_stove() = 0;
  virtual void process_value_from_stove(int value_from_stove) = 0;
```

esphome/components/micronova/micronova.h
```diff
@@ -0,0 +1,164 @@
...
+// Main component class
+class MicroNova : public PollingComponent, public uart::UARTDevice {
+ public:
+  MicroNova() {}
+
+  void setup() override;
+  void loop() override;
+  void update() override;
+  void dump_config() override;
+  void register_micronova_listener(MicroNovaSensorListener *l) { micronova_listeners_.push_back(l); }
```
jesserockz: We like to always prefix member variables with `this->` so it is always clear

Please make sure to update it everywhere.
```suggestion
  void register_micronova_listener(MicroNovaSensorListener *l) { this->micronova_listeners_.push_back(l); }
```

esphome/components/micronova/micronova.h
```diff
@@ -0,0 +1,164 @@
...
+  void write_address(uint8_t location, uint8_t address, uint8_t data);
+  int read_stove_reply();
+
+  void set_enable_rx_pin(GPIOPin *enable_rx_pin) { this->enable_rx_pin_ = enable_rx_pin; }
+
+  void set_current_stove_state(uint8_t s) { current_stove_state_ = s; }
+  uint8_t get_current_stove_state() { return current_stove_state_; }
+
+  void set_stove(MicroNovaSwitchListener *s) { stove_switch_ = s; }
+  MicroNovaSwitchListener *get_stove_switch() { return stove_switch_; }
```
jesserockz:
```suggestion
  void set_current_stove_state(uint8_t s) { this->current_stove_state_ = s; }
  uint8_t get_current_stove_state() { return this->current_stove_state_; }

  void set_stove(MicroNovaSwitchListener *s) { this->stove_switch_ = s; }
  MicroNovaSwitchListener *get_stove_switch() { return this->stove_switch_; }
```

esphome/components/micronova/number/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+            max_value=40,
+            step=thermostat_temperature_config.get(CONF_STEP),
+        )
+        cg.add(numb.set_micronova_object(mv))
+        cg.add(mv.register_micronova_listener(numb))
+        cg.add(
+            numb.set_memory_location(
+                thermostat_temperature_config.get(CONF_MEMORY_LOCATION, 0x20)
+            )
+        )
```
jesserockz: These values are something that should go into the config as a default, and not hidden from the user in `to_code`. 

Two ways to do this are to add it to the final extend which will override the `MICRONOVA_LISTENER_SCHEMA`, 

or (maybe the better option) you can make `MICRONOVA_LISTENER_SCHEMA` a function/generator that will return the schema with the defaults if all number/sensor/switch etc will be setting these.

esphome/components/micronova/switch/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+CONFIG_SCHEMA = cv.Schema(
+    {
+        cv.GenerateID(CONF_MICRONOVA_ID): cv.use_id(MicroNova),
+        cv.Optional(CONF_STOVE): switch.switch_schema(
+            MicroNovaSwitch,
+            icon=ICON_POWER,
+        )
+        .extend(MICRONOVA_LISTENER_SCHEMA)
+        .extend({cv.Optional(CONF_MEMORY_DATA_OFF, default=0x06): cv.hex_int_range()})
+        .extend({cv.Optional(CONF_MEMORY_DATA_ON, default=0x01): cv.hex_int_range()}),
```
jesserockz: These should be in a single extend
```suggestion
        .extend(
            {
                cv.Optional(CONF_MEMORY_DATA_OFF, default=0x06): cv.hex_int_range(),
                cv.Optional(CONF_MEMORY_DATA_ON, default=0x01): cv.hex_int_range(),
            }
        ),
```

jesserockz (changes requested):
Mostly looks good, just a few changes below please =)

esphome/components/micronova/micronova.h
Jorre05: Correct, done.

esphome/components/micronova/micronova.h
Jorre05: Done, I hope I've spotted them all.

esphome/components/micronova/micronova.h
Jorre05: Done.

esphome/components/micronova/switch/__init__.py
Jorre05: Done.

esphome/components/micronova/number/__init__.py
Jorre05: Not entirely sure what you mean by function/generator. But, I've done my best.

esphome/components/micronova/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+            cv.GenerateID(): cv.declare_id(MicroNova),
+            cv.Required(CONF_ENABLE_RX_PIN): pins.gpio_output_pin_schema,
+        }
+    )
+    .extend(uart.UART_DEVICE_SCHEMA)
+    .extend(cv.polling_component_schema("60s"))
+)
+
+
+def MICRONOVA_LISTENER_SCHEMA(default_memory_location, default_memory_address):
```
jesserockz, line 50: This is exactly what I meant =)

esphome/components/micronova/number/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        .extend(
+            MICRONOVA_LISTENER_SCHEMA(
+                default_memory_location=0x20, default_memory_address=0x7D
+            )
+        )
+        .extend(
+            {cv.Optional(CONF_MEMORY_WRITE_LOCATION, default=0xA0): cv.hex_int_range()},
+            {cv.Optional(CONF_STEP, default=1.0): cv.float_range(min=0.1, max=10.0)},
+        ),
```
jesserockz:
```suggestion
        .extend(
            {
                cv.Optional(CONF_MEMORY_WRITE_LOCATION, default=0xA0): cv.hex_int_range(),
                cv.Optional(CONF_STEP, default=1.0): cv.float_range(min=0.1, max=10.0),
            }
        ),
```

jesserockz (dismissed):

jesserockz (approved):

PR #4761, opened 2023-04-29
Add device support for modbus

playya:

## Types of changes

- [ X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

## Test Environment

- [ ] ESP32
- [X ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
substitutions:
  device_name: "Modbus Device Test"
  device_id: "modbus_device_test"
  node_name: "modbus-device-test"
  slave_id: "0x1"

esphome:
  name: ${node_name}
esp32:
  framework:
    type: esp-idf
    advanced:
      ignore_efuse_mac_crc: true
uart:
  - id: mod_bus_uart
#    rx_pin: 9
#    tx_pin: 10
    rx_pin: 16
    tx_pin: 17
    baud_rate: 9600
    data_bits: 8
    parity: NONE
    stop_bits: 1

modbus:
  uart_id: mod_bus_uart
  id: mod_bus_server
  flow_control_pin: 5

modbus_device:
  - id: ${device_id}
    address: 1 # slave address

number:
  - platform: modbus_device
    modbus_device_id: ${device_id}
    id: battery_capacity_number
    name: "Battery Cap Number"
    address: 0x9001
    value_type: U_WORD
    read_lambda: |-
      ESP_LOGD("main","Modbus Number incoming value = %f",x);
      uint16_t b_capacity = x ;
      data.push_back(b_capacity);
      return x * 1.0 ;
    multiply: 1.0

```

## Checklist:


playya:
Hey @martgras How do you like moving some code from modbus_controller to modbus?
Currently I only implemented reading registers from number components, but may add more if this get's a chance to be accepted.

SeByDocKy:
What is  the difference with modbus_controller:?

playya:
This allows To implement device on modbus which May bei controlled by modbus Controllers.

JeroenVanOort:
Thank you very much for your work @playya, this is just what I am looking for!

I gave it a try, but I got an error:

```
src/esphome/components/modbus/modbus.cpp: In member function 'bool esphome::modbus::Modbus::parse_modbus_byte_(uint8_t)':
src/esphome/components/modbus/modbus.cpp:108:22: error: 'class std::vector<unsigned char>' has no member named 'pop_front'; did you mean 'front'?
     this->rx_buffer_.pop_front(byte);
                      ^~~~~~~~~
                      front
```

playya:
@jeffborg I pushed an update.

JeroenVanOort:
Thanks for the changes, the code now compiles!

However, I don't think all messages are parsed correctly.

Logs:
```
[07:52:56][D][modbus_device:085]: Received Server request 0A
[07:52:56][D][modbus_device:085]: Received Server request 0A
[07:52:56][D][modbus_device:085]: Received Server request 0A
[07:52:56][D][modbus_device:085]: Received Server request 0A
[07:52:56][D][modbus_device:085]: Received Server request 0A
[07:52:56][D][modbus_device:085]: Received Server request 0A
[07:52:56][D][uart_debug:114]: <<< 0A:03:50:0C:00:06:15:B0
```

It is seeing 0A as the function code, but that is the device address. For the function code, it should look one byte further. Either in or `get_message_info`, or by adding +1 to they payload when calling it.

Also, I get a bunch of errors when setting the log level to verbose.

JeroenVanOort:
@martgras I think we could do with a bit of background here. I've been looking into this (allowing ESPHome to act as a Modbus RTU server/slave) and I'm starting to doubt that a `modbus_device` component is what is needed; we might be able to get this to work with `modbus_controller`. But to make up my mind about that, I would like to know more about the thoughts that went into the Modbus-related component of ESPHome. Would you be willing to help us with that?

Questions:
1. What is behind the term 'controller'? It doesn't seem to be a term used often in the Modbus world. I guess it would be a generic term for something that is either a client or a server. I'm tempted to call it a 'device', which is also a term used in the current code.
2. Do you think a `modbus_controller` could be made to act as a server? And if so, did you plan for it? And if so, could you give us some hints on how you imagined this to work? And if even if not, what do you think needs to be done to make this work?

jpeletier:
Hello all,

As a contributor to ESPHome's `modbus_controller` and reviewing the example/draft code changes proposed in this PR to understand its intent around a new "modbus_device" platform, I think it is taking the wrong approach.

I believe that something more along the lines of a generic server that allows you to interact or bind to existing sensors and components would be more useful, whereas having to implement all basic-type components (switch, light, number...) as a "modbus_device" is a lot more code and a lot more maintenance work in the future while risking more bugs.

Another issue is that if for example, what you just want to do is to expose as a modbus server some complex, already-existing component such as a RGB light or some HVAC device, you would have to define one or more  "modbus_device" sensors/components of different types and then sync the state between them. So you still have to do binding and it is worse, because you have to keep both copies of the same state in sync.

This is a repeatable pattern. For example imagine you want to expose some components via a REST API. Would you then create "rest_api_number", "rest_api_switch", "rest_api_select" ... and so on? No. It would be easier to just enable listening on some URL endpoints and then giving you some trigger/script/lambda facilities to bind REST calls to anything already existing within ESPHome. And if you want something that just stores a number (like this PR's example), just bind it to a `platform: template` Number and you're done.

Some of you in this thread have tried and even forked my [modbus-server](https://github.com/epiclabs-uc/esphome-modbus-server) component. I know I have not given it the time it deserved as of late, but I think it is the right kind of approach. I acknowledge it is a minimal demo because it just gives you lambdas and then you have to do all the binding in there by hand, but indeed it provides most of what most people need with very little code.

I would support anyone who takes that concept and re-implements it, perhaps doing away with [modbus-server](https://github.com/epiclabs-uc/esphome-modbus-server)'s usage of a third-party library and just reusing modbus-related code already in ESPHome which is beautifully demonstrated in this draft. A first version can just have lambdas, a second one could provide triggers/script support so you can bind to other components using pure YAML.

> @JeroenVanOort: I would like to know more about the thoughts that went into the Modbus-related component of ESPHome. Would you be willing to help us with that?

For me `modbus_controller` is just a tool to proxy external modbus devices so they exist as ESPHome components. The name "controller" I believe comes from the fact that you are connecting to some external _controller_ that is in charge of some other system. It is true that it is a bit confusing and could also be called "modbus_client" or "modbus device"  (not to confuse with this draft PR's concept of "modbus_device" though), but I guess a name change now would make things even more confusing :-)

I hope this is helpful and clarifying. I am happy to discuss the architecture for a "modbus_server" component built-in into ESPHome and whant to thank @playya and everyone for contributing with the ideas and the working draft.

Cheers,
Javier

playya:
Hi @jpeletier,

I also used your server implementation, but had issues with re pin on my MAX485 breakout board.
I think using gpio + lib did not always set the pin until buffers are flushed completely and I had some incomplete messages.

I did start this implementation without any external libraries like modbus_controller to have controller over this behaviour and I currently se no incomplete messages.

Yes this code moves some code from modbus_controller to modbus to avoid duplication inc modbus_device.
I would like to provide all those base types for modbus_device which allows users to be used with actions to set those variables.
each type will get write transform functions to allow generic implementation of write functions and read_transform for read functions.

From my point of view specific modbus_controller follwing modbus protocool may be easier to be implemented by creating a ModbusController and register fields using add_sensor_item. This would only require python boilerplate code, exceot for write_transform,/read_transform. I try to achieve the same for modbus_device and one can easily implement own modbus devices.

@JeroenVanOort 
1. I think @jpeletier explained this quite well
2. I think we need to split this into 2 components because the receiver has the info if this is a request (device) or a reponse (controller) for same function codes. Also custom commands can be handled by overriding get_message_info.

JeroenVanOort:
Thank you both for responding!

I've been digging through the code and doing some experiments, and I think adding Modbus RTU server support to ESPHome can be broken down into the following parts:

1. Parsing bytes & checking CRC
    In order to receive the correct number of bytes and do the CRC, we need to know if we are acting as a server or a client (handling a request or a response), because the number of bytes for a request is fixed as opposed to a response where the number of data bytes is included in the message. @playya has chosen to do this by moving the code that determines the data length and offset from the `modbus` component to the `modbus_controller` and `modbus_device` components. This means that the device address first needs to be parsed in order to call `get_message_info` on the right `modbus_controller` or `modbus_device` instance.
    That will most likely work, but I am not sure if it is the best approach. Another way would be to add a config option to `modbus`, where users can set the role (client/server) for a `modbus` instance. The default would be 'client' to be backwards compatible. I just want to bring this up as an option, so we can think about it. Possibly, this would remove the need for a `modbus_device` component and thereby reduce both the extra code needed and the confusion about client/server/master/slave/controller/device.
2. Configuring the server registers
    When using ESPHome as a client, it makes a lot of sense to configure registers using sensors (sensor, binary_sensor, switch, etc). But when using ESPHome as a server, I think it would make much more sense to do something like this:
    
    ```yaml
    modbus_controller:
      - address: 0x2
        server_registers:
        - address: 0x1000
          value_type: S_DWORD
          read_lambda: |-
            return 42.42;
          write_lambda: |-
            // Optional
    ```
    I think this would also reduce the amount of code we need to add.

Please share what your thoughts are on this.

JeroenVanOort:
I had some time on my hands today and made this: https://github.com/esphome/esphome/pull/4874

playya:
1. I think my approach is more OOP-ish and allows to delegate calculating the size in case of custom messages not defined by modbus protocol.
2. I have chosen modbus_device because it represents a (virtual/fake) modbus device on the bus and not controlling a device on bus. I do not add much code. Most code is moved from controller to modbus to make this available for modbus_device. Also read/write is handled differently in sub components: if you write a value on controller it will be send to the device. A modbus device receives a request and publish the value on sensor.

jpeletier:
> I had some time on my hands today and made this: #4874

@JeroenVanOort This exactly what I am talking about. I gave it a good look and pointed to a very few things but nothing important. A net of +126 lines of code and it is **backwards compatible**, which is very important. From reading the code, I understand writing registers is TBD.

 I have yet to test the implementation, I will do so soon.

@playya you can have what you need with @JeroenVanOort 's PR in #4874, just use template component, or any component like this:

```yaml

# This switch can be controlled by HomeAssistant and by a Modbus remote client.
switch:
  - platform: gpio
    id: myswitch
    name: MyModbusSwitch
    pin: 25

modbus_controller:
  - id: modbus_evse_server
    modbus_id: modbus_server
    address: 0x2
    server_registers:
      - start_address: 0x0000
        value_type: U_WORD
        read_lambda: |-
          return id(myswitch).state;
        write_lambda: |-
          if( id(myswitch).state != !!value ) // value contains the register value sent by the client
            id(myswitch).toggle();
```

Worst case, you can leverage #4874 and build what you want on top of it, as an external component, so I suggest we center our efforts on this very much simpler approach.

playya:
We can also follow the role approach, if we can use it like modbus_controller client with number, sensor, etc.
One should be able to implement a modbus server in pure yaml.

This should also help to reuse definitions for client/server implementations.

PR #4770, opened 2023-05-02
Add Alarm Control Panel

grahambrown11:
# Implement Alarm Control Panel

- [ ] Other

A new `alarm_control_panel` that takes a list of `binary_sensors` that triggers the alarm when any of the sensor change to on, with the relevant API stuff to automatically create the Alarm Control Panel in HomeAssistant and provide state changes.

**Pull requests in progress:**
- ESPHome API: esphome/aioesphomeapi#427
- HomeAssistant: home-assistant/core#92357
- Docs: esphome/esphome-docs#2891

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
esphome:
  name: alarm-panel
  friendly_name: Alarm
  comment: Alarm Control Panel

esp8266:
  board: d1_mini

logger:

api:

alarm_control_panel:
  id: alarm
  name: Panel
  codes:
    - "1234"
  requires_code_to_arm: true
  arming_time: 10
  delay_time: 15
  trigger_time: 30
  binary_sensors:
    - input: zone_1
    - input: zone_2
      bypass_armed_home: true
    - input: ha_test
  on_state:
    then:
      - lambda: !lambda |-
          ESP_LOGD("TEST", "State change %s", id(alarm)->to_string(id(alarm)->get_state()).c_str());
  on_triggered:
    then:
      - switch.turn_on: siren
  on_cleared:
    then:
      - switch.turn_off: siren

binary_sensor:
  - platform: gpio
    id: zone_1
    name: Zone 1
    device_class: door
    pin:
      number: D1
      mode: INPUT_PULLUP
      inverted: True
  - platform: gpio
    id: zone_2
    name: Zone 2
    device_class: motion
    pin:
      number: D2
      mode: INPUT_PULLUP
      inverted: True
  - platform: homeassistant
    id: ha_test
    name: Zone 3
    entity_id: input_boolean.test_switch

switch:
  - platform: gpio
    id: siren
    name: Siren
    icon: mdi:alarm-bell
    pin: D7

```

## Checklist:


dougiteixeira:
It would be very useful to have the option of configuring multiple codes to control the alarm, so it would be possible for each person to have their own personal code.

Is this viable?

grahambrown11:
> It would be very useful to have the option of configuring multiple codes to control the alarm, so it would be possible for each person to have their own personal code.
> 
> Is this viable?

@dougiteixeira I'll add it to the list for future enhancements

nagyrobi:
> > It would be very useful to have the option of configuring multiple codes to control the alarm, so it would be possible for each person to have their own personal code.
> > Is this viable?
> 
> @dougiteixeira I'll add it to the list for future enhancements

Yes this is a must. Moreover, combined with a [Wiegand](https://esphome.io/components/wiegand.html) reader, not only numeric codes could be used, but also proxy cards, which are actually strings of various lengths. Each family member could have a separate proxy keyring or card.
Same with [Grow Fingerprint Reader](https://esphome.io/components/fingerprint_grow.html).

grahambrown11:
@nagyrobi 

> not only numeric codes could be used

The code is a string cos if your code started with a zero and it was stored as a numeric then the leading zero is dropped...

> combined with a Wiegand reader

For now you could use a HomeAssistant Automation to translate the many codes from the Wiegand reader
 into 1 to disarm... but will definatly add more codes in future release, for local ESPHome automation.
I just want to get this out as there are a lot of moving parts between ESPHome and HomeAssistant.

grahambrown11:
Multiple code support has now been added

esphome/components/alarm_control_panel/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+CONF_BYPASS_ARMED_HOME = "bypass_armed_home"
+CONF_REQUIRES_CODE_TO_ARM = "requires_code_to_arm"
+CONF_ARMING_HOME_TIME = "arming_home_time"
+CONF_ARMING_AWAY_TIME = "arming_away_time"
+CONF_DELAY_TIME = "delay_time"
+CONF_TRIGGER_TIME = "trigger_time"
+CONF_ON_TRIGGERED = "on_triggered"
+CONF_ON_CLEARED = "on_cleared"
+
+AlarmPanelNS = cg.esphome_ns.namespace("alarm_control_panel")
```
jesserockz:
```suggestion
alarm_control_panel_ns = cg.esphome_ns.namespace("alarm_control_panel")
```

esphome/components/alarm_control_panel/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+AlarmControlPanelComponent = AlarmPanelNS.class_("AlarmControlPanel", cg.Component)
```
jesserockz: Variable name should match the actual class name please.
```suggestion
AlarmControlPanel = AlarmPanelNS.class_("AlarmControlPanel", cg.Component)
```

esphome/components/alarm_control_panel/alarm_control_panel.h
```diff
@@ -0,0 +1,200 @@
...
+   */
+  AlarmControlPanelState get_state();
+
+ protected:
+  // in order to store last panel state in flash
+  ESPPreferenceObject pref_;
+  // the list of binary sensors that the alarm_panel monitors
+  std::vector<binary_sensor::BinarySensor *> sensors_;
+  // the list of sensor ids that the alarm_panel bypass when in armed_home
+  std::vector<uint32_t> bypass_when_home_;
```
jesserockz: You should really just store a pointer to the binary_sensor

esphome/components/alarm_control_panel/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+        cv.Required(CONF_INPUT): cv.use_id(binary_sensor.BinarySensor),
+        cv.Optional(CONF_BYPASS_ARMED_HOME, default=False): cv.boolean,
+    }
+)
+
+CONFIG_SCHEMA_ALARM_CONTROL_PANEL = cv.ENTITY_BASE_SCHEMA.extend(
+    {
+        cv.GenerateID(): cv.declare_id(AlarmControlPanelComponent),
+        cv.Optional(CONF_CODES): cv.ensure_list(cv.string_strict),
+        cv.Optional(CONF_REQUIRES_CODE_TO_ARM, False): cv.boolean,
```
jesserockz: There should be a validation on `CONF_REQUIRES_CODE_TO_ARM` because it is ignored if you dont have any codes

esphome/components/alarm_control_panel/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+        cv.Optional(CONF_ARMING_HOME_TIME, "0s"): cv.positive_time_period_seconds,
+        cv.Optional(CONF_ARMING_AWAY_TIME, "0s"): cv.positive_time_period_seconds,
+        cv.Optional(CONF_DELAY_TIME, "0s"): cv.positive_time_period_seconds,
+        cv.Optional(CONF_TRIGGER_TIME, "0s"): cv.positive_time_period_seconds,
+        cv.Required(CONF_BINARY_SENSORS): cv.ensure_list(CONFIG_SCHEMA_BINARY_SENSOR),
```
jesserockz: I don't think binary_sensors should be **required**. personally I might want to write custom logic for my binary sensors (or any other sensor type) based on other factors and manually "trigger" the ACP to go into pending/triggered mode via yaml actions.

esphome/components/alarm_control_panel/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+AlarmControlPanelCondition = AlarmPanelNS.class_(
+    "AlarmControlPanelCondition", automation.Condition
+)
+
+CONFIG_SCHEMA_BINARY_SENSOR = cv.Schema(
+    {
+        cv.Required(CONF_INPUT): cv.use_id(binary_sensor.BinarySensor),
+        cv.Optional(CONF_BYPASS_ARMED_HOME, default=False): cv.boolean,
+    }
+)
```
jesserockz: This should be wrapped in `cv.maybe_simple_id` to allow people not having to define a sub-object in yaml but just have a list of ids

esphome/components/alarm_control_panel/alarm_control_panel.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+void AlarmControlPanel::loop() {
+  // change from ARMING to ARMED_x after the arming_time_ has passed
+  if (this->current_state_ == AlarmControlPanelState::ARMING) {
+    auto delay = this->arming_away_time_;
+    if (this->desired_state_ == AlarmControlPanelState::ARMED_HOME) {
+      delay = this->arming_home_time_;
+    }
+    if ((millis() - this->last_update_) > delay) {
+      ESP_LOGD(TAG, "After Arming Time");
```
jesserockz: This log message seems a bit random

esphome/components/alarm_control_panel/alarm_control_panel.cpp
```diff
@@ -0,0 +1,268 @@
...
+uint32_t AlarmControlPanel::get_supported_features() {
+  // Home Assistant values:
+  // ARM_HOME = 1
+  // ARM_AWAY = 2
+  // ARM_NIGHT = 4
+  // TRIGGER = 8
+  // ARM_CUSTOM_BYPASS = 16
+  // ARM_VACATION = 32
+  return 1 + 2 + 8;
+}
```
jesserockz: This should be changed to use constants within here, and also should be set based on features users are actually using. For instance I would never use ARM_HOME in my house so I would not want HA to show it.

You don't have to implement all the modes in this PR, but at least ARM_HOME should be optional based on config.

jesserockz (changes requested):
One big thing missing from this is `calls` As per all of the other entities types that can be controlled (from with YAML or API) have a `Call` class that is used to `control` them. Please modify this to follow that same pattern.

I think the base ACP code should be more basic, and we should look at making platforms similar to all the other entity types with most of this code falling into the `template` platform under ACP.

There's probably more, but these requests might change the code too much to bother reviewing everything right now. Please reach out if you have questions.

esphome/components/alarm_control_panel/__init__.py
grahambrown11: I'm not sure I fully understand your usecase, in my mind an alarm system is a bunch of binary sensors that trigger the alarm when armed...
You can always use a template binary sensor to turn non binary sensors (like ultrasonic distance) into a binary on/off when over/under a certain value...

grahambrown11:
> One big thing missing from this is `calls` As per all of the other entities types that can be controlled (from with YAML or API) have a `Call` class that is used to `control` them. Please modify this to follow that same pattern.

Sure I'll work on that, didn't know there were patterns to follow 😅 

> I think the base ACP code should be more basic, and we should look at making platforms similar to all the other entity types with most of this code falling into the `template` platform under ACP.

I don't understand - I thought the `template` platform was for using Lambda which is advanced 🤔 this is basic from a user of ESPHome is it not?

esphome/components/alarm_control_panel/alarm_control_panel.h
grahambrown11: For the logic in the loop, I ignore the sensor when in the `bypass_when_home_` vector, I don't see how a pointer will help here... I have a vector of binary sensors above...

esphome/components/alarm_control_panel/__init__.py
nagyrobi: I think Jesse thinks of the possibility to use the system the other way around too: have actions triggered from the other parts of the config. Eg. an extreme case a binary sensor could change the state of the alarm fine-tuned from _Binary Sensor Filters_ / _Binary Sensor Automation_ without having to create copies of it with different settings for other uses (eg. the same binary sensor would be useful for both Alarm and A/C overriding, but with different trigger delay settings). 

So it would be very well feasible that someone wouldn't want to specify any `binary_sensors` in Alarm Control Panel, but would just use Alarm's Automation Actions to trigger it from various sensor's `on_...` sections.

I think this is just really about making binary_sensors optional.

jesserockz: I'm not saying remove the `binary_sensors`, just don't force people to only have `binary_sensors` as the way to trigger the alarm from `armed` to `pending`. 
One use case could be instant triggering, where some sensors could be used to instantly trigger the alarm/sirens instead of going into pending/countdown mode.

grahambrown11: Cool, I guess a less opinionated system is more flexible for everyone

RoganDawes:
I've been working on a UART integration with my Texecom Premier 832 alarm panel, which exposes the various sensors states as binary inputs, and provides controls for away arm/stay arm/disable, zone bypass, etc. The Texecom also has multiple partitions which can be armed and disarmed independently, but I'd love to be able to just extend the basic Alarm Control Panel classes to gain the automatic interfaces into e.g. the webserver and Home Assistant.

jesserockz (approved):

PR #4776, opened 2023-05-03
Add support for TMP1075 temperature sensor

sybrenstuvel:

Add support for the [TMP1075](https://www.ti.com/product/TMP1075), which is a "temperature sensor with I2C interface in industry standard LM75 form factor and pinout".

This is implemented as an i2c component. The configuration register values are exposed as a `struct` with fields for each setting, and setter functions for those fields. The component's `setup()` function will send the configuration to the device once, regardless of how many of those setters were called. When calling those setters to change settings (for example in a lambda expression) the function `write_config()` should be called as well in order to actually send the config to the device again.

I've added various `CONF_xxx` constants. To keep the changes local, they are currently in the component's `sensor.py` file. Depending on the review feedback I can of course move those into the `esphome.const` module.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation:** esphome/esphome-docs#2895

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
sensor:
  - platform: tmp1075
    name: TMP1075 Temperature
    id: radiator_temp
    update_interval: 10s
    i2c_id: i2c_bus_1
    conversion_rate: 27.5ms
    alert:
      limit_low: 50          # temperature in °C
      limit_high: 75         # temperature in °C
      fault_count: 1         # count between 1 and 4, inclusive
      polarity: active_high  # enum 'active_high' or 'active_low'
```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs)


esphome/components/tmp1075/tmp1075.cpp
```diff
@@ -0,0 +1,146 @@
...
+  ESP_LOGCONFIG(TAG, "  fault_count: %d", config_.fields.faults);
+  ESP_LOGCONFIG(TAG, "  polarity   : %d", config_.fields.polarity);
+  ESP_LOGCONFIG(TAG, "  alert_mode : %d", config_.fields.alert_mode);
+  ESP_LOGCONFIG(TAG, "  shutdown   : %d", config_.fields.shutdown);
+}
+
+void TMP1075Sensor::set_alert_limit_low(const float temp) { this->alert_limit_low_ = temp; }
+void TMP1075Sensor::set_alert_limit_high(const float temp) { this->alert_limit_high_ = temp; }
+void TMP1075Sensor::set_oneshot(const bool oneshot) { config_.fields.oneshot = oneshot; }
+void TMP1075Sensor::set_conversion_rate(const enum EConversionRate rate) { config_.fields.rate = rate; }
```
kroimon: Please move simple setter and getter methods to the header file in accordance with the rest of the codebase.

esphome/components/tmp1075/tmp1075.cpp
```diff
@@ -0,0 +1,146 @@
...
+  if (faults > 4) {
+    ESP_LOGE(TAG, "'%s' - fault_count too high: %d", this->name_.c_str(), faults);
+    return;
+  }
+  config_.fields.faults = faults - 1;
+}
+void TMP1075Sensor::set_alert_polarity(const bool polarity) { config_.fields.polarity = polarity; }
+void TMP1075Sensor::set_alert_function(const enum EAlertFunction function) { config_.fields.alert_mode = function; }
+
+void TMP1075Sensor::load_config_() {
```
kroimon: This is never called, is it?

sybrenstuvel: True! It's a leftover from when I wanted to load the IC configuration at startup. Later I decided that it's better to just have ESPHome fully define the IC's config instead. I'll remove the code.

kroimon (approved):

kbx81 (approved):

PR #4792, opened 2023-05-08
Fix ezo parsing

alfredopironti:

The current parsing of the EZO sensor had a few issues that wouldn't make them work. Namely:

- The returned value would be truncated always (not only for EZO_READ), which would not allow any other command to actually work (cd5dae3)
- The temperature-compensated readings were not properly parsed (f91f691)
- Additionally, the component was previously logging binary data (f2dfba2)

This is tested against both a pH and ORP EZO sensors. I've carefully reviewed all EZO datasheets, and they all follow the same pattern, so this code should work with every EZO sensor available at time of writing. The only caveat is the RGB EZO sensor which, by nature, on EZO_READ returns 3 comma-separated values (R,G,B). The current implementation only returns the "R" component. Support for the RGB sensor entails a breaking change, where "EZO_READ" returns an array of values. I'd be happy to implement such a change, but didn't want to introduce a breaking change at this stage.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** N/A (no documentation updates required)

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

Typical use case reading the EZO_CALIBRATION status:

```yaml

button:
  - platform: template
    name: "Retrieve pH calibration status"
    on_press:
      - lambda: |-
         id(ezo_ph).get_calibration();

sensor:
  - platform: ezo
    id: ezo_ph
    address: 99
    name: "pH sensor"
    unit_of_measurement: "pH"
    accuracy_decimals: 2
    on_calibration:
      - text_sensor.template.publish:
          id: ph_calibration_status
          state: !lambda |-
            return x;

```

Without this patch, the "on_calibration" callback is never invoked, due to parsing issues.

## Checklist:

### DIFF
--- a/esphome/components/ezo/ezo.cpp
+++ b/esphome/components/ezo/ezo.cpp
@@ -108,18 +108,16 @@ void EZOSensor::loop() {
 
   ESP_LOGV(TAG, "Received buffer \"%s\" for command type %s", &buf[1], EZO_COMMAND_TYPE_STRINGS[to_run->command_type]);
 
-  if ((buf[0] == 1) || (to_run->command_type == EzoCommandType::EZO_CALIBRATION)) {  // EZO_CALIBRATION returns 0-3
-    // some sensors return multiple comma-separated values, terminate string after first one
-    for (size_t i = 1; i < sizeof(buf) - 1; i++) {
-      if (buf[i] == ',') {
-        buf[i] = '\0';
-        break;
-      }
-    }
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Can't test it, but code looks fine.

PR #4797, opened 2023-05-09
Add support for Grove tb6612 fng

max246:

I am working with this handy motor driver which can drive DC and Stepper motors : https://wiki.seeedstudio.com/Grove-I2C_Motor_Driver-TB6612FNG/
The good thing about this driver is the connectivity to be used over i2c rather than the usual hbridge and you can set the motor driver to standby to stop wasting power.

Methods implemented:
- grove_i2c_motor.motor_run
- grove_i2c_motor.motor_break
- grove_i2c_motor.motor_stop
- grove_i2c_motor.standby
- grove_i2c_motor.no_standby

At the current state, only DC motor is implemented but the code for the stepper motor is already ready to be expose via an action. I dont currently have a stepper motor so in my case I can only test a DC motor

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2906

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ x ] ESP8266
- [ ] RP2040

```yaml
i2c:
  sda: 4
  scl: 5
  scan: true
  id: bus_a
grove_i2c_motor:
    id: test_motor
    name: motor_outside
    address: 0x14

switch:
  - platform: template
    name: open_vent
    id: open_vent
    optimistic: True
    on_turn_on: 
      then:
        - grove_i2c_motor.motor_run:
            channel: 1
            speed: 255
            direction: BACKWARD
            id: test_motor
        - delay: 10s
        - switch.turn_on: stop_motor
        - switch.turn_off:  open_vent



# Example config.yaml

```

## Checklist:
  - [ x ] The code change is tested and works locally.


tests/test3.yaml
```diff
@@ -1171,3 +1171,5 @@ daly_bms:
 qr_code:
   - id: homepage_qr
     value: https://esphome.io/index.html
+
```
HeMan: These extra lines could be removed.

esphome/components/grove_i2c_motor/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+DIRECTION_TYPE = {
+    "FORWARD": 1,
+    "BACKWARD": 2,
+}
+
+CONFIG_SCHEMA = (
+    cv.Schema(
+        {
+            cv.Required(CONF_ID): cv.declare_id(GROVE_TB6612FNG),
+            cv.Required(CONF_NAME): cv.string,
```
kroimon: What is the name used for?

esphome/components/grove_i2c_motor/grove_i2c_motor.cpp
```diff
@@ -0,0 +1,194 @@
...
+      return;
+    }
+  } else {
+    if (this->write_register(GROVE_MOTOR_DRIVER_I2C_CMD_CCW, buffer_, 2) != i2c::ERROR_OK) {
+      ESP_LOGW(TAG, "Run motor failed!");
+      this->status_set_warning();
+      return;
+    }
+  }
+  delayMicroseconds(100000);
```
kroimon: Do you really need a 100 ms delay here? That feels overly long, especially if no other command directly follows.
Same question for the other delays in the code.

You could also remember the time of the last sent command and only delay right before sending the next command, if the time since the last command was too short.

max246: Probably I can remove them, it came from the offical library

esphome/components/grove_i2c_motor/__init__.py
max246: I suppose it was nice to have a name but maybe I can make it optional?

kroimon: If it's just cosmetics because it's not used in code, remove them. Users can always just use comments in the YAML if they need additional information.

esphome/components/grove_i2c_motor/grove_i2c_motor.h
```diff
@@ -0,0 +1,224 @@
+#pragma once
+
+#include "esphome/components/i2c/i2c.h"
+#include "esphome/core/component.h"
+#include "esphome/core/hal.h"
+#include "esphome/core/automation.h"
+
+/*
```
descipher, line 9: Embeding driver code may not be the most effective method, please consider using this method:

  platformio_libs:
    - https://github.com/Seeed-Studio/Grove_Motor_Driver_TB6612FNG.git

max246: And writer a class as helper to use the imported library?

descipher: You should be able to just include the external Grove_Motor_Driver_TB6612FNG.h and use the same calls in the local component. So essentially yes.

descipher (approved):
Looks good to me.

max246:
> Looks good to me.

I tried to use the library provided without the GitHub link, but is behind a version which had compile issues.

descipher:
> > Looks good to me.
> 
> I tried to use the library provided without the GitHub link, but is behind a version which had compile issues.

Yes, hit that type of issue b4, compliler barfs but the code is fine.

esphome/components/grove_i2c_motor/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+@automation.register_action(
+    "grove_i2c_motor.motor_run",
+    GROVETB6612FNGMotorRunAction,
+    cv.Schema(
+        {
+            cv.Required(CONF_ID): cv.use_id(GROVE_TB6612FNG),
+            cv.Required(CONF_CHANNEL): cv.int_range(min=0, max=1),
+            cv.Required(CONF_SPEED): cv.int_range(min=0, max=255),
```
jesserockz: You access these as templatable below, but they are not in the schema? Same for other actions below.
```suggestion
            cv.Required(CONF_CHANNEL): cv.templatable(cv.int_range(min=0, max=1)),
            cv.Required(CONF_SPEED): cv.templatable(cv.int_range(min=0, max=255)),
```

esphome/components/grove_i2c_motor/grove_i2c_motor.cpp
```diff
@@ -0,0 +1,183 @@
...
+  }
+  this->set_i2c_address(addr);
+}
+
+void GroveMotorDriveTB6612FNG::dc_motor_run(uint8_t chl, int16_t speed) {
+  if (speed > 255) {
+    speed = 255;
+  } else if (speed < -255) {
+    speed = -255;
+  }
```
jesserockz:
```suggestion
  speed = clamp<int16_t>(speed, -255, 255);
```

esphome/components/grove_i2c_motor/grove_i2c_motor.cpp
```diff
@@ -0,0 +1,183 @@
...
+  if (this->write_register(GROVE_MOTOR_DRIVER_I2C_CMD_SET_ADDR, &addr, 1) != i2c::ERROR_OK) {
+    ESP_LOGW(TAG, "Set new i2c address failed!");
+    this->status_set_warning();
+    return;
+  }
+  this->set_i2c_address(addr);
+}
+
+void GroveMotorDriveTB6612FNG::dc_motor_run(uint8_t chl, int16_t speed) {
```
jesserockz: No need to shorten variable names, just makes it harder to understand
```suggestion
void GroveMotorDriveTB6612FNG::dc_motor_run(uint8_t channel, int16_t speed) {
```

esphome/components/grove_i2c_motor/grove_i2c_motor.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+  if (steps > 0) {
+    cw = 1;
+  }
+  // stop
+  else if (steps == 0) {
+    this->stepper_stop();
+    return;
+  } else if (steps == -32768) {
+    steps = 32767;
```
jesserockz:
```suggestion
  } else if (steps == INT16_MIN) {
    steps = INT16_MAX
```

esphome/components/grove_i2c_motor/grove_i2c_motor.cpp
```diff
@@ -0,0 +1,183 @@
...
+  } else {
+    steps = -steps;
+  }
+
+  if (rpm < 1) {
+    rpm = 1;
+  } else if (rpm > 300) {
+    rpm = 300;
+  }
```
jesserockz:
```suggestion
  rpm = clamp<uint16_t>(rpm, 1, 300);
```

esphome/components/grove_i2c_motor/grove_i2c_motor.cpp
```diff
@@ -0,0 +1,183 @@
...
+  // 4=>infinite ccw  5=>infinite cw
+  uint8_t cw = (is_cw) ? 5 : 4;
+  // 0.1ms_per_step
+  uint16_t ms_per_step = 0;
+
+  if (rpm < 1) {
+    rpm = 1;
+  } else if (rpm > 300) {
+    rpm = 300;
+  }
```
jesserockz:
```suggestion
  rpm = clamp<uint16_t>(rpm, 1, 300);
```

esphome/components/grove_i2c_motor/grove_i2c_motor.h
```diff
@@ -0,0 +1,224 @@
...
+
+  void play(Ts... x) override {
+    auto chl = this->chl_.value(x...);
+    auto speed = this->speed_.value(x...);
+    this->motor_->dc_motor_run(chl, speed);
+  }
+
+ protected:
+  GroveMotorDriveTB6612FNG *motor_;
+};
```
jesserockz: If you also `#include "esphome/core/helpers.h" at the top, each action can be made slightly simpler:
```suggestion
template<typename... Ts> class GROVETB6612FNGMotorRunAction : public Action<Ts...>, public Parented<GroveMotorDriveTB6612FNG> {
 public:
  TEMPLATABLE_VALUE(uint8_t, chl)
  TEMPLATABLE_VALUE(uint16_t, speed)

  void play(Ts... x) override {
    auto chl = this->chl_.value(x...);
    auto speed = this->speed_.value(x...);
    this->parent_->dc_motor_run(chl, speed);
  }
};
```

jesserockz (changes requested):
Generally looks good, just a few "tips" to align better with the codebase =)

max246:
@jesserockz could this be approved?

jesserockz (changes requested):
Please add the actions to the test yaml file too

max246:
> Please add the actions to the test yaml file too

Should be all good now

jesserockz (dismissed):

jesserockz (approved):

PR #4809, opened 2023-05-11
Tuya: Prevent loop when setting colors on case-sensitive dps

richardhopton:

If the TuyaMCU expects lower case color data, the current code gets stuck in an infinite loop
1) ESPHome sets datapoint to uppercase color
2) TuyaMCU sets the datapoint to lowercase color which is treated as a change
3) ESPHome updates state based on reported value from MCU and triggers ESP to set the dp again...

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/tuya/light/tuya_light.cpp
+++ b/esphome/components/tuya/light/tuya_light.cpp
@@ -57,35 +57,43 @@ void TuyaLight::setup() {
         return;
       }
 
+      float red, green, blue;
       switch (*this->color_type_) {
         case TuyaColorType::RGBHSV:
         case TuyaColorType::RGB: {
-          auto red = parse_hex<uint8_t>(datapoint.value_string.substr(0, 2));
-          auto green = parse_hex<uint8_t>(datapoint.value_string.substr(2, 2));
-          auto blue = parse_hex<uint8_t>(datapoint.value_string.substr(4, 2));
-          if (red.has_value() && green.has_value() && blue.has_value()) {
-            auto rgb_call = this->state_->make_call();
-            rgb_call.set_rgb(float(*red) / 255, float(*green) / 255, float(*blue) / 255);
-            rgb_call.perform();
-          }
+          auto red_int = parse_hex<uint8_t>(datapoint.value_string.substr(0, 2));
+          auto green_int = parse_hex<uint8_t>(datapoint.value_string.substr(2, 2));
+          auto blue_int = parse_hex<uint8_t>(datapoint.value_string.substr(4, 2));
+          if (!red_int.has_value() || !green_int.has_value() || !blue_int.has_value())
+            return;
+
+          red = float(*red_int) / 255;
+          green = float(*green_int) / 255;
+          blue = float(*blue_int) / 255;
           break;
         }
         case TuyaColorType::HSV: {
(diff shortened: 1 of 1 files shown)


esphome/components/tuya/light/tuya_light.cpp
```diff
@@ -206,7 +214,7 @@ void TuyaLight::write_state(light::LightState *state) {
         break;
       }
     }
-    this->parent_->set_string_datapoint_value(*this->color_id_, color_value);
+    parent_->set_string_datapoint_value(*this->color_id_, color_value);
```
richardhopton: this change was made for consistency with the bunch of other `parent->...` invocations in this method.

jesserockz: I would rather the others all be changed to `this->` to be more consistent with the rest of the codebase. (I am trying to weed out the instances where `this->` was missed when I catch them.

jesserockz (commented):
Apart from my above comment about `this->` it looks good.

I don't think this has anything to do with case-sensitivity, but rather if the device sends responds back at all after setting the colour it will get into a loop. Good solution either way.

richardhopton:
> Apart from my above comment about `this->` it looks good.
> 
> I don't think this has anything to do with case-sensitivity, but rather if the device sends responds back at all after setting the colour it will get into a loop. Good solution either way.

Actually, you're right it should loop but in `set_string_datapoint_value` there is a check to not see the datapoint value if it's already that value... which under normal circumstances stops a loop from happening at all. The issue here is that some TuyaMCUs need capitalized color values or they won't work and others don't care if you set upper or lower case but they lowercase the value. Since 00ff00 and 00FF00 are different so the equality check in `set_string_datapoint_value` doesn't stop the re-write, and the loop is allowed to continue.

esphome/components/tuya/light/tuya_light.cpp
richardhopton: I did wonder - let me update them all to `this->parent_...` before you merge.

esphome/components/tuya/light/tuya_light.cpp
```diff
@@ -57,37 +57,45 @@ void TuyaLight::setup() {
...
-          }
+          auto red_int = parse_hex<uint8_t>(datapoint.value_string.substr(0, 2));
+          auto green_int = parse_hex<uint8_t>(datapoint.value_string.substr(2, 2));
+          auto blue_int = parse_hex<uint8_t>(datapoint.value_string.substr(4, 2));
+          if (!red_int.has_value() || !green_int.has_value() || !blue_int.has_value())
+            return;
+
+          red = float(*red_int) / 255;
+          green = float(*green_int) / 255;
+          blue = float(*blue_int) / 255;
```
ssieb:
```suggestion
          auto rgb = parse_hex<uint32_t>(datapoint.value_string.substr(0, 6));
          if (!rgb.has_value())
            return;

          red = (*rgb >> 16) / 255.0f;
          green = ((*rgb >> 8) & 0xff) / 255.0f;
          blue = (*rgb & 0xff) / 255.0f;
```

jesserockz (approved):

PR #4818, opened 2023-05-14
Rp2040 pio ledstrip

Papa-DMan:

While waiting for fastled to push a release with their rp2040 support, finishing my uni courses, and some encouragement from members in this community I decided to take a look at writing a write a very basic led strip driver for the rp2040 from scratch

This driver utilizes the programmable IO state machines in the RP2040, I wrote out the assembly code for a couple of known chipset timings.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2917

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040

```yaml
# Example config.yaml
light:
  - platform: rp2040_pio_led_strip
    name: led_strip
    id: led_strip
    pin: GPIO13
    num_leds: 10
    pio: 0
    rgb_order: GRB
    chipset: WS2812B

```

## Checklist:


jesserockz (commented):
Love it. 

I have a question now for you... 

The .pio files are the same except the timings right? If so, can we auto generate the .pio file from python and then also generate the .pio.h file from that. This way we only have to add the chipsets/timings into the python file and the rest of the code can just expect a single .pio / .pio.h file to exist at compile time for the users chosen chipset or custom timings as I did in `esp32_rmt_led_strip`

Papa-DMan:
totally doable. on it!

Papa-DMan:
Okay I've hit an issue here, I can generate the files and can ask the platformio api to run the assembler included in the rp2040 platform to assemble the .pio files to the .pio.h files. The issue is that I'm not sure where the generated files should reside. I could generate them to the component folder, however what I would really want is to generate them over to the .esphome build directory before the platformio compile command is executed. I'm currently generating these files during the to_code(config) function call. I was able to get the build path from the core but since the to_code runs before write_code is called the build directory isn't created and populated yet.

jesserockz:
Yeah I thought something like this might happen. Push what you have done and Ill take a look

jesserockz:
I have pushed changes that should work, but I am getting this error when i run it:
```
Using tool-rp2040tools@1.0.2 package
Error: [Errno 2] No such file or directory: 'pioasm'
```

My changes will allow any future component to also write pio files to be generated

Papa-DMan:
Okay I've figured it out. When the api call was made to run `platformio pkg exec --package platformio/tool-rp2040tools -- pioasm pio_path pio_path.h` it was picking up my local pioasm installation instead. Upon removing it I got the error:
 `Using tool-rp2040tools@1.0.2 package
Error: [Errno 2] No such file or directory: 'pioasm'`
like you did. Apparently rp2040tools doesn't come with the pioasm like I thought it did. Upon some further digging I discovered that there is in fact a separate platformio package by earlephilhower:
`earlephilhower/tool-pioasm-rp2040-earlephilhower`
that would provide the pioasm tool. I installed that package locally and changed the api call to use that package instead, and when compiling test6.yaml the assembly portion doesn't fail out.

jesserockz:
I have pushed some more changes to make the whole thing very dynamic, and allows multiple led strips etc. Untested, but compiles.

Papa-DMan:
Okay so I've just tested it with my scope and it is putting out the right times both when selecting a chipset and using custom timings in the yaml. However all the data is zero instead of the pixel data regardless of what the light is set to.

jesserockz:
I will set up a board today to actually test it

jesserockz:
Ok I have fixed it and it now sends the actual pixel out, but its only sending a single pixel, not the 8 i have configured

jesserockz:
Is this fully working for you with multiple pixels?

Papa-DMan:
It is somewhat working with multiple Pixels. I am only getting about half of the configured Pixels getting lit up, and when they are the color isn't correct. On my 300 LED WS2812B test strip I set to full red the result is a pattern of 8 pixels at red 9 pixels at blue and 8 pixels at green for the first half of the strip, the second half is off.

jesserockz:
That is the exact result I was getting. Will try your latest commit

jesserockz:
Dang off-by-1 bugs...haha

Working perfectly now

jesserockz (approved):

PR #4833, opened 2023-05-15
[ILI9XXX] Improve display transfer performance & add second FB

Fabian-Schmidt:

Improve transfer speed to SPI display - example WT32-SC01 480x320px display 16 bit color from ~215ms to 105ms.

Use improvement to add a second framebuffer so full display redraw worst case takes ~170ms. Shortcut for full screen refresh is ~115ms.

Second framebuffer is optional and used to detect changes by comparing last display to current display. Goal is to make writing of display lambda functions easier. As I can now use `fill` operation etc.

Typical ESP32 devices with display have PSRam. Smallest possible PSRam is 2MB. One framebuffer with 480x320px and 16 bit color uses 300kb. So there should always be enough memory for second framebuffer.

All performance gain only on ESP32 Arduino. ESP32 IDF SPI is very very slow.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -185,15 +262,16 @@ void ILI9XXXDisplay::display_() {
...
   for (uint16_t row = 0; row < h; row++) {
     uint32_t pos = start_pos + (row * width_);
+    uint32_t pos_16 = pos + pos;
     uint32_t rem = w;
 
     while (rem > 0) {
       uint32_t sz = std::min(rem, ILI9XXX_TRANSFER_BUFFER_SIZE);
       // ESP_LOGVV(TAG, "Send to display(pos:%d, rem:%d, zs:%d)", pos, rem, sz);
-      buffer_to_transfer_(pos, sz);
+      buffer_to_transfer_(pos, pos_16, sz);
```
nielsnl68: `
      this->buffer_to_transfer_(pos, pos_16, sz);
`

to make it more consistent met the other changes you made below.

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -208,45 +286,70 @@ void ILI9XXXDisplay::display_() {
...
   this->x_high_ = 0;
   this->y_high_ = 0;
 }
 
-uint32_t ILI9XXXDisplay::buffer_to_transfer_(uint32_t pos, uint32_t sz) {
-  for (uint32_t i = 0; i < sz; ++i) {
-    switch (this->buffer_color_mode_) {
-      case BITS_8_INDEXED:
-        transfer_buffer_[i] = display::ColorUtil::color_to_565(
+uint32_t ILI9XXXDisplay::buffer_to_transfer_(uint32_t pos, uint32_t pos_16, uint32_t sz) {
```
nielsnl68: Not sure of i see the  benefits to send `pos_16` as parameter in compare with just create the variable only when you are in `BITS_16` mode.

nielsnl68 (changes requested):
Thanks @Fabian-Schmidt for taking over the touch to make the ili9xxx platform component much faster and better.

I have added 2 small change requests.

esphome/components/ili9xxx/ili9xxx_display.cpp
Fabian-Schmidt: fixed

esphome/components/ili9xxx/ili9xxx_display.cpp
Fabian-Schmidt: I was playing with performance optimizations and this one is not worth it. Removed the parameter.

Fabian-Schmidt:
Thanks @nielsnl68 for this very quick feedback on the changes.

PR #4853, opened 2023-05-18
[Logger] ESP32 S3 serial logger

Fabian-Schmidt:

The ESP32 S2 & S3 have build in USB peripheral support. Often it is used instead of external USB-to-UART bridge (CP210x / FTDI). It is used for flashing and must also be used for serial logging.

This PR adds Arduino support for it and set is as default logger for ESP32 S2 & S3.

I am unsure if `USB_CDC` or `USB_SERIAL_JTAG` is the correct name. Both are very similar in ESP32-S3 and USB JTag is a superset of USB CDC. My understanding is Arduino `USBSerial` uses USB CDC aka 'USB OTG Console' (see https://github.com/espressif/arduino-esp32/blob/8ee5f0a11e5423a018e0f89146e05074466274db/cores/esp32/HWCDC.h#L105) ESP docs for reference:
- [USB OTG Console](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-guides/usb-otg-console.html)
- [USB Serial/JTAG Controller Console](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-guides/usb-serial-jtag-console.html)

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32-S2
- [x] ESP32-S3
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
esphome:
  name: t-display-s3-touch-idf
esp32:
  board: esp32-s3-devkitc-1
logger:
```

## Checklist:


spectrumjade:
I've been trying to incorporate the ESP32-S2 into some projects and I found that by default (on my board at least) the CDC logger was made available as `Serial` rather than `USBSerial`. I noticed the same thing with the ESP32-C3. This actually causes a compile-time failure due to incompatible types when using the UART component (for which I have a fix proposed in #4869).

What's interesting is that the S3 board seems to have a different set of compile flags in PlatformIO than the S2 and C3. Both of those two have `ARDUINO_USB_CDC_ON_BOOT` enabled as well, which has the effect of assigning the CDC logger to `Serial`. I think that your PR would run into issues in this case because `USBSerial` would be undefined.

https://github.com/platformio/platform-espressif32/blob/70fc09635ba59272d97d06b0c909baee36fecfec/boards/lolin_s2_mini.json#L10
https://github.com/platformio/platform-espressif32/blob/70fc09635ba59272d97d06b0c909baee36fecfec/boards/lolin_c3_mini.json#L10

jesserockz:
Related: https://github.com/esphome/esphome/pull/4658

Fabian-Schmidt:
Hi @spectrumjade 
I think it is actually worse. It is not just that S3 boards have different set of flags. It is just some boards have the flag and some don't. So some S3 have `DARDUINO_USB_CDC_ON_BOOT` and some don't and same with S2 and C3. I am unsure if this is related to the circuit on the device or based on who created the json. 
Unsure how to set an appropriate default in ESP Home given this scenario.

Fabian-Schmidt:
Should I strip this PR of the C++ code changes and just leave the python changes of the default values?
Or should this be incorporated into PR #4658 and I abandon this one?

jesserockz:
@Fabian-Schmidt if you want to strip this PR down to just the config validation changes I can then rebase and add that into #4658

Fabian-Schmidt:
@jesserockz I stripped down this PR to just the config validation changes.

esphome/components/logger/__init__.py
```diff
@@ -147,6 +147,8 @@ def validate_local_no_higher_than_global(value):
                 CONF_HARDWARE_UART,
                 esp8266=UART0,
                 esp32=UART0,
+                esp32_s2=USB_CDC,
+                esp32_s3=USB_CDC,
```
kbx81, line 175: Based on [similar work I did](https://github.com/esphome/esphome/pull/5559), I'd generally recommend using `USB_JTAG_SERIAL` whenever possible. It has some advantages over CDC (I'll spare you the details here, unless you really want to know).

kbx81: Should we also add the C3 here, too?

kbx81: Note that the S2 _only_ supports CDC, the S3 and C3 both support JTAG over USB, and last but not least the C3 does not support CDC at all.

kbx81: This will require some fixing since CDC doesn't quite work correctly on the S3 and C3 when using IDF...so it shouldn't be the default in that case.

kbx81:
The fix for the failing test is in #4658

kbx81 (approved):

PR #4854, opened 2023-05-18
mqtt: add ESP-IDF >= 5.0 support

stintel:

Add ESP-IDF >= 5.0 support to the MQTT component.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/mqtt/mqtt_backend_idf.cpp
+++ b/esphome/components/mqtt/mqtt_backend_idf.cpp
@@ -11,6 +11,7 @@ namespace mqtt {
 static const char *const TAG = "mqtt.idf";
 
 bool MQTTBackendIDF::initialize_() {
+#if ESP_IDF_VERSION_MAJOR < 5
   mqtt_cfg_.user_context = (void *) this;
   mqtt_cfg_.buffer_size = MQTT_BUFFER_SIZE;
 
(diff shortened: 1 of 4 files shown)


esphome/components/mqtt/mqtt_backend_idf.cpp
```diff
@@ -80,7 +116,9 @@ void MQTTBackendIDF::mqtt_event_handler_(const Event &event) {
       ESP_LOGV(TAG, "MQTT_EVENT_CONNECTED");
       // TODO session present check
       this->is_connected_ = true;
+#if ESP_IDF_VERSION_MAJOR < 5
       this->on_connect_.call(!mqtt_cfg_.disable_clean_session);
+#endif
```
jesserockz: Does this mean no `on_connect` callback for idf 5+?

stintel: Unfortunately I don't remember why I did this two weeks ago. I'll have to dig back into it, but as I'm currently travelling it might be a while.

jesserockz: Im guessing because the `mqtt_cfg_.disable_clean_session` has changed in idf 5+. Funny thing though is that the boolean is not actually exposed in YAML/lambdas and could probably just be removed from the trigger entirely.

stintel: Shall I just remove it then?

jesserockz: Yes, if we decide to expose it to the yaml later it can be added back.

stintel: Done.

esphome/components/mqtt/mqtt_backend_idf.cpp
```diff
@@ -80,7 +116,6 @@ void MQTTBackendIDF::mqtt_event_handler_(const Event &event) {
       ESP_LOGV(TAG, "MQTT_EVENT_CONNECTED");
       // TODO session present check
       this->is_connected_ = true;
-      this->on_connect_.call(!mqtt_cfg_.disable_clean_session);
```
jesserockz, line 83: What I meant by remove it, was remove the argument to the trigger so it can be called from both versions:
It will require changes to the definition of `on_connect_` to rmeove the arg etc.
```
this->on_connect_.call()
```

jesserockz (changes requested):

esphome/components/mqtt/mqtt_backend_idf.cpp
stintel: Thanks. I couldn't get it figured out.

jesserockz (approved):

PR #4861, opened 2023-05-19
Add ES8311 audio codec component

kroimon:

Implements a component to interface with the ES8311 audio codec ([Datasheet](http://www.everest-semi.com/pdf/ES8311%20PB.pdf)).

I could not find any publicly available documentation for the I2C and I2S interfaces, so this is based on information found in other available open source drivers like the one from [espressif/esp-bsp](https://github.com/espressif/esp-bsp/tree/a402bd3e55ebd383095e83b45dd4be178c76c6a9/components/es8311) and [espressif/esp-adf](https://github.com/espressif/esp-adf/tree/a74a26f1f2c2f84a7c3631bdd34d365a26704fcc/components/audio_hal/driver/es8311).

Besides others, this chips is used on the Espressif ESP32-S3-BOX. Other development efforts regarding this hardware is being coordinated [here](https://github.com/esphome/feature-requests/issues/2239).

## ⚠️ Work in Progress
This is still WIP. The PR is here to track progress and encourage feedback!

Missing bits and pieces:

- Not fully tested
- No configuration options yet
- Should have a way to properly use MCLK which needs to have support by the i2s_audio component.
- Probably more

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** TODO 

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
i2c:
  scl: GPIO18
  sda: GPIO8

i2s_audio:
  i2s_lrclk_pin: GPIO47
  i2s_bclk_pin: GPIO17
  #i2c_mclk_pin: GPIO2

es8311:
  address: 0x18
```

## Checklist:


rpatel3001:
FYI, there is a full datasheet and user guide available here: https://github.com/espressif/esp-dev-kits/issues/24#issuecomment-781314125

rpatel3001:
Looks like [793e22c](https://github.com/esphome/esphome/pull/4861/commits/793e22cddcee013e47360ff7bac0a59e5b8301f2) needs a slight fix, the types in the templates for remap() in set_volume and get_volume need to be reversed.

also, i figure the volume here should be set to max and left to be controlled by media_player

kroimon:
> the types in the templates for remap() in set_volume and get_volume need to be reversed.

🤦 Oops, you're right. Thanks, fixed!

> also, i figure the volume here should be set to max and left to be controlled by media_player

For the media player probably yes.
For the speaker component, it would probably make sense to leave volume control to the DAC, wouldn't it?

I'm not sure how to best handle volume control for external DACs to be honest...

rpatel3001:
For a small change, exposing the initial volume setting as a config option for the codec would be an easy fix, at least for being able to set the volume statically. 

A bigger fixer for dynamic changes might be to add an abstract class for codecs/dacs and have media_player optionally take the id of one of those to override it's internal volume control.

Slightly unrelated, does it make sense to add mclk_freq and sample_rate params to i2s_audio then have the codec/adc add the i2s bus id as a parameter? That way they can pull the data directly, the way media_player pulls pin numbers, instead of having to manually keep the two configs in sync. I suppose it depends if the shared bus is supposed to have a fixed frequency or if each component should take control and reconfigure it when active (preventing simultaneous use).

kroimon:
Yeah you are right, I will add the initial volume as a config option, probably just with 100% as default.

The mclk frequency and sample rate can unfortunately not be configured statically. The media player component sets the sample rate based on the media file it plays.
This also makes it very hard to impossible to use the same I2S port for microphone and media player at the same time, as the voice assistant always expects a 16000 Hz sample rate at the moment... I'm not yet sure how to best implement all of this.
But in general: yes, I think the main i2s_audio component should be the one containing (and probably setting) all necessary options for I2S, ensuring interlocks etc.

BTW I already pinged you on Discord: Can you please open a PR for your 'add_i2s_mclk' branch? We will definitely need that.

rpatel3001:
I think the ideal way forward is a major refactor of media_player to not depend on arduino, and of the i2s_audio components in general to centralize i2s bus setup. It's probably worth sketching out use cases and doing some real architecture work with nabu casa folks before writing too much code. Especially since I hope we'll eventually get esp-adf, esp-sr, wakenet, etc into esphome as well.

sehraf:
> to centralize i2s bus setup. It's probably worth sketching out use cases and doing some real architecture work

i second that, https://github.com/esphome/esphome/pull/4884 already starts to change parameters

nielsnl68:
PLease resolve the conflicts.

kroimon:
This might not be necessary anymore as @jesserockz is working on the S3 Box and we can now make use of the components from esp-adf.

sehraf:
> This might not be necessary anymore as @jesserockz is working on the S3 Box and we can now make use of the components from esp-adf.

Can I track the process somewhere? Will this also include the ac101 by any chance?

PR #4866, opened 2023-05-20
Added base64 helper

freekode:

Added encode and decode funcitons for base64
Copied from [here](https://stackoverflow.com/questions/180947/base64-decode-snippet-in-c)

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2935

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
esp32_camera:
  ...
  on_image:
      then:
        - lambda: |-
            const char* image_data_char = reinterpret_cast<char*>(image.data);
            std::string encodedData = base64_encode(&image_data_char[0], image.length);
            bool result = id(mqtt_client).publish("cam/image", encodedData);
```

## Checklist:


freekode:
I'm not sure which tests to add.
I tested it locally only.

max246:
Maybe would be good to add a note in the esphome doc? If it gets approved is handy to know that i can be used for espcamera or just generic encoding.

freekode:
Sure thing. I added explanation to esp32 camera docs. If you think there is another better place, jut let me know.

kroimon (approved):
Looks good to me!

esphome/core/helpers.h
```diff
@@ -429,6 +429,12 @@ std::string value_accuracy_to_string(float value, int8_t accuracy_decimals);
 /// Derive accuracy in decimals from an increment step.
 int8_t step_to_accuracy_decimals(float step);
 
+static const std::string BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                                        "abcdefghijklmnopqrstuvwxyz"
+                                        "0123456789+/";
+std::string base64_encode(const char *buf, unsigned int buf_len);
```
jesserockz:
```suggestion
std::string base64_encode(const uint8_t *buf, size_t buf_len);
```

jesserockz: Might be nice to also create overloads for these to they are consistent, ie decoding can take a buffer to write to and return the length as other buffer apis do.

And an overload to pass a std::vector to `encode` as well

jesserockz (changes requested):

jesserockz (approved):

PR #4884, opened 2023-05-24
Allow i2s microphone bits per sample to be configured

jesserockz:

This brings a significant increase in audio quality for some microphones and allows others to work at all.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2943

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
+++ b/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
@@ -132,11 +132,10 @@ void I2SAudioMicrophone::read_() {
     return;
   }
   samples.resize(samples_read);
-  for (size_t i = 0; i < samples_read; i++) {
-    if (this->bits_per_sample_ == I2S_BITS_PER_SAMPLE_16BIT) {
-      samples[i] = reinterpret_cast<int16_t *>(this->buffer_)[i];
-      continue;
-    } else if (this->bits_per_sample_ == I2S_BITS_PER_SAMPLE_32BIT) {
+  if (this->bits_per_sample_ == I2S_BITS_PER_SAMPLE_16BIT) {
+    memcpy(samples.data(), this->buffer_, bytes_read);
+  } else if (this->bits_per_sample_ == I2S_BITS_PER_SAMPLE_32BIT) {
+    for (size_t i = 0; i < samples_read; i++) {
       int32_t temp = reinterpret_cast<int32_t *>(this->buffer_)[i] >> 14;
       samples[i] = clamp<int16_t>(temp, INT16_MIN, INT16_MAX);
       continue;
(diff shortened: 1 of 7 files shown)


esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
```diff
@@ -48,7 +48,7 @@ void I2SAudioMicrophone::start_() {
   i2s_driver_config_t config = {
       .mode = (i2s_mode_t) (I2S_MODE_MASTER | I2S_MODE_RX),
       .sample_rate = 16000,
-      .bits_per_sample = I2S_BITS_PER_SAMPLE_16BIT,
+      .bits_per_sample = I2S_BITS_PER_SAMPLE_32BIT,
```
sehraf: Doesn't this break external ADCs, which expect 16 bits per sample and (currently) can't be reconfigured?

jesserockz: Potentially maybe...
I guess I can make it configurable and dynamically scale the data if you have it set to not 16-bit which was on my mind to do.
Thanks for the push =)

jesserockz:
Thanks for the feedback @sehraf, I have now made the input bits configurable and it will scale down to 16-bit before passing to other components like voice_assistant.

kroimon:
While I like the idea very much in general, I wonder if we should really add the new `bits_per_sample` option to the microphone or if it wouldn't be better to add it to the base `i2s` component.

The current I2S media player aside, the sample rate and bit depth are inherent parameters of the bus itself and if we want to move to a more flexible implementation allowing full-duplex use of microphone and speaker, this might be a breaking change later.
For reference, [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s3/api-reference/peripherals/i2s.html#full-duplex) is the latest esp-idf i2s driver documentation for full-duplex mode for the ESP32-S3.

kroimon:
Also, having these settings go through the main `i2c` component as a hub would allow components interfacing external ADC/DAC audio codecs to get the current configuration from there to be sent to the codecs as those require having a matching configuration.

jesserockz:
> While I like the idea very much in general, I wonder if we should really add the new bits_per_sample option to the microphone or if it wouldn't be better to add it to the base i2s component.

ESPHome does not support duplex (And I dont see a need for it in the near future) and each component (microphone/media_player) reconfigure the i2s bus with their own individual config. This allows mics and speakers of different bits/channels etc to be able to work "together" on the same board.

The only reason for the common i2s_audio component with the clock pins is so that multiple components do not try to (re)configure the bus at the same time and it holds a lock per component to prevent this.

PR #4910, opened 2023-06-01
airthings_wave: refactor to eliminate code duplication

kpfleming:

Refactoring of `airthings_wave_mini` and `airthings_wave_plus` components into a new `airthings_wave_base` component, eliminating code duplication.

## Types of changes

- [X] Other (refactoring only)

## Test Environment

- [X] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/airthings_wave_base/airthings_wave_base.cpp
+++ b/esphome/components/airthings_wave_base/airthings_wave_base.cpp
@@ -1,0 +1,83 @@
+#include "airthings_wave_base.h"
+
+#ifdef USE_ESP32
+
+namespace esphome {
+namespace airthings_wave_base {
+
+static const char *const TAG = "airthings_wave_base";
+
+void AirthingsWaveBase::gattc_event_handler(esp_gattc_cb_event_t event, esp_gatt_if_t gattc_if,
+                                            esp_ble_gattc_cb_param_t *param) {
+  switch (event) {
+    case ESP_GATTC_OPEN_EVT: {
+      if (param->open.status == ESP_GATT_OK) {
+        ESP_LOGI(TAG, "Connected successfully!");
+      }
+      break;
+    }
+
+    case ESP_GATTC_DISCONNECT_EVT: {
+      ESP_LOGW(TAG, "Disconnected!");
+      break;
+    }
+
+    case ESP_GATTC_SEARCH_CMPL_EVT: {
+      this->handle_ = 0;
+      auto *chr = this->parent()->get_characteristic(this->service_uuid_, this->sensors_data_characteristic_uuid_);
+      if (chr == nullptr) {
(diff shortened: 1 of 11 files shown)


esphome/components/airthings_wave_mini/sensor.py
```diff
@@ -1,82 +1,26 @@
...
     CONF_ID,
-    CONF_HUMIDITY,
-    CONF_TVOC,
-    CONF_PRESSURE,
-    CONF_TEMPERATURE,
-    UNIT_PARTS_PER_BILLION,
-    ICON_RADIATOR,
 )
 
-DEPENDENCIES = ["ble_client"]
```
jesserockz, line 22: Please leave the DEPENDENCIES here as well as they show up nicer as errors in the editor on the real components

esphome/components/airthings_wave_base/airthings_wave_base.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+void AirthingsWaveBase::request_read_values_() {
+  auto status = esp_ble_gattc_read_char(this->parent()->get_gattc_if(), this->parent()->get_conn_id(), this->handle_,
+                                        ESP_GATT_AUTH_REQ_NONE);
+  if (status) {
+    ESP_LOGW(TAG, "Error sending read request for sensor, status=%d", status);
+  }
+}
+
+AirthingsWaveBase::AirthingsWaveBase() : PollingComponent(10000) {}
```
jesserockz: This should not have actually been in the main component, safe to remove.
```suggestion
```

jesserockz (changes requested):

esphome/components/airthings_wave_mini/sensor.py
kpfleming: Sure, I can just `import` that from the base into the real components.

esphome/components/airthings_wave_base/airthings_wave_base.cpp
kpfleming: Does this mean the `PollingComponent` constructor does not need to be called, or that it should be called from the constructors of the real components?

jesserockz: It doesn't need to be called, the update interval is set inside `cg.register_component`

esphome/components/airthings_wave_mini/sensor.py
kpfleming: Will push an update now.

esphome/components/airthings_wave_base/airthings_wave_base.cpp
kpfleming: Will push an update now.

kpfleming:
@jesserockz Requested changes have been made and CI is passing again. Thanks for the review!

jesserockz (approved):

PR #4911, opened 2023-06-02
Sigma delta fix

PlainTechEnthusiast:

The output pin, if configured, in the Sigma Delta Output was never initialized. This worked as long as the pin had been used as an output since power up since a software reset after an OTA didn't seem to reset the pin registers. Though I'm sure that last bit is probably expected behavior.

This PR also implements dump_config() and adds a missing header for automation.h.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/sigma_delta_output/sigma_delta_output.cpp
+++ b/esphome/components/sigma_delta_output/sigma_delta_output.cpp
@@ -27,5 +27,31 @@ void SigmaDeltaOutput::dump_config() {
   LOG_FLOAT_OUTPUT(this);
 }
 
+void SigmaDeltaOutput::update() {
+  this->accum_ += this->state_;
+  const bool next_value = this->accum_ > 0;
+
+  if (next_value) {
+    this->accum_ -= 1.;
+  }
+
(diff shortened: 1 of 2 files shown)


PlainTechEnthusiast:
I should be able to take care of the lint issue later today and amend the PR

HxYxI:
-

HxYxI (approved):
Text

PlainTechEnthusiast:
Ready for review

jesserockz (changes requested):
Since you are making changes, could you also move the `update` function into the cpp file?

Thanks

PlainTechEnthusiast:
> Since you are making changes, could you also move the `update` function into the cpp file?

Done and tested on an ESP32. I went back and forth on including it initially, I guess I could have just included it as a separate commit.

Thanks

jesserockz (approved):

PR #4916, opened 2023-06-04
Initial support for power management

silverchris:

Enables lightsleep/Modem sleep/Tickless Idle for power savings
On my test ESP32, it drops power consumption from 40mA continuous to 10mA average with bursts to 40mA.

I am marking this as a breaking change, as it will need components to be updated to work with it, though it shouldn't break anything unless it's enabled.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
pm:
  min_freq_mhz: 40
  max_freq_mhz: 240
  tickless: True

```

## Checklist:


silverchris:
Resolves https://github.com/esphome/feature-requests/issues/371

Open to suggestions on how to make this fit in better, as well

Unfocused:
I'm wondering how this could fit in with dedicated power management chips, like the AXP192 that's fairly common on devices from M5Stack and some other manfuacturers of ESP32 devkits with both batteries and power-hungry peripherals. 

There's a few different components floating around to add support for that chip:

- https://github.com/martydingo/esphome-axp192
- https://github.com/leoedin/esphome-axp192
- https://github.com/Pionizer/pionizer-axp192
- https://github.com/airy10/esphome-m5stickC/tree/master/components/axp192
- https://gitlab.com/geiseri/esphome_extras/-/tree/main/components/axp192

I've seen (but not yet needed) components to add support for a couple of other power mangement chips too. And I have at least one ESP32 devkit that just uses a GPIO pin to enable/disable power to peripherals for power saving.

However, there's not much agreement/consistency on how components for these chips are implemented or configured. And currently those power management components are configured separately and differently from this power management component, which seems both messy and confusing. I wonder if this PR could help there. 

Given there's a bunch of different power mangement chips, it'd make sense for each of those chips to be a platform under a power management domain. In which case, I wonder if this PR makes more sense as a ESP32 power management platform. ie:

```yaml
power_management:
  - platform: esp32_pm
    min_freq_mhz: 40
    max_freq_mhz: 240
    tickless: True
```

(As an aside: `pm` is very ambiguous. Would be much clearer if named something like `power_management`.)

AXP192 would be another platform, as would other CPUs like RP2040. 

In addition to controlling power to peripherals, chips like AXP192 can usually monitor things like power usage, battery life, battery charging, etc. So it'd be great if this PR allowed some automation to configure the power mangement at runtime, even if it's just something simple like `power_management.enable` and `power_management.disable` actions. That would allow something like:

```yaml
power_management:
  - platform: esp32_pm
    id: pm_esp32
    min_freq_mhz: 10
    max_freq_mhz: 40
  - platform: axp192
    id: pm_axp192

binary_sensor:
  - platform: axp192_batterypowered
    on_press:
      - power_management.enable: pm_esp32
      - power_management.enable: pm_axp192
    on_release:
      - power_management.disable: pm_esp32
      - power_management.disable: pm_axp192
```

(That obviously needs to be better named, since "enabling" for this PR means reducing power usage, but "enabling" for the AXP192 means increasing power usage.)

silverchris:
@Unfocused 
That does make some sense. I had been roughly thinking about how to support the RP2040, but I haven't played with it's features for sleep and power management yet, though I do have a couple in my drawer waiting for me :)

The AXP192 is fairly different from what I was thinking about with this, but I do think it fits under the same umbrella.

Will do some thinking!

I also completely agree with renaming it from pm to power_management

EternityForest:
> I also completely agree with renaming it from pm to power_management

If it's not going to cover anything but the CPU itself, maybe it could even be cpu_power_management?


External peripherals like sensors and stuff seem really hard to handle in a general way. Some need start time, you have to not try to read data from things that aren't ready,  so something like an ADC has to know if its connected to something that needs power...


Should there be a advanced WakeLock type system, where componentscan request a specific domain be turned off(Sensors, high power sensors, motors, etc), and another component can say "Nope, not ready yet", and nodes can track changes in state and respond appropriately?





Or what about just adding a full general purpose state machine component with rules, events, and timers?  I don't like to use "Pure" state machines, l like to extend them with timers.

Your code would look like 

```
statemachine:
    - name: foo
      default: on

      states:
            on:
                timer: 2000
                on_enter: (Turn on your stuff here, set up PM)
                rules:
                    timer: off

           off:
                on_enter: (Set up low power stuff here)
                rules:
                    button: 
                        goto: on
                        condition: (Some lambda that disabled turning on)

# Trigger the button event when someone presses the button 
```

And as a bonus you can expose the current state of the state machine to the frontend for visibility.


Maybe you could somehow give *Every* component a call to pause it, so loop never happens, so that you could manually block things that shouldn't happen while a peripheral is off?  The majority of polling sensors could be paused like that without rewriting everything.

Doing it all manually in a state machine would be nice and explicit, lend itself well to GUI tools later, and solve a lot of other problems too.


Maybe you could have a global "Event Bus" for simple predefined enum events, and just define one for "Change power level', that could trigger all SMs?

Edit: There's already a state machine library and probably others(https://github.com/muxa/esphome-state-machine) but I would rather give it a go myself or fork it, I think state machines enhanced with timers are far more useful.

They do look like they plan to add conditions, but timers are WONTFIX(https://github.com/muxa/esphome-state-machine/issues/40), they want you to use delays in an action instead, which I don't think is ideal for trying to use it as the standard way to do power management, the syntax is a bit spaghettiful.

EternityForest:
Should  we hold an ESP_PM_APB_FREQ_MAX lock when using UART frequencies above 500k that are too much for the TICK clock source?  I don't have time for extensive testing but I think I have it done: https://github.com/EternityForest/esphome/commit/e88a6cb725c10368b1418526ea651e4351ca609e

esphome/components/esp32_pm/esp32_pm.h
```diff
@@ -0,0 +1,51 @@
...
+  float get_setup_priority() const override { return setup_priority::BUS; }
+  void set_tickless(bool tickless) override;
+  void loop() override;
+  void dump_config() override;
+
+  std::unique_ptr<power_management::PMLock> get_lock(std::string name, power_management::PmLockType lock) override;
+
+ private:
+  uint16_t min_freq_ = 40;
+  uint16_t max_freq_ = 240;
```
luar123, line 41: Defaults depend on ESP variant. I would suggest to set both to `CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ`. With this DFS would be disabled by default and it would work for all variants.

esphome/components/esp32_pm/power_management.py
```diff
@@ -0,0 +1,34 @@
...
+
+pm_ns = cg.esphome_ns.namespace("esp32_pm")
+PM = pm_ns.class_("ESP32PowerManagement", cg.Component)
+
+CONFIG_SCHEMA = cv.Schema(
+    {
+        cv.GenerateID(): cv.declare_id(PM),
+        cv.Optional(CONF_MIN_FREQ_MHZ, default=40): cv.uint16_t,
+        cv.Optional(CONF_MAX_FREQ_MHZ, default=240): cv.uint16_t,
+        cv.Optional(CONF_TICKLESS, default=False): cv.boolean,
```
luar123, line 17: Defaults depend on ESP variant. I would suggest to set both to CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ. With this DFS would be disabled by default and it would work for all variants.

esphome/components/esp32_pm/esp32_pm.cpp
```diff
@@ -0,0 +1,142 @@
...
+std::unique_ptr<power_management::PMLock> ESP32PowerManagement::get_lock(std::string name,
+                                                                         power_management::PmLockType lock) {
+  return make_unique<ESPPMLock>(name, lock);
+}
+
+ESPPMLock::ESPPMLock(const std::string &name, power_management::PmLockType lock) {
+  name_ = name;
+  lock_ = lock;
+#ifdef CONFIG_PM_ENABLE
+  esp_pm_lock_create(ESP_PM_NO_LIGHT_SLEEP, 0, "Output Lock", &this->pm_lock_);
```
luar123, line 121: Is it a good idea to always create/delete locks?

luar123 (changes requested):
This is exactly what I was looking for, any plans to merge it?
I am working on zigbee support with esp32h2 and with PM enabled the idle current consumption could drop from ~30mA to close to 0 without deep-sleep.

Hedda:
FYI, new battery-powered Seeed Studio IoT Button is based on an ESP32-C6 SoC; ”_supports ESPHome or Zigbee firmware_” (though their provided Zigbee firmware is not based on ESPHome, this hardware might be good proof-of-concept as it as can technically support WiFi, Zigbee, or Thread):

- https://www.cnx-software.com/2025/04/13/battery-powered-seeed-studio-esp32-c6-iot-button-esphome-zigbee-firmware/

- https://www.seeedstudio.com/Seeed-IoT-Button-p-6419.html

- https://github.com/Seeed-Projects/Seeed_IoT_Button

- https://wiki.seeedstudio.com/getting_started_with_seeed_iot_button/

- https://files.seeedstudio.com/wiki/iot_button_zigbee/Seeed_Studio_IoT_Button_PCB.zip

- https://www.aliexpress.com/item/1005008801906619.html
- 

image

Hedda:
Any updates on this?

PR #4926, opened 2023-06-08
Move ESPTime into core esphome namespace

jesserockz:

This moves the `ESPTime` class into a core "helper" file so it can be used without `time` components being required.

This is a breaking change for any `external_components` that use `ESPTime` as it has moved out of the `time::` namespace, and should now use `#include "esphome/core/time.h"`.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** Part of https://github.com/esphome/feature-requests/issues/2267

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


kbx81 (approved):
It's about time

PR #4928, opened 2023-06-08
Handling of sensor read quality refined

lukasl96:
It's now possible to receive sensor values even when the sensor thinks its read quality is low

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
+++ b/esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
@@ -17,6 +17,7 @@ void MopekaProCheck::dump_config() {
   LOG_SENSOR("  ", "Temperature", this->temperature_);
   LOG_SENSOR("  ", "Battery Level", this->battery_level_);
   LOG_SENSOR("  ", "Reading Distance", this->distance_);
+  LOG_SENSOR("  ", "Sensor Reading Quality", this->quality_level_);
 }
 
 /**
@@ -70,17 +71,15 @@ bool MopekaProCheck::parse_device(const esp32_ble_tracker::ESPBTDevice &device) 
     uint32_t distance_value = this->parse_distance_(manu_data.data);
     SensorReadQuality quality_value = this->parse_read_quality_(manu_data.data);
     ESP_LOGD(TAG, "Distance Sensor: Quality (0x%X) Distance (%dmm)", quality_value, distance_value);
-    if (quality_value < QUALITY_HIGH && quality_value > QUALITY_LOW) {
-      ESP_LOGW(TAG, "Possibly poor sensor read quality.");
-    }
-    if (quality_value < QUALITY_MED && quality_value > QUALITY_NONE) {
-      ESP_LOGW(TAG, "Poor sensor read quality.");
-    }
     if (quality_value <= QUALITY_NONE) {
       // if really bad reading set to 0
       ESP_LOGW(TAG, "Really bad sensor read quality.");
       ESP_LOGW(TAG, "Setting distance to 0");
       distance_value = 0;
+    } else if (quality_value < QUALITY_MED ) {
+      ESP_LOGW(TAG, "Poor sensor read quality.");
+    } else if (quality_value < QUALITY_HIGH) {
+      ESP_LOGW(TAG, "Possibly poor sensor read quality.");
     }
 
     // update distance sensor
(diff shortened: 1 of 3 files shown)


esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
```diff
@@ -70,11 +70,15 @@ bool MopekaProCheck::parse_device(const esp32_ble_tracker::ESPBTDevice &device)
     uint32_t distance_value = this->parse_distance_(manu_data.data);
     SensorReadQuality quality_value = this->parse_read_quality_(manu_data.data);
     ESP_LOGD(TAG, "Distance Sensor: Quality (0x%X) Distance (%dmm)", quality_value, distance_value);
-    if (quality_value < QUALITY_HIGH) {
-      ESP_LOGW(TAG, "Poor read quality.");
+    if (quality_value < QUALITY_HIGH && quality_value > QUALITY_LOW) {
```
spbrogan: If i am reading this correctly there is an odd overlap for these conditionals and log messages.  I would suggest a more traditional if/else if structure and print a unique log message for each quality.

esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
```diff
@@ -70,11 +70,15 @@ bool MopekaProCheck::parse_device(const esp32_ble_tracker::ESPBTDevice &device)
...
+      ESP_LOGW(TAG, "Possibly poor sensor read quality.");
     }
-    if (quality_value < QUALITY_MED) {
+    if (quality_value < QUALITY_MED && quality_value > QUALITY_NONE) {
+      ESP_LOGW(TAG, "Poor sensor read quality.");
```
spbrogan: instead of allowing low quality readings without letting the user know that their sensor is reading poorly I wonder if we want to something different. 

One idea is we could expose the read quality as a sensor.  That way the HA user can at least see what the quality reading is.  

and/or 

create a config item per sensor that allows you to specify the minimum reading quality you want reported in the level and distance sensors.
or
 create multiple distance and level sensors pairs.  1 pair for high only.  1 pair for med and above.  1 pair for low and above.  Then you could choose which sensor pair you wanted.  

and/or 

Enable some algorithm to ignore N number of bad readings and not change the level or distance sensor on a bad reading unless there have been no better readings within N reports. 

Thoughts?

spbrogan (changes requested):

esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
lukasl96: I already have a custom component with sensor reading quality as an own sensor. I will add the code later this day.

lukasl96: I committed the changes but the tests are failing.
I don't understand why they are failing because i use the same code as a custom component in my environment.

esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
```diff
@@ -94,6 +97,20 @@ bool MopekaProCheck::parse_device(const esp32_ble_tracker::ESPBTDevice &device)
...
+        quality_level_percent = 100;
+      } else if (quality_value = QUALITY_MED) {
+        quality_level_percent = 66.6;
+      } else if (quality_value = QUALITY_LOW) {
+        quality_level_percent = 33.3;
+      } else {
+        quality_level_percent = 0;
+      }
+      this->quality_level_->publish_state(quality_level_percent);
+    }
```
jesserockz, line 113: I don't like the fact this is converting the fixed enums to a fixed percentage.

I feel like this would just be better as a text_sensor with the names as they are.

esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
```diff
@@ -94,6 +97,20 @@ bool MopekaProCheck::parse_device(const esp32_ble_tracker::ESPBTDevice &device)
       }
       this->level_->publish_state(tank_level);
     }
+    // update quality level sensor
+    if (this->quality_level_ != nullptr) {
+      uint8_t quality_level_percent = 0;
+      if (quality_value = QUALITY_HIGH) {
```
spbrogan, line 103: these should all be "==" as you are doing assignment rather than comparison.

esphome/components/mopeka_pro_check/mopeka_pro_check.cpp
spbrogan: you could also just return the quality value which is int between 0 - 3

spbrogan:
I kind of like the idea of adding an optional config item of value between 0-3 which could be the quality reading in which to not report.  And maybe another optional config item of the value between 0-3 which all quality readings below will be reported as zero.

phurth:
I like the idea of passing the quality value and allowing the user to configure whether or not to ignore at a certain level. This better aligns with the Mopeka app behavior, where low (1) quality readings are still reported to the user. I have one sensor that often reports a quality of 1, but still seems accurate enough. I'd like the ability to allow that reading, but maybe still ignore 0 quality readings.

PR #4710, opened 2023-04-19
Add runtime online image support

guillempages:
# What does this implement/fix?

This PR implements the possibility of downloading PNG images at runtime and showing them in a compatible display.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/1463

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2076
## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
esphome:
  # ...
  on_boot:
    then:
      - wait_until: api.connected
      - component.update: example_image
      - 
#http_request needs to be explicitly added and configured:
http_request:
  #Arduino does not support SSL verification; on arduino this must be explicitly disabled (on ESP-IDF it's not necessary).
  verify_ssl: false

online_image:
  - url: "https://www.example.org/image.png"
    format: PNG
    id: example_image
    type: RGB565
    on_download_finished:
      - component.update: my_display
    on_error:
      - logger.log: "Could not download the image"

display:
  id: my_display
  # ...
  lambda: |-
    it.image(0, 0, id(example_image));
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4931, opened 2023-06-10
display: Improve Image rendering by removing usage of virtual functions

ayufan:

The virtual functions have high overhead. In case of image rendering they are being called for each pixel. This change improves rendering by about 5%.

Since Image does no longer expose virtual functions for pixels, fix Animation to shift `data_buffer`.

For 150x150 image, with RGB565:

- before: `28987 us`
- after: `30600 us`

It was tested on ESP32S3.

## Types of changes

- [x] Other: Performance improvement

## Test Environment

- [ ] ESP32
- [x] ESP32S3
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
animation:
  - id: image_startup
    file: "images/AfterPrinting/AfterPrinting.gif"
    resize: 150x150
    type: RGB565

display:
  lambda: |-
    id(image_startup).next_frame();
    auto start_us = micros();
    it.image(0, 0, id(image_startup));
    ESP_LOGE("display", "image draw: %ld", dt_us);
```

## Checklist:

### DIFF
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -15,6 +15,25 @@ static const char *const TAG = "display";
 const Color COLOR_OFF(0, 0, 0, 255);
 const Color COLOR_ON(255, 255, 255, 255);
 
+static int image_type_to_bpp(ImageType type) {
+  switch (type) {
+    case IMAGE_TYPE_BINARY:
+      return 1;
+    case IMAGE_TYPE_GRAYSCALE:
+      return 8;
+    case IMAGE_TYPE_RGB565:
+      return 16;
+    case IMAGE_TYPE_RGB24:
+      return 24;
+    case IMAGE_TYPE_RGBA:
+      return 32;
+    default:
+      return 0;
+  }
+}
+
+static int image_type_to_width_stride(int width, ImageType type) { return (width * image_type_to_bpp(type) + 7u) / 8u; }
+
 void Rect::expand(int16_t horizontal, int16_t vertical) {
   if (this->is_set() && (this->w >= (-2 * horizontal)) && (this->h >= (-2 * vertical))) {
     this->x = this->x - horizontal;
@@ -755,35 +774,7 @@ void Animation::set_frame(int frame) {
 }
 
 void Animation::update_data_start_() {
-  int bpp = 0;
-
-  switch (this->type_) {
-    case IMAGE_TYPE_BINARY:
-      bpp = 1;
-      break;
-
-    case IMAGE_TYPE_GRAYSCALE:
-      bpp = 8;
-      break;
-
-    case IMAGE_TYPE_RGB565:
-      bpp = 16;
-      break;
-
-    case IMAGE_TYPE_RGB24:
-      bpp = 24;
-      break;
-
-    case IMAGE_TYPE_RGBA:
-      bpp = 32;
-      break;
-
-    default:
-      break;
-  }
-
-  const uint32_t width_aligned = (this->width_ * bpp + 7u) / 8u;
-  const uint32_t image_size = width_aligned * this->height_;
+  const uint32_t image_size = image_type_to_width_stride(this->width_, this->type_) * this->height_;
   this->data_start_ = this->animation_data_start_ + image_size * this->current_frame_;
 }
 
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -699,78 +699,9 @@ Image::Image(const uint8_t *data_start, int width, int height, ImageType type)
     : width_(width), height_(height), type_(type), data_start_(data_start) {}
 int Image::get_current_frame() const { return 0; }
 
-bool Animation::get_pixel(int x, int y) const {
-  if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)
-    return false;
-  const uint32_t width_8 = ((this->width_ + 7u) / 8u) * 8u;
-  const uint32_t frame_index = this->height_ * width_8 * this->current_frame_;
-  if (frame_index >= (uint32_t) (this->width_ * this->height_ * this->animation_frame_count_))
-    return false;
-  const uint32_t pos = x + y * width_8 + frame_index;
-  return progmem_read_byte(this->data_start_ + (pos / 8u)) & (0x80 >> (pos % 8u));
-}
-Color Animation::get_rgba_pixel(int x, int y) const {
-  if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)
-    return Color::BLACK;
-  const uint32_t frame_index = this->width_ * this->height_ * this->current_frame_;
-  if (frame_index >= (uint32_t) (this->width_ * this->height_ * this->animation_frame_count_))
-    return Color::BLACK;
-  const uint32_t pos = (x + y * this->width_ + frame_index) * 4;
-  return Color(progmem_read_byte(this->data_start_ + pos + 0), progmem_read_byte(this->data_start_ + pos + 1),
-               progmem_read_byte(this->data_start_ + pos + 2), progmem_read_byte(this->data_start_ + pos + 3));
-}
-Color Animation::get_color_pixel(int x, int y) const {
-  if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)
-    return Color::BLACK;
-  const uint32_t frame_index = this->width_ * this->height_ * this->current_frame_;
-  if (frame_index >= (uint32_t) (this->width_ * this->height_ * this->animation_frame_count_))
-    return Color::BLACK;
-  const uint32_t pos = (x + y * this->width_ + frame_index) * 3;
-  Color color = Color(progmem_read_byte(this->data_start_ + pos + 0), progmem_read_byte(this->data_start_ + pos + 1),
-                      progmem_read_byte(this->data_start_ + pos + 2));
-  if (color.b == 1 && color.r == 0 && color.g == 0 && transparent_) {
-    // (0, 0, 1) has been defined as transparent color for non-alpha images.
-    // putting blue == 1 as a first condition for performance reasons (least likely value to short-cut the if)
-    color.w = 0;
-  } else {
-    color.w = 0xFF;
-  }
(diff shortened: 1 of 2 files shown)


nielsnl68:
Greate work @ayufan,

Do you think we can simplify the `get_color()` a bit more by combining those four in to one function?

ayufan:
@nielsnl68 I have to think about this, maybe in next PR, since there's an overhead for `switch`. I'm playing with various ways to speed-up the bitblit since this code related to rendering is crazy slow, and it is possible to make it easily 20x faster.

So, far I'm unifying everything to be derivative of `Image`, next will be making `Glyph` rendering to be also done through `Image`.

esphome/components/display/display_buffer.cpp
```diff
@@ -797,12 +728,16 @@ void Animation::next_frame() {
     this->loop_current_iteration_ = 1;
     this->current_frame_ = 0;
   }
+
+  this->update_data_start_();
 }
 void Animation::prev_frame() {
```
nielsnl68, line 753: I'm missing the `next_frame()` method or am i overlooking it?

esphome/components/display/display_buffer.cpp
```diff
@@ -815,6 +750,41 @@ void Animation::set_frame(int frame) {
       this->current_frame_ = this->animation_frame_count_ - abs_frame;
     }
   }
+
+  this->update_data_start_();
+}
+
+void Animation::update_data_start_() {
```
nielsnl68, line 776: It would be nice when this function moves to the  `Image` class. Where current_frame_ is always set to **0**.

esphome/components/display/display_buffer.cpp
```diff
@@ -815,6 +750,41 @@ void Animation::set_frame(int frame) {
...
+    case IMAGE_TYPE_RGBA:
+      bpp = 32;
+      break;
+
+    default:
+      break;
+  }
+
+  const uint32_t width_aligned = (this->width_ * bpp + 7u) / 8u;
+  const uint32_t image_size = width_aligned * this->height_;
```
nielsnl68: When you try to speed up this classes, i think you should also try not to create to many unneeded variables just to calculate a result, are you not agree?

esphome/components/display/display_buffer.h
```diff
@@ -532,11 +532,11 @@ class Font {
...
-  virtual bool get_pixel(int x, int y) const;
-  virtual Color get_color_pixel(int x, int y) const;
-  virtual Color get_rgba_pixel(int x, int y) const;
-  virtual Color get_rgb565_pixel(int x, int y) const;
-  virtual Color get_grayscale_pixel(int x, int y) const;
+  bool get_pixel(int x, int y) const;
+  Color get_color_pixel(int x, int y) const;
+  Color get_rgba_pixel(int x, int y) const;
+  Color get_rgb565_pixel(int x, int y) const;
+  Color get_grayscale_pixel(int x, int y) const;
```
nielsnl68, line 539: As wrote above, maybe we can combine this functions to one as well?

nielsnl68 (changes requested):
Suc6

esphome/components/display/display_buffer.cpp
ayufan: A few lines up.

esphome/components/display/display_buffer.cpp
ayufan: @nielsnl68 This will not work. This method is only applicable to `Animation`, since Animation knows that it has frames and Animation implements frames rotation.

esphome/components/display/display_buffer.h
ayufan: This is too big of a change for now. This color blitting needs to be completely reworked, and I would prefer to look at this separately.

The purpose of this PR is to do targeted change that does not break existing interfaces. Someone could have esphome integration that depends on `get_color_pixel`.

nielsnl68:
> So, far I'm unifying everything to be derivative of `Image`, next will be making `Glyph` rendering to be also done through `Image`.

Fantastic idea, makes my Switchplate work much simpler :D

BTW, pleace checkout @guillempages work as well and try to work together ;)

esphome/components/display/display_buffer.cpp
ayufan: It does not matter. Those are added for readability purposes, they are optimised by the compiler anyway.

guillempages:
I also spent some time trying to unify everything into "get_color" but found out that it was faster to actually keep the 4 methods separate 😕

I'd be glad to be proven wrong though.

Regarding "unvirtualizing" those methods, please hold on. I'm working in adding support for runtime images (i.e. downloaded at runtime), and that code is relying on the virtual methods 😕 (#4710)

There's might be a way around that, I'm not sure. My problem is that current images need to read from program memory, but the runtime images need to read from heap memory (or psram when available)

ayufan:
> Regarding "unvirtualizing" those methods, please hold on. I'm working in adding support for runtime images (i.e. downloaded at runtime), and that code is relying on the virtual methods 😕

@guillempages I have easy trick for you then: Pass execution to `image->draw()`.

```c++
class DisplayBuffer {
public:
  void image(int x, int y, BaseImage *image, Color color_on = COLOR_ON, Color color_off = COLOR_OFF) {
    image->draw(this, x, y, color_on, color_off);
  }

class BaseImage {
public:
  virtual void draw(DisplayBuffer *buffer, int x, int y...) = 0;
};

class Image : BaseImage {
public:
}

class CompressedImage : BaseImage {};
```

It will be cheaper to decode image online and render it right away, since the larger images will use a lot of memory. The `CompressedImage` could then figure out the best way to decode in a target format of `DisplayBuffer` if it starts exposing what this format is.

We pay heafty price for reading those one-by-one today.

guillempages:
Ok, I was just answering blindly to @nielsnl68 comment. I'll need to take a look at your proposal then, and see if I can leverage that. 🙂

nielsnl68:
> Ok, I was just answering blindly to @nielsnl68 comment. I'll need to take a look at your proposal then, and see if I can leverage that. 🙂

that is way i pinged you on this PR  👍🏻

esphome/components/display/display_buffer.cpp
```diff
@@ -15,6 +15,25 @@ static const char *const TAG = "display";
...
+    case IMAGE_TYPE_GRAYSCALE:
+      return 8;
+    case IMAGE_TYPE_RGB565:
+      return 16;
+    case IMAGE_TYPE_RGB24:
+      return 24;
+    case IMAGE_TYPE_RGBA:
+      return 32;
+    default:
+      return 0;
```
guillempages, line 31: I'm not sure if it is a good idea to set the default to 0.
Maybe not having a default case at all? Depending on how the compilation warnings are setup, this should fail compilation (or at least give a warning) if a new type is added to the enum, and not added to the switch.

ayufan: Yes, this is good idea, to ensure that all cases are always covered.

esphome/components/display/display_buffer.h
guillempages: Also (as mentioned on another PR), having the image type check at runtime has a ~10% performance penalty.
I did some tests regarding that on https://github.com/esphome/esphome/pull/4600#discussion_r1153808915
(having a single `get_pixel` method that chooses which method to use internally basically amounts to the same as the "switch-within-for" use case)

I'd strongly advise against having this single `get_pixel` (unless you are able to find some performance trick to avoid having to check the image type at every pixel)

ayufan: The current implementation of `display` is far from being performance focused. It is to be signifcantly improved, since working on individual pixels will newer work fast. Here you can find twisting the problem around: https://github.com/esphome/esphome/pull/4932 and gaining a few more %.

guillempages:
Now I did look at the code :-D

In general it looks good, but I'm still concerned about removing the "virtual" calls; since there might be external components (or the already mentioned PR #4710) that rely on the `get_xxx_pixel` methods being virtual.

I find the animation part a great idea though.

ayufan:
> In general it looks good, but I'm still concerned about removing the "virtual" calls; since there might be external components (or the already mentioned PR https://github.com/esphome/esphome/pull/4710) that rely on the get_xxx_pixel methods being virtual.

Well. I cannot test that :) I think that in general people do update their components with breaking changes from `esphome`. I did it a few times :)

esphome/components/display/display_buffer.h
guillempages: It's not about making it "fast", it's about not making it "slower" :-D

guillempages: Let's see where this leads

ayufan:
@guillempages

Thanks for the review of all PR. I appreciate all the comments :)

Can you guide me which one of those I should address, or should I wait for more feedback? There are code-style ones like handling `default:` and `ImageAlign::HORIZONTAL_ALIGN` which are great suggestions, but I'm unsure how to handle the functional suggestions.

guillempages:
@ayufan I'm not a maintainer of the repo, nor a code-owner of any part of it, so I cannot tell you which part will bring the greatest chances of success :-D @nielsnl68 mentioned me, because I have also done some modifications to the image and animation parts lately (and as I mentioned, I still have two of them open).

From experience, it takes a while until some senior reviewer looks at the PR. So you might want to start thinking about that, and maybe implementing some of them, if you think they are valid/worth it.

ayufan:
> From experience, it takes a while until some senior reviewer looks at the PR. So you might want to start thinking about that, and maybe implementing some of them, if you think they are valid/worth it.

I don't want to open too many and have them lingering :)

guillempages:
You can modify them locally, and do a force-push, so that you stay at three PRs. If the changes are directly related to one PR, it's ok to have more than one commit on that branch. You then need to rebase the other PRs, of course.

I also had three PRs that based on each other, and Jesse put the latter two to draft, so that it's clear that there is no point in reviewing them yet (since the code is part of the other PRs, *and* might change because of suggestions/requests on the previous ones). So I had to wait until each one got merged, and then rebase and set the next one to "ready for review".

jesserockz (approved):

PR #4932, opened 2023-06-10
display: add `BaseImage` and provide only `Image::get_pixel` method

ayufan:

This is based on top of: https://github.com/esphome/esphome/pull/4931.

- Does add `BaseImage::draw` which is called by `DisplayBuffer::image`
- Does hide all `Image::get_*_pixel` methods and makes them unsafe since they are called locally
- Does add `Image::get_pixel` and makes it safe
- Does add `Image::draw` to use unsafe methods to draw images quicker
- Makes it faster by about ~2.5%

For 150x150 image, with RGB565:

```
before: 28987 us
after: 28288 us
It was tested on ESP32S3.
```

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected): Changes `Image` interface.
- [x] Other: Improves performance

## Test Environment

- [ ] ESP32
- [x] ESP32S3
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -325,63 +325,8 @@ void DisplayBuffer::vprintf_(int x, int y, Font *font, Color color, TextAlign al
     this->print(x, y, font, color, align, buffer);
 }
 
-void DisplayBuffer::image(int x, int y, Image *image, Color color_on, Color color_off) {
-  bool transparent = image->has_transparency();
-
-  switch (image->get_type()) {
-    case IMAGE_TYPE_BINARY: {
-      for (int img_x = 0; img_x < image->get_width(); img_x++) {
-        for (int img_y = 0; img_y < image->get_height(); img_y++) {
-          if (image->get_pixel(img_x, img_y)) {
-            this->draw_pixel_at(x + img_x, y + img_y, color_on);
-          } else if (!transparent) {
-            this->draw_pixel_at(x + img_x, y + img_y, color_off);
-          }
-        }
-      }
-      break;
-    }
-    case IMAGE_TYPE_GRAYSCALE:
-      for (int img_x = 0; img_x < image->get_width(); img_x++) {
-        for (int img_y = 0; img_y < image->get_height(); img_y++) {
-          auto color = image->get_grayscale_pixel(img_x, img_y);
-          if (color.w >= 0x80) {
-            this->draw_pixel_at(x + img_x, y + img_y, color);
-          }
-        }
-      }
-      break;
-    case IMAGE_TYPE_RGB565:
-      for (int img_x = 0; img_x < image->get_width(); img_x++) {
-        for (int img_y = 0; img_y < image->get_height(); img_y++) {
-          auto color = image->get_rgb565_pixel(img_x, img_y);
-          if (color.w >= 0x80) {
-            this->draw_pixel_at(x + img_x, y + img_y, color);
-          }
-        }
-      }
-      break;
-    case IMAGE_TYPE_RGB24:
-      for (int img_x = 0; img_x < image->get_width(); img_x++) {
-        for (int img_y = 0; img_y < image->get_height(); img_y++) {
-          auto color = image->get_color_pixel(img_x, img_y);
-          if (color.w >= 0x80) {
-            this->draw_pixel_at(x + img_x, y + img_y, color);
-          }
-        }
-      }
-      break;
-    case IMAGE_TYPE_RGBA:
-      for (int img_x = 0; img_x < image->get_width(); img_x++) {
-        for (int img_y = 0; img_y < image->get_height(); img_y++) {
-          auto color = image->get_rgba_pixel(img_x, img_y);
-          if (color.w >= 0x80) {
-            this->draw_pixel_at(x + img_x, y + img_y, color);
-          }
-        }
-      }
-      break;
-  }
+void DisplayBuffer::image(int x, int y, BaseImage *image, Color color_on, Color color_off) {
+  image->draw(x, y, this, color_on, color_off);
 }
 
 #ifdef USE_GRAPH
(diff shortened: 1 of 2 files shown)


ayufan:
cc @guillempages @nielsnl68

ayufan:
@guillempages I would propose to add `CompressedImage : public BaseImage` that would figure out best how to decode and `draw` the image.

This would help making the `CompressedImage` to be optimised. I will play with `Image` to improve `Image::draw` to introduce `bitblt`.

esphome/components/display/display_buffer.cpp
```diff
@@ -637,23 +601,91 @@ Font::Font(const GlyphData *data, int data_nr, int baseline, int height) : basel
...
-  if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)
-    return Color::BLACK;
   const uint32_t pos = (x + y * this->width_) * 4;
   return Color(progmem_read_byte(this->data_start_ + pos + 0), progmem_read_byte(this->data_start_ + pos + 1),
                progmem_read_byte(this->data_start_ + pos + 2), progmem_read_byte(this->data_start_ + pos + 3));
 }
-Color Image::get_color_pixel(int x, int y) const {
-  if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)
-    return Color::BLACK;
+Color Image::get_rgb24_pixel(int x, int y) const {
```
ayufan: The `  if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)` is checked only in a public interface: `Image::get_pixel`.

esphome/components/display/display_buffer.cpp
```diff
@@ -637,23 +601,91 @@ Font::Font(const GlyphData *data, int data_nr, int baseline, int height) : basel
...
+        }
+      }
+      break;
+  }
+}
+Color Image::get_pixel(int x, int y, Color color_on, Color color_off) const {
   if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)
-    return false;
+    return color_off;
+  switch (this->type_) {
```
guillempages, line 663: Have you tried doing performance measurements on that?

I did try that approach IIRC on a previous PR (#4600) and found out that the overhead of the switch on every pixel was pretty noticeable. The measurements and tests I did should still be in a comment on that PR.

Yes, it makes the code more readable, having a generic `get_pixel`, but unfortunately it also makes it slower. 😕

esphome/components/display/display_buffer.cpp
```diff
@@ -637,23 +601,91 @@ Font::Font(const GlyphData *data, int data_nr, int baseline, int height) : basel
...
+          auto color = this->get_rgba_pixel_(img_x, img_y);
+          if (color.w >= 0x80) {
+            display->draw_pixel_at(x + img_x, y + img_y, color);
+          }
+        }
+      }
+      break;
+  }
+}
+Color Image::get_pixel(int x, int y, Color color_on, Color color_off) const {
```
guillempages, line 660: I see that this method is added "for convenience", but not actually used in the perfromance-related code (i.e. the hot loop).
That's good ;-)

It might be a good idea to add a comment to the method, explicitly mentioning that it introduces a performance penalty, and for usage within loops, the explicit methods should be used.

esphome/components/display/display_buffer.h
```diff
@@ -529,24 +529,33 @@ class Font {
   int height_;
 };
 
-class Image {
+class BaseImage {
```
guillempages, line 532: This is something that would need to be confirmed by @jesserockz, but IMHO it would be nice to split the "image" class(es) into its own file; specially now that it's getting more funcitonality. I.e. leave the generic display_buffer classes here, and create a new image.h/.cpp file with the image code.

What I'm not really sure is whether it would also be good to put those new files in the "image" directory (together with their __init__.py file) or leave them here.

The same could be done for animations, of course.

esphome/components/display/display_buffer.h
```diff
@@ -529,24 +529,33 @@ class Font {
...
 
   void set_transparency(bool transparent) { transparent_ = transparent; }
   bool has_transparency() const { return transparent_; }
 
  protected:
+  bool get_binary_pixel_(int x, int y) const;
+  Color get_rgb24_pixel_(int x, int y) const;
+  Color get_rgba_pixel_(int x, int y) const;
+  Color get_rgb565_pixel_(int x, int y) const;
+  Color get_grayscale_pixel_(int x, int y) const;
```
guillempages, line 557: Is there a reason why you are making those protected?
The public `get_pixel` method has a performance drawback, so it would be nice if users of the class could also be able to use the better-performing methods, instead of forcing a switch on every "get_pixel".

esphome/components/display/display_buffer.cpp
ayufan: Of course, any switch will cost a lot. This is why there's `Image::draw` that ideally should be implemented with direct bitblt.

esphome/components/display/display_buffer.h
ayufan: Of course I agree with it. The tricky part about new component is that the `Font` and `Image` are deeply integrated in interface of `DisplayBuffer`. But as a low hanging fruit, it might be good enough to introduce `image.h`, `font.h` and alike to split implementation.

esphome/components/display/display_buffer.h
ayufan: @guillempages You should never use those methods. Those methods are unsafe, thus why they are protected. They do not check `x`, `y`, nor compatibility with `type`.

Also when grepping the code I did not see anything that would use them directly. They were previously used by `DisplayBuffer::draw`, but this responsibility was migrated to in-class `Image::draw` method that can much better in the future optimise image buffer access.

The current `Image::draw` follows what is being done with `QrCode::draw` and `Graph::draw`.

guillempages: I was thinking of the end-customer; either for custom/external components, or in a lambda, where the image is being copied pixel-by-pixel. I agree that they are not safe, and should normally not be used, but actually explicitly *preventing* them from being used is overkill IMHO.

guillempages: and regarding the `Image::draw`, I have no objections to that; it might make some use cases easier. I'm not sure whether it then complicates the bitblt implementation though, since this would be strongly coupled to the display itself, and not the image. But that's something for the future; currently there is no bitblt implementation anyway.

ayufan: > where the image is being copied pixel-by-pixel

Unsure. The copy pixel-by-pixel regardless what people do will want this to do is a bad idea. The crazy overhead is in converting the image data into 32bit RGBA, and then converting this into display representation, which for cases like `Image::RGB565` and `Display::RGB565` results in a very slow bitblit.

You mentioned that this cost 10% on performance penalty, but have something fully safe. I think this is acceptable tradeoff for a niche use-case. I would rather force people to propose much better ways to manipulate pixel-data, since what is currently being done is very slow even on esp32s3 being very close to being not useable :) Like for 240x240 it can achieve at max 5FPS?

ayufan: > and regarding the Image::draw, I have no objections to that; it might make some use cases easier. I'm not sure whether it then complicates the bitblt implementation though, since this would be strongly coupled to the display itself, and not the image. But that's something for the future; currently there is no bitblt implementation anyway.

I have idea how to approach this, but I wanted first to better define responsibilities and improve what is offered now :)

esphome/components/display/display_buffer.h
ayufan: Actually! We could move now `Image` into `components/image/image.cpp`. Given the newly added `BaseImage`. Similar to `Font`. This would use abstract interface that would be implemented by `image::Image`.

guillempages: I've checked with Jesse on discord, and he gave the OK to the split into their own files. But the "new" image.cpp/h files should stay in the "components/display" directory, and not be moved to the "image" directory

ayufan: > But the "new" image.cpp/h files should stay in the "components/display" directory, and not be moved to the "image" directory

Do you have idea why?

guillempages: The discussion in discord mentioned:

> are they 'image handling for the display component'? Then they belong to display

> classic domain-driven design thingie - things that are named the same may be different things. Especially hard with large OSS projects

I have created a PR splitting the code (I want to add something to the animation class as well, and this will make it easier).
You might want to pull my branch and rebase yours on top of that, so that you don't need to redo the work twice: #4950

jesserockz (approved):

esphome/components/display/display_buffer.h
```diff
@@ -529,24 +529,33 @@ class Font {
...
-  Color get_rgb565_pixel(int x, int y) const;
-  Color get_grayscale_pixel(int x, int y) const;
-  int get_width() const;
-  int get_height() const;
+  Color get_pixel(int x, int y, Color color_on = COLOR_ON, Color color_off = COLOR_OFF) const;
+  int get_width() const override;
+  int get_height() const override;
   ImageType get_type() const;
 
-  virtual int get_current_frame() const;
```
nielsnl68, line 544: I added this method here to allow being used in my `SwitchPlate` component.
Could you please move it to the `BaseImage` class, it should normally return 0 (zero)? 
And only change value with animated images.
Thanks

nielsnl68 (commented):
Thanks for your work. I have just one important request :

esphome/components/display/display_buffer.h
ayufan: @nielsnl68 Can you link me to PR? Just out of curiosity?

nielsnl68:
I did not made a PR for it. The project is still to Alfa to create a PR.
but here is my repo: https://github.com/nielsnl68/esphome-components/tree/main/components/switchplate

PR #4933, opened 2023-06-10
display: allow to align image with `ImageAlign`

ayufan:

This is based on https://github.com/esphome/esphome/pull/4932

This allows to render image that is aligned, similar to `TextAlign`.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/2996

## Test Environment

- [ ] ESP32
- [x] ESP32S3
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -311,6 +311,37 @@ void DisplayBuffer::vprintf_(int x, int y, Font *font, Color color, TextAlign al
 }
 
 void DisplayBuffer::image(int x, int y, BaseImage *image, Color color_on, Color color_off) {
+  this->image(x, y, image, ImageAlign::TOP_LEFT, color_on, color_off);
+}
+
+void DisplayBuffer::image(int x, int y, BaseImage *image, ImageAlign align, Color color_on, Color color_off) {
+  auto x_align = ImageAlign(int(align) & (int(ImageAlign::HORIZONTAL_ALIGNMENT)));
+  auto y_align = ImageAlign(int(align) & (int(ImageAlign::VERTICAL_ALIGNMENT)));
+
+  switch (x_align) {
+    case ImageAlign::RIGHT:
+      x -= image->get_width();
+      break;
+    case ImageAlign::CENTER_HORIZONTAL:
+      x -= image->get_width() / 2;
+      break;
+    case ImageAlign::LEFT:
+    default:
+      break;
+  }
+
+  switch (y_align) {
+    case ImageAlign::BOTTOM:
+      y -= image->get_height();
+      break;
+    case ImageAlign::CENTER_VERTICAL:
+      y -= image->get_height() / 2;
+      break;
+    case ImageAlign::TOP:
+    default:
+      break;
+  }
+
   image->draw(x, y, this, color_on, color_off);
 }
 
--- a/esphome/components/display/display_buffer.h
+++ b/esphome/components/display/display_buffer.h
@@ -74,6 +74,54 @@ enum class TextAlign {
   BOTTOM_RIGHT = BOTTOM | RIGHT,
 };
 
+/** ImageAlign is used to tell the display class how to position a image. By default
+ * the coordinates you enter for the image() functions take the upper left corner of the image
+ * as the "anchor" point. You can customize this behavior to, for example, make the coordinates
+ * refer to the *center* of the image.
+ *
+ * All image alignments consist of an X and Y-coordinate alignment. For the alignment along the X-axis
+ * these options are allowed:
+ *
+ * - LEFT (x-coordinate of anchor point is on left)
+ * - CENTER_HORIZONTAL (x-coordinate of anchor point is in the horizontal center of the image)
+ * - RIGHT (x-coordinate of anchor point is on right)
+ *
+ * For the Y-Axis alignment these options are allowed:
+ *
+ * - TOP (y-coordinate of anchor is on the top of the image)
+ * - CENTER_VERTICAL (y-coordinate of anchor is in the vertical center of the image)
+ * - BOTTOM (y-coordinate of anchor is on the bottom of the image)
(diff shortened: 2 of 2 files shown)


ayufan:
cc @guillempages :)

This adds the https://github.com/esphome/esphome/pull/4933/commits/ab63404917557d96671e022f800c61aeb5782b26 on top of other incremental changes.

esphome/components/display/display_buffer.cpp
```diff
@@ -306,63 +325,41 @@ void DisplayBuffer::vprintf_(int x, int y, Font *font, Color color, TextAlign al
...
-          }
-        }
-      }
+    case ImageAlign::CENTER_VERTICAL:
+      y -= image->get_height() / 2;
+      break;
+    case ImageAlign::TOP:
+    default:
       break;
   }
```
guillempages: Please, look at how the text_align is doing that, and try to keep it consistent.

It is probably better readable/easier to follow, if you create a method `get_image_bounds` like `get_text_bounds`.

esphome/components/display/display_buffer.cpp
```diff
@@ -306,63 +325,41 @@ void DisplayBuffer::vprintf_(int x, int y, Font *font, Color color, TextAlign al
...
-          if (image->get_pixel(img_x, img_y)) {
-            this->draw_pixel_at(x + img_x, y + img_y, color_on);
-          } else if (!transparent) {
-            this->draw_pixel_at(x + img_x, y + img_y, color_off);
-          }
-        }
-      }
+void DisplayBuffer::image(int x, int y, BaseImage *image, ImageAlign align, Color color_on, Color color_off) {
+  auto x_align =
+      ImageAlign(int(align) & (int(ImageAlign::LEFT) | int(ImageAlign::CENTER_HORIZONTAL) | int(ImageAlign::RIGHT)));
```
guillempages: You might want to define a constant "HORIZONTAL_ALIGNMENT" and use that instead of the 3-way-or. IMHO this would make it slightly more readable.

esphome/components/display/display_buffer.cpp
```diff
@@ -306,63 +325,41 @@ void DisplayBuffer::vprintf_(int x, int y, Font *font, Color color, TextAlign al
...
+  auto y_align =
+      ImageAlign(int(align) & (int(ImageAlign::TOP) | int(ImageAlign::CENTER_VERTICAL) | int(ImageAlign::BOTTOM)));
```
guillempages: You might want to define a constant "VERTICAL_ALIGNMENT" and use that instead of the 3-way-or. IMHO this would make it slightly more readable.

esphome/components/display/display_buffer.cpp
```diff
@@ -306,63 +325,41 @@ void DisplayBuffer::vprintf_(int x, int y, Font *font, Color color, TextAlign al
...
-      }
+    case ImageAlign::CENTER_VERTICAL:
+      y -= image->get_height() / 2;
+      break;
+    case ImageAlign::TOP:
+    default:
       break;
   }
+
+  image->draw(x, y, this, color_on, color_off);
```
guillempages, line 345: What if the resulting `x` or `y` are actually out-of-bounds? (e.g. negative, or greater than display's width/height)

The actual `draw` method would try to write pixels outside of the display; which could then also lead to crashes.
(well, I concede that becoming greater than width/height would also be a problem with the current code, since `TOP_LEFT` is the default anyway). Nevertheless it might be worth adding a check here (or on the "real" `draw` method).

ayufan: @guillempages The `x` and `y` being negative will work fine, the image will be cut to display. And nothing would crash, since there's a check for bounds when setting a pixel. This is no different to the case where `x+image_width` is larger than screen size, the image will be cropped and pixels will be discarded by `set_pixel_at`.

esphome/components/display/display_buffer.cpp
ayufan: This is good idea.

esphome/components/display/display_buffer.cpp
ayufan: I was thinking about that, but tbh this seems to not be needed outside of this method. So, in such case and given the complexity of that code it seemed reasonable to not introduce `get_text_bounds`.

I'm actually debating with myself whether to open PR that would move `get_text_bounds` to `Font::get_text_bounds_` and add `Font::print` method.

esphome/components/display/display_buffer.cpp
guillempages: That's not always true. I've encountered a display that actually did crash if you tried to write a pixel out-of-bounds.

ayufan: @guillempages Yes, you are right. It might happen since the `draw_pixel_at` does call `draw_absolute_pixel_internal` which might or might not check bounds: https://github.com/esphome/esphome/blob/dev/esphome/components/display/display_buffer.cpp#LL160C9-L160C37.

However, this does not change that this is being problem today. Easiest way to crash if the `draw_absolute_pixel_internal` is unsafe is to write: `display.image(1000, 1000, id(image))` :)

This is also the problem for the `void DisplayBuffer::print(int x, int y, Font *font, Color color, TextAlign align, const char *text)`.

The easiest way could be ensuring that `draw_pixel_at` is always safe to call. However, this changed without adding some bitblt would negatively impact performance.

jesserockz (dismissed):

jesserockz:
@ayufan Please add a docs PR to add the new overloaded function

nielsnl68 (approved):

ayufan:
@jesserockz https://github.com/esphome/esphome-docs/pull/2996 docs are in here

PR #4941, opened 2023-06-13
Add support for ESP32-S3-BOX-Lite displays

guillempages:
# What does this implement/fix?

The ESP32-S3-BOX-Lite display seems to have an ILI9xxx driver (regardless of the documentation mentioning st7789v).
Add the needed configuration so that it works.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

Related to esphome/feature-requests/issues/2239 (Not a full fix)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2988

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
display:
  - platform: ili9xxx
    model: S3BOX_LITE
    cs_pin: GPIO5
    dc_pin: GPIO4
    reset_pin: GPIO48
    id: lcd
    # Width = 320, Height = 240
    lambda: |-
      it.fill(Color(32, 32, 0));
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #4942, opened 2023-06-13
Add support for S3 box display

guillempages:

Add support for the ILI9xxx display for the ESP32-S3-Box.

This is currently based on #4941 to avoid merge conflicts later on (i.e. it is also showing those changes). Once that is merged, I'll rebase it to dev to make review easier.

## Types of changes

- [ ] Other

Related to https://github.com/esphome/feature-requests/issues/2239 (not a full fix)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2294
## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
display:
  - platform: ili9xxx
    model: S3BOX_LITE
    cs_pin: GPIO5
    dc_pin: GPIO4
    reset_pin: GPIO48
    id: lcd
    lambda: |-
      it.fill(Color(32, 32, 0));


```

## Checklist:


nielsnl68:
Hi @guillempages, do you know which display driver is used in this device?
Imho, It would be better to use the name of the display driver instead of the device name where the display is used in.

kroimon:
I think it would be better to just make the existing models more generic by offering additional YAML options for mirroring and color orders (the different options available via the MADCTL register) instead of adding more and more models which just use different values for this register.

nielsnl68:
> I think it would be better to just make the existing models more generic by offering additional YAML options for mirroring and color orders (the different options available via the MADCTL register) instead of adding more and more models which just use different values for this register.

i agree 100%

guillempages:
> do you know which display driver is used in this device?
It would be better to use the name of the display driver instead of the device name where the display is used in.

According to the documentation by the manufacturer, it is an ILI9342C; which is a problem, since we already have a configuration for that. But using that configuration, the display is mirrored; which prevents reusing that configuration.

After discussing this with @jesserockz and @kbx81 we agreed on the "S3-BOX" (and "S3-BOX-LITE" for #4941) names.

> I think it would be better to just make the existing models more generic by offering additional YAML options for mirroring and color orders (the different options available via the MADCTL register) instead of adding more and more models which just use different values for this register.

My initial idea was actually also to add the possibility to change color order and inversion as additional options.

But in the discussion mentioned above, the consensus was to add additional models to the existing driver, instead of the generic approach.

guillempages:
According to @rpatel3001 it's working as expected 🙂

kbx81 (approved):

PR #4944, opened 2023-06-14
Construct web_server assets at build time instead of run time

bdraco:

Construct web_server assets at build time instead of run time

The index page was constructed every time a request was made. We can pre-compute it during build and store it in flash

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/web_server/web_server.h
```diff
@@ -40,31 +57,37 @@ class WebServer : public Controller, public Component, public AsyncWebHandler {
...
-   */
-  void set_css_include(const char *css_include);
-
   /** Set the URL to the script that's embedded in the index page. Defaults to
    * https://esphome.io/_static/webserver-v1.min.js
    *
    * @param js_url The url to the web server script.
    */
-  void set_js_url(const char *js_url);
+  void set_js_url(const char *js_url);
```
bdraco:
```suggestion
  void set_js_url(const char *js_url);
```

esphome/components/web_server/web_server.cpp
```diff
@@ -308,49 +308,41 @@ void WebServer::handle_index_request(AsyncWebServerRequest *request) {
...
-  stream->print(F("</body></html>"));
-#endif
-
   request->send(stream);
 }
+#elif USE_WEBSERVER_VERSION == 2
+void WebServer::handle_index_request(AsyncWebServerRequest *request) {
+  AsyncWebServerResponse *response =
+      request->beginResponse_P(200, "text/html", ESPHOME_WEBSERVER_INDEX_HTML, ESPHOME_WEBSERVER_INDEX_HTML_SIZE);
+  // All content is controlled and created by user - so allowing all origins is fine here.
```
jesserockz: Is this comment required here?

bdraco:
```suggestion
```

bdraco: Not anymore. It was a duplicate header since it was already set in the `.h`

esphome/components/web_server/__init__.py
```diff
@@ -81,6 +81,37 @@ def validate_local(config):
...
 
 
+def build_index_html(config) -> str:
+    html = "<!DOCTYPE html><html><head><meta charset=UTF-8><link rel=icon href=data:>"
+    css_include = config.get(CONF_CSS_INCLUDE)
+    js_include = config.get(CONF_JS_INCLUDE)
+    if css_include:
+        html += "<link rel=stylesheet href=/0.css>"
+    if config[CONF_CSS_URL]:
+        html += '<link rel=stylesheet href="{config[CONF_CSS_URL]}">'
```
jesserockz:
```suggestion
        html += f'<link rel=stylesheet href="{config[CONF_CSS_URL]}">'
```

jesserockz (approved):

bdraco:
Thanks.  Just need to find a bit more ram somewhere else to keep my esp8266 garage door opener a bit more stable 😉

PR #4956, opened 2023-06-17
display/font: optimise font rendering by about 25%

ayufan:

This does contains changes from https://github.com/esphome/esphome/pull/4956. They can be merged together or separately. The commit added on top of the other PR: https://github.com/esphome/esphome/pull/4956/commits/ebc33bc0b71efa3a791a9fc148bce368984a8fd2.

- optimise font rendering by about 25%
- add `BaseFont` and introduce `Font::draw` methods
- since the `Font::draw` is introduced the pixel loop is optimised to avoid repetitive operations
- ESP32S3: before: `1840ms`, after: `1411ms`
- the difference will be bigger on ESP8266 due to memory access pattern
- move `BaseFont`, `BaseImage` and `COLOR_*` to `displaybuffer.h` to make image/font only depend on `displaybuffer`, but not other way around
   
## Types of changes

- [x] Other: Performance improvement

## Test Environment

- [ ] ESP32
- [x] ESP32S3 (Fly Halo)
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
display:
      auto_clear_enabled: true
      lambda: |-
        auto start_us = micros();
        it.print(120, 20, id(roboto), TextAlign::TOP_CENTER, "Hello World!");
        auto dt_us = micros() - start_us;
        ESP_LOGE("logo_page", "text draw: %ld", dt_us);
```

## Checklist:

### DIFF
--- a/esphome/components/display/font.cpp
+++ b/esphome/components/display/font.cpp
@@ -8,25 +8,23 @@ namespace display {
 
 static const char *const TAG = "display";
 
-bool Glyph::get_pixel(int x, int y) const {
-  const int x_data = x - this->glyph_data_->offset_x;
-  const int y_data = y - this->glyph_data_->offset_y;
-  if (x_data < 0 || x_data >= this->glyph_data_->width || y_data < 0 || y_data >= this->glyph_data_->height)
-    return false;
-  const uint32_t width_8 = ((this->glyph_data_->width + 7u) / 8u) * 8u;
-  const uint32_t pos = x_data + y_data * width_8;
-  return progmem_read_byte(this->glyph_data_->data + (pos / 8u)) & (0x80 >> (pos % 8u));
-}
 void Glyph::draw(int x_at, int y_start, DisplayBuffer *display, Color color) const {
   int scan_x1, scan_y1, scan_width, scan_height;
   this->scan_area(&scan_x1, &scan_y1, &scan_width, &scan_height);
 
-  const int glyph_x_max = scan_x1 + scan_width;
-  const int glyph_y_max = scan_y1 + scan_height;
-  for (int glyph_x = scan_x1; glyph_x < glyph_x_max; glyph_x++) {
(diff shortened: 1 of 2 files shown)


guillempages:
@ayufan I have no objections to this commit; just a small nitpick. I think the removal of the "font.h", "animation.h" and "image.h" from the display_buffer.h file should be considered a breaking change, since it will force external components that are extending those classes to change their code (i.e. add the extra include) to allow it to compile.
This is of course not a big deal, but should be mentioned explicitly IMHO.

ayufan:
@guillempages Good call. I wonder how big of the deal it is to have this breaking change? Or should we avoid this?

I'm have another PR ready that moves to make those behave the same as `graph`, or `qr_code` does:

- `image.cpp/.h` into `components/image`
- `animation.cpp/.h` into `components/animation`
- `font.cpp/.h` into `components/font`

We can do this since they are not longer interlocked with `display_buffer.cpp` due to introduction of `BaseFont` and `BaseImage`, and they are rather a very specific custom implementation that require initialising those components anyway. The purpose is to make `components/display` clean to contain only display related buffering.

@jesserockz What is your take on the above move? This is PoC branch: https://github.com/ayufan-research/esphome/tree/display-move-font-image-anim

guillempages:
Regarding the breaking change; I don't see it as a big deal; IMHO just mentioning it should be enough. But that's @jesserockz's call.

Regarding the moving of the files, I personally would find it clearer, if the actual h/cpp files are together in the same folder as their python __init__ files (makes it easier to modify things together, and probably the `component: xxx` labels on github would then also be set more meaningfully (i.e. there is no need to mention `component: display` if it's not the base display, but e.g. the animation which is changing).

jesserockz (approved):

jesserockz:
I have added the breaking-change label to https://github.com/esphome/esphome/pull/4963 which moved and removed the headers.

PR #4959, opened 2023-06-17
Add actions to animation

guillempages:

Add actions to be able to easily change the current frame to the animation component.
Instead of needing to use a lambda and c++ code, allow a simple yaml action to switch to the next or previous frame, or to set a specific frame.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3001

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
animation:
  - id: fire
    file: fire.gif

interval:
  - interval: 200ms
    then:
       - animation.next_frame: fire
# Alternative 1:
#       - animation.prev_frame: fire
# Alternative 2:
#       - animation.set_frame:
#          - id: fire
#            frame: 5
       - component.update: disp

display:
  - platform: <xyz>
     id: disp
     lambda: |-
       it.image(0, 0, id(fire));
```

## Checklist:

### DIFF
--- a/esphome/components/animation/__init__.py
+++ b/esphome/components/animation/__init__.py
@@ -1,6 +1,6 @@
 import logging
 
-from esphome import core
+from esphome import automation, core
 from esphome.components import display, font
 import esphome.components.image as espImage
 from esphome.components.image import CONF_USE_TRANSPARENCY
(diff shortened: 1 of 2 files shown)


esphome/components/animation/__init__.py
```diff
@@ -74,7 +87,35 @@ def validate_cross_dependencies(config):
...
+)
+PREV_FRAME_SCHEMA = automation.maybe_simple_id(
+    {
+        cv.GenerateID(): cv.use_id(Animation_),
+    }
+)
+SET_FRAME_SCHEMA = cv.Schema(
+    {
+        cv.GenerateID(): cv.use_id(Animation_),
+        cv.Required(CONF_FRAME): cv.positive_int,
```
jesserockz:
```suggestion
        cv.Required(CONF_FRAME): cv.uint16_t,
```

esphome/components/display/animation.cpp
```diff
@@ -1,6 +1,7 @@
 #include "animation.h"
 
 #include "esphome/core/hal.h"
+#include "esphome/core/log.h"
```
jesserockz: Why is this needed? There are no logs in this file?

jesserockz (changes requested):

esphome/components/display/animation.cpp
guillempages: Ouch. I removed the debug logs I was using while implementing, and forgot to remove the include.

Good spot, thanks!

jesserockz (approved):

PR #4985, opened 2023-06-21
Adding New SM300D2 V2 hardware component.

Duker4444:
Adding New SM300D2 V2 hardware component.

There was release of a new hardware version of the SM300D2 sensor that sends 19 bytes vs the old 17 byte stream.
This broke the original code since the data no longer aligned up properly causing it to error out with a "Preamble error".

I'm calling it V2 since it's not compatible with the V1 editions.

I've also changed the serial trigger to be on the 300ms timing between datasets vs a specific character.
This allows the system to pickup the data as a whole string every time. " No more Preamble errors.. "
It just waits for the delay and triggers as soon as it receives data. Resolving differing Device addresses.

I'm still working out the checksum. validation if needed..
however Both of my V2 sensors are reading fine now.

Related issue or feature (if applicable): fixes
This fixes the following SM300D2 Sensor - Invalid Preamble - Reading Data Time out #3767

Test Environment
 ESP32
 ESP8266
Notes:

You should see in the logs the device address and function type in my case shows type 17 meaning "7 in 1"
The Sensors by default update approx 1-2mins unless a shorter update interval is specified.
A Reset Button has been coded in. you should see it show up as an option on the device.
New Sensors have a 48hr hardware "burn-in" before they are really accurate.
Avoid using the WebServer option since this seems to cause the ESP to hang over time.
Changes:
Improved logging and removed the preamble issue all together.

Sample Config for ESPHOME

uart:
rx_pin:
number: GPIO3
mode: INPUT_PULLDOWN
baud_rate: 9600
rx_buffer_size: 1024

sensor:

platform: sm300d2v2
co2:
name: "Office CO2 Value"

formaldehyde:
name: "Office Formaldehyde Value"

tvoc:
name: "Office TVOC Value"

pm_2_5:
name: "Office PM2.5 Value"

pm_10_0:
name: "Office PM10 Value"

temperature:
name: "Office Temperature Value"

humidity:
name: "Office Humidity Value"


jesserockz (commented):
Most of this code looks the same apart from the additional sensors and the offset in the response. 
Could this be combined with the existing platform and have a model config variable added to switch these things instead of a whole duplicate platform?

Duker4444:
So im fairly new at programming in c++, but eventually, I'd like to take an entirely different approach in parsing out the dataset that would work for both types..

I would look at creating two functions with a detection piece early on to figure out which type im working with.. and then use the right function to pull the data and parse it out.

My skills are not quite there yet.

I only have v2 hardware.. it would be nice to get a v1 where i could develop against it as well.

The two are very similar  since I based the code off the v1. But since the length of the dataset is so different.  The v1 simply wouldn't work. 

I did notice that the owner seems to be missing on the original code. I could at least incorporate some of the code improvements into it that I've placed in my v2 edition on the trigger.

But I would be limited on testing without a v1.  This v2 stinks since there's almost no documentation out there on it. It's like the vendor released it and didn't say anything at all to the public.  I only discovered it on another similar product they made.

My eventual goal would be to make one component.. 

It would be great to find a co-developer to work with..

Duker4444:
Suggestions are greatly welcomed.. :)

PR #4991, opened 2023-06-22
display: fix white screen on binary displays

ayufan:

This a bug introduced by https://github.com/esphome/esphome/pull/4600.

It changed `COLOR_OFF` for reasons not related to the merged PR.
The potential problem with `COLOR_OFF` is steaming from sometimes
a wrong assumption that `Color` represents `RGBA`, where-as it does
in fact represents `RGBW`.

In case of some displays the `W` is considered as a white-level,
and binary displays expect the value of a pixel to be `0` for pixel to be off.
The binary pixel is checked via `Color#is_on()` which does look at `this->raw_32 != 0`.

This does not change how of transparent images are displayed on color/grayscale displays.
Since for transparent pixels are not render at all.

It does not fix rendering of the `PNG` images on binary displays (which is broken currently),
as incorrect luminance is used.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue): https://github.com/esphome/issues/issues/4587
- [ ] Other

## Test Environment

- [ ] ESP32: TTGO with LoRa, and `SSD1306` (binary) 
- [x] ESP32S3: Fly HALO with `gc9a01` (RGB565)
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
esp32:

esphome:
  name: lora8

i2c:
  sda: SDA
  scl: SCL

logger:

font:
  - file: "gfonts://Roboto"
    id: roboto
    size: 20

image:
  - id: standby_image
    file: "rgb.png"
    resize: 128x64
    type: BINARY
    use_transparency: true

display:
  - platform: ssd1306_i2c
    model: "SSD1306 128x64"
    address: 0x3C
    lambda: |-
      it.print(0, 0, id(roboto), "Hello World");
      it.image(0, 0, id(standby_image));
```

## Checklist:

### DIFF
--- a/esphome/components/display/display_buffer.cpp
+++ b/esphome/components/display/display_buffer.cpp
@@ -12,7 +12,7 @@ namespace display {
 
 static const char *const TAG = "display";
 
-const Color COLOR_OFF(0, 0, 0, 255);
+const Color COLOR_OFF(0, 0, 0, 0);
 const Color COLOR_ON(255, 255, 255, 255);
 
 void DisplayBuffer::init_internal_(uint32_t buffer_length) {


randybb:
For now tested with SSD1306 and it is working fine.SH1107 is still broken - like wrong init.

ayufan:
> SH1107 is still broken - like wrong init

I guess, it is another issue :)

nagyrobi:
SH1107 (which is still driven by ssd1306 component; eg. [FeatherWing OLED - 128x64](https://www.adafruit.com/product/4650)) works with this patch:

```yaml
  - source: github://bekriebel/esphome@sh1107_fix
    components:
      - ssd1306_base
      - ssd1306_i2c
      - ssd1306_spi
```

Anybody who has review capabilites would care to make a PR out of it?

randybb:
> > SH1107 is still broken - like wrong init
> 
> I guess, it is another issue :)

yea, I thought it was merged as I tested it a long time ago :D - so we need to wait for that PR.

guillempages:
> 
> Anybody who has review capabilites would care to make a PR out of it?

Actually you don't need review capabilities to create a PR; just to get it merged ;-)
so if @nikitakuklev or @bekriebel 
has already a patch, it would be easiest if he/she creates the PR

guillempages (approved):
Looks good.
Seems to fix the issue (according to the feedback), and it does not break the transparency implementation.

kbx81 (approved):

PR #4996, opened 2023-06-23
Draft: Testing PR for `display_buffer` approach for display drivers

ayufan:

The purpose of this to provide a new way of writing display drivers:

### Design goals

- All existing display should continue working, but their implementation should be marked as deprecated
- Display driver should not manage framebuffer, only send data directly to the target device
- Display driver should use native device rotation where-is possible
- Display driver should declare what is native buffer format for buffer transfer
- Framebuffer should provide an allocated buffer of choice
- Image/Font and all other functions should prefer to use `draw_pixels` or `draw_pixels_format` in all cases which would convert if needed from non-native format to native and render by lines
- Pixel formats should provide a uniform and convenient pixel mapping between all pixel formats
- Transparancy should be implemented using Pixel Format, where the color key should part of Pixel Format implementation, similar to how `PixelA1` and `PixelW1` are implemented

### Performance

- previously took 170ms to render 240x240/RGB565
- framebuffer: with direct blit to framebuffer, it takes 8ms
- framebuffer: with conversion from RGB888 to RGB565, it takes 21ms
- direct to display: with conversion from RGB888 over SPI to RGB565, it takes 56ms
- direct to display: RGB565 to SPI RGB565 it takes 30ms
- displaying framebuffer, it takes around 30ms to be transfered over SPI

- previously took 1700us to render the front from example
- framebuffer: it takes around 400ms now

> The routines for pixel conversion are not yet optimised.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```yaml
display:
  - id: lcd_display
    platform: display_gc9a01
    reset_pin: GPIO13
    cs_pin: GPIO10
    dc_pin: GPIO9
    rotation: 0
    auto_clear_enabled: false

  - id: framebuffer
    platform: display_buffer
    width: 240
    height: 240
    format: RGB565
    auto_clear_enabled: true
    lambda: |-
      start_us = micros();
      it.image(120, 120, id(logo), ImageAlign::CENTER);
      ESP_LOGE("logo_page", "mellow_logo2: %ld", micros() - start_us);

      start_us = micros();
      it.print(120, 120, id(roboto), TextAlign::TOP_CENTER, "Hello World!");
      ESP_LOGE("logo_page", "roboto: %ld", micros() - start_us);

      start_us = micros();
      id(framebuffer).draw(id(lcd_display));
      ESP_LOGE("logo_page", "send: %ld", micros() - start_us);

      ESP_LOGE("logo_page", "refresh: %ld", micros() - refresh_us);
      refresh_us = micros();
```

## Checklist:

### DIFF
--- a/esphome/components/display/__init__.py
+++ b/esphome/components/display/__init__.py
@@ -46,6 +46,16 @@ DISPLAY_ROTATIONS = {
     270: display_ns.DISPLAY_ROTATION_270_DEGREES,
 }
 
+PIXEL_TYPES = {
+    "BINARY": display_ns.enum("PixelW1"),
+    "W2": display_ns.enum("PixelW2"),
+    "W4": display_ns.enum("PixelW4"),
+    "GRAYSCALE": display_ns.enum("PixelW8"),
+    "RGB332": display_ns.enum("PixelRGB332"),
+    "RGB565": display_ns.enum("PixelRGB565_BE"),
+    "RGB24": display_ns.enum("PixelRGB888"),
+    "RGBA": display_ns.enum("PixelRGBA8888")
+}
 
 def validate_rotation(value):
     value = cv.string(value)
--- a/esphome/components/display/display.h
+++ b/esphome/components/display/display.h
@@ -4,6 +4,7 @@
 #include <vector>
 
 #include "rect.h"
+#include "pixel_formats.h"
 
 #include "esphome/core/color.h"
 #include "esphome/core/automation.h"
@@ -483,7 +484,15 @@ class Display {
    */
   bool clip(int x, int y);
 
+  virtual PixelFormat get_native_pixel_format() = 0;
+  bool draw_pixels_at(int x, int y, int w, int h, const uint8_t *data, int stride, PixelFormat data_format, Color color_on = COLOR_ON, Color color_off = COLOR_OFF);
+
  protected:
+  virtual uint8_t *get_native_pixels_(int y) { return nullptr; }
+  virtual bool draw_pixels_(int x, int y, int w, int h,
+    const uint8_t *data, int data_line_size, int data_stride) { return false; }
+  virtual bool filled_rectangle_(int x1, int y1, int width, int height, Color color);
+
   bool clamp_x_(int x, int w, int &min_x, int &max_x);
   bool clamp_y_(int y, int h, int &min_y, int &max_y);
   void vprintf_(int x, int y, BaseFont *font, Color color, TextAlign align, const char *format, va_list arg);
--- a/esphome/components/display/display.h
+++ b/esphome/components/display/display.h
@@ -409,6 +409,9 @@ class Display {
 
   /// Internal method to set the display writer lambda.
   void set_writer(display_writer_t &&writer);
+  void set_writer(display_buffer_writer_t writer) {
+    this->set_writer([writer](Display &display) { return writer((display::DisplayBuffer &) display); });
+  }
 
   void show_page(DisplayPage *page);
   void show_next_page();
@@ -500,6 +503,8 @@ class Display {
 class DisplayPage {
  public:
   DisplayPage(display_writer_t writer);
+  DisplayPage(display_buffer_writer_t writer)
+      : DisplayPage([writer](Display &display) { return writer((display::DisplayBuffer &) display); }) {}
   void show();
   void show_next();
   void show_prev();
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -38,6 +38,7 @@ IMAGE_TYPE = {
     "RGB565": ImageType.IMAGE_TYPE_RGB565,
     "RGB24": ImageType.IMAGE_TYPE_RGB24,
     "RGBA": ImageType.IMAGE_TYPE_RGBA,
+    "JPEG": ImageType.IMAGE_TYPE_JPEG,
 }
 
 CONF_USE_TRANSPARENCY = "use_transparency"
@@ -347,6 +348,14 @@ async def to_code(config):
                     continue
                 pos = x + y * width8
                 data[pos // 8] |= 0x80 >> (pos % 8)
+    elif config[CONF_TYPE] in ["JPEG"]:
+        output = io.BytesIO()
+        image.save(output, format='JPEG')
+        data = list(output.getvalue())
+
+        cg.add_define("USE_JPEGDEC")
+        cg.add_library("JPEGDEC", "1.2.7")
+
     else:
         raise core.EsphomeError(
             f"Image f{config[CONF_ID]} has an unsupported type: {config[CONF_TYPE]}."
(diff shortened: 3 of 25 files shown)


esphome/components/display_gc9a01/__init__.py
guillempages, line 1: IMHO, adding an extra display should be made in a separate PR; this one is big enough as it is ;-)

esphome/components/display/__init__.py
```diff
@@ -45,6 +45,14 @@
     270: display_ns.DISPLAY_ROTATION_270_DEGREES,
 }
 
+PIXEL_TYPES = {
+    "BINARY": display_ns.enum("PixelW1"),
+    "GRAYSCALE": display_ns.enum("PixelW8"),
+    "RGB332": display_ns.enum("PixelRGB332"),
+    "RGB565": display_ns.enum("PixelRGB565_BE"),
```
guillempages, line 55: Why "BE" and not the standard RGB565?

esphome/components/image/__init__.py
```diff
@@ -28,14 +28,17 @@
...
+
+ImageType = image_ns.enum("ImageType")
 IMAGE_TYPE = {
     "BINARY": ImageType.IMAGE_TYPE_BINARY,
     "TRANSPARENT_BINARY": ImageType.IMAGE_TYPE_BINARY,
     "GRAYSCALE": ImageType.IMAGE_TYPE_GRAYSCALE,
     "RGB565": ImageType.IMAGE_TYPE_RGB565,
     "RGB24": ImageType.IMAGE_TYPE_RGB24,
     "RGBA": ImageType.IMAGE_TYPE_RGBA,
+    "JPEG": ImageType.IMAGE_TYPE_JPEG,
```
guillempages, line 41: I'm not 100% sure about this, but we might want to separate the compression methods into a separate enum/configuration.
(Thinking about my online-image PR here. If it gets separated, it could also implement an "AUTO" type, that would detect the content-type from html).
Also it could separate the input format (JPEG, PNG, GIF,...) from the desired framebuffer format.

For pre-loaded images, both "IMAGE_TYPE" and "IMAGE_FORMAT" could then be made exclusive, so you would still be able to store compressed images in the flash.
And having separate options could also make it easier to instantiate different base classes (e.g. BufferImage vs CompressedImage) later on, based on the specified option.

esphome/components/image/image.cpp
```diff
@@ -124,11 +172,92 @@ Color Image::get_grayscale_pixel_(int x, int y) const {
...
+
+static int jpeg_draw_by_native(JPEGDRAW *pDraw)
+{
+  display::Display *display = (display::Display*)pDraw->pUser;
+  display->draw_pixels_at(
+    pDraw->x, pDraw->y,
+    pDraw->iWidth, pDraw->iHeight,
+    (const uint8_t*)pDraw->pPixels,
+    pDraw->iBpp * pDraw->iWidth * pDraw->iHeight / 8,
+    display->get_native_pixel_format()
```
guillempages: why do you need to ask "display" for its native_pixel_format, to pass it to its own "draw_pixels_at" method? Couldn't the display method take care of getting its own known parameters?

esphome/components/image/__init__.py
ayufan: > I'm not 100% sure about this, but we might want to separate the compression methods into a separate enum/configuration.

I'm actually not sure. It feels just fine for `ImageType`, where `JPEG` is another type of image supported. There's not really an additional `FORMAT` needed, since this is already covered by `PixelFormat`. Image Type being sometimes RAW (RGB565) or Compressed (JPEG) is likely a good place to have this.

As for the image detection (of compressed formats), there could be a method to detect image type based on a few bytes of header, but I don't think that there should be `AUTO` for such.

esphome/components/display/__init__.py
ayufan: The current RGB565 implemented in a code is in fact `BE`, where-as most of libraries to default refer `LE`. I did not make my mind yet, but I think I will make `PixelRGB565_LE` to make it explicit. Other pixel formats do have stable endiness.

Correctness: Sometimes it is BE, sometimes it is LE.

esphome/components/image/image.cpp
ayufan: The `draw_pixels_at` is meant to be able to render data buffer of any format. In this case we have a buffer in a native format, thus why the native format is fetched.

The `draw_pixels_at` depending on a format, and a type of display might: do direct draw, do direct bitblt to buffer, or convert to draw, convert to buffer.

guillempages: But then, you could have an overloaded version of the method, that implicitly uses the display's own native pixel format. It really doesn't feel good, to have to ask a value from a class, only to pass it to a method of that class ;-)

ayufan: This is Draft, not a complete work. So, you are being slightly nitpicky :) Look at wider change in this work. I have this call right now due for the convenience.

esphome/components/display_gc9a01/__init__.py
ayufan: Per. Discord:

> all those commits are to be merged in separate PR: https://github.com/esphome/esphome/pull/4996/commits, each represent a single functional non-breaking and compilable change

esphome/components/display/__init__.py
```diff
@@ -45,6 +45,16 @@
     270: display_ns.DISPLAY_ROTATION_270_DEGREES,
 }
 
+PIXEL_TYPES = {
+    "BINARY": display_ns.enum("PixelW1"),
+    "W2": display_ns.enum("PixelW2"),
+    "W4": display_ns.enum("PixelW4"),
```
nielsnl68, line 52: Where stands W2 and W4 for? maybe replace them for more understandable terms?

ayufan: The idea is White 2 bits, or White 4 bits. Which in other is Gray 2 / 4 / 8.

xdecock:
This mr would help a lot with ePapers, i've implemented the 2bit grayscale on an epaper and handling both the color and the protocol in waveshare is a nightmare. Getting better separation would probably help.

I've seen a [Dual buffer pull request](https://github.com/esphome/esphome/pull/4833)  also for the il9xxx which might get easier with you work.

nielsnl68:
hi @ayufan could you please commit this changes also to my temporary display repo
at https://github.com/nielsnl68/esphome-display we can use that as testbed to try all different setups and changes etc.

esphome/components/display/__init__.py
clydebarrow: So maybe `GRAYSCALE_4BPP` and the like?

esphome/components/display/display.h
```diff
@@ -480,7 +484,15 @@ class Display {
    */
   bool clip(int x, int y);
 
+  virtual PixelFormat get_native_pixel_format() = 0;
+  bool draw_pixels_at(int x, int y, int w, int h, const uint8_t *data, int stride, PixelFormat data_format, Color color_on = COLOR_ON, Color color_off = COLOR_OFF);
```
clydebarrow, line 488: Can you document what the parameters mean exactly? I would assume `stride` is the line width in the source data, but that would be nice to know. And why are there two color values as parameters?

As a general principle I like to see public methods, especially those with long argument lists, have a documentation comment block. We don't currently have good developer docs for ESPHome but having the public API documented will really help.

PR #5001, opened 2023-06-25
display: split `DisplayBuffer` and `Display`

ayufan:

This is a change taken from the https://github.com/esphome/esphome/pull/4996.

It is non-breaking change to split large `DisplayBuffer` into two classes:
- `Display`: to provide all generic display functions
- `DisplayBuffer`: to provide backward compatible legacy way for managing display buffers and software screen rotation

At some point the `DisplayBuffer` will be renamed to `RotatedDisplay` since its primary function is to implement
a software rotated displays. The usage of `Display` expects that all display will implement screen rotation in hardware.

This does not change any existing interfaces.

## Types of changes

- [x] Other: refactor of Display class

## Test Environment

- [x] ESP32S3
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:


kbx81 (approved):
I'm fine with this, but as it's a bit of an architectural change, I think another set of eyes would be great before we merge.

esphome/components/display/display_buffer.cpp
```diff
@@ -73,333 +67,6 @@ void HOT DisplayBuffer::draw_pixel_at(int x, int y, Color color) {
   this->draw_absolute_pixel_internal(x, y, color);
   App.feed_wdt();
```
nielsnl68, line 68: Just a reminder for later and possible out of scope for this PR, not sure of we should feed the doc here anymore.

esphome/components/display/display_buffer.h
```diff
@@ -2,568 +2,35 @@
...
-  virtual void fill(Color color);
-  /// Clear the entire screen by filling it with OFF pixels.
-  void clear();
-
   /// Get the width of the image in pixels with rotation applied.
-  int get_width();
+  int get_width() override;
   /// Get the height of the image in pixels with rotation applied.
-  int get_height();
+  int get_height() override;
```
nielsnl68, line 20: Should this not also moved to the `Display` class, or do you have a reason to let those here? 

UPDATE: Don't my suggestion here, it is already made virtual in the `Display();` class.

esphome/components/display/display_buffer.h
```diff
@@ -2,568 +2,35 @@
...
-
-  /// Internal method to set the display rotation with.
-  void set_rotation(DisplayRotation rotation);
-
-  // Internal method to set display auto clearing.
-  void set_auto_clear(bool auto_clear_enabled) { this->auto_clear_enabled_ = auto_clear_enabled; }
+  void draw_pixel_at(int x, int y, Color color) override;
 
   virtual int get_height_internal() = 0;
   virtual int get_width_internal() = 0;
```
nielsnl68, line 26: same as above?

esphome/components/display/display_buffer.h
```diff
@@ -2,568 +2,35 @@
...
-   * return rect for active clipping region
-   */
-  Rect get_clipping();
-
-  bool is_clipping() const { return !this->clipping_rectangle_.empty(); }
 
  protected:
-  void vprintf_(int x, int y, BaseFont *font, Color color, TextAlign align, const char *format, va_list arg);
-
   virtual void draw_absolute_pixel_internal(int x, int y, Color color) = 0;
```
nielsnl68, line 29: same as above.

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -7,6 +7,17 @@ namespace touchscreen {
...
 
+void Touchscreen::set_display(display::Display *display) {
+  this->display_ = display;
+  this->display_width_ = display->get_width();
+  this->display_height_ = display->get_height();
+  this->rotation_ = static_cast<TouchRotation>(display->get_rotation());
+
+  if (this->rotation_ == ROTATE_90_DEGREES || this->rotation_ == ROTATE_270_DEGREES) {
+    std::swap(this->display_width_, this->display_height_);
+  }
```
nielsnl68, line 18: Is it an idea to move this part to the `setup();` method, then we know the display is fully setup as well.

esphome/components/display/display.h
```diff
@@ -0,0 +1,575 @@
...
+  DisplayRotation rotation_{DISPLAY_ROTATION_0_DEGREES};
+  optional<display_writer_t> writer_{};
+  DisplayPage *page_{nullptr};
+  DisplayPage *previous_page_{nullptr};
+  std::vector<DisplayOnPageChangeTrigger *> on_page_change_triggers_;
+  bool auto_clear_enabled_{true};
+  std::vector<Rect> clipping_rectangle_;
+};
+
+class DisplayPage {
```
nielsnl68, line 499: When we break everything separate files would `DisplayPage` class and events etc. not fit in its own file?

nielsnl68 (commented):
Here are some suggestions that i think would make it better,

esphome/components/display/display_buffer.h
ayufan: I don't understand the question. The `get_width/height` implementation is specific to the rotation of display. This is not universally good, so at least I made a decision that this is a responsibility of who is implementing `Display` to figure out what to do with `rotation` and provide `width/height`.

esphome/components/display/display_buffer.h
ayufan: Same as above. This is implementation detail related to how rotation is handled.

esphome/components/touchscreen/touchscreen.cpp
ayufan: I'm not changing behavior of existing code outside of `display`, as this is not purpose of this PR.

esphome/components/display/display_buffer.h
ayufan: The software rotation support is specific implementation done by `DisplayBuffer`. It is not universally needed.

esphome/components/touchscreen/touchscreen.cpp
nielsnl68: I understand, this change should be in an new PR. just wanted your opinion ;)

esphome/components/display/display_buffer.h
nielsnl68: It is not only used for rotation alone. It is also used for the max size with the touchscreen component and can be used in LAMBDA to figure out the size of the screen.

ayufan: Yes, but `get_width/height` is exposed to represent a renderable part of the display. And behavior of `get_width/height` is not changed. Components should not use `_internal` since this is implementation specific.

esphome/components/display/display.h
ayufan: We could maybe it now or in separate PR, but I would prefer if someone could do it in separate PR :)

nielsnl68: Although i think it is easier to do it in this PR. I can live with doing it in a next PR.

esphome/components/display/display_buffer.h
nielsnl68: Oeps, i have been sleeping when i wrote this comment above :( please disregard this statement's :(

nielsnl68 (approved):

PR #5003, opened 2023-06-25
display: Add helper methods to `Display::clip` and `Display::clamp_x/y_`

ayufan:

This is a change taken from the https://github.com/esphome/esphome/pull/4996. 
This requires changes from: https://github.com/esphome/esphome/pull/5001.

It is additive change adding `clip` and `clamp_x/y_` methods:

- `clip` checks if `x/y` is within display boundaries, and clipping region
- `clamp_x/y_` clamps `x/x+w` to display boundaries and clipping region, returning `min_x (inclusive)` and `max_x (exclusive)`

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32S3
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/display/display.cpp
+++ b/esphome/components/display/display.cpp
@@ -269,10 +269,7 @@ void Display::do_update_() {
   } else if (this->writer_.has_value()) {
     (*this->writer_)(*this);
   }
-  // remove all not ended clipping regions
-  while (is_clipping()) {
-    end_clipping();
-  }
+  this->clear_clipping_();
 }
 void DisplayOnPageChangeTrigger::process(DisplayPage *from, DisplayPage *to) {
   if ((this->from_ == nullptr || this->from_ == from) && (this->to_ == nullptr || this->to_ == to))
@@ -322,12 +319,50 @@ void Display::shrink_clipping(Rect add_rect) {
     this->clipping_rectangle_.back().shrink(add_rect);
   }
 }
-Rect Display::get_clipping() {
+Rect Display::get_clipping() const {
   if (this->clipping_rectangle_.empty()) {
     return Rect();
   } else {
     return this->clipping_rectangle_.back();
   }
+}
+void Display::clear_clipping_() { this->clipping_rectangle_.clear(); }
+bool Display::clip(int x, int y) {
+  if (x < 0 || x >= this->get_width() || y < 0 || y >= this->get_height())
+    return false;
+  if (!this->get_clipping().inside(x, y))
+    return false;
+  return true;
+}
+bool Display::clamp_x_(int x, int w, int &min_x, int &max_x) {
+  min_x = std::max(x, 0);
+  max_x = std::min(x + w, this->get_width());
+
+  if (!this->clipping_rectangle_.empty()) {
+    const auto &rect = this->clipping_rectangle_.back();
+    if (!rect.is_set())
+      return false;
+
+    min_x = std::max(min_x, (int) rect.x);
+    max_x = std::min(max_x, (int) rect.x2());
+  }
+
+  return min_x < max_x;
+}
+bool Display::clamp_y_(int y, int h, int &min_y, int &max_y) {
+  min_y = std::max(y, 0);
+  max_y = std::min(y + h, this->get_height());
+
+  if (!this->clipping_rectangle_.empty()) {
+    const auto &rect = this->clipping_rectangle_.back();
+    if (!rect.is_set())
+      return false;
+
+    min_y = std::max(min_y, (int) rect.y);
+    max_y = std::min(max_y, (int) rect.y2());
+  }
+
+  return min_y < max_y;
 }
 
 DisplayPage::DisplayPage(display_writer_t writer) : writer_(std::move(writer)) {}
--- a/esphome/components/display/display.h
+++ b/esphome/components/display/display.h
@@ -472,14 +472,21 @@ class Display {
    *
    * return rect for active clipping region
    */
-  Rect get_clipping();
+  Rect get_clipping() const;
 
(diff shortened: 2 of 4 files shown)


esphome/components/display/display.h
```diff
@@ -0,0 +1,582 @@
...
+  DisplayRotation get_rotation() const { return this->rotation_; }
+
+  /** Get the type of display that the buffer corresponds to. In case of dynamically configurable displays,
+   * returns the type the display is currently configured to.
+   */
+  virtual DisplayType get_display_type() = 0;
+
+  /** Set the clipping rectangle for further drawing
+   *
+   * @param[in]  rect:       Pointer to Rect for clipping (or NULL for entire screen)
```
guillempages: You mention "Pointer to Rect ... or NULL", but the parameter type is a plain Rect, not a pointer

esphome/components/display/rect.h
```diff
@@ -16,19 +16,19 @@ class Rect {
...
 
   void extend(Rect rect);
   void shrink(Rect rect);
 
-  bool inside(Rect rect, bool absolute = true);
-  bool inside(int16_t test_x, int16_t test_y, bool absolute = true);
-  bool equal(Rect rect);
+  bool inside(Rect rect, bool absolute = true) const;
+  bool inside(int16_t test_x, int16_t test_y, bool absolute = true) const;
+  bool equal(Rect rect) const;
```
guillempages, line 31: would it make sense to add an `operator==` method for that? currently not used, but might make the code nicer

guillempages: Also `extend` and `shrink` could be handled with `operator+` and `operator-`

esphome/components/display/display.cpp
```diff
@@ -0,0 +1,378 @@
...
+}
+void Display::clear_clipping_() { this->clipping_rectangle_.clear(); }
+bool Display::clip(int x, int y) {
+  if (x < 0 || x >= this->get_width() || y < 0 || y >= this->get_height())
+    return false;
+  if (!this->get_clipping().inside(x, y))
+    return false;
+  return true;
+}
+bool Display::clamp_x_(int x, int w, int &min_x, int &max_x) {
```
guillempages, line 337: should this method be `const`?

esphome/components/display/display.cpp
```diff
@@ -0,0 +1,378 @@
...
+    if (!rect.is_set())
+      return false;
+
+    min_x = std::max(min_x, (int) rect.x);
+    max_x = std::min(max_x, (int) rect.x2());
+  }
+
+  return min_x < max_x;
+}
+bool Display::clamp_y_(int y, int h, int &min_y, int &max_y) {
```
guillempages, line 352: should this be `const`?

esphome/components/display/display.h
```diff
@@ -0,0 +1,582 @@
...
+
+  bool is_clipping() const { return !this->clipping_rectangle_.empty(); }
+
+  /** Check if pixel is within region of display.
+   */
+  bool clip(int x, int y);
+
+ protected:
+  bool clamp_x_(int x, int w, int &min_x, int &max_x);
+  bool clamp_y_(int y, int h, int &min_y, int &max_y);
```
guillempages, line 485: what are those methods actually needed for? It would be easier to review/understand (at least to me) if I could see how they are going to be used ;-)

esphome/components/display/rect.h
nielsnl68: The above is part of my clipping update. @ayufan has only moved it to its own file.
Your idea is something i will love to add @guillempages after this release is committed i will add those.

esphome/components/display/display.h
nielsnl68: You should have reviewed my PR better :P
But you are right that is for sure a leftover from the code that i used to implementing this methods.

Suggestion:
>  * @param[in]  rect:        Use a Rect object for clipping

guillempages: I don't remember reviewing your PR at all ;-P

esphome/components/display/display.h
ayufan: @guillempages Not yet, they are in: https://github.com/esphome/esphome/pull/4996.

esphome/components/display/display.cpp
ayufan: Yes, but I cannot change it yet due to `get_width()` not being `const`. This is something to be fixed in interface of this class, which forces to also update everywhere the `get_width_internal()`. The same goes for `get_height`.

guillempages: If all the displays (including external) need to be updated anyway, it would be good IMHO to do that in a single release, so that external displays do not need to change the code twice...
OTOH; forcing get_width_internal to be const might have deeper implications: Some display implementation might theoretically do a lazy implementation (i.e. get the actual width from the controller when it is read for the first time, and then store it as a variable). This would then not be possible in a const method.
Not sure how big an issue this would be though.

esphome/components/display/display.h
guillempages: IMHO, it would be good to add some documentation to the methods, even if they are only "internal protected" methods ;-)

kbx81 (approved):

kbx81:
@ayufan @guillempages @nielsnl68 I'm alright with merging this -- please mark the conversations as "resolved" so I know if/when it's ok to do so. 😇

esphome/components/display/display.h
ayufan: I can do it in about 4wks time :) or can do it in a follow-up PR.

PR #2956, opened 2021-12-28
Handle Nextion displays that are in Protocol Reparse Mode

masto:
# What does this implement/fix? 

If a Nextion display is in Protocol Reparse Mode, the standard communication protocol doesn't work, and it's necessary to send an escape sequence to get out of it.

This makes it possible to communicate with a Sonoff NSPanel (and presumably other devices in the same state).

Tip from fvanroie via
https://community.home-assistant.io/t/sonoff-nspanel-smart-scene-wall-switch-by-itead-coming-soon-on-kickstarter/332962/27

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [X] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).
  
If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #5016, opened 2023-06-27
Handle Nextion displays that are in Protocol Reparse Mode

edwardtfn:
This is an update of https://github.com/esphome/esphome/pull/2956 (from @masto) as the proposed workaround in the comments there is not solving the problem in many cases and that PR is failing compiling since ESPHome v2023.6.0.

If a Nextion display is in Protocol Reparse Mode, the standard communication protocol doesn't work, and it's necessary to send an escape sequence to get out of it.

This makes it possible to communicate with a Sonoff NSPanel (and presumably other devices in the same state).

Tip from fvanroie via
https://community.home-assistant.io/t/sonoff-nspanel-smart-scene-wall-switch-by-itead-coming-soon-on-kickstarter/332962/27

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** Please take a look at #2956 

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/nextion/nextion.cpp
+++ b/esphome/components/nextion/nextion.cpp
@@ -46,7 +46,8 @@ bool Nextion::check_connect_() {
     this->reset_(false);
 
     this->ignore_is_setup_ = true;
-    this->send_command_("boguscommand=0");  // bogus command. needed sometimes after updating
+    this->send_command_("boguscommand=0");            // bogus command. needed sometimes after updating
+    this->send_command_("DRAKJHSUYDGBNCJHGJKSHBDN");  // escape Protocol Reparse mode if we're in it
     this->send_command_("connect");
 
     this->comok_sent_ = millis();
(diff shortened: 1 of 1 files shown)


jesserockz:
This can be sent from YAML without the need to override the whole nextion component.

```yaml
esphome:
  on_boot:
    priority: 601
    then:
      - lambda: id(my_nextion).send_command_printf("DRAKJHSUYDGBNCJHGJKSHBDN");

```

edwardtfn:
As I mentioned in my original comment, this workaround doesn't work.

jesserockz:
In any case, there is no point opening a PR that adds this code as it will never be merged. 
You can keep the branch in your own fork, but no need to "spam" the PR list with it.

masto:
IMO the bigger problem with this fix, as currently proposed, is that it forces the display out of reparse mode, thereby "unfixing" anyone who who actually wants it to be.

PR #4849, opened 2023-05-17
[ILI9xxx] Update ili9xxx_init.h code for the ILI9488 display for correct white balance

lucasreiners:
# What does this implement/fix?

Recently I had a project using an ILI9488 based display. I found out, that the white balance of the display is incorrect - during the initialization sequence of the display, some gamma values are set, which I improved in this pull request (I also used the official documentation for the display, but tweaked the values further.

You can see the difference in the following image (left is AFTER the change, right is BEFORE - the large numbers on the right are now displayed correctly in white, on the right with the existing init code, the values are displayed in a light blue instead of white):
CDA8ACB4-EC1D-4210-A950-FF938F080D98-COLLAGE

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** n/a

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** 
n/a

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #5022, opened 2023-06-28
[ILI9xxx] Revert "Update ili9xxx_init.h for correct white balance (#4849)"

### LINKED ISSUE #4603: ili9488 white looks Green/blueish
### The problem

I just installed 2023.6.1, which includes [esphome#4849](https://github.com/esphome/esphome/pull/4849)
I actually never had white balance or color issues before, now my display looks like what this pull request actually should fix.
In other words: what wasn't broken before in my case, now is.
it worked fine with ESPHome 2023.5.5

2023.6.1

Home Assistant Add-on

_No response_

ESP32

### Board

lolin_d32_pro

ILI9xxx

### Example YAML snippet

```yaml
display:
  - id: my_display
    platform: ili9xxx
    model: ili9488
    cs_pin: 14
    dc_pin: 27
    reset_pin: 33
    auto_clear_enabled: true
```

_No response_

### Additional information

_No response_

lnicolas83:

revert commit #4849 that breaks correct display of colors.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4603

## Test Environment

- [ x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:


nielsnl68:
Sorry, i do not advice to roll back the changes. Because it worked fine for other people.
My suggestion in the short term is to add a new ili9488 display driver with the for your display, we could name for example `ili9488_A` and ili9488_B' etc.

Better is introducing a new property that can be used to choose which color palate needs to be used.

I hope you understand my resistance to this PR.

lnicolas83:
Can you look at the before and after pictures of commit 4849:

Before: true white and true blue
after: fake white and blue

Wouldn't it be better to restore the default colormap (and therefore undo commit 4849) and then have someone provide a PR to optionally change it?

Also, if a user who doesn't want true white (#FFFFFF) and true blue (#0000FF) can also just use another value in their yaml config.

nielsnl68:
As I said above, someone with a different ili9488 display, did a lot of work to get his display work properly with the right colors.
That is why i suggest to allow both colormaps to exist and that the end-user can choose which he need.

lnicolas83:
OK, I understand your logic, the person managed to make his screen work and it doesn't matter if he breaks the operation of the others, which worked very well before.

I'm going to stay on version 2023.5.5 in case other people decide to break the existing one (instead of just adding a feature without breaking anything).

THANKS.

PR #5038, opened 2023-07-02
Fix binary_sensor initial publish

### LINKED ISSUE #2537: Use of delayed_on_off filter in Binary Sensor Component breaks HA sensor updates
### The problem

Given code example, on program upload / reset, sensor in Home Assistant does not follow pin state nor change value at all.
Automations in ESPHome continue to function however.  I.E. "switch.turn_off: dev_genstartstopssr_plb150" would fire after 10 second delay.

2021.9.3

Home Assistant Add-on

core-2021.10.1

ESP32

### Board

Adafruit Feather ESP32

Binary Sensor Filters

### Example YAML snippet

```yaml
- platform: gpio
    pin: GPIO36
    name: "Dev-(+3v3_U)"
    id: util_power
    filters:
      - delayed_on_off: 10s
    on_release:
      then:
        - switch.turn_off: dev_genstartstopssr_plb150
```

_No response_

### Additional information

_No response_

dudanov:

Fixes an error in the initial publication of the state of the binary sensor during the use of filters.

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/2537

https://community.home-assistant.io/t/esphome-fails-to-update-sensor-state-in-ha-when-using-4-second-delay-filters-workaround/347541

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/binary_sensor/filter.cpp
+++ b/esphome/components/binary_sensor/filter.cpp
@@ -21,7 +21,7 @@ void Filter::output(bool value, bool is_initial) {
 }
 void Filter::input(bool value, bool is_initial) {
   auto b = this->new_value(value, is_initial);
-  if (b.has_value()) {
+  if (b.has_value() | !this->dedup_.has_value()) {
     this->output(*b, is_initial);
   }
 }
--- a/esphome/components/binary_sensor/filter.cpp
+++ b/esphome/components/binary_sensor/filter.cpp
@@ -22,7 +22,7 @@ void Filter::output(bool value, bool is_initial) {
 void Filter::input(bool value, bool is_initial) {
   auto b = this->new_value(value, is_initial);
   if (b.has_value() || !this->dedup_.has_value()) {
-    this->output(*b, is_initial);
+    this->output(b.value_or(false), is_initial);
   }
 }
 
--- a/esphome/components/binary_sensor/filter.cpp
+++ b/esphome/components/binary_sensor/filter.cpp
@@ -21,7 +21,7 @@ void Filter::output(bool value, bool is_initial) {
 }
 void Filter::input(bool value, bool is_initial) {
   auto b = this->new_value(value, is_initial);
-  if (b.has_value() | !this->dedup_.has_value()) {
+  if (b.has_value() || !this->dedup_.has_value()) {
     this->output(*b, is_initial);
   }
 }


esphome/components/binary_sensor/filter.cpp
```diff
@@ -21,7 +21,7 @@ void Filter::output(bool value, bool is_initial) {
 }
 void Filter::input(bool value, bool is_initial) {
   auto b = this->new_value(value, is_initial);
-  if (b.has_value()) {
+  if (b.has_value() | !this->dedup_.has_value()) {
```
jesserockz:
```suggestion
  if (b.has_value() || !this->dedup_.has_value()) {
```

dudanov: ha-ha. And how did I make such a ridiculous mistake... Binary values are to blame... thanks, Jesse!

esphome/components/binary_sensor/filter.cpp
```diff
@@ -21,7 +21,7 @@ void Filter::output(bool value, bool is_initial) {
 }
 void Filter::input(bool value, bool is_initial) {
   auto b = this->new_value(value, is_initial);
-  if (b.has_value()) {
+  if (b.has_value() || !this->dedup_.has_value()) {
```
jesserockz, line 24: Hmm, if b doesnt have a value, then what is this actually going to publish?

jesserockz:
Im gonna close this as I think the filters should not alter the data unexpectedly like this. #5091 should fix the problem you were having by publishing false as initial. Let me know if something else is still required though.

dudanov:
> Im gonna close this as I think the filters should not alter the data unexpectedly like this. #5091 should fix the problem you were having by publishing false as initial. Let me know if something else is still required though.

Jesse, I wouldn't create a PR or make corrections if it wasn't required. #5091  does not fix the whole problem. Since the method `publish_initial_state` still runs through all the filters that delay the publication of the signal. If you are familiar with automation, then there are no and should not be indeterminate states, exceptions in electrical engineering are the so-called Z-states. The standard automation library has functional blocks TON, TOF, etc., they all have certain states at the output. This PR is necessary as it solves this problem.

jesserockz:
The main issue I see here is that if you specifically have a filter that delays any state being reported until you need a true or false from startup, this code will break that by forcing a `false` return.

A `BinarySensor` has `.has_state()` to tell you if this sensor has a state published yet or not.

I'll reopen, but not satisfied with the solution (yet)

dudanov:
> A `BinarySensor` has `.has_state()` to tell you if this sensor has a state published yet or not.

In lambda, yes, I can. But this is not an option, especially since the user interface cannot be configured out of the box to display `false` instead of `undefined`.

PR #5051, opened 2023-07-04
Preparing color.h for IDF >= 5

HeMan:

Compiler started to give warnings about operator= in color.h when preparing other components for IDF >= 5. Adding copy constructor and changing is_on to const gave a suggestion from compiler to change argument to const reference on display-classes. It's a breaking change for external modules based on the display class.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/core/color.h
+++ b/esphome/core/color.h
@@ -43,7 +43,9 @@ struct Color {
                                                             b((colorcode >> 0) & 0xFF),
                                                             w((colorcode >> 24) & 0xFF) {}
 
-  inline bool is_on() ALWAYS_INLINE { return this->raw_32 != 0; }
+  inline Color(const Color &other) : raw_32(other.raw_32) {}
+
+  inline bool is_on() const ALWAYS_INLINE { return this->raw_32 != 0; }
 
   inline bool operator==(const Color &rhs) {  // NOLINT
     return this->raw_32 == rhs.raw_32;
--- a/esphome/components/addressable_light/addressable_light_display.cpp
+++ b/esphome/components/addressable_light/addressable_light_display.cpp
@@ -52,7 +52,7 @@ void AddressableLightDisplay::display() {
   }
 }
 
-void HOT AddressableLightDisplay::draw_absolute_pixel_internal(int x, int y, Color color) {
+void HOT AddressableLightDisplay::draw_absolute_pixel_internal(int x, int y, const Color &color) {
   if (x >= this->get_width_internal() || x < 0 || y >= this->get_height_internal() || y < 0)
     return;
 
--- a/esphome/components/addressable_light/addressable_light_display.h
+++ b/esphome/components/addressable_light/addressable_light_display.h
@@ -47,7 +47,7 @@ class AddressableLightDisplay : public display::DisplayBuffer, public PollingCom
  protected:
   int get_width_internal() override;
   int get_height_internal() override;
-  void draw_absolute_pixel_internal(int x, int y, Color color) override;
+  void draw_absolute_pixel_internal(int x, int y, const Color &color) override;
   void update() override;
 
   light::LightState *light_state_;
--- a/esphome/components/display/display.cpp
+++ b/esphome/components/display/display.cpp
@@ -12,10 +12,10 @@ static const char *const TAG = "display";
 const Color COLOR_OFF(0, 0, 0, 0);
 const Color COLOR_ON(255, 255, 255, 255);
 
-void Display::fill(Color color) { this->filled_rectangle(0, 0, this->get_width(), this->get_height(), color); }
+void Display::fill(const Color &color) { this->filled_rectangle(0, 0, this->get_width(), this->get_height(), color); }
 void Display::clear() { this->fill(COLOR_OFF); }
 void Display::set_rotation(DisplayRotation rotation) { this->rotation_ = rotation; }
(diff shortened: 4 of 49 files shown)


guillempages:
Are you sure this is the right way to go? When the "=" operator is used now, won't this convert the color first to a 32 bit int, do some bit-magic and create a new `Color`? Won't we lose performance here? When drawing images, the "=" operator is being used on the hot loop IIRC.

HeMan:
> Are you sure this is the right way to go? When the "=" operator is used now, won't this convert the color first to a 32 bit int, do some bit-magic and create a new `Color`? Won't we lose performance here? When drawing images, the "=" operator is being used on the hot loop IIRC.

No, I'm not sure. The message I got was:
```
src/esphome/core/color.h: In member function 'esphome::Color esphome::Color::operator+(const esphome::Color&) const':
src/esphome/core/color.h:116:12: warning: implicitly-declared 'constexpr esphome::Color::Color(const esphome::Color&)' is deprecated [-Wdeprecated-copy]
  116 |     return ret;
      |            ^~~
src/esphome/core/color.h:61:17: note: because 'esphome::Color' has user-provided 'esphome::Color& esphome::Color::operator=(const esphome::Color&)'
   61 |   inline Color &operator=(const Color &rhs) ALWAYS_INLINE {  // NOLINT
      |                 ^~~~~~~~
src/esphome/core/color.h: In member function 'esphome::Color esphome::Color::operator-(const esphome::Color&) const':
src/esphome/core/color.h:139:12: warning: implicitly-declared 'constexpr esphome::Color::Color(const esphome::Color&)' is deprecated [-Wdeprecated-copy]
  139 |     return ret;
      |            ^~~
src/esphome/core/color.h:61:17: note: because 'esphome::Color' has user-provided 'esphome::Color& esphome::Color::operator=(const esphome::Color&)'
   61 |   inline Color &operator=(const Color &rhs) ALWAYS_INLINE {  // NOLINT
      |                 ^~~~~~~~
src/esphome/core/color.h: In member function 'esphome::Color esphome::Color::gradient(const esphome::Color&, uint8_t)':
src/esphome/core/color.h:164:12: warning: implicitly-declared 'constexpr esphome::Color::Color(const esphome::Color&)' is deprecated [-Wdeprecated-copy]
  164 |     return new_color;
      |            ^~~~~~~~~
src/esphome/core/color.h:61:17: note: because 'esphome::Color' has user-provided 'esphome::Color& esphome::Color::operator=(const esphome::Color&)'
   61 |   inline Color &operator=(const Color &rhs) ALWAYS_INLINE {  // NOLINT
      |                 ^~~~~~~~

```
I'll try adding an explicit copy constructor instead, could that work?

guillempages (approved):
looks good to me

HeMan:
> looks good to me

To bad it failed CI... Reworking it now

esphome/core/color.h
```diff
@@ -43,7 +43,9 @@ struct Color {
                                                             b((colorcode >> 0) & 0xFF),
                                                             w((colorcode >> 24) & 0xFF) {}
 
-  inline bool is_on() ALWAYS_INLINE { return this->raw_32 != 0; }
+  inline Color(const Color &other_color) : raw_32(other_color.raw_32) {}
```
guillempages: AFAIK the C++ convention for copy constructors is to call the original "other", but I might be mistaken here.
```suggestion
  inline Color(const Color &other) : raw_32(other.raw_32) {}
```

guillempages (commented):
Wow, what a rabbit hole you got yourself into :-D

The changes look good to me; just a small suggestion regarding name, but feel free to ignore it.

This might have conflicts with #5001 though; ~~IMHO it would be better to wait until that one is merged (hopefully soon), and rebase on top of that.~~ Now that it has been merged, you should probably rebase.

guillempages:
Also this should probably be marked as a breaking change; if you had to change the const'ness of all the displays, external components will also break and need to change that.

(Don't get me wrong; I still think these changes you are making are meaningful and good)

guillempages (approved):
Looks good to me.

As mentioned above this might be a breaking change for external display components. IMHO it should be put in the same release as #5002 (which is also a breaking change for external display components), so that those components do not need to react to breaking changes twice.

ayufan:
@HeMan I wonder, can we just remove the offended `operator =` instead? It seems to serve no function, and is rather slower due to copying each byte individually.

Adding `const Color &` everywhere is not the best idea (from performance point of view) since it makes to pass a pointer, and getting the value requires doing dereference of the object, where as the current `Color` does fit the 32bit register.

Also the description says "Compiler started to give warnings about operator= in color.h when preparing other components for IDF >= 5.". It would help to have those warning listed in the description.

IMHO. Just remove `operator = (const Color &`.

cc @guillempages

HeMan:
> Reference

Hmm, @guillempages had the suspicion that that it would cast to uint32_t to let the other operator= do it's magic, but I tried to remove both operators and everything seem to work fine. I think that is the best solution, but I'll create another PR to let you decide which to use.

ayufan:
@HeMan Thanks. I think we should merge the https://github.com/esphome/esphome/pull/5070, not this PR.

PR #5061, opened 2023-07-05
Prepare some components for IDF >= 5

HeMan:

Prepares enough components for IDF >= 5 so that test5.yaml works with latest version. 

## Types of changes

- [X] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/bedjet/bedjet_hub.cpp
+++ b/esphome/components/bedjet/bedjet_hub.cpp
@@ -3,6 +3,7 @@
 #include "bedjet_hub.h"
 #include "bedjet_child.h"
 #include "bedjet_const.h"
+#include <cinttypes>
 
 namespace esphome {
 namespace bedjet {
@@ -373,7 +374,7 @@ void BedJetHub::gattc_event_handler(esp_gattc_cb_event_t event, esp_gatt_if_t ga
       if (this->last_notify_ == 0 || delta > MIN_NOTIFY_THROTTLE || this->force_refresh_) {
         // Set reentrant flag to prevent processing multiple packets.
         this->processing_ = true;
-        ESP_LOGVV(TAG, "[%s] Decoding packet: last=%d, delta=%d, force=%s", this->get_name().c_str(),
+        ESP_LOGVV(TAG, "[%s] Decoding packet: last=%" PRId32 ", delta=%" PRId32 ", force=%s", this->get_name().c_str(),
                   this->last_notify_, delta, this->force_refresh_ ? "y" : "n");
         bool needs_extra = this->codec_->decode_notify(param->notify.value, param->notify.value_len);
 
@@ -523,11 +524,11 @@ void BedJetHub::dispatch_status_() {
 
       ESP_LOGI(TAG, "[%s] Still waiting for first GATT notify event.", this->get_name().c_str());
     } else if (diff > NOTIFY_WARN_THRESHOLD) {
-      ESP_LOGW(TAG, "[%s] Last GATT notify was %d seconds ago.", this->get_name().c_str(), diff / 1000);
+      ESP_LOGW(TAG, "[%s] Last GATT notify was %" PRId32 " seconds ago.", this->get_name().c_str(), diff / 1000);
     }
 
     if (this->timeout_ > 0 && diff > this->timeout_ && this->parent()->enabled) {
-      ESP_LOGW(TAG, "[%s] Timed out after %d sec. Retrying...", this->get_name().c_str(), this->timeout_);
+      ESP_LOGW(TAG, "[%s] Timed out after %" PRId32 " sec. Retrying...", this->get_name().c_str(), this->timeout_);
       // set_enabled(false) will only close the connection if state != IDLE.
       this->parent()->set_state(espbt::ClientState::CONNECTING);
       this->parent()->set_enabled(false);
--- a/esphome/components/esp32_ble/ble_uuid.cpp
+++ b/esphome/components/esp32_ble/ble_uuid.cpp
@@ -4,6 +4,7 @@
 
 #include <cstring>
 #include <cstdio>
+#include <cinttypes>
 #include "esphome/core/log.h"
 
 namespace esphome {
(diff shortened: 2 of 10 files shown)


esphome/components/bedjet/bedjet_hub.cpp
```diff
@@ -373,7 +374,7 @@ void BedJetHub::gattc_event_handler(esp_gattc_cb_event_t event, esp_gatt_if_t ga
       if (this->last_notify_ == 0 || delta > MIN_NOTIFY_THROTTLE || this->force_refresh_) {
         // Set reentrant flag to prevent processing multiple packets.
         this->processing_ = true;
-        ESP_LOGVV(TAG, "[%s] Decoding packet: last=%d, delta=%d, force=%s", this->get_name().c_str(),
+        ESP_LOGVV(TAG, "[%s] Decoding packet: last=%" PRId32 ", delta=% " PRId32 ", force=%s", this->get_name().c_str(),
```
jhansche: Looks like there's a space added here? Won't be a valid replacement

jhansche (changes requested):
For the record: Only looked at `bedjet`

esphome/components/bedjet/bedjet_hub.cpp
HeMan: Ah, good you spotted that! Fixing it.

DAVe3283:
Tested `max31865` on an ESP8266 and it compiles/runs as expected. Can't test the ESP-IDF as I don't have a spare 31865 to hook to an ESP32. But the changes look good to me.

jhansche:
Just thinking out loud here: is there any reason we wouldn't be able to handle these automatically in the ESP_LOG* macro(s)?

I guess it would involve either converting the %d arguments to the expected type at runtime, or doing a string replacement of the %d format placeholder, which might be more effort than we want to go through 🤔

esphome/components/esp32_ble/ble_uuid.cpp
```diff
@@ -166,7 +167,7 @@ std::string ESPBTUUID::to_string() const {
     case ESP_UUID_LEN_16:
       return str_snprintf("0x%02X%02X", 6, this->uuid_.uuid.uuid16 >> 8, this->uuid_.uuid.uuid16 & 0xff);
```
jesserockz, line 168: Shouldnt this be changed to match too?

jhansche: I would expect that %X should be unambiguous? It sounds like it's just (u)int(32) that's changing?

Edit: nm I see now that the next one changed below. This seems odd to me, since an 8bit int in hex format shouldn't be ambiguous 🤔

esphome/components/max31865/max31865.cpp
```diff
@@ -45,14 +46,15 @@ void MAX31865Sensor::update() {
     config = this->read_register_(CONFIGURATION_REG);
     fault_detect_time = micros() - start_time;
     if ((fault_detect_time >= 6000) && (config & 0b00001100)) {
-      ESP_LOGE(TAG, "Fault detection incomplete (0x%02X) after %uμs (datasheet spec is 600μs max)! Aborting read.",
+      ESP_LOGE(TAG,
+               "Fault detection incomplete (0x%02X) after %" PRIu32 "μs (datasheet spec is 600μs max)! Aborting read.",
```
jhansche, line 50: Another %X here too

esphome/components/sen5x/sen5x.cpp
```diff
@@ -252,7 +253,7 @@ void SEN5XComponent::dump_config() {
   ESP_LOGCONFIG(TAG, "  Firmware version: %d", this->firmware_version_);
```
jhansche, line 253: I admit I'm not entirely sure what the goal is, or why some placeholders are replaced but others aren't.

If it's necessary for the new IDF, could we come up with a lint check that would make these easier to spot?

The new macro admittedly feels clunky to me 🤷‍♂️

esphome/components/esp32_ble/ble_uuid.cpp
HeMan: Hmm, it compiles both without changes and when changed to `PRIX16`. It could be that uint16_t doesn't change between IDF 4 and IDF 5. Anyway, I'll match with PRIX16.

esphome/components/max31865/max31865.cpp
HeMan: Similar resolution here as in https://github.com/esphome/esphome/pull/5061#discussion_r1257772707, uint8_t doesn't change between IDF 4 and 5, but I change to match.

HeMan:
When I started to look in to it, not sure if all the format specifiers should be macro'ified. Changing all won't give any benefits now, but could reduce work in the future if datatypes changes. 
Just changing the bare minimum is easier since it fails to compile on IDF => 5, and it reduces noise in that the PR's gets smaller.
What do you say @jesserockz @jhansche?

jhansche:
> Just changing the bare minimum is easier since it fails to compile on IDF => 5, and it reduces noise in that the PR's gets smaller.

My concern is that the new ESP_LOG* invocations look clunky and overly verbose. It doesn't look like a natural logging statement. That's why my [earlier question](https://github.com/esphome/esphome/pull/5061#issuecomment-1622446107) was trying to propose ideas to handle this in a different way - one that _doesn't_ require changing every place we use `%u` for example; instead it would be handled by the ESP_LOG* macros internally, but only on IDF5+. I can't think of a magical way to do that that doesn't also become a heavy lift, unfortunately (like a string replacement JIT at runtime)

~What exactly is the change? Is there a link to documentation explaining what is actually causing this? Sounds like you're saying that IDF5 changes the definition of `uint32_t` (etc)?~
EDIT: Found the link to docs: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/migration-guides/release-5.x/5.0/gcc.html#int32-t-and-uint32-t-for-xtensa-compiler
It seems very weird to me, that this is their recommended solution. Inlining a formatting char via macro at compile time just doesn't feel like the most reasonable way to handle this.. 🤔  I might even go as far as to say it looks bad 😬

esphome/components/esp32_ble/ble_uuid.cpp
jesserockz: Either that, or change the arguments, considering they are all technically bit shifted down to uint8_t anyway, we could cast each one to keep the format string simpler.

esphome/components/waveshare_epaper/waveshare_epaper.cpp
```diff
@@ -250,7 +251,7 @@ void WaveshareEPaperTypeA::dump_config() {
       ESP_LOGCONFIG(TAG, "  Model: 2.9inV2");
       break;
   }
-  ESP_LOGCONFIG(TAG, "  Full Update Every: %u", this->full_update_every_);
+  ESP_LOGCONFIG(TAG, "  Full Update Every: %" PRIu32, this->full_update_every_);
```
kbx81, line 254: Extra `%` symbol left here?

```suggestion
  ESP_LOGCONFIG(TAG, "  Full Update Every: " PRIu32, this->full_update_every_);
```

kbx81 (approved):

kbx81:
I'm fine with this -- any final comments before we merge?

PR #5068, opened 2023-07-07
binary_sensor: Validate max_length for on_click/on_double_click

kpfleming:

Raise an error if max_length is not higher than min_length. Such a configuration will not work as the user expects, as the automation with the invalid max_length will never be triggered. Validation for `on_multi_click` already ensures the same thing.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/binary_sensor/__init__.py
+++ b/esphome/components/binary_sensor/__init__.py
@@ -359,6 +359,18 @@ def validate_multi_click_timing(value):
 validate_device_class = cv.one_of(*DEVICE_CLASSES, lower=True, space="_")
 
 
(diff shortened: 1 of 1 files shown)


kpfleming:
I discovered this problem by setting `min_length` to 1s, and not setting `max_length`, assuming that the result would be what I wanted... it was not, since the default for `max_length` is 350ms.

jesserockz (approved):

jesserockz:
Thanks

PR #5069, opened 2023-07-07
New 'Duty Time' sensor component

dudanov:

The component `Duty Time` is used to calculate the operating time of something, for example, a light bulb in `seconds`.

It is configured by one of the logical sources: a `binary_sensor` or a `lambda` function.

The main loop checks the source for a change in the input signal. The auxiliary interval loop `update()` of the `PollingComponent` class periodically updates the sensor.

If necessary, it is possible to add an auxiliary sensor to record the time of the last switch-on.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3049

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example
# Test "object"
switch:
  - platform: template
    name: Test Object
    id: my_switch
    optimistic: on

# binary_sensor
binary_sensor:
  - platform: template
    id: my_binary_sensor
    name: Object State
    lambda: "return id(my_switch).state;"

# Reset Button
button:
  - platform: template
    name: Reset Duty Time
    on_press:
      sensor.duty_time.reset: id_duty

# duty_time component
sensor:
  - platform: duty_time
    id: id_duty
    name: Duty Time
    restore: on
    last_time:
      name: Last Duty Time
    # binary_sensor as logical source
    sensor: my_binary_sensor
    # .. EOR lambda
    #lambda: "return id(my_switch).state;"
```

## Checklist:

### DIFF
--- a/esphome/components/duty_time/duty_time_sensor.cpp
+++ b/esphome/components/duty_time/duty_time_sensor.cpp
@@ -47,7 +47,7 @@ void DutyTimeSensor::setup() {
 }
 
 void DutyTimeSensor::set_value_(const uint32_t sec) {
-  this->last_time_ = 0;           // initial ms correction
+  this->last_time_ = 0;
   if (this->last_state_)
     this->last_time_ = millis();  // last time with 0 ms correction
   this->total_sec_ = sec;
--- a/esphome/components/duty_time/duty_time_sensor.cpp
+++ b/esphome/components/duty_time/duty_time_sensor.cpp
@@ -47,10 +47,11 @@ void DutyTimeSensor::setup() {
 }
 
 void DutyTimeSensor::set_value_(const uint32_t sec) {
-  this->edge_ms_ = 0;
-  this->edge_sec_ = sec;
-  this->last_update_ = millis();
-  this->publish_and_save_(sec, 0);
+  this->last_time_ = 0;           // initial ms correction
+  if (this->last_state_)
+    this->last_time_ = millis();  // last time with 0 ms correction
+  this->total_sec_ = sec;
+  this->publish_and_save_(0);
 }
 
 void DutyTimeSensor::process_state_(const bool state) {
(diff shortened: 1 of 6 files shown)


SeByDocKy:
Sound an interesting addition. Can be useful for example to build a kind of "historical stats" if you can add a "period" field to compute the ratio. More adding an action to "reset" the sensor would be great too

dudanov:
> Sound an interesting addition. Can be useful for example to build a kind of "historical stats" if you can add a "period" field to compute the ratio. More adding an action to "reset" the sensor would be great too

[Documentation](https://deploy-preview-3049--esphome.netlify.app/components/sensor/duty_time.html)

esphome/components/duty_time/duty_time_sensor.cpp
```diff
@@ -0,0 +1,100 @@
+#include "duty_time_sensor.h"
+#include "esphome/core/hal.h"
+
+namespace esphome {
+namespace duty_time_sensor {
+
+static const char *const TAG = "duty_time_sensor";
+
+void DutyTimeSensor::set_sensor(const binary_sensor::BinarySensor *const sensor) {
+  this->func_ = [sensor]() { return sensor->state; };
```
jesserockz: Would it not be better to attach to the state callback and call process when it changes instead of checking every loop?

dudanov: I did that at first. But the lambda solution, which is common for both cases, seemed to me to be common, with less code. The loop method is still overridden in the component and each loop gets control, so let both cases work through a lambda.

dudanov: Jesse, if you strongly advise on the use of callback functions, I will implement it without any problems. All the same, there is also a question of the general style of the project. :)

dudanov: Done, @jesserockz :)
https://github.com/esphome/esphome/pull/5069/commits/8492b47e43f5eb527969349639ef183925b4079e

jesserockz (approved):

PR #5074, opened 2023-07-09
Make docker use pip installed pillow

### LINKED ISSUE #4398: Docker: Pillow old version
### The problem

When adding BDF fonts (created in Fontforge) that use the 10646-1 codepage, ESPHome will hang (on Pillow), not allowing certain characters like the Euro symbol. All BDF fonts are interpreted as Latin-1 character sets only.  It may also have problems with Unicode TTF fonts.

I think this may have been fixed in recent versions of Pillow but after going into the container, I can see the version of Pillow is quite old and since it is system-owned, unsure how to update it.

Is my only way to update pillow by recreating the container myself?

2023.3.2

Docker

N/A

ESP32

### Board

ESP8266 / ESP32

python pillow

### Example YAML snippet

```yaml
font: 
  - file: ehmtx/MatrixClock-Light8-Unicode.bdf #
    id: ehmtx_font
    glyphs:  |
      ! "#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz€
```

```txt
This is while compiling for the EspHoMaTriX (ehmtx) project:

https://github.com/lubeda/EsphoMaTrix
```

### Additional information

You can try out this font here:

https://github.com/darkpoet78/MatrixClockFonts/blob/main/8-series/MatrixClock-Light8-Unicode.bdf

HeMan:

Don't apt install pillow to make docker container use version from requirements_optional.txt

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4398

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,16 +22,22 @@ RUN \
         python3=3.9.2-3 \
         python3-pip=20.3.4-4+deb11u1 \
         python3-setuptools=52.0.0-4 \
-        python3-pil=8.1.2+dfsg-0.3+deb11u1 \
         python3-cryptography=3.3.2-1 \
         python3-venv=3.9.2-3 \
         iputils-ping=3:20210202-1 \
         git=1:2.30.2-1+deb11u2 \
         curl=7.74.0-1.3+deb11u7 \
         openssh-client=1:8.4p1-5+deb11u1 \
-        libcairo2=1.16.0-5 \
-        python3-cffi=1.14.5-1 \
-    && rm -rf \
+        python3-cffi=1.14.5-1; \
+    if [ "$TARGETARCH$TARGETVARIANT" = "armv7" ]; then \
+        apt-get install -y --no-install-recommends \
+          build-essential=12.9 \
+          python3-dev=3.9.2-3 \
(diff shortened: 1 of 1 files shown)


jesserockz (changes requested):
Removing build dependencies in a different dockerfile line does not free the space they originally took in the image due to the way docker layers work.

HeMan:
> Removing build dependencies in a different dockerfile line does not free the space they originally took in the image due to the way docker layers work.

It wasn't so much the space, it was more to keep similar content in the container after the change. Should I update the PR?

jesserockz:
I don't think we even need those extra packages, I just built the image with `python-pil` removed and it still have pillow installed from requirements_optional correctly.

HeMan:
> I don't think we even need those extra packages, I just built the image with `python-pil` removed and it still have pillow installed from requirements_optional correctly.

It's only the ARMv7 builds that need them, I'll update the Dockerfile to only install there.

jesserockz (approved):

PR #5093, opened 2023-07-12
Grove amend name

max246:

@glmnet  pointed out I picked a wrong name when I made the driver name, this should reflect how it is usually done

## Types of changes

- [x] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3059

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
i2c:
  sda: 4
  scl: 5
  scan: true
  id: bus_a
grove_tb6612fng:
    id: test_motor
    name: motor_outside
    address: 0x14

switch:
  - platform: template
    name: open_vent
    id: open_vent
    optimistic: True
    on_turn_on: 
      then:
        - grove_tb6612fng.motor_run:
            channel: 1
            speed: 255
            direction: BACKWARD
            id: test_motor
        - delay: 10s
        - switch.turn_on: stop_motor
        - switch.turn_off:  open_vent
```

## Checklist:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -104,7 +104,7 @@ esphome/components/gp8403/* @jesserockz
 esphome/components/gpio/* @esphome/core
 esphome/components/gps/* @coogle
 esphome/components/graph/* @synco
-esphome/components/grove_i2c_motor/* @max246
+esphome/components/grove_tb6612fng/* @max246
 esphome/components/growatt_solar/* @leeuwte
 esphome/components/haier/* @paveldn
 esphome/components/havells_solar/* @sourabhjaiswal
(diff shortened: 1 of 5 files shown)


randybb:
But to be exact this is still not a great name. I would call it "seedstudio i2c motor driver", because this component is driving that stm32f0, not the motor driver itself. And "grove" refers to the  HY2.0-4P for i2c called "grove", so it doesn't make sense.

max246:
> he motor dr

Grove is the "family name" of the sensors which is stated in the doc as well https://wiki.seeedstudio.com/Grove-I2C_Motor_Driver-TB6612FNG/

There are 3 types of this board which have different drive, therefore the name TB6612FNG. Someone can extend by interfacing the others motor drivers but for consumption wise, this board benefits of a standby mode which is better suited for IoT stuff.

glmnet:
@randybb comment is very accurate, I wasn't aware there was an MCU in in front of the chip nor that grove was the name given to the connector.
Naming it seed studio could conflict a new product from the company which also uses i2c to drive motors, who knows.
Also removing the tb6612fng part from it could lead to confusion as there are many i2c boards which are likely not compatible.
Having said that I'm ok with this name as it clearly points to the "off the shelf" hardware this component can drive.

jesserockz (approved):

PR #5101, opened 2023-07-14
ESP-IDF SPI LED strip (APA102/SK9822)

dentra:

This PR adds support for SPI LED strips on APA102 and SK9822 chips. This is especially true for the ESP-IDF platform.
Tested with my Lilygo Dongle S3 with one led.

~~I will do PR in the docs a little later.~~

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3064

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
light:
  - platform: esp32_spi_led_strip
    name: "Led Light"
    data_pin: GPIO40
    clock_pin: GPIO39
    num_leds: 1
    rgb_order: RGB
    max_refresh_rate: 50us
```

## Checklist:

### DIFF
--- a/esphome/components/esp32_spi_led_strip/led_strip.cpp
+++ b/esphome/components/esp32_spi_led_strip/led_strip.cpp
@@ -1,3 +1,4 @@
+#ifdef USE_ESP_IDF
 #include "esphome/core/log.h"
 
 #include "led_strip.h"
@@ -71,7 +72,7 @@ void LedStripSpi::setup() {
 
   // clear all pixels and set brightness to maximum, skipping start frame
   for (int i = 1; i <= this->num_leds_; i++) {
-    this->buf_[i * this->frame_size_] = 0xFF;
+    this->buf_[i * FRAME_SIZE] = 0xFF;
   }
   this->spi_flush_();
 }
@@ -174,7 +175,7 @@ light::ESPColorView LedStripSpi::get_view_internal(int32_t index) const {
       b = OFFSET_B;
       break;
   }
-  auto offset = (index + 1) * this->frame_size_;  // skip start frame
+  auto offset = (index + 1) * FRAME_SIZE;  // skip start frame
   return {
       this->buf_ + offset + r,     // red
       this->buf_ + offset + g,     // green
@@ -187,3 +188,4 @@ light::ESPColorView LedStripSpi::get_view_internal(int32_t index) const {
 
 }  // namespace esp32_spi_led_strip
 }  // namespace esphome
+#endif
--- a/esphome/components/esp32_spi_led_strip/led_strip.cpp
+++ b/esphome/components/esp32_spi_led_strip/led_strip.cpp
@@ -33,7 +33,7 @@ void LedStripSpi::dump_config() {
   ESP_LOGCONFIG(TAG, "ESP32 SPI LED Strip:");
   ESP_LOGCONFIG(TAG, "  Data Pin: GPIO%d", this->data_pin_);
   ESP_LOGCONFIG(TAG, "  Clock Pin: GPIO%d", this->clock_pin_);
-  ESP_LOGCONFIG(TAG, "  Number of LEDs: %d" PRId16, this->num_leds_);
+  ESP_LOGCONFIG(TAG, "  Number of LEDs: %d", this->num_leds_);
   ESP_LOGCONFIG(TAG, "  RGB Order: %s", rgb_order_to_string(this->rgb_order_));
   if (this->max_refresh_rate_) {
     ESP_LOGCONFIG(TAG, "  Max Refresh Rate: %" PRIu32 " Hz", this->max_refresh_rate_);
--- a/esphome/components/esp32_spi_led_strip/led_strip.cpp
+++ b/esphome/components/esp32_spi_led_strip/led_strip.cpp
@@ -1,4 +1,4 @@
-#ifdef USE_ESP_IDF
+#ifdef USE_ESP32
 #include "esphome/core/log.h"
 
 #include "led_strip.h"
@@ -28,7 +28,7 @@ inline static const char *rgb_order_to_string(RGBOrder order) {
 }
 
 void LedStripSpi::dump_config() {
-  ESP_LOGCONFIG(TAG, "ESP-IDF SPI LED Strip:");
+  ESP_LOGCONFIG(TAG, "ESP32 SPI LED Strip:");
   ESP_LOGCONFIG(TAG, "  Data Pin: GPIO%d", this->data_pin_);
   ESP_LOGCONFIG(TAG, "  Clock Pin: GPIO%d", this->clock_pin_);
   ESP_LOGCONFIG(TAG, "  Number of LEDs: %d", this->num_leds_);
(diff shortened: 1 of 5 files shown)


kbx81:
Ah! I was just thinking we needed this. 😄
Please be sure to add a simple test into one of the `test*.yaml` files. I'll try to test this here and review this week.
Thanks for working on this! 🍻

jesserockz (commented):
Why is this restricted to esp-idf? It should work on arduino too as esp-idf headers/code are available there

dentra:
added test and support of arduino platform

kbx81:
Tested this today on my Adafruit Funhouse board and it works beautifully! Now I can use ESP-IDF alone on this hardware. 😁

That said, it's getting into some trouble when using the Arduino framework. It appears it keeps crashing -- although it doesn't crash _immediately_ (seems to take about a minute or two). This board does not have a proper serial adapter (it depends on the ESP's USB interface) so I can't see the logs. I'll see if I can try it on different hardware that allows me to see what's up, but I figured I'd mention it here in the meantime. 😇

Overall, good work!

esphome/components/esp32_spi_led_strip/led_strip.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+void LedStripSpi::dump_config() {
+  ESP_LOGCONFIG(TAG, "ESP32 SPI LED Strip:");
+  ESP_LOGCONFIG(TAG, "  Data Pin: GPIO%d", this->data_pin_);
+  ESP_LOGCONFIG(TAG, "  Clock Pin: GPIO%d", this->clock_pin_);
+  ESP_LOGCONFIG(TAG, "  Number of LEDs: %d", this->num_leds_);
+  ESP_LOGCONFIG(TAG, "  RGB Order: %s", rgb_order_to_string(this->rgb_order_));
+  if (this->max_refresh_rate_) {
+    ESP_LOGCONFIG(TAG, "  Max Refresh Rate: %u Hz", this->max_refresh_rate_);
```
kbx81: For IDF 5+ compatibility:
```suggestion
    ESP_LOGCONFIG(TAG, "  Max Refresh Rate: %" PRIu32 " Hz", this->max_refresh_rate_);
```

kbx81 (changes requested):
Quick thing I found relating to IDF 5+ compatibility

esphome/components/esp32_spi_led_strip/led_strip.cpp
kbx81: ...you'll also need to `#include <cinttypes>`

dentra: done.

dentra: do I also change %d -> PRId16 for num_leds_ and %d -> PRId8 for data_pin_ and clock_pin_?

kbx81: No need for those (but it also won't hurt) -- this is a change made in IDF 5+ that only affects 32-bit `int` types. 😄

kbx81:
Alrighty, I've been testing this pretty extensively today and have some feedback...and I apologize in advance for this because it's a bit lengthy. 😇 Regardless, I'm excited to have this component available and I'm sure it'll be widely used, so we need to get it just right. 😄

I'm testing with the [Funhouse board from Adafruit](https://www.adafruit.com/product/4985) because it has five APA102s along the top of it -- very convenient to use for this purpose. I've been building with either of these two configurations:

```yaml
esp32:
  board: adafruit_funhouse_esp32s2
  framework:
    type: arduino
    version: 2.0.9
    platform_version: 5.4.0
```

```yaml
esp32:
  board: adafruit_funhouse_esp32s2
  framework:
    type: esp-idf
    version: 5.0.2
    platform_version: 6.3.2
```

I've also tried Arduino with `platform_version: 6.3.2`.

In any event, here's what I've found, largely based on this board:

**SPI `display` + `esp32_spi_led_strip`**

Something very bad is happening when the ST7789 (and I'm going to make the leap guess other SPI displays, too) is used with this component, regardless of the framework. I have not dug into it deeply, but my gut is telling me that it's some conflict between this component's SPI initialization and what ESPHome is doing for the display. This brings us to a suggestion:

Instead of defining the data and clock pins within this component and having it do its own initialization of the SPI, should we instead use ESPHome's built-in SPI handling and then point this component at that? In other words, perhaps `esp32_spi_led_strip` could inherit from `SPIDevice` and (in the config) it would get a `spi_id` telling it which interface instance to use (which, naturally, includes the pins) -- I think this would mitigate any conflicts as it relates to the initialization of the hardware. It's not uncommon to use a display with LEDs such as this.

**Color ordering**

This is short and sweet: for the LEDs on the Funhouse board, I use the `BGR` order with `fastled_spi` but I have to use `RGB` with `esp32_spi_led_strip `. Not a big deal, but some folks might complain/be confused/think something is wrong if they switch from that to this.

**Arduino**

When built with the Arduino framework and including the changes I have in [this PR](https://github.com/esphome/esphome/pull/5116), I am getting some *bizarre* behavior. It's an odd combination of things that I haven't dug in to pin down just yet, but, to name a few:

- Wi-fi sometimes won't connect -- it scans, finds the network, but connecting continuously fails; eventually it falls back to starting the fallback AP but this also fails, logging errors about failing to allocate buffer memory (???)
- When it *does* connect to wi-fi, it logs a connection from HA but either:
  - HA indicates all of its entities are unavailable
  - Everything operates normally for a minute or two but then the connection stalls and drops until the ESP is reset
- It'll crash randomly, although this is infrequent

Casual observation: when running an effect (such as the built-in Twinkle), the effect freezes at the same time the HA connection stalls/drops (as described above). It's not completely freezing/crashing, though, as it continues to log readings from the other sensors on the board (seen via serial logs). There's something really weird going on there...

I'm completely at a loss to explain what's going on -- I've never seen behavior this erratic. 😅 I probably need more time to sort through it and to try to pin down (a) specific cause(s).

**ESP-IDF**

When built with ESP-IDF, the SPI display issue aside, everything seems to work just fine. 🎉

Ok, I think that's everything for now. Thanks for reading all this. If I work out anything else, I'll report back...

dentra:
@kbx81 thank you for such a detailed review and testing. 
I will try to look deeper to use SPIDevice. Changing color ordering is very simple, I will do that.

dentra:
SPIDevice does not use hardware capabilities for ESP-IDF, but does it only for Arduino. Thus, there may be a collision of using the same spi_bus_id when compiling in the Arduino environment. In this component is is hardcoded to SPI2_HOST (looking for a solution).

kbx81:
> SPIDevice does not use hardware capabilities for ESP-IDF, but does it only for Arduino. Thus, there may be a collision of using the same spi_bus_id when compiling in the Arduino environment. In this component is is hardcoded to SPI2_HOST (looking for a solution).

Jesse and I were discussing this yesterday. I'm going to do some work to move IDF onto the SPI hardware. I had no idea it was bit-banging when I made this suggestion. Will fix soon.

kbx81:
Thank you for working on this! Just to let you know, the SPI component was refactored in #5311 and the author wrote #5288 in parallel; as such, it's fully updated and using `SPIDevice`, pulling in hardware full support when/where it's able & configured. As such, we're going to move forward with #5288. Regardless, thank you for your contribution.

PR #5104, opened 2023-07-16
Add timeout filter

clydebarrow:

Add a new filter to publish `NaN` when no data is provided within a specified time window.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3068

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
sensor:
  - platform: vbus
    model: deltasol_bs_2009
    temperature_1:
      name: Collector Temp
      filters:
        timeout: 60s
```

## Checklist:

### DIFF
--- a/esphome/components/sensor/__init__.py
+++ b/esphome/components/sensor/__init__.py
@@ -217,6 +217,7 @@ OffsetFilter = sensor_ns.class_("OffsetFilter", Filter)
 MultiplyFilter = sensor_ns.class_("MultiplyFilter", Filter)
 FilterOutValueFilter = sensor_ns.class_("FilterOutValueFilter", Filter)
 ThrottleFilter = sensor_ns.class_("ThrottleFilter", Filter)
+TimeoutFilter = sensor_ns.class_("TimeoutFilter", Filter, cg.Component)
 DebounceFilter = sensor_ns.class_("DebounceFilter", Filter, cg.Component)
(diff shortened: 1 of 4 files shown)


nagyrobi:
This is very welcome. It's useful for ModBUS stuff too.

I wonder would it worth to implement an option to give a specific lambda value to fall back to if required, not only NAN.

For example, i use a wireless and a wired temp sensor measuring the same room, as an input for a thermostat. The wireless sensor has higher precision and can be placed more optimally, and uses a battery. The wired sensor has a lower precision and could only be placed in a worse location.

This filter would allow to set up a very convenient priority system, which in case of timeout from the wireless sensor (due to interference or battery drain or whatever), could fall back to the wired sensor or any other value (a number etc) shown in a lambda.

clydebarrow:
> would it worth to implement an option to give a specific lambda value to fall back to if required, not only NAN.

No, because now this filter is proposed, you can then use a `lambda` filter after the `timeout` to replace the NaN with another value, e.g. the value of the other sensor (accessed via `id(sensor_id).state`.) You can't do that right now because without the `timeout` filter there is no way of triggering the filter chain in the absence of data.

Or to rephrase that, already done!

jesserockz (approved):

PR #5105, opened 2023-07-16
Coolix IR protocol improvements

dudanov:

The `coolix` protocol in many air conditioners _does not follow strict rules_, resulting in the `climate_ir` `coolix` component not being further developed to provide advanced control and feedback capabilities. **This PR marks the beginning of this work.**

For example, in my air conditioner, the `SLEEP` command is sent in two different packets: the first is the `SLEEP` command, and the second sends the standard state. Before this changes, the decoder ignored the data, and now it outputs:

`[remote.coolix:093]: Received unstrict Coolix: [0xB2E003, 0xB21F98]`

Some commands consist even of one packet, for example, `SWING_STEP`. Now the decoder processes it correctly and outputs:

`[remote.coolix:095]: Received unstrict Coolix: [0xB20FE0]`

As for the breakdown of automation compatibility, I can assume that it is used by an extremely small number of users, since the protocol is specific and is used mainly as part of the `climate_ir` `coolix` component. Actually, not so long ago I removed it from the component to `remote_base` _without any issue_.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3071

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
binary_sensor:
 - platform: remote_receiver
    coolix: 0xB21F98 # in first or second packet
 - platform: remote_receiver
    coolix:
      first: 0xB2E003
      second: 0xB21F98 # optional. if none - event generated only on one packet commands

```

## Checklist:

### DIFF
--- a/esphome/components/remote_base/__init__.py
+++ b/esphome/components/remote_base/__init__.py
@@ -358,7 +358,7 @@ COOLIX_BASE_SCHEMA = cv.Schema(
         cv.Required(CONF_FIRST): cv.hex_int_range(0, 16777215),
         cv.Optional(CONF_SECOND, default=0): cv.hex_int_range(0, 16777215),
         cv.Optional(CONF_DATA): cv.invalid(
-            "'data' option has been removed in ESPHome 2023.7. "
+            "'data' option has been removed in ESPHome 2023.8. "
             "Use the 'first' and 'second' options instead."
         ),
     }
--- a/esphome/components/remote_base/__init__.py
+++ b/esphome/components/remote_base/__init__.py
@@ -355,8 +355,8 @@ async def canalsatld_action(var, config, args):
 
 COOLIX_BASE_SCHEMA = cv.Schema(
     {
-        cv.Required(CONF_FIRST): cv.hex_uint32_t,
-        cv.Optional(CONF_SECOND, default=0): cv.hex_uint32_t,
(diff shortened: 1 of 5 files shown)


esphome/components/remote_base/__init__.py
```diff
@@ -349,19 +351,48 @@ async def canalsatld_action(var, config, args):
...
-COOLIX_SCHEMA = cv.Schema({cv.Required(CONF_DATA): cv.hex_uint32_t})
 
 
-@register_binary_sensor("coolix", CoolixBinarySensor, COOLIX_SCHEMA)
+COOLIX_BASE_SCHEMA = cv.Schema(
+    {
+        cv.Required(CONF_FIRST): cv.hex_int_range(0, 16777215),
+        cv.Optional(CONF_SECOND, default=0): cv.hex_int_range(0, 16777215),
+        cv.Optional(CONF_DATA): cv.invalid(
+            "'data' option has been removed in ESPHome 2023.7. "
```
glmnet: .8

glmnet (dismissed):
Looks good and works ok. thank you

esphome/components/remote_base/__init__.py
dudanov: Corrected. But now a new review is required @glmnet .

PR #5108, opened 2023-07-16
Tuya Number Scaling

### LINKED ISSUE #4309: Scaling for Tuya number component
### The problem

The device sends values that are scaled by 10. I want to write these values to a number (to display in the frontend and change) and then send them back using the same number. The number component doesn't allow scaling like the sensor component.

2023.2.0

Home Assistant Add-on

_No response_

ESP8266

### Board

_No response_

Tuya number

### Example YAML snippet

_No response_

_No response_

### Additional information

There is the option to write the number to a global variable and display it as a sensor template, but this creates two entities of the same number. If the number component would have a scaling option, it would make the code much cleaner and easier to use.
This, in my opinion is a very common feature for others that are using Tuya MCU or something similar where values are multiplied to only use integers instead of float.

InvncibiltyCloak:

Uses the Tuya Number component's step as the scaling from ESPHome float to Tuya integer.

In the Tuya ecosystem, all numeric datapoints are integers. Therefore, if the value needs to include floating point precision, the value has a scaling factor applied. For example, a pH reading from 0.00 to 15.00 is reported as integers from 0 to 1500 with a scaling of 0.01. 

Currently in ESPHome the Tuya Number component has no support for scaling. However, this concept is effectively already built into the Number component with the Step parameter capturing the intent. 

This PR changes the behavior of Tuya Number to use the Step parameter as the scaling for the displayed number. 

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/issues#4309

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3075

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
tuya:
  id: tuya_mcu

number:
  - platform: tuya
    name: "pH Low Limit"
    number_datapoint: 108
    min_value: 0
    max_value: 15
    step: 0.01
```

## Checklist:


InvncibiltyCloak:
Not sure if this one would be considered a breaking change.
It definitely affects existing config files if people were using `step` values other than 1.

But considering the behavior was mostly unusable with `step < 1` (decimal places get truncated to integers for Tuya), it seems unlikely anyone would care. I would try to class it as a "New Feature".

prigorus:
Is there any chance to push this feature? I am waiting some time for this and would really help me.

ssieb (approved):

PR #5110, opened 2023-07-17
Dashboard: use Popen() on Windows

kuba2k2:

The PR adds a Popen()-based process spawning method to the Dashboard code, so that the dashboard can be tested on Windows. Nothing changes for Linux/Docker users, it only adds Windows compatibility (which would raise a Python exception prior to that).

Windows doesn't support non-blocking pipes; use Popen() with a reading thread instead.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -25,6 +25,7 @@ import tornado.ioloop
 import tornado.iostream
 import tornado.netutil
 import tornado.process
+import tornado.queues
 import tornado.web
 import tornado.websocket
 import yaml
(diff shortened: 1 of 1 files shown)


jesserockz:
Just gave this a quick test on windows and it prints an exit code of 0 way before the process is finished
image

kuba2k2:
That could mean that readline() can return an empty line even if the process is running. I'll give it a fix. On my machine though, there was no problems whatsoever.

kuba2k2:
That should fix it, I've added a check to make sure the process has terminated, before closing the thread.

jesserockz (approved):
Thanks!

PR #5111, opened 2023-07-17
Init colorama in ESPHome main

kuba2k2:

Call `colorama.init()` in ESPHome CLI tools, to fix terminal color support on Windows (especially in editors such as VSCode, having their own terminal support).

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/__main__.py
+++ b/esphome/__main__.py
@@ -670,7 +670,6 @@ POST_CONFIG_ACTIONS = {
 
 
 def parse_args(argv):
-    colorama.init()
     options_parser = argparse.ArgumentParser(add_help=False)
     options_parser.add_argument(
         "-v", "--verbose", help="Enable verbose ESPHome logs.", action="store_true"
@@ -954,6 +953,8 @@ def parse_args(argv):
 
 
 def run_esphome(argv):
+    colorama.init()
+
     args = parse_args(argv)
     CORE.dashboard = args.dashboard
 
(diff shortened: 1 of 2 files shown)


jesserockz:
Which part exactly is this fixing?
Do you have a before and after screenshot?

kuba2k2:
This fixes colored text output, where otherwise ANSI color codes are shown instead.

Take a look (before/after):
obraz

(the 2nd one errors out, because it doesn't have the Popen() PR checked out).

Nevertheless, it's always good to init Colorama whenever working with colored text output.

jesserockz:
Weird, it was colouring for me when I was testing the popen PR just fine on windows.

Since it is going to be init'd here, can you remove the init from the log file?

kuba2k2:
Sure, I didn't really notice it was in `log.py` already. It seems that the whole issue was that `init()` was placed *after* the `logging` calls - placing it before them fixes coloring.

Anyway, I removed it from `log.py`, and moved to `run_esphome()` since it seems more appropriate.

jesserockz:
If moving it before the `logging` calls fixes the issue on windows, I would rather keep in inside the log.py file in that case...
(sorry to go around in circles)

kuba2k2:
Done.

jesserockz (approved):

jesserockz:
Thanks!

PR #5133, opened 2023-07-20
relative path fix got ingress

### LINKED ISSUE #4709: Add-on options relative_url & streamer_mode dont work
### The problem

The add-on configuration options `relative_url` & `streamer_mode` don't work
Adding a `relative_url` the dashboard gets a `404: Not Found` page
Turning on `streamer_mode` also doesn't appear to change anything the SSID is still logged when going to the device logs

2023.6

Home Assistant Add-on

2023.6

ESP8266

### Board

_No response_

_No response_

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

grahambrown11:
If the Relative Path setting in the add-on is set the ingress from HA does not work, so when HA ingress hits the root path this will redirect to the path in the settings.
Also ensured the path starts and ends with a slash otherwise it's prefix and not a path then all the APIs from the frontend break...

## Types of changes

- [ ] Other

fixes https://github.com/esphome/issues/issues/4709

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:


grahambrown11:
Closing, will open an new PR to remove the setting from the add-on config as per the comment in HA - https://github.com/home-assistant/core/pull/97002#issuecomment-1646126090

grahambrown11:
Created https://github.com/esphome/home-assistant-addon/pull/99 to remove the setting

PR #5137, opened 2023-07-21
pca9554 cache reads

hwstar:

Performance improvement. This update changes the way digital_read() operates. Instead of doing an I2C transaction for each call to digital read, it will now just fetch a recent copy of the last I2C transaction. This recent copy is updated approximately every 10mS in the loop() function.

## Types of changes

- [X] Other

## Test Environment

- [X ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:
  - [X ] The code change is tested and works locally.


esphome/components/pca9554/pca9554.cpp
```diff
@@ -26,14 +26,25 @@ void PCA9554Component::setup() {
...
 }
+
+void PCA9554Component::loop() {
+  // Called approximately 10 ms on average (100 Hz). Note: during power up
+  // initialization it will be called at a higher rate.
+  // The read_inputs_() method will cache the input values from the chip.
+  this->read_inputs_();
+  // Clear all the previously read flags.
+  for (int i = 0; i < 8; i++)
+    this->was_previously_read_[i] = false;
```
jesserockz: This should be a little bit better
```suggestion
  std::fill(this->was_previously_read_.begin(), this->was_previously_read_.end(), false);
```

Also add `#include <algorithm>` at the top

hwstar:
Changes made.

hwstar:
Changed from Bit vector to uint8_t per ssieb's request.

ssieb (approved):

PR #5139, opened 2023-07-22
image: allow the image to by auto-loaded by animation

ayufan:
This fixes the case where only `animation:` is present and `image:` is auto-loaded.

This introduces `MULTI_CONF_NO_DEFAULT=True` that allows to include image source code only without initialising a default object.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

Per `test8.yaml` included in PR.

## Checklist:

### DIFF
--- a/esphome/__main__.py
+++ b/esphome/__main__.py
@@ -12,7 +12,7 @@ import argcomplete
 
 from esphome import const, writer, yaml_util
 import esphome.codegen as cg
-from esphome.config import iter_components, read_config, strip_default_ids
+from esphome.config import iter_component_configs, read_config, strip_default_ids
 from esphome.const import (
     ALLOWED_NAME_CHARS,
     CONF_BAUD_RATE,
@@ -196,7 +196,7 @@ def write_cpp(config):
 def generate_cpp_contents(config):
     _LOGGER.info("Generating C++ source...")
 
-    for name, component, conf in iter_components(CORE.config):
+    for name, component, conf in iter_component_configs(CORE.config):
         if component.to_code is not None:
             coro = wrap_to_code(name, component)
             CORE.add_job(coro, conf)
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -36,6 +36,7 @@ _LOGGER = logging.getLogger(__name__)
 DOMAIN = "image"
 DEPENDENCIES = ["display"]
 MULTI_CONF = True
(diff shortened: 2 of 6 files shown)


nielsnl68:
@ayufan , although this fix is mostly for the `animation` issue.
The fix itself is a more common one. May i suggest to change the title more to that change then to the animation issue.

geiseri:
@ayufan Wouldn't this better be addressed by the image component directly?  I am not sure what other components it could impact to warrant this being more general.  This is a special case where the image functionality can be used without an actual component.  I could argue it is "being used wrong" as other animation implementations use sprites.  It would be a single image with all of the frames together and the memory is blit from each frame offset.  Technically I think that is what it is doing under the hood anyway.  The only other place I could see this is `color`.  You might want to subclass that one to provide a dynamic behavior similar to animation. 

@nielsnl68 I also have this issue in a custom component that is a factory that constructs images.  I moved it subclass image, but remained with the same issue.  Would it be better to have it named `SUBCLASS_COMPONENT` or `IMPLEMENTS_COMPONENT`?

guillempages:
@nielsnl68 it is not only animation; IIRC there was someone on discord complaining about this issue for a font, and my online_image implementation also has the same issue.

@ayufan as niels mentioned, it might be a good idea to handle this generically; not animation specific. I created a PR a while ago to add the test that you are adding for the animation; we might want to split that, add the tests for animation and font in my PR (#5057) and keep the fix here.

guillempages:
@ayufan you might want to add "Fixes https://github.com/esphome/issues/issues/4719, https://github.com/esphome/issues/issues/4726" to the description as well (and there might be other duplicate issues as well)

jesserockz:
I started to look into this again and remembered this PR and I think its actually the best way forward =)

jesserockz (approved):

jesserockz:
Tested working locally and also with https://github.com/esphome/esphome/pull/5218 (after adding the new flag to `uart`)

jesserockz (changes requested):
Rebase required

(I'm just using this PR to test a new bot on the ESPHome github repo)

ayufan:
@jesserockz This is rebased.

ayufan:
@jesserockz This is rebased.

jesserockz:
Thanks @ayufan

jesserockz (approved):

PR #5142, opened 2023-07-22
fix to PR # 3887 MQTT connection not using discovery: false nor discovery_prefix

luka6000:
fixing MQTT connection not using _discovery: false_ nor _discovery_prefix: topic_prefix_

## Types of changes

- [ ] Other

https://github.com/esphome/issues/issues/4703

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
mqtt:
  broker: 192.168.0.99
  username: devel
  password: anypass
  discovery: false
  discovery_prefix: 'io_username/feeds'
  birth_message:
  will_message:
  log_topic:
```

## Checklist:

### DIFF
--- a/esphome/components/haier/climate.py
+++ b/esphome/components/haier/climate.py
@@ -89,7 +89,7 @@ SUPPORTED_SWING_MODES_OPTIONS = {
 
 SUPPORTED_CLIMATE_MODES_OPTIONS = {
     "OFF": ClimateMode.CLIMATE_MODE_OFF,  # always available
-    "AUTO": ClimateMode.CLIMATE_MODE_AUTO,  # always available
+    "HEAT_COOL": ClimateMode.CLIMATE_MODE_HEAT_COOL,  # always available
     "COOL": ClimateMode.CLIMATE_MODE_COOL,
     "HEAT": ClimateMode.CLIMATE_MODE_HEAT,
     "DRY": ClimateMode.CLIMATE_MODE_DRY,
--- a/esphome/components/haier/haier_base.cpp
+++ b/esphome/components/haier/haier_base.cpp
@@ -72,7 +72,7 @@ HaierClimateBase::HaierClimateBase()
   this->traits_ = climate::ClimateTraits();
   this->traits_.set_supported_modes({climate::CLIMATE_MODE_OFF, climate::CLIMATE_MODE_COOL, climate::CLIMATE_MODE_HEAT,
                                      climate::CLIMATE_MODE_FAN_ONLY, climate::CLIMATE_MODE_DRY,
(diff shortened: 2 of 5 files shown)


esphome/mqtt.py
```diff
@@ -124,8 +124,13 @@ def on_disconnect(client, userdata, result_code):
...
 def show_discover(config, username=None, password=None, client_id=None):
-    topic = "esphome/discover/#"
-    _LOGGER.info("Starting log output from %s", topic)
+    if CONF_DISCOVERY_PREFIX in config[CONF_MQTT]:
+        topic = f"{config[CONF_DISCOVERY_PREFIX]}/esphome/discover"
+        topics = f"{config[CONF_DISCOVERY_PREFIX]}/esphome/discover/#"
+    else:
+        topic = "esphome/discover"
+        topics = "esphome/discover/#"
+    _LOGGER.info("Starting log output from %s", topics)
```
jesserockz: The `CONF_DISCOVERY_PREFIX` has a default value of `homeassistant` so it is always in the config and cannot be removed.

luka6000: line 245 says something different
discovery_prefix = config[CONF_MQTT].get(CONF_DISCOVERY_PREFIX, "homeassistant")

Maybe you're talking about discovery_prefix variable and it's default?

jesserockz: https://github.com/esphome/esphome/blob/dev/esphome/components/mqtt/__init__.py#L191-L193

luka6000: it works just as expected with _discovery: false_ but I get your point. So what do you suggest? Change this to CONF_TOPIC_PREFIX and CONF_DISCOVERY or hardcode _homeassistant/esphome/..._?

jesserockz: My point here is that with this new code, all the topics are going to change to `homeassistant/esphome/discover` etc which is undesired as technically this discover information has nothing to do with homeassistant.

luka6000:
@jesserockz please review again. I've used CONF_TOPIC_PREFIX instead of CONF_DISCOVERY_PREFIX. And the discovery flag is still respected.

jesserockz:
`CONF_TOPIC_PREFIX` has the same issue that it has a default value of the value from `esphome` -> `name` and will not be empty unless people explicitly add `topic_prefix: ""`, but even then the code checks `if CONF_TOPIC_PREFIX in config` which is always going to be true.

I think for this PR, just make the changes related to disabling the pub/sub if `discovery: false` and make a separate PR or come chat on discord about a better solution if we want to prefix the topic for this part.

PR #5143, opened 2023-07-23
Add BMI160 support

flaviut:

Adds support for [the Bosch BMI160 IMU](https://www.bosch-sensortec.com/products/motion-sensors/imus/bmi160/)

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/3086

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
sensor:
  - platform: bmi160
    address: 0x68
    update_interval: 60s
    acceleration_x:
      name: "BMI160 Accel X"
    acceleration_y:
      name: "BMI160 Accel Y"
    acceleration_z:
      name: "BMI160 Accel z"
    gyroscope_x:
      name: "BMI160 Gyro X"
    gyroscope_y:
      name: "BMI160 Gyro Y"
    gyroscope_z:
      name: "BMI160 Gyro z"
    temperature:
      name: "BMI160 Temperature"
```

## Checklist:


esphome/components/bmi160/bmi160.cpp
```diff
@@ -0,0 +1,244 @@
...
+  // need to wait (max delay in datasheet) because we can't send commands while another is in progress
+  delay(5);
+
+  ESP_LOGV(TAG, "  Bringing gyroscope out of sleep...");
+  if (!this->write_byte(BMI160_REGISTER_CMD, (uint8_t) Cmd::GYRO_SET_PMU_MODE | (uint8_t) GyroPmuMode::NORMAL)) {
+    this->mark_failed();
+    return;
+  }
+  ESP_LOGV(TAG, "  Waiting for gyroscope to wake up...");
+  delay(81);
```
flaviut: Unfortunately I don't see a better way to do this--I could do a loop checking the I2C register if it's done yet, but that's a lot of work to maybe save 30ms (typical duration of this step is 50ms, max is 81ms)

clydebarrow: You could extract the succeeding code into another function and use `set_timeout()` to call it later.

You'd need a flag for setup complete which you would test in `update()`.

flaviut: Good call! Thank you and done.

esphome/components/bmi160/bmi160.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    case 1:
+      ESP_LOGV(TAG, "  Bringing gyroscope out of sleep...");
+      if (!this->write_byte(BMI160_REGISTER_CMD, (uint8_t) Cmd::GYRO_SET_PMU_MODE | (uint8_t) GyroPmuMode::NORMAL)) {
+        this->mark_failed();
+        return;
+      }
+      ESP_LOGV(TAG, "  Waiting for gyroscope to wake up...");
+      // wait between 51 & 81ms, doing 81 to be safe
+      this->set_timeout(81, [this]() { this->internal_setup_(2); });
```
clydebarrow: If 81ms is the maximum per the datasheet, I'd allow a little extra in case the timeout call happens early. There is no guarantee on the accuracy of the timeout, and while I'm fairly certain it will not happen early, an extra few ms delay will do no harm. I'd round it up to 100ms.

esphome/components/bmi160/bmi160.cpp
```diff
@@ -0,0 +1,260 @@
...
+  float accel_x = (float) data[3] / (float) INT16_MAX * 16 * GRAVITY_EARTH;
+  float accel_y = (float) data[4] / (float) INT16_MAX * 16 * GRAVITY_EARTH;
+  float accel_z = (float) data[5] / (float) INT16_MAX * 16 * GRAVITY_EARTH;
+
+  uint8_t raw_temperature;
+  if (this->read_register(BMI160_REGISTER_DATA_GYRO_X_LSB, &raw_temperature, 2, true) != i2c::ERROR_OK) {
+    this->status_set_warning();
+    return;
+  }
+  int16_t int_temperature = *reinterpret_cast<int16_t *>(&raw_temperature);
```
clydebarrow: This relies on the executing architecture being little-endian. Better to explicitly shift and add the bytes, that way it's portable across architecures.

esphome/components/bmi160/sensor.py
```diff
@@ -0,0 +1,84 @@
...
+    STATE_CLASS_MEASUREMENT,
+    UNIT_METER_PER_SECOND_SQUARED,
+    ICON_SCREEN_ROTATION,
+    UNIT_DEGREE_PER_SECOND,
+    UNIT_CELSIUS,
+)
+
+DEPENDENCIES = ["i2c"]
+
+CONF_ACCEL_X = "accel_x"
```
clydebarrow: There are already similar config constants defined in esphome.const:
```
CONF_ACCELERATION_X = "acceleration_x"
CONF_ACCELERATION_Y = "acceleration_y"
CONF_ACCELERATION_Z = "acceleration_z"

```
I know that other gyro components define their own, but if they're in the system files I would think it better to use those. There are currently no GYRO versions in const.py - maybe you could add those as part of the PR.

esphome/components/bmi160/sensor.py
```diff
@@ -0,0 +1,84 @@
...
+CONF_GYRO_Z = "gyro_z"
+
+bmi160_ns = cg.esphome_ns.namespace("bmi160")
+BMI160Component = bmi160_ns.class_(
+    "BMI160Component", cg.PollingComponent, i2c.I2CDevice
+)
+
+accel_schema = sensor.sensor_schema(
+    unit_of_measurement=UNIT_METER_PER_SECOND_SQUARED,
+    icon=ICON_BRIEFCASE_DOWNLOAD,
```
clydebarrow: There are icons defined in const.py for acceleration in three axes.

```
ICON_ACCELERATION_X = "mdi:axis-x-arrow"
ICON_ACCELERATION_Y = "mdi:axis-y-arrow"
ICON_ACCELERATION_Z = "mdi:axis-z-arrow"
```

Or just `ICON_ACCELERATION` if you don't care about differentiating them.

clydebarrow (changes requested):

flaviut:
Good call on the icons & configs! Fixed & tested everything.

flaviut:
Hi I want to check in on this. I understand if it won't be merged, but I would like to at least have the icon changes merged so I can keep this custom integration in an upstreamable state.

kbx81 (approved):

PR #5145, opened 2023-07-24
Allow BTM and RRM to be explicitly disabled on Arduino framework

### LINKED ISSUE #4725: Can't explicitly disable wifi BTM or RRM
### The problem

`enable_btm` and `enable_rrm` can only be set to `true` for ESP32 devices using the `esp-idf` framework, but setting these to `false` on an ESP8266 isn't valid, instead the settings have to be omitted.

2023.7.0

Home Assistant Add-on

_No response_

ESP8266

### Board

_No response_

Wifi

### Example YAML snippet

```yaml
---
substitutions:
  enable_btm: 'false'
  enable_rrm: 'false'

esphome:
  name: 'test'

esp8266:
  board: 'esp01_1m'

wifi:
  enable_btm: '${enable_btm}'
  enable_rrm: '${enable_rrm}'
```

```txt
Failed config

wifi: [source test.yaml:13]
  
  This feature is only available with frameworks ['esp-idf'].
  enable_btm: false
  
  This feature is only available with frameworks ['esp-idf'].
  enable_rrm: false
```

### Additional information

_No response_

joshuaspence:

Currently the following configuration fails when using the Arduino framework:

```yaml
---
substitutions:
  enable_btm: 'false'
  enable_rrm: 'false'

esphome:
  name: 'test'

esp8266:
  board: 'esp01_1m'

wifi:
  enable_btm: '${enable_btm}'
  enable_rrm: '${enable_rrm}'
```

The intention of the code is to only allow `enable_btm` and `enable_rrm` to be set to `true` when using ESP-IDF, but this isn't consistent with the actual behaviour of the code.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** Fixes esphome/issues#4725.

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
---
substitutions:
  enable_btm: 'false'
  enable_rrm: 'false'

esphome:
  name: 'test'

esp8266:
  board: 'esp01_1m'

wifi:
  enable_btm: '${enable_btm}'
  enable_rrm: '${enable_rrm}'
```

## Checklist:

### DIFF
--- a/esphome/components/wifi/__init__.py
+++ b/esphome/components/wifi/__init__.py
@@ -78,6 +78,12 @@ def validate_channel(value):
     return value
 
 
+def validate_esp_idf_feature(value):
+    if value:
+        return cv.only_with_esp_idf(value)
+    return value
+
+
 AP_MANUAL_IP_SCHEMA = cv.Schema(
     {
         cv.Required(CONF_STATIC_IP): cv.ipv4,
@@ -280,10 +286,10 @@ CONFIG_SCHEMA = cv.All(
                 cv.decibel, cv.float_range(min=8.5, max=20.5)
             ),
             cv.SplitDefault(CONF_ENABLE_BTM, esp32_idf=False): cv.All(
-                cv.boolean, cv.only_with_esp_idf
+                cv.boolean, validate_esp_idf_feature
             ),
             cv.SplitDefault(CONF_ENABLE_RRM, esp32_idf=False): cv.All(
-                cv.boolean, cv.only_with_esp_idf
+                cv.boolean, validate_esp_idf_feature
             ),
             cv.Optional(CONF_PASSIVE_SCAN, default=False): cv.boolean,
             cv.Optional("enable_mdns"): cv.invalid(


jesserockz:
These config keys are invalid to be set to any value (including false) on anything other than esp32 using esp-idf.

There is no reason to even include them in a config for an esp8266, even to set to false as there is not code to enable or disable these features for esp8266, rp2040 or esp32 arduino

joshuaspence:
The reason I want to do things this way is that I use a package for configuring wifi and a separate package for configuring the platform. Allowing setting these fields to false means that I can set them to false from within my ESP32 package. Otherwise I would need to create separate packages for:

- ESP32 with wifi
- ESP32 without wifi (ethernet devices)
- ESP8266

Although not a huge inconvenience, I felt that the current implementation was likely not intentional/expected.

This is possibly related to https://github.com/esphome/feature-requests/issues/2163, although that issue is focussed on lists rather than dicts.

jesserockz:
I had a feeling you might say packages...haha

No, the current implementation is 100% intentional and is used is multiple places in the codebase to add config variables only for certain platforms.

joshuaspence:
Alright then, thanks for the explanation.

PR #5149, opened 2023-07-24
Max31850

lylavoie:

Implements support for MAX31850 thermocouple, based on Dallas sensor component.  Deals with the unit conversions and error handling provided within the MAX31850 chip.

## Types of changes

- [ X ] New feature (non-breaking change which adds functionality)
- [ ] Other

## Test Environment

- [ X ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
sensor:
  - platform: dallas
    address: 0x973c01d075bd6028
    name: Boiler Water Supply Temp
  - platform: dallas
    address: 0x082d4c0f49858c3b
    name: Boiler Exhaust Temp
    chipset: max31850
```

## Checklist:
  - [ X ] The code change is tested and works locally.

  - [ X ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).


lylavoie:
Checking if any further review / changes are needed? 
Similarly the documentation update has also been submitted: esphome/esphome-docs#3089

esphome/components/dallas/sensor.py
```diff
@@ -27,6 +39,9 @@
             cv.Optional(CONF_ADDRESS): cv.hex_uint64_t,
             cv.Optional(CONF_INDEX): cv.positive_int,
             cv.Optional(CONF_RESOLUTION, default=12): cv.int_range(min=9, max=12),
```
jesserockz, line 41: If the resolution cannot be changed on certain chipsets, then config validations should exist to block it being set.

jesserockz (changes requested):

randybb:
Is the chipset definition needed? Because 1wire 64-bit ID contains a 8-bit family code, so it can be always autodiscovered https://www.analog.com/en/technical-articles/1wire-software-resource-guide-device-description.html (Maxim 1-Wire device types are missing in this list).

esphome/components/dallas/sensor.py
lylavoie: This is not completely possible, because of the "auto" chipset, where the chipset isn't known until the i2c connection is made and the device ID is read.  We could validate the resolution / chipset, if it's not set to auto.  Do you have a preference?

lylavoie:
@randybb If I am ready the want the sensors are reporting correctly, it seems like the MAX31850 overload the 0x3B value.  The output below is from scanning the bus on my setup, and the 0x3248000ec426833b sensor is my MAX31850, while the other is a standard DS18B20 temp probe.  Unfortunately, the overloaded the value, bit bit stuffing for the two sensors is different.  That was the original reason I started this work.

[18:35:32][D][dallas.sensor:085]:   Found sensors:
[18:35:32][D][dallas.sensor:087]:     0x973c01d075bd6028
[18:35:32][D][dallas.sensor:087]:     0x3248000ec426833b

The other option would be to break the MAX31850 to a separate component, but I don't think that configuration would support sharing the same bus/wire with the other Dallas sensors.

randybb:
The dallas component is really old and it was never meant to be universal, so it is supporting only DS18B20 sensors. Probably a major rewrite would be needed to support various types of sensors/IO/bus bridges (at least the scanner could show the device family). Let's wait for the devs what they think about that.

lylavoie:
We could refactor the dallas component into a 1-wire base component and then sensor components, similar to the structure of SPI and other buses.  That would allow the different chipsets / sensors to become independent, at the cost of some duplicated code (between different sensors) and the bus component.

randybb:
And ideally when it could be used for i2c 1-wire masters like is mentioned here https://github.com/esphome/feature-requests/issues/621

PR #5161, opened 2023-07-27
Force heater off on setup

rufuswilson:

forces the heater to be turned off on setup
on my custom built PCB using the sensirion chips, the heater was enabled by default

## Types of changes

- [ ] Other

https://github.com/esphome/issues/issues/2887

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3149

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (commented):
How would someone turn it back on?

rufuswilson:
This is a good question. From the different discussions, I understood it should have been turned off by default.


I was going to use the following as a temporary workaround to turn it off (I haven't tested it yet). Something similar but with the `0x306D` command should turn it back on again. What do you think?
```
esphome:
  on_boot:
    - priority: 100
      then:
      - lambda: |-
      - lambda: |-
          uint16_t SHT3XD_COMMAND_HEATER_ENABLE = 0x306D;
          uint16_t SHT3XD_COMMAND_HEATER_DISABLE = 0x3066;
          id(sht31d)->write_command(SHT3XD_COMMAND_HEATER_DISABLE);
```

jesserockz:
Well the best and most user friendly way would be to have:
```yaml

sensor:
  - platform: sht3xd
    heater_at_start: on/off
    
on_...:
  - sht3xd.heater_on:
  - sht3xd.heater_off:
```

rufuswilson:
That is a very good suggestion however it goes well beyond my current coding capabilities

jesserockz:
In that case, the actions we can leave for another time. But adding a config boolean for heater at startup should be done so a user can configure it properly and this does not become a breaking/unexpected change to all users already using it.

esphome/components/sht3xd/sensor.py
```diff
@@ -38,6 +38,11 @@
             ),
         }
     )
+    .extend(
+        {
+            cv.Required("heater_enabled", default=True): cv.boolean,
+        }
+    )
```
jesserockz: This can go directly inside the schema, and also needs to define `CONF_HEATER_ENABLED = "heater_enabled"` higher up in the file
```suggestion
        },
        cv.Optional(CONF_HEATER_ENABLED, default=True): cv.boolean,
    )
```

esphome/components/sht3xd/sensor.py
```diff
@@ -48,6 +53,8 @@ async def to_code(config):
     await cg.register_component(var, config)
     await i2c.register_i2c_device(var, config)
 
+    cg.add(var.set_heater_enabled(config["heater_enabled"]))
```
jesserockz:
```suggestion
    cg.add(var.set_heater_enabled(config[CONF_HEATER_ENABLED]))
```

jesserockz (changes requested):
c++ code looks good, just a few changes to the python side

jesserockz (approved):
Thanks for making the changes

rufuswilson:
Great! Just created a PR for the docs.
https://github.com/esphome/esphome-docs/pull/3149

jesserockz (approved):

PR #5174, opened 2023-08-01
feat: add AS5600 component/sensor

### LINKED ISSUE #1459: Support for AS5600 I2C Position Sensor

It would be fantastic to see the AS5600 module supported in ESPHome to allow single-chip support for all manner of 360° contactless rotational position sensing. 

 I wish to use this initially in a self-designed weather station for wind direction but expect to see it used for blinds, as a digital potentiometer equivalent for position inputs for dimmers and many similar use cases. The low cost adds to the attraction!

**Additional context**

More details at https://ams.com/as5600

ammmze:
The AS5600 is a magnetic rotary position encoder with a 12-bit resolution, allowing for 4096 positions/steps. There's been an open feature request for this for a couple years and a custom component implementation using a 3rd party library. This component uses the esphome native i2c implementation, not the 3rd party library.

resolves: esphome/feature-requests#1459

The AS5600 is a magnetic rotary position encoder with a 12-bit resolution, allowing for 4096 positions/steps. There's been an open feature request for this for a couple years and a custom component implementation using a 3rd party library. This component uses the esphome native i2c implementation, not the 3rd party library.

## Types of changes

**Related issue or feature (if applicable):** resolves esphome/feature-requests#1459

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3304

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
i2c:
  sda: 25
  scl: 21
  scan: true
  id: bus_a

as5600:
    id: my_as5600
    dir_pin: 22
    start_position: -90deg
    range: 180deg

sensor:
  - platform: as5600
    update_interval: 1s
    name: Percent
    unit_of_measurement: '%'
    accuracy_decimals: 2
    icon: mdi:percent
    filters:
      - delta: 1
      - lambda: 'return (x / as5600::POSITION_COUNT) * 100;'
  - platform: as5600
    update_interval: 1s
    name: Angle
    unit_of_measurement: °
    accuracy_decimals: 2
    icon: mdi:rotate-right
    out_of_range_mode: nan
    filters:
      - delta: 1
      - lambda: 'return x * as5600::RAW_TO_DEGREES * id(my_as5600).get_range_scale();'
  - platform: as5600
    update_interval: 1s
    name: Position
    filters:
      - delta: 1
    raw_position:
      name: Raw Position
      filters:
        - delta: 1
    gain:
      name: Gain
      filters:
        - delta: 1
    magnitude:
      name: Magnitude
      filters:
        - delta: 1
    status:
      name: Status
      filters:
        - delta: 1

```

## Checklist:

### DIFF
--- a/esphome/components/as5600/__init__.py
+++ b/esphome/components/as5600/__init__.py
@@ -1,0 +1,228 @@
+from esphome import pins
+import esphome.codegen as cg
+import esphome.config_validation as cv
+from esphome.components import i2c
+from esphome.const import (
+    CONF_ID,
+    CONF_DIR_PIN,
+    CONF_DIRECTION,
+    CONF_HYSTERESIS,
+    CONF_RANGE,
+)
+
+CODEOWNERS = ["@ammmze"]
+DEPENDENCIES = ["i2c"]
+MULTI_CONF = True
+
+as5600_ns = cg.esphome_ns.namespace("as5600")
+AS5600Component = as5600_ns.class_("AS5600Component", cg.Component, i2c.I2CDevice)
+
+DIRECTION = {
+    "CLOCKWISE": 0,
+    "COUNTERCLOCKWISE": 1,
+}
+
+POWER_MODE = {
+    "NOMINAL": 0,
+    "LOW1": 1,
+    "LOW2": 2,
+    "LOW3": 3,
+}
+
+HYSTERESIS = {
+    "NONE": 0,
+    "LSB1": 1,
+    "LSB2": 2,
+    "LSB3": 3,
+}
+
+SLOW_FILTER = {
+    "16X": 0,
+    "8X": 1,
+    "4X": 2,
+    "2X": 3,
+}
+
+FAST_FILTER = {
+    "NONE": 0,
+    "LSB6": 1,
+    "LSB7": 2,
+    "LSB9": 3,
+    "LSB18": 4,
+    "LSB21": 5,
+    "LSB24": 6,
+    "LSB10": 7,
+}
+
+CONF_ANGLE = "angle"
+CONF_RAW_ANGLE = "raw_angle"
+CONF_RAW_POSITION = "raw_position"
+CONF_WATCHDOG = "watchdog"
+CONF_POWER_MODE = "power_mode"
+CONF_SLOW_FILTER = "slow_filter"
+CONF_FAST_FILTER = "fast_filter"
+CONF_START_POSITION = "start_position"
+CONF_END_POSITION = "end_position"
+
+
+RESOLUTION = 4096
+MAX_POSITION = RESOLUTION - 1
+ANGLE_TO_POSITION = RESOLUTION / 360
+POSITION_TO_ANGLE = 360 / RESOLUTION
+# validate min range of 18deg (per datasheet) ... though i seem to get valid values down to a range of 192steps (16.875deg)
+MIN_RANGE = round(18 * ANGLE_TO_POSITION)
+
+
+def angle(min=-360, max=360):
+    return cv.All(
+        cv.float_with_unit("angle", "(°|deg)"), cv.float_range(min=min, max=max)
+    )
+
+
+def angle_to_position(value, min=-360, max=360):
+    try:
+        value = angle(min=min, max=max)(value)
+        return (RESOLUTION + round(value * ANGLE_TO_POSITION)) % RESOLUTION
+    except cv.Invalid as e:
+        raise cv.Invalid(f"When using angle, {e.error_message}")
+
+
+def percent_to_position(value):
+    value = cv.possibly_negative_percentage(value)
+    return (RESOLUTION + round(value * RESOLUTION)) % RESOLUTION
+
+
+def position(min=-MAX_POSITION, max=MAX_POSITION):
+    """Validate that the config option is a position.
+    Accepts integers, degrees, or percentage (of 360 degrees).
+    """
+
+    def validator(value):
+        if isinstance(value, str) and value.endswith("%"):
+            value = percent_to_position(value)
+
+        if isinstance(value, str) and (value.endswith("°") or value.endswith("deg")):
+            return angle_to_position(
+                value,
+                min=round(min * POSITION_TO_ANGLE),
+                max=round(max * POSITION_TO_ANGLE),
+            )
+
+        return cv.int_range(min=min, max=max)(value)
+
+    return validator
+
+
+def position_range():
+    """Validate that value given is a valid range for the device.
+    A valid range is one of the following:
+    - a value of 0 (meaning full range)
+    - 18 thru 360 degrees
+    - negative 360 thru negative 18 degrees (notes: these are normalized to their positive values, accepting negatives is for convenience)
+    """
+    zero_validator = position(min=0, max=0)
+    negative_validator = cv.Any(
+        position(min=-MAX_POSITION, max=-MIN_RANGE),
+        zero_validator,
+    )
+    positive_validator = cv.Any(
+        position(min=MIN_RANGE, max=MAX_POSITION),
+        zero_validator,
+    )
+
+    def validator(value):
+        is_negative_str = isinstance(value, str) and value.startswith("-")
+        is_negative_num = isinstance(value, (float, int)) and value < 0
+        if is_negative_str or is_negative_num:
+            return negative_validator(value)
+        return positive_validator(value)
+
+    return validator
+
+
+def has_valid_range_config():
+    """Validate that that the config start + end position results in a valid
+    positional range, which must be >= 18degrees
+    """
+    range_validator = position_range()
+
+    def validator(config):
+        # if we don't have an end position, then there is nothing to do
+        if CONF_END_POSITION not in config:
+            return config
+
+        # determine the range by taking the difference from the end and start
+        range = config[CONF_END_POSITION] - config[CONF_START_POSITION]
+
+        # but need to account for start position being greater than end position
+        # where the range rolls back around the 0 position
+        if config[CONF_END_POSITION] < config[CONF_START_POSITION]:
+            range = RESOLUTION + config[CONF_END_POSITION] - config[CONF_START_POSITION]
+
+        try:
+            range_validator(range)
+            return config
+        except cv.Invalid as e:
+            raise cv.Invalid(
+                f"The range between start and end position is invalid. It was was {range} but {e.error_message}"
+            )
+
+    return validator
+
+
+CONFIG_SCHEMA = cv.All(
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(AS5600Component),
+            cv.Optional(CONF_DIR_PIN): pins.gpio_input_pin_schema,
+            cv.Optional(CONF_DIRECTION, default="CLOCKWISE"): cv.enum(
+                DIRECTION, upper=True
+            ),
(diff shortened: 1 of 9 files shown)


fran6120:
Hi!

So far I was using this sensor in esphome by integrating as an external component from here: https://github.com/esphome/feature-requests/issues/1459

Now, seeing that you have written a custom library for esphome, I have been encouraged to try it.

The utility that I give is to obtain the direction in degrees of a 3D printed weather vane and for the moment it works great.

The only problem that I have found for my use, is that the lambda function that returns the value of the position in degrees, sends both the value "0 deg" and "360 deg". This problem already happened with the previous external component.

I don't think it's the most elegant way to do it, but it can be solved like this:

```
 - lambda: |-
       float deg = floor (x * as5600::RAW_TO_DEGREES * id(my_as5600).get_range_scale());
       if (deg > 359) {return deg - 360;}
       else {return deg;}
```

Perhaps it would be more convenient to have the option of obtaining the sensor value in degrees or raw, in the sensor configuration itself without having to use a lambda, but only as a suggestion, since it works great as it is.

Thanks you!

ammmze:
Thank you for your feedback! It is greatly appreciated!

> The only problem that I have found for my use, is that the lambda function that returns the value of the position in degrees, sends both the value "0 deg" and "360 deg".

That is interesting that you would get 360 deg because `x` should be a value equal to or between `0` and `4095`. So a max degree _should_ come through as `359.912109375` (eq `4095 * (360 / 4096)`). And then using `floor`, in theory that should come out to 359. I will look more into this.

Also for a slightly more elegant solution you could use the modulo operator. Replace the last 2 lines of your lambda with `return deg % 360;`.

Some examples of this:
```
  0 % 360 = 0
  1 % 360 = 1
180 % 360 = 180
359 % 360 = 359
360 % 360 = 0
361 % 360 = 1
```

> Perhaps it would be more convenient to have the option of obtaining the sensor value in degrees or raw, in the sensor configuration itself without having to use a lambda, but only as a suggestion, since it works great as it is.

That is something I am open to doing. I had thought about doing that but I wasn't sure if that is common in these esphome components. I haven't really come across other components that give an option to change the reporting format, that was really the primary reason why I didn't implement it. Although then if we do make an option to have it convert to degrees within the component, we'd then need to decide does it floor the degrees to give a whole degree, does it round to a certain decimal place, etc. The use case for any of them could be valid...this is probably a contributing reason for why we don't really see other components doing options for conversion built-in to the component.

fran6120:
You are right! only using `floor` solve the 0 and 360 issue.

The other part of the code that no longer works, was used with the previous library, when adding an angle start offset.  Function that you have implemented with `start_position`

Thanks you!

esphome/components/as5600/as5600.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  // calculate allowed range of motion by taking the start from the end
+  // but only if the end is greater than the start. If the start is greater
+  // than the end position, then that means we take the start all the way to
+  // reset point (i.e. 0 deg raw) and then we that with the end position
+  uint16_t range = this->raw_max_ > this->start_position_ ? this->raw_max_ - this->start_position_
+                                                          : (4095 - this->start_position_) + this->raw_max_;
+
+  // range scale is ratio of actual allowed range to the full range
+  this->range_scale_ = range / 4095.0;
```
jesserockz:
```suggestion
  this->range_scale_ = range / 4095.0f;
```

esphome/components/as5600/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+    CONF_ID,
+    CONF_DIR_PIN,
+    CONF_DIRECTION,
+    CONF_HYSTERESIS,
+    CONF_RANGE,
+)
+
+CODEOWNERS = ["@ammmze"]
+DEPENDENCIES = ["i2c"]
+AUTO_LOAD = ["sensor"]
```
jesserockz: Please alter the PR code to work without autoloading `sensor`
```suggestion
```

jesserockz (changes requested):

esphome/components/as5600/as5600.cpp
ammmze: 👍🏻  Done!

esphome/components/as5600/__init__.py
ammmze: 👍🏻 Done! Good catch...guess I just forgot to remove that after I moved the sensors to `sensor.py`.

ammmze:
Looks like my [esphome-docs PR](https://github.com/esphome/esphome-docs/pull/3098) got closed and locked since the PRs have just been sitting idle for so long waiting for reviews. I've just rebased it and pushed new changes. Are you, @jesserockz, able to unlock it and re-open it? Or do I need to open a new PR in esphome-docs?

jesserockz:
@ammmze GitHub wont let me reopen it because you force pushed your branch afterwards. Please make a new one. Sorry about that.

esphome/components/as5600/sensor.py
```diff
@@ -0,0 +1,120 @@
...
+    ENTITY_CATEGORY_DIAGNOSTIC,
+    CONF_MAGNITUDE,
+    CONF_STATUS,
+    CONF_POSITION,
+)
+from . import as5600_ns, AS5600Component
+
+CODEOWNERS = ["@ammmze"]
+DEPENDENCIES = ["as5600"]
+AUTO_LOAD = ["as5600"]
```
jesserockz: This should be one or the other, not both.

jesserockz (changes requested):
Please move the sensor class to its own file `as5600_sensor.h/cpp` and place it in a `sensor` subdirectory. Also rename `sensor.py` to be `sensor/__init__.py` in that new subdirectory. 
This will also require altering some of the main component code as it should not know about AS5600Sensor. This is something that many ESPHome components do already if you need examples.

ammmze:
> Please move the sensor class to its own file `as5600_sensor.h/cpp` and place it in a `sensor` subdirectory. Also rename `sensor.py` to be `sensor/__init__.py` in that new subdirectory. This will also require altering some of the main component code as it should not know about AS5600Sensor. This is something that many ESPHome components do already if you need examples.

These changes have been completed and I've opened a [new PR in esphome-docs](https://github.com/esphome/esphome-docs/pull/3304) to replace the one that got closed.

esphome/components/as5600/sensor/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_parented(var, config[CONF_AS5600_ID])
+    await cg.register_component(var, config)
+    await sensor.register_sensor(var, config)
+
+    if CONF_OUT_OF_RANGE_MODE in config:
+        cg.add(var.set_out_of_range_mode(config[CONF_OUT_OF_RANGE_MODE]))
```
jesserockz: We have been slowly converting ESPHome to_code functions to use this format for optional config entries. Please alter them to match =)
```suggestion
    if out_of_range_mode_config := config.get(CONF_OUT_OF_RANGE_MODE):
        cg.add(var.set_out_of_range_mode(out_of_range_config_mode))
```

esphome/components/as5600/sensor/as5600_sensor.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+AS5600Sensor::AS5600Sensor(){};
+
+float AS5600Sensor::get_setup_priority() const { return setup_priority::DATA; }
+
+void AS5600Sensor::dump_config() {
+  LOG_SENSOR("", "AS5600 Sensor", this);
+  ESP_LOGCONFIG(TAG, "  Out of Range Mode: %u", this->out_of_range_mode_);
+  LOG_UPDATE_INTERVAL(this);
+}
```
jesserockz: This should also run LOG_SENSOR for all of the sub-sensors

esphome/components/as5600/sensor/as5600_sensor.h
```diff
@@ -0,0 +1,43 @@
...
+#include "esphome/components/sensor/sensor.h"
+#include "esphome/components/as5600/as5600.h"
+
+namespace esphome {
+namespace as5600 {
+
+class AS5600Sensor : public PollingComponent, public Parented<AS5600Component>, public sensor::Sensor {
+ public:
+  AS5600Sensor();
+
```
jesserockz: Not doing anything useful
```suggestion
```

esphome/components/as5600/sensor/as5600_sensor.cpp
```diff
@@ -0,0 +1,79 @@
...
+static const uint8_t REGISTER_ANGLE_RAW = 0x0C;  // 16 bytes / R
+static const uint8_t REGISTER_ANGLE = 0x0E;      // 16 bytes / R
+
+// Status registers
+static const uint8_t REGISTER_STATUS = 0x0B;     // 8 bytes  / R
+static const uint8_t REGISTER_AGC = 0x1A;        // 8 bytes  / R
+static const uint8_t REGISTER_MAGNITUDE = 0x1B;  // 16 bytes / R
+
+AS5600Sensor::AS5600Sensor(){};
+
```
jesserockz:
```suggestion
```

esphome/components/as5600/sensor/as5600_sensor.h
```diff
@@ -0,0 +1,43 @@
...
+  void update() override;
+  void dump_config() override;
+  float get_setup_priority() const override;
+
+  void set_angle_sensor(sensor::Sensor *angle_sensor) { angle_sensor_ = angle_sensor; }
```
jesserockz: Please make sure all member variables are prefixed with `this->` as per our [codebase standards](https://esphome.io/guides/contributing#codebase-standards)

```suggestion
  void set_angle_sensor(sensor::Sensor *angle_sensor) { this->angle_sensor_ = angle_sensor; }
```

jesserockz (changes requested):
Thanks for making the requested changes and splitting the sensor code out.

A few more things below, but nothing serious =)

ammmze:
> Thanks for making the requested changes and splitting the sensor code out.
> 
> A few more things below, but nothing serious =)

Thank you for taking the time to look through this again! I've made the requested changes.

esphome/components/as5600/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+
+    if dir_pin_config := config.get(CONF_DIR_PIN):
+        pin = await cg.gpio_pin_expression(dir_pin_config)
+        cg.add(var.set_dir_pin(pin))
+
+    if end_position_config := config.get(CONF_END_POSITION):
+        cg.add(var.set_end_position(end_position_config))
+
+    if range_config := config.get(CONF_RANGE):
+        cg.add(var.set_range(range_config))
```
jesserockz: These need to have `None` checks as they could be 0 integers which would be falsy and be skipped over.
```suggestion
    if (end_position_config := config.get(CONF_END_POSITION, None)) is not None:
        cg.add(var.set_end_position(end_position_config))

    if (range_config := config.get(CONF_RANGE, None)) is not None:
        cg.add(var.set_range(range_config))
```

jesserockz (approved):

jesserockz:
Dont forget to take a look at the docs PR changes. Once both PRs are approved they can be merged =)

PR #5191, opened 2023-08-03
Pn532 non blocking scan

### LINKED ISSUE #4382: PN532 component blocks for 100ms every update
### The problem

The `read_data` function on pn532_i2c component block the code execution for up to 100ms. Since it's called at least once every `update_interval`, it can consume significant processing time (>10% if update_interval is the default of 1s, and a lot more if the interval is less for quicker tag detection), in addition to hogging the I²C bus while doing that. It will always timeout those 100ms if no tag is presented to the reader. This is especially noticeable when running other timing-sensitive operations (e.g.: rtttl buzzer, fast-updating sensors).
```
bool PN532I2C::read_data(std::vector<uint8_t> &data, uint8_t len) {
  delay(1);

  std::vector<uint8_t> ready;
  ready.resize(1);
  uint32_t start_time = millis();
  while (true) {
    if (this->read_bytes_raw(ready.data(), 1)) {
      if (ready[0] == 0x01)
        break;
    }

    if (millis() - start_time > 100) {
      ESP_LOGV(TAG, "Timed out waiting for readiness from PN532!");
      return false;
    }
  }

  data.resize(len + 1);
  this->read_bytes_raw(data.data(), len + 1);
  return true;
}
```
The ESP will be stuck in the `while(true)` loop until 100ms pass, or a tag is presented thus causing PN532 to respond with `0x01`.

I am working on an asynchronous implementation shifting waiting for a `0x01` from the chip to the `loop()` function, that seems to be working fine.

2023.3.2

Home Assistant Add-on

2023.3.5

ESP32-IDF

### Board

WT32-ETH01

pn532, pn532_i2c

### Example YAML snippet

```yaml
i2c:
  scl: GPIO32
  sda: GPIO33
  scan: True
  frequency: 100kHz
  timeout: 2ms  # see https://github.com/esphome/issues/issues/4340

pn532_i2c:
  update_interval: 250ms
  id: rfid
  on_tag:
    then:
      - binary_sensor.template.publish:
          id: tag_present
          state: ON
```

_No response_

### Additional information

_No response_

sharkydog:

Do not block loop() while waiting for tags to appear (command InListPassiveTarget).
Reading and writing NDEF is still blocking.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4382

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/pn532/pn532.cpp
+++ b/esphome/components/pn532/pn532.cpp
@@ -127,14 +127,14 @@ void PN532::loop() {
   if (!this->requested_read_)
     return;
 
-  int8_t ready = this->read_ready_(false);
-  if (ready == -1)
+  auto ready = this->read_ready_(false);
+  if (ready == WOULDBLOCK)
     return;
 
   bool success = false;
   std::vector<uint8_t> read;
 
-  if (ready) {
+  if (ready == READY) {
     success = this->read_response(PN532_COMMAND_INLISTPASSIVETARGET, read);
   } else {
     this->send_ack_();  // abort still running InListPassiveTarget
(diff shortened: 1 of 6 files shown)


esphome/components/pn532/pn532.cpp
```diff
@@ -286,12 +296,59 @@ bool PN532::read_ack_() {
...
+}
 void PN532::send_nack_() {
   ESP_LOGV(TAG, "Sending NACK for retransmit");
   this->write_data({0x00, 0x00, 0xFF, 0xFF, 0x00, 0x00});
   delay(10);
 }
 
+int8_t PN532::read_ready_(bool block) {
+  static uint32_t start_time = 0;
+  static int8_t ready = -1;
```
jesserockz: This prevents using multiple pn532 at the same time...
Which is not common, but there was a request for it so I added it

jesserockz (changes requested):

esphome/components/pn532/pn532.cpp
sharkydog: then move those as instance properties?

sharkydog: And maybe convert `int8_t ready` [-1,0,1] to enum [WOULDBLOCK,TIMEOUT,READY]?

sharkydog: Also should I add `yield()` at the end of the loop (if block==true) as in pn532_spi before this PR?

github-k8n:
Perfect, I tested the PR using 
external_components:
  - source: github://pr#5191
    components: [ pn532, pn532_i2c, pn532_spi ]

 and it pretty much solved the issue of pn532 blocking everything else all the time.
However i noticed that it still seems to have a delay (130ms) right after i scan a tag (but not at any other time), not sure if this is related here, just wanted to bring it up in case there is some edgecase that was forgotten:

EDIT: HOWEVER, when running an addressable light in a "scanner" config where it was previously pretty obvious when somehting was blocking, this does not seem to show any blocking even if the logs say otherwise...


[17:39:31][D][pn532:200]: Found new tag '0F-14-57-34'
[17:39:31][D][pn532:379]: Waiting to read next tag
[17:39:31][W][component:204]: Component pn532 took a long time for an operation (0.13 s).
[17:39:31][W][component:205]: Components should block for at most 20-30ms.

sharkydog:
Yep, reading is still blocking.
The whole component need to be reworked to fix this, probably along with the nfc component.

sharkydog:
Also, if interest in this PR "sparks", I won't have the time to update and test it to current dev in the next 3-4 weeks, so I will happily give up credit to someone who wants to speed things up.

jesserockz (approved):

PR #5195, opened 2023-08-03
Allow components to be internal with respect to publishing via MQTT

### LINKED ISSUE #375: Silent option for MQTT (for use alongside Native API)

Add a silent option to MQTT Client which suppresses automatic reporting of ESPHome component (sensor, etc...) values but still allows one to use MQTT features in ESPHome such as triggers, actions (publish), and sensors (subscribe).

When using the Native API alongside the MQTT Client (with discovery set to false) the values of all ESPHome components are also reported over MQTT automatically creating a lot of unnecessary traffic which can add up over a lot of ESPs. 

For example, one may wish to use MQTT alongside the Native API, for:
- Home Assistant sensors that expect MQTT such as [MQTT Room Presence](https://www.home-assistant.io/components/mqtt_room/)
- Simple communication between ESPs (with minimal setup/overhead and not reliant on Home Assistant)
- Commerical devices that have MQTT interfaces
- etc...

My primary use case (Room Presence) could also be done on the Home Assistant side with an automation (for each ESP32) and disabling MQTT on the ESP32 but I think that wouldn't be as clean a solution as directly on the ESPs (i.e. edge computing). 

**Additional context**

I ran across this issue since I have just started using ESPHome (migrated from a custom solution and have been loving it so far - great job btw!) and wanted to setup Room Presence. This is fairly simply using the existing tools (BLE rssi sensor, lambda filter to convert to distance, on_value action with mqtt_publish_json) - I would like to write this up in the cookbook section when I get the chance. The only issue is the extra, unnecessary (IMO) traffic this creates when also using the Native API.

kahrendt:

When using the native API and MQTT, the ``internal`` setting makes no differentiation between the two. This pull request allows every MQTT component to be separately marked internal or not. If ``internal_mqtt`` is set to true, then the component will never publish its state via MQTT. If ``internal_mqtt`` is set to false, then the component will always publish its state via MQTT.

There is an additional configuration option for the base mqtt component called ``internal_mqtt_default``. If set to ``copy``, then all MQTT components will, by default, follow the regular ``internal`` setting. If set to ``internal``, all components, by default, do not publish their state via MQTT. If set to ``external``, all components, by default, publish their state via MQTT. Do note that setting the ``internal_mqtt`` on individual components overrides this default setting.

The default options give the current behavior, so this is not a breaking change.

Note that if you want to test this PR out, you cannot just use it via ``external_components``, as there are some modifications to the core. You need to install this PR specifically to use the new features. (Please let me know if there is a better way?)

I struggled with how to name these options as there are really three possibilities here; force publish, force not publish, or mimic current behavior. Please give any suggestions with regard to the terminology or to the implementation itself!

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/feature-requests#375

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3107

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

mqtt:
  broker: 10.0.0.2
  username: !secret mqtt_username
  internal_mqtt_default: internal

sensor:
  - platform: template
    id: test_sensor
    lambda: |-
      return 5.0;
  
  - platform: copy
    source_id: test_sensor
    name: "API External, MQTT Internal" # will publish via the API, but not through MQTT
    internal: false
    internal_mqtt: true  
  
  - platform: copy
    source_id: test_sensor
    name: "API Internal, MQTT External" # will publish via MQTT but not through API
    internal: true
    internal_mqtt: false #

  - platform: copy
    source_id: test_sensor
    name: "API External, MQTT Internal by Default" # will publish via the API, but not via MQTT due to global default

```

## Checklist:


SeByDocKy:
Ho !!!! it's a great addition....

PR #5198, opened 2023-08-04
Fix AHT10 / AHT20 communication

flaminggoat:

Fixes communication with AHT20 modules & presumably AHT10 modules (though I don't have an AHT10 to test with). The implementation before this PR sends the messages shown in the scope capture, which include a byte of 0x00 after the device address, which is not what is specified in the [AHT10](https://kaimte.com/upload/attach/20200514/2786ec587a09412dcd95609fe91e00e4.pdf) or [AHT20 datasheet](https://cdn.sparkfun.com/assets/d/2/b/e/d/AHT20.pdf) and results in the AHT20 sending a NACK. So I am not sure how this code was working at all previously?
image
This PR corrects that, and allows communication with the module.

In addition the AHT20 datasheet indicates that the calibration/initialisation command is different for the AHT20 compared to the AHT10 module, so a variant config option has been added to allow selection of AHT10 or AHT20 modules. I found that my AHT20 module would not calibrate unless the command specified in the data sheet was used.

AHT20 Datasheet commands
image

AHT10 Datasheet commands
image

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** possibly related https://github.com/esphome/issues/issues/3941

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3221

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
i2c:
  sda: 6
  scl: 7
  scan: false
  id: bus_a
  frequency: 200kHz
    
sensor:
  - platform: aht10
    variant: aht20
    i2c_id: bus_a
    id: aht_20
    address: 0x38
    temperature:
      name: "Temperature"
    humidity:
      name: "Humidity"
    update_interval: 5s
```

## Checklist:
image

### DIFF
--- a/esphome/components/aht10/aht10.cpp
+++ b/esphome/components/aht10/aht10.cpp
@@ -42,7 +42,7 @@ void AHT10Component::setup() {
       ESP_LOGCONFIG(TAG, "Setting up AHT10");
   }
 
-  if (this->write(calibrate_cmd, sizeof(AHT10_CALIBRATE_CMD)) != i2c::ERROR_OK) {
+  if (this->write(calibrate_cmd, sizeof(AHT20_CALIBRATE_CMD)) != i2c::ERROR_OK) {
     ESP_LOGE(TAG, "Communication with AHT10 failed!");
     this->mark_failed();
     return;
--- a/esphome/components/aht10/aht10.cpp
+++ b/esphome/components/aht10/aht10.cpp
@@ -30,12 +30,17 @@ static const uint8_t AHT10_CAL_ATTEMPTS = 10;
 static const uint8_t AHT10_STATUS_BUSY = 0x80;
 
 void AHT10Component::setup() {
-  const uint8_t *calibrate_cmd = AHT10_CALIBRATE_CMD;
-  if (this->variant_ == "aht20") {
-    calibrate_cmd = AHT20_CALIBRATE_CMD;
+  const uint8_t *calibrate_cmd;
+  switch (this->variant_) {
+    case AHT10Variant::AHT20:
+      calibrate_cmd = AHT20_CALIBRATE_CMD;
+      ESP_LOGCONFIG(TAG, "Setting up AHT20");
+      break;
+    case AHT10Variant::AHT10:
+    default:
+      calibrate_cmd = AHT10_CALIBRATE_CMD;
+      ESP_LOGCONFIG(TAG, "Setting up AHT10");
   }
-
-  ESP_LOGCONFIG(TAG, "Setting up AHT10 (variant: %s)", this->variant_.c_str());
 
   if (this->write(calibrate_cmd, sizeof(AHT10_CALIBRATE_CMD)) != i2c::ERROR_OK) {
     ESP_LOGE(TAG, "Communication with AHT10 failed!");
@@ -53,7 +58,7 @@ void AHT10Component::setup() {
     }
     ++cal_attempts;
     if (cal_attempts > AHT10_CAL_ATTEMPTS) {
-      ESP_LOGE(TAG, "AHT10 calibration hung!");
+      ESP_LOGE(TAG, "AHT10 calibration timed out!");
       this->mark_failed();
       return;
     }
--- a/esphome/components/aht10/aht10.cpp
+++ b/esphome/components/aht10/aht10.cpp
@@ -26,11 +26,11 @@ static const uint8_t AHT10_MEASURE_CMD[] = {0xAC, 0x33, 0x00};
 static const uint8_t AHT10_DEFAULT_DELAY = 5;    // ms, for calibration and temperature measurement
 static const uint8_t AHT10_HUMIDITY_DELAY = 30;  // ms
 static const uint8_t AHT10_ATTEMPTS = 3;         // safety margin, normally 3 attempts are enough: 3*30=90ms
-static const uint8_t AHT10_CAL_ATTEMPTS = 10; 
+static const uint8_t AHT10_CAL_ATTEMPTS = 10;
 static const uint8_t AHT10_STATUS_BUSY = 0x80;
 
 void AHT10Component::setup() {
-  const uint8_t * calibrate_cmd = AHT10_CALIBRATE_CMD;
+  const uint8_t *calibrate_cmd = AHT10_CALIBRATE_CMD;
   if (this->variant_ == "aht20") {
     calibrate_cmd = AHT20_CALIBRATE_CMD;
   }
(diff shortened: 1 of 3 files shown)


esphome/components/aht10/aht10.h
```diff
@@ -13,13 +15,15 @@ class AHT10Component : public PollingComponent, public i2c::I2CDevice {
...
   float get_setup_priority() const override;
+  void set_variant(std::string variant) { this->variant_ = std::move(variant); }
 
   void set_temperature_sensor(sensor::Sensor *temperature_sensor) { temperature_sensor_ = temperature_sensor; }
   void set_humidity_sensor(sensor::Sensor *humidity_sensor) { humidity_sensor_ = humidity_sensor; }
 
  protected:
   sensor::Sensor *temperature_sensor_{nullptr};
   sensor::Sensor *humidity_sensor_{nullptr};
+  std::string variant_ = "aht10";
```
jesserockz: Please change this to be an `enum` similar to other components that have a `model` configuration

jesserockz (changes requested):

esphome/components/aht10/aht10.h
flaminggoat: Thanks for reviewing, I have changed to use an enum. I can add a PR for the documentation if this approach is okay.  My only concern is I don't understand how it could have been working before for the AHT10 with the `write_bytes(0,...` function. The updated code is inline with other AHT10 libraries such as the [adafruit one](https://github.com/adafruit/Adafruit_AHTX0/blob/d0aff01e79cc36ab0c0df01e2b17663fc99b0d02/Adafruit_AHTX0.cpp#L134C19-L134C19)

esphome/components/aht10/aht10.h
```diff
@@ -1,25 +1,31 @@
...
   float get_setup_priority() const override;
+  void set_variant(AHT10Variant variant) { this->variant_ = variant; }
 
   void set_temperature_sensor(sensor::Sensor *temperature_sensor) { temperature_sensor_ = temperature_sensor; }
   void set_humidity_sensor(sensor::Sensor *humidity_sensor) { humidity_sensor_ = humidity_sensor; }
 
  protected:
   sensor::Sensor *temperature_sensor_{nullptr};
   sensor::Sensor *humidity_sensor_{nullptr};
+  AHT10Variant variant_ = AHT10Variant::AHT10;
```
jesserockz:
```suggestion
  AHT10Variant variant_{};
```

esphome/components/aht10/aht10.cpp
```diff
@@ -20,36 +20,48 @@ namespace esphome {
...
+      ESP_LOGCONFIG(TAG, "Setting up AHT20");
+      break;
+    case AHT10Variant::AHT10:
+    default:
+      calibrate_cmd = AHT10_CALIBRATE_CMD;
+      ESP_LOGCONFIG(TAG, "Setting up AHT10");
+  }
 
-  if (!this->write_bytes(0, AHT10_CALIBRATE_CMD, sizeof(AHT10_CALIBRATE_CMD))) {
+  if (this->write(calibrate_cmd, sizeof(AHT10_CALIBRATE_CMD)) != i2c::ERROR_OK) {
```
jesserockz: This probably should have a size const instead of using just the aht10 size.

jesserockz (changes requested):

esphome/components/aht10/aht10.cpp
flaminggoat:
```suggestion
  if (this->write(calibrate_cmd, sizeof(AHT20_CALIBRATE_CMD)) != i2c::ERROR_OK) {
```

jesserockz: You missed the important part of what I said... 
Instead of using the size of aht10 or aht20, just define a new const that is the expected size to use for both

esphome/components/aht10/aht10.cpp
```diff
@@ -20,36 +20,48 @@ namespace esphome {
...
+      ESP_LOGCONFIG(TAG, "Setting up AHT20");
+      break;
+    case AHT10Variant::AHT10:
+    default:
+      calibrate_cmd = AHT10_CALIBRATE_CMD;
+      ESP_LOGCONFIG(TAG, "Setting up AHT10");
+  }
 
-  if (!this->write_bytes(0, AHT10_CALIBRATE_CMD, sizeof(AHT10_CALIBRATE_CMD))) {
+  if (this->write(calibrate_cmd, sizeof(AHT20_CALIBRATE_CMD)) != i2c::ERROR_OK) {
```
jesserockz: At the top of this file add:
```c++
static const uint8_t CALIBRATE_CMD_LENGTH = 3;
```

and use that here

SasaKaranovic:
Tested the PR and can confirm it's working as expected with AHT20 sensor (AHT20 + BMP280 board) and ESP32 (ESP32 C3 Super Mini).

Configuration file used is below:
```
esphome:
  name: sk-test
  platform: ESP32
  board: lolin_c3_mini

# Enable logging
logger:

# Enable Home Assistant API
api:
  encryption:
    key: "---redacted---"

ota:
  password: "---redacted---"

wifi:
  ssid: "---redacted---"
  password: "---redacted---"

web_server:
  port: 80

i2c:
  sda: 8
  scl: 9
  scan: true
  id: bus_a

sensor:
  - platform: bmp280
    temperature:
      id: 'bmp_temp'
      name: "Outside Temperature"
      oversampling: 16x
    pressure:
      id: 'bmp_pressure'
      name: "Outside Pressure"
    address: 0x77
    update_interval: 10s

  - platform: aht10
    variant: aht20
    i2c_id: bus_a
    id: aht_20
    address: 0x38
    temperature:
      id: 'aht_temp'
    humidity:
      id: 'aht_humid'
      name: "Living Room Humidity"
    update_interval: 10s

external_components:
  - source: github://pr#5198
    components: [ aht10 ]

```

esphome/components/aht10/aht10.cpp
flaminggoat: done

flaminggoat:
As I guess there were no complaints with the general design of the change I have now added a PR for the documentation too https://github.com/esphome/esphome-docs/pull/3221

heintjeput:
This did solve an issue for me with the AHT21 sensor. Is there a way this can be merged into main?

jesserockz (approved):

Wouter-Lemoine:
I can also confirm this PR works for an "AHT30" sensor on a Wemos D1 mini using the aht20 variant. The aht30 also outputs a 7th byte containing a CRC according to the datasheet, but it doesn't seem to get in the way of the functionality.

kristof-dragon:
I can also confirm, that this PR solved the "Communication with AHT10 failed!" issue, where the I2C scan would detect the sensor, but it would not work.

I tested it on the latest v2023.11.6 ESPhome with
AHT20 + BMP280 sensor board on ESP32
AHT20 + ENS160 sensor board on ESP8266

z3roq:
Also can confirm that this fix "Communication with AHT10 failed!" issue for me with ESP32 C3 SuperMini using AHT20 + BMP280 cheap chinese board.

PR #5200, opened 2023-08-04
i2c: fix build on ESP-IDF >= 5.1

stintel:

As of version 5.1, i2c_port_t is an enum.

Compile error before this change:

```
src/esphome/components/i2c/i2c_bus_esp_idf.cpp:18:33: error: invalid conversion from 'int' to 'i2c_port_t' [-fpermissive]
   18 |   static i2c_port_t next_port = 0;
      |                                 ^
      |                                 |
      |                                 int
src/esphome/components/i2c/i2c_bus_esp_idf.cpp:19:20: error: no 'operator++(int)' declared for postfix '++' [-fpermissive]
   19 |   port_ = next_port++;
      |           ~~~~~~~~~^~
```

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** NA

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** NA

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
esphome:
  name: bug

esp32:
  board: esp32-c3-devkitm-1
  framework:
    platform_version: https://github.com/platformio/platform-espressif32
    type: esp-idf
    version: 5.1.0

i2c:
  sda: 4
  scl: 5

```

## Checklist:

### DIFF
--- a/esphome/components/i2c/i2c_bus_esp_idf.cpp
+++ b/esphome/components/i2c/i2c_bus_esp_idf.cpp
@@ -15,8 +15,13 @@ static const char *const TAG = "i2c.idf";
 
 void IDFI2CBus::setup() {
   ESP_LOGCONFIG(TAG, "Setting up I2C bus...");
-  static i2c_port_t next_port = 0;
-  port_ = next_port++;
+  static i2c_port_t next_port = I2C_NUM_0;
+  port_ = next_port;
+#if I2C_NUM_MAX > 1
+  next_port = (next_port == I2C_NUM_0) ? I2C_NUM_1 : I2C_NUM_MAX;
+#else
+  next_port = I2C_NUM_MAX;
+#endif
 
   recover_();
 
--- a/esphome/components/i2c/i2c_bus_esp_idf.cpp
+++ b/esphome/components/i2c/i2c_bus_esp_idf.cpp
@@ -23,11 +23,11 @@ void IDFI2CBus::setup() {
   next_port = I2C_NUM_MAX;
 #endif
 
-if (port_ == I2C_NUM_MAX) {
-  ESP_LOGE(TAG, "Too many I2C buses configured");
-  this->mark_failed();
-  return;
-}
+  if (port_ == I2C_NUM_MAX) {
+    ESP_LOGE(TAG, "Too many I2C buses configured");
+    this->mark_failed();
+    return;
+  }
 
   recover_();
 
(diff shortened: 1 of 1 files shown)


jesserockz (changes requested):
This still needs to increment the port in each setup...

stintel:
> This still needs to increment the port in each setup...

Sorry about that. When I initially did this change I was in "let's just get things to compile" mode. Should have reviewed before opening the PR.

Should be better now. I don't like how it looks, but don't have any better ideas.

esphome/components/i2c/i2c_bus_esp_idf.cpp
```diff
@@ -15,8 +15,18 @@ static const char *const TAG = "i2c.idf";
 
 void IDFI2CBus::setup() {
   ESP_LOGCONFIG(TAG, "Setting up I2C bus...");
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 1, 0)
```
jesserockz: I actually dont think we need to check for idf 5.1, all of the used "variables" are available on the current 4.4, just as defines instead of the enums

stintel: Good catch! Updated and compile-tested with IDF 4.4 and 5.1.

esphome/components/i2c/i2c_bus_esp_idf.cpp
```diff
@@ -15,8 +15,13 @@ static const char *const TAG = "i2c.idf";
...
   ESP_LOGCONFIG(TAG, "Setting up I2C bus...");
-  static i2c_port_t next_port = 0;
-  port_ = next_port++;
+  static i2c_port_t next_port = I2C_NUM_0;
+  port_ = next_port;
+#if I2C_NUM_MAX > 1
+  next_port = (next_port == I2C_NUM_0) ? I2C_NUM_1 : I2C_NUM_MAX;
+#else
+  next_port = I2C_NUM_MAX;
+#endif
```
jesserockz, line 25:
```suggestion

if (port_ == I2C_NUM_MAX) {
  ESP_LOGE(TAG, "Too many I2C buses configured");
  this->mark_failed();
  return;
}

```

jesserockz (approved):

PR #5206, opened 2023-08-05
Daly bms performance

andreashergert1984:

due to the not optimal performance of sending all 8 requests in on loop the maximum loop time exeeds.
image

This improvements splits those requests up so that the sending as well as response handling is split up in multiple update-calls.
The loop time is now not warning anymore.

## Types of changes

- [X] Performance improvement

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

```yaml
daly_bms:
  update_interval: 1s
```

## Checklist:

### DIFF
--- a/esphome/components/daly_bms/daly_bms.cpp
+++ b/esphome/components/daly_bms/daly_bms.cpp
@@ -21,8 +21,8 @@ static const uint8_t DALY_REQUEST_TEMPERATURE = 0x96;
 
 void DalyBmsComponent::setup() {}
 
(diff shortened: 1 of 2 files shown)


jesserockz (changes requested):
The better way to solve this is to set request_id_ to a value on `update()`, and `loop()` should be checking that value and also check it is not waiting for a response before sending the next request.

jesserockz:
duplicate of #3388

jesserockz:
Closing as a duplicate of #3388

PR #3255, opened 2022-02-27
Online PNG image support

guillempages:
# What does this implement/fix?

This PR implements the possibility of downloading PNG images at runtime and showing them in a compatible display.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** relates to https://github.com/esphome/feature-requests/issues/1463

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2076

## Test Environment

- [x] ESP32 (Inkplate10)
- [ ] ESP32 IDF
- [ ] ESP8266

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
online_image:
  - url: "https://www.example.org/image.png"
    id: example_image
    format: PNG

display:
  # ...
  lambda: |-
    it.image(0, 0, id(example_image));

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #5214, opened 2023-08-08
Allow images to be downloaded from URLs

landonr:

## Types of changes

- [ ] Other

related #3255 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2409

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
image:
  - id: launch_image
    file: https://raw.githubusercontent.com/landonr/homeThing/main/docs/homeThing%20logo.svg
    resize: 48x48
    type: TRANSPARENT_BINARY

```

## Checklist:


landonr:
@guillempages wassup

esphome/components/image/__init__.py
```diff
@@ -47,6 +48,8 @@
 
 SOURCE_LOCAL = "local"
 SOURCE_MDI = "mdi"
+SOURCE_WEB = "web"
+CONF_FILETYPE = "filetype"
```
guillempages: Are you using the `filetype` config option? I couldn't find it (but I might have overlooked that)

esphome/components/image/__init__.py
```diff
@@ -74,6 +77,56 @@ def download_mdi(value):
     return value
 
 
+def get_filename_from_url(url):
+    # Regular expression pattern to match the file name at the end of the URL
+    pattern = r"/([^/]+)$"
+    match = re.search(pattern, url)
+    if match:
+        file_name = match.group(1)
```
guillempages: For consistency with the method's name (and the rest of the code), you might want to rename "file_name" to a single word filename (valid for all instances)
```suggestion
        filename = match.group(1)
```

esphome/components/image/__init__.py
```diff
@@ -74,6 +77,56 @@ def download_mdi(value):
     return value
 
 
+def get_filename_from_url(url):
+    # Regular expression pattern to match the file name at the end of the URL
+    pattern = r"/([^/]+)$"
```
guillempages: You can put two subgroups in the regex, and then parse filename and extension with a single call.
```suggestion
    pattern = r"/([^/]+)(\.([^./]+))?$"
```
`group(1)` would be filename (without extension) and `group(3)`  the extension (only if group(2) finds a match).
Then you could return a tuple instead of a single value, and avoid the next method.

esphome/components/image/__init__.py
```diff
@@ -74,6 +77,56 @@ def download_mdi(value):
...
+
+
+def get_file_type(file_name):
+    # Regular expression pattern to match the file extension
+    pattern = r"\.([^.]+)$"
+    match = re.search(pattern, file_name)
+    if match:
+        file_type = match.group(1)
+        return file_type
+    return None
```
guillempages: See above; it can be merged into "get_filename_from_url"

esphome/components/image/__init__.py
```diff
@@ -74,6 +77,56 @@ def download_mdi(value):
     return value
 
 
+def get_filename_from_url(url):
+    # Regular expression pattern to match the file name at the end of the URL
+    pattern = r"/([^/]+)$"
+    match = re.search(pattern, url)
+    if match:
+        file_name = match.group(1)
+        file_name = re.sub(r"%20", "", file_name)  # Replace %20 with empty string
```
guillempages: "%20" is normally a white space; it might be better to replace it with "_" for readability?

esphome/components/image/__init__.py
```diff
@@ -74,6 +77,56 @@ def download_mdi(value):
...
+        file_type = match.group(1)
+        return file_type
+    return None
+
+
+def _compute_local_image_path(value) -> Path:
+    image_id = get_filename_from_url(value[CONF_URL])
+    filetype = get_file_type(value[CONF_URL])
+    base_dir = Path(CORE.config_dir) / ".esphome" / DOMAIN / f"{filetype}"
+    return base_dir / f"{image_id}.{filetype}"
```
guillempages: Instead of a new method doing (mostly) the same, the already existing `_compute_local_icon_path` method above could be re-used. You could add a second parameter "extension" or similar to pass the filetype.

esphome/components/image/__init__.py
```diff
@@ -74,6 +77,56 @@ def download_mdi(value):
...
+
+
+def download_image(value):
```
guillempages, line 117: This method has a lot of common code with `download_mdi`, that could be reused.

Maybe put the common part into a new internal method, and just keep the URL/MDI name parsing part in the respective "public" download methods?

esphome/components/image/__init__.py
```diff
@@ -139,6 +192,19 @@ def validate_file_shorthand(value):
                 CONF_ICON: icon,
             }
         )
+    if value.startswith("https"):
```
guillempages: I'm also in favor of the "https only" Internet, but there are people still using non-encrypted http ;-) (Especially in local-only setups)
```suggestion
    if value.startswith("http://") or value.startswith("https://"):
```

esphome/components/image/__init__.py
```diff
@@ -139,6 +192,19 @@ def validate_file_shorthand(value):
...
+        validate_cairosvg_installed(value)
```
guillempages: cairosvg should only be needed for svg images, but the URL could be another kind of image (e.g. PNG). 
I'm not 100% sure how to handle this best though.

esphome/components/image/__init__.py
```diff
@@ -139,6 +192,19 @@ def validate_file_shorthand(value):
...
+
+        # match = re.search(r"([a-zA-Z0-9\-]+)", value)
+        # if match is None:
+        #     raise cv.Invalid("Could not parse mdi icon name.")
+        # icon = match.group(1)
```
guillempages: Any reason you left this in?
```suggestion
```

esphome/components/image/__init__.py
```diff
@@ -74,6 +77,56 @@ def download_mdi(value):
...
+
+def download_image(value):
+    image_id = get_filename_from_url(value[CONF_URL])
+    path = _compute_local_image_path(value)
+    if path.is_file():
+        return value
+    url = value[CONF_URL]
+    _LOGGER.debug("Downloading %s image from %s", image_id, url)
+    try:
+        req = requests.get(url, timeout=MDI_DOWNLOAD_TIMEOUT)
```
guillempages: Since now the timeout is used as well for image download, you might want to rename the constant.
```suggestion
        req = requests.get(url, timeout=IMAGE_DOWNLOAD_TIMEOUT)
```

guillempages (commented):
@landonr sorry, was AFK for a while.

I think your PR could be improved by reusing some of the MDI download code; left some comments.

esphome/components/image/__init__.py
guillempages: OTOH, it might be a good idea to use a hash instead of the actual filename (like the fonts are doing), to allow having similar images coming from different URLs (e.g. different weather icons, from separate providers; you might have `http://example.com/cool_icons/sunny.svg` and `https://example.org/big_colored_icon/sunny.svg`.)

For MDI icons this is not an issue, and it is easier to debug if the icon name is visible instead of the hash

landonr:
sweet thanks for the review @guillempages! Hope you had a nice time afk! As I mentioned in #5254 i'll do those changes first so I can pull the duplicate code into here

esphome/external_files.py
```diff
@@ -0,0 +1,81 @@
...
+def is_file_recent(file_path: str, refresh: TimePeriodSeconds) -> bool:
+    if os.path.exists(file_path):
+        creation_time = os.path.getctime(file_path)
+        current_time = datetime.now().timestamp()
+        return current_time - creation_time <= refresh.total_seconds
+    return False
+
+
+def compute_local_file_dir(name: str, domain: str) -> Path:
+    base_directory = Path(CORE.config_dir) / ".esphome" / domain
```
jesserockz: This should be using the data_dir, not the config_dir
```suggestion
    base_directory = Path(CORE.data_dir) / domain
```

esphome/components/image/__init__.py
jesserockz: The `download_image` function should take care of this by checking the file type either from the webserver Content-Type, or by inspecting the file after its downloaded and then check if cairosvg is installed.

esphome/components/image/__init__.py
```diff
@@ -74,6 +84,28 @@ def download_mdi(value):
     return value
 
 
+def download_image(value):
```
jesserockz, line 117: There is a lot of overlap with download_mdi here, they should probably be combined

esphome/external_files.py
```diff
@@ -0,0 +1,81 @@
...
+    parsed_url = urlparse(url)
+
+    # Get the path component
+    path = unquote(parsed_url.path)
+
+    # Extract file name and extension
+    file_name = os.path.basename(path)
+
+    file_base_name, file_extension = os.path.splitext(file_name)
+    return file_base_name, file_extension
```
jesserockz: This needs to handle when the filename is not directly in the url

jesserockz (changes requested):

esphome/external_files.py
landonr: > jesserockz

i don't understand. Do you mean we should download the file first before checking for info?

jesserockz: I just want to make sure that URLs like `https://example.com/image` which could legitimately be an image download redirect work. Right now if someone was to use this url, the code only looks for a filename and extension based on the end of the url which is hacky at best...
Most download servers will give something like `Content-Disposition: attachment; filename="filename.jpg"` header which should be used for the local filename. This is the filename the browser would save a file as in such a case.

landonr: I'm not getting a Content-Disposition back from home assistant when i try to load a media play image like the following
> http://172.16.1.67:8123/api/media_player_proxy/media_player.beam?token=f6a40dc8c712ca8c1f3a599ee15286283f197bd696f5f59b0d3481185fe14c4e&cache=b311e6b1cf141036

Also since .beam is the entity_id, it shows up like the file extension when splitting the URL

what do you think about downloading the file and checking its type, and deleting it if it's not correct?

jesserockz: Ah yes this is a very good example of what I meant. So yes, the file should be downloaded in the validator and actual content type checked. 
Another option is to allow the user to specify the format/type of the image as another config param?

landonr: I have it working for everything but svg images hidden behind a redirect. I could add `type` as a config param but wouldn't I need to list out every supported image type?

esphome/external_files.py
```diff
@@ -0,0 +1,145 @@
...
+                _LOGGER.debug(
+                    "has_remote_file_changed: File not modified since %s",
+                    local_modification_time_str,
+                )
+                return False
+            _LOGGER.debug("has_remote_file_changed: File modified")
+            return True
+        except requests.exceptions.RequestException as e:
+            raise cv.Invalid(
+                f"Could check if {url} has changed, please check if file exists "
```
jesserockz:
```suggestion
                f"Could not check if {url} has changed, please check if file exists "
```

jesserockz (changes requested):
Tested and working with pngs so far.
One minor text change below so far and will take another look very soon

jesserockz (changes requested):
Tested and working with pngs so far.
One minor text change below so far and will take another look very soon

esphome/external_files.py
```diff
@@ -0,0 +1,145 @@
...
+    _LOGGER.debug("get file info url %s", url)
+    r = requests.get(
+        url,
+        allow_redirects=True,
+        timeout=NETWORK_TIMEOUT,
+        headers={"User-agent": "esphome"},
+    )
+    if r.status_code != 200:
+        raise cv.Invalid(
+            f"Could check {url} info, check if file exists " f"({r.status_code}"
```
jesserockz: Another one here =)
```suggestion
            f"Could not check {url} info, check if file exists " f"({r.status_code}"
```

esphome/external_files.py
jesserockz: This is the `home.svg` download from the mdi repo
```py
>>> import mimetypes
>>> mimetypes.MimeTypes().guess_type("home.svg")
('image/svg+xml', None)
```

landonr: should I use mimetypes to guess the image types for all image types?

jesserockz:
@landonr I have pushed some kinda big changes that removed the content disposition and url filename guessing and just storing the file in any case with a hash of the url

The actual file type is determined when the code actually needs to read and decode the file just so it can follow a different path for svg types.

If you are happy/unhappy with the changes, please let me know =)

Jesse

landonr:
> @landonr I have pushed some kinda big changes that removed the content disposition and url filename guessing and just storing the file in any case with a hash of the url
> 
> The actual file type is determined when the code actually needs to read and decode the file just so it can follow a different path for svg types.
> 
> If you are happy/unhappy with the changes, please let me know =)
> 
> Jesse

great idea, I learned a lot from this PR! Thanks!

esphome/components/image/__init__.py
```diff
@@ -43,34 +52,74 @@
...
-        return value
-    url = f"https://raw.githubusercontent.com/Templarian/MaterialDesign/master/svg/{mdi_id}.svg"
-    _LOGGER.debug("Downloading %s MDI image from %s", mdi_id, url)
+def _compute_local_image_path(value: dict) -> Path:
+    url = value[CONF_URL]
+    h = hashlib.new("sha256")
+    h.update(url.encode())
+    key = h.hexdigest()[:8]
+    base_dir = external_files.compute_local_file_dir(DOMAIN)
+    return base_dir / key
```
landonr, line 76: I think it'd like to use this for fonts also, if that works I think it should go into external_files

jesserockz: The same code is used for packages and external_components to clone the git repo into a local hashed. 

For now I am going to merge this PR. If you want to move the code to be reusable, please make another PR to not hold this feature back from being released =)

jesserockz (approved):

alexd321:
Docs linked are wrong. Links to victron component
Should be https://github.com/esphome/esphome-docs/pull/3409

fawick:
Having seen this released just with 2023.11.04 just now, I wonder whether the web source images are fetched only once in the firmware build stage or whether the ESPHome-running device actually downloads them (possibly regularly) at runtime. 

I have a use case for the latter, which I currently handle with pr#4710.

landonr:
This is for images downloaded once on build. Looking forward to #4710

fawick:
Thank you for clarifying!

PR #5254, opened 2023-08-15
download font from url on build

landonr:

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3261

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
external_components:
  - source:
      type: local
      path: ../compile-online-image/esphome/components
    components: [
      font,
      external_files
    ]

font:
  - file: 
      url: "https://github.com/IdreesInc/Monocraft/releases/download/v3.0/Monocraft.ttf"
      # refresh: 10s
      type: web
    id: web_font
    size: 24
    glyphs: '^…*/|\$#<>!?"%()[]+=,-_.:°0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ abcdefghijklmnopqrstuvwxyz&êÊîÎâÂôÔéÉíÍáÁóÓčëËüÜïÏöÖøØṣṢñãõÑÃÕ''’'

  - file: 
      url: "https://github.com/IdreesInc/Monocraft/releases/download/v3.0/Monocraft.ttf"
      # refresh: 10s
      type: web
    id: web_font_medium
    size: 20
    glyphs: '^…*/|\$#<>!?"%()[]+=,-_.:°0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ abcdefghijklmnopqrstuvwxyz&êÊîÎâÂôÔéÉíÍáÁóÓčëËüÜïÏöÖøØṣṢñãõÑÃÕ''’'

  - file: "fonts/iosevka.ttf"
    id: local_font
    size: 16
    glyphs: '^…*/|\$#<>!?"%()[]+=,-_.:°0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ abcdefghijklmnopqrstuvwxyz&êÊîÎâÂôÔéÉíÍáÁóÓčëËüÜïÏöÖøØṣṢñãõÑÃÕ''’'
  - file: "https://github.com/landonr/homeThing/raw/main/fonts/materialdesignicons-webfont.ttf" 
    id: material_font_large
    size: 20
    glyphs: [
      '󰐊', # mdi-play
    ]
  - file: "gfonts://Roboto"
    id: roboto
    size: 20


spi:
  clk_pin: GPIO18
  mosi_pin: GPIO19

display:
  - platform: st7789v
    model: TTGO_TDisplay_135x240
    id: my_display
    cs_pin: GPIO5
    dc_pin: GPIO16
    reset_pin: GPIO23
    rotation: 90
    update_interval: 10s
    lambda: |-
      it.print(0, 0, id(web_font), "web font");
      it.print(0, 25, id(web_font_medium), "web font medium");
      it.print(0, 50, id(local_font), "local font");
      it.print(0, 75, id(material_font_large), "󰐊");
      it.print(0, 100, id(roboto), "gfont");
      return;
```

## Checklist:

image


esphome/components/font/__init__.py
```diff
@@ -109,9 +110,42 @@ def _compute_gfonts_local_path(value) -> Path:
...
+
+
+def get_file_type(file_name):
+    # Regular expression pattern to match the file extension
+    pattern = r"\.([^.]+)$"
+    match = re.search(pattern, file_name)
+    if match:
+        file_type = match.group(1)
+        return file_type
+    return None
```
guillempages: As in #5214, this could be combined into the get_filename_from_url method

esphome/components/font/__init__.py
```diff
@@ -109,9 +110,42 @@ def _compute_gfonts_local_path(value) -> Path:
     return _compute_local_font_dir(name) / "font.ttf"
 
 
+def get_filename_from_url(url):
+    # Regular expression pattern to match the file name at the end of the URL
+    pattern = r"/([^/]+)$"
+    match = re.search(pattern, url)
+    if match:
+        file_name = match.group(1)
```
guillempages: You might want to rename to filename for consistency with the method name
```suggestion
        filename = match.group(1)
```

esphome/components/font/__init__.py
```diff
@@ -109,9 +110,42 @@ def _compute_gfonts_local_path(value) -> Path:
...
+        file_name = re.sub(r"%20", "", file_name)  # Replace %20 with empty string
```
guillempages: "%20" is normally a space; maybe replace by "_" for better readability?

esphome/components/font/__init__.py
```diff
@@ -191,6 +225,35 @@ def download_gfonts(value):
 )
 
 
+def download_font(value):
```
guillempages: Maybe combine with `download_gfont`? (i.e. write the common functionality only once and keep the methods only for parsing the URLs/names)

esphome/components/font/__init__.py
```diff
@@ -207,6 +270,14 @@ def validate_file_shorthand(value):
             data[CONF_WEIGHT] = weight[1:]
         return FILE_SCHEMA(data)
 
+    if value.startswith("https"):
```
guillempages: unencrypted http is not fully dead yet ;-)
```suggestion
    if value.startswith("http://") or value.startswith("https://"):
```

esphome/components/font/__init__.py
```diff
@@ -109,9 +110,42 @@ def _compute_gfonts_local_path(value) -> Path:
...
+    # Regular expression pattern to match the file extension
+    pattern = r"\.([^.]+)$"
+    match = re.search(pattern, file_name)
+    if match:
+        file_type = match.group(1)
+        return file_type
+    return None
+
+
+def _compute_local_image_path(value) -> Path:
```
guillempages: Could this be combined with the gfont one?
Also, the hash calculation used for gfonts might also be useful here, to allow downloading diffenent fonts with the same name (but different URLs)

esphome/components/font/__init__.py
```diff
@@ -364,6 +436,9 @@ async def to_code(config):
     elif conf[CONF_TYPE] == TYPE_GFONTS:
         path = _compute_gfonts_local_path(conf)
         font = load_ttf_font(path, config[CONF_SIZE])
+    elif conf[CONF_TYPE] == TYPE_WEB:
+        path = _compute_local_image_path(conf).as_posix()
```
guillempages: Could the `as_posix()` call be done within the `compute_path` method? (this comment might become irrelevant if you do combine it with the `_compute_gfonts_local_path` method).

landonr:
thanks for the review @guillempages i'll do the changes this week. It might make sense to move out the `compute_local_path` stuff to a new file because it can be used between fonts and images.

I've also been meaning to add an option for `refresh: 1day` so images can be updated

esphome/components/font/__init__.py
landonr: good idea, I had to call it from the higher level config, but it looks a lot cleaner

esphome/components/font/__init__.py
landonr: done!

esphome/components/font/__init__.py
landonr: i renamed the method instead because file_name was used elsewhere already

esphome/components/font/__init__.py
landonr: I'll check back on this. I combined the methods but I might be able to fix the hashing to support different fonts with the same name

esphome/components/font/__init__.py
landonr: I removed this, but I haven't tested it yet

esphome/components/font/__init__.py
landonr: long live http. what do you suggest? I can look for a regex to match URLs if you think that works?

esphome/components/font/__init__.py
landonr: i don't really understand this comment, can you suggest the change? How would the method return 2 different options

landonr:
> thanks for the review @guillempages i'll do the changes this week. It might make sense to move out the `compute_local_path` stuff to a new file because it can be used between fonts and images.
> 
> I've also been meaning to add an option for `refresh: 1day` so images can be updated

@guillempages i updated the external_files stuff to a new component called `external_files`. I also added the `source_refresh` key, but made it default to `never`

esphome/components/font/__init__.py
guillempages: You can return a tuple:
```
return (file_name, file_type)
```

and when calling the method you can do
```
  filename, extension = get_file_from_url(url)
```
And you would fill both variables with a single call. Python magic ;-)

esphome/components/external_files/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    base_directory = Path(CORE.config_dir) / ".esphome" / domain / name
+    if not os.path.exists(base_directory):
+        os.makedirs(base_directory)
+
+    recent_directory = find_recent_directory(base_directory, refresh)
+    if recent_directory:
+        return recent_directory
+
+    current_time = int(datetime.now().timestamp())
+    sub_directory = Path(os.path.join(base_directory, str(current_time)))
```
guillempages: I don't think it's a good idea to hard-code the timestamp into the directory name. Both Linux and Windows filesystems do store metadata related to the file like when was it created; no need to make this explicit in the directory name.

esphome/components/external_files/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    ]
+
+    for sub_directory in subdirectories:
+        if is_directory_recent(os.path.join(base_directory, sub_directory), refresh):
+            return os.path.join(base_directory, sub_directory)
+
+    return None
+
+
+def compute_local_file_dir(name: str, domain: str, refresh: TimePeriodSeconds) -> Path:
```
guillempages: IMHO "refresh" would only be relevant when deciding whether to re-download or not, but the whole directory creation methods should be agnostic to the refresh time.

esphome/components/external_files/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+from datetime import datetime
+from esphome.core import CORE, TimePeriodSeconds
+
+CODEOWNERS = ["@landonr"]
+
+
+def is_directory_recent(directory_path: str, refresh: TimePeriodSeconds) -> bool:
+    if os.path.exists(directory_path):
+        current_time = datetime.now().timestamp()
+        dir_timestamp = int(os.path.basename(directory_path))
```
guillempages: You should read the timestamp from the file stats; not from the directory name.

esphome/components/external_files/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+        return current_time - dir_timestamp <= refresh.total_seconds
+    return False
+
+
+def find_recent_directory(base_directory: str, refresh: TimePeriodSeconds) -> Path:
```
guillempages: I'm not sure this method is needed at all (I know you are calling it; I mean I don't think this should be called ;-) )

esphome/components/external_files/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def compute_local_file_dir(name: str, domain: str, refresh: TimePeriodSeconds) -> Path:
+    base_directory = Path(CORE.config_dir) / ".esphome" / domain / name
+    if not os.path.exists(base_directory):
+        os.makedirs(base_directory)
+
+    recent_directory = find_recent_directory(base_directory, refresh)
+    if recent_directory:
+        return recent_directory
```
guillempages: The directory should be unrelated to the timestamp.

esphome/components/external_files/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+    if not os.path.exists(sub_directory):
+        os.makedirs(sub_directory)
+
+    return sub_directory
+
+
+def get_file_name_from_url(url):
+    # Regular expression pattern to match the file name at the end of the URL
+    pattern = r"/([^/]+)$"
```
guillempages: As mentioned before, you could combine "filename" and "extension" into a single regex, to optimize the code a bit. If you then `return (filename, extension)`, callers can get the split for free...

esphome/components/external_files/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+        file_name = re.sub(r"%20", "_", file_name)  # Replace %20 with '_'
+
+        # Remove the file extension
+        file_name = re.sub(r"\.[^.]+$", "", file_name)
+
+        return file_name
+    return None
+
+
+def get_file_type(file_name):
```
guillempages: ...And then you wouldn't need this method.

esphome/components/font/__init__.py
```diff
@@ -8,6 +8,7 @@
 import requests
 
 from esphome import core
+from esphome.components import external_files
```
guillempages: I'm not sure if this is actually a component, or would be a utility; but that's not for me to decide; that should be decided by the project owners (Jesse & Co)

esphome/components/font/__init__.py
```diff
@@ -142,25 +134,48 @@ def validate_weight_name(value):
...
-    url = f"https://fonts.googleapis.com/css2?family={value[CONF_FAMILY]}:wght@{wght}"
+def get_font_url(value):
+    if value[CONF_TYPE] == TYPE_GFONTS:
+        wght = value[CONF_WEIGHT]
+        return (
+            f"https://fonts.googleapis.com/css2?family={value[CONF_FAMILY]}:wght@{wght}"
+        )
+    if value[CONF_TYPE] == TYPE_WEB:
+        return value[CONF_URL]
+    return ""
```
guillempages: Maybe better return None, instead of an empty string?
```suggestion
    return None
```

esphome/components/font/__init__.py
```diff
@@ -142,25 +134,48 @@ def validate_weight_name(value):
...
-        raise cv.Invalid(
-            f"Could not download font for {name}, please check the fonts exists "
-            f"at google fonts ({e})"
+
+def get_font_name(value):
+    if value[CONF_TYPE] == TYPE_GFONTS:
+        return f"{value[CONF_FAMILY]}@{value[CONF_WEIGHT]}"
+    if value[CONF_TYPE] == TYPE_WEB:
+        return external_files.get_file_name_from_url(value[CONF_URL])
+    return ""
```
guillempages:
```suggestion
    return None
```

esphome/components/font/__init__.py
```diff
@@ -142,25 +134,48 @@ def validate_weight_name(value):
...
+
+
+def get_font_path(value):
+    if value[CONF_TYPE] == TYPE_GFONTS:
+        name = f"{value[CONF_FAMILY]}@{value[CONF_WEIGHT]}@{value[CONF_ITALIC]}@v1"
+        return (
+            Path(
+                external_files.compute_local_file_dir(name, DOMAIN, value[CONF_REFRESH])
```
guillempages: Previously, the directory contained a hashed version of the font name; now it contains the name in plaintext. I don't have a strong opinion either way, but I assume the previous implementation was done this way for a reason; maybe Jesse does have some input on that.

esphome/components/font/__init__.py
```diff
@@ -142,25 +134,48 @@ def validate_weight_name(value):
...
         )
+    if value[CONF_TYPE] == TYPE_WEB:
+        font_id = external_files.get_file_name_from_url(value[CONF_URL])
+        name = f"{font_id}@{value[CONF_WEIGHT]}@{value[CONF_ITALIC]}@v1"
+        file_name = font_id + "." + external_files.get_file_type(value[CONF_URL])
+        file_path = Path(
+            external_files.compute_local_file_dir(name, DOMAIN, value[CONF_REFRESH])
+        )
+        return file_path / file_name
+    return ""
```
guillempages:
```suggestion
    return None
```

esphome/components/font/__init__.py
```diff
@@ -142,25 +134,48 @@ def validate_weight_name(value):
...
+    if value[CONF_TYPE] == TYPE_GFONTS:
+        name = f"{value[CONF_FAMILY]}@{value[CONF_WEIGHT]}@{value[CONF_ITALIC]}@v1"
+        return (
+            Path(
+                external_files.compute_local_file_dir(name, DOMAIN, value[CONF_REFRESH])
+            )
+            / "font.ttf"
         )
+    if value[CONF_TYPE] == TYPE_WEB:
+        font_id = external_files.get_file_name_from_url(value[CONF_URL])
```
guillempages: You created the method "get_font_name", but are not using that?

esphome/components/font/__init__.py
```diff
@@ -173,23 +188,50 @@ def download_gfonts(value):
...
+
+
+def download_web_font(value):
+    name = get_font_name(value)
+    url = get_font_url(value)
+    path = get_font_path(value)
+    if path.is_file():
+        return value
+    try:
+        req = requests.get(url, timeout=30)
```
guillempages: Maybe create a constant for the timeout?

esphome/components/font/__init__.py
```diff
@@ -173,23 +188,50 @@ def download_gfonts(value):
...
+EXTERNAL_FONT_SCHEMA = cv.Schema(
     {
-        cv.Required(CONF_FAMILY): cv.string_strict,
         cv.Optional(CONF_WEIGHT, default="regular"): cv.Any(
             cv.int_, validate_weight_name
         ),
         cv.Optional(CONF_ITALIC, default=False): cv.boolean,
-    },
-    download_gfonts,
+        cv.Optional(CONF_REFRESH, default="never"): cv.All(
```
guillempages: Not sure about the "refresh" value. Maybe a more automatic system, using the ETAG header from the HTTP request would be better? If the server properly implements that, it would return an id on the etag header, that would indicate whether the file changed on the server or not. (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)

esphome/components/font/__init__.py
```diff
@@ -254,10 +308,13 @@ def _file_schema(value):
         cv.Optional(CONF_SIZE, default=20): cv.int_range(min=1),
         cv.GenerateID(CONF_RAW_DATA_ID): cv.declare_id(cg.uint8),
         cv.GenerateID(CONF_RAW_GLYPH_ID): cv.declare_id(GlyphData),
-    }
+    },
 )
 
-CONFIG_SCHEMA = cv.All(validate_pillow_installed, FONT_SCHEMA)
+CONFIG_SCHEMA = cv.All(
```
guillempages: Was this change necessary because of the CI formatting checks (black, pyflake,...)?

esphome/components/external_files/__init__.py
landonr: i thought about that but I wasn't sure if that was a solid solution. It's an easy switch if you think that's better!

esphome/components/font/__init__.py
landonr: interesting, i'll check that out! I made it default to never because it could return a new directory if the build took over a second. This should solve that

landonr:
thanks @guillempages i'll get to work on this later today or tomorrow

esphome/components/font/__init__.py
landonr: got it working, thanks for the suggestion!

esphome/components/font/__init__.py
landonr: I forgot to remove this, thanks!

esphome/components/font/__init__.py
landonr: how do you think we should keep track of the etag? In the file name or directory?

esphome/components/external_files/__init__.py
landonr, line 1: I moved this to a new file so we can reuse it for images

jesserockz: This is probably better off in a helper/util file, not a component file

landonr: I'm going to reuse it for external images, i'll move it out to a helper file

esphome/components/font/__init__.py
```diff
@@ -162,27 +179,58 @@ def download_gfonts(value):
...
+
+def download_web_font(value):
+    name = get_font_name(value)
+    url = get_font_url(value)
+    path = get_font_path(value)
+
+    if external_files.is_file_recent(
+        path, value[CONF_REFRESH]
+    ) or external_files.check_etag_equality(url, path):
+        return value
```
landonr: return the font if the file is recent, or the local etag matches the remote

landonr:
@jesserockz updated with changes from #5214

esphome/components/font/__init__.py
```diff
@@ -221,17 +277,23 @@ def validate_file_shorthand(value):
...
 
 def _file_schema(value):
     if isinstance(value, str):
         return validate_file_shorthand(value)
-    return TYPED_FILE_SCHEMA(value)
+    typed_schema = TYPED_FILE_SCHEMA(value)
+    if typed_schema[CONF_TYPE] == TYPE_WEB:
+        download_web_font(typed_schema)
+    elif typed_schema[CONF_TYPE] == TYPE_GFONTS:
+        download_gfont(typed_schema)
```
jesserockz: I think these should still be on the relevant SCHEMAs as extras...

jesserockz (changes requested):

esphome/components/font/__init__.py
landonr: Thanks! Updated, let me know what you think

esphome/components/font/__init__.py
```diff
@@ -161,26 +200,45 @@ def download_gfonts(value):
...
+        cv.Optional(CONF_DOWNLOAD): download_gfont,
+    }
+).extend(EXTERNAL_FONT_SCHEMA)
+
+WEB_FONT_SCHEMA = EXTERNAL_FONT_SCHEMA.extend(
+    {
+        cv.Required(CONF_URL): cv.string_strict,
+        cv.Optional(CONF_DOWNLOAD): download_web_font,
+    }
 )
```
jesserockz: No, it should be like this I think.
```suggestion
GFONTS_SCHEMA = cv.All(
    EXTERNAL_FONT_SCHEMA.extend(
        {
            cv.Required(CONF_FAMILY): cv.string_strict,
        }
    ),
    download_gfont,
)

WEB_FONT_SCHEMA = cv.All(
    EXTERNAL_FONT_SCHEMA.extend(
        {
            cv.Required(CONF_URL): cv.string_strict,
        }
    ),
    download_web_font,
)
```

jesserockz (changes requested):

esphome/components/font/__init__.py
landonr: nice I couldn't figure out how to get the cv.All to have a schema. I had to pass in the type because it wasn't available in the conf
Updated, hopefully it looks good!

jesserockz (dismissed):

jesserockz (dismissed):

clydebarrow (approved):
Tested the earlier font changes with this, all seems well.

jesserockz:
Thanks @clydebarrow. Im just fixing up the docs PR now

PR #5275, opened 2023-08-17
fix to PR # 3887 MQTT connection not using discovery: false

luka6000:

Setting
```
discovery: false
```
will make MQTT client go silent with any discovery info forced by PR3887. This will allow using external MQTT brokers like io.adafruit which enforces specific topic names

## Types of changes

- [ ] Other

https://github.com/esphome/issues/issues/4703

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
mqtt:
  broker: 192.168.0.99
  username: devel
  password: anypass
  discovery: false
  birth_message:
  will_message:
  log_topic:
```

## Checklist:

### DIFF
--- a/esphome/components/mqtt/mqtt_client.cpp
+++ b/esphome/components/mqtt/mqtt_client.cpp
@@ -66,25 +66,28 @@ void MQTTClientComponent::setup() {
   }
 #endif
 
-  this->subscribe(
-      "esphome/discover", [this](const std::string &topic, const std::string &payload) { this->send_device_info_(); },
-      2);
+  if (this->is_discovery_enabled()) {
+    this->subscribe(
+        "esphome/discover", [this](const std::string &topic, const std::string &payload) { this->send_device_info_(); },
+        2);
 
-  std::string topic = "esphome/ping/";
-  topic.append(App.get_name());
-  this->subscribe(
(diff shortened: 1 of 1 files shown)


esphome/components/mqtt/mqtt_client.cpp
```diff
@@ -85,42 +87,45 @@ void MQTTClientComponent::send_device_info_() {
...
   std::string topic = "esphome/discover/";
   topic.append(App.get_name());
-  this->publish_json(
-      topic,
-      [](JsonObject root) {
-        auto ip = network::get_ip_address();
-        root["ip"] = ip.str();
-        root["name"] = App.get_name();
+
+  if (this->is_discovery_enabled()) {
```
jesserockz: I feel this should just be an early return at the top of this function. Otherwise it still builds the topic string above for no reason.

luka6000: sure, you're right. Just pushed a change with this

luka6000: @jesserockz can you please review this change? MQTT client at current state doesn't work or is very hard to use with generaly available MQTT brokers

jesserockz (approved):

PR #5283, opened 2023-08-20
Fix Waveshare 7.5v2 epaper screens are always powered on

### LINKED ISSUE #4739: Some Waveshare ePaper Screens are always powered on
### The problem

The waveshare 7.5v2 displays are always on (not asleep) which is damaging the display over time. I've already lost a driver hat and a waveshare 7.5v2 screen running esphome with this epaper screen type.

If you use a "strong" light source such as a smartphone LED and point it directly at the display (~1cm) after the display has refreshed ~3 seconds ago, it will go dark at this point.

But it shouldn't, because the refresh of the epaper display has already been completed and the pixels shouldn't react to light after this step. I suspected that ESPhome would never put the screen into deep sleep mode. Waveshare warns of display damage if the display is left on for an extended period of time.

https://www.waveshare.com/wiki/7.5inch_e-Paper_HAT_(B)_Manual#ESP32.2F8266:~:text=Note%20that%20the%20screen%20cannot%20be%20powered%20on%20for%20a%20long%20time.%20When%20the%20screen%20is%20not%20refreshed%2C%20please%20set%20the%20screen%20to%20sleep%20mode%20or%20power%20off%20it.%20Otherwise%2C%20the%20screen%20will%20remain%20in%20a%20high%20voltage%20state%20for%20a%20long%20time%2C%20which%20will%20damage%20the%20e%2DPaper%20and%20cannot%20be%20repaired!

As you can see at https://esphome.io/api/waveshare__epaper_8cpp_source.html#l00111:~:text=%C2%A0void%20HOT%20WaveshareEPaper7P5InV2%3A%3Adisplay()%20%7B  the waveshare 7.5v2 display will not be powered off (command(0x02)) after refeshing the content.

The display is in power-on mode for the whole time. This will certainly damage the epaper display. I've added the necessary commands to test the deep sleep mode of the display and my smartphone LED check described above does not work anymore (= the screen does not go dark on the pointed spot anymore).

This should be fixed in the code.

```
    on_state: 
      then:
        - if:
            condition:
              - lambda: 'return !(id(display_update_in_progress) == 1);'
            then:
              - globals.set:
                  id: display_update_in_progress
                  value: "1"
              - logger.log: "Pre display update"
              - display.page.show: overview
              - lambda: |- 
                  id(epaper_display).initialize(); <<<<<<<<<<<<<<< must be called in this test code because the display could be powered off 
              - component.update: epaper_display
              - logger.log: "Post display call, should take 4 seconds now"
              - delay: 4s
              - logger.log: "delay over"
              - lambda: |- 
                  id(epaper_display).command(0x02); <<<<<<<<<<<<<<< will set the display to sleep.
              - logger.log: "After display update"
              - delay: 4s
              - globals.set:
                  id: display_update_in_progress
                  value: "0"
[...]
```

<= 2023.7.0

Home Assistant Add-on

2023.7.3

ESP32

### Board

Lolin D32 Pro

ESPHome display component

### Example YAML snippet

```yaml
spi:
  clk_pin: 18
  mosi_pin: 23
  miso_pin: 19


display:
  - platform: waveshare_epaper
    id: epaper_display
    reset_duration: 2ms
    setup_priority: 1000
    cs_pin: 5
    dc_pin: 0
    busy_pin: 15
    reset_pin: 2
    model: 7.50inv2
    update_interval: never 
    

    pages:
      - id: overview
        lambda: |-
          auto time = id(sntp_time).now();

          // Top status header
          it.print(6, 4, id(font_11pt), "Home Assistant");
          it.printf(95, 4, id(font_11pt), "(ESPhome version: %s | SSID: %s)",id(version_id).state.c_str(),id(ssid_id).state.c_str());
          it.print(680, 4, id(font_11pt), "ref.:");
          it.strftime(705, 4, id(font_11pt), "%d.%m.%Y %H:%M", id(sntp_time).now());
          it.line(2, 16, 798, 16);
          it.line(2, 17, 798, 17);

...

  - platform: homeassistant
    id: esp_waterme_s210000001_moisture
    name: "esp_waterme_s210000001_moisture"
    entity_id: sensor.esp_waterme_s210000001_moisture
    accuracy_decimals: 1
    on_value: 
      then:
         - component.update: epaper_display
```

_No response_

### Additional information

_No response_

phoenixswiss:

As I've stated on [#4739](https://github.com/esphome/issues/issues/4739) Waveshare 7.5v2 displays are always powered on (not asleep) which causes damage to the display over time. Waveshare warns of display damage if the display is left on for an extended period of time.

[Waveshare documentation](https://www.waveshare.com/wiki/7.5inch_e-Paper_HAT_(B)_Manual#ESP32.2F8266:~:text=Note%20that%20the%20screen%20cannot%20be%20powered%20on%20for%20a%20long%20time.%20When%20the%20screen%20is%20not%20refreshed%2C%20please%20set%20the%20screen%20to%20sleep%20mode%20or%20power%20off%20it.%20Otherwise%2C%20the%20screen%20will%20remain%20in%20a%20high%20voltage%20state%20for%20a%20long%20time%2C%20which%20will%20damage%20the%20e%2DPaper%20and%20cannot%20be%20repaired!)

~~I've created a new version of the 7.50inv2 display variant that fixes this problem.~~ I've updated the existing code of the 7.50inv2 display variant to fix this problem.

- Turn on display only during refresh
- Fixed the `wait_until_idle_()` function to work properly with Waveshare 7.5v2 screens (Warning: You also need to invert the busy pin in the yaml config!)
- Increased the timeout because the display is slow and needs at least 6 seconds to refresh

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [#4739](https://github.com/esphome/issues/issues/4739)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#[3143](https://github.com/esphome/esphome-docs/pull/3143)

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
display:
  - platform: waveshare_epaper
    id: epaper_display
    reset_duration: 2ms
    setup_priority: 1000
    cs_pin: 5
    dc_pin: 0
    busy_pin: 
      number: 15
      inverted: True
    reset_pin: 2
    model: 7.50inv2
    update_interval: never 
```

## Checklist:

### DIFF
--- a/esphome/components/waveshare_epaper/waveshare_epaper.cpp
+++ b/esphome/components/waveshare_epaper/waveshare_epaper.cpp
@@ -1561,66 +1561,7 @@ void WaveshareEPaper7P5In::dump_config() {
   LOG_PIN("  Busy Pin: ", this->busy_pin_);
   LOG_UPDATE_INTERVAL(this);
 }
-void WaveshareEPaper7P5InV2::initialize() {
-  // COMMAND POWER SETTING
-  this->command(0x01);
-  this->data(0x07);
-  this->data(0x07);
-  this->data(0x3f);
-  this->data(0x3f);
-  this->command(0x04);
-
-  delay(100);  // NOLINT
-  this->wait_until_idle_();
-  // COMMAND PANEL SETTING
-  this->command(0x00);
-  this->data(0x1F);
-
-  // COMMAND RESOLUTION SETTING
-  this->command(0x61);
-  this->data(0x03);
-  this->data(0x20);
-  this->data(0x01);
-  this->data(0xE0);
-  // COMMAND ...?
-  this->command(0x15);
-  this->data(0x00);
-  // COMMAND VCOM AND DATA INTERVAL SETTING
-  this->command(0x50);
-  this->data(0x10);
-  this->data(0x07);
-  // COMMAND TCON SETTING
-  this->command(0x60);
-  this->data(0x22);
-}
-void HOT WaveshareEPaper7P5InV2::display() {
-  uint32_t buf_len = this->get_buffer_length_();
-  // COMMAND DATA START TRANSMISSION NEW DATA
-  this->command(0x13);
-  delay(2);
-  for (uint32_t i = 0; i < buf_len; i++) {
-    this->data(~(this->buffer_[i]));
-  }
-
-  // COMMAND DISPLAY REFRESH
-  this->command(0x12);
-  delay(100);  // NOLINT
-  this->wait_until_idle_();
-}
-
-int WaveshareEPaper7P5InV2::get_width_internal() { return 800; }
-int WaveshareEPaper7P5InV2::get_height_internal() { return 480; }
-void WaveshareEPaper7P5InV2::dump_config() {
-  LOG_DISPLAY("", "Waveshare E-Paper", this);
-  ESP_LOGCONFIG(TAG, "  Model: 7.5inV2rev2");
-  LOG_PIN("  Reset Pin: ", this->reset_pin_);
-  LOG_PIN("  DC Pin: ", this->dc_pin_);
-  LOG_PIN("  Busy Pin: ", this->busy_pin_);
-  LOG_UPDATE_INTERVAL(this);
-}
-
-/* 7.50inV2Fixed */
-bool WaveshareEPaper7P5InV2Fixed::wait_until_idle_() {
+bool WaveshareEPaper7P5InV2::wait_until_idle_() {
   if (this->busy_pin_ == nullptr) {
     return true;
   }
@@ -1628,7 +1569,7 @@ bool WaveshareEPaper7P5InV2Fixed::wait_until_idle_() {
   const uint32_t start = millis();
   while (this->busy_pin_->digital_read()) {
     this->command(0x71);
-    if (millis() - start > 10000) {
+    if (millis() - start > this->idle_timeout_()) {
       ESP_LOGE(TAG, "Timeout while displaying image!");
       return false;
     }
@@ -1637,8 +1578,7 @@ bool WaveshareEPaper7P5InV2Fixed::wait_until_idle_() {
   }
   return true;
 }
-void WaveshareEPaper7P5InV2Fixed::initialize() {
-
+void WaveshareEPaper7P5InV2::initialize() {
   // COMMAND POWER SETTING
   this->command(0x01);
   this->data(0x07);
@@ -1647,11 +1587,10 @@ void WaveshareEPaper7P5InV2Fixed::initialize() {
   this->data(0x3f);
 
   //we don't want the display to be powered at this point
-  //this->command(0x04); 
 
   delay(100);  // NOLINT
   this->wait_until_idle_();
-  
+
   // COMMAND VCOM AND DATA INTERVAL SETTING
   this->command(0x50);
   this->data(0x10);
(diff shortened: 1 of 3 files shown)


jesserockz:
Why not just apply the changes to the existing model instead of creating a whole new one?

phoenixswiss:
@jesserockz  I'm a big fan of not changing the functionality of an existing feature/profile/variant. With the wait_until_idle_() function broken in the 7.50inv2 variant, we have had to solve this problem outside of the component until now. It is up to you, I can modify the code and change the 7.50inv2 display variant in my PR.

jesserockz:
I just feel that we keep slapping new models, whereas the model is supposed to match the hardware

https://phdcomics.com/comics.php?f=1531

phoenixswiss:
@jesserockz I've changed the code as you suggested and edited the docs accordingly.

phoenixswiss:
@jesserockz Is something missing to accept this pull request?

jesserockz:
> @jesserockz Is something missing to accept this pull request?

Not from you, just my free time to come back to it =)

esphome/components/waveshare_epaper/waveshare_epaper.cpp
```diff
@@ -1582,34 +1610,54 @@ void WaveshareEPaper7P5InV2::initialize() {
...
+
+  ESP_LOGI(TAG, "Before command(0x02) (>> power off)");
+  this->command(0x02);
+  this->wait_until_idle_();
+  ESP_LOGI(TAG, "After command(0x02) (>> power off)");
 }
 
 int WaveshareEPaper7P5InV2::get_width_internal() { return 800; }
 int WaveshareEPaper7P5InV2::get_height_internal() { return 480; }
+uint32_t WaveshareEPaper7P5InV2::idle_timeout_() { return 10000; }
```
jesserockz, line 1660: 10 seconds wait time in the loop that uses this value will trigger the watchdog timer causing a full reset of the ESP...

phoenixswiss:
@jesserockz nope, we feed the watchdog in WaveshareEPaper7P5InV2::wait_until_idle_() on line 1576 =) we must set the timeout value to such a high number because the display refresh cycle is very slow and can take up to 9sec.

jesserockz:
Lol, I missed that.

jesserockz:
These screens really need to be made into state machines so the rest of the esphome application can get on with business during those 9 seconds. But that's separate from this pr

esphome/components/waveshare_epaper/waveshare_epaper.cpp
```diff
@@ -1582,34 +1610,54 @@ void WaveshareEPaper7P5InV2::initialize() {
...
+
   // COMMAND DISPLAY REFRESH
   this->command(0x12);
   delay(100);  // NOLINT
   this->wait_until_idle_();
+
+  ESP_LOGI(TAG, "Before command(0x02) (>> power off)");
+  this->command(0x02);
+  this->wait_until_idle_();
+  ESP_LOGI(TAG, "After command(0x02) (>> power off)");
```
kbx81: Hmm, is there a reason for using "information" level? Feels like this should be "verbose" or even "very verbose".

kbx81:
```suggestion
  ESP_LOGV(TAG, "Before command(0x02) (>> power off)");
  this->command(0x02);
  this->wait_until_idle_();
  ESP_LOGV(TAG, "After command(0x02) (>> power off)");
```

phoenixswiss: @kbx81 Because turning off the display is absolutely key to prevent permanent HW damage and the hole point of this PR

@jesserockz Is there any chance to get this PR approved soon? These displays are getting more damaged every day the old code is running...

kbx81: I completely understand the importance of the function and the point of the PR; that said, if the display is instructed to power off, I should be able to assume it is powered/powering off. Messages in the log indicating this is happening are extraneous under normal circumstances.

phoenixswiss: Ok, I will change the log severity to verbose.

kbx81: Sounds good and thank you for working on this! 😄

phoenixswiss: @kbx81 @jesserockz  I've changed the log severity.

kbx81 (approved):

esphome/components/waveshare_epaper/waveshare_epaper.cpp
kbx81: I see, thanks again!

PR #5287, opened 2023-08-21
Add config to allow suppression of warnings about use of strapping pins

### LINKED ISSUE #2168: Suppress Strapping Pin Warning

I'd like a configuration option on a pin to disable compiler warnings about strapping pins.

When compiling code that uses strapping pins a compiler warning is issued

```
WARNING GPIO8 is a Strapping PIN and should be avoided.
Attaching external pullup/down resistors to strapping pins can cause unexpected failures.
```

**Additional context**

As noted in the FAQ "It’s recommended to avoid them unless you have a pressing need to use them and you have reviewed the expected boot voltage levels of these pins from the ESP datasheet."

This is the case for me. This is a tuya device which I've done a chip replacement on, replacing the tuya chip with an esp-c3-12f. I can't change which pins are used. And I've tested and confirmed that the boot voltage on these pins does not cause an issue. So for this device I'd like to suppress that warning. Having it as a config option is valuable because it would warn me if I inadvertently use one of these pins in a future project. But once I've confirmed that it's not an issue on a particular device, I'd like to disable the warning.

clydebarrow:

This change implements a new config option for GPIO pins `ignore_strapping_warning:` which if set to `true` will suppress warnings for that pin about it being also a strapping pin. 

Although some care has to be taken with strapping pins, there are numerous situations in which they can safely be used for I/O and this allows a user to prevent spurious warnings.

Some refactoring of code has also been done to centralise the check for and warning about strapping pins, eliminating duplicated code.

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2168
## Types of changes

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3146

## Test Environment

- [x] ESP32
- [x] ESP32 IDF

```yaml
# Example config.yaml
spi:
  mosi_pin:
    number: GPIO2

```

## Checklist:


jesserockz (commented):
I'm not sure I like the idea of hiding the warnings, they are not too obtrusive, but are very informative. Many people copy their config from online and once this is available, everyone will slap this line in there as well and then people will report more issues without any warnings...

Just my quick thoughts after seeing this...

clydebarrow:
If they've seen the warning and deliberately chosen to hide it, surely it's now their problem? There are use cases where the strapping pins have to be used (safely) and having warnings that can't be turned off is a Bad Idea because it encourages the attitude that warnings don't matter.

I do find the warnings obtrusive - and while they are informative, they don't have to be forever informative when they no longer matter. IMHO.

jesserockz:
> surely it's now their problem

Haha you think too much of some people. Even **with** the warning they come for help.

I get what you mean about warnings becoming something that is normal to ignore.

Ok, I will accept this change, but I also would like the config validation to fail if someone sets it to true on a non-strapping pin to stop people blanket putting it everywhere... (I have not thought if this will be easy or not...)

kbx81:
> Many people copy their config from online and once this is available, everyone will slap this line in there as well and then people will report more issues without any warnings...

This is precisely why I worry about this change, myself.

Experienced users are aware of the concept of "normalizing warnings"--which, make no mistake, isn't a good thing--but inexperienced users who don't get a warning because they copied & pasted some bits of config from some social site are worse off because they'll simply never see the warning. Personally, between normalizing warnings and just not having a warning because of "user error", I think not seeing the warning is worse because it degrades the experience for users who are new to the platform (ESPHome, Espressif hardware, electronics or otherwise).

> I also would like the config validation to fail if someone sets it to true on a non-strapping pin to stop people blanket putting it everywhere... (I have not thought if this will be easy or not...)

Solid agree -- it should only be a valid option for strapping pins, failing validation otherwise. This should help to somewhat mitigate the issue of too much copy/pasting.

clydebarrow:
> would like the config validation to fail if someone sets it to true on a non-strapping pin

Great idea. Consider it done.

clydebarrow:
I also removed the note in the warning about being able to use the strapping: option. This could go in the linked FAQ page - that way the information that the warning can be suppressed can be delivered with proper context.

esphome/const.py
```diff
@@ -697,6 +697,7 @@
 CONF_STOP = "stop"
 CONF_STOP_ACTION = "stop_action"
 CONF_STORE_BASELINE = "store_baseline"
+CONF_STRAPPING = "strapping"
```
jesserockz: I wonder if we should go as far as to making it like this?
```suggestion
CONF_IGNORE_STRAPPING_WARNING = "ignore_strapping_warning"
```

jesserockz (approved):

PR #5288, opened 2023-08-21
Native SPI RGB LED component

clydebarrow:

## Types of changes

- [ ] Other

This new component supports the APA102 and similar SPI addressable RGB LEDs. While there is support for these LEDs in other components, these rely on external libraries supporting multiple different chip types and don't work with ESP-IDF.

This component supports only the SPI (two wire) LED interface and leverages the existing ESPHome `spi` component so works on both Arduino and ESP-IDF platforms, is lightweight and performant. It has only been tested on ESP32 but should work with any platform supported by ESPHome that implements SPI.

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3147
## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
    spi:
      mosi_pin: GPIO3
      clk_pin: GPIO9

    light:
      - platform: apa102
        num_leds: 30
        color_correct: [80%, 60%, 100%]
        id: rgb_led
        name: "RGB LED Strip"
        data_rate: 1MHz
```

## Checklist:


jesserockz:
This is kinda a duplicate of #5101 which does need work to use SPIDevice like this PR does, but implements a generic SPI led strip driver which would be preferred over having multiple specific ones like this PR implements. Perhaps somehow a merge of the two PRs is in order?

clydebarrow:
> generic SPI led strip driver which would be preferred over having multiple specific ones like this PR implements

Well I don't have any of the other LEDs on hand to test with, however looking at the datasheet for the SK9822 and the HD107 it appears the code I already wrote will work with them, as they have the identical frame format and BGR order. So maybe all that's needed is a name change (maybe spi_rgb_led) and some extra documentation. 

I don't want to make it some kind of umbrella component that tries to do too much. Keeping it simple makes keeping it working easy. IOW I would resist adding the one-wire LED protocol support which requires quite different code.

clydebarrow:
The APA107 is a 1-wire chip, so not SPI compatible. The HD108 is SPI but has 16 bit data, so that's a possibility to add, but I'd prefer to do that later when I have some to test with. They're new and more expensive so probably not widely in use yet.

So right now it seems the APA102, APA102C, HD107 and SK9822 will work with my code and they are the main SPI type RGB leds available.

kbx81:
> The APA107 is a 1-wire chip

FWIW, I think the APA107 is also a 2-wire interface with clock and data lines -- at least, everything I'm finding on it right now seems to indicate this. 😇

clydebarrow:
I found two conflicting datasheets for APA107. Maybe there are two different chips from different manufacturers.

https://www.rose-lighting.com/wp-content/uploads/sites/53/2019/10/apa107-LED.pdf

https://www.superlightingled.com/PDF/APA107.pdf

clydebarrow:
Anyway, the 2 wire APA107 should work just fine with this component. I renamed it to `spi_rgb_led` and updated the docs to reflect the fact it works with chips other than APA102.

jesserockz:
`spi_led_strip` would be better to match `esp32_rmt_led_strip` and `rp2040_pio_led_strip`

esphome/components/spi_led_strip/spi_led_strip.h
```diff
@@ -0,0 +1,91 @@
+#pragma once
+
+#include "esphome/core/component.h"
```
kbx81, line 3: I believe you need to add `defines.h` to ensure `SPIRGBLED_DATA_RATE` is picked up. [See also.](https://github.com/esphome/esphome/pull/5344)

```suggestion
#include "esphome/core/component.h"
#include "esphome/core/defines.h"
```

clydebarrow: Is `defines.h` something new? Because it worked as it was.

But now that you have merged the new SPI PR, that won't be necessary. I'll update it to dynamically set the rate rather than using the #define kludge.

kbx81: > Is `defines.h` something new? Because it worked as it was.

It'll likely work because many other components include it; it's where any defines added to the project are written by `cg.add_define`. If by some chance nothing includes it, the macro wouldn't work. 😇

> But now that you have merged the new SPI PR, that won't be necessary. I'll update it to dynamically set the rate rather than using the #define kludge.

Sounds great! I'll leave you to it, then... 😁

kbx81: Also, feel free to nuke my last commits -- was just trying to resolve a merge conflict and I fudged up the codeowners. 😅

clydebarrow: Ok pending the CI run, should be good to go. Tested here with the changes.

kbx81: Awesome! I'll give it another test on my FunHouse board 😄

kbx81 (approved):

PR #5291, opened 2023-08-22
Use gzip compression for the web server component's static resources

EternityForest:

Compresses static resources in the web server, encoding with gzip in python's stdlib, and decoding client side using the content-encoding header.

On my computer, it brings the size of the V2 js down to 11320 bytes, from 33.2kb currently.

As far as I know, all common browsers have supported this for 15+ years, and even things like Lynx now have support,
so I do not attempt to inspect request headers, or provide any options to control the behavior, I just assume that it is basically universal.

Edit for transparency: Original PR had one import in the wrong place that my IDE missed, failing pylint, I used force-push to keep history neat.

## Types of changes

- [ ] Other

I don't believe there is an issue in any repo I can see, but I do see compression mentioned on the roadmap here: https://github.com/esphome/esphome-webserver

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

Note: Does not change anything a user should notice.

```yaml
# Example config.yaml
web_server:
  port: 80
  js_include: "./www.js"
  js_url: ""
  version: 2
```

## Checklist:

Nothing user exposed changed, but perhaps a note should be added somewhere, so people are aware that
serving locally is more practical now?

### DIFF
--- a/esphome/components/web_server/__init__.py
+++ b/esphome/components/web_server/__init__.py
@@ -110,7 +110,9 @@ def build_index_html(config) -> str:
     return html
 
 
-def add_resource_as_progmem(resource_name: str, content: str, compress: bool = True) -> None:
+def add_resource_as_progmem(
+    resource_name: str, content: str, compress: bool = True
+) -> None:
     """Add a resource to progmem."""
     content_encoded = content.encode("utf-8")
     if compress:
--- a/esphome/components/web_server/__init__.py
+++ b/esphome/components/web_server/__init__.py
@@ -1,3 +1,4 @@
+import gzip
 import esphome.codegen as cg
 import esphome.config_validation as cv
 from esphome.components import web_server_base
@@ -112,6 +113,7 @@ def build_index_html(config) -> str:
 def add_resource_as_progmem(resource_name: str, content: str) -> None:
     """Add a resource to progmem."""
     content_encoded = content.encode("utf-8")
+    content_encoded = gzip.compress(content_encoded)
     content_encoded_size = len(content_encoded)
     bytes_as_int = ", ".join(str(x) for x in content_encoded)
     uint8_t = f"const uint8_t ESPHOME_WEBSERVER_{resource_name}[{content_encoded_size}] PROGMEM = {{{bytes_as_int}}}"
--- a/esphome/components/web_server/__init__.py
+++ b/esphome/components/web_server/__init__.py
@@ -110,10 +110,11 @@ def build_index_html(config) -> str:
     return html
 
 
-def add_resource_as_progmem(resource_name: str, content: str) -> None:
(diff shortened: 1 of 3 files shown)


bdraco:
I had this in the original PR but removed it since it made the index page larger

EternityForest:
@bdraco Didn't know there was an original PR? Did it just get closed?

I'm getting 158 uncompressed vs 142 compressed,  when requesting the HTML with requests(Easier than hacking a direct readout during compile), then manually calling gzip.compresss.

Am I calculating wrong?  Or is the problem just that the actual adding of the gz header itself was taking more code space?  

Saving 16 bytes isn't quite enough to offset that I wouldn't imagine, so should I just special case the index HTML and serve it uncompressed?

If you use the cloud version the longer https://oi.esphome.io/v2/www.js URL adds a few more bytes, and the V1 URLs re even longer,  so maybe with V1 you'd still be saving bytes compressed, and it's best to keep it simple and compress all resources?

bdraco:
It was merged as https://github.com/esphome/esphome/pull/4944

bdraco:
https://github.com/esphome/esphome/pull/4944/commits/633bd883f49e60a07f3314cffd6c7f75a67bc35d
https://github.com/esphome/esphome/pull/4944/commits/59236022525289d554022318ffbdd5e1c4b81f60

These commits removed the gzip support after it was added earlier in the PR

bdraco:
It probably doesn't matter enough either way as most people are going to use the cloud assets anyways.

bdraco:
It will increase the latency just a bit on the client to uncompress the index page though.

Compressing the css and js makes complete sense to me.

I'd either add a comment so we at least know why we did it later that we are compressing the index for consistency and not for performance reasons ...or not compress the index

EternityForest:
Allright, will do, I'll wait to see if anyone else wants to weigh in on it... I think I'm going to put the computer away and head out soon after the embarrassing fight I just had with the CI bots on the other PR!

What's the ESPHome style for handling that kind of minor edits? Ammend/squash as you go or keep history till the end?

bdraco:
It will get squashed at the end

EternityForest:
Alright, nobody else has chimed in, so I just went with the conservative approach and left off gzip for the index.

bdraco:
Not in a position to test at the moment, but it LGTM 👍

EternityForest:
Further obvious n00b question since I'm not familiar with the DevOps process here:  Should I be keeping this branch up to date, or leaving it so it stays in one state till everything is reviewed, and then fixing any conflicts after?

bdraco:
Pleas keep it up to date if conflict occur.  PRs in a conflict state are usually skipped during reviews

jesserockz (approved):

PR #5293, opened 2023-08-22
Redefine tag according to the selected framework

andrewjswan:
Redefine the output of the tag (esp-arduino or esp_idf) in the log, according to the selected framework.

If there are different ESPs with different frameworks, there is confusion when viewing the log, it is better to specify a tag (esp-arduino or esp_idf), depending on the chosen framework.

## Types of changes

- [X] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/core/log.cpp
+++ b/esphome/core/log.cpp
@@ -52,8 +52,11 @@ int HOT esp_idf_log_vprintf_(const char *format, va_list args) {  // NOLINT
   auto *log = logger::global_logger;
   if (log == nullptr)
     return 0;
-
+#if defined(USE_ESP32_FRAMEWORK_ARDUINO)
+  log->log_vprintf_(ESPHOME_LOG_LEVEL, "esp-arduino", 0, format, args);
+#else
   log->log_vprintf_(ESPHOME_LOG_LEVEL, "esp-idf", 0, format, args);
+#endif
 #endif
   return 0;
 }


HeMan (approved):

andrewjswan:
@HeMan all work is finished at the time of publication of the PR, no more changes are expected.

HeMan (approved):

jesserockz:
This function wraps any `printf` calls from within esp-idf. Hence why the tag is esp-idf...
When using Arduino, the printf are still coming from esp-idf...

andrewjswan:
> This function wraps any `printf` calls from within esp-idf. Hence why the tag is esp-idf... When using Arduino, the printf are still coming from esp-idf...

Look at conditional compilation, this function works for IDF, and for Esp32 with Arduino framework, but it always writes esp_idf, when parsing the logs this leads to misunderstandings.

jesserockz:
You are not understanding the issue, the logs that use `printf(...)` come from esp-idf code, whether you are using arduino or esp-idf. 
If you saw a log that has the tag `esp-arduino` then you would expect to find that log source code in the https://github.com/espressif/arduino-esp32, but it wont be because it still comes from https://github.com/espressif/esp-idf which arduino has compiled in.

The tag points to the source of the log message, which will always be esp-idf code, not the arduino code

andrewjswan:
Then why is this change, working?

jesserockz:
define working?

kbx81:
Arduino is a layer of code that sits on top of IDF.

When you build an ESPHome config using the Arduino framework, IDF code is still included because Arduino depends on this (to some extent). By building your ESPHome config with IDF, you're simply excluding the Arduino layer.

Even if you've built your config with the Arduino framework, IDF functions can and do still log messages from "underneath" Arduino. What Jesse is trying to call out is that the change you're suggesting here obfuscates that which results in misleading messages in the logs.

The tag in the logs needs to reflect what code specifically logs the message, not what framework was chosen at build time.

andrewjswan:
@kbx81 Then this tag should be removed completely, it makes only confusion when in the Arduino log, writes IDF. This is not correct.

kbx81:
Removing the tag also creates confusion because then it's impossible to determine where the message comes from.

I think understanding the platform architecture(s) here is more important than how a message is tagged in the logs. You can't effectively troubleshoot if you don't understand the platform architecture.

andrewjswan:
esp-idf, does not give an understanding of where this output was called into the log. The problem here is not in understanding, but in the log, because logs are posted on forums, chats, and the config is not always, so there should be either an unambiguous tag type esphome, or a tag displaying the framework.

andrewjswan:
And once again. Since we are in this Define (esp_32_arduino or esp_idf) then the log output function will be called in any case, only in the case of esp_32_arduino the corresponding tag will be output. That's all.

jesserockz:
@andrewjswan Can you please post an example of the log you are seeing?

andrewjswan:
Only screenshot ... ESPHome, Arduino Framework
IMG_20230906_080515_512

jesserockz:
Ok thanks for that. 

So do you understand what a TAG is for in the ESPHome logs? It tells you the source of the log message, in this case there is `wifi` and `esp-idf` this tells me if I want to find the code that logged the message, I should look at the wifi component, or in the esp-idf codebase respectively. 

If we change the tag from `esp-idf` to `esp-arduino`, that would point a developer to look at the arduino codebase to find the source of the log, very important if it was an error, not just info/debug like the ones above. Except that you would do a global search for the log pattern and you would not find it anywhere and then wonder "why can't I find it, it was logged from esp-arduino..."


Another way to read it:

The function `esp_idf_log_vprintf_` is used in `logger.cpp:262` to set a callback for any logs to be printed -> `esp_log_set_vprintf(esp_idf_log_vprintf_);`.

This `esp_log_set_vprintf()` is inside `esp-idf` and only `esp-idf` it has absolutely nothing to do with Arduino. Even if you choose to use Arduino as the framework, there is logging done from `esp-idf` as evidenced in your screenshot above.

andrewjswan:
Here was a problem that was tried to solve, and everyone thought that the project was built with IDF framework, and could not understand what the problem was, but it turned out that the project was built on Arduino Framework, and as soon as it was found out, everything was solved.
IMG_20230906_101252_125

PR #5296, opened 2023-08-22
Make string globals persist-able using fixed size allocations

EternityForest:

This PR enables the restore feature for global variables of type std::string, by adding a new class for the special case.

It allocates a user-configurable fixed-size preference object(Defaulting to 63 characters), with a single length byte, for compatibility with the existing preference backends. Strings too big to fit are simply ignored and not saved.

Edit:  I *think* I have my IDE set up for proper linting!

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3151

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

This entry adds text to the variable at every boot,  so you can confirm that it works in the web server.
```yaml
esphome:
  name: test
  on_boot:
    - lambda: 'id(my_global_string2) = ("Blah");'


text_sensor:
  - platform: template
    lambda: "return id(my_global_string2);"
    id: my_global_string_tester

globals:
  - id: my_global_string2
    type: std::string
    restore_value: yes
    initial_value: '"DefaultVal"'
    max_restore_data_length: 22

```

## Checklist:

### DIFF
--- a/esphome/components/globals/globals_component.h
+++ b/esphome/components/globals/globals_component.h
@@ -65,6 +65,64 @@ template<typename T> class RestoringGlobalsComponent : public Component {
   ESPPreferenceObject rtc_;
 };
 
+// Use with string or subclasses of strings
+template<typename T, uint8_t SZ> class RestoringGlobalStringComponent : public Component {
+ public:
+  using value_type = T;
+  explicit RestoringGlobalStringComponent() = default;
+  explicit RestoringGlobalStringComponent(T initial_value) { this->value_ = initial_value; }
+  explicit RestoringGlobalStringComponent(
+      std::array<typename std::remove_extent<T>::type, std::extent<T>::value> initial_value) {
+    memcpy(this->value_, initial_value.data(), sizeof(T));
+  }
+
+  T &value() { return this->value_; }
+
+  void setup() override {
+    char temp[SZ];
+    this->rtc_ = global_preferences->make_preference<uint8_t[SZ]>(1944399030U ^ this->name_hash_);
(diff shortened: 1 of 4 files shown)


EternityForest:
Working on the text input component here has revealed a pretty big issue i didn't really pay attention to: Whether then max size limit config option should include the length byte or not. Off by ones are surprisingly confusing!

I'm starting to think maybe it's best to just hide that one byte, and have the number the user sees be the actual max usable length, since ESP already has tons of hidden NVS overhead.

esphome/components/globals/globals_component.h
```diff
@@ -65,6 +65,64 @@ template<typename T> class RestoringGlobalsComponent : public Component {
...
+  explicit RestoringGlobalStringComponent(
+      std::array<typename std::remove_extent<T>::type, std::extent<T>::value> initial_value) {
+    memcpy(this->value_, initial_value.data(), sizeof(T));
+  }
+
+  T &value() { return this->value_; }
+
+  void setup() override {
+    char temp[SZ];
+    this->rtc_ = global_preferences->make_preference<uint8_t[SZ]>(1944399030U ^ this->name_hash_);
```
jesserockz, line 83: This should also hash in the configured length to invalidate old data if you change the length between firmwares.

EternityForest:
I thought the prefs library already had length change invalidation at a
lower level?

jesserockz (approved):
Ah yes thanks!

PR #5302, opened 2023-08-24
Hardware PWM for ESP8266 using I2S

matemaciek:
# Hardware PWM for ESP8266 using I2S

ESP8266 does not have a hardware PWM, so let's add one. We can (ab)use I2S DMA for this.

Disadvantages:
- Bound to Rx pin

Advantages:
- Stable, thanks to not relying on CPU
- Works for high frequencies (in theory: up to some MHz)
- High resolution for low and medium frequencies

## This is very much work in progress

Code still needs more testing and cleanup, but it already works. Comments and sugestions are more than welcome (-: Especially usefull would be if someone wants to check behaviour on highest frequencies, my tools are not advanced enough.

Short term plans:
- Disable clock pins and make sure they are usable for other purposes
- Theoretical calculation of frequency precision and duty cycle resolution
- Documentation
- Tests
- Extract code to new Arduino library

## Types of changes

## Test Environment

- [x] ESP8266

```yaml
# Example config.yaml
output:
  - platform: esp8266_hw_pwm
    frequency: 108kHz
    id: switch_pwm

switch:
  - platform: output 
    name: 'Flowing'
    id: buzzer
    output: switch_pwm
    on_turn_on:
      then:
        - output.turn_on: switch_pwm
        - output.set_level:
            id: switch_pwm
            level: "50%"
```

## Checklist:

### DIFF
--- a/esphome/components/esp8266_hw_pwm/__init__.py
+++ b/esphome/components/esp8266_hw_pwm/__init__.py
@@ -1,0 +1,1 @@
+CODEOWNERS = ["@matemaciek"]
--- a/esphome/components/esp8266_hw_pwm/esp8266_hw_pwm.cpp
+++ b/esphome/components/esp8266_hw_pwm/esp8266_hw_pwm.cpp
@@ -1,0 +1,37 @@
+#ifdef USE_ESP8266
+
+#include "esp8266_hw_pwm.h"
+#include "esphome/core/macros.h"
+#include "esphome/core/defines.h"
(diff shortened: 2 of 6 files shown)


max246:
Since this module can be limited to be used only for ESP8266, we can just rename to from `esp8266_hw_pwm `to `hardware_pwm`
It seems like you can do the same hack with a normal esp32, the option to enable for esp32 can be added in the future.

PR #5298, opened 2023-08-23
Update PSRAM config params for IDF4+

kbx81:
# What does this implement/fix?

Minor change to update IDF configuration params when PSRAM is enabled. Based on [this](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-reference/kconfig.html#config-spiram).

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [x] ESP32-S2 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #5311, opened 2023-08-29
Refactor SPI code; Add ESP-IDF hardware support

### LINKED ISSUE #4827: No hardware SPI support for ESP-IDF
### The problem

The SPI component supports hardware SPI for Arduino, but not ESP-IDF. The software implementation that it falls back to has an apparent maximum speed of around 600kHz. The Arduino hw implementation has no problem with high speeds as expected.

1. The fact that a software implementation is always used for ESP-IDF should be documented;
2. The limitations of the sw implementation should be documented;
3. A warning should be issued if a clock rate greater than available is requested;
4. Hardware support on ESP-IDF should be implemented.

I am exploring option 4.

Example spi implementation with 4MHz clock requested:

```
esp32:
  board: esp32dev
  framework:
    type: esp-idf

spi:
  mosi_pin:
    number: GPIO26
  clk_pin:
    number: GPIO25

light:
  - platform: spi_rgb_led
    num_leds: 1
    color_correct: [80%, 60%, 100%]
    id: rgb_led
    name: "RGB LED"
    data_rate: 4MHz
```

If built for Arduino the results are as expected (clean 4MHz clock). On ESP-IDF the clock looks like this - a maximum speed of 609kHz, but with some clock cycles stretched (probably due to interrupts delaying the sw spi code.)

spi_clock

2023.9.0-dev

pip

_No response_

ESP32-IDF

### Board

tinypico

spi

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

clydebarrow:

## Types of changes

- [x] Other

This is a major rewrite of the SPI code - but no breaking API changes - that makes it easy to add support
for different platforms, and then adds support for hardware SPI on the ESP-IDF platform.

It has been tested on ESP32 and ESP8266, both hardware and software SPI, in multiple modes and
with multiple devices. It has also been tested on RP2040, but works only in software mode (as does the current SPI implementation, so no loss of functionality.) I expect to get HW SPI on RP2040 working once I have some debug capabilities.

There is also a new component suitable for SPI testing, and quick implementation of access to SPI peripherals not currently supported. This should lessen the need in the future for new components to
support specific peripherals as they can be implemented with a few lines of YAML.

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4827

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3156

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040

```yaml
# Example config.yaml
    spi:
        clk_pin: GPIO14
        mosi_pin: GPIO27
        miso_pin: GPIO26

    spi_device:
        id: spidev
        data_rate: 2MHz
        mode: 3
        bit_order: lsb_first

   on...:
     then:
       - lambda: !lambda |-
           id(spidev).enable();
           id(spidev).write_byte(0x4F);
           id(spidev).disable();

```

## Checklist:


kbx81:
I guess you didn't see my comment on https://github.com/esphome/issues/issues/4827, haha

clydebarrow:
> I guess you didn't see my comment on [esphome/issues#4827](https://github.com/esphome/issues/issues/4827), haha

I did, but I was nearly done when I saw it.

kbx81:
No worries, it just means you get a tougher review since it's fresh in my head. 😅
Joking aside, you've definitely done a lot more than I was planning to do initially, anyway...thank you for working on this! I'll take a look and get you some feedback.

clydebarrow:
Tough review is good. Since it's a component used by lots of others, it has to be right. I have tested it pretty thoroughly but testing other SPI based components would be useful.

esphome/components/spi/spi_esp_idf.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+SPIBus *SPIComponent::get_next_bus(GPIOPin *clk, GPIOPin *sdo, GPIOPin *sdi) {
+  spi_host_device_t channel;
+
+  if (bus_list.empty())
+    return nullptr;
+  channel = bus_list.back();
+  bus_list.pop_back();
+  return new SPIBusHw(clk, sdo, sdi, channel);
+}
```
kbx81: One of the things we've been discussing and wanting to implement is moving the SPI hardware assignment to outside of the C code.

In the branch I've been working on, I'd added a `requested_spi_host` parameter that the user can define if desired or required. Without this, it's difficult for someone who doesn't know C/C++ to determine which interface hardware they might get.

Care to work this in, too? 😄 I can give you some hints if you want to discuss...

kbx81 (commented):
Just had a quick look over this -- thanks again for tackling it. I'll do some testing here, as well, but for now I wanted to get some discussion going around re-working the "automagic" assignment of the ESP's SPI hardware. See the comment below...

kbx81:
Hmm, testing on the [Adafruit Funhouse](https://www.adafruit.com/product/4985) board and getting this accompanied by a blank display:
```
[20:58:59][D][esp-idf:000]: E (2577) spi_master: check_trans_valid(694): txdata transfer > host maximum
[20:58:59]
[20:58:59][E][spi-esp-idf:056]: Transmit failed - err 258
```

(Side note: might be helpful to display that error code in hexadecimal as they have them [here](https://github.com/espressif/esp-idf/blob/32472536715d674c160a2565795895e0273f8cde/components/esp_common/include/esp_err.h))

I had addressed this in my branch by exposing a `max_transfer_size` parameter the user can define in the yaml. It's helpful to make this buffer large when it's used with displays. Then (`write_array` as an example):

```c
esp_err_t ret;
spi_transaction_t txt;
size_t count = 0;
size_t xmit_length = length > this->max_transfer_size_ ? this->max_transfer_size_ : length;
while (xmit_length) {
  memset(&txt, 0, sizeof(txt));  // zero out the transaction
  txt.length = xmit_length * 8;  // transmit xmit_length bytes
  txt.tx_buffer =
      data +
      (count++ * this->max_transfer_size_);  // set pointer to transmit buffer with offset, increment count
  ret = spi_device_polling_transmit(this->dev_handle_.value(), &txt);  // make it so
  assert(ret == ESP_OK);                                               // should have had no issues
  if (xmit_length < this->max_transfer_size_) {
    return;
  } else {
    xmit_length = length - (this->max_transfer_size_ * count);
    xmit_length = xmit_length > this->max_transfer_size_ ? this->max_transfer_size_ : xmit_length;
  }
}
```

esphome/components/spi/spi_esp_idf.cpp
```diff
@@ -0,0 +1,125 @@
...
+ public:
+  SPIBusHw(GPIOPin *clk, GPIOPin *sdo, GPIOPin *sdi, spi_host_device_t channel)
+      : SPIBus(clk, sdo, sdi), channel_(channel) {
+    spi_bus_config_t buscfg = {};
+    buscfg.mosi_io_num = Utility::get_pin_no(sdo);
+    buscfg.miso_io_num = Utility::get_pin_no(sdi);
+    buscfg.sclk_io_num = Utility::get_pin_no(clk);
+    buscfg.quadwp_io_num = -1;
+    buscfg.quadhd_io_num = -1;
+    buscfg.max_transfer_sz = 32;
```
kbx81: I added a member variable and function to allow this to be defined pre-setup. Something like:
```c
void set_max_transfer_size(int max_transfer_size) { this->max_transfer_size_ = max_transfer_size; }
```
Then, in the component schema:

```python
            cv.Optional(CONF_MAX_TRANSFER_SIZE, default=4092): cv.All(
                cv.Range(min=4, max=4092), cv.only_with_esp_idf
            ),
```
...along with a constant for `CONF_MAX_TRANSFER_SIZE` to allow the user to define this. 😄

kbx81: (The default value can be made something more reasonable; I was just using that specific value for testing. haha)

esphome/components/spi/spi_esp_idf.cpp
clydebarrow: Could be done easily enough. The config option `requested_spi_host` or maybe just `interface` (it's in the `spi` context so `spi` in the option is redundant - it's an advanced option no need to be too verbose IMHO) and  with `auto` or a number as the argument. The existing code to allocate from the vector would need to change to mark slots as used. Would need config validation to check for conflicts.

I have assumed BTW that interfaces once assigned are never released until reboot. AFAIK ESPHome components are assumed to be eternal.

esphome/components/spi/spi_esp_idf.cpp
clydebarrow: Any reason to limit it to ESP-IDF? Should be ok with Arduino.

clydebarrow: I think...

esphome/components/spi/spi_esp_idf.cpp
kbx81: > The config option requested_spi_host or maybe just interface (it's in the spi context so spi in the option is redundant - it's an advanced option no need to be too verbose IMHO) 

Absolutely -- I hadn't given much thought to the name yet. This was just what I had been using as I've already got this built. 😉

> The existing code to allocate from the vector would need to change to mark slots as used. Would need config validation to check for conflicts.

Whatever works -- I took a "simple" approach in that the config validation just checks if any SPI bus instance specifies a hardware interface and, if it does, it considers it a required parameter for all of the SPI bus instances. In other words, if you specify one of them, it forces you to specify them all. It might sound aggressive but, realistically it's pretty uncommon to have more than two or three. I don't think it's asking too much and it avoids what I came to realize is a lot of other work for no real gain.

> I have assumed BTW that interfaces once assigned are never released until reboot. AFAIK ESPHome components are assumed to be eternal.

Totally fair assumption, IMHO. 😇

esphome/components/spi/spi_esp_idf.cpp
kbx81: Try it! 😄 My focus was mostly on IDF. It seems to Just Work™️ in Arduino-land and I didn't really look at how it's being handled there.

clydebarrow: Ok, keep on digging, I'll tackle these things as a block.

kbx81: If you're on Discord and want to chat there, ping me -- same username. 😄
https://discord.com/channels/429907082951524364/1144355652370714634

kbx81:
Related: https://github.com/esphome/feature-requests/issues/1414

clydebarrow:
There are some strange things in the RP2040 Arduino/Pico-SDK interface so I'm shelving HW SPI support for RP2040 for now. SPI on RP2040 will work, but only provide SW implementation.

esphome/components/spi/spi_esp_idf.cpp
clydebarrow: > ```
> [20:58:59][E][spi-esp-idf:056]: Transmit failed - err 258
> ```

Interestingly I can't reproduce this. I can transfer 4008 bytes with the original code (sets max_transfer to 32) or updated code and configuring any size. 

Arduino breaks transfers into chunks of 64 bytes or less. I presume that IDF allocates a buffer of `max_transfer_size` for DMA purposes, so will be faster for large transfers - but my inability to reproduce the problem has me wondering.

I've tried to set up a test on a Lilygo T5 4.7 e-paper board and it works with Arduino, but on esp-idf the buffer allocation in PSRAM fails.

kbx81: Hmmm interesting -- until just the other day, PSRAM was not working in IDF: https://github.com/esphome/esphome/pull/5298

Maybe make sure you've rebased and added `psram:` into the config you are testing if/when you are testing with IDF. 😇 This is what I had to do.

clydebarrow: My branch is up-to-date with dev, but what `param:` do I need to add?

Also just realised my 4008 bytes test might not be big enough - it's possible that IDF is allowing up to 4092 irrespective of the `.max_transfer_size` value.

kbx81: > My branch is up-to-date with dev, but what `param:` do I need to add?

All you need to add is `psram:` -- nothing more. 😄

> Also just realised my 4008 bytes test might not be big enough - it's possible that IDF is allowing up to 4092 irrespective of the `.max_transfer_size` value.

Yes, I believe 4092 is a hard limit of the DMA mechanism it uses -- not sure how that interacts with the `.max_transfer_sz` parameter, but I seem to recall bumping into the limit just the same when I made it smaller. I'll re-test and confirm shortly.

Also, which version of IDF are you testing? We default to 4.4.x right now but 5.0.x is our new target -- I've been testing my changes against both. 😇

clydebarrow: Oh, you meant `psram:` :-)

kbx81: That's `psram:` not `param:` as autocorrect keeps so helpfully trying to fix for me. 😅

Jorre05:
Hi,

I'm eagerly trying this PR to drive a SPI connected display connected to an ESP32-C3 mini. But the code fails to compile. A C3 variant only has 3 SPI's ( https://www.espressif.com/sites/default/files/documentation/esp32-c3_datasheet_en.pdf )

```c
src/esphome/components/spi/spi_esp_idf.cpp:19:16: error: 'SPI3_HOST' was not declared in this scope
     SPI2_HOST, SPI3_HOST
                ^~~~~~~~~
```

clydebarrow:
> Hi,
> 
> I'm eagerly trying this PR to drive a SPI connected display connected to an ESP32-C3 mini. But the code fails to compile. A C3 variant only has 3 SPI's ( https://www.espressif.com/sites/default/files/documentation/esp32-c3_datasheet_en.pdf )
> 
> ```c
> src/esphome/components/spi/spi_esp_idf.cpp:19:16: error: 'SPI3_HOST' was not declared in this scope
>      SPI2_HOST, SPI3_HOST
>                 ^~~~~~~~~
> ```



> C3 variant only has 3 SPI's
Good pickup. Will attend to - do you know of any other variants without SPI3?

Jorre05:
> Good pickup. Will attend to - do you know of any other variants without SPI3?

No, sorry.

clydebarrow:
> > Good pickup. Will attend to - do you know of any other variants without SPI3?
> 
> No, sorry.

Ok, do you have an example yaml I can test with for the C2?

Jorre05:
> Ok, do you have an example yaml I can test with for the C2?

I'ts quite some yaml, but I'll provide a stripped-down version. Standby...

Jorre05:
Here you go

```yml
esphome:
  name: test-device

esp32:
  framework:
    type: esp-idf
  board: esp32-c3-devkitm-1
  variant: esp32c3

logger:
  level: DEBUG

external_components:
  - source: github://pr#5311
    components:
      - spi

spi:
  clk_pin: GPIO6
  mosi_pin: GPIO7

**Comment by clydebarrow** [comment]
Thanks. Just pushed version with changes as required. Compiles your yaml fine now.

**Comment by Jorre05** [comment]
Yes! I have no physical access to the device now, but the code compiles, uploads and the devices responds as normal. Logging indicates 'Using HW SPI: 0'. I'll check the display when I get home. 

Thanks!

**Comment by clydebarrow** [comment]
Yeah, so 0 means the first user-accessible SPI device, i.e. SPI2 on that chip. There is documentation to come :-)

**Comment by Jorre05** [comment]
I can confirm, my display (st7735) works perfectly.

File: esphome/components/spi/spi_arduino.cpp [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+using SPIBusDelegate = SPIClass;
+#endif
+
+// list of available buses
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables, cppcoreguidelines-interfaces-global-init)
+static std::vector<std::function<SPIBusDelegate *()>> bus_list = {
+#ifdef USE_ESP32
+    [] { return &SPI; },  // NOLINT(cppcoreguidelines-interfaces-global-init)
+#if defined(USE_ESP32_VARIANT_ESP32C3) || defined(USE_ESP32_VARIANT_ESP32C2) || defined(USE_ESP32_VARIANT_ESP32C6)
+#elif defined(USE_ESP32_VARIANT_ESP32S2) || defined(USE_ESP32_VARIANT_ESP32S3)
```
jesserockz: Maybe a comment to say this block is intentionally blank?

clydebarrow:
I moved the hardware validation and selection into Python. Now the `interface:` choices include `spi2` and `spi3`. See the updated esphome-docs for the full blurb. I avoided allowing `vspi` and `fspi` since these are not names officially used by Espressif in their docs (those terms refer to the internal bus signals, not the SPI controllers themselves and have simply been abused by Arduino.)

kbx81 (approved):

PR #5317, opened 2023-08-30
Bugfix: disable channels after IO if multiple tca9548a  I2C multiplexers are configured

kahrendt:

If there are multiple tca9548a I2C multiplexers configured on the same bus, then there are potential address conflicts. If you have two components with the same address attached to two tca9548a multiplexers, they will both attempt to communicate at the same time. This PR disables all the channels on a tca9548a after any IO if there is more than one multiplexer configured.

I removed the MULTI_CONF option in ``__init__.py``, and used a `cv.ensure_list` to determine if more than tca9548a is configured. If only one tca9548a is configured, then the  component does not disable all channels and matches the current behavior.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

tca9548a:
  - address: 0x70
    id: multiplex0
    i2c_id: direct_bus
    channels:
      - bus_id: multplex0channel0
        channel: 0
  - address: 0x71
    id: multiplex1    
    i2c_id: direct_bus
    channels:
      - bus_id: multplex1channel0
        channel: 0

```

## Checklist:

### DIFF
--- a/esphome/components/tca9548a/__init__.py
+++ b/esphome/components/tca9548a/__init__.py
@@ -11,10 +11,9 @@ tca9548a_ns = cg.esphome_ns.namespace("tca9548a")
 TCA9548AComponent = tca9548a_ns.class_("TCA9548AComponent", cg.Component, i2c.I2CDevice)
 TCA9548AChannel = tca9548a_ns.class_("TCA9548AChannel", i2c.I2CBus)
 
-MULTI_CONF = True
-
 CONF_BUS_ID = "bus_id"
-CONFIG_SCHEMA = (
+
+TCA9548A_COMPONENT_SCHEMA = (
     cv.Schema(
         {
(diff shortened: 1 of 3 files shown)


jesserockz (changes requested):
I dont think we need to care if there are multiple multiplexers. Just disable the channel always after a transaction.

Also don't mark here as failed, just log an error and `set_status_error` perhaps.

kahrendt:
Sounds good, thanks for the feedback! I reverted back to the original ``__init__.py`` with ``MULTI_CONF=True``. The component now disables all channels after any communication. I removed the ``current_channel_`` variable and checks, as the channel will need to be written every time if things are working correctly.

jesserockz (approved):

PR #5318, opened 2023-08-30
Modbus Controller added some features

0x3333:

Added some helpers and functionalities to Modbus Controller:

* Exposed `get_command_queue_length()` method as public, in my case, it is useful to know if the module has sent all commands(In case we update very frequently).
* If the Modbus module didn't respond, we mark the module `offline`(Currently not exposed to UI or API just via code).
* Added `offline_skip_updates`, if the module is offline, we skip some updates so we don't have to wait too much for the timeout of some module and keep reading others in the same bus.

Default functionally has not changed.

Some background about this PR:

I have a home automation bus with 10 devices. Some devices are high-speed (Mainly input modules) that I update constantly using a custom component(I update every time the command queue is empty, that's why I need `get_command_queue_length()` public). For a high-speed module, if it goes offline, it is marked as faulty and is not updated for `offline_skip_updates` cycles allowing the other modules to be updated without the timeout of the faulty module affecting them too much(My timeout is already low, set to 500ms).

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/3163

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
modbus_controller:
  - id: modbus_input_0
    address: 0x1
    modbus_id: modbus_0
    offline_skip_updates: 50
```

## Checklist:

  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs). ***Not yet***


esphome/components/modbus_controller/modbus_controller.h
```diff
@@ -409,7 +409,8 @@ class ModbusCommandItem {
 
 class ModbusController : public PollingComponent, public modbus::ModbusDevice {
  public:
-  ModbusController(uint16_t throttle = 0) : command_throttle_(throttle){};
+  ModbusController(uint16_t throttle = 0, uint16_t offline_skip_updates = 0)
+      : command_throttle_(throttle), offline_skip_updates_(offline_skip_updates){};
```
jesserockz: We dont need the variables in both the constructor and as setters. Lets just change to setters only to match the codebase.
```suggestion
```

esphome/components/modbus_controller/__init__.py
```diff
@@ -206,8 +208,11 @@ async def add_modbus_base_properties(
 
 
 async def to_code(config):
-    var = cg.new_Pvariable(config[CONF_ID], config[CONF_COMMAND_THROTTLE])
-    cg.add(var.set_command_throttle(config[CONF_COMMAND_THROTTLE]))
+    command_throttle = config[CONF_COMMAND_THROTTLE]
+    offline_skip_updates = config[CONF_OFFLINE_SKIP_UPDATES]
+    var = cg.new_Pvariable(config[CONF_ID], command_throttle, offline_skip_updates)
+    cg.add(var.set_command_throttle(command_throttle))
+    cg.add(var.set_offline_skip_updates(offline_skip_updates))
```
jesserockz:
```suggestion
    var = cg.new_Pvariable(config[CONF_ID])
    cg.add(var.set_command_throttle(config[CONF_COMMAND_THROTTLE]))
    cg.add(var.set_offline_skip_updates(config[CONF_OFFLINE_SKIP_UPDATES]))
```

jesserockz (changes requested):

esphome/components/modbus_controller/modbus_controller.h
0x3333: `throttle` was already present in the constructor, should I remove it and let only the setter?

Based on the next comment, sounds like that. Will remove.

jesserockz (approved):

0x3333:
I'll create a PR regarding the docs.

0x3333:
Created Docs [PR](https://github.com/esphome/esphome-docs/pull/3163)

PR #5320, opened 2023-08-31
7789 controller fixes take 2

clydebarrow:

The 7789 driver was set up to use mode 3 SPI - high clock, trailing edge. The datasheet says it should use mode 0 - low clock, leading edge. Implementing this negates the need for the extra SLPOUT as per the previous PR.

SPI clock has also been bumped to 40 MHz - the datasheet says clock high and low periods are minimum 7ns, which means the maximum clock rate is around 70MHz.

This has been tested against the current dev branch (Arduino only) as well as PR https://github.com/esphome/esphome/pull/5311 (Arduino and ESP-IDF 5.0).

## Types of changes

**Related issue or feature (if applicable):** Relates to: https://github.com/esphome/feature-requests/issues/901

## Test Environment

- [x] ESP32
- [x] ESP32 IDF

```yaml
# Example config.yaml
esphome:
  name: t-embed

esp32:
  board: esp32-s3-devkitc-1
  variant: esp32s3
  framework:
    type: arduino

logger:
api:
ota:
wifi:
  ssid: !secret wifi_ssid

power_supply:
  - id: power_on
    pin: GPIO46

spi:
  clk_pin: GPIO12
  mosi_pin: GPIO11

display:
  - platform: st7789v
    power_supply: power_on
    model: "Custom"
    eightbitcolor: False
    rotation: 270
    width: 170
    height: 320
    offset_width: 0
    offset_height: 35
    backlight_pin: GPIO15
    cs_pin: GPIO10
    dc_pin: GPIO13
    reset_pin: GPIO9
    id: disp
    lambda: |-
      auto rand_col = Color::random_color();
      it.filled_rectangle(0, 0, it.get_width() - 1, it.get_height() - 1, rand_col);
```

## Checklist:

### DIFF
--- a/esphome/components/st7789v/st7789v.cpp
+++ b/esphome/components/st7789v/st7789v.cpp
@@ -5,6 +5,7 @@ namespace esphome {
 namespace st7789v {
 
 static const char *const TAG = "st7789v";
+static const size_t TEMP_BUFFER_SIZE = 512;
 
 void ST7789V::setup() {
   ESP_LOGCONFIG(TAG, "Setting up SPI ST7789V...");
@@ -205,15 +206,23 @@ void ST7789V::write_display_data() {
   this->dc_pin_->digital_write(true);
 
   if (this->eightbitcolor_) {
+    uint8_t temp_buffer[TEMP_BUFFER_SIZE];
+    size_t temp_index = 0;
     for (int line = 0; line < this->get_buffer_length_(); line = line + this->get_width_internal()) {
       for (int index = 0; index < this->get_width_internal(); ++index) {
         auto color = display::ColorUtil::color_to_565(
             display::ColorUtil::to_color(this->buffer_[index + line], display::ColorOrder::COLOR_ORDER_RGB,
                                          display::ColorBitness::COLOR_BITNESS_332, true));
-        this->write_byte((color >> 8) & 0xff);
-        this->write_byte(color & 0xff);
+        temp_buffer[temp_index++] = (uint8_t) (color >> 8);
+        temp_buffer[temp_index++] = (uint8_t) color;
+        if (temp_index == TEMP_BUFFER_SIZE) {
+          this->write_array(temp_buffer, TEMP_BUFFER_SIZE);
+          temp_index = 0;
+        }
       }
     }
+    if (temp_index != 0)
+      this->write_array(temp_buffer, temp_index);
   } else {
     this->write_array(this->buffer_, this->get_buffer_length_());
   }
--- a/esphome/components/st7789v/st7789v.cpp
+++ b/esphome/components/st7789v/st7789v.cpp
@@ -228,7 +228,7 @@ void ST7789V::init_reset_() {
     delay(1);
     // Trigger Reset
     this->reset_pin_->digital_write(false);
-    delay(10);
+    delay(1);
     // Wake up
     this->reset_pin_->digital_write(true);
   }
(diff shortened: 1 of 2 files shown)


kbx81:
The issue with clock rates higher than 20MHz is this:

```
[13:40:36][D][esp-idf:000]: E (331) spi_hal: spi_hal_cal_clock_conf(94): When work in full-duplex mode at frequency > 26.7MHz, device cannot read correct data.
[13:40:36]Try to use IOMUX pins to increase the frequency limit, or use the half duplex mode.
[13:40:36]Please note the SPI master can only work at divisors of 80MHz, and the driver always tries to find the closest frequency to your configuration.
[13:40:36]Specify ``SPI_DEVICE_NO_DUMMY`` to ignore this checking. Then you can output data at higher speed, or read data at your
[13:40:36][D][esp-idf:000]: E (371) spi_master: spi_bus_add_device(366): assigned clock speed not supported
[13:40:36]
[13:40:36][E][spi-esp-idf:042]: Add device failed - err 106
[13:40:36][W][spi-esp-idf:053]: spi_setup called before initialisation
[13:40:36][D][esp-idf:000]: E (422) spi_master: check_trans_valid(684): invalid dev handle
[13:40:36]
[13:40:36][E][spi-esp-idf:085]: Transmit failed - err 102
```

On some boards, the display is hardwired and the pins cannot be changed -- and we don't currently have a way to change the data rate via YAML (although, in one case, this has been worked around -- see the ILI-series component).

Regardless, I think we should stick to a data rate that by default does not render errors. Thoughts? 😄

clydebarrow:
> Regardless, I think we should stick to a data rate that by default does not render errors. Thoughts? 😄

Absolutely agree. With my new SPI code it would be possible to add a YAML config to change it, then have the best of all worlds.

clydebarrow:
I did try to test on other pins, and did not see that message, but the UART logging on my S3 board isn;t working well and logging via the network misses the setup phase. It would explain why I saw no clock though :-o

kbx81:
> I did try to test on other pins, and did not see that message, but the UART logging on my S3 board isn;t working well and logging via the network misses the setup phase. It would explain why I saw no clock though :-o

Indeed, it seems that particular message only appears once at init time and never again...and the result is that the SPI device/client fails to register and so any communication that follows simply fails -- and does so silently if you don't happen to be watching the logs. 😇

esphome/components/st7789v/st7789v.cpp
```diff
@@ -5,6 +5,7 @@ namespace esphome {
 namespace st7789v {
 
 static const char *const TAG = "st7789v";
+static const size_t TEMP_BUFFER_SIZE = 512;
```
kbx81: Out of concern for memory-constrained devices (specifically the '8266) how do you feel about making this a little smaller? ...maybe 256 or even 128? I realize larger is better for performance, though, so there is a trade-off. Thoughts?

clydebarrow: I tested initially with 128 and it was pretty good, the gain going to 512 was noticeable but not huge. So 128 it is.

kbx81 (approved):

esphome/components/st7789v/st7789v.cpp
clydebarrow: At least with memory-constrained devices all the RAM is DMA-capable :-)

kbx81: Thanks again! 🍻

PR #5327, opened 2023-09-01
Add startup_delay to interval.

clydebarrow:

Add a new configuration option for the `interval` component to delay the first execution after boot.

## Types of changes

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3158

## Test Environment

- [x] ESP32
- [x] ESP32 IDF

```yaml
# Example config.yaml
interval:
  - interval: 5s
    startup_delay: 10s
    then:
      - logger.log: Interval called
```

## Checklist:


nagyrobi:
This is useful!

nagyrobi:
Can't remember right now, but can this be negative?

Afaik when the device boots, there's one empty cycle passed before the first call. How about shortening that?

clydebarrow:
> Can't remember right now, but can this be negative?

I don't see how:

```
Optional(
                CONF_STARTUP_DELAY, default="0s"
            ): cv.positive_time_period_milliseconds
```

> Afaik when the device boots, there's one empty cycle passed before the first call. How about shortening that?

I don't believe that's the case. Here's the interval setup code:

```
  item->interval = interval;
  item->last_execution = now - offset - interval;
  item->last_execution_major = this->millis_major_;
```
So it seems it sets `last_execution` in the past, which should trigger the first `update()` call pretty soon. My motivation for this was that my `interval:` calls were happening too soon - before other components had completed setup - so I did not get the impression there was any initial delay.

clydebarrow:
@jesserockz Can we get this merged? I'd like to not have to keep merging it to my dev branches.

jesserockz (approved):

PR #5329, opened 2023-09-01
Make uart error message go away

clydebarrow:

When using ESP-IDF 5.02 this message appears in the log:
```
[22:55:08]E (170) uart: uart_param_config(726): Invalid src_clk
```

While this appears to be innocuous it's distracting. This PR prevents that message appearing.

## Types of changes

## Test Environment

- [x] ESP32 IDF

```yaml
# Example config.yaml
esp32:
  board: esp32-s3-devkitc-1
  variant: esp32s3
  framework:
    type: esp-idf
    version: 5.0.2
    platform_version: 6.3.2
```

## Checklist:


kbx81 (approved):
Tested on ESP32 and ESP32-S2 and confirmed the error no longer appears.
Thanks for fixing this @clydebarrow!

PR #5337, opened 2023-09-03
Add Lilygo T-Embed to st7789v display config.

clydebarrow:

This adds the the [T-Embed board](https://github.com/Xinyuan-LilyGO/T-Embed) display to supported models.

Some questions for reviewers:

I have hard-coded several pins since these are fixed in the T-Embed. First question is whether this is a good idea - I think it is since it reduces the knowledge required by a user to use the display.

However the `power_supply` id still needs to be provided (it's hard-wired as per the sample yaml) and the spi pins still need to be specified in an spi block.

It would be nice to hard-code these as well, but since they live in different components I have no idea how to do that. It would basically mean creating extra hidden components - hidden is fine since they can serve no other purpose, e.g. the SPI is hardwired to the display so can't be used for anything else.

This issue could also simply be dealt with via documentation.

Another question is if I should duplicate this for the TTGO display which also is hardwired.

Final question - the width, height and offsets are currently hard-coded in the .cpp file for each model. IMHO this should be moved into the Python code. I have not done this yet but it would be easy enough.

## Types of changes

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3161

## Test Environment

- [x] ESP32
- [x] ESP32 IDF

```yaml
# Example config.yaml

# Would be nice to provide this automagically.
power_supply:
  - id: power_on
    pin: GPIO46

# Ditto
spi:
  clk_pin: GPIO12
  mosi_pin: GPIO11

font:
  - file: "gfonts://Roboto"
    id: roboto
    size: 20

display:
  - platform: st7789v
    power_supply: power_on
    model: lilygo T EMBED 170X320
    lambda: |-
      auto rand_col = Color::random_color();
      it.filled_rectangle(0, 0, it.get_width() - 1, it.get_height() - 1, rand_col);
      it.print(it.get_width()/2, it.get_height()/2, id(roboto), TextAlign::CENTER, "Centered");
```

## Checklist:


clydebarrow:
I have gone ahead and moved all the configuration into Python. This simplifies the C++ code, eliminates duplication between the Python and C++ and makes it simpler to add new displays.

TTGO now has default pins specified.

All preset configuration is overridable which allows variations between board versions to be handled.

esphome/components/st7789v/display.py
```diff
@@ -29,55 +31,113 @@
...
+            CONF_OFFSET_HEIGHT: 52,
+            CONF_OFFSET_WIDTH: 40,
+        }
+    ),
+    "ADAFRUIT_FUNHOUSE_240X240": model_spec(
+        presets={
+            CONF_HEIGHT: 240,
+            CONF_WIDTH: 240,
+            CONF_OFFSET_HEIGHT: 0,
+            CONF_OFFSET_WIDTH: 0,
```
kbx81, line 60: We can add these here, too, I think:
```suggestion
            CONF_OFFSET_WIDTH: 0,
            CONF_CS_PIN: "GPIO40",
            CONF_DC_PIN: "GPIO39",
            CONF_RESET_PIN: "GPIO41",
```

kbx81:
> This issue could also simply be dealt with via documentation.

It's probably reasonable to add a table to the doc that calls out the pins for known boards that are directly supported. I'm not sure about attempting to auto-configure SPI since, in some cases, the SPI bus could be shared with other components/devices.

Overall, I think this is a nice change as it simplifies the yaml (and C/C++ code) a good bit. 😄

clydebarrow:
From what I've seen these boards dedicate pins to the display so sharing the SPI is probably not a big issue, but the effort to automatically assign them outweighs the benefit I think.

I've updated the docs but can add a table to make the information clearer.

esphome/components/st7789v/display.py
```diff
@@ -29,55 +31,116 @@
 ST7789VRef = ST7789V.operator("ref")
 ST7789VModel = st7789v_ns.enum("ST7789VModel")
 
+MODEL_ENUM = "model_enum"
```
kbx81: I think this is unused

kbx81 (approved):
LGTM

PR #5338, opened 2023-09-03
Added Handling for Nack "file not found"

croessi:
If DFPlayer sends NACK witch argument 0x06 indicating that the file was not found, still the component stays in is_playing =true state. This fixes this faulty behavior, however might not be the cleanest solution.

If NACK is received with "File not found errorcode 0x06" the is_playing state is set to false.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```

## Checklist:

### DIFF
--- a/esphome/components/dfplayer/dfplayer.cpp
+++ b/esphome/components/dfplayer/dfplayer.cpp
@@ -101,6 +101,10 @@ void DFPlayer::loop() {
             ESP_LOGV(TAG, "Nack");
             this->ack_set_is_playing_ = false;
             this->ack_reset_is_playing_ = false;
+            if (argument == 6) {
+              ESP_LOGV(TAG, "File not found");
+              this->is_playing_  = false;
+            }
           case 0x41:
             ESP_LOGV(TAG, "Ack ok");
             this->is_playing_ |= this->ack_set_is_playing_;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: esphome/components/dfplayer/dfplayer.cpp [hunk 1]
```diff
@@ -101,6 +101,10 @@ void DFPlayer::loop() {
             ESP_LOGV(TAG, "Nack");
             this->ack_set_is_playing_ = false;
             this->ack_reset_is_playing_ = false;
+            if (argument == 6) {
+              ESP_LOGV(TAG, "File not found");
+              this->is_playing_ = false;
+            }
```
glmnet, line 107: Looks good, however I just noticed a break; might be missing here, I don't see reason to fall to 0x41 as `is_playing_` isn't going to be updated (may be it was the original intention and it didn't work)

glmnet: Also it will log Ack ok which is not true, so definitely add a break here then good to go.

croessi: Thanks, I added the break. Was already wondering if this was intentional. If pipeline is green, please approve.

glmnet (approved):

PR #5348, opened 2023-09-06
time: Make std::string version of strftime() avoid runaway memory allocations

kpfleming:

`ESPTime::strftime()` can crash due to a memory allocation failure if the supplied format string contains unsupported specifiers (in particular '%s' when using the ESP-IDF framework) or if the format string does not produce any output.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml` to reproduce the problem (works with Arduino framework, crashes with ESP-IDF framework):
```yaml
time:
  - id: _time
    platform: sntp
    servers:
      - <IP address of NTP server>
    timezone: EST5EDT,M3.2.0,M11.1.0
    on_time_sync:
      - logger.log:
          format: "%s"
          args:
            - id(_time).now().strftime("%s").c_str()
```

## Checklist:

### DIFF
--- a/esphome/core/time.cpp
+++ b/esphome/core/time.cpp
@@ -49,6 +49,11 @@ std::string ESPTime::strftime(const std::string &format) {
   struct tm c_tm = this->to_c_tm();
   size_t len = ::strftime(&timestr[0], timestr.size(), format.c_str(), &c_tm);
   while (len == 0) {
+    if (timestr.size() >= 128) {
+      // strftime has failed for reasons unrelated to the size of the buffer
+      // so return a formatting error
+      return "ERROR";
+    }
     timestr.resize(timestr.size() * 2);
     len = ::strftime(&timestr[0], timestr.size(), format.c_str(), &c_tm);
   }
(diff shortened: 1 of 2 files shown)


kpfleming:
Thanks to @ssieb for helping me track down this issue!

kpfleming:
If you prefer, here's another implementation which avoids all dynamic allocations (except for the return value) and resizing:

```c++
std::string ESPTime::strftime(const std::string &format) {                                                                                                                                                                                                     
  static char output[128];                                                                                                                                                                                                                                     
  struct tm c_tm = this->to_c_tm();                                                                                                                                                                                                                            
  size_t len = ::strftime(&output[0], sizeof(output), format.c_str(), &c_tm);                                                                                                                                                                                  
  if (len == 0) {                                                                                                                                                                                                                                              
      return "ERROR";                                                                                                                                                                                                                                          
  } else {                                                                                                                                                                                                                                                     
    return std::string(&output[0], len);                                                                                                                                                                                                                       
  }                                                                                                                                                                                                                                                            
}
```

jesserockz (approved):

PR #5362, opened 2023-09-08
esp32: Update copy_files() for rm components

openxzx:
Hi, I think it would be better to update the rmtree function.


jesserockz (commented):
This was my original code until I realised the files would not be removed if you removed all components

PR #5364, opened 2023-09-08
Fix repeat.count = 0 case

Xmister:

repeat.count = 0 right now acts the same as repeat.count = 1, which should not be the case.
There are plenty of instances when a lambda returning 0 should mean the code needs to be skipped.

## Types of changes

## Test Environment

- [x] ESP32

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (approved):
LGTM

PR #5379, opened 2023-09-12
Add IDF 5 test yaml, add adc to IDF tests, fix adc for IDF 5

kbx81:

- Adds an IDF 5 test yaml (`test11.yaml`)
- Adds adc to IDF tests
- Fixes adc for IDF 5

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/adc/adc_sensor.h
+++ b/esphome/components/adc/adc_sensor.h
@@ -62,7 +62,11 @@ class ADCSensor : public sensor::Sensor, public PollingComponent, public voltage
   adc1_channel_t channel1_{ADC1_CHANNEL_MAX};
   adc2_channel_t channel2_{ADC2_CHANNEL_MAX};
   bool autorange_{false};
-  esp_adc_cal_characteristics_t cal_characteristics_[4] = {};  // array size based on adc_atten_t enum
+#if ESP_IDF_VERSION_MAJOR >= 5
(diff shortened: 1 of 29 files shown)


esphome/components/adc/adc_sensor.h
```diff
@@ -62,7 +62,7 @@ class ADCSensor : public sensor::Sensor, public PollingComponent, public voltage
   adc1_channel_t channel1_{ADC1_CHANNEL_MAX};
   adc2_channel_t channel2_{ADC2_CHANNEL_MAX};
   bool autorange_{false};
-  esp_adc_cal_characteristics_t cal_characteristics_[ADC_ATTEN_MAX] = {};
+  esp_adc_cal_characteristics_t cal_characteristics_[4] = {};  // array size based on adc_atten_t enum
```
jesserockz: I can see they removed this enum value from idf 5, but they also added a new one `SOC_ADC_ATTEN_NUM`.
It would need some ifdefs to wrap with the version, but would be better than a hardcoded value (4) in code...

kbx81: Ah! I did not find that -- good catch. Solid agree, I'll update it...

jesserockz (approved):

PR #5387, opened 2023-09-13
Allow to specify global build directory

werwolfby:

Default implementation use `.esphome/build` folder to store build artifacts. This folder can be relatively big. I run `esphome` container in kubernetes and all `/config` folder is mounted as remote storage. This is fine for small pregenerated json and `idedata`. But this is not best option for build temporary files. Instead I can use local storage for `.esphome/build` folders.

This PR allows to specify `ESPHOME_BUILD_PATH` environment variable that will be used as default build folder path. This build path still can be overridden by individual components over `build_path` option (as described in [doc](https://esphome.io/components/esphome.html)), but introduce default folder that can be specified.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3183

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/docker/docker_entrypoint.sh
+++ b/docker/docker_entrypoint.sh
@@ -21,4 +21,10 @@ export PLATFORMIO_PLATFORMS_DIR="${pio_cache_base}/platforms"
 export PLATFORMIO_PACKAGES_DIR="${pio_cache_base}/packages"
 export PLATFORMIO_CACHE_DIR="${pio_cache_base}/cache"
 
+# If /build is mounted, use that as the build path
+# otherwise use path in /config (so that builds aren't lost on container restart)
+if [[ -d /build ]]; then
+    export ESPHOME_BUILD_PATH=/build
+fi
+
 exec esphome "$@"
--- a/esphome/core/config.py
+++ b/esphome/core/config.py
@@ -198,9 +198,10 @@ def preload_core_config(config, result):
     CORE.data[KEY_CORE] = {}
 
     if CONF_BUILD_PATH not in conf:
-        build_path_env = os.environ["ESPHOME_BUILD_PATH"]
+        if "ESPHOME_BUILD_PATH" in os.environ:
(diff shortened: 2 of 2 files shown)


jesserockz (commented):
I recently made changes (#5374) that (re)moved the entire `.esphome` directory for HA addon containers. Perhaps the same logic could be used here and alter the entire `relative_internal_path` to be something else if there is a root folder mounted like `/build`?

werwolfby:
@jesserockz thanks for suggestion. I still think it is worth to separate `.esphome` (or `/data` in case of your PR) and `/build` folder.

And not sure how it will work if I'm running it as separate container, not a HA addon.

> I didn't check deeply your PR, yet

jesserockz:
tbh, Just working with `/data` in regular containers if it exists/is mounted would be a good step. Then `/config` is only for config/user created files.

werwolfby:
Please correct me if I'm wrong, cause I don't know the codebase enough. To work with `/data` folder I need to pretend to be a *HA Addon* by specifying `ESPHOME_IS_HA_ADDON` env variable. Otherwise it will use [`return self.relative_config_path(".esphome")`](https://github.com/esphome/esphome/pull/5374/files#diff-47ebb996ee5f7db414545915b30bb0a77691ac6d21642d13e22d2ba4d81a028dR561).

Does setting this env variable has other side effects?

### Update

According to code it affects only using `/data` folder, which is some that I really can use! I'll try it and will let you know, thanks for the help!

jesserockz:
There might be more in the future and I wouldn't fake being an addon, but we can set another env var in the entrypoint as you did in this PR id `/data` exists and then check that inside `relative_internal_path`

dentra:
@werwolfby you do not need to pretend HA addon, have a look at PR #5417.

PR #5398, opened 2023-09-16
Proposal: Test yaml for each component

Fabian-Schmidt:

This is a proposal to allow creating a test yaml configuration per component inside the individual component folder.
This is additional to the 'hero' full component tests in the `tests` folder.

The aim is to have the tests close to the component code and avoid the large test code. Ideally the test code is also the example from the documentation.
The test runner is `scripts/test_build_components`. It tries to minimize specific platform knowledge.
For each platform is a base file in the e.g. `build_components_base.esp32-ard.yaml`. The runner will make a copy of this file and use the esphome `packages` feature to load one component for testing.
As a component author to write a the test I create a file `./tests/components/[target_component]/[test_name].[target_platform].yaml`.  The runner picks up the file and based on the filename determines the target platform.
I added a few examples in this PR to showcase how this could look like.
I tried to use the command line substitution feature when possible. It does not work for package file path.

## TODO

- [x] Include in CI pipeline

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040

```yaml
# Example config.yaml

```

## Checklist:


jesserockz:
Hi @Fabian-Schmidt.

I was actually talking to @kbx81 last week about this exact idea and we also have some ideas. 
I am focused on the beta this week until the release next week, but I will then come back and do a proper review of this then hopefully.

Cheers,
Jesse

kbx81:
Circling back to this...we definitely want to improve our tests/testing where possible...and this is a step in the right direction. 🍻

One of the items @jesserockz and I discussed was using Python to build out the testing framework -- it's more consistent with most of the rest of the project and offers an advantage in that it would be easier to auto-determine some aspects of what should be tested and how. For example, we could determine by any given component module which framework(s) it can or cannot be used with and, consequently, test only/exactly what's allowed by the component's validation. In this sense, it would rely less on the (manual) creation of test.yaml files which would offer a greater degree of automation, further reduce errors and make it somewhat easier to maintain.

Any thoughts? 😄

Fabian-Schmidt:
> Circling back to this...we definitely want to improve our tests/testing where possible...and this is a step in the right direction. 🍻
> 
> One of the items @jesserockz and I discussed was using Python to build out the testing framework -- it's more consistent with most of the rest of the project and offers an advantage in that it would be easier to auto-determine some aspects of what should be tested and how. For example, we could determine by any given component module which framework(s) it can or cannot be used with and, consequently, test only/exactly what's allowed by the component's validation. In this sense, it would rely less on the (manual) creation of test.yaml files which would offer a greater degree of automation, further reduce errors and make it somewhat easier to maintain.
> 
> Any thoughts? 😄

I understand the request and thought of a similar approach at first.

I avoided it due to the necessary complexity and 'magic'/logic baked into such test runner framework.
To avoid reinventing everything best to choose an existing test framework from python and add only necessary features.
Also it could be wise to offer C++ level unit test and not just does it compile tests.

I am very uncomfortable in writing python and therefore choose a kiss and familiar approach and implemented a test runner in bash.

This proposal achieves the goal of component level build test with minimal additional effort by the person implementing the component. I argue it is  almost zero as the person creating the component has to create such config for local testing and the docs. 
The big benefit of this approach is full transparency of what is going on. I fear chasing Python test runner issues.

Does this proposal hinder/block any future implementation of an intelligent test runner?

nielsnl68:
I think this is a good idea @kbx81. 

>In addition with your idea is something what i already talked about earlier with @jesserockz.
>That is my suggestion to moving all related test scripts to there own component, in my honest opinion this will also help with a better maintaining of the test's it self, the CI test procedure could become much faster because it test only the changed component.

~~I would have worked on this idea already, weren't I already having too many project ideas I'm working on at the moment.~~ Not needed anymore, thanks to @Fabian-Schmidt, most appreciated.

EDIT: Oeps, i should have read the OP message first before reponding to @kbx81 message only. Thanks for doing the work i was thinking about.

Thinking a bit more; why not allowing both way's?  for component creators/updaters that want an easy test could do just an YAML test script. When they want a thorough test they could a python script like @kbx81 suggested.

In addition to this test i would like to see an basic `unit_test` component. that can do some basic function and method testing.

nielsnl68:
@Fabian-Schmidt, I like your approach the only think i would have done different is moving the test scripts in a separate folder something like:

> As a component author to write a the test I create a file **./esphome/components/[target_component]/_test/[test_name].[target_platform].yaml** or **./esphome/components/[target_component]/.test/[test_name].[target_platform].yaml**

This keeps the component folder clean, imho.

Fabian-Schmidt:
> @Fabian-Schmidt, I like your approach the only think i would have done different is moving the test scripts in a separate folder something like:
> 
> > As a component author to write a the test I create a file **./esphome/components/[target_component]/_test/[test_name].[target_platform].yaml** or **./esphome/components/[target_component]/.test/[test_name].[target_platform].yaml**
> 
> This keeps the component folder clean, imho.

I moved the component test files into a folder `tests` to have the same name as the root test folder and also other projects.
The proposed name `.test` has the downside of being a hidden folder and `_test` is unusual.

Fabian-Schmidt:
I have added the component build test to the ci pipeline.

nielsnl68:
> > @Fabian-Schmidt, I like your approach the only think i would have done different is moving the test scripts in a separate folder something like:
> > > As a component author to write a the test I create a file **./esphome/components/[target_component]/_test/[test_name].[target_platform].yaml** or **./esphome/components/[target_component]/.test/[test_name].[target_platform].yaml**
> > 
> > 
> > This keeps the component folder clean, imho.
> 
> I moved the component test files into a folder `tests` to have the same name as the root test folder and also other projects. The proposed name `.test` has the downside of being a hidden folder and `_test` is unusual.

About the `.test` name i do agree on that and let it skip for now. The main reason i suggested the prefixes is that normally in espHome (as you for sure know) the subfolders are used for storing platform extensions. When we use just `test` then it could be interfere with a future test component for example. Although the '_` prefix is unusual, it is being used in other software packages as well for the same sort reason.

Fabian-Schmidt:
> About the `.test` name i do agree on that and let it skip for now. The main reason i suggested the prefixes is that normally in espHome (as you for sure know) the subfolders are used for storing platform extensions. When we use just `test` then it could be interfere with a future test component for example. Although the '_` prefix is unusual, it is being used in other software packages as well for the same sort reason.

I renamed the component test folder from `tests` to `_test`.

Fabian-Schmidt:
I would like to proceed on this PR.

Is this Proposal for testing components considered/acceptable?

Should I change something about it?

What could be next steps for this PR?

nielsnl68:
Just wondering, did you think about having multiple test scripts for different scenarios like:
One with network (wifi) and one without network to validate that in both cases it still works etc.?

Fabian-Schmidt:
> Just wondering, did you think about having multiple test scripts for different scenarios like: One with network (wifi) and one without network to validate that in both cases it still works etc.?

This is just a smoke build test and does not run the actual code.
Does the presence of network (wifi) could cause build issues with some components?

That said the build runner does not care about the details. So the `tests/test_build_components/build_components_base.*.yaml` files can be modified to include network or alternatively additional base files can be added.

I wanted to keep the tests small so they can run quickly.

nielsnl68:
> That said the build runner does not care about the details. So the `tests/test_build_components/build_components_base.*.yaml` files can be modified to include network or alternatively additional base files can be added.

When this is the case then my question is answered positive no need to extend it any further :D

kbx81:
I will be revisiting this this coming week. Please standby. 😇

kbx81:
Sorry for the delay! We generally like this approach. We have a short list of questions/concerns/things we want to make sure are handled. Here we go...

- How will this react when there are changes in core? In general, if/when core is changed, we believe all tests for all components should run. We're unclear on if/how that's handled here, but we need to make sure it's covered.
- How are dependencies handled? For example, if I make a change to the base `sensor` component, all components that depend on/inherit from it should also run. Since each component is its own (Python) module, it's probably reasonable to extract dependency information that way -- you may want to look at some of the helper functions in `esphome/loader.py`.
- We think that the yaml files for each component should be moved into a dedicated test folder -- for example: `tests/components/adc` We realize it's convenient to have the tests in the same directory as the component but we have concerns about that becoming messy as support for more and more chips and frameworks is added.
- As for naming, we want to try to keep it simple yet specific. Something like either:
  - `test.platform.yaml` or...
  - `test.*.yaml`
  We think `SensorAnyPlatform.yaml` is a bit confusing and adds complexity, for example.
- For template components, we think it would be good to consolidate as much as possible in one yaml file, as we're concerned about how long it is taking to run individual tests.

Hopefully this all makes sense -- let us know what you think and we are happy to chat more about it. 😄 Thanks for working on this! 🍻

jesserockz:
Just to add a small bit:

> For template components, we think it would be good to consolidate as much as possible in one yaml file, as we're concerned about how long it is taking to run individual tests.

This was meant more as any component that implements multiple platforms (sensor, switch etc) should only create 1 yaml per target chip (esp32, esp8266 etc) because I noticed that a single template sensor test took 7 mins, and there should already be multiple template sensors in that yaml file as there are multiple exclusive ways to configure a template sensor.

Fabian-Schmidt:
Thanks @kbx81 for the feedback. Here my response:

> * How will this react when there are changes in core? In general, if/when core is changed, we believe all tests for all components should run. We're unclear on if/how that's handled here, but we need to make sure it's covered.
* Currently I handle core changes like any other change. As part of the CI pipeline the script `script/list-components.py --changed` will identify that the component is changed and only issue a test for this one changed component.
* I can change the CI pipeline to run all tests if a core component was modified.

> * How are dependencies handled? For example, if I make a change to the base `sensor` component, all components that depend on/inherit from it should also run. Since each component is its own (Python) module, it's probably reasonable to extract dependency information that way -- you may want to look at some of the helper functions in `esphome/loader.py`.
* Dependencies are not handled. In the CI pipeline the script `script/list-components.py --changed` will only identify modified components.
* The issue writing a logic to identify the modified child components is not easy. Because the dependency tree cannot be created static, because  `AUTO_LOAD` is a function and can therefore have a dynamic outcome.
  * For example if the `psram` component is modified I expect that [ili9xxx](https://github.com/esphome/esphome/blob/dev/esphome/components/ili9xxx/display.py) is tested. But it uses `AUTO_LOAD` function so depending if `CORE.is_esp32` is set I will get a different results.
* I was aiming for a KISS (keep it simple) approach and therefore ignored this issue.

> * We think that the yaml files for each component should be moved into a dedicated test folder -- for example: `tests/components/adc` We realize it's convenient to have the tests in the same directory as the component but we have concerns about that becoming messy as support for more and more chips and frameworks is added.
* In the proposal the tests are in a separate folder inside the component. Just to make sure I understood the idea is to move them like this:
  * Proposal: `/esphome/components/[target_component]/_test/[test_name].[target_platform].yaml`
  * Change: `/tests/components/[target_component]/[test_name].[target_platform].yaml`

> * As for naming, we want to try to keep it simple yet specific. Something like either:
>   
>   * `test.platform.yaml` or...
>   * `test.*.yaml`
>     We think `SensorAnyPlatform.yaml` is a bit confusing and adds complexity, for example.
* There will be multiple tests per component and they must have a unique name. So the test name would then be:
  * `[test_name].[target_platform].yaml`
    * `test_name`- unique name given to this test.
    * `target_platform` - Either `*` if the test is applicable to all existing platforms or a named platform like `esp32-ard` or `esp32`.
  * In the proposal the `*` for target_platform is optional and be omitted. I can remove this feature.
  * I don't want to enforce `test` as a string into the filename as the folder name does contain the text `test` already. So I see little benefit.

> * For template components, we think it would be good to consolidate as much as possible in one yaml file, as we're concerned about how long it is taking to run individual tests.
* I share the concern. I just added one template component to showcase. This must not be the final test file.

> 
> Hopefully this all makes sense -- let us know what you think and we are happy to chat more about it. 😄 Thanks for working on this! 🍻

Yes it does make sense.

jesserockz:
> I can change the CI pipeline to run all tests if a core component was modified.

Yes please.

> The issue writing a logic to identify the modified child components is not easy. Because the dependency tree cannot be created static, because AUTO_LOAD is a function and can therefore have a dynamic outcome.
For example if the psram component is modified I expect that [ili9xxx](https://github.com/esphome/esphome/blob/dev/esphome/components/ili9xxx/display.py?rgh-link-date=2023-11-28T21%3A02%3A37Z) is tested. But it uses AUTO_LOAD function so depending if CORE.is_esp32 is set I will get a different results.

Ah, yes I forgot about the dynamic auto loads. I do not have any plans to make `dependencies` dynamic though so that one would be fine to safely work with atm. Also, if `esphome/components/sensor` is modified, it should run any component test that has a `sensor` platform (not sure how this would be easily scripted though, and might be simpler just to run everything).

> In the proposal the tests are in a separate folder inside the component. Just to make sure I understood the idea is to move them like this:
Proposal: /esphome/components/[target_component]/_test/[test_name].[target_platform].yaml
Change: /tests/components/[target_component]/[test_name].[target_platform].yaml

Correct

> I don't want to enforce test as a string into the filename as the folder name does contain the text test already. So I see little benefit.

Just to also be clear on one above point, each test file takes several minutes to run, so the better approach would be to combine them all into a single test file (per chip platform of course) to cut down on the actual number of `esphome compile` commands run.

> I don't want to enforce test as a string into the filename as the folder name does contain the text test already. So I see little benefit.

We are not saying enforce the naming, but as this PR will be the basis of future PRs to come, lets keep it simple for now.

kbx81:
@Fabian-Schmidt just a friendly check-in...are you continuing to work on this? It is something we are interested in getting in sooner rather than later. That said, I realize the holidays are upon us so time may be constrained. Let me/us know if we can help. 😄

Fabian-Schmidt:
Hi,
I made the following changes:

- Move the folder of the test yaml files to `/tests/components/[target_component]/[test_name].[target_platform].yaml`.
- Add logic to `list-components.py` to find changes depending on current component changes based on dependencies and auto_load fields.
So for example a change in `esp32_ble_tracker` should now rebuild the tests of all components depending on it. 
The logic will also detect nested changes. So a change in `esp32_ble` will run the tests for `esp32_ble_tracker` and all its components.
- I flashed out the `template` test file a bit to cover more cases.

I believe this addresses the reported feedback (so far). Please correct me if I missed someones message.

kbx81:
Looking good! I've been holding off a bit so we could focus on the December release. We'll circle back after the holidays, but overall I think this is in a pretty good place. 😄

kbx81:
@Fabian-Schmidt I attempted to merge `dev` in but as we now have some pytest CI it looks like this is failing on Windows due to the file name. Could you take a quick look?

Fabian-Schmidt:
> @Fabian-Schmidt I attempted to merge `dev` in but as we now have some pytest CI it looks like this is failing on Windows due to the file name. Could you take a quick look?

Yes. I forgot `*` is a forbidden character in windows for filenames.
The proposal does contain `tests/components/template/test.*.yaml`  to highlight that this test must run for all platforms (ESP32, RP2040 ...).
I update the filename indicator to `all` to highlight that a component compile test must run for all platforms.

jesserockz (approved):
LGTM. Lets see how it goes =)

kbx81 (approved):

PR #5406, opened 2023-09-18
Speed up (and fix) ili9xxx display component.

### LINKED ISSUE #5089: Display ST7789v reset pin set as Required when it should be set as Optional but when not there tries to set the pin to 41.
### The problem

```yaml
# TTGO-Camera Plus 001
# https://www.lilygo.cc/en-ca/products/t-camera-plus
# https://github.com/Xinyuan-LilyGO/esp32-camera-screen/blob/master/schematic/OV2640_V2.0.pdf

substitutions:  
  # SPI
  spi_clk: GPIO21
  spi_mosi: GPIO19
  spi_miso: GPIO22
  
  # Display Physical Connector
  #1: BLK = Backlight
  #2: D/C = Data/Command
  #3: RES = Reset
  #4: SDA = Serial Data or SPI MOSI
  #5: SCL = Serial Clock or SPI SCK
  #6: VCC (3.3V)
  #7: Ground
  dis_cs: GPIO12 # Chip Select
  dis_dc: GPIO15 # Data/Command
  dis_bk: GPIO2 # Backlight
  
display:
  - platform: st7789v
    id: st7789_display
    model: Custom
    height: 240
    width: 240
    offset_width: 0
    offset_height: 0
    backlight_pin: $dis_bk
    cs_pin: $dis_cs
    dc_pin: $dis_dc
    #reset_pin: 0
```

I first was getting this error when I had it set to model: Adafruit_Funhouse 240x240

```
display.st7789v: [source /config/ttgo-camera-plus.yaml:631]
  
  Invalid pin number: 41 (must be 0-39).
```

Jesse told me to use model: Custom and set the params for it as that model above is for an ESP32-S2.

After doing so I got this message:
```
"both dc_pin and reset_pin must be specified"
```

Jesse looked at it and showed this
```c++
    if CONF_DC_PIN not in config or CONF_RESET_PIN not in config:
        raise cv.Invalid(f"both {CONF_DC_PIN} and {CONF_RESET_PIN} must be specified")
```

Not sure about the GPIO41 but Jesse said it only exists on an ESP32-S3 not the ESP32-DOWDQ6 on this module.

My board is set to esp32dev so I'm not sure why it is trying to set gpio41 to it when you don't specify the reset pin

```yaml
esp32:
  board: esp32dev
  framework:
    #type: arduino
    type: esp-idf
```

The documentation should show it as optional as well.

The display does work though when I set the reset pin to an arbitrary GPIO number like 0 above.

Thank you.

2023.10.6

Docker

Issue with ESPHome not HA

ESP32

### Board

TTGO-Camera Plus

Display ST7789v

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

clydebarrow:

* Improve the update speed of the ili9xxx display driver, for 16 bit colour at least (should also be some improvement for 8 bit.) Speed in 16 bit mode is roughly tripled (full screen write reduced from 120ms to 40ms @ 40MHz SPI.)
* Fix a bug in `display.py` that meant use of `color_palette: GRAYSCALE` did not build. See https://github.com/esphome/issues/issues/4896.

Tested on [this display](https://protosupplies.com/product/tft-lcd-3-2-240x320-rgb-spi-display-with-touchscreen/) in 16 bit and 8 bit modes (Grayscale and image adaptive). ~~Not tested on an 18 bit display since I don't have one.~~

Also tested in 16 and 18 bit mode on an ILI9481 (320x480). This required another fix (now included in this PR) for https://github.com/esphome/issues/issues/4903

Now also adds support for the st7789v to the ili9xxx component. This indirectly fixes https://github.com/esphome/issues/issues/5089, which is an issue only in the st7789v component.

Adds support for `power_supply:` config and LCD panel driver configs to swap and mirror the x/y axes (see example config below.)

## Types of changes

- [x] Other

Pull request in esphome-docs with documentation (if applicable): esphome/esphome-docs#3200

## Test Environment

- [x] ESP32
- [x] ESP32 IDF

```yaml
# Example config.yaml
time:
  - platform: sntp
    id: ntp

spi:
  clk_pin: GPIO18
  mosi_pin: GPIO23
  interface: hardware

font:
  - file: "gfonts://Roboto"
    id: roboto_40
    size: 40
    glyphs: '!"%()+=,-_.:°0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ abcdefghijklmnopqrstuvwxyz@'

output:
  - platform: ledc
    pin: 32
    id: gpio_32_backlight_pwm

# Define a monochromatic, dimmable light for the backlight
light:
  - platform: monochromatic
    output: gpio_32_backlight_pwm
    name: "Display Backlight"
    id: back_light
    restore_mode: ALWAYS_ON

display:
  - platform: ili9xxx
    model: st7789v
    power_supply: power_on
    height: 170
    width: 320
    offset_height: 35
    offset_width: 0
    panel_setup:
      swap_xy: true
      mirror_x: false
      mirror_y: true
      color_order: bgr
    data_rate: 80MHz
    cs_pin: 10
    dc_pin: GPIO13
    reset_pin: GPIO9
    lambda: |-
      it.print(0, 0, id(roboto_40), Color(0, 255, 0), "Hello World!");
      it.print(0, 86, id(roboto_40), Color(255, 0, 0), "Hello @esphome_!");
      it.strftime(120, 200, id(roboto_40), TextAlign::BASELINE_LEFT, "%H:%M:%S", id(ntp).now());
```

## Checklist:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs)


esphome/components/ili9xxx/display.py
```diff
@@ -140,8 +140,6 @@ async def to_code(config):
         rhs = []
         for x in range(256):
             rhs.extend([HexInt(x), HexInt(x), HexInt(x)])
-        prog_arr = cg.progmem_array(config[CONF_RAW_DATA_ID], rhs)
-        cg.add(var.set_palette(prog_arr))
```
nielsnl68: Why did you remove this?

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,73 @@ void ILI9XXXDisplay::update() {
...
-
-    while (rem > 0) {
-      uint32_t sz = std::min(rem, ILI9XXX_TRANSFER_BUFFER_SIZE);
-      // ESP_LOGVV(TAG, "Send to display(pos:%d, rem:%d, zs:%d)", pos, rem, sz);
-      buffer_to_transfer_(pos, sz);
+           "height:%d, mode=%d, 18bit=%d)",
+           this->x_low_, this->y_low_, this->x_high_, this->y_high_, w, h, this->buffer_color_mode_,
+           this->is_18bitdisplay_);
+
+  set_addr_window_(0, this->y_low_, this->width_ - 1, this->y_high_);
```
nielsnl68: why did you removed the width window here?

nielsnl68 (changes requested):
Looks good indeed.
I have just 2 questions below.

esphome/components/ili9xxx/display.py
clydebarrow: It was a duplicate, and caused a compile time error due to the same id being defined twice.

esphome/components/ili9xxx/ili9xxx_display.cpp
clydebarrow: By ignoring the min/max x values and just updating full lines it's possible to do a single SPI write of the modified part of the buffer. The speedup due to eliminating SPI transaction setup overhead more than compensates for the possible redundant data being written.

esphome/components/ili9xxx/display.py
clydebarrow: Lines 178-180 are executed for both GRAYSCALE and ADAPTIVE mode.

nielsnl68: Oww that is a good find 👍🏻

esphome/components/ili9xxx/ili9xxx_display.cpp
nielsnl68:
```suggestion
  set_addr_window_(this->x_low_, this->y_low_, this->x_high_, this->y_high_);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,73 @@ void ILI9XXXDisplay::update() {
...
+  uint8_t transfer_buffer[ILI9XXX_TRANSFER_BUFFER_SIZE];
   // check if something was displayed
   if ((this->x_high_ < this->x_low_) || (this->y_high_ < this->y_low_)) {
     ESP_LOGV(TAG, "Nothing to display");
     return;
   }
 
-  set_addr_window_(this->x_low_, this->y_low_, w, h);
+  // we will only update the changed rows to the display
+  uint16_t const w = this->width_;
```
nielsnl68:
```suggestion
  uint16_t const w = this->x_high_ - this->x_low_ + 1;  // NOLINT
```
See reason below.

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,73 @@ void ILI9XXXDisplay::update() {
...
+  uint16_t const h = this->y_high_ - this->y_low_ + 1;  // NOLINT
+  size_t pos = this->y_low_ * this->width_;
```
nielsnl68:
``` suggestion
  size_t pos = (this->y_low_ * this->width_) + this->x_low_;
```
See reason below.

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,73 @@ void ILI9XXXDisplay::update() {
...
-      // ESP_LOGVV(TAG, "Send to display(pos:%d, rem:%d, zs:%d)", pos, rem, sz);
-      buffer_to_transfer_(pos, sz);
+           "height:%d, mode=%d, 18bit=%d)",
+           this->x_low_, this->y_low_, this->x_high_, this->y_high_, w, h, this->buffer_color_mode_,
+           this->is_18bitdisplay_);
+
+  set_addr_window_(0, this->y_low_, this->width_ - 1, this->y_high_);
+  if (this->buffer_color_mode_ == BITS_16 && !this->is_18bitdisplay_) {
+    // 16 bit mode maps directly to display format
+    this->write_array(this->buffer_ + pos * 2, h * w * 2);
```
nielsnl68:
``` suggestion
    while  (rem-- != 0) {
       this->write_array(this->buffer_ + pos * 2,  w * 2);
       pos = pos + this->width_;
       rem = rem - w;
    }
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,73 @@ void ILI9XXXDisplay::update() {
...
       } else {
-        this->write_array16(transfer_buffer_, sz);
+        put16_be(transfer_buffer + idx, color_val);
+        idx += 2;
+      }
+      if (idx == ILI9XXX_TRANSFER_BUFFER_SIZE) {
+        this->write_array(transfer_buffer, idx);
+        idx = 0;
+        App.feed_wdt();
       }
```
nielsnl68, line 258:
``` suggestion
      }
      if ( idx % w == 0) {
         pos = pos + (this->width_ - this->x_high_) + this->x_low_;
      }

**Review by nielsnl68 — CHANGES_REQUESTED** [review]
Although your idea of doing bulk writes is nice at first I think it is only useful with small display's but not when we use bigger display's.
I suggest to update the `pos` for every row you send to the display. 
Like the examples below.

File: esphome/components/ili9xxx/ili9xxx_display.cpp (hunk 6 above)
Comment (clydebarrow) [reply esphome/components/ili9xxx/ili9xxx_display.cpp:258]: That won't work - `idx` is an index into the transfer buffer, not the display buffer. It will need another variable to track the position in the display buffer.

Comment (clydebarrow) [reply esphome/components/ili9xxx/ili9xxx_display.cpp:258]: So I applied your suggestions (with another variable) and it crashes (watchdog timeout.)

Comment (nielsnl68) [reply esphome/components/ili9xxx/ili9xxx_display.cpp:258]: Can you commit the changes, so i can see where it did go wrong.

Comment (clydebarrow) [reply esphome/components/ili9xxx/ili9xxx_display.cpp:258]: Sure but here are some benchmark numbers to consider - I fudged your code enough to send each line separately, so even though the display is wrong, the timing will be the same. This is on a 320 pixel wide display  - not sure what you consider a "bigger" display. 

I tested two scenarios - full screen redraw (always the case unless `auto_clear_enabled` is set to `false`) and toggling a 40x40 pixel rectangle - so a tiny part of the whole screen changed.

#### My PR as it stands:

Full screeen redraw: 47ms
40x40 redraw:           6ms (12800 pixels transmitted.)

#### Line by line as per your suggestion:

Full screen redraw: 185ms
40x40 redraw: 5ms (1600 pixels transmitted)

So there is only a small saving when very small areas are redrawn (and the time then is going to be small anyway) and a huge hit when bigger areas are redrawn.  This is due to the relatively high overhead of starting an SPI transaction.

Now bear in mind that even if only small areas of the screen are redrawn, the full bounding box of the changes is redrawn, so two small changes on opposite sides of the screen mean nearly the entire width is considered to have changed.

So there will be some very special cases where there is a small benefit to going line-by-line but in my opinion a typical use-case is likely to see worse performance, not better.

So I'm happy to commit your suggestions for you to debug (it won't even compile as it stands) but at this point I don't believe it's worth the effort.

Comment (clydebarrow) [reply esphome/components/ili9xxx/ili9xxx_display.cpp:258]: Those numbers BTW are just the execution time of `void ILI9XXXDisplay::display_()` - I added test code to measure duration using `millis()`.

Comment (clydebarrow) [reply esphome/components/ili9xxx/ili9xxx_display.cpp:258]: Oh, and the actual display is 240 wide, not 320 (I have it rotated 90 degrees :-) So on a wider display there will be more benefit to line by line, but large areas will still suffer greatly.

File: esphome/components/ili9xxx/ili9xxx_display.cpp (hunk 5 above)
Comment (nielsnl68) [reply esphome/components/ili9xxx/ili9xxx_display.cpp]: oeps, a bit to hasty copy past there:
```c++
  while  (rem != 0) {
```

and maybe add a `App.feed_wdt();` in the while loop as well ;)

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,80 @@ void ILI9XXXDisplay::update() {
...
-      // ESP_LOGVV(TAG, "Send to display(pos:%d, rem:%d, zs:%d)", pos, rem, sz);
-      buffer_to_transfer_(pos, sz);
+           "height:%d, mode=%d, 18bit=%d)",
+           this->x_low_, this->y_low_, this->x_high_, this->y_high_, w, h, this->buffer_color_mode_,
+           this->is_18bitdisplay_);
+
+  set_addr_window_(this->x_low_, this->y_low_, this->x_high_, this->y_high_);
+  if (this->buffer_color_mode_ == BITS_16 && !this->is_18bitdisplay_) {
+    // 16 bit mode maps directly to display format
+    while  (rem-- != 0) {
```
nielsnl68: Oeps my mistake to hasty with copy paste here

```suggestion
    while  (rem != 0) {
       App.feed_wdt();
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,80 @@ void ILI9XXXDisplay::update() {
...
-
-    while (rem > 0) {
-      uint32_t sz = std::min(rem, ILI9XXX_TRANSFER_BUFFER_SIZE);
-      // ESP_LOGVV(TAG, "Send to display(pos:%d, rem:%d, zs:%d)", pos, rem, sz);
-      buffer_to_transfer_(pos, sz);
+           "height:%d, mode=%d, 18bit=%d)",
+           this->x_low_, this->y_low_, this->x_high_, this->y_high_, w, h, this->buffer_color_mode_,
+           this->is_18bitdisplay_);
+
+  set_addr_window_(this->x_low_, this->y_low_, this->x_high_, this->y_high_);
```
nielsnl68:
```suggestion
  set_addr_window_(this->x_low_, this->y_low_, this->x_high_, this->y_high_);
  size_t rem = h * w;  // remaining number of pixels to write
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -162,86 +170,80 @@ void ILI9XXXDisplay::update() {
...
+  if (this->buffer_color_mode_ == BITS_16 && !this->is_18bitdisplay_) {
+    // 16 bit mode maps directly to display format
+    while  (rem-- != 0) {
+       this->write_array(this->buffer_ + pos * 2,  w * 2);
+       pos = pos + this->width_;
+       rem = rem - w;
+    }
+  } else {
+    size_t rem = h * w;  // remaining number of pixels to write
```
nielsnl68, line 226:
```suggestion
```

esphome/components/ili9xxx/ili9xxx_display.cpp
nielsnl68: cool test. Not so cool results :(

> This is on a 320 pixel wide display - not sure what you consider a "bigger" display.
480x320 and more. like the ili93488 or ili93486 displays

Can we talk about this on discord later just DM me and i will comeback to you when i have time later today. My name is the same as here.

nielsnl68 (changes requested):

esphome/components/ili9xxx/ili9xxx_display.cpp
clydebarrow: You didn't try compiling did you?

```
src/esphome/components/ili9xxx/ili9xxx_display.cpp:194:13: error: 'rem' was not declared in this scope; did you mean 'drem'?
```

With all due respect, I'm not inclined to try to apply major changes that you haven't even built or tested, especially when I believe the end result will be inferior to what I have proposed. I'm going to revert the suggestions, happy to consider further suggestions if well-founded.

esphome/components/ili9xxx/ili9xxx_display.cpp
clydebarrow: Bedtime for me now, could not find you on discord. I'm cly_de_barrow (discord reserves Clyde for its AI bot apparently.)

clydebarrow: So here's how I see it:

* my code as it stands is faster than the existing implementation in _all_ cases, _much_ faster in most cases.
* It would nonetheless be possible to improve the speed for updates with narrow widths;
* The problem is that there is over 100µs of setup time for each SPI transaction (see scope capture below)
* To speed up narrow updates without ruining the general case will require a run-time adaptation choosing between different approaches;
* The complexity of that means significant time to implement and test, which I don't have.

So, I suggest going with the PR as is (preferably after finding someone to test the 18 bit display case) and if anyone has the time and motivation to improve it even further they can do so.



scope_1 copy

clydebarrow: And a few more numbers for larger displays (tested timing only, not functionality). Partial updates are 40 pixels wide, 200 lines high. Times in ms for `ILI9XXXDisplay::display_()`.
```
                                  dev branch                    PR
                               full       partial        full      partial
ILI9486 (480x320, 16 bit).      338           23          96         58
ILI9488 (480x320, 18 bit)      7400          410         417        262
```

So for 18 bit the improvement is major for all cases. For 16 bit the worst case is much better, best case still fast enough that any improvement is unlikely to be noticeable.

esphome/components/ili9xxx/display.py
jesserockz: @clydebarrow Can you put this fix into its own pr so I can pull into the beta and release today?

clydebarrow: https://github.com/esphome/esphome/pull/5433

esphome/components/ili9xxx/display.py
```diff
@@ -77,20 +105,33 @@ def _validate(config):
...
+)
 CONFIG_SCHEMA = cv.All(
     font.validate_pillow_installed,
     display.FULL_DISPLAY_SCHEMA.extend(
         {
             cv.GenerateID(): cv.declare_id(ili9XXXSPI),
             cv.Required(CONF_MODEL): cv.enum(MODELS, upper=True, space="_"),
             cv.Optional(CONF_DIMENSIONS): cv.dimensions,
+            cv.Optional(CONF_WIDTH): cv.int_,
+            cv.Optional(CONF_HEIGHT): cv.int_,
```
jesserockz: This should be like this
```suggestion
            cv.Optional(CONF_DIMENSIONS): cv.Any(
                cv.dimensions,
                cv.Schema({
                    cv.Optional(CONF_WIDTH): cv.int_,
                    cv.Optional(CONF_HEIGHT): cv.int_,
                }),
            ),
```

esphome/components/ili9xxx/display.py
```diff
@@ -120,7 +165,16 @@ async def to_code(config):
...
-
+    if CONF_PANEL_SETUP in config:
+        panel = config[CONF_PANEL_SETUP]
+        mad = COLOR_ORDERS[panel[CONF_COLOR_ORDER]] | 0x8000
+        if panel[CONF_MIRROR_Y]:
+            mad |= 0x80
+        if panel[CONF_MIRROR_X]:
+            mad |= 0x40
+        if panel[CONF_SWAP_XY]:
+            mad |= 0x20
```
nielsnl68: I prefer that is done in the class it self, not all drivers could work the same way.

esphome/components/ili9xxx/display.py
```diff
@@ -46,18 +49,43 @@ def AUTO_LOAD():
...
 CONF_LED_PIN = "led_pin"
 CONF_COLOR_PALETTE_IMAGES = "color_palette_images"
 CONF_INVERT_DISPLAY = "invert_display"
+CONF_MIRROR_X = "mirror_x"
+CONF_MIRROR_Y = "mirror_y"
+CONF_SWAP_XY = "swap_xy"
+CONF_PANEL_SETUP = "panel_setup"
+CONF_COLOR_ORDER = "color_order"
+CONF_OFFSET_HEIGHT = "offset_height"
+CONF_OFFSET_WIDTH = "offset_width"
```
nielsnl68, line 75:
```suggestion
CONF_OFFSET_y= "offset_x"
CONF_OFFSET_Y = "offset_y"
```

esphome/components/ili9xxx/display.py
```diff
@@ -77,20 +105,33 @@ def _validate(config):
...
+PANEL_SCHEMA = cv.Schema(
+    {
+        cv.Optional(CONF_SWAP_XY, default=False): cv.boolean,
+        cv.Optional(CONF_MIRROR_X, default=False): cv.boolean,
+        cv.Optional(CONF_MIRROR_Y, default=False): cv.boolean,
+        cv.Optional(CONF_COLOR_ORDER, default="BGR"): cv.one_of(
+            *COLOR_ORDERS.keys(), upper=True
+        ),
+    }
+)
```
nielsnl68: I do not see the real need to move this into it own section, all the other settings are basicly for the panel as well.

esphome/components/ili9xxx/display.py
```diff
@@ -101,7 +142,11 @@ def _validate(config):
             cv.Optional(CONF_COLOR_PALETTE_IMAGES, default=[]): cv.ensure_list(
                 cv.file_
             ),
+            cv.Optional(CONF_PANEL_SETUP): PANEL_SCHEMA,
             cv.Optional(CONF_INVERT_DISPLAY): cv.boolean,
+            cv.Optional(CONF_OFFSET_HEIGHT, default=0): cv.int_,
+            cv.Optional(CONF_OFFSET_WIDTH, default=0): cv.int_,
```
nielsnl68:
```suggestion
            cv.Optional(CONF_OFFSET_X, default=0): cv.int_,
            cv.Optional(CONF_OFFSET_Y, default=0): cv.int_,
```

esphome/components/ili9xxx/display.py
```diff
@@ -131,10 +185,15 @@ async def to_code(config):
...
+    if CONF_POWER_SUPPLY in config:
+        ps = await cg.get_variable(config[CONF_POWER_SUPPLY])
+        cg.add(var.set_power_supply(ps))
+
     if CONF_DIMENSIONS in config:
         cg.add(
-            var.set_dimentions(config[CONF_DIMENSIONS][0], config[CONF_DIMENSIONS][1])
+            var.set_dimensions(config[CONF_DIMENSIONS][0], config[CONF_DIMENSIONS][1])
         )
+    cg.add(var.set_offsets(config[CONF_OFFSET_WIDTH], config[CONF_OFFSET_HEIGHT]))
```
nielsnl68:
```suggestion
    cg.add(var.set_offsets(config[CONF_OFFSET_X], config[CONF_OFFSET_Y]))
```

nielsnl68 (changes requested):

esphome/components/ili9xxx/display.py
clydebarrow: This is correct for all the currently supported devices. If something different is required in the future then it should be able to be handled in Python anyway. I am of the understanding from earlier discussions that it is preferred that as much setup-time configuration as possible should be handled in the Python code - is that correct @jesserockz ?

esphome/components/ili9xxx/display.py
clydebarrow: Fixed in next push.

esphome/components/ili9xxx/display.py
clydebarrow: While `x` and `y` are probably more technically correct descriptions, maintaining the naming correspondence with the dimension values make sense to me, and  matches what was done in the st7789v driver. I don't think there is any chance of it being misinterpreted.

esphome/components/ili9xxx/display.py
clydebarrow: Each chip init sequence already sets these values to something. By making the entire group optional, the default values magically coded in the initialisation sequence will remain if the group is absent, otherwise it will be overridden by these settings, even the defaults. If the inner values were moved to the outer schema, they could not have default values since it would be necessary to test if any of them were user-specified.

clydebarrow: Another option would be to move the swap/mirror options under `rotation:` instead (and move `color_order` to the top level.) So it would look like this:

```
    color_order: bgr
    rotation:
      mirror_x: true
      mirror_y: true
      swap_xy: false
```
Alternatively
```
    rotation: 90
```
And that would be mapped to `swap_xy` and `mirror_x` (I think - have to confirm the correct combinations.)

This would eliminate the need for software rotation.

esphome/components/ili9xxx/display.py
clydebarrow: I have moved this config into the CPP code. for various reasons it made sense.

esphome/components/ili9xxx/display.py
clydebarrow: I've gone with the latter option.

esphome/components/ili9xxx/display.py
```diff
@@ -35,6 +36,7 @@ def AUTO_LOAD():
 )
 
 ILI9XXXColorMode = ili9XXX_ns.enum("ILI9XXXColorMode")
+ColorOrder = display.display_ns.enum("ColorMode")
 
 MODELS = {
     "M5STACK": ili9XXX_ns.class_("ILI9XXXM5Stack", ili9XXXSPI),
```
nielsnl68: I do not know if i missed up here: but `ili9xxxSPI` should be `ILI9XXXDisplay` i think.

nielsnl68 (changes requested):

esphome/components/ili9xxx/ili9xxx_display.h
```diff
@@ -4,10 +4,13 @@
 #include "ili9xxx_defines.h"
 #include "ili9xxx_init.h"
 
+#ifdef USE_POWER_SUPPLY
+#include "esphome/components/power_supply/power_supply.h"
+#endif
```
jesserockz: Not needed anymore?

jesserockz (changes requested):
Can you split the power supply stuff into a separate PR?

esphome/components/ili9xxx/display.py
```diff
@@ -26,38 +30,83 @@ def AUTO_LOAD():
...
+            config[CONF_MIRROR_Y] = False
+        elif rotation == 270:
+            config[CONF_SWAP_XY] = True
+            config[CONF_MIRROR_X] = False
+            config[CONF_MIRROR_Y] = True
+        elif rotation == 180:
+            config[CONF_SWAP_XY] = False
+            config[CONF_MIRROR_X] = True
+            config[CONF_MIRROR_Y] = True
+
```
jesserockz: The `config` dict is not a place to add new keys to during a validation to use later. It should only set keys that are part of the schema already.

For example, if `rotation` has potential to be an object, then this validation should set the values inside the object, not on the top level config.

esphome/components/ili9xxx/display.py
```diff
@@ -101,7 +161,21 @@ def _validate(config):
             cv.Optional(CONF_COLOR_PALETTE_IMAGES, default=[]): cv.ensure_list(
                 cv.file_
             ),
-            cv.Optional(CONF_INVERT_DISPLAY): cv.boolean,
+            cv.Optional(CONF_INVERT_DISPLAY): cv.invalid(
+                "invert_display is deprecated - use invert_colors "
```
jesserockz:
```suggestion
                "'invert_display' has been removed - please use 'invert_colors'"
```

esphome/components/ili9xxx/display.py
clydebarrow: The problem is that the top-level `display` component expects CONF_ROTATION, if present, to be a scalar value. The only alternative I can think of is to move the sub-schema options (`SWAP_XY` etc) under a different key (`transform:` ?) and make it exclusive with ROTATION.

clydebarrow: And the problem with that is that specifying `rotation` would not then map to the flags necessary to set up hardware rotation. That would have to be moved into the C++ code.

Or I could disallow use of `rotation` completely and just allow `transform` with appropriate documentation.

jesserockz: Having a CONF_TRANSFORM that is exclusive with CONF_ROTATION would be fine.

jesserockz (approved):
LGTM

PR #5411, opened 2023-09-20
Fix RC Switch protocol not transmitting correctly via IR

### LINKED ISSUE #2044: remote.rc_switch supports IR receiving but not IR transmitting
**Operating environment/Installation (Hass.io/Docker/pip/etc.):**

Docker

**ESP (ESP32/ESP8266, Board/Sonoff):**

ESP32

**ESPHome version (latest production, beta, dev branch)**

ESPHome version 1.17.1 

**Affected component:**

* https://esphome.io/components/remote_transmitter.html
* https://esphome.io/components/remote_receiver.html

**Description of problem:**

I have a few IR remotes that ESPHome's IR decoder decodes as `RCSwitch Raw`.
For example, here are a few of the decodes from my Lasko heater's IR remote:
```
RCSwitch Raw: protocol=4 data='11011000000'
RCSwitch Raw: protocol=4 data='11011000010'
RCSwitch Raw: protocol=4 data='11011000100'
RCSwitch Raw: protocol=4 data='11011000001'
RCSwitch Raw: protocol=4 data='11011001000'
RCSwitch Raw: protocol=4 data='11011010000
```

However, the RCSwitch Raw service `remote_transmitter.transmit_rc_switch_raw` does not support sending IR commands. All calls to `remote_transmitter.transmit_rc_switch_raw` with an IR transmitter appear to do nothing. Additionally the `RCSwitchBase::transmit` function appears to set the frequency to 0 for every transmission:
https://github.com/esphome/esphome/blob/dev/esphome/components/remote_base/rc_switch_protocol.cpp#L57

1.  If the `RCSwitch` remote type is meant to be used for RF only, it should not be used when decoding IR signals.
2. If the `RCSwitch` remote type is meant to support IR, then the `remote_transmitter.transmit_rc_switch_raw` function should work, and not have the frequency set to 0.

Also, when using `dump: all` in `remote_receiver` raw codes are not printed if another decoder is able to successfully decode the signal. Since my codes are being decoded by `RCSwitch`, the raw codes are not printed. This can be worked around by explicitly setting `dump: raw` to forcefully show all raw codes.

Sending any other IR codes using `remote_transmitter.transmit_raw` or `remote_transmitter.transmit_nec` on this node works as expected.

**Problem-relevant YAML-configuration entries:**
```yaml
substitutions:
  node_name: remote

packages:
  device_base: !include common/device_base.yaml

esphome:
  name: ${node_name}
  platform: ESP32
  board: nodemcu-32s

remote_receiver:
  pin:
    number: GPIO34
    inverted: True
    mode: INPUT_PULLUP
  dump: all

remote_transmitter:
  pin: GPIO32
  carrier_duty_percent: 50%
  id: ir_trans

  services:
    - service: remote_send_raw_command
      variables:
        command: int[]
      then:
        - remote_transmitter.transmit_raw:
            carrier_frequency: 38kHz
            code: !lambda 'return command;'
    - service: remote_send_nec_command
      variables:
        address: int
        command: int
      then:
            address: !lambda 'return address;'
            command: !lambda 'return command;'
 
binary_sensor:
  - platform: gpio
    pin: 
      number: GPIO26
      mode: INPUT_PULLUP
      inverted: True
    id: tv_power
    internal: true # don't expose to HA
    filters:
      - delayed_off: 1s # debounce, when powering on, the TV turns off briefly
      - delayed_on: 100ms

switch:
  - platform: template
    name: "TV Power"
    lambda: |-
      if (id(tv_power).state) {
        return true;
      } else {
        return false;
      }
    turn_on_action:
          address: 0x20DF
          command: 0x54AB
    turn_off_action:
          address: 0x20DF
          command: 0xA45B

  - platform: template
    name: "test switch"
    turn_on_action:
      - remote_transmitter.transmit_rc_switch_raw:
          protocol: 4
          code: '11011000000'
          transmitter_id: ir_trans
          repeat: 5
```

**Logs (if applicable):**

```
PASTE DEBUG LOG HERE

```

**Additional information and things you've tried:**

I've tried decoding the Raw IR from these remotes in [IRScrutinizer](https://github.com/bengtmartensson/IrScrutinizer) which detects the protocol as `F12_relaxed`. If this is correct (I'm not 100% sure), then it would imply that this signal is being incorrectly decoded as RCSwitch and should be RAW since ESPHome's IR libraries do not natively support `F12_relaxed`.

michd:

Enables the use of RC Switch protocol with IR Remote transmitters, by setting the carrier frequency to 38kHz.

The RC Switch protocol can be used for both RF and IR remotes. RF Remotes don't require generating a carrier signal, so, as pointed out in the [remote transmitter configuration variables docs](https://esphome.io/components/remote_transmitter#configuration-variables), that carrier frequency can be nullified by setting the duty cycle to 100%.

Since this protocol is also used with IR devices, a 50% duty cycle should still result in a carrier frequency being generated. Therefore, this sets the frequency to 38kHz in the protocol implementation, allowing the user to cancel it out in YAML config.

When this carrier frequency is left to 0, then even with the duty cycle set to 50%, infrared transmitters don't include the carrier, and thus de-modulators don't pick up the signal at all.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [#2044](https://github.com/esphome/issues/issues/2044)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** n/a, docs already say to use `carrier_duty_cycle: 100%` for RF transmitters.

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
remote_transmitter:
  pin:
    number: GPIO13
  carrier_duty_percent: 50%

button:
  - platform: template
    name: "Power"
    on_press:
      remote_transmitter.transmit_rc_switch_raw:
        # Reverse engineered protocol for
        # Dreo Solaris Slim H3 Space Heater remote
        code: '10110000001'
        protocol:
          pulse_length: 60
          sync: [24, 4]
          zero: [8, 20]
          one: [23, 5]
        repeat:
          times: 5
          wait_time: 8ms
# Other buttons omitted for brevity
```

## Checklist:

### DIFF
--- a/esphome/components/remote_base/rc_switch_protocol.cpp
+++ b/esphome/components/remote_base/rc_switch_protocol.cpp
@@ -54,7 +54,7 @@ void RCSwitchBase::sync(RemoteTransmitData *dst) const {
   }
 }
 void RCSwitchBase::transmit(RemoteTransmitData *dst, uint64_t code, uint8_t len) const {
-  dst->set_carrier_frequency(0);
+  dst->set_carrier_frequency(38000);
   this->sync(dst);
   for (int16_t i = len - 1; i >= 0; i--) {
     if (code & ((uint64_t) 1 << i)) {


michd:
Hi! Just wondering if there's anything further I should do on this one. :slightly_smiling_face:

michd:
I'm still invested in getting this merged so I don't have to manually patch it again every time I update esphome. :)

roger-:
Can confirm this worked for me with a Lasko standing fan (just needed `pulse_length: 440` and `sync: [3, 18]`).

I run ESPHome in Docker, so as a hack I apply the patch by running the following whenever I restart the container:

```bash
docker exec -it <contained id> sed -i 's/set_carrier_frequency(0)/set_carrier_frequency(38000)/g' /esphome/esphome/components/remote_base/rc_switch_protocol.cpp
```

Would be great to see this merged.

mr-sneezy:
Hey anyone, is this issue with RC_switch_raw for IR transmitters fixed yet (two years on) ?
The reason I ask is I'm setting up an IR blaster based on ESP8266 D1 Mini and the Lolin IR shield, and on rc_switch_raw output I see no 38khz carrier on the pulse data in TX on my oscilloscope. 
I do see the 38Khz carrier on other ESPhome IR protocols and they get received by their respective devices so it seems to be an RC_switch_raw issue...

jesserockz (approved):

mr-sneezy:
Great to see some activity regarding the problem. I'd like to test the PR from Dev.
I believe I need to add this to my ESPhome yaml for the device and recompile but it did not work so far (as in still no 38khz).
Did I interpret the method incorrectly to fetch a PR to test in the component?
```
external_components:
# shorthand
  - source: github://pr#5411
    components: [ remote_transmitter ]
```

PR #5413, opened 2023-09-21
dallas: limit addresses to 64 bits

oddstr13:

Limits dallas temperature sensor address to 64 bits.
Currently, all bits that exceed this limit is silently dropped, causing unexpected behavior for users moving from Arduino where address is given as an array, which drops bytes on the other end of the address.

 https://discord.com/channels/429907082951524364/1154371632731799603

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Invalid

dallas:
  pin: 2

sensor:
  - id: ds_temp
    platform: dallas
    address: 0xcd3ce1045784a128067925
```

```yaml
# Valid

dallas:
  pin: 2

sensor:
  - id: ds_temp
    platform: dallas
    address: 0xcd3ce1045784a128
```

## Checklist:

### DIFF
--- a/esphome/components/dallas/sensor.py
+++ b/esphome/components/dallas/sensor.py
@@ -24,7 +24,7 @@ CONFIG_SCHEMA = cv.All(
     ).extend(
         {
             cv.GenerateID(CONF_DALLAS_ID): cv.use_id(DallasComponent),
-            cv.Optional(CONF_ADDRESS): cv.hex_int,
+            cv.Optional(CONF_ADDRESS): cv.hex_uint64_t,
             cv.Optional(CONF_INDEX): cv.positive_int,
             cv.Optional(CONF_RESOLUTION, default=12): cv.int_range(min=9, max=12),
         }


oddstr13:
There are probably more tests we can do to validate the addresses of DS18B20, DS18S20 and DS1820 sensors.
DS18B20 is identified by 0x28, and DS1820/DS18S20 are identified with 0x10 for example.
There are likely more family codes that work with the same temperature library, so care must be taken there if that kind of validation is desired.
The address also contains a 8-bit CRC, which we could verify.

https://www.analog.com/media/en/technical-documentation/data-sheets/DS18B20.pdf (see section «64-BIT Lasered ROM code»)

esphome/components/dallas/sensor.py
```diff
@@ -24,7 +24,7 @@
     ).extend(
         {
             cv.GenerateID(CONF_DALLAS_ID): cv.use_id(DallasComponent),
-            cv.Optional(CONF_ADDRESS): cv.hex_int,
+            cv.Optional(CONF_ADDRESS): cv_dallas_address,
```
jesserockz: We have a validator for this already =)
```suggestion
            cv.Optional(CONF_ADDRESS): cv.hex_uint64_t,
```

oddstr13: Not familiar with the magic yet, thanks :)

jesserockz (approved):

PR #5423, opened 2023-09-24
Feat/component poller suspend

### LINKED ISSUE #2301: Suspend update for PollingComponent

When a device is playing music or voice and has a configured screen, there is a temporary pause in the sound playback when the screen refreshes. It appears as a quarter-second click. The following entries can be seen in the logs:
```
[12:06:00][V][component:204]: Component st7789v.display took a long time for an operation (0.37 s).
[12:06:00][V][component:205]: Components should block for at most 20-30ms.
```

I suggest adding a component with the ability to temporarily pause its operation (refresh). It doesn't necessarily have to be done for the screen; it would be beneficial to have this feature for all components. Here's an example of implementation:
```yaml
then:
  - component.suspend: main_display
  - delay: 2s
  - component.update: main_display
  - delay: 2s
  - component.resume: main_display
```
It is important to allow for manual updates of a component even when it is in a suspended state.

**Additional context**

Use voice_assistant: with display:

xdecock:

This pr aims at allowing suspend / resume of polling component update, this is mainly in place to help out with "slow component" when managing audio interractions.

PS: Took a bit of time to understand how to start the version of my pull request, tested and it does work as i expected

Tested on a D1 Mini & and C3 Pico

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2301

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3205

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
esphome:
  name: test
  friendly_name: test
  platformio_options:
    board_build.flash_mode: dio
  on_boot:
    then:
      - logger.log: Wait
      - delay: 10s
      - component.suspend: adc_pin_3
      - logger.log: Stopped
      - delay: 10s
      - logger.log: Resume
      - component.resume: adc_pin_3
      - delay: 10s
      - logger.log: Stop 2
      - component.suspend: adc_pin_3
      - delay: 10s
      - logger.log: Start Slow
      - component.resume:
          id: adc_pin_3
          update_interval: 10s


esp32:
  board: lolin_c3_mini
  framework:
    type: esp-idf

# Enable logging
logger:
  level: VERBOSE
  hardware_uart: USB_SERIAL_JTAG

sensor:
  - platform: adc
    pin: GPIO3
    name: "Test 1"
    update_interval: 2s
    id: adc_pin_3

```

## Checklist:

### DIFF
--- a/esphome/automation.py
+++ b/esphome/automation.py
@@ -13,7 +13,6 @@ from esphome.const import (
     CONF_TIME,
     CONF_UPDATE_INTERVAL,
 )
-from esphome.core import TimePeriod
 from esphome.schema_extractors import SCHEMA_EXTRACT, schema_extractor
 from esphome.util import Registry
 
@@ -333,7 +332,7 @@ async def component_suspend_action_to_code(config, action_id, template_arg, args
 )
 async def component_resume_action_to_code(config, action_id, template_arg, args):
     comp = await cg.get_variable(config[CONF_ID])
-    var = cg.new_Pvariable(action_id, template_arg, comp);
+    var = cg.new_Pvariable(action_id, template_arg, comp)
     if CONF_UPDATE_INTERVAL in config:
         cg.add(var.set_update_interval(config[CONF_UPDATE_INTERVAL]))
     return var
--- a/esphome/automation.py
+++ b/esphome/automation.py
@@ -326,7 +326,9 @@ async def component_suspend_action_to_code(config, action_id, template_arg, args
     maybe_simple_id(
         {
             cv.Required(CONF_ID): cv.use_id(cg.PollingComponent),
-            cv.Optional(CONF_UPDATE_INTERVAL): cv.positive_time_period_milliseconds,
+            cv.Optional(CONF_UPDATE_INTERVAL): cv.templatable(
+                cv.positive_time_period_milliseconds
+            ),
         }
     ),
 )
--- a/esphome/automation.py
+++ b/esphome/automation.py
@@ -69,6 +69,8 @@ WhileAction = cg.esphome_ns.class_("WhileAction", Action)
 RepeatAction = cg.esphome_ns.class_("RepeatAction", Action)
 WaitUntilAction = cg.esphome_ns.class_("WaitUntilAction", Action, cg.Component)
 UpdateComponentAction = cg.esphome_ns.class_("UpdateComponentAction", Action)
+SuspendComponentAction = cg.esphome_ns.class_("SuspendComponentAction", Action)
+ResumeComponentAction = cg.esphome_ns.class_("ResumeComponentAction", Action)
 Automation = cg.esphome_ns.class_("Automation")
 
 LambdaCondition = cg.esphome_ns.class_("LambdaCondition", Condition)
@@ -301,7 +303,7 @@ async def lambda_action_to_code(config, action_id, template_arg, args):
 
 @register_action(
     "component.suspend",
-    UpdateComponentAction,
+    SuspendComponentAction,
     maybe_simple_id(
         {
             cv.Required(CONF_ID): cv.use_id(cg.PollingComponent),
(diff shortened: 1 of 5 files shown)


nagyrobi:
Would it be possible to start the poller in suspended state at boot?

The use case is, for example having to wait for a slave device to start up (as it's slower than the esp), before starting to poll it.

xdecock:
> Would it be possible to start the poller in suspended state at boot?
> 
> The use case is, for example having to wait for a slave device to start up (as it's slower than the esp), before starting to poll it.

I thought about that, and imo, the most elegant way to do that, would be to add an action:
```yaml
  component.set_update_interval;
    id: component_id
    interval: xxxs
```

This mr do allow to "dynamically update the interval" 

The only thing i did not understand is how to create an action with that template.

That being, said, if i'm guided about how to implement those type of actions, i would probably do it as a separate pull request.

I'd then solve this exact scenario this way:
```
fast_component:
  id: fc_id
  update_interval: never
  
slow_component:
  id: sc_id
  on_startup:
    - component.set_interval:
      id: fc_id
      update_interval: 5s 
      resume: false #not sure about this or the next line
    - component.resume: fc_id
```

jesserockz:
Instead of another action to change the interval, you could add an `Optional` config item to the `start` action for `update_interval`

xdecock:
> Instead of another action to change the interval, you could add an `Optional` config item to the `start` action for `update_interval`

So more something in the vein of

```yaml
component.resume:
  id: fc_id
  update_interval: 3s
```

An extreme version would allow to merge those two actions via

```yaml
component.set_update_interval:
  id: fc_id
  update_interval: never

component.set_update_interval:
  id: fc_id
  update_interval: 5s

# possible helpers
component.set_update_interval:
  id: fc_id
  action: suspend

component.set_update_interval:
  id: fc_id
  action: resume
```

If this would work, i'd be happy with a pointer to the correct class to declare the correct structure in python, I'm quite happy to update my mr but i'm struggling a bit to find how things work.

xdecock:
Fast tentative,

I'll debug it later today if it doesn't work, if the logic is sound, i'll redo my local tests, and update the documentation accordingly

xdecock:
Updated the example, seems it works correctly on my local devices. i've struggled a bit to understand the code generation logic, so i'm happy with any correction i might have to do

esphome/automation.py
```diff
@@ -303,6 +306,38 @@ async def component_update_action_to_code(config, action_id, template_arg, args)
...
+    return cg.new_Pvariable(action_id, template_arg, comp)
+
+
+@register_action(
+    "component.resume",
+    ResumeComponentAction,
+    maybe_simple_id(
+        {
+            cv.Required(CONF_ID): cv.use_id(cg.PollingComponent),
+            cv.Optional(CONF_UPDATE_INTERVAL): cv.positive_time_period_milliseconds,
```
jesserockz: In the cpp class you define this as `templatable`, but the config side does not allow the use of a lambda to set.
```suggestion
            cv.Optional(CONF_UPDATE_INTERVAL): cv.templatable(cv.positive_time_period_milliseconds),
```

This also needs some changes in function to match

jesserockz (changes requested):

esphome/automation.py
xdecock: Updated, along side the test it runs as (i) expected it to work

xdecock:
> Would it be possible to start the poller in suspended state at boot?
> 
> The use case is, for example having to wait for a slave device to start up (as it's slower than the esp), before starting to poll it.

this is now possible by starting with update_interval: never, and doing a component.resume when the slow component booted up, and set the interval with the resume action

jesserockz (approved):

PR #5436, opened 2023-09-27
Adding display features

clydebarrow:

Some display improvements that should have no impact on existing code, but will facilitate further enhancements.

* ~Hoist `PollingComponent` out of individual display classes and make it a parent class of `DisplayBuffer`. All existing display components already implement this so no breaking change.~ (moved to another PR)
* Add a new function to allow potentially efficient drawing of multiple pixels. The default implementation will work for all displays but is slow, but can be overridden by individual components (with knowledge of buffer format) to optimise this procedure.

The second change is aimed at efficient implementation of the LVGL display interface. This has been tested with LVGL.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:
This is a sample implemention of the draw function for ili9xx. Runs  5-10 times faster than the reference implementation (for 16 bit colour.)

```
void ILI9XXXDisplay::draw_pixels_in_window(int x_start, int y_start, int w, int h, const uint8_t *ptr,
                                           display::ColorOrder order, display::ColorBitness bitness, bool big_endian,
                                           int x_offset, int y_offset, int x_pad) {
  // optimal case is when everybody uses 16 bit big-endian colour format. Anything else we hand off.
  if (this->buffer_color_mode_ != BITS_16 || bitness != display::COLOR_BITNESS_565 ||
      order != display::COLOR_ORDER_RGB || !big_endian) {
    DisplayBuffer::draw_pixels_in_window(x_start, y_start, w, h, ptr, order, bitness, big_endian);
    return;
  }

  ESP_LOGV(TAG, "copy with rotation %d degrees", (int) this->rotation_);
  size_t line_stride = x_start + w + x_pad;
  uint16_t *src_ptr;
  uint16_t *dst_ptr;
  switch (this->rotation_) {
    case display::DISPLAY_ROTATION_0_DEGREES: {
      src_ptr = ((uint16_t *) ptr) + y_offset * line_stride + x_offset;
      dst_ptr = ((uint16_t *) this->buffer_) + y_start * this->width_ + x_start;
      for (int y = 0; y != h; y++) {
        memcpy(dst_ptr, src_ptr, w * 2);
        src_ptr += line_stride;
        dst_ptr += this->width_;
      }
      return;
    }
    case display::DISPLAY_ROTATION_180_DEGREES: {
      for (int py = 0; py != h; py++) {
        src_ptr = ((uint16_t *) ptr) + (y_offset + py) * line_stride + x_offset;
        dst_ptr = ((uint16_t *) this->buffer_) + (this->height_ - y_start - py) * this->width_ - x_start;
        for (int px = 0; px != w; px++) {
          *--dst_ptr = *src_ptr++;
        }
      }
      break;
    }
    case display::DISPLAY_ROTATION_90_DEGREES: {
      for (int py = 0; py != h; py++) {
        src_ptr = ((uint16_t *) ptr) + (y_offset + py) * line_stride + x_offset;
        dst_ptr = ((uint16_t *) this->buffer_) + x_start * this->width_ + this->width_ - y_start - py - 1;
        for (int px = 0; px != w; px++) {
          *dst_ptr = *src_ptr++;
          dst_ptr += this->width_;
        }
      }
      break;
    }
    case display::DISPLAY_ROTATION_270_DEGREES: {
      for (int py = 0; py != h; py++) {
        src_ptr = ((uint16_t *) ptr) + (y_offset + py) * line_stride + x_offset;
        dst_ptr = ((uint16_t *) this->buffer_) + (this->height_ - x_start) * this->width_ + y_start + py;
        for (int px = 0; px != w; px++) {
          dst_ptr -= this->width_;
          *dst_ptr = *src_ptr++;
        }
      }
      break;
    }
  }


```

nielsnl68:
@clydebarrow Would it not much better and faster when the display it self handles the rotation?

esphome/components/display/display_color_utils.h
```diff
@@ -3,8 +3,13 @@
...
 namespace display {
-enum ColorOrder : uint8_t { COLOR_ORDER_RGB = 0, COLOR_ORDER_BGR = 1, COLOR_ORDER_GRB = 2 };
-enum ColorBitness : uint8_t { COLOR_BITNESS_888 = 0, COLOR_BITNESS_565 = 1, COLOR_BITNESS_332 = 2 };
+enum ColorOrder : uint8_t { COLOR_ORDER_RGB = 0, COLOR_ORDER_BGR = 1, COLOR_ORDER_GRB = 2, COLOR_ORDER_UNKNOWN = 255 };
+enum ColorBitness : uint8_t {
+  COLOR_BITNESS_888 = 0,
+  COLOR_BITNESS_565 = 1,
+  COLOR_BITNESS_332 = 2,
+  COLOR_BITNESS_UNKNOWN = 255
+};
```
nielsnl68: Could you use the same *enum* values as used in ili9xxx?

```c++
enum ILI9XXXColorMode {
  BITS_8 = 0x08,  // would corspondent to COLOR_BITNESS_332
  BITS_8_INDEXED = 0x09,
  BITS_16 = 0x10, // would corspondent to COLOR_BITNESS_565
  BITS_18 = 0x12, // would corspondent to COLOR_BITNESS_666 (*)
  BITS_24 = 0x18, //  would corspondent to COLOR_BITNESS_888 (*)
  BITS_UNKNOWN = 0xFF, //  would corspondent to COLOR_BITNESS_UNKNOWN  (*)
};
```

esphome/components/display/display_buffer.h
```diff
@@ -12,7 +12,7 @@
 namespace esphome {
 namespace display {
 
-class DisplayBuffer : public Display {
+class DisplayBuffer : public Display, public PollingComponent {
```
nielsnl68: I do like to move the `PollingComponent` to the *DisplayBuffer. But i'm not sure of this should be in this PR.

esphome/components/display/display_buffer.h
```diff
@@ -22,6 +22,40 @@ class DisplayBuffer : public Display {
   /// Set a single pixel at the specified coordinates to the given color.
   void draw_pixel_at(int x, int y, Color color) override;
 
+  /// Given a buffer of pixels encoded in the nominated format, draw these into the display.
+  /// The naive implementation here will work in all cases, but can be overridden by sub-classes
+  /// in order to optimise the procedure.
+  /// TODO: handle one-bit-per-pixel monochrome
+  virtual void draw_pixels_in_window(int x_start, int y_start, int w, int h, const uint8_t *ptr, ColorOrder order,
```
nielsnl68, line 44: Sorry, i do not think that the ColorOrder, bitness and big_endian should be handled by this function. To me it is just a Base setting that should be set by the driver it self.

nielsnl68 (changes requested):
top work

clydebarrow:
> @clydebarrow Would it not much better and faster when the display it self handles the rotation?

Yes, it would. But not all drivers can handle rotation - the ili9xx series could do it easily(it's not currently implemented) but that's not going to be true for all displays. So it needs to be something implemented on a per-driver basis.

The idea would be for the upper-level (`DisplayBuffer`) to have a user-specified rotation set, and the driver to optionally provide an 'effective rotation'. If the driver can implement the rotation in hardware, then the effective rotation is set to 0 degrees. The effective rotation is what will be used by the `draw_pixel` functions to do the mapping rather than the user-specified rotation.

That's another task on top of what I'm doing here and should be a separate PR.

esphome/components/display/display_buffer.h
clydebarrow: Sure, that is quite separate to the other stuff, I just bundled it in because it was easy and I thought non-controversial, but could just as easily be a separate PR.

There is also the question of whether it should be in the `DisplayBuffer` or `Display` class - I have no strong opinion either way but if there's no good reason to limit it to `DisplayBuffer` then I would guess it should go at the higher level.

esphome/components/display/display_color_utils.h
clydebarrow: Those values were already defined in `display_color_utils.h` (except `UNKNOWN`) and are used by the color conversion functions in `ColorUtils` so I did not want to rework all that code.

Currently the indexed color case is not provided for there - that's another can of worms so for now I just left it to the driver.

esphome/components/display/display_buffer.h
clydebarrow: I think you misunderstand - this is not setting anything in the driver. Let me explain (at some length):

The use case for this is where there is a rendering component (i.e. LVGL) that can draw stuff into some kind of buffer. Then there is a display component that is responsible for displaying stuff on a screen. The objective is that any rendering component in ESPHome should be able to be plugged into any display component and it should Just Work. The secondary objective is that it should be _possible_ to make it not just work, but work fast, if the right choices are made, but it should never not work.

Each driver has potentially a different internal buffer layout which is unknown (and unknowable) to the rendering component.  So to make it Just Work the base case is that every pixel in the render buffer is copied, with colour translation, to the display (via `draw_pixel_at` which is implemented on a per-driver basis.) Using that mechanism, the rendering component and the display component are coupled only the notion of x-y position and colour space.

However that is going to be very slow in many cases. So the `draw_pixels_in_window` method allows the rendering component to say to the display driver 'here's a block of pixels stored in a buffer, corresponding to this display window, and this is how they are encoded. Go show it'. The default implementation in `DisplayBuffer` just falls back to the naive pixel-by-pixel implementation, but as a virtual method it can be re-implemented on a per-driver basis to make things faster.

Now if the render buffer and the display buffer use different colour encodings e.g. then there isn't much improvement that can be made, but if e.g. they both use RGB565 big-endian format, then the data can be copied straight from one buffer to another without colour conversion. Even better if the effective rotation is zero, `memcpy` can be used. That's all shown in the example implementation I posted - https://github.com/esphome/esphome/pull/5436#issuecomment-1737302151.

So the colour format information in the function call isn't setting anything in the driver - it has its own internal notion of colour encoding - but a description of how the supplied render buffer is encoded.

This means that to get the full performance a user will have to choose the right settings for both the display driver and the renderer in the YAML - but if they don't match it still works - just not as fast as it might.

I used the already-defined types from `display_color_utils.h` so that the existing colour conversion routines could be used - no sense in re-inventing the wheel.

nielsnl68 (approved):

clydebarrow:
I have removed the `PollingComponent` hoisting and will submit a separate PR - I agree with @ayufan that it should go in `Display`, but the Python code also needs work as the component registration should be handled by the display registration.
I think this is now ready to go.

esphome/components/display/display_color_utils.h
clydebarrow: Actually the indexed case is probably never going to work as there would need to be agreement between the renderer and the display driver on the palette to be used. I think that belongs in the Too Hard basket. I expect the majority of cases to use 16 bit RGB565 big-endian so that's where the focus should be. If saving RAM by use of an 8 bit display buffer matters, then it's probably not a chip suited to running something like LVGL.

esphome/components/display/display_buffer.h
```diff
@@ -22,38 +22,65 @@ class DisplayBuffer : public Display, public PollingComponent {
...
+              color_value = ptr[source_idx_mod + 0] + (ptr[source_idx_mod + 1] << 8) + (ptr[source_idx_mod + 2] << 16);
+            }
+            break;
+          case COLOR_BITNESS_A888:
+            source_idx_mod = source_idx * 4;
+            color_value = (ptr[source_idx_mod + 2] << 16) + (ptr[source_idx_mod + 1] << 8) + ptr[source_idx_mod + 0];
             break;
         }
-        this->draw_pixel_at(x, y, ColorUtil::to_color(color_value, order, bitness));
+        this->draw_pixel_at(x + x_start, y + y_start, ColorUtil::to_color(color_value, order, bitness));
```
nielsnl68, line 79: I could not find it, but make you sure that `x + x_start` and ` y + y_start` does not go outside the display window?

esphome/components/ili9xxx/ili9xxx_display.h
```diff
@@ -48,6 +48,9 @@ class ILI9XXXDisplay : public display::DisplayBuffer,
 
  protected:
   void draw_absolute_pixel_internal(int x, int y, Color color) override;
+  virtual void draw_pixels_in_window(int x_start, int y_start, int w, int h, const uint8_t *ptr,
```
nielsnl68:
``` suggestion
   void draw_pixels_in_window(int x_start, int y_start, int w, int h, const uint8_t *ptr,
```
i belief the `virtual` is only needed the first time this method was declared

clydebarrow: Correct, I already fixed that.

esphome/components/display/display_buffer.h
clydebarrow: Same way that the existing code (doesn't) ensure that -  it relies on the calling code to pass valid values. 

However looking at `draw_pixel_at` while it does not check that the x/y coords are within the display boundaries, it does check that they are within the clipping rectangle, _if one is set_. As far as I can tell the clipping rectangle is _never_ set, i.e. there are no usages of the clipping functions I can find in the code base.

Right now `get_clipping` returns an unset `Rect` if no clipping is set. It should probably instead return a rectangle representing the display limits. Want me to fix that?

In any case `draw_pixels_in_window()` should also check the clipping rectangle. I'll attend to that.

nielsnl68: That is right, i added the clipping future to be used with my `switchplate` component. That component was using it a lot.
Maybe we should set the default clipping regio to the screensize?

clydebarrow: Yes, agree - shall I add that to this PR?

nielsnl68: yeah, i belief it is better here.

PR #5438, opened 2023-09-27
Add combination sensor and remove absorbed kalman_combinator component

kahrendt:

This component adds a combination sensor that mimics the combine state of several sensor helper functionality in Home Assistant.

It is a breaking change, as it incorporates the kalman_combinator component. I have added a warning and instructions to fix this change to that component's sensor.py.

It also adds an additional type not present in HA, which is a linear combination of sensors. This allows the user to specify a constant that the sensor state is multiplied by before summing. This allows (among other things) you to subtract a sensor state from another sensor state without needing a template sensor and lambdas.

If there is a preference for leaving the kalman_combinator component as is, I'll make that adjustment. I have added @Cat-Ion as a codeowner, since their Kalman combinator code is unmodified. I will also happily adjust that depending on their wishes.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3216

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

  - platform: combination
    type: kalman
    name: Kalman-filtered temperature
    process_std_dev: 0.00139
    sources:
      - source: scd30_temperature
        error: !lambda |-
          return 0.4 + std::abs(x - 25) * 0.023;
      - source: scd4x_temperature
        error: 1.5
  - platform: combination
    type: linear
    name: Linearly combined temperatures
    sources:
      - source: scd30_temperature
        constant: !lambda |-
          return 0.4 + std::abs(x - 25) * 0.023;
      - source: scd4x_temperature
        constant: 1.5
  - platform: combination
    type: max
    name: Max of combined temperatures
    sources:
      - source: scd30_temperature
      - source: scd4x_temperature
  - platform: combination
    type: mean
    name: Mean of combined temperatures
    sources:
      - source: scd30_temperature
      - source: scd4x_temperature
  - platform: combination
    type: median
    name: Median of combined temperatures
    sources:
      - source: scd30_temperature
      - source: scd4x_temperature
  - platform: combination
    type: min
    name: Min of combined temperatures
    sources:
      - source: scd30_temperature
      - source: scd4x_temperature
  - platform: combination
    type: most_recently_updated
    name: Most recently updated of combined temperatures
    sources:
      - source: scd30_temperature
      - source: scd4x_temperature
  - platform: combination
    type: range
    name: Range of combined temperatures
    sources:
      - source: scd30_temperature
      - source: scd4x_temperature
  - platform: combination
    type: sum
    name: Sum of combined temperatures
    sources:
      - source: scd30_temperature
      - source: scd4x_temperature

```

## Checklist:


jesserockz:
It would be a bit cleaner if `CombinationComponent` was a class with `handle_new_value` as virtual and a sub class for each type that overrode that function (and anything else required). This way all of the type specific variables can be self contained in the right class.

kahrendt:
> It would be a bit cleaner if `CombinationComponent` was a class with `handle_new_value` as virtual and a sub class for each type that overrode that function (and anything else required). This way all of the type specific variables can be self contained in the right class.

Good idea! I refactored the code to reflect that structure.

jesserockz (approved):

jesserockz:
Thanks for making those changes. Much cleaner now =)

PR #5439, opened 2023-09-27
Add spi support for ade7953

angelnu:

Sensor ADE7953 supports multiple buses. This adds support for SPI to the existing i2c implementation.

## Types of changes

- [ ] Other

- implements https://github.com/esphome/feature-requests/issues/2067
- dependency for https://github.com/esphome/feature-requests/issues/1784

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3217

## Test Environment

- [X] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml for spi

sensor:
  - platform: ade7953_spi
    id: ade7953_1
    spi_id: spi_bus_1
    cs_pin: GPIO04
    irq_pin: GPIO35
    voltage:
      name: "Voltage"
      filters:
      - delta: 1
    current_a:
      name: "Current A"
      filters:
      - delta: 0.01
    current_b:
      name: "Current B"
      filters:
      - delta: 0.01
    active_power_a:
      name: "Power A"
      id: active_power_1
      # active_power is inverted, so multiply by -1
      filters:
      - multiply: -1
      - delta: 1
    active_power_b:
      name: "Power B"
      id: active_power_2
      # active_power is inverted, so multiply by -1
      filters:
      - multiply: -1
      - delta: 1
    update_interval: 1s

```

## Checklist:


esphome/components/ade7953_base/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+            default="1x",
+        ): cv.one_of(*PGA_GAINS, lower=True),
+        cv.Optional(CONF_CURRENT_GAIN_A, default=0x400000): cv.hex_int_range(
+            min=0x100000, max=0x800000
+        ),
+        cv.Optional(CONF_CURRENT_GAIN_B, default=0x400000): cv.hex_int_range(
+            min=0x100000, max=0x800000
+        ),
+    }
+).extend(cv.polling_component_schema("1s"))
```
jesserockz: This needs to stay at 60 seconds default
```suggestion
).extend(cv.polling_component_schema("60s"))
```

esphome/components/ade7953_base/ade7953_base.cpp
```diff
@@ -0,0 +1,114 @@
...
+  ade_read_32(AVGAIN_32, &val);
+  ESP_LOGCONFIG(TAG, "  AVGAIN_32: 0x%08jX", (uintmax_t) val);
+  ade_read_32(AWGAIN_32, &val);
+  ESP_LOGCONFIG(TAG, "  AWGAIN_32: 0x%08jX", (uintmax_t) val);
+  ade_read_32(BIGAIN_32, &val);
+  ESP_LOGCONFIG(TAG, "  BIGAIN_32: 0x%08jX", (uintmax_t) val);
+  ade_read_32(BVGAIN_32, &val);
+  ESP_LOGCONFIG(TAG, "  BVGAIN_32: 0x%08jX", (uintmax_t) val);
+  ade_read_32(BWGAIN_32, &val);
+  ESP_LOGCONFIG(TAG, "  BWGAIN_32: 0x%08jX", (uintmax_t) val);
```
jesserockz: `dump_config` should not really be doing comms, it should only be printing out the existing configuration, or previously read data

esphome/components/ade7953_base/ade7953_base.h
```diff
@@ -0,0 +1,129 @@
...
+      // this->ade_write_8(0x0010, 0x04);
+      this->ade_write_8(0x00FE, 0xAD);
+      this->ade_write_16(0x0120, 0x0030);
+      // Set gains
+      this->ade_write_8(PGA_IA_8, pga_ia_);
+      this->ade_write_8(PGA_IB_8, pga_ib_);
+      this->ade_write_32(AIGAIN_32, aigan_);
+      this->ade_write_32(BIGAIN_32, bigan_);
+      this->is_setup_ = true;
+    });
```
jesserockz: What is the point of the timeout here? 
Nothing was done beforehand that would need to be waited on...

esphome/components/ade7953_i2c/ade7953_i2c.h
```diff
@@ -0,0 +1,91 @@
...
+
+namespace esphome {
+namespace ade7953_i2c {
+
+class AdE7953I2c : public ade7953_base::ADE7953, public i2c::I2CDevice {
+ public:
+  void dump_config() override;
+
+ protected:
+  bool ade_write_8(uint16_t reg, uint8_t value) override {
```
jesserockz: Please move all of these functions into the cpp file.

esphome/components/ade7953_i2c/sensor.py
```diff
@@ -0,0 +1,28 @@
...
+
+CONFIG_SCHEMA = (
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(ADE7953),
+        }
+    )
+    .extend(ade7953_base.ADE7953_CONFIG_SCHEMA)
+    .extend(i2c.i2c_device_schema(0x38))
+    .extend(cv.polling_component_schema("1s"))
```
jesserockz:
```suggestion
```

esphome/components/ade7953_spi/sensor.py
```diff
@@ -0,0 +1,28 @@
...
+
+CONFIG_SCHEMA = (
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(ADE7953),
+        }
+    )
+    .extend(ade7953_base.ADE7953_CONFIG_SCHEMA)
+    .extend(spi.spi_device_schema())
+    .extend(cv.polling_component_schema("1s"))
```
jesserockz:
```suggestion
```

esphome/components/ade7953_base/ade7953_base.h
```diff
@@ -0,0 +1,129 @@
...
+    active_power_a_sensor_ = active_power_a_sensor;
+  }
+  void set_active_power_b_sensor(sensor::Sensor *active_power_b_sensor) {
+    active_power_b_sensor_ = active_power_b_sensor;
+  }
+
+  void set_reactive_power_a_sensor(sensor::Sensor *reactive_power_a) { reactive_power_a_sensor_ = reactive_power_a; }
+  void set_reactive_power_b_sensor(sensor::Sensor *reactive_power_b) { reactive_power_b_sensor_ = reactive_power_b; }
+
+  void setup() override {
```
jesserockz: Content of this should be in cpp file please.

jesserockz (changes requested):
Partial review below

esphome/components/ade7953_base/ade7953_base.h
angelnu: The chip needs up to 100ms to be ready after a reset. Since we do not always have access to the ready interrupt this is what the documentation recommends.
It was like this in the previous implementation - I can add a comment.

esphome/components/ade7953_base/ade7953_base.cpp
angelnu: My intention is to provide feedback on the actual values from the registers without querying them in each sensor refresh.
I went with the dump_config as this is not called to often but I can see why it does not fit the esphome architecture.

Do you have a recommendation or should I come up with a proposal? For example, we could threat them as (optional) sensors even if they do not change often. Another option would be to read the value on setup only but then if the chip gets reset they actual values might have changed.

My feeling is that just covering them as sensors is good enough. They are only required for initial debugging really so most configs do not need to include them as sensors.

WDYT?

esphome/components/ade7953_base/ade7953_base.h
angelnu: Done

esphome/components/ade7953_i2c/ade7953_i2c.h
angelnu: Done

esphome/components/ade7953_base/ade7953_base.h
angelnu: I added a comment to explain the 100ms delay.

esphome/components/ade7953_base/ade7953_base.cpp
jesserockz: Either as sensors, or gather the data once in `setup` and then print in `dump_config`

angelnu: Moved @jesserockz  - resolving.

angelnu:
@notsonominal you also mentioned in discord that this worked for you, right?

esphome/components/ade7953_base/__init__.py
```diff
@@ -0,0 +1,194 @@
...
+    "2x": 0b001,
+    "4x": 0b010,
+    "8x": 0b011,
+    "16x": 0b100,
+    "22x": 0b101,
+}
+
+ADE7953_CONFIG_SCHEMA = cv.Schema(
+    {
+        # cv.GenerateID(): cv.declare_id(ADE7953),
```
jesserockz: No commented code please
```suggestion
```

esphome/components/ade7953_i2c/sensor.py
```diff
@@ -0,0 +1,27 @@
...
+import esphome.config_validation as cv
+from esphome.components import i2c, ade7953_base
+from esphome.const import CONF_ID
+
+
+DEPENDENCIES = ["i2c"]
+AUTO_LOAD = ["ade7953_base"]
+
+ade7953_ns = cg.esphome_ns.namespace("ade7953_i2c")
+ADE7953 = ade7953_ns.class_("AdE7953I2c", cg.PollingComponent, i2c.I2CDevice)
```
jesserockz: This should be showing that is extends `ade7953_base::ADE7953`

The class stub should be created in the _base python file and imported here.

```suggestion
ADE7953 = ade7953_ns.class_("AdE7953I2c", ade7953_base.ADE7953, i2c.I2CDevice)
```

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,85 @@
...
+  i2c::ErrorCode err = write(reg_data, 2);
+  if (err != i2c::ERROR_OK)
+    return true;
+  uint8_t recv[2];
+  err = read(recv, 2);
+  if (err != i2c::ERROR_OK)
+    return true;
+  *value = 0;
+  *value |= ((uint32_t) recv[0]) << 8;
+  *value |= ((uint32_t) recv[1]);
```
jesserockz: Much cleaner =)

Need to add `#include "esphome/core/helpers.h"` to the top of the file

```suggestion
  *value = encode_uint16(recv[0], recv[1]);
```

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return true;
+  uint8_t recv[4];
+  err = read(recv, 4);
+  if (err != i2c::ERROR_OK)
+    return true;
+  *value = 0;
+  *value |= ((uint32_t) recv[0]) << 24;
+  *value |= ((uint32_t) recv[1]) << 16;
+  *value |= ((uint32_t) recv[2]) << 8;
+  *value |= ((uint32_t) recv[3]);
```
jesserockz: Same as above
```suggestion
  *value = encode_uint32(recv[0], recv[1], recv[2], recv[3]);
```

esphome/components/ade7953_spi/ade7953_spi.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+bool AdE7953Spi::ade_read_32(uint16_t reg, uint32_t *value) {
+  this->enable();
+  this->write_byte16(reg);
+  this->transfer_byte(0x80);
+  *value = 0;
+  *value |= this->read_byte() << 24;
+  *value |= this->read_byte() << 16;
+  *value |= this->read_byte() << 8;
+  *value |= this->read_byte();
```
jesserockz: Same here as i2c

jesserockz (changes requested):

angelnu:
Thanks for the review @jesserockz - changes made and testcases passing. Ready to merge from my side.

esphome/components/ade7953_spi/ade7953_spi.cpp
```diff
@@ -64,11 +65,9 @@ bool AdE7953Spi::ade_read_32(uint16_t reg, uint32_t *value) {
   this->enable();
   this->write_byte16(reg);
   this->transfer_byte(0x80);
-  *value = 0;
-  *value |= this->read_byte() << 24;
-  *value |= this->read_byte() << 16;
-  *value |= this->read_byte() << 8;
-  *value |= this->read_byte();
+  uint8_t recv[4];
+  read_array(recv, 4);
```
jesserockz:
```suggestion
  this->read_array(recv, 4);
```

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,81 @@
...
+  reg_data[0] = reg >> 8;
+  reg_data[1] = reg >> 0;
+  i2c::ErrorCode err = write(reg_data, 2);
+  if (err != i2c::ERROR_OK)
+    return true;
+  uint8_t recv[2];
+  err = read(recv, 2);
+  if (err != i2c::ERROR_OK)
+    return true;
+  *value = 0;
```
jesserockz: Probably don't need this line
```suggestion
```

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,81 @@
...
+bool AdE7953I2c::ade_write_16(uint16_t reg, uint16_t value) {
+  std::vector<uint8_t> data;
+  data.push_back(reg >> 8);
+  data.push_back(reg >> 0);
+  data.push_back(value >> 8);
+  data.push_back(value >> 0);
+  return write(data.data(), data.size()) != i2c::ERROR_OK;
+}
+bool AdE7953I2c::ade_write_32(uint16_t reg, uint32_t value) {
+  std::vector<uint8_t> data;
```
jesserockz: Since we know the future size we can allocate it once at the start
```suggestion
  std::vector<uint8_t> data(6);
```

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+bool AdE7953I2c::ade_write_8(uint16_t reg, uint8_t value) {
+  std::vector<uint8_t> data;
+  data.push_back(reg >> 8);
+  data.push_back(reg >> 0);
+  data.push_back(value);
+  return write(data.data(), data.size()) != i2c::ERROR_OK;
+}
+bool AdE7953I2c::ade_write_16(uint16_t reg, uint16_t value) {
+  std::vector<uint8_t> data;
```
jesserockz:
```suggestion
  std::vector<uint8_t> data(4);
```

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+static const char *const TAG = "ade7953";
+
+void AdE7953I2c::dump_config() {
+  ESP_LOGCONFIG(TAG, "ADE7953_i2c:");
+  LOG_I2C_DEVICE(this);
+  ade7953_base::ADE7953::dump_config();
+}
+bool AdE7953I2c::ade_write_8(uint16_t reg, uint8_t value) {
+  std::vector<uint8_t> data;
```
jesserockz:
```suggestion
  std::vector<uint8_t> data(3);
```

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,81 @@
...
+  data.push_back(reg >> 8);
+  data.push_back(reg >> 0);
+  data.push_back(value);
+  return write(data.data(), data.size()) != i2c::ERROR_OK;
```
jesserockz: Please make sure you prefix member functions with `this->`
```suggestion
  return this->write(data.data(), data.size()) != i2c::ERROR_OK;
```

jesserockz (changes requested):

angelnu:
Thanks for the new additions @jesserockz - all merged.

esphome/components/ade7953_i2c/ade7953_i2c.cpp
```diff
@@ -0,0 +1,80 @@
...
+  data.push_back(value);
+  return this->write(data.data(), data.size()) != i2c::ERROR_OK;
+}
+bool AdE7953I2c::ade_write_16(uint16_t reg, uint16_t value) {
+  std::vector<uint8_t> data(4);
+  data.push_back(reg >> 8);
+  data.push_back(reg >> 0);
+  data.push_back(value >> 8);
+  data.push_back(value >> 0);
+  return write(data.data(), data.size()) != i2c::ERROR_OK;
```
jesserockz: There are some more places to add `this->` prefix. I just dont add all of them in a review as I might as well just update the branch at that point and then it doesnt help you for future PRs =)

angelnu: @jesserockz - ok, I think I got them all now. Sorry that I missed your comment about other occurrences.

esphome/components/ade7953_spi/ade7953_spi.cpp
angelnu: added in a later commit

esphome/components/ade7953_base/ade7953_base.h
jesserockz: This is not done yet?

esphome/components/ade7953/sensor.py
jesserockz, line 1: I have been on the fence about this since aliasing is nice for people that currently use the platform, but we tend to just remove them with a validation error informing users they need to make a change or see the documentation etc.

Please search the codebase for `CONFIG_SCHEMA = cv.invalid("...")` to see other examples.

jesserockz (changes requested):

esphome/components/ade7953/sensor.py
angelnu: ok - removed alias @jesserockz

jesserockz (approved):

PR #5445, opened 2023-09-28
Do not enable SHT3x heater by default. Fixes #4886.

### LINKED ISSUE #4886: SHT3x now enables heater by default
### The problem

PR #5161 seems to enable the sensor heater by default despite its title: `cv.Optional(CONF_HEATER_ENABLED, default=True): cv.boolean`, and later `if (!this->write_command(heater_enabled_ ? SHT3XD_COMMAND_HEATER_ENABLE : SHT3XD_COMMAND_HEATER_DISABLE)) {`

2023.10.0-dev

Home Assistant Add-on

_No response_

ESP8266

### Board

_No response_

sht3xd

### Example YAML snippet

_No response_

_No response_

### Additional information

This can be worked around by setting `heater_enabled: false`, but I don't see the need for a breaking change.

However there in general there is no way to make a piece of code that changes a setting where previously no forced setting code existed non-breaking to someone. However I think the heater off is the reasonable setting for the majority of users.

jkl1337:

The SHT sensor heater is for special use and defaulting them to on breaks the vast majority of prior configs and seems quite surprising. This appears to have been an inadvertent side effect of PR #5161.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4886

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

## Checklist:

### DIFF
--- a/esphome/components/sht3xd/sensor.py
+++ b/esphome/components/sht3xd/sensor.py
@@ -38,7 +38,7 @@ CONFIG_SCHEMA = (
                 device_class=DEVICE_CLASS_HUMIDITY,
                 state_class=STATE_CLASS_MEASUREMENT,
             ),
-            cv.Optional(CONF_HEATER_ENABLED, default=True): cv.boolean,
+            cv.Optional(CONF_HEATER_ENABLED, default=False): cv.boolean,
         },
     )
     .extend(cv.polling_component_schema("60s"))


jesserockz:
From my understanding of the previous PR, the heater was always turned on anyway before the PR. 
That PR simply allowed it to be disabled at startup instead. 

Changing the default now would be a breaking change to everyone using it now.

jkl1337:
No, the original PR broke everything. Please see the issue [#4886](https://github.com/esphome/issues/issues/4886) filed and now the people piling up complaining that the latest release broke things.

The heater is only for special use cases. It should not be on for normal use of the sensor, please see data sheet for how it is used.

Apparently the original breaking PR author had a device that turned it on by default which is likely undesirable (also, note, not the norm the PR states “on my custom built PCB using the sensirion chips, the heater was enabled by default”) and they needed code to turn it off. Note their comment: "This is a good question. From the different discussions, I understood it should have been turned off by default." The original PR even says to allow *disabling* the heater. It's very confusing that the default was made to enable it. I would imagine the original PR author was testing with an explicit `heater_enabled: False` and missed that the default was incorrect.

jesserockz:
Ok understood. Happy to merge this change in but as its breaking and its easy enough for people to change yaml to false, it wont be released until 2023.10.0

jesserockz (approved):

jesserockz:
Could you also please make a docs PR targeting `next` to flip the default value there too?

kbx81 (approved):

PR #5447, opened 2023-09-28
Fix an Issue with IR Remote Climate and Whirlpool protocol toggle

the-mentor:
# What does this fix?
In short because the whirlpool protocol is based on toggle we need to block the receiver while transmitting to prevent the system from getting confused. 

## Types of changes

Related issues:
https://github.com/esphome/issues/issues/4650
https://github.com/esphome/issues/issues/4468

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```
 remote_transmitter:
  id: remote_transmitter_id
  pin: GPIO14
  carrier_duty_percent: 50%

remote_receiver:
  id: "remote_receiver_id"
  pin:
    number: GPIO5
    inverted: True
  tolerance: 55%
  dump: all

climate:
  - platform: whirlpool
    name: "Study AC"
    model: "DG11J1-91"
    sensor: room_temperature
    transmitter_id: remote_transmitter_id
    receiver_id: remote_receiver_id
```

## Checklist:

### DIFF
--- a/esphome/components/whirlpool/whirlpool.h
+++ b/esphome/components/whirlpool/whirlpool.h
@@ -48,7 +48,7 @@ class WhirlpoolClimate : public climate_ir::ClimateIR {
   /// Handle received IR Buffer
   bool on_receive(remote_base::RemoteReceiveData data) override;
   /// Set the time of the last transmission.
-  int32_t last_transmit_time_;
+  int32_t last_transmit_time_{};
 
   bool send_swing_cmd_{false};
   Model model_;
--- a/esphome/components/whirlpool/whirlpool.h
+++ b/esphome/components/whirlpool/whirlpool.h
@@ -48,7 +48,7 @@ class WhirlpoolClimate : public climate_ir::ClimateIR {
   /// Handle received IR Buffer
   bool on_receive(remote_base::RemoteReceiveData data) override;
   /// Set the time of the last transmission.
-  int32_t last_transmit_time_ = millis() + 500;
+  int32_t last_transmit_time_;
 
   bool send_swing_cmd_{false};
   Model model_;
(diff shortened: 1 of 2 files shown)


the-mentor:
Ping @kpfleming following our discord chat

clydebarrow (changes requested):
You should make the variable an instance member - declare it in the header file under `protected:`. Call it `last_transmit_time_` (note the trailing underscore) and refer to it in your code as `this->last_transmit_time_`. "t1" is not descriptive and a global is not appropriate here - it should not be visible outside the component.

esphome/components/whirlpool/whirlpool.h
```diff
@@ -47,6 +47,8 @@ class WhirlpoolClimate : public climate_ir::ClimateIR {
   void transmit_state() override;
   /// Handle received IR Buffer
   bool on_receive(remote_base::RemoteReceiveData data) override;
+  /// Set the time of the last transmission.
+  int32_t last_transmit_time_ = millis() + 500;
```
clydebarrow:
```suggestion
  int32_t last_transmit_time_;
```

clydebarrow (changes requested):

clydebarrow (approved):
Looks good to me.

esphome/components/whirlpool/whirlpool.h
```diff
@@ -47,6 +47,8 @@ class WhirlpoolClimate : public climate_ir::ClimateIR {
   void transmit_state() override;
   /// Handle received IR Buffer
   bool on_receive(remote_base::RemoteReceiveData data) override;
+  /// Set the time of the last transmission.
+  int32_t last_transmit_time_;
```
clydebarrow: Sorry, I led you astray there, this should have an initialiser, just not `millis()`. Initiialising to `{}` will make it zero, so guaranteed to be in the past.

```suggestion
  int32_t last_transmit_time_{};
```

jesserockz (approved):

PR #5449, opened 2023-09-28
Improve dualstack and IPv6 support

HeMan:

Makes ESPHome work better with IPv6 by adding a datatype for multiple IP addresses. This adds support for reporting all addresses it got, via `wifi_info`/`ethernet_info` text-sensors, via MQTT and as part of debug message.
It makes MQTT resolve both IPv6 and IPv4 addresses, and connects over IPv6 (as well as over IPv4 as before).
Breaks external components that assumes ESPHome only has one IP address.

## Types of changes

- [x] Other

https://github.com/esphome/issues/issues/5126

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3234

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP32 Ethernet
- [x] ESP8266
- [x] RP2040

```yaml
esphome:
  name: ipv6mqtt

network:
  enable_ipv6: true

esp32:
  board: esp32dev
  framework:
    type: esp-idf

wifi:
  ssid: !secret wifissd
  password: !secret wifipwd

sensor:
  - platform: wifi_signal
    name: "WiFi Signal Sensor"
    update_interval: 60s

text_sensor:
  - platform: wifi_info
    ip_address:
      name: ESP IP Address
      address_0:
        name: ESP IP Address 0
      address_1:
        name: ESP IP Address 1
      address_2:
        name: ESP IP Address 2
      address_3:
        name: ESP IP Adress 3

debug:

logger:
  level: VERY_VERBOSE

web_server:
  port: 80

api:

ota:

mqtt:
  broker: mqtt6.example.com
  discovery: true


```

## Checklist:


esphome/components/ethernet/ethernet_component.h
```diff
@@ -58,7 +58,7 @@ class EthernetComponent : public Component {
   void set_clk_mode(emac_rmii_clock_mode_t clk_mode, emac_rmii_clock_gpio_t clk_gpio);
   void set_manual_ip(const ManualIP &manual_ip);
 
-  network::IPAddress get_ip_address();
+  network::IPAddresses get_ip_address();
```
jesserockz: I feel like if you are changing the functionality to return a list, the name of the function should also change...
```suggestion
  network::IPAddresses get_ip_addresses();
```

jesserockz (changes requested):

esphome/components/ethernet/ethernet_component.h
HeMan: Good suggestion! Should I also update `WiFiComponent::wifi_sta_ip` to `WiFiComponent::wifi_sta_ips`?

esphome/components/ethernet_info/ethernet_info_text_sensor.h
```diff
@@ -12,10 +12,18 @@ namespace ethernet_info {
...
+    if (ips != this->last_ips_) {
+      this->last_ips_ = ips;
+      for (auto &ip : ips) {
+        if (ip.is_set()) {
+          address_results += ip.is_ip4() ? "IPv4: " : "IPv6: ";
+          address_results += ip.str();
+          address_results += "\n";
+        }
+      }
+      this->publish_state(address_results);
```
jesserockz: Not sure I like this sensor changing to just list all addresses. This could be a breaking change for people that use it currently.

jesserockz: Maybe have a second sensor for ipv6_address?

esphome/components/network/__init__.py
```diff
@@ -16,12 +17,14 @@
...
     {
         cv.Optional(CONF_ENABLE_IPV6, default=False): cv.boolean,
+        cv.Optional(CONF_MIN_IPV6_ADDR_COUNT, default=0): cv.positive_int,
     }
 )
 
 
 async def to_code(config):
     cg.add_define("ENABLE_IPV6", config[CONF_ENABLE_IPV6])
+    cg.add_define("MIN_IPV6_ADDR_COUNT", config[CONF_MIN_IPV6_ADDR_COUNT])
```
jesserockz: This should follow the format we use through ESPHome. `USE_NETWORK_MIN_IPV6_ADDRESS_COUNT`

(Really, `ENABLE_IPV6` should have followed this format too)

esphome/components/wifi/wifi_component.cpp
```diff
@@ -190,11 +190,12 @@ void WiFiComponent::set_fast_connect(bool fast_connect) { this->fast_connect_ =
 void WiFiComponent::set_btm(bool btm) { this->btm_ = btm; }
 void WiFiComponent::set_rrm(bool rrm) { this->rrm_ = rrm; }
 #endif
-network::IPAddress WiFiComponent::get_ip_address() {
+network::IPAddresses WiFiComponent::get_ip_addresses() {
   if (this->has_sta())
-    return this->wifi_sta_ip();
+    return this->wifi_sta_ip_addresses();
+  // TODO: fix soft ap
```
jesserockz: Is this still a TODO?

jesserockz (changes requested):

esphome/components/network/__init__.py
HeMan: Should I change `ENABLE_IPV6` while I'm at it? Is `USE_NETWORK_IPV6` a good name or do you have a better suggestion?

esphome/components/wifi/wifi_component.cpp
HeMan: Nope, removing the comment.

esphome/components/ethernet_info/ethernet_info_text_sensor.h
HeMan: This is a tricky one. Going dual stack will change the behavior for the entire system with that it usually has more than one address. Having different API endpoints for different addresses is also something that I like to avoid.
On the other hand, I'm not really pleased with adding prefixes to the address and treat is a "textblob" with newlines since it makes it harder to machine parse.
Is there a way to dynamically add and remove sensors? Have an array of sensors that could contain empty/unset addresses? Would that break something else?

esphome/components/ethernet_info/ethernet_info_text_sensor.h
```diff
@@ -12,19 +12,31 @@ namespace ethernet_info {
...
-      this->publish_state(network::IPAddress(ip).str());
+    std::string address_results;
+    auto ips = ethernet::global_eth_component->get_ip_addresses();
+    if (ips != this->last_ips_) {
+      this->last_ips_ = ips;
+      this->publish_state(ips[0].str());
+      uint8_t sensor = 0;
+      for (auto &ip : ips) {
+        if (ip.is_set()) {
+          if (ip_sensors_[sensor]) {
```
jesserockz: Please alter to explicitly check for `nullptr`
```suggestion
          if (ip_sensors_[sensor] != nullptr) {
```

esphome/components/ethernet_info/text_sensor.py
```diff
@@ -18,17 +18,25 @@
     {
         cv.Optional(CONF_IP_ADDRESS): text_sensor.text_sensor_schema(
             IPAddressEsthernetInfo, entity_category=ENTITY_CATEGORY_DIAGNOSTIC
-        ).extend(cv.polling_component_schema("1s"))
+        )
+        .extend(cv.polling_component_schema("1s"))
+        .extend(
+            {
+                cv.Optional(f"a{x}"): text_sensor.text_sensor_schema(
```
jesserockz: I feel like the config key should be a bit more descriptive than `aX`

MatthewCroughan:
Just commenting to let you know that this fixed all of my IPv6 woes. AAAA records now resolve, MQTT broker can be set to an IPv6 address. In addition, IPv6 addresses will not be accidentally parsed as if they were domain names. I was able to test your changes simply on an esp32 with `nix`.

###### Getting your fork of ESPHome and putting it on my `$PATH` by overriding the existing [Nix derivation from Nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/tools/misc/esphome/default.nix#L103)
1. `nix shell --impure --expr '(import (builtins.getFlake "github:nixos/nixpkgs/nixos-23.11") {}).esphome.overrideAttrs { src = builtins.fetchGit { url = "https://github.com/HeMan/esphome.git"; rev = "8b4d425516f4e9eba48b18f1338aff2698765dff"; allRefs = true; }; }'`

2. `esphome run esphome.yaml`


###### esphome.yaml

```yml
esphome:
  name: esp32home1

esp32:
  board: esp32dev
  framework:
    type: arduino

network:
    enable_ipv6: true

logger:
  level: VERBOSE

mqtt:
  broker: "0203:7f46:0257:a116:6b75:a134:4b73:cc36"
```

MatthewCroughan:
It seems like the web interface doesn't listen on the v6 address, meaning it is inaccessible via IPv6.

HeMan:
> It seems like the web interface doesn't listen on the v6 address, meaning it is inaccessible via IPv6.

Web interface for Arduino on ESP32 depends on https://github.com/esphome/AsyncTCP/pull/5

esphome/components/wifi_info/text_sensor.py
```diff
@@ -37,7 +37,16 @@
     {
         cv.Optional(CONF_IP_ADDRESS): text_sensor.text_sensor_schema(
             IPAddressWiFiInfo, entity_category=ENTITY_CATEGORY_DIAGNOSTIC
-        ).extend(cv.polling_component_schema("1s")),
+        )
+        .extend(cv.polling_component_schema("1s"))
+        .extend(
+            {
+                cv.Optional(f"a{x}"): text_sensor.text_sensor_schema(
```
jesserockz: Same here

esphome/components/wifi_info/wifi_info_text_sensor.h
```diff
@@ -3,25 +3,37 @@
...
-      this->last_ip_ = ip;
-      this->publish_state(ip.str());
+    auto ips = wifi::global_wifi_component->wifi_sta_ip_addresses();
+    if (ips != this->last_ips_) {
+      this->last_ips_ = ips;
+      this->publish_state(ips[0].str());
+      uint8_t sensor = 0;
+      for (auto &ip : ips) {
+        if (ip.is_set()) {
+          if (ip_sensors_[sensor]) {
```
jesserockz: Same here

jesserockz (changes requested):

HeMan:
Libretiny depends on https://github.com/esphome/AsyncTCP/pull/9

esphome/components/ethernet_info/ethernet_info_text_sensor.h
```diff
@@ -12,19 +12,31 @@ namespace ethernet_info {
...
+          }
+        }
+      }
     }
   }
 
   float get_setup_priority() const override { return setup_priority::ETHERNET; }
   std::string unique_id() override { return get_mac_address() + "-ethernetinfo"; }
   void dump_config() override;
+  void add_ip_sensors(uint8_t index, text_sensor::TextSensor *s) { ip_sensors_[index] = s; }
```
jesserockz:
```suggestion
  void add_ip_sensors(uint8_t index, text_sensor::TextSensor *s) { this->ip_sensors_[index] = s; }
```

esphome/components/ethernet_info/ethernet_info_text_sensor.h
```diff
@@ -12,19 +12,31 @@ namespace ethernet_info {
...
-      this->publish_state(network::IPAddress(ip).str());
+    std::string address_results;
+    auto ips = ethernet::global_eth_component->get_ip_addresses();
+    if (ips != this->last_ips_) {
+      this->last_ips_ = ips;
+      this->publish_state(ips[0].str());
+      uint8_t sensor = 0;
+      for (auto &ip : ips) {
+        if (ip.is_set()) {
+          if (ip_sensors_[sensor] != nullptr) {
```
jesserockz:
```suggestion
          if (this->ip_sensors_[sensor] != nullptr) {
```

esphome/components/wifi_info/wifi_info_text_sensor.h
```diff
@@ -3,25 +3,37 @@
...
-      this->last_ip_ = ip;
-      this->publish_state(ip.str());
+    auto ips = wifi::global_wifi_component->wifi_sta_ip_addresses();
+    if (ips != this->last_ips_) {
+      this->last_ips_ = ips;
+      this->publish_state(ips[0].str());
+      uint8_t sensor = 0;
+      for (auto &ip : ips) {
+        if (ip.is_set()) {
+          if (ip_sensors_[sensor] != nullptr) {
```
jesserockz:
```suggestion
          if (this->ip_sensors_[sensor] != nullptr) {
```

esphome/components/wifi_info/wifi_info_text_sensor.h
```diff
@@ -3,25 +3,37 @@
...
+            sensor++;
+          }
+        }
+      }
     }
   }
   float get_setup_priority() const override { return setup_priority::AFTER_WIFI; }
   std::string unique_id() override { return get_mac_address() + "-wifiinfo-ip"; }
   void dump_config() override;
+  void add_ip_sensors(uint8_t index, text_sensor::TextSensor *s) { ip_sensors_[index] = s; }
```
jesserockz:
```suggestion
  void add_ip_sensors(uint8_t index, text_sensor::TextSensor *s) { this->ip_sensors_[index] = s; }
```

jesserockz (changes requested):
**This** almost looks good, just a couple of missing things :wink:

HeMan:
> **This** almost looks good, just a couple of missing things 😉

Fixed! Found an unused variable as well.

jesserockz (approved):

PR #5451, opened 2023-09-29
IDF 5 fixes for various components from test1.yaml

kbx81:

IDF 5 fixes for various components compile-tested in our existing `test1.yaml` file.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/ota/automation.h
+++ b/esphome/components/ota/automation.h
@@ -54,11 +54,7 @@ class OTAEndTrigger : public Trigger<> {
   }
 };
 
-#if ESP_IDF_VERSION_MAJOR >= 5
-class OTAErrorTrigger : public Trigger<long int> {
-#else
-class OTAErrorTrigger : public Trigger<int> {
-#endif
+class OTAErrorTrigger : public Trigger<uint8_t> {
  public:
   explicit OTAErrorTrigger(OTAComponent *parent) {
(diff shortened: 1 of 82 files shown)


HeMan (approved):

esphome/components/ota/automation.h
```diff
@@ -54,7 +54,11 @@ class OTAEndTrigger : public Trigger<> {
   }
 };
 
+#if ESP_IDF_VERSION_MAJOR >= 5
+class OTAErrorTrigger : public Trigger<long int> {
+#else
 class OTAErrorTrigger : public Trigger<int> {
+#endif
```
jesserockz: Why does this need to change?

Also the `error` variable is always a `uint8_t` anyway, so the trigger can probably be changed to `uint8_t` as well...

kbx81: Oh sweet! I didn't dig in quite that deep. The compiler complained about the type and it led me here. Let me take another look...

kbx81: Pushed a change, let me know if this looks right...

jesserockz (approved):

PR #5458, opened 2023-10-01
Tuya Number: split "multiply" to a separate option

### LINKED ISSUE #4928: Tuya Number "step" attribute serves double duty
### The problem

I am configuring a dehumidifier with the Tuya MCU component. Unfortunately there is no dehumidifier HA component support, nor support for Climate devices with no target temperature, so I am configuring these as individual sensors etc.

One of the data points is, well, the "target humidity" :smiley: The MCU accepts values from 30 to 80, in increments of 5 - so 30, 35, 40, 45, etc. Values in between (e.g. 32) are ignored, and nothing is configured on the MCU.

So I naturally tried configuring this with:
```yaml
  - platform: tuya
    name: "Target humidity"
    number_datapoint: 4
    min_value: 30`
    max_value: 80
    step: 5
```

Unfortunately this resulted into nonsensical values, like "325", because -as a careful look at the documentation reveals- the `step` attribute is _"used as the scaling factor for the Number"_. This MCU reports and accepts just "65" for "65%".

The Modbus Controller Number documentation (which I have not tested) seems to suggest that it has a separate `step` attribute, and a `multiply` attribute. The Template Number has a `step`, and no multiply. The Tuya Climate component has a `temperature_multiplier` attribute. The Sensor components have filters, of course.

Having a way to set the `step` separately from the multiplier feels useful. It's impossible to create a valid configuration for my MCU as things are. Perhaps that would be a feature request, but I'm choosing to file this as an issue because calling it "step" is a) misleading b) inconsistent with other integrations.

2023.9.1

Home Assistant Add-on

_No response_

ESP8266

### Board

_No response_

_No response_

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

paravoid:

Commit 9d4f471 (PR #5108), part of ESPHome 2023.9.1, introduced a multiplication factor for Tuya Number, to be used when converting from a float (say, pH 5.2) to an integer-only data-point (say, 5200).

However, that commit overloaded the "step" attribute, incorrectly assuming that all stepped values are supposed to be multiplied. There are devices in the wild that expect integers to be moved in steps, but do not require any multiplication; for example, a dehumidifier that accepts an integer for target humidity, but only in steps of 5%.

With the aforementioned commit, it's impossible to correctly configure such a device. Moreover, despite documented, it is counter-intuitive, as "step" (CONF_STEP) now does something different compared to other ESPHome components.

To resolve this, revert "step" to its previous function, and add a separate "multiply" attribute, using the standard CONF_MULTIPLY key. This follows what other components do, such as for example the Modbus Controller Number component, thereby providing an additional knob in a standard way.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) : _sort of: this breaks 2023.9.1 configs, but restores <= 2023.9.0 configs_
- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/issues#4928

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3224

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040

```yaml
tuya:

number:
  - platform: tuya
    name: "Target humidity"
    number_datapoint: 4
    min_value: 30`
    max_value: 80
    step: 5
```

```yaml
tuya:

number:
  - platform: tuya
    name: "pH Low Limit"
    number_datapoint: 108
    min_value: 0
    max_value: 15
    multiply: 100
```

## Checklist:

### DIFF
--- a/esphome/components/tuya/number/__init__.py
+++ b/esphome/components/tuya/number/__init__.py
@@ -6,6 +6,7 @@ from esphome.const import (
     CONF_NUMBER_DATAPOINT,
     CONF_MAX_VALUE,
     CONF_MIN_VALUE,
+    CONF_MULTIPLY,
     CONF_STEP,
 )
 from .. import tuya_ns, CONF_TUYA_ID, Tuya
@@ -31,6 +32,7 @@ CONFIG_SCHEMA = cv.All(
             cv.Required(CONF_MAX_VALUE): cv.float_,
             cv.Required(CONF_MIN_VALUE): cv.float_,
             cv.Required(CONF_STEP): cv.positive_float,
+            cv.Optional(CONF_MULTIPLY, default=1.0): cv.float_,
         }
     )
     .extend(cv.COMPONENT_SCHEMA),
@@ -49,7 +51,8 @@ async def to_code(config):
         step=config[CONF_STEP],
     )
 
-    paren = await cg.get_variable(config[CONF_TUYA_ID])
-    cg.add(var.set_tuya_parent(paren))
+    cg.add(var.set_write_multiply(config[CONF_MULTIPLY]))
+    parent = await cg.get_variable(config[CONF_TUYA_ID])
+    cg.add(var.set_tuya_parent(parent))
 
     cg.add(var.set_number_id(config[CONF_NUMBER_DATAPOINT]))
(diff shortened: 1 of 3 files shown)


paravoid:
@InvncibiltyCloak, since you authored this multiplication feature, I'd appreciate your review! Turns out that PR did introduce a regression I'm afraid :)

@ssieb & @jesserockz, also tagging you since you reviewed & merged the original change.

jesserockz (approved):
This makes total sense. Thanks for fixing

PR #5463, opened 2023-10-01
Displays with resolutions of 320x480 or more failing on ESP32

fsiek:
with 520kB RAM or less on allocating memory for the frame buffer.

By reducing the display size and clearing the border via SPI the setup of the display will no longer fail but only getting a small border.

Display setup will no longer fail on displays with higher resolutions. The buffered display size will be reduced to the resolution the largest free memory block will make possible. The unaccessable border will be cleared via a SPI fill function during setup.
Did this to make [ESP32-3248S035](https://de.aliexpress.com/item/1005005542773413.html) board including a 3.5" ST7796 display work. 

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
esphome:
  name: Test1
  friendly_name: Test1

esp32:
  board: az-delivery-devkit-v4
  framework: 

    type: esp-idf
    version: recommended

# Enable logging
logger:
  level: VERY_VERBOSE

# Enable Home Assistant API
api:
  encryption:
    key: "..."

ota:
  password: "..."

wifi:
  ssid: !secret wifi_ssid

  ap:
    ssid: "Test1 Fallback Hotspot"
    password: "..."

captive_portal:

color:

  - id: my_blue
    red: 0%
    green: 0%
    blue: 100%

spi:
  - id: lcd
    clk_pin: GPIO14
    mosi_pin: GPIO13
    miso_pin: GPIO12

display:
  - platform: ili9xxx
    model: ST7796
    spi_id: lcd
    cs_pin: GPIO15
    dc_pin: GPIO2
    reset_pin: GPIO23
    dimensions: 320x480
    rotation: 270
    update_interval: 1s
    lambda: |-
      it.filled_rectangle(0, 0, it.get_width(), it.get_height(), id(my_blue));

```

## Checklist:

### DIFF
--- a/esphome/components/ili9xxx/ili9xxx_display.cpp
+++ b/esphome/components/ili9xxx/ili9xxx_display.cpp
@@ -42,7 +42,7 @@ void ILI9XXXDisplay::setup() {
   }
 }
 
-void ILI9XXXDisplay::downsize_resolution(uint16_t bytes_per_pixel) {
+void ILI9XXXDisplay::downsize_resolution_(uint16_t bytes_per_pixel) {
   uint32_t max_mem = heap_caps_get_largest_free_block(MALLOC_CAP_8BIT);
   while (this->get_buffer_length_() * bytes_per_pixel > max_mem) {
     this->width_ = this->width_ - 2;
--- a/esphome/components/ili9xxx/ili9xxx_display.cpp
+++ b/esphome/components/ili9xxx/ili9xxx_display.cpp
@@ -24,7 +24,7 @@ void ILI9XXXDisplay::setup() {
   if (this->buffer_color_mode_ == BITS_16) {
     if (this->get_buffer_length_() * 2 > max_mem) {
       downsize_resolution(2);  // Reduce size.
-      this->fill_spi(this->x_high_, this->y_high_, this->x_low_, this->y_low_, 0x0000);  // fill display without using the buffer to clear border around buffered screen
+      this->fill_spi(this->x_high_, this->y_high_, this->x_low_, this->y_low_, 0x0000);
     }
     this->init_internal_(this->get_buffer_length_() * 2);
     if (this->buffer_ != nullptr) {
@@ -34,8 +34,8 @@ void ILI9XXXDisplay::setup() {
   }
   if (this->get_buffer_length_() > max_mem) {
     downsize_resolution(1);  // Reduce size
-    this->fill_spi(this->x_high_, this->y_high_, this->x_low_, this->y_low_, 0x0000);  // fill display without using the buffer to clear border around buffered screen
-  }
+    this->fill_spi(this->x_high_, this->y_high_, this->x_low_, this->y_low_, 0x0000);
+    }
   this->init_internal_(this->get_buffer_length_());
   if (this->buffer_ == nullptr) {
     this->mark_failed();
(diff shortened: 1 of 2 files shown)


clydebarrow:
I would suggest that rather than trying to automagically change the display size, that there should be a config option to specify the border sizes. That will be more robust but still allow you to achieve your goal.

In any case the reduced size would need to be reflected in the getwidth/getheight calls, and be taken into account in the drawing methods. I don't see that right now.

nielsnl68:
I agree with @clydebarrow

fsiek:
Paint routines using get_width_internal() and get_height_internal() and returning the variables width_ and height_ that are adjusted to the maximum possible size.

Because the maximum available memory block depends strongly on the configuration used, it is not predictable how large the margin must be to fit the frame buffer into memory.

```
esp32:
  board: az-delivery-devkit-v4
  framework: 
    type: arduino
    version: recommended
```
gives you
```
[09:16:20][D][ili9xxx:054]: Largest free heap block: 110580 bytes
[09:16:20][D][ili9xxx:055]: Reduced display size to 262x422
```

But with this configuration
```
esp32:
  board: az-delivery-devkit-v4
  framework: 
    type: esp-idf
    version: recommended
    sdkconfig_options:
      CONFIG_FREERTOS_UNICORE: y
      CONFIG_ESP_SYSTEM_MEMPROT_FEATURE: n
      CONFIG_ESP32_IRAM_AS_8BIT_ACCESSIBLE_MEMORY: y
      CONFIG_LWIP_IPV6: n
```
you will get ~20% more available display size
```
[08:56:22][D][ili9xxx:054]: Largest free heap block: 131072 bytes
[08:56:22][D][ili9xxx:055]: Reduced display size to 290x450
```

clydebarrow:
> Because the maximum available memory block depends strongly on the configuration used, it is not predictable how large the margin must be to fit the frame buffer into memory.

Precisely. The margin needs to be chosen, if necessary by trial and error, by the user. It's not going to be acceptable to get some random size since to make any practical use of the display you have to know in advance how big the display area is so you can layout the screen.
One option, already available, is to use 8 bit colours. Another approach would be to have a buffer with half the display resolution (i.e. 1/4 the buffer size) and write 4 pixels for each one in the buffer.

fsiek:
It's just a fallback in case the memory is not sufficient. Even with 8 bit color you need a memory block of 153kB for a 320x480 display. Most ESP32 boards have just 320kB but even with a 520kB board you need to be lucky to get 130kB for the frame buffer.

It's very annoying to work with such displays as you would not await memory problems with a 520kB board. It took me some days to figure out, how to circumvent the memory problems and make it run with ESPHome.

Display_Home_Assistant2

ssieb:
All ESP32 chips have the same amount of RAM.  The difference is just in how it's described.
I also don't agree with this type of random resizing of the display.

fsiek:
> All ESP32 chips have the same amount of RAM.

That's not true. Most boards have 320kB but AZ-Delivery ESP-32 Dev Kit C V4 and FireBeetle-ESP32 have 520kB. [https://registry.platformio.org/platforms/platformio/espressif32/boards](url)

Have here a ESP32-3248S035 that has 520kB, too.

ssieb:
From https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/memory-types.html :
`There is 520 KB of available SRAM (320 KB of DRAM and 200 KB of IRAM) on the ESP32. `
Some vendors only list the 320 and others list the full 520.  Only the 320 is available for general use.

clydebarrow:
Offset_height and width are now included in the ili9xxx driver.

PR #5472, opened 2023-10-02
St7789v and SPI data rate

### LINKED ISSUE #4914: ESPHome 2023.9.0 broke SPI (with st7789v display)
### The problem

Have been happy user of ESP32-board with st7789v display, no issues. Now did an update to 2023.9.0 (from previous, I think it was 2023.8.3) and suddenly the display stays blank. I can see a few flickers on backlight but that's it.

I would think this is an SPI issue since using "interface: software" I can get the attached display to work, albeit really slow.:
[W][component:204]: Component st7789v.display took a long time for an operation (1.85 s).

Before update it was more functional with:
[W][component:204]: Component st7789v.display took a long time for an operation (0.27 s).

2023.9.0

Home Assistant Add-on

2023.9.3

ESP32

### Board

az-delivery-devkit-v4

SPI st7789v

### Example YAML snippet

```yaml
esp32:
  board: esp32dev # Also tried az-delivery-devkit-v4 same result
  framework:
    type: arduino
--
--
# Display
spi:
  clk_pin: GPIO18
  mosi_pin: GPIO23
  interface: software # This is only option that works, hardware, spi, spi1, spi2 do not

display:
  - platform: st7789v
    model: Custom
    height: 240
    width: 240
    offset_height: 0
    offset_width: 0
    rotation: 90
    eightbitcolor : yes
    #update_interval: 20ms
    backlight_pin: GPIO16
    cs_pin: GPIO25
    dc_pin: GPIO26
    reset_pin: GPIO27
    lambda: .....
```

_No response_

### Additional information

_No response_

clydebarrow:

WIP

Issue https://github.com/esphome/issues/issues/4914 reports st7789v display not working after 2023.9.0. Possibilities are some SPI change (unlikely, it's well tested now) or changes made to the st7789v component. This PR explores possible fixes.

I have tested 2023.9.0 on an ST7789V board without problem, as has @kbx81 so the reported issue must be not common to all displays.

Possibilities:

* ~Data rate - default rate was changed from 10MHz to 20MHz. This may not be suitable for displays that are not integrated into a board.~
* SPI mode - it was changed from 3 to 0, per the ST7789V datasheet, and that worked on multiple boards at the time. It's possible there are some "st7789v" chips that require a different mode.

This PR moves the SPI data rate into the common SPI base schema thus making it configurable for all SPI devices. Will get OP for the issue report to test with lower data rates.

It now also allows setting the SPI mode so the OP can test with different modes.
 

Also fixes the data rate for the MAX6675 component - see issue #4932

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4914 and https://github.com/esphome/issues/issues/4932

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3236

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
display:
  - platform: st7789v
    power_supply: power_on
    model: LILYGO_T-EMBED_170X320
    spi_mode: mode0
    data_rate: 10MHz
```

## Checklist:


clydebarrow:
Confirmed by the OP in https://github.com/esphome/issues/issues/4914 that selecting mode 3 fixes his board. I will update the docs to note that this may be necessary and can consider this done.

kbx81 (approved):

PR #5475, opened 2023-10-03
Publish the `pulse_meter` total when setting the total

### LINKED ISSUE #4941: Pulse Meter Sensor does not set Value
### The problem

Pulse Meter Sensor does not set Value. Everything works in ESPHome (dev).

2023.9.2

Home Assistant Add-on

core-2023.9.3

ESP8266

### Board

nodemcu

_No response_

### Example YAML snippet

```yaml
- platform: template
    name: "Reset 0"
    on_press:
      - pulse_meter.set_total_pulses:
         id: pulse_meter_id
         value: "0"
```

_No response_

### Additional information

_No response_

TrentHouliston:

When the pulse total count is manually set the new state should be published.
Since this sensor should be a total increasing sensor when analysing if the total drops it can be seen as a continuation point.

If the initial value you set to isn't published it's impossible to know how big the "total increase" was for the step after that.

This PR changes the set total to also publish the value. 

Also added myself to the code owners since I've touched this component enough times now that at the very least I'm probably someone to be notified for future reviews.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4941

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

```yaml
# Reset at midnight
time:
  - platform: homeassistant
    on_time:
      seconds: 0
      minutes: 0
      hours: 0
      then:
        - pulse_meter.set_total_pulses:
            id: meter
            value: 0

sensor:
  - platform: pulse_meter
    pin: GPIO05    
    id: meter
    total:
      name: "Total Pulses"
```

## Checklist:


TrentHouliston:
By this same logic it may also makes sense to publish a 0 on boot.

However that one I can see potentially causing issues if someone is attempting to restore a saved value on boot as it would publish 0, then their restored value.

Also I'm not sure where the best place to publish a 0 on boot would be...

ssieb:
Maybe it would be worth considering adding support for restore value.

TrentHouliston:
I think for this PR it's complete as is, and there is a method that you can use to restore on boot anyway (an on boot lambda). It might be better to do it that way since then you could restore from other sources than flash.

ssieb (approved):

PR #5479, opened 2023-10-04
Fix: Incorporate unique serial number in preference's hash for multiple Sensirion sensors

kahrendt:

Currently, the ``sgp4x``, ``spg30``, and ``sen5x`` components store the baseline values in the preferences using a hash derived from the compilation time. If you have multiple sensors on different I2C buses, then the same baseline preference will be used for each sensor. This PR changes it so that the hash computation uses the compilation time (so the baseline still resets with any new firmware) and each sensor's unique serial number.

This PR isn't a breaking change, as the current behavior will already recompute the baseline after flashing a new firmware.

I have tested it locally with two SGP40 sensors. The ``sgp30`` and ``sen5x`` components should also work, as the changes follow the same logic. The existing test cases already have the baseline stored (by default), so the modifications are tested at least through compilation.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

  - platform: sgp4x
    i2c_id: multplex0channel0
    store_baseline: true
    voc:
      name: "VOC Index 0"
    compensation:
      temperature_source: temp_sensor
      humidity_source: humidity_sensor
  - platform: sgp4x
    i2c_id: multplex0channel1
    store_baseline: true
    voc:
      name: "VOC Index 1"
    compensation:
      temperature_source: temp_sensor
      humidity_source: humidity_sensor    
```

## Checklist:

### DIFF
--- a/esphome/components/sen5x/sen5x.cpp
+++ b/esphome/components/sen5x/sen5x.cpp
@@ -1,5 +1,6 @@
 #include "sen5x.h"
 #include "esphome/core/hal.h"
+#include "esphome/core/helpers.h"
 #include "esphome/core/log.h"
 #include <cinttypes>
 
@@ -135,8 +136,8 @@ void SEN5XComponent::setup() {
       ESP_LOGD(TAG, "Firmware version %d", this->firmware_version_);
 
       if (this->voc_sensor_ && this->store_baseline_) {
-        uint32_t combined_serial = (uint32_t(this->serial_number_[0]) << 24) |
-                                   (uint32_t(this->serial_number_[1]) << 16) | (uint32_t(this->serial_number_[2]));
+        uint32_t combined_serial =
+            encode_uint24(this->serial_number_[0], this->serial_number_[1], this->serial_number_[2]);
         // Hash with compilation time and serial number
         // This ensures the baseline storage is cleared after OTA
         // Serial numbers are unique to each sensor, so mulitple sensors can be used without conflict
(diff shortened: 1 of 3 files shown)


esphome/components/sen5x/sen5x.cpp
```diff
@@ -135,9 +135,12 @@ void SEN5XComponent::setup() {
       ESP_LOGD(TAG, "Firmware version %d", this->firmware_version_);
 
       if (this->voc_sensor_ && this->store_baseline_) {
-        // Hash with compilation time
+        uint32_t combined_serial = (uint32_t(this->serial_number_[0]) << 24) |
+                                   (uint32_t(this->serial_number_[1]) << 16) | (uint32_t(this->serial_number_[2]));
```
jesserockz: This probably needs formatting and also `#include "esphome/core"helpers.h"
```suggestion
        uint32_t combined_serial = encode_uint24(this->serial_number_[0], this->serial_number_[1], this->serial_number_[2]);
```

jesserockz (changes requested):

esphome/components/sen5x/sen5x.cpp
kahrendt: Thanks!

jesserockz (approved):

PR #5483, opened 2023-10-05
Bug: Unwanted change resistance in x9c component

fizista:

There is an unwanted change in the resistance value.

Just call set_level(20) several times. The resistance will change.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** n/a

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
switch:
  - platform: template
    id: test_repeat
    name: "X9C Const value 20"
    restore_mode: ALWAYS_OFF
    optimistic: True
    on_turn_on:
      - output.set_level: 
          id: x9c_pot
          level: 20
    on_turn_off:
      - output.set_level: 
          id: x9c_pot
          level: 20

output:
  - platform: x9c
    id: x9c_pot
    cs_pin: GPIO15
    inc_pin: GPIO4
    ud_pin: GPIO2
    initial_value: 0.01
    inverted: True

```

## Checklist:


EtienneMD:
I see what you are saying, chip should not be selected with INC pin in High state. It never is thought as the pin is selected high prior to CS pin being lowered. Any advantage in leaving it high all the time when not in use?

fizista:
You're right. The high state on INC is not needed.

esphome/components/x9c/x9c.cpp
```diff
@@ -25,7 +29,6 @@ void X9cOutput::trim_value(int change_amount) {
 
   delayMicroseconds(100);  // Let value settle
 
-  this->inc_pin_->digital_write(false);
```
EtienneMD: I'd leave this in, as this prevents saving the counter position to non-volatile memory. An action could be added in the future to enable that feature.

EtienneMD (changes requested):

jesserockz (approved):

PR #5486, opened 2023-10-06
Add support for PN7160

kbx81:

Adds support for PN7160 NFCCs.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3241

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

See [documentation preview](https://deploy-preview-3241--esphome.netlify.app/components/binary_sensor/pn7160).

## Checklist:


esphome/components/pn7160/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+async def pn7160_simple_action_to_code(config, action_id, template_arg, args):
+    paren = await cg.get_variable(config[CONF_ID])
+    return cg.new_Pvariable(action_id, template_arg, paren)
+
+
+async def setup_pn7160(var, config):
+    await cg.register_component(var, config)
+
+    if CONF_DWL_REQ_PIN in config:
+        pin = await cg.gpio_pin_expression(config[CONF_DWL_REQ_PIN])
```
jesserockz: Can we change all of these to the new format? (same for 7150)
```suggestion
    if dwl_req_pin_config := config.get(CONF_DWL_REQ_PIN):
        pin = await cg.gpio_pin_expression(dwl_req_pin_config)
```

esphome/components/pn7160/automation.h
```diff
@@ -0,0 +1,142 @@
...
+
+template<typename... Ts> class EmulationOffAction : public Action<Ts...> {
+ public:
+  explicit EmulationOffAction(PN7160 *a_pn7160) : pn7160_(a_pn7160) {}
+
+  void play(Ts... x) override { this->pn7160_->set_tag_emulation_off(); }
+
+ protected:
+  PN7160 *pn7160_;
+};
```
jesserockz: Similar to `Conditions` these can be `Parented<>` and drop the constructor and protected field

```suggestion
template<typename... Ts> class EmulationOffAction : public Action<Ts...>, public Parented<PN7160> {
 public:
  void play(Ts... x) override { this->pn7160_->set_tag_emulation_off(); }
};
```

esphome/components/pn7160/binary_sensor.py
jesserockz, line 1: Maybe this should be a common `nfc` binary sensor, similar to `touchscreen`?

jesserockz (changes requested):

esphome/components/pn7160/__init__.py
kbx81: Done

esphome/components/pn7160/automation.h
kbx81: Done

esphome/components/pn7160/binary_sensor.py
kbx81: That's a good idea -- maybe good to do in a follow-up PR and make it so the other NFC components can use it, too?

jesserockz: If this becomes a followup, it will be breaking for this brand new binary sensor platform just to remove it again...

kbx81: Good point 😅

kbx81: (For posterity) The `binary_sensor` was removed for now; we'll implement it as its own component before too long.

jesserockz (approved):

PR #5497, opened 2023-10-08
Fixed precision for Nextion sensor with float values

olegtarasov:

Nextion Sensor component doesn't work correctly if `precision` is set for `float` components. Nextion doesn't support floats natively, it just emulates them by multiplying or dividing integers by powers of 10. Nextion sensor in ESPHome already has `precision` property, but it only multiplies values when sending them to Nextion. When values are received from Nextion, they should be divided by the appropriate amount, but in fact they are not. This pull request fixes that.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml
esphome:
  name: test

esp32:
  board: mhetesp32minikit
  framework:
    type: arduino

uart:
  id: uart_2
  rx_pin: GPIO16
  tx_pin: GPIO17
  baud_rate: 115200

sensor:
  - platform: nextion
    id: test_sensor
    name: "Test sensor"
    component_name: page0.x0
    precision: 1
    update_interval: 1s
```

## Checklist:

### DIFF
--- a/esphome/components/nextion/sensor/nextion_sensor.cpp
+++ b/esphome/components/nextion/sensor/nextion_sensor.cpp
@@ -76,9 +76,15 @@ void NextionSensor::set_state(float state, bool publish, bool send_to_nextion) {
     }
   }
 
+  float published_state = state;
   if (this->wave_chan_id_ == UINT8_MAX) {
     if (publish) {
-      this->publish_state(state);
+      if (this->precision_ > 0) {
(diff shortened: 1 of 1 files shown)


kbx81 (approved):
LGTM -- thanks for fixing this!

PR #5505, opened 2023-10-09
Don't allow entity category "CONFIG" for sensors

edenhaus:

The configuration category "config" is for read/write entities that control things. A sensor can never control things (and thus the change triggering because of this change are incorrect to begin with).
See https://github.com/home-assistant/core/pull/101471

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/sensor/__init__.py
+++ b/esphome/components/sensor/__init__.py
@@ -190,9 +190,7 @@ def validate_datapoint(value):
 
 
 _SENSOR_ENTITY_CATEGORIES = {
-    k: v
-    for k, v in cv.ENTITY_CATEGORIES.items()
-    if k != ENTITY_CATEGORY_CONFIG
+    k: v for k, v in cv.ENTITY_CATEGORIES.items() if k != ENTITY_CATEGORY_CONFIG
 }
 
 
--- a/esphome/components/sensor/__init__.py
+++ b/esphome/components/sensor/__init__.py
@@ -85,6 +85,7 @@ from esphome.const import (
     DEVICE_CLASS_WATER,
     DEVICE_CLASS_WEIGHT,
     DEVICE_CLASS_WIND_SPEED,
+    ENTITY_CATEGORY_CONFIG,
 )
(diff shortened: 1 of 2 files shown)


esphome/components/sensor/__init__.py
```diff
@@ -188,6 +189,17 @@ def validate_datapoint(value):
     return validate_datapoint({CONF_FROM: cv.float_(a), CONF_TO: cv.float_(b)})
 
 
+_SENSOR_ENTITY_CATEGORIES = {
+    k: v
+    for k, v in cv._ENTITY_CATEGORIES.items()  # pylint: disable=protected-access
```
jesserockz: Happy with the overall change, but can we just change this to be non-protected instead?
```suggestion
    for k, v in cv.ENTITY_CATEGORIES.items()
```

jesserockz (changes requested):

esphome/components/sensor/__init__.py
edenhaus: Yes sure :) I thought that it was protected so everybody uses the function `entity_category` instead

esphome/components/sensor/__init__.py
```diff
@@ -188,6 +189,17 @@ def validate_datapoint(value):
     return validate_datapoint({CONF_FROM: cv.float_(a), CONF_TO: cv.float_(b)})
 
 
+_SENSOR_ENTITY_CATEGORIES = {
+    k: v
+    for k, v in cv.ENTITY_CATEGORIES.items()  # pylint: disable=protected-access
```
jesserockz: Still needed?
```suggestion
    for k, v in cv.ENTITY_CATEGORIES.items()
```

jesserockz (dismissed):
LGTM, Thanks

jesserockz (approved):

PR #5511, opened 2023-10-11
Add Keeloq RF protocol

marshn:

A simple implementation for the KeeLoq protocol for the Microchip HCS301 encoder chip. It can send and receive the
keeloq packets but does not verify or generate the encrypted parts of the data. It is useful for receiving signals from
keyfobs and can also be used with receivers which do not rigorously check the encrypted data. 

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3255
## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
remote_receiver:
  pin: 
    number: GPIO37 
  dump: 
    - keeloq
  tolerance: 30%
  filter: 250us
  idle: 4ms
  on_keeloq:
      then:
        - lambda: |-
            ESP_LOGD("main", "KeeLoq: Addr=%04x Cmd=%02x Rpt=%d Bat=%d Encrypted=%08x", x.address, x.command, x.repeat, x.vlow, x.encrypted);

   on_ ... :
      then:
        - remote_transmitter.transmit_keeloq:
            address: '0x57ffe7b'
            command: '0x02'
            code: '0xd19ef0a9'
            repeat:
              times: 3
              wait_time: 15ms
```

## Checklist:


jesserockz:
I see a comment about random hopping code etc, but I don't see anywhere in the code that generates this?

marshn:
Thanks for taking a look Jesse.

You are correct that this PR can send and receive the keeloq packets, but does not verify or generate the encrypted (rolling code) parts of the data packet. It is still useful for receiving signals from keyfobs (to trigger automations, for example) and can also be used to send packets to keeloq receivers which do not rigorously check the encrypted data (which are more common than you might think). The into to the PR briefly explains this.

I my case I am using the PR to control an electric gate which incorporates a keeloq receiver. The gate software suffers from a common bug which means it will respond to retransmitted keeloq packets (simply alternating between two valid packets resyncronises the gate receiver each time one of the packets is sent). The packets are obtained using the receiver part of the PR and the dump function.

The ability to generate and verify the encrypted part of the data packet is only useful if the secret key (which is programmed into the transmitter) and/or the 'Manufacturers Key' is available. In most cases neither key is available. If keys became available the code to generate and/or check rolling code could be written as a lambda or incorporated into a new PR.

jesserockz:
> The ability to generate and verify the encrypted part of the data packet is only useful if the secret key (which is programmed into the transmitter) and/or the 'Manufacturers Key' is available. In most cases neither key is available. If keys became available the code to generate and/or check rolling code could be written as a lambda or incorporated into a new PR.

This is not exactly true as the flipper zero is able to decode the keys from a transmission and reproduce the correct rolling codes. Maybe the subghz code from flipperzero can be used as a reference to encode and decode within ESPHome correctly?

marshn:
The flipper zero code does incorporate some Manufacturer Keys (MF) which have been obtained by various means, so it can verify/generate rolling codes for a limited number of manufacturers whos MF keys have been compromised. Where Flipper Zero does not include an MF key is is only able to verify/check the unencrypted parts of the transmissions, as this PR does.

The code to generate/verify rolling codes could be incorporated in to esphome (PR https://github.com/esphome/esphome/pull/3733 did so, but was closed), but without the MF code is of no use. PR 3733 did not incorporate any keys.

The keys for my application are not avaiable, so generating/verifying the encrypted parts of the transmission is not possible.

It is possible, but expensive (in time/CPU cycles) to extract the MF keys from a keeloq receiver, but this would be beyond the scope of even advanced esphome users. A keeloq transmitter key can also be extracted from a keyfob, but this is only of use to generate/verify transmissions from that device, which is a more limited use.

The Flipper Code does not contain the MF keys in plain text, they are encrypted to discourage others from doing as you suggest. They are decrypted before use and are not stored in plaintext. The Flipper Zero authors request that the keeloq MF keys are not decrypted and circulated, and I would not do so.

In summary, yes the Flipper Zero code (or the original Microchip code) could be incorporated into esphome. However, the security of keeloq is in keys, without those is is not possible to verify/check encrypted packets. Only a subset of the keys are available in encrypted form in the Flipper code. This PR is useful without the ability to generate/verify the encrypted parts of keeloq transmissions.

jesserockz:
Haha no problem, I never looked into how it was done there, only knew that it was.

jesserockz (approved):

PR #5531, opened 2023-10-13
Add an ability to restore number state from flash

olegtarasov:

I've added an ability to save and load number states from flash, just as switches do. I've added 2 new configuration variables to number core:

* `restore_mode` — can be either `ENABLED` or `DISABLED`.
* `restore_default_state` — default value that is assigned to a number if `restore_mode` is ENABLED, but no state was saved in flash.

I've tried to keep the implementation as close as possible to the way switch restore is implemented. It allows to have some extensibility and to extend restore options, which are now binary, to acommodate more options.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
number:
  # Boiler config
  - platform: opentherm
    t_dhw_set:
      id: boilotron_dhw_setpoint
      name: "Hot Water target temperature"
      restore_mode: enabled
      restore_default_state: 60
```

## Checklist:


ssieb:
Most numbers don't make sense to have a restore option because they are representing the current state of a device.  That's why only the template number currently has a restore option.  You will need to go through all the users of number and remove the option from each one that shouldn't restore.  Just like was done for switch.

olegtarasov:
@ssieb You are right. My goal was to add number persistence to OpenTherm external component and I thought that other components would benefit from adding this to number core. I didn't take into account that this change entails much more work, which I'm not ready to take on right now, since I'm new to ESPHome project and didn't figure out all the implementation details. I will add number persistence directly to OpenTherm component for now.

Thank you for your feedback!

PR #5541, opened 2023-10-16
st7789 using mode3 when no cs pin

rletendu:

The ST7789 display component uses CS pin with SPI mode 0. But when the hwd implementation physically tight CS pin to GND, the idle state of SCK line should be high and leading the SPI mode being used move to mode 3 instead of 0.

## Types of changes

- [ ] Other

Issue seems having being discussed multiple times for users experimenting issues with ST7789 with no real outcomes
Recently while trying to use  ESP Home with such litle gadget [Geekmagic with ESPHome](https://community.home-assistant.io/t/installing-esphome-on-geekmagic-smart-weather-clock-smalltv-pro/618029) I figure out the issue.
This PR is only addressing the SPI mode concern with no CS pin, the ESP8266 memory limitation might be addressed later on in a different PR since this is more a workaround for frame buffer size limitation on this platform.

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040

The fix is fully transparent and detects if no CS pin being declared in the display st7789 component.
If no CS, the SPI mode is forced to 3.

```yaml

font:
  - file: 'OpenSans-Regular.ttf'
    id: font1
    size: 30

color:
    - id: color_green
      red: 0%
      green: 100%
      blue: 0%

spi:
  clk_pin: GPIO18
  mosi_pin: GPIO23
  interface: hardware
  id: spihwd

display:
  - platform: st7789v
    model: "Custom"
    spi_id: spihwd
    height: 240
    width: 240
    offset_height: 0
    offset_width: 0
    dc_pin: GPIO02
    reset_pin: GPIO04
    id: disp

        
    lambda: |-
      it.rectangle(0,  0, it.get_width()-1, it.get_height()-1, id(color_green));

```

## Checklist:

### DIFF
--- a/esphome/components/st7789v/st7789v.cpp
+++ b/esphome/components/st7789v/st7789v.cpp
@@ -15,7 +15,7 @@ void ST7789V::setup() {
 #endif
   // If no CS pin st7789 needs to operate in SPI mode 3 instead of mode 0
   if (this->cs_ == nullptr) {
-    ESP_LOGD(TAG,"No CS pin : Using SPI mode 3");
+    ESP_LOGD(TAG, "No CS pin : Using SPI mode 3");
     this->set_mode(spi::MODE3);
   }
   this->spi_setup();
--- a/esphome/components/st7789v/st7789v.cpp
+++ b/esphome/components/st7789v/st7789v.cpp
@@ -13,6 +13,11 @@ void ST7789V::setup() {
   this->power_.request();
   // the PowerSupply component takes care of post turn-on delay
 #endif
+  // If no CS pin st7789 needs to operate in SPI mode 3 instead of mode 0
+  if (this->cs_ == nullptr) {
+    ESP_LOGD(TAG,"No CS pin : Using SPI mode 3");
+    this->set_mode(spi::MODE3);
+  }
   this->spi_setup();
   this->dc_pin_->setup();  // OUTPUT
 


kbx81:
@clydebarrow any thoughts on this? 😄

jesserockz (changes requested):
This should be done in config validation

kbx81:
Ah, yes, sorry, commented before looking at the code (it was late).

As Jesse suggested, we will not unexpectedly change the SPI mode on the C/C++ side, especially now that it can be configured in yaml (see `spi_mode` in the updated docs [here](https://next.esphome.io/components/display/st7789v#configuration-variables)).

We are finding that some of these controllers/modules seem to work better with mode 0 while others work better with mode 3; I don't think we should assume that the CS pin plays into this. It's certainly possible, but it's probably best to let the user choose which works best for their display, especially since it is now quite easy to change in the yaml config.

rletendu:
Many thanks for looking into this and sorry did not have to look to the beta version...
The SPI configuration now visible at yaml level is definitely the most flexible option !

nielsnl68:
Is this something that could benefit other SPI controlled displays as well?

kbx81:
> Is this something that could benefit other SPI controlled displays as well?

Yes, I do believe so @nielsnl68. 😉

clydebarrow:
> Is this something that could benefit other SPI controlled displays as well?

Any other SPI device including displays that uses the pre-defined `spi_device_schema()` already has this implemented, as it is now handled in `register_spi_device()`.

PR #5550, opened 2023-10-17
Fan cycle no off option

skykingjwc:
Adds a feature to the fan component so that the fan will not turn off when calling the cycle_speed action if so desired by the user

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes nothing.

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3275

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
esphome:
    name: fantest

esp32:
  board: esp32dev
  framework:
    type: esp-idf   #also tested with Arduino

output:
  - platform: ledc  #using ledc because we need a float output
    id: my_output_id
    inverted: false
    pin: GPIO19

logger:

fan:
    - platform: speed
      id: test_speed_fan
      speed_count: 3
      off_speed_cycle: True
      output: my_output_id
      on_speed_set:
        then:
          - logger.log: Fan speed was changed!

binary_sensor:
  - platform: gpio
    name: Cycle Fan Speed
    id: cyclespeed
    pin:
      number: 18
      mode:
        input: true
        pulldown: true

    filters:
      - delayed_on: 10ms

    # Optional variables:
    on_press:
      - fan.cycle_speed: test_speed_fan
      - logger.log: "Cycle speed clicked"

```

## Checklist:


esphome/components/speed/fan/__init__.py
```diff
@@ -40,3 +42,6 @@ async def to_code(config):
     if CONF_DIRECTION_OUTPUT in config:
         direction_output = await cg.get_variable(config[CONF_DIRECTION_OUTPUT])
         cg.add(var.set_direction(direction_output))
+
+    if CONF_OFF_SPEED_CYCLE in config:
+        cg.add(var.set_offcycle(config[CONF_OFF_SPEED_CYCLE]))
```
jesserockz: Since you added a default, it will always be in config
```suggestion
   cg.add(var.set_offcycle(config[CONF_OFF_SPEED_CYCLE]))
```

.gitignore
jesserockz, line 1: If you want to add a line to here, please add it in its own PR.

esphome/components/fan/automation.h
```diff
@@ -60,13 +60,22 @@ template<typename... Ts> class CycleSpeedAction : public Action<Ts...> {
...
+
+        if (speed > supported_speed_count && off_speed_cycle) {
+          // was running at max speed, off speed cycle enabled, so turn off
           speed = 1;
           auto call = this->state_->turn_off();
           call.set_speed(speed);
           call.perform();
-        } else {
+        }
+        else if (speed > supported_speed_count && !off_speed_cycle) {
```
jesserockz:
```suggestion
        } else if (speed > supported_speed_count && !off_speed_cycle) {
```

esphome/components/fan/automation.h
```diff
@@ -60,13 +60,22 @@ template<typename... Ts> class CycleSpeedAction : public Action<Ts...> {
...
+          // was running at max speed, off speed cycle disabled,  set to lowest speed
+          auto call = this->state_->turn_on();
+          call.set_speed(1);
+          call.perform();
+        }
+        else {
```
jesserockz:
```suggestion
        } else {
```

tests/fantests.yaml
jesserockz, line 1: Can you add the config option and action test to test1.yaml please instead of a new file.

jesserockz (changes requested):

jesserockz (changes requested):
THinking about it, I actually feel like this should just be an option on the cycle action itself and not on the fan. It will be simpler and less code changes required

esphome/components/fan/automation.h
skykingjwc: Resolved by deac151c7f592b106161603b4d668ca0091efbb0

esphome/components/fan/automation.h
skykingjwc: Resolved by https://github.com/esphome/esphome/commit/deac151c7f592b106161603b4d668ca0091efbb0

esphome/components/speed/fan/__init__.py
skykingjwc: Resolved by b1ebc789db86ae2c1d2792cf3d48621c82a42f49

tests/fantests.yaml
skykingjwc: Done locally, but it fails because more code changes are required for other components.

skykingjwc:
> THinking about it, I actually feel like this should just be an option on the cycle action itself and not on the fan. It will be simpler and less code changes required

I agree after seeing the tests fail.
I would be happy to update the code to do it this way, could you elaborate a little more or point me to an example?  I'm not sure how to code an option for an action.   Thanks.

skykingjwc:
> > THinking about it, I actually feel like this should just be an option on the cycle action itself and not on the fan. It will be simpler and less code changes required
> 
> I agree after seeing the tests fail. I would be happy to update the code to do it this way, could you elaborate a little more or point me to an example? I'm not sure how to code an option for an action. Thanks.

Never mind, I used the output component as an example and I think I have it working as an action option.  Should I make a new PR and cancel this one, or just merge all my changes from the branch with the action option?

jesserockz:
Closing in favour of #5564

PR #5559, opened 2023-10-18
Improv Serial support via USB CDC and JTAG

kbx81:

Ensures that Improv Serial works over USB CDC & JTAG interfaces.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/logger/logger.cpp
+++ b/esphome/components/logger/logger.cpp
@@ -3,9 +3,13 @@
 
 #ifdef USE_ESP_IDF
 #include <driver/uart.h>
+
+#if defined(USE_ESP32_VARIANT_ESP32C3) || defined(USE_ESP32_VARIANT_ESP32C6) || defined(USE_ESP32_VARIANT_ESP32S3) || \
+    defined(USE_ESP32_VARIANT_ESP32H2)
 #include <driver/usb_serial_jtag.h>
 #include <esp_vfs_dev.h>
 #include <esp_vfs_usb_serial_jtag.h>
+#endif
 
 #include "freertos/FreeRTOS.h"
 #include "esp_idf_version.h"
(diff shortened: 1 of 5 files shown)


esphome/components/logger/logger.cpp
```diff
@@ -120,19 +185,19 @@ void HOT Logger::log_message_(int level, const char *tag, int offset) {
...
-        uart_ == UART_SELECTION_USB_SERIAL_JTAG
+        this->uart_ == UART_SELECTION_USB_SERIAL_JTAG
 #elif defined(USE_ESP32_VARIANT_ESP32S3)
-        uart_ == UART_SELECTION_USB_CDC || uart_ == UART_SELECTION_USB_SERIAL_JTAG
+        this->uart_ == UART_SELECTION_USB_CDC || this->uart_ == UART_SELECTION_USB_SERIAL_JTAG
 #else
         /* DISABLES CODE */ (false)  // NOLINT
 #endif
     ) {
       puts(msg);
```
jesserockz, line 197: Should we use the same `esp_usb_console_write_buf(...)` and `usb_serial_jtag_write_bytes(...)` for logging with those interfaces instead of `puts` here?

kbx81: We certainly can but I left it alone if only because of its elegance. Its use here doesn't seem to present any issues/conflicts with the new USB bits here and we'd be trading a single line of code for...a lot more. haha
Still, if you think we should (arguably for consistency), I'm game to re-work it.

jesserockz (approved):

PR #5563, opened 2023-10-18
Incorrect ESP32 Strapping PIN Defined

descipher:

WARNING GPIO4 is a Strapping PIN and should be avoided.
GPIO4 should be GPIO5

## Types of changes

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040


jesserockz (approved):
Oops, thanks

PR #5602, opened 2023-10-26
Add area (zone) to esphome core config to be suggested through API and MQTT.

Cossid:

Add area (zone) to esphome core config to be suggested through API and MQTT.

Related PRs:
- https://github.com/esphome/aioesphomeapi/pull/607
- https://github.com/home-assistant/core/pull/102834

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3297

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [X] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: light-switch
  area: Living Room

```

## Checklist:


Cossid:
Note, this add a `suggested_area` for MQTT, and adds a `area` value to the API message/DeviceInfo, but I do not currently have a dev environment for Home Assistant, so this will need changes on the ESPHome API integration in Home Assistant to consume the new variable.

esphome/components/api/api_pb2.h
jesserockz, line 1: As per the comment in the header of this file, it is autogenerated from `api.proto` with the mentioned script

jesserockz (changes requested):
Onboard with the change, but the api side needs to be done properly with api.proto.

Cossid:
Looking into it.

Cossid:
@jesserockz I can't seem to get protobuf-compiler to install, even in a virtual environment via pip.  I have adjusted the proto file and manually adjusted the file it would have compiled accordingly.  Is there a better way to process/verify this?

jesserockz:
The devcontainer/ `ghcr.io/esphome/esphome:lint` should have the tooling installed

Cossid:
`:lint` does not appear to be valid, and `:dev` does not appear to have protobuf-compiler

jesserockz:
Sorry, I typed it from memory while out, its `ghcr.io/esphome/esphome-lint:dev`

Cossid:
Thanks for updating, I wasn't finding `protoc` available in esphome-lint:dev either.
Curious, why 16 vs 15?

jesserockz:
15 is taken already by the `bluetooth_proxy_feature_flags`

Cossid:
Ah, it was out of order, I see, thanks!

jesserockz:
Just to match mqtt, ive changed the protoc name

Cossid:
Sure, I was on the fence about that myself.

jesserockz:
Maybe the yaml config should match too :thinking:

jesserockz:
Im currently writing the HA core side too

Cossid:
the "suggested" part didn't even occur to me until I saw the pre-existing constant, but I think I'd prefer to leave it as just `area` in the esphome config.  Aside from HA integration, I've got thoughts of using this field for organizing alternate dashboard listings, in which case "suggested" wouldn't really fit as well.

jesserockz:
image

jesserockz (approved):

nagyrobi:
It would have been nice to consider (perhaps an override) option for each component to be in a different area.

There are use cases when ESPHome acts as a central hub for switches, sensors, etc. these being spread out in different areas.

But I guess this needs support from HA side too..

jesserockz:
@nagyrobi You can still move entities to a different area inside Home Assistant. They can be individually changed from the device area. 

This feature is only the "suggested area" that pre-fills the area selection when you first add a device to HA.

PR #5615, opened 2023-10-28
Skip gpio validation

amcfague:

Adds an option to skip validation on pins. For custom boards, there are often differing usages--for instance, the [Spotpear ESP32C3](https://www.spotpear.com/index/product/detail/id/1232.html) uses GPIO12 and 13 to control the LEDs, unlike on the devkit. This enables skipping validation checks for specific pins.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3699

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esp32:
  board: esp32s3box
  variant: ESP32S3

output:
  - platform: gpio
    id: board_info_ed
    pin:
      # This pin is reserved on the ESP32S3!
      number: 26
      ignore_pin_validation_error: true
```

## Checklist:


jesserockz:
If the `board` is known, it can be added to the `esp32/boards.py` file which will allow the known pins to be used for general purpose. Validation is already skipped for listed pins in that file.

amcfague:
> If the `board` is known, it can be added to the `esp32/boards.py` file which will allow the known pins to be used for general purpose. Validation is already skipped for listed pins in that file.

To do so requires also adding platforms to [platform-esspresif32](https://github.com/platformio/platform-espressif32/new/develop/boards), which is a bit more in depth (and for a lot of these cheapo boards, probably not worthwhile).

jesserockz:
In that case, I feel like it should be a similar option to https://github.com/esphome/esphome/pull/5287

amcfague:
> In that case, I feel like it should be a similar option to #5287

Sounds very reasonable; let me update.

amcfague:
Hey @jesserockz, I combined what would have been two overlapping functions into one, `validate_configs`, that runs on the entire object.

esphome/components/esp32/gpio.py
```diff
@@ -148,6 +153,20 @@ def validate_supports(value):
         )
 
     value = _esp32_validations[variant].usage_validation(value)
+
+    ignore_pin_validation_warning = value[CONF_IGNORE_PIN_VALIDATION_WARNING]
```
jesserockz: Similar to the `CONF_IGNORE_STRAPPING_WARNING`, this should also raise an error if you set `ignore_pin_validation_warning: true` and the pin is not one that would normally raise an error. This is to stop people putting it on every pin config.

amcfague: Good idea! I've added this.

esphome/const.py
```diff
@@ -337,6 +337,7 @@
 CONF_IF = "if"
 CONF_IGNORE_EFUSE_MAC_CRC = "ignore_efuse_mac_crc"
 CONF_IGNORE_OUT_OF_RANGE = "ignore_out_of_range"
+CONF_IGNORE_PIN_VALIDATION_WARNING = "ignore_pin_validation_warning"
```
jesserockz: Should this be error instead of warning?
```suggestion
CONF_IGNORE_PIN_VALIDATION_ERROR = "ignore_pin_validation_error"
```

amcfague:
Hey @jesserockz! It's been a minute, but I merged with dev and this should be all up to date now, if you wanted to give it a pass!

jesserockz (approved):

PR #5630, opened 2023-10-30
Fix armv7 docker builds

bdraco:
https://github.com/esphome/esphome/pull/5606 started failing 4 days ago due to the openssl security update

- fix the openssl version
- pin the compiler version so we get better errors in the future


bdraco:
Looks like its the openssl security update causing the issue

kbx81 (approved):

bdraco:
Cheers

PR #5635, opened 2023-10-31
Add SHT2X

dmaasland:

Add the SHT2x sensor component.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3347

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
   sensor:
    - platform: sht2x
       temperature:
       humidity:
         name: "Living Room Humidity"
```

## Checklist:


esphome/components/sht2x/sensor.py
```diff
@@ -0,0 +1,58 @@
...
+)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await i2c.register_i2c_device(var, config)
+
+    if CONF_TEMPERATURE in config:
+        sens = await sensor.new_sensor(config[CONF_TEMPERATURE])
```
kahrendt:
```
    if temperature_config := config.get(CONF_TEMPERATURE):
        sens = await sensor.new_sensor(temperature_config)
```

esphome/components/sht2x/sensor.py
```diff
@@ -0,0 +1,58 @@
...
+        cg.add(var.set_temperature_sensor(sens))
+
+    if CONF_HUMIDITY in config:
+        sens = await sensor.new_sensor(config[CONF_HUMIDITY])
```
kahrendt:
```
    if humidity_config := config.get(CONF_HUMIDITY):
        sens = await sensor.new_sensor(humidity_config)
```

esphome/components/sht2x/sht2x.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+uint16_t SHT2XComponent::read_raw_value_() {
+  uint8_t buffer[3];
+  uint8_t crc;
+  uint16_t result;
+
+  if (this->read(buffer, 3) != i2c::ERROR_OK) {
+    this->status_set_warning();
+  }
+  crc = crc8_(buffer, 2);
```
kahrendt: `crc = this->crc8_(buffer, 2);`

esphome/components/sht2x/sht2x.cpp
```diff
@@ -0,0 +1,123 @@
...
+  return result;
+}
+
+float SHT2XComponent::get_temperature_() {
+  if (this->write(&SHT2X_COMMAND_TEMPERATURE, 1) != i2c::ERROR_OK) {
+    ESP_LOGE(TAG, "Reading temperature error");
+  };
+
+  delay(SHT2X_DELAY_TEMPERATURE);  // NOLINT
+  uint16_t raw_temperature = read_raw_value_();
```
kahrendt: Change this everywhere you call a class function (see https://esphome.io/guides/contributing.html#codebase-standards)
`  uint16_t raw_temperature = this->read_raw_value_();`

esphome/components/sht2x/sht2x.cpp
```diff
@@ -0,0 +1,123 @@
...
+  float temperature = -46.85 + (175.72 / 65536.0) * raw_temperature;
+  return temperature;
+}
+
+float SHT2XComponent::get_humidity_() {
+  if (this->write(&SHT2X_COMMAND_HUMIDITY, 1) != i2c::ERROR_OK) {
+    ESP_LOGE(TAG, "Reading humidity error");
+  }
+
+  delay(SHT2X_DELAY_HUMIDITY);  // NOLINT
```
kahrendt: Avoid the delay here too

esphome/components/sht2x/sht2x.h
```diff
@@ -0,0 +1,34 @@
...
+#include <cinttypes>
+
+namespace esphome {
+namespace sht2x {
+
+/// This class implements support for the SHT2x family of temperature+humidity i2c sensors.
+class SHT2XComponent : public PollingComponent, public sensirion_common::SensirionI2CDevice {
+ public:
+  void set_temperature_sensor(sensor::Sensor *temperature_sensor) { temperature_sensor_ = temperature_sensor; }
+  void set_humidity_sensor(sensor::Sensor *humidity_sensor) { humidity_sensor_ = humidity_sensor; }
```
kahrendt: Don't forget to add ``this->`` in this file as well.

kahrendt (changes requested):
I don't have a sensor to test it with, but I looked through your code and added comments for a few things that I have missed in my previous PRs. Nice work overall!

esphome/components/sht2x/sht2x.cpp
kahrendt: I must have dropped the first comment for get_temperature. You should use ``set_timeout`` instead of NOLINT and a delay to avoid blocking the main loop. It will require you to restructure your code a bit to make it work.

dmaasland:
Thanks for your review @kahrendt. I think i've resolved all of them :)

esphome/components/sht2x/sht2x.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+void SHT2XComponent::update() {
+  if (this->status_has_warning()) {
+    ESP_LOGD(TAG, "Retrying to reconnect the sensor.");
+    this->write_command(SHT2X_COMMAND_SOFT_RESET);
+  }
+
+  this->handle_humidity_();
+  this->set_timeout(SHT2X_DELAY_HUMIDITY + 10, [this]() { this->handle_temperature_(); });
```
kahrendt: Move the check for if ``humidity_sensor_`` and ``temperature_sensor_`` are not ``nullptr`` here. If a sensor is not configured, then the handle code will never need to run in the first place (assuming the SHT2X is okay with only ever requesting one type of measurement).

dmaasland: Done! Should be fine. Added a small check to remove the delay if the humidity sensor isn't available.

kahrendt (approved):
Looks good to me! Hopefully a maintainer can take a look soon as well so that it can get merged.

PR #5636, opened 2023-10-31
Ensure that all uses of strncpy in wifi component are safe.

kpfleming:

`strncpy` must not be allowed to overwrite the last byte in the destination buffer.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/wifi/wifi_component.cpp
+++ b/esphome/components/wifi/wifi_component.cpp
@@ -261,8 +261,8 @@ void WiFiComponent::set_sta(const WiFiAP &ap) {
 void WiFiComponent::clear_sta() { this->sta_.clear(); }
 void WiFiComponent::save_wifi_sta(const std::string &ssid, const std::string &password) {
   SavedWifiSettings save{};
-  strncpy(save.ssid, ssid.c_str(), sizeof(save.ssid));
-  strncpy(save.password, password.c_str(), sizeof(save.password));
+  strncpy(save.ssid, ssid.c_str(), sizeof(save.ssid) - 1);
+  strncpy(save.password, password.c_str(), sizeof(save.password) - 1);
   this->pref_.save(&save);
   // ensure it's written immediately
   global_preferences->sync();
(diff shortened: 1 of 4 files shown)


jesserockz (approved):

bergdahl:
I might be wrong, but if the string is len-1 it copies all characters to the buffer, but it will not add the \0 character.

kpfleming:
That is correct. In all cases where `strncpy` is used in ESPHome the buffer is initialized to all-zeros (or the last byte is set to zero) before `strncpy` is called; the result is that the string in the destination buffer will be zero-terminated, even if the source string was longer than the buffer could hold.

jesserockz:
This has broken cases where people actually use 32 character SSIDs

nagyrobi:
> This has broken cases where people actually use 32 character SSIDs

Whoops.

clydebarrow:
I think the misconception here was that the ssid and password buffers should have been null-terminated, whereas it looks to me that they were simply designed to be null-padded.

kpfleming:
Sorry about that, but the use of `strncpy` to copy them around does imply that they are C strings, which are supposed to be null-terminated :-)

Something will need to be changed in this code when PlatformIO/ESP-IDF are upgraded to a version which includes GCC 12 (or possibly even GCC 11), since it will generate errors for these calls to `strncpy` with a length that matches the size of the destination buffer. That's what prompted this PR in the first place (building with PlatformIO 6.4.0 and ESP-IDF 5.1.0).

clydebarrow:
So it should be changed to `memcpy` with a calculated length (minimum of destination and source length). `strncpy` in general is simply a Bad Idea, though in this case it happens to achieve the desired result.

Of course if you boil it down, null-terminated strings are a Bad Idea :-)

PR #5644, opened 2023-10-31
Null topic_prefix disables MQTT publishing/subscription unless topic is explicitly configured

kahrendt:

This PR is an extension of esphome/esphome#5213.

When MQTT's``topic_prefix`` is null, this PR sets the default behavior for MQTT to not publish or subscribe to any topics. If a topic is explicitly configured, then it will publish or subscribe as expected. This allows users to quickly disable most components from using MQTT.

This is potentially a breaking change. It was possible to set the ``topic_prefix`` to null before, and topics would use no prefix, though I don't believe this is commonly done. There is no change in behavior if a user does not configure a ``topic_prefix`` or if the ``topic_prefix`` configured is not null.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3316

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

mqtt:
  broker: 10.0.0.2
  username: livingroom
  topic_prefix: null

```

## Checklist:

### DIFF
--- a/esphome/components/mqtt/__init__.py
+++ b/esphome/components/mqtt/__init__.py
@@ -136,48 +136,44 @@ def validate_config(value):
     # If the topic prefix is not null and these messages are not configured, then set them to the default
     # If the topic prefix is null and these messages are not configured, then set them to null
     if CONF_BIRTH_MESSAGE not in value:
-        out[CONF_BIRTH_MESSAGE] = (
-            {
+        if topic_prefix != "":
+            out[CONF_BIRTH_MESSAGE] = {
                 CONF_TOPIC: f"{topic_prefix}/status",
                 CONF_PAYLOAD: "online",
                 CONF_QOS: 0,
                 CONF_RETAIN: True,
             }
-            if topic_prefix != ""
-            else {}
-        )
+        else:
+            out[CONF_BIRTH_MESSAGE] = {}
     if CONF_WILL_MESSAGE not in value:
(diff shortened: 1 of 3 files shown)


esphome/components/mqtt/__init__.py
```diff
@@ -133,33 +133,51 @@ def validate_config(value):
...
+        out[CONF_BIRTH_MESSAGE] = (
+            {
+                CONF_TOPIC: f"{topic_prefix}/status",
+                CONF_PAYLOAD: "online",
+                CONF_QOS: 0,
+                CONF_RETAIN: True,
+            }
+            if topic_prefix != ""
+            else {}
+        )
```
jesserockz: While I like the use of ternary operator personally, I think it makes it harder to read this
```suggestion
        if topic_prefix != "":
            out[CONF_BIRTH_MESSAGE] = {
                CONF_TOPIC: f"{topic_prefix}/status",
                CONF_PAYLOAD: "online",
                CONF_QOS: 0,
                CONF_RETAIN: True,
            }
        else:
            out[CONF_BIRTH_MESSAGE] = {}
```

jesserockz (changes requested):

jesserockz (approved):

PR #5646, opened 2023-11-01
Migrate component finder into core code.

clydebarrow:

Move the function to find a given component within the global config into core code.

This follows on from https://github.com/esphome/esphome/pull/5634

## Types of changes

- [x] Other - new function in `core/config.py` for developer use.

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

The tests already added in the previous PR suffice for this.

### DIFF
--- a/esphome/components/pca9554/__init__.py
+++ b/esphome/components/pca9554/__init__.py
@@ -10,7 +10,7 @@ from esphome.const import (
     CONF_INVERTED,
     CONF_OUTPUT,
 )
-from esphome.core import CORE, ID
+from esphome.config import get_config_for_component
 
 CODEOWNERS = ["@hwstar", "@clydebarrow"]
 DEPENDENCIES = ["i2c"]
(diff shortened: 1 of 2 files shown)


clydebarrow:
@jesserockz continuing discussion here;

This moves the "find_component" functionality into core. I'm not sure a generic pin checker is feasible, since that would require a rigid schema for expander components. Open to suggestions though.

Are you happy with the logic for finding the component? I think it should be robust for all the use cases I can think of, and I have done some tests for both multi and non-multi conf which worked.

ssieb:
If the validation hasn't been done, the values you read might not be correct.  At what point do default values and conversions get done?

jesserockz:
Im pretty sure defaults and conversions are only done when that particular component is validated.

This kind of code would only be useful for pin expanders as every other component can have `FINAL_VALIDATE_SCHEMA`

jesserockz:
I decided to try play around with FINAL_VALIDATE a little bit and I think I have something working. Ill push my stuff up soon to its own PR

ssieb:
> Im pretty sure defaults and conversions are only done when that particular component is validated.

Doesn't that mean this usage would be risky?  The number of pins could be set by the default.

jesserockz:
Yes, lets move this to discord. Ill start a thread

PR #5649, opened 2023-11-01
wifi: Don't build SoftAP/DHCPS support unless 'ap' is in config.

kpfleming:

Optimization: don't build the SoftAP and DHCP server code (in the ESP-IDF framework) unless the ESPHome config includes `ap`; don't build AP setup code in the `wifi` component either in that case.

Also updates version of `flake8` in pre-commit config to avoid a false negative when checking `__init__.py` in the wifi component.

## Types of changes

- [X] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/wifi/__init__.py
+++ b/esphome/components/wifi/__init__.py
@@ -403,6 +403,10 @@ async def to_code(config):
             lambda ap: cg.add(var.set_ap(wifi_network(conf, ap, ip_config))),
         )
         cg.add(var.set_ap_timeout(conf[CONF_AP_TIMEOUT]))
+        cg.add_define("ENABLE_WIFI_AP")
+    elif CORE.is_esp32 and CORE.using_esp_idf:
+        add_idf_sdkconfig_option("CONFIG_ESP_WIFI_SOFTAP_SUPPORT", False)
+        add_idf_sdkconfig_option("CONFIG_LWIP_DHCPS", False)
 
     cg.add(var.set_reboot_timeout(config[CONF_REBOOT_TIMEOUT]))
     cg.add(var.set_power_save_mode(config[CONF_POWER_SAVE_MODE]))
--- a/esphome/components/wifi/__init__.py
+++ b/esphome/components/wifi/__init__.py
@@ -403,7 +403,7 @@ async def to_code(config):
             lambda ap: cg.add(var.set_ap(wifi_network(conf, ap, ip_config))),
         )
         cg.add(var.set_ap_timeout(conf[CONF_AP_TIMEOUT]))
-        cg.add_define("ENABLE_WIFI_AP")
+        cg.add_define("USE_WIFI_AP")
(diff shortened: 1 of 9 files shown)


ChillingSilence:
cACK, sounds nice for me as someone who has a bunch of PoE devices that don't have any WiFi / AP support :)

esphome/components/wifi/__init__.py
```diff
@@ -397,6 +397,10 @@ def add_sta(ap, network):
             lambda ap: cg.add(var.set_ap(wifi_network(conf, ap, ip_config))),
         )
         cg.add(var.set_ap_timeout(conf[CONF_AP_TIMEOUT]))
+        cg.add_define("ENABLE_WIFI_AP")
```
jesserockz: To match other defines in ESPHome we follow `USE_<component>[_<something>]`

```suggestion
        cg.add_define("USE_WIFI_AP")
```

.pre-commit-config.yaml
jesserockz, line 1: I have bumped this is its own pr https://github.com/esphome/esphome/pull/5658

esphome/components/wifi/wifi_component.cpp
```diff
@@ -152,6 +154,7 @@ void WiFiComponent::loop() {
     }
 
     if (this->has_ap() && !this->ap_setup_) {
+#if ENABLE_WIFI_AP
```
jesserockz: This might as well be one line above so there is no empty if block

jesserockz (changes requested):

kpfleming:
> cACK, sounds nice for me as someone who has a bunch of PoE devices that don't have any WiFi / AP support :)

This PR is unlikely to be useful for you; if your YAML does not have the `wifi` component then none of the WiFi code (all of it, not just SoftAP) should be linked into the binary. It's still being compiled, and there's no easy to way to avoid compiling *all* of the WiFi code.

esphome/components/wifi/__init__.py
kpfleming: Making this change now; should I also add `USE_WIFI_AP` to `defines.h` for static analyzer/IDE usage?

esphome/components/wifi/wifi_component.cpp
kpfleming: Good point, I've found a few other places where the entire block should be inside the ifdef-endif too.

esphome/components/wifi/__init__.py
jesserockz: Yes =)

jesserockz (commented):
Please do add to `defines.h` then this PR is good to merge

kpfleming:
Done, and CI tests found another bug which has also been fixed :-)

jesserockz (approved):

PR #5651, opened 2023-11-01
Add Healthcheck to Dockerfile

JcMinarro:
Add [Healthcheck](https://docs.docker.com/engine/reference/builder/#healthcheck) status to the docker container

## Types of changes

- [ ] Other

This functionality was initially added on https://github.com/esphome/esphome/pull/1492 and updated on https://github.com/esphome/esphome/pull/1517, but looks like "by error" It was removed on this [refactor PR ](https://github.com/esphome/esphome/pull/2338/files#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcL20-L22) (cc/ @OttoWinter) 

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -101,6 +101,10 @@ ENV USERNAME="" PASSWORD=""
 # Expose the dashboard to Docker
 EXPOSE 6052
 
+# Run healthcheck (heartbeat)
+HEALTHCHECK --interval=30s --timeout=30s \
+  CMD curl --fail http://localhost:6052/version -A "HealthCheck" || exit 1
+
 COPY docker/docker_entrypoint.sh /entrypoint.sh
 
 # The directory the user should mount their configuration files to


docker/Dockerfile
```diff
@@ -101,6 +101,10 @@ ENV USERNAME="" PASSWORD=""
 # Expose the dashboard to Docker
 EXPOSE 6052
 
+# Run healthcheck (heartbeat)
+HEALTHCHECK --interval=30s --timeout=30s \
+  CMD curl --fail http://localhost:6052/version -A "HealthCheck || exit 1
```
jesserockz: There is a single `"` on this line, is that correct?

JcMinarro: Ups, I can see a local change on my laptop  with this second `"` that wasn't been committed :facepalm:

JcMinarro: Updated :pray:

jesserockz (approved):

PR #5666, opened 2023-11-03
Checks for pins used in multiple places

clydebarrow:

Check that I/O pins are not (erroneously) used in more than one place for different functions.

Pin schema option `allow_other_uses: true` will suppress the errors on a pin-by-pin basis.

To facilitate implementation across different components, a new function in `pins.py` has been defined to provide a base schema for gpio pins.

~~Still to be done: update various I/O expander components to make use of this.~~

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3366

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [x] BK72xx
- [x] RTL87xx

```yaml
# Example config.yaml
binary_sensor:
  - platform: gpio
    internal: true
    name: Pin 1_1
    pin:
      pca9554: p_c_a
      other_uses: false
      number: 9

```

## Checklist:


esphome/core/config.py
```diff
@@ -169,6 +170,7 @@ def validate_version(value: str):
             cv.Optional(
                 CONF_COMPILE_PROCESS_LIMIT, default=_compile_process_limit_default
             ): cv.int_range(min=1, max=multiprocessing.cpu_count()),
+            cv.Optional(CONF_CHECK_PIN_USE, default=True): cv.boolean,
```
jesserockz: I dont like global config options that can allow bad configs as tutorials add them and then everyone copy/pastes it into every config.

esphome/components/esp32/gpio.py
```diff
@@ -161,33 +157,22 @@ def validate_supports(value):
...
-        cv.GenerateID(): cv.declare_id(ESP32InternalGPIOPin),
-        cv.Required(CONF_NUMBER): validate_gpio_pin,
-        cv.Optional(CONF_MODE, default={}): cv.Schema(
-            {
-                cv.Optional(CONF_INPUT, default=False): cv.boolean,
-                cv.Optional(CONF_OUTPUT, default=False): cv.boolean,
-                cv.Optional(CONF_OPEN_DRAIN, default=False): cv.boolean,
-                cv.Optional(CONF_PULLUP, default=False): cv.boolean,
-                cv.Optional(CONF_PULLDOWN, default=False): cv.boolean,
-            }
```
jesserockz, line 177: Note that esp32 doesnt allow analog in the modes

esphome/components/libretiny/gpio.py
```diff
@@ -186,26 +186,11 @@ def validate_gpio_usage(value):
...
-                cv.Optional(CONF_OPEN_DRAIN, default=False): cv.boolean,
-                cv.Optional(CONF_PULLUP, default=False): cv.boolean,
-                cv.Optional(CONF_PULLDOWN, default=False): cv.boolean,
-            }
-        ),
-        cv.Optional(CONF_INVERTED, default=False): cv.boolean,
-    },
+BASE_PIN_SCHEMA = cv.All(
+    pins.gpio_base_schema(ArduinoInternalGPIOPin, validate_gpio_pin),
+    validate_gpio_usage,
```
jesserockz: This uses `add_extra` as there is potential for this to be extended and you cannot extend an `All`

esphome/pins.py
```diff
@@ -161,3 +212,49 @@ def check_strapping_pin(conf, strapping_pin_list, logger):
...
+    :param mode_validator: A validator function for the pin mode
+    :param invertable: If the pin supports hardware inversion
+    :return: A schema for the pin
+    """
+    mode_dict = dict(map(lambda m: (cv.Optional(m, default=False), cv.boolean), modes))
+    schema = cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(pin_type),
+            cv.Required(CONF_NUMBER): number_validator,
+            cv.Optional(CONF_OTHER_USES): cv.boolean,
```
jesserockz: Maybe a better name?
```suggestion
            cv.Optional(CONF_ALLOW_OTHER_USES): cv.boolean,
```

esphome/pins.py
```diff
@@ -10,16 +12,67 @@
...
+class PinRegistry(dict):
+    def __init__(self):
+        super().__init__()
+        self.pins_used = {}
+
+    def increment_count(self, component, config):
+        if (
+            CORE.raw_config[CONF_ESPHOME].get(CONF_CHECK_PIN_USE) is False
+            or CONF_NUMBER not in config
+            or config.get(CONF_OTHER_USES) is True
```
jesserockz: Similar to the ignore strapping warning and potentially the ignore_validation warning, I think that an error should be raised if the pin is only used once but this is set, to stop people putting it everywhere.

jesserockz (changes requested):

esphome/core/config.py
clydebarrow: Agreed, but the test yamls need a way of supressing this - I added the per-pin suppression for a couple but some have numerous instances where it is needed. Do you have any other ideas to achieve this or should I just bite the bullet and add it to all overloaded pins in the test yamls?

esphome/__main__.py
```diff
@@ -389,7 +389,8 @@ def command_config(args, config):
         output = re.sub(
             r"(password|key|psk|ssid)\: (.+)", r"\1: \\033[5m\2\\033[6m", output
         )
-    safe_print(output)
+    if not CORE.quiet:
+        safe_print(output)
```
jesserockz: "Unrelated" changes should go in their own PR for easier tracing and in case of needing to be reverted etc

jesserockz (changes requested):
Overall looks good. Please just split the `quiet` stuff into its own PR then this can merge :tada:

esphome/__main__.py
clydebarrow: Oh, forgot I had that in there - I needed it when cleaning up the test yamls because "esphome config" was the easiest way to find the issues and it insisted on spitting out the whole config on success.

jesserockz (approved):

PR #5670, opened 2023-11-03
Refactor ATM90E32 to reduce blocking time and improve accuracy.

descipher:

Simplify code and increase SPI data rate to 1Mhz.
Adds manufactures application guide calibration capability. (no user changes)
Adds Harmonic Power, Phase and Peak Current optional outputs.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

See updated docs https://github.com/esphome/esphome-docs/pull/3573

## Checklist:


esphome/components/atm90e32/atm90e32.cpp
```diff
@@ -156,19 +189,18 @@ void ATM90E32Component::dump_config() {
   LOG_SENSOR("  ", "Frequency", this->freq_sensor_);
   LOG_SENSOR("  ", "Chip Temp", this->chip_temperature_sensor_);
 }
-float ATM90E32Component::get_setup_priority() const { return setup_priority::DATA; }
+
+float ATM90E32Component::get_setup_priority() const { return setup_priority::BUS; }
```
jesserockz: This should be `IO` at the earliest as it relies on the spi bus to be setup.
```suggestion
float ATM90E32Component::get_setup_priority() const { return setup_priority::IO; }
```

esphome/components/atm90e32/atm90e32.h
```diff
@@ -38,38 +43,68 @@ class ATM90E32Component : public PollingComponent,
...
+    uint8_t num_reads = 5;
+    uint64_t total_value = 0;
+    for (int i = 0; i < num_reads; ++i) {
+      uint32_t measurement_value = read32_(ATM90E32_REGISTER_IRMS + phase, ATM90E32_REGISTER_IRMSLSB + phase);
+      total_value += measurement_value;
+    }
+    uint32_t average_value = total_value / num_reads;
+    uint32_t current_offset = ~average_value + 1;
+    return current_offset & 0xFFFF;  // Take the lower 16 bits
+  }
```
jesserockz: Can the contents of these be moved to the cpp file please?

esphome/components/atm90e32/atm90e32.cpp
```diff
@@ -6,75 +6,88 @@ namespace esphome {
 namespace atm90e32 {
 
 static const char *const TAG = "atm90e32";
-
+void ATM90E32Component::loop() {
+  const int32_t current_millis = millis();
+  if (current_millis - last_periodic_millis > 400) {
+    last_periodic_millis = current_millis;
```
jesserockz: Fetching the data should only be done once per `update` call.

Perhaps another solution could be to have a state machine {idle, update, publish}. `update()` can set from idle to update, `loop()` can fetch the data in `update` state then change state to `publish` which `loop()` can pick up next time around (~16ms later).

jesserockz (changes requested):

esphome/components/atm90e32/atm90e32.h
descipher: For sure, what's the threshold for moving em out to the cpp, 3 or more lines?

esphome/components/atm90e32/atm90e32.cpp
descipher: Ok, But I think the priority description should change then. It's not clear to me, do we set the priority of the component to a value lower than the interfaces it will use or at them?

Here the description is misleading unless we indicate that we need to be lower than i2c or spi when we depend on a BUS.

e.g.
const float esphome::setup_priority::BUS
For communication buses like i2c/spi

esphome/components/atm90e32/atm90e32.cpp
descipher: The objective here is to do a local SPI data fetch into an array store at a resonable frequency. The update overide does the publish from the pre-populated local array. I will look at a state machine to see if thats more effective. May I ask why the concern on this?

esphome/components/atm90e32/atm90e32.cpp
jesserockz: Lower number = lower priority = setup later

jesserockz: If a component is set at the same priority level, it could be setup out of order

esphome/components/atm90e32/atm90e32.h
jesserockz: To be honest, not sure...haha
But these functions are reading data from the bus, so they definitely don't belong in the header.

esphome/components/atm90e32/atm90e32.cpp
jesserockz: There is no point in updating local data every 400ms when the update interval could be 60 seconds. Its a waste of resource and cpu time when it is just thrown away.

The local store doesn't need to update any faster than the configured `update_interval`

esphome/components/atm90e32/atm90e32.cpp
descipher: Ok, I'm a bit confused by this .. we init the SPI interface in this component so wouldn't this require a higher priority.

e.g. from setup the first thing we call is:
  this->spi_setup(); and there is and option to have multiple instances in the configuration.

descipher: I think I see it now, SPI is a separate component and has it's own priority setting.

jesserockz: Yes, the SPI base component needs to do its own periperhal setup before any components try to register themselves with it.

esphome/components/atm90e32/atm90e32.cpp
descipher: Fair enough. I don't think I can do it the way you envisioned though, it only works well if the array data is ready to go b4 the update polling event. I can look at preempting but I think the exact polling pit may be unpredicatable. I will measure it and see.

jesserockz: Data doesnt have to be ready when `update` is called, it can take however long it needs before its actually published. `update` can be thought of a trigger to start the process to get sensors updated.

descipher: The accuracy suffers by doing that. See: https://github.com/CircuitSetup/Expandable-6-Channel-ESP32-Energy-Meter/issues/88

CircuitSetup: @descipher now that I think about it, performance could suffer with setups that are doing, for example, 24+ current channels at 60 second update intervals. Could the local polling be dependent on the update interval, maybe?

descipher: I don't think its a performance issue at all. It is maximum of 12 SPI channels at 2Mhz reading directly into memory. The processing time calculates as 2us of SPI data transfer processing per polling every 400ms. Which is nothing compared to the API publishing process of 46ms * 12  or 552ms unless you turn serial logging off. However it would be good if you could test that many since I cannot or at least somewhere close to it.

If we do the publishing during the local poll it will have a large delay between channels with 12 of them.

CircuitSetup: @descipher okay, that makes sense. I was thinking that storing 24+ current channels in memory every 400ms until the update is called would overrun the buffer, but you obviously have a better understanding of how that works than me. 

@jesserockz what needs to happen to get this pushed through?

esphome/components/atm90e32/atm90e32_reg.h
```diff
@@ -131,10 +131,12 @@ static const uint16_t ATM90E32_REGISTER_IOFFSETN = 0x6E;  // N Current Offset
 
 /* ENERGY REGISTERS */
 static const uint16_t ATM90E32_REGISTER_APENERGYT = 0x80;  // Total Forward Active
+static const uint16_t ATM90E32_REGISTER_APENERGY = 0x81;   // Forward Active Reg Base
 static const uint16_t ATM90E32_REGISTER_APENERGYA = 0x81;  // A Forward Active
```
danieltwagner, line 135: seems like all of these A/B/C register names will become unused once you reference them with base + offset, mind removing them?

esphome/components/atm90e32/atm90e32.cpp
danieltwagner: Could you expand on why reading these registers in `loop()` is better than reading them in `update()`? It's not obvious to me. My impression is that the issue was that reading these registers at 200 kHz might mean the value updates in between reads and so reading, say, voltage across different phases might be out of sync if new data arrives between reads.

The linked issue has a lengthy conversation but I take it this is the relevant insight?
> What does provide value is the quick polling local memory retrieval which have no additional delays.

Where the code runs (`update()` or `loop()`) shouldn't make a difference on how quickly these values are read. What this change would accomplish is minimizing time waiting for the SPI reads to finish before exiting `update()`, but otherwise it should be identical. Am I missing something?

esphome/components/atm90e32/atm90e32_reg.h
descipher: True, but I have not referenced everyone of them so only instances where I added the Base value can be cleaned and not all of them have a base. There is value as documentation allowing them to remain.

esphome/components/atm90e32/atm90e32.cpp
descipher: update() triggers only once and holds up other callbacks (single thread) while it processes. If we publish direct chip reads from update we will miss the single PIT physical chip read sample and thus the values will dither from channel to channel significantly. A phase publishing call takes too much time even at 2Mhz. We are not waiting on the read we are waiting on the publishing code to complete before the next direct read. This crosses the 220ms chip read frequency which results in value changes on the same phase over two diffent SPI ports. loop() is in a different scheduling context and is not a thread blocking contributor. @jesserockz and I spoke on this last night we are going to test a bool flag comming in on update() and then process everything out of loop() based on the flag state to see if that's a better solution than the 400ms local work interval. As long as that does not result in the 0 phase value read problem during an atm90e32 chip and esp boot cycle we should be in good shape.

danieltwagner: Thanks for the explanation, that makes sense. Could you read all values first (essentially the code that's currently in `loop()`) and then publish (the code that's currently in `update()`) after you read the values? That would seem conceptually simpler than a (binary) state machine where business logic is spread across two functions.

descipher: That approach was tried using a preempt strategy, it fails. The scheduler is random so there is no predictable time slot to prempt the update() interval. That also results in 0 value reads on restart and thats a show stopper.

danieltwagner: I'm not sure I'm following. Just to be clear, what I was suggesting was something like this:
```
void ATM90E32Component::update() {
  for (uint8_t phase = 0; phase < 3; phase++) {
    if (this->phase_[phase].voltage_sensor_ != nullptr) {
      this->phase_[phase].voltage_ = this->get_phase_voltage_(phase);
    }
  }
  
  [... read all other values ...]
  
  for (uint8_t phase = 0; phase < 3; phase++) {
    if (this->phase_[phase].voltage_sensor_ != nullptr) {
      this->phase_[phase].voltage_sensor_->publish_state(this->get_local_phase_voltage_(phase));
    }
  }
  
  [... publish all other values ...]
}
```

Why would this result in 0 value reads and where does the need to preempt anything come from?

danieltwagner: Is the concern that the execution of `update()` might get interrupted in order to do some other work? (Is it an RTOS Task?) If so, presumably the same might happen for the component's `loop()`?

descipher: Yep, data PIT would be good however now we have the thread blocking problem, it will be well over 50ms with just 2 phases and with 12 its at least 6 * 50ms.

descipher: If we do it with a flag we no longer block because the polling interval only waits for the flag to be updated, it's even better than the original effort I came up with provided that I can solve the 0 value initial read problem which should be resolvable.

Extrapilot1: Max data rate is 1.1Mhz (900khz for calibration), so a 2Mhz data rate is likely a problem

There is a related concern- as regards the max window size on the peak and sag detector registers. They rollover at a max of 256ms, where the value for that window is available only during the following window. If you want to be able to actually monitor peaks, which tend to be sub-50ms events, there is a need to poll those registers at a high rate, where others can comfortably be at 60 seconds, or minutes, depending on which. Ideally, this could be configured per entity vs per chip, so that the SPI bus doesn’t become limiting, and the CPU isnt dealing with a bunch of data no one needs/wants.

descipher: Thanks for catching that, will adjust the rate to 1Mhz. We cannot reach full throttle therefor the hardware interval of 900khz would not be reached under the current safe delay read constraint.

descipher: @jesserockz bool flag method works well on my local testing rigs, thank-you. By setting it to true on init it resolves the 0 value publishing issue on startup. Just need some additional users to validate the change.

CircuitSetup (approved):
looks good!

descipher:
Adding 3 additional capabilities of the ATM90E32 chip, phase angle,  harmonic power and peak current. Docs to follow, will post the completion event with PR details in the next few days.

CircuitSetup:
@descipher thanks for all your hard work with this! 

@jesserockz does anything else need to change with this for it to be merged?

descipher:
> @descipher thanks for all your hard work with this!
> 
> @jesserockz does anything else need to change with this for it to be merged?

I will add the docs shortly lots going on at the moment, this part is good to go and can be committed with or without the added docs for the features. There will also be a follow on PR part which deals with selectable higher polling data functions of the advanced peak current and other data output.

descipher:
@jesserockz This is ready for a final QA, there is more work coming to address a specific need to poll peak_current at a higher rate than the interval. When I have time I will make it an option via a number comp with a publish timeout so it does not bloat an upstream HA DB with excess data.

jesserockz (dismissed):

esphome/components/atm90e32/atm90e32.h
```diff
@@ -20,71 +25,102 @@ class ATM90E32Component : public PollingComponent,
...
     uint32_t cumulative_reverse_active_energy_{0};
   } phase_[3];
+
   sensor::Sensor *freq_sensor_{nullptr};
   sensor::Sensor *chip_temperature_sensor_{nullptr};
   uint16_t pga_gain_{0x15};
   int line_freq_{60};
   int current_phases_{3};
+  bool publish_interval_flag_{true};
+  bool peak_current_signed_{true};
```
jesserockz: This initialised value should match the default in config, otherwise its kinda confusing.

descipher: Corrected.

jesserockz (approved):

PR #5673, opened 2023-11-04
Remove page jump on Nextion startup

edwardtfn:

Nextion loads page 0 by default on power on, so there's no need to call that page on setup. As users could define the initial page to load on the boot sequence of the TFT file, this will conflict with this command and therefore I'm suggesting to replace it by a optional parameter to be set to Nextion component and, if not set, no page jump will occur.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3336

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:


kbx81:
I think the original motivation for this was so that the display is on a known page at boot. As you mentioned, yes, the display will land on page 0 when it boots, but it's possible that the ESP resets while the Nextion does not. This would result in the display being on an indeterminate page when the ESP starts which could be bad depending on how the Nextion's code is designed.

Rather than simply remove this, I think it would be better to expose it as a configuration variable so that the user can decide on a page to set at boot. If the user doesn't define this, then we can simply leave it unchanged.

edwardtfn:
I see you point, although this could be achieved with a call to `goto_page("0")` on the `on_setup` of the Nextion component, which would make more sense, in my opinion. So if you need to ensure a specific page this could be the best anyways.
But I will take a look on the idea of having as a configurable variable. That could even be implemented to support the page name in addition to the page I'd, which would be nice.

edwardtfn:
Done!
I've added support only to "page id" (and not to "page name") to keep consistency with `wake_up_page`.
@kbx81, please let me know your thoughts.

jesserockz (approved):
LGTM. WIll leave the final review and merge up to @kbx81

kbx81 (approved):
Double approval. 😄 LGTM, also -- thanks for working on this!

PR #5677, opened 2023-11-05
Fix compile with latest esp-idf on esp32c6

DAVe3283:

Trying to compile with the latest esp-idf for an esp32c6 (specifically the OLIMEX ESP32-C6-EVB) with logging set to VERBOSE or VERY_VERBOSE resulted in a number of components failing to compile.

This is mostly due to incorrect printf modifiers for \*int32_t types.
#4671 addressed some of the format strings but only for default logging.
This PR addresses as many as I could get to compile on an esp32c6, which is hopefully most of them now.

Tested on ESP32-C6 (esp-idf), ESP32 (arduino), and an ESP8266 (arduino).

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: esp32c6
  friendly_name: esp32c6
  # OLIMEX ESP32-C6-EVB
  # platformio esp32-c6-devkitm-1 claims 320K RAM which appears wrong per the datasheet
  # platformio esp32-c6-devkitc-1 has RAM right, but has 8MB flash, we actually have 4MB, so fix it:
  platformio_options:
    board_build.flash_size: 4MB
    board_upload.flash_size: 4MB
    board_upload.maximum_size: 4194304

esp32:
  board: esp32-c6-devkitc-1
  variant: esp32c6
  framework:
    type: esp-idf
    version: latest
    platform_version: "https://github.com/stintel/platform-espressif32#esp32-c6-test"

logger:
  level: VERY_VERBOSE
api:
ota:
wifi:
  ssid: !secret wifi_ssid
captive_portal:

# These won't compile with VERBOSE logging prior to this PR
binary_sensor:
  # Button
  - platform: gpio
    name: BUT1
    pin:
      number: 9
      mode:
        input: true
        pullup: true
      inverted: true
light:
  - platform: status_led
    name: "Status LED"
    pin:
      inverted: true
      number: 8
```

## Checklist:

### DIFF
--- a/esphome/components/cse7761/cse7761.cpp
+++ b/esphome/components/cse7761/cse7761.cpp
@@ -217,7 +217,7 @@ void CSE7761Component::get_data_() {
     this->voltage_sensor_->publish_state(voltage);
   }
 
-  for (int channel = 0; channel < 2; channel++) {
+  for (uint8_t channel = 0; channel < 2; channel++) {
     // Active power = PowerPA * PowerPAC * 1000 / 0x80000000
     float active_power = (float) this->data_.active_power[channel] / this->coefficient_by_unit_(POWER_PAC);  // W
     float amps = (float) this->data_.current_rms[channel] / this->coefficient_by_unit_(RMS_IAC);             // A
--- a/esphome/components/cse7761/cse7761.cpp
+++ b/esphome/components/cse7761/cse7761.cpp
@@ -217,7 +217,7 @@ void CSE7761Component::get_data_() {
     this->voltage_sensor_->publish_state(voltage);
   }
 
-  for (uint32_t channel = 0; channel < 2; channel++) {
+  for (int channel = 0; channel < 2; channel++) {
     // Active power = PowerPA * PowerPAC * 1000 / 0x80000000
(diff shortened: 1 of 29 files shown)


esphome/components/cse7761/cse7761.cpp
```diff
@@ -217,7 +217,7 @@ void CSE7761Component::get_data_() {
     this->voltage_sensor_->publish_state(voltage);
   }
 
-  for (uint32_t channel = 0; channel < 2; channel++) {
+  for (int channel = 0; channel < 2; channel++) {
```
jesserockz: This is going to be 0, 1, or 2. I think it can just be `uint8_t`?
```suggestion
  for (uint8_t channel = 0; channel < 2; channel++) {
```

jesserockz (dismissed):
I hate how messy log lines have had to become because of these macros :crying_cat_face:

DAVe3283:
Yeah, it makes them a pain to read for sure, but I can't think of a better way to deal with it either.

jesserockz (approved):

PR #5681, opened 2023-11-06
Refactor dashboard zeroconf support

bdraco:

- Switch to using a ServiceBrowser to avoid flooding the network with A queries
- Use zeroconf's internal cache where possible
- This will be a bit slower to detect when a device is offline due to a hard failure / unplug because it won’t send a good bye (it will when the ttl of the record expires and the callback happens) but that seems worth it to not overload the network with multicast traffic 

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import base64
 import binascii
 import codecs
@@ -15,7 +17,6 @@ import shutil
 import subprocess
 import threading
 from pathlib import Path
-from typing import Optional
 
 import tornado
 import tornado.concurrent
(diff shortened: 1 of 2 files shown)


jesserockz (dismissed):
Changes working for me

bdraco:
I found an issue in testing. When the device is adopted it takes a while to show online because the host mapping isn't updated right away

jesserockz (approved):
Ah I didnt test adoption before, but did just now and its working after latest changes.

bdraco:
thanks

PR #5687, opened 2023-11-07
Implement a memory cache for dashboard entries to avoid frequent disk reads

bdraco:

We now `stat()` the files to be loaded and if we already have the entry with the same properties in the cache, we avoid re-creating the `DashboardEntry` object and in turn reading the json file from disk

The api would call `/devices` over and over which would hit the disk over and over.

Screenshot 2023-11-06 at 8 05 00 PM
Screenshot 2023-11-06 at 8 03 59 PM

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -154,6 +154,16 @@ class DashboardSettings:
             )
 
         entry_cache = self._entry_cache
+
+        # Remove entries that no longer exist
+        removed: list[str] = []
+        for file in entry_cache:
+            if file not in path_to_cache_key:
+                removed.append(file)
+
+        for file in removed:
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Looks good to me. Dashboard is functioning correctly still

bdraco:
Thanks. 

We could probably do better with some type of inotify watcher to avoid the stat but I figure that might not be portable enough and since this solved 99% of the disk I/o that would be diminishing returns

PR #5689, opened 2023-11-07
Add gradient option to addressable color wipe effect

lifeisafractal:

This PR adds a new option to the addressable color wipe effect to enable gradient transitions between colors.

This also addresses the code/doc discrepancy of `num_leds` being required/options respectively. The approach taken is to honor the docs and allow it to be optional (see docs PR for details). The only catch here is that I couldn't figure out how to get voluptuous to statefully reference previous entries in a list as a default value, so my approach was to tackle this during code generation which I know isn't the preferred method.

For the new `gradient` option, I used the same approach above. Maybe a better approach for both is explicit rather than implicit, and make `num_leds` required, and have `gradient` optional, but default to `false`. I'm looking for opinions and input on this.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3339

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
light:
  - platform: neopixelbus
    type: RGB
    pin: D2
    num_leds: 50
    method: BIT_BANG
    variant: ws2811
    id: lights
    name: "Holiday Lights"
    effects:
      - addressable_color_wipe:
          name: thanksgiving
          add_led_interval: 30ms
          reverse: false
          colors:
            - # yellow
              red: 100%
              green: 91%
              blue: 0%
              num_leds: 10
              gradient: true
            - # red
              red: 100%
              green: 6%
              blue: 18%
            - # orange solid
              red: 100%
              green: 63%
              blue: 0%
              gradient: false
              num_leds: 100
            - # orange
              red: 100%
              green: 63%
              blue: 0%
              gradient: true
              num_leds: 10
```

## Checklist:

### DIFF
--- a/esphome/components/light/addressable_light_effect.h
+++ b/esphome/components/light/addressable_light_effect.h
@@ -152,7 +152,6 @@ class AddressableColorWipeEffect : public AddressableLightEffect {
   uint32_t add_led_interval_{};
   size_t leds_added_{0};
   bool reverse_{};
-  bool gradient_{};
 };
 
 class AddressableScanEffect : public AddressableLightEffect {
--- a/esphome/components/light/addressable_light_effect.h
+++ b/esphome/components/light/addressable_light_effect.h
@@ -108,6 +108,7 @@ class AddressableColorWipeEffect : public AddressableLightEffect {
   void set_colors(const std::vector<AddressableColorWipeEffectColor> &colors) { this->colors_ = colors; }
   void set_add_led_interval(uint32_t add_led_interval) { this->add_led_interval_ = add_led_interval; }
   void set_reverse(bool reverse) { this->reverse_ = reverse; }
+  void set_gradient(bool gradient) { this->gradient_ = gradient; }
   void apply(AddressableLight &it, const Color &current_color) override {
     const uint32_t now = millis();
     if (now - this->last_add_ < this->add_led_interval_)
@@ -118,7 +119,14 @@ class AddressableColorWipeEffect : public AddressableLightEffect {
     else
       it.shift_right(1);
     const AddressableColorWipeEffectColor color = this->colors_[this->at_color_];
-    const Color esp_color = Color(color.r, color.g, color.b, color.w);
+    Color esp_color = Color(color.r, color.g, color.b, color.w);
+    if(this->gradient_) {
+      size_t next_color_index = (this->at_color_ + 1) % this->colors_.size();
+      const AddressableColorWipeEffectColor next_color = this->colors_[next_color_index];
+      const Color next_esp_color = Color(next_color.r, next_color.g, next_color.b, next_color.w);
+      uint8_t gradient = 255 * ((float)this->leds_added_ / color.num_leds);
+      esp_color = esp_color.gradient(next_esp_color, gradient);
+    }
     if (this->reverse_)
       it[-1] = esp_color;
     else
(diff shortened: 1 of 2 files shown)


lifeisafractal:
It's been a bit with no review. Do I need to do another step I missed? Happy to help out to get this feature in; I think it's really neat!

esphome/components/light/effects.py
```diff
@@ -385,7 +386,8 @@ async def addressable_rainbow_effect_to_code(config, effect_id):
                 cv.Optional(CONF_BLUE, default=1.0): cv.percentage,
                 cv.Optional(CONF_WHITE, default=1.0): cv.percentage,
                 cv.Optional(CONF_RANDOM, default=False): cv.boolean,
-                cv.Required(CONF_NUM_LEDS): cv.All(cv.uint32_t, cv.Range(min=1)),
+                cv.Optional(CONF_NUM_LEDS): cv.All(cv.uint32_t, cv.Range(min=1)),
+                cv.Optional(CONF_GRADIENT): cv.boolean,
```
kroimon: Please add default values for the optional settings here for use during validation.

esphome/components/light/effects.py
```diff
@@ -399,7 +401,11 @@ async def addressable_color_wipe_effect_to_code(config, effect_id):
     cg.add(var.set_add_led_interval(config[CONF_ADD_LED_INTERVAL]))
     cg.add(var.set_reverse(config[CONF_REVERSE]))
     colors = []
+    num_leds = 1
+    gradient = False
```
kroimon: Do not add default values during code generation.

esphome/components/light/effects.py
```diff
@@ -408,7 +414,8 @@ async def addressable_color_wipe_effect_to_code(config, effect_id):
                 ("b", int(round(color[CONF_BLUE] * 255))),
                 ("w", int(round(color[CONF_WHITE] * 255))),
                 ("random", color[CONF_RANDOM]),
-                ("num_leds", color[CONF_NUM_LEDS]),
+                ("num_leds", num_leds),
+                ("gradient", gradient),
```
kroimon: After adding defaults to the validation, remove the `num_leds` and `gradient` variables above and just use:
```suggestion
                ("num_leds", config[CONF_NUM_LEDS]),
                ("gradient", config[CONF_GRADIENT]),
```

esphome/components/light/addressable_light_effect.h
```diff
@@ -144,6 +152,7 @@ class AddressableColorWipeEffect : public AddressableLightEffect {
   uint32_t add_led_interval_{};
   size_t leds_added_{0};
   bool reverse_{};
+  bool gradient_{};
```
kroimon: Unused?

kroimon (changes requested):

esphome/components/light/effects.py
lifeisafractal: done.

esphome/components/light/effects.py
lifeisafractal: fixed.

esphome/components/light/effects.py
lifeisafractal: fixed.

esphome/components/light/addressable_light_effect.h
lifeisafractal: correct, left from earlier implementation when trying things out. Removed.

lifeisafractal:
All PR feedback addressed. I went back to the approach of defaults in the validator as is the custom here. The documentation PR has been updated to reflect this as well.

kroimon (approved):

jesserockz (approved):

PR #5616, opened 2023-10-28
feat: Add ESP32 BLE enable/disable automations

Rapsssito:
# What does this implement/fix?

Adds `ble.enable` and `ble.disable` actions like the existing `wifi.enable` and `wifi.disable` (with the corresponding `ble.enabled` condition). Along with this, it adds the parameter `enable_on_boot` to the `esp32_ble` component to select if the service should start at boot or not.

Currently, the BLE controller, once it is setup, keeps running forever. This is far from desirable in projects with very high energy limitations. For example, `esp32_improv` turns on the BLE controller for a service that becomes useless once the device has connected to a WiFi network. The `ble.enable` and `ble.disable` actions will empower users to shutdown the whole BLE controller after it is not needed.

Updated components dependent on `esp32_ble`:
- [x] `esp32_improv` stack (i.e. including `esp32_ble_server`)
- [x] `esp32_ble_tracker`

_Note: I have noticed that "esp32_ble_beacon" shares a lot of code with "esp32_ble". It would be ideal if both components shared the same codebase so "esp32_ble_beacon" would also get these new features. An ideal solution will be creating a single BLE controller that abstracts all the BLE functionality from the rest of the components. That will also simplify the implementation of NimBLE (https://github.com/esphome/feature-requests/issues/810)_

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2418

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3303

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Tested in a ESP32-C3-WROOM-02 chip.

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
esphome:
  name: test-esp-enable
  friendly_name: test-esp-enable
  platformio_options:
    board_build.flash_mode: dio # Required for ESP-IDF framework
  on_boot:
    - priority: 260
      then:
        # Wait 5s before enabling BLE
        - delay: 5s
        - ble.enable:
        # Disable BLE to enable it again, just to test everything can ramp up again correctly
        - delay: 5s
[...]

PR #5690, opened 2023-11-07
Actions to enable and disable WireGuard connection

droscy:

This PR add actions to enable and disable the WireGuard component and conditions to check if the vpn link is enabled and active.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** this PR should also address a feature requested here https://github.com/esphome/feature-requests/issues/1444#issuecomment-1621489477 by @snechiporenko

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3342

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# configure wireguard as usual
wireguard:
  [...]

button:                                                                                                                                                                                                                                       
  - platform: template                                                                                                                                                                                                                        
    name: 'Toggle WireGuard'                                                                                                                                                                                                                  
    entity_category: config                                                                                                                                                                                                                   
    on_press:                                                                                                                                                                                                                                 
      - if:                                                                                                                                                                                                                                   
          condition: wireguard.enabled                                                                                                                                                                                                        
          then:                                                                                                                                                                                                                               
            - wireguard.disable:                                                                                                                                                                                                              
          else:                                                                                                                                                                                                                               
            - wireguard.enable:                                                                                                                                                                                                               
                                                                                                                                                                                                                                              
  - platform: template                                                                                                                                                                                                                        
    name: 'Log WireGuard status'                                                                                                                                                                                                              
    entity_category: config                                                                                                                                                                                                                   
    on_press:                                                                                                                                                                                                                                 
      - if:                                                                                                                                                                                                                                   
          condition: wireguard.peer_online                                                                                                                                                                                                    
          then:                                                                                                                                                                                                                               
            - logger.log: 'wireguard remote peer is online'                                                                                                                                                                                   
          else:                                                                                                                                                                                                                               
            - logger.log: 'wireguard remote peer is offline'
```

## Checklist:


esphome/components/wireguard/wireguard.cpp
```diff
@@ -199,7 +209,35 @@ void Wireguard::set_address_sensor(text_sensor::TextSensor *sensor) { this->addr
...
+  this->enabled_ = true;
+  ESP_LOGI(TAG, "WireGuard enabled");
+  this->publish_enabled_state();
+}
+
+void Wireguard::disable() {
+  this->enabled_ = false;
+  this->stop_connection_();
+  ESP_LOGI(TAG, "WireGuard disabled");
+  this->publish_enabled_state();
```
jesserockz: Can you do this in a further loop outside of the action itself?

See https://github.com/esphome/esphome/pull/5616#pullrequestreview-1714120196 for details on how I changed for ble (the same needs to be done for wifi too)

jesserockz (changes requested):

esphome/components/wireguard/wireguard.cpp
droscy: Can I simply replace `this->stop_connection_();` with the following?

```c++
this->defer(std::bind(&Wireguard::stop_connection_, this));
```

jesserockz: If it works correctly, yeah probably

droscy: I've just pushed the requested change.

esphome/components/wireguard/wireguard.cpp
```diff
@@ -301,6 +339,8 @@ void resume_wdt() {
 
 std::string mask_key(const std::string &key) { return (key.substr(0, 5) + "[...]="); }
 
+Wireguard *global_wireguard;  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
```
jesserockz: ~~Is this really needed?~~

Ah lol, I collapsed the .h file and didnt notice its use.

jesserockz: But in saying that, we try to only add globals like this when there are cross component calls, which there are not here. Wi-Fi and BLE have it because there is other code that calls them. 
For the actions inside WireGuard, the config schema should use GenerateID to pass a reference of the wireguard component to the actions and add `Parented<WireGuard>` so they can just use `this->parent_`

jesserockz (changes requested):

esphome/components/wireguard/wireguard.cpp
droscy: Ok, I've switched from global pointer to local parent, removing the global pointer.

jesserockz (approved):

PR #5694, opened 2023-11-08
Add support for fan preset modes

### LINKED ISSUE #1661: Add support for fan preset modes

add a configuration option to allow the use of defined preset modes for fans that support them.

adding preset modes to ESP based fans would allow more flexibility than just supporting percentage speeds.

**Additional context**

mill1000:

Adds preset mode support to fan components.  Requires  https://github.com/esphome/aioesphomeapi/pull/616 and for Home Assistant users: https://github.com/home-assistant/core/pull/103781

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/1661

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3362

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# For testing speed fan presets
fan:
  - platform: speed
    name: "Test Fan"
    id: test_fan
    output: gpio_19
    oscillation_output: gpio_18
    direction_output: gpio_17
    preset_modes:
      - Auto
      - Sleep
      - Some Other Preset
    on_preset_set:
      then:
        lambda: |-
          if (id(test_fan).preset_mode == "Auto") {
            // Auto enabled, do automatic things
          }
```

## Checklist:


mill1000:
Embarrassed that I keep jumping the gun on my pushes and not passing CI...

bdraco:
I've only had time to do a cursory look through but I wanted to make sure to point out the below:

https://developers.home-assistant.io/docs/core/entity/fan?_highlight=fan#preset-modes

> Preset modes should not include named (manual) speed settings as these should be represented as percentages.

> Manually setting a speed must disable any set preset mode. If it is possible to set a percentage speed manually without disabling the preset mode, create a switch or service to represent the mode.


You may have already implemented it this way, but I wanted to make sure I mentioned it now since it may save some time later

mill1000:
> https://developers.home-assistant.io/docs/core/entity/fan?_highlight=fan#preset-modes
> 
> > Preset modes should not include named (manual) speed settings as these should be represented as percentages.

Hm...not sure how to feel about this one. Was the intent to prohibit presets that affect speed entirely, or to prevent an integration from using presets as the sole means to control speed. e.g. Is it unreasonable to want 1 % granularity but have presets for specific speeds (i.e. Low, Med, High)

> > Manually setting a speed must disable any set preset mode. If it is possible to set a percentage speed manually without disabling the preset mode, create a switch or service to represent the mode.
> 
> You may have already implemented it this way, but I wanted to make sure I mentioned it now since it may save some time later

As currently implemented (in speed fan) this is true. Any command to change the speed, direction or oscillation will disable the preset.

mill1000:
Poked around with adding preset support to H-Bridge fan as well, but holding off on any continued work until it's decided whether preset modes with speeds are not allowed.

esphome/components/copy/fan/copy_fan.cpp
```diff
@@ -46,6 +49,8 @@ void CopyFan::control(const fan::FanCall &call) {
     call2.set_speed(*call.get_speed());
   if (call.get_direction().has_value())
     call2.set_direction(*call.get_direction());
+  if (!call.get_preset_mode().empty())
+    call2.set_preset_mode(call.get_preset_mode());
```
mill1000, line 53: I think this logic is incorrect. Copy fan should probably always pass preset mode, even if empty.

bdraco:
> Hm...not sure how to feel about this one. Was the intent to prohibit presets that affect speed entirely, or to prevent an integration from using presets as the sole means to control speed. e.g. Is it unreasonable to want 1 % granularity but have presets for specific speeds (i.e. Low, Med, High)

The intent was to make sure that presets are not simple speeds like `Low` `Low Medium` `Med` `Medium High` `High` etc as these should be handled by percentage.  The goal is to model a fan so that when you ask for 20% between different types of fans you get the nearest to 20% as possible. We also want to be able to control the fan in languages other than english.

Preset modes should be something like `Auto`, `Breeze`, `Cycle`. These may control the speed but it should not be a 1:1 mapping to a specific speed.  

Preset modes are generally opaque to anything outside the Home Assistant UI since they generally aren't controllable by voice or translatable.

esphome/components/copy/fan/copy_fan.cpp
mill1000: On second thought this is OK as in. `call2` will default to an empty `preset_mode` so there's no need to copy an empty `preset_mode` from `call` to `call2`

mill1000:
So what do you feel is the path forward here? Does allowing control of direction and oscillation ensure that "presets are not simple speeds", or should I rip out the speed fan implementation?

bdraco:
> So what do you feel is the path forward here? Does allowing control of direction and oscillation ensure that "presets are not simple speeds", or should I rip out the speed fan implementation?

I think its ok. ESPHome fans are going to be mostly DIY, and I don't think it makes it any easier for some vendor to build an ESPHome fan and only have `low`,`medium`,`high` presets modes instead of speeds which is what we really want to avoid.

mill1000:
Cool. I'll continue with the h-bridge work and additional test cases.

nagyrobi:
@bdraco my use case: my fans can be controlled 0-100% to any value. However, having to drag a small slider is unconfortable so simply having additional presets with say 33%, 66%, and 100% as shortcuts (names could vary) would improve the user experience of the family nembers. 

@mill1000 Are the values templatable?

Templating could be handy to have the ability to change the values at runtime. Eg have a template number component to specify the actual speed of `Breeze` with a lambda (thus from the ui), rather than having to reflash firmware.

bdraco:
> @bdraco my use case: my fans can be controlled 0-100% to any value. However, having to drag a small slider is unconfortable so simply having additional presets with say 33%, 66%, and 100% as shortcuts (names could vary) would improve the user experience of the family nembers.

I get what you are saying, but that is a frontend problem. We shouldn't let frontend design dictate how the backend works, we should improve the frontend instead.

nagyrobi:
Why a frontend problem? We would want to use both: presets for quick speed access, and the slider too, for special occasions which need intermediate levels.

We only have the slider now, and can't have the presets.

Or we could have the slider "degraded" to only support a few levels instead of 100 - which, honestly, is quite stupid...

bdraco:
> Why a frontend problem? We would want to use both: presets for quick speed access, and the slider too, for special occasions which need intermediate levels.

We shouldn't add multiple ways to do the same thing on the backend as we don't want to have to maintain multiple systems to do the same thing.

There are already frontend options to set a preset speed, one of them is a button card:

```yaml
show_name: true
show_icon: true
type: button
entity: fan.alexander_master_bedroom_2
tap_action:
  action: call-service
  service: fan.set_percentage
  target:
    entity_id: fan.alexander_master_bedroom_2
  data:
    percentage: 18
icon: mdi:fan-speed-2
```

nagyrobi:
@bdraco the only issue with this approach is that it only considers Home Assistant's UI as the single usage possibility with ESPHome. There are also other open source user interfaces which rely on HA core, but don't need a PC/browser to control it (see https://www.openhasp.com for example). Also ESPHome is supported (even at api level) by other open source systems, like iobroker for example, which is a good thing.

bdraco:
> @bdraco the only issue with this approach is that it only considers Home Assistant's UI as the single usage possibility with ESPHome. There are also other open source user interfaces which rely on HA core, but don't need a PC/browser to control it (see https://www.openhasp.com for example). Also ESPHome is supported (even at api level) by other open source systems, like iobroker for example, which is a good thing.

This PR is proposing to add support to Home Assistant so I'm reviewing it in that context.

mill1000:
Ok. I've tested the preset mode on hbridge, speed and copy fan. At this time I'm satisfied with the function and have marked this as ready for review

esphome/components/fan/fan.cpp
```diff
@@ -143,20 +173,36 @@ void Fan::save_state_() {
   state.oscillating = this->oscillating;
   state.speed = this->speed;
   state.direction = this->direction;
+
+  if (this->get_traits().supports_preset_modes() && !this->preset_mode.empty()) {
+    const auto &preset_modes = this->get_traits().supported_preset_modes();
+    // Store index of current preset mode
+    auto it = std::find(preset_modes.begin(), preset_modes.end(), this->preset_mode);
+    if (it != preset_modes.end())
```
bdraco: Please name variables to show intent

bdraco:
The one big thing that looks missing is that when the user manually sets a speed, it should override and clear the preset mode

esphome/components/speed/fan/speed_fan.cpp
```diff
@@ -27,9 +36,18 @@ void SpeedFan::control(const fan::FanCall &call) {
   if (call.get_direction().has_value())
     this->direction = *call.get_direction();
 
+  if (!ignore_preset) {
```
mill1000: I don't love this logic and I've worked up a different implementation, but I still need to test it

mill1000:
> The one big thing that looks missing is that when the user manually sets a speed, it should override and clear the preset mode

In my testing this has worked as you've described. From HA, any manual change to speed, direction or oscillation causes a call to be made with an empty `preset_mode` which clears the preset.

bdraco:
> > The one big thing that looks missing is that when the user manually sets a speed, it should override and clear the preset mode
> 
> In my testing this has worked as you've described. From HA, any manual change to speed, direction or oscillation causes a call to be made with an empty `preset_mode` which clears the preset.

👍  I was looking for the code here, and didn't think about HA will send it empty.

I don't yet have a fan running ESPHome unfortunately (I ordered some equipment for testing)

mill1000:
@nagyrobi 

> Are the values templatable?
> 
> Templating could be handy to have the ability to change the values at runtime. Eg have a template number component to specify the actual speed of Breeze with a lambda (thus from the ui), rather than having to reflash firmware.

No they aren't templatable. 

I looked into this briefly, and I don't _think_ there's a technical limitation to templating them but I wasn't able to find another component that used templates outside of actions (except of course template components themselves).

esphome/components/speed/fan/speed_fan.cpp
mill1000: How's the name `preset_iterator`. Wasn't sure what else to call it.

bdraco:
Test hardware arrived while I was traveling. I'll try to get something built to test with next week after the holiday

bdraco:
I didn't get a chance to test this out yet and I'm traveling again tomorrow but I'm going to throw the test hardware in my travel bag so I'll be able to test it on the road

bdraco:
Tested speed fan and everything is working as expected AFAICT

mill1000:
Sweet!

esphome/components/hbridge/fan/__init__.py
```diff
@@ -28,6 +37,19 @@
...
 
+_PRESET_MODES_SCHEMA = cv.All(
+    cv.ensure_list(
+        cv.All(
+            {
+                cv.Required(CONF_NAME): cv.string_strict,
+                cv.Optional(CONF_SPEED): cv.int_range(1),
+                cv.Optional(CONF_DIRECTION): cv.enum(FAN_DIRECTION_ENUM, upper=True),
+            },
+            cv.has_at_least_one_key(CONF_SPEED, CONF_DIRECTION),
```
filmkorn: To implement an `auto` mode that controls the fan speed (and perhaps direction) based on other factors (ie air pollution or temperature), it seems obsolete to set at least speed and direction. In this case I'd want the preset name to be nothing but an enum that I can use as input and trigger for an automation.

bdraco: It seems like driving the preset mode via an automation might be a better design where the preset mode effectively turns on an automation and setting manual speed or changing presets will turn it off

mill1000: Either way for that to be possible we'd need a new automation trigger, something like `on_preset_set`

bdraco: That makes sense.  I really like the idea that the preset mode drives an automation as it's much more flexible and inline with what we think of as being a preset mode

mill1000: Alright I'll take a swing at it

mill1000: Added a new `on_preset_set` trigger

```yaml
    on_preset_set:
      then:
        - logger.log: "A preset was set"
````

**Comment by mill1000** [comment]
Updated the docs PR to include the `on_preset_set` trigger and added some info on how to programmatically access and change the preset mode.

**Comment by nagyrobi** [comment]
Within the automation, you detect which preset was set by name? Name can be any string (spaces, diacritics, etc)?

**Comment by mill1000** [comment]
Yes by name, and AFAIK any string.

File: esphome/components/speed/fan/__init__.py [hunk 5]
```diff
@@ -1,18 +1,40 @@
...
+    cv.ensure_list(
+        {
+            cv.Required(CONF_NAME): cv.string_strict,
+            cv.Optional(CONF_SPEED): cv.int_range(1),
+            cv.Optional(CONF_OSCILLATING): cv.boolean,
+            cv.Optional(CONF_DIRECTION): cv.enum(FAN_DIRECTION_ENUM, upper=True),
+        }
+    ),
+    cv.Length(min=1),
+)
```
bdraco: Since preset modes are an advanced feature, I think it would be better to not have to support both manually configured and the automation.  They can do all of this in the automation anyways.

mill1000: Gotcha. Just pushed an automation-only approach.

Presets are now defined as a list of names
```yaml
preset_modes:
  - Auto
  - Sleep
```

mill1000:
@bdraco Anything else you need from me to move this along?

bdraco:
Nope, at this point I'm waiting for another dev to take a look

jesserockz (approved):
Code looks fine, but what do presets actually do on a fan? Doesn't seem like there is any functional change?

mill1000:
> but what do presets actually do on a fan? Doesn't seem like there is any functional change?

For speed fan and h-bridge they do nothing without configuring a matching `on_preset_set` automation.

Other platforms might have predefined presets to enable device specific modes. (e.g. an automatic or sleep mode).

PR #5697, opened 2023-11-08
Generate partitions.csv based on flash size

kbx81:

This change will generate `partitions.csv` based on `flash_size` in `esp32` config.

The default is `4MB` to maintain compatibility; other supported values are `8MB`, `16MB` and `32MB`.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3345

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esp32:
  board: esp32dev
  flash_size: 16MB
```

## Checklist:

### DIFF
--- a/esphome/components/esp32/__init__.py
+++ b/esphome/components/esp32/__init__.py
@@ -386,10 +386,21 @@ FRAMEWORK_SCHEMA = cv.typed_schema(
 )
 
 
+FLASH_SIZES = [
+    "4MB",
+    "8MB",
+    "16MB",
+    "32MB",
+]
+
+CONF_FLASH_SIZE = "flash_size"
 CONFIG_SCHEMA = cv.All(
     cv.Schema(
         {
             cv.Required(CONF_BOARD): cv.string_strict,
+            cv.Optional(CONF_FLASH_SIZE, default="4MB"): cv.one_of(
+                *FLASH_SIZES, upper=True
+            ),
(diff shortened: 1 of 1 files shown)


balloob:
Can we also make this a value that can be passed to the wizard, and if so, it's added to the YAML ? It would also be nice if we could set this when adopting a device.

randybb:
or autodiscovered after first flashing

jesserockz:
Adopting a device probably doesn't need to know about this as the pre-flashed firmware should have the correct setting for the device.

balloob:
Empty ESPHome device firmware flashed through esphome.io/projects or web.esphome.io would need to know the flash size.

jesserockz:
For those we can probably create a firmware file per flash size and esptool.js can get the size and flash the correct file in the first place?
Could be a new feature of the manifest to add an optional `flash_size` to each `build` to match against, just like the `chipFamily` is matched?

balloob:
Yeah it's been a feature requested before for ESP Web Tools. Looks like we can move that forward.

jesserockz (approved):

PR #5714, opened 2023-11-09
2023.11.0b3

jesserockz:
**Do not merge, release script will automatically merge**
- Bump aioesphomeapi to 18.2.7 [esphome#5706](https://github.com/esphome/esphome/pull/5706)
- Update Dockerfile to use piwheels for armv7 [esphome#5709](https://github.com/esphome/esphome/pull/5709)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,6 +5,7 @@
 # One of "docker", "hassio"
 ARG BASEIMGTYPE=docker
 
+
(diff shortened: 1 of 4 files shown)


bdraco:
https://github.com/esphome/esphome/pull/5715 should be good to pull in if CI passes (I'm pretty sure it will)

Note that it still has to build zeroconf on arm7 because piwheels doesn't have the wheel yet

PR #5720, opened 2023-11-09
Speed up OTAs

bdraco:

The upload buffer was limited to 1024 which limited the speed of the OTA. By increasing the buffer size to 8192, the OTA can happen around 20-30% faster for most cases (YMMV with network performance and how busy the ESP is)

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/espota2.py
+++ b/esphome/espota2.py
@@ -40,6 +40,10 @@ MAGIC_BYTES = [0x6C, 0x26, 0xF7, 0x5C, 0x45]
 
 FEATURE_SUPPORTS_COMPRESSION = 0x01
 
+
+UPLOAD_BLOCK_SIZE = 8192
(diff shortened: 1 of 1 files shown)


balloob (approved):
What happens if the ESP is unable to keep up?

bdraco:
> What happens if the ESP is unable to keep up? 

Same as before, The socket with get back pressure and the tcp stack will wait to send.  Right now the buffer can't be filled in time because user space can't get the data into kernel space fast enough because it's so small that the python code/user space switching is too slow

bdraco:
thanks

PR #5724, opened 2023-11-10
dashboard: Use mdns cache when available if device connection is OTA

bdraco:

Since we already have a service browser running, we likely already know the IP of the device we want to connect to so we can replace `--device OTA` with the address if its in the cache (ie seen in the last 120 seconds aka host ttl) to avoid the esphome app having to look it up again (and multicast another query to the network)

To accomplish this, we need to migrate the tornado websocket handler to use coroutines since tornado already supports them natively. The way we are currently starting the webserver is deprecated and going away so we will also migrate partially to `asyncio` with some wrappers for the un-migrated parts https://www.tornadoweb.org/en/stable/guide/structure.html#the-main-coroutine

Before this change zeroconf is starting its own event loop in a thread, after the change, a single event loop is used.
Screenshot 2023-11-12 at 6 08 06 PM
Screenshot 2023-11-12 at 6 07 58 PM

In a future pr we can avoid spawning another process to run the logs which will further reduce the number of zeroconf instances running at the same time

This should help reduce cases where the system runs out of buffer space aka https://github.com/esphome/issues/issues/2669 https://github.com/esphome/issues/issues/2534

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/dashboard/async_adapter.py
+++ b/esphome/dashboard/async_adapter.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+import asyncio
+import threading
+
+
+class ThreadedAsyncEvent:
+    """This is a shim to allow the asyncio event to be used in a threaded context.
+
+    When more of the code is moved to asyncio, this can be removed.
+    """
+
+    def __init__(self) -> None:
+        """Initialize the ThreadedAsyncEvent."""
+        self.event = threading.Event()
+        self.async_event: asyncio.Event | None = None
+        self.loop: asyncio.AbstractEventLoop | None = None
+
+    def async_setup(
+        self, loop: asyncio.AbstractEventLoop, async_event: asyncio.Event
+    ) -> None:
+        """Set the asyncio.Event instance."""
+        self.loop = loop
+        self.async_event = async_event
+
+    def async_set(self) -> None:
+        """Set the asyncio.Event instance."""
+        self.async_event.set()
+        self.event.set()
+
+    def set(self) -> None:
+        """Set the event."""
+        self.loop.call_soon_threadsafe(self.async_event.set)
+        self.event.set()
+
+    def wait(self) -> None:
+        """Wait for the event."""
+        self.event.wait()
+
+    async def async_wait(self) -> None:
+        """Wait the event async."""
+        await self.async_event.wait()
+
+    def clear(self) -> None:
+        """Clear the event."""
+        self.loop.call_soon_threadsafe(self.async_event.clear)
+        self.event.clear()
+
+    def async_clear(self) -> None:
+        """Clear the event async."""
+        self.async_event.clear()
+        self.event.clear()
+
+    def is_set(self) -> bool:
+        """Return if the event is set."""
+        return self.event.is_set()
--- a/esphome/dashboard/dashboard.py
+++ b/esphome/dashboard/dashboard.py
@@ -1001,8 +1001,6 @@ class MDNSStatus:
     async def async_resolve_host(self, host_name: str) -> str | None:
         """Resolve a host name to an address in a thread-safe manner."""
         if aiozc := self.aiozc:
-            # Currently we do not do any I/O and only
-            # return the cached result (timeout=0)
             return await aiozc.async_resolve_host(host_name)
         return None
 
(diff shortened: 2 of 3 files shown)


esphome/dashboard/dashboard.py
```diff
@@ -963,24 +977,41 @@ def get(self, platform: str):
...
+        """Resolve a host name to an address in a thread-safe manner."""
+        if aiozc := self.aiozc:
+            # Currently we do not do any I/O and only
+            # return the cached result (timeout=0)
+            return await aiozc.async_resolve_host(host_name)
+        return None
+
     def _refresh_hosts(self):
         """Refresh the hosts to track."""
         entries = _list_dashboard_entries()
```
bdraco: This should probably run in the executor since the first iteration will do some disk IO

bdraco:
Unrelated to this PR, but I just realized zeroconf is processing queries when the service registry is empty.

This is a use case that was never considered because I'm so used to working in HA core which will always have services in the registry.

Its probably using 2-5x as much cpu time as needed for when there are no answers in the registry if we can short the answer query path.

I will have to solve that in zeroconf: https://github.com/python-zeroconf/python-zeroconf/pull/1307

bdraco:
I can't believe I missed this use case for so many years especially considering how much it cut the run time 🤦 
https://github.com/python-zeroconf/python-zeroconf/pull/1307 releasing now.. will bump in a separate PR

bdraco:
Existing problem that I did not fix, but should be possible now that we can run jobs in the executor in a future PR: the `get_serial_ports` helper is blocking the tornado event loop
 
Screenshot 2023-11-12 at 6 10 19 PM

esphome/dashboard/async_adapter.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import annotations
+
+import asyncio
+import threading
+
+
+class ThreadedAsyncEvent:
```
bdraco, line 7: The goal is to get rid of this but we need it for now to avoid having to get rid of all the threads in a single pr

bdraco: https://github.com/esphome/esphome/pull/5749

esphome/dashboard/dashboard.py
```diff
@@ -1211,11 +1244,26 @@ def post(self, configuration=None):
...
+        return self._mdns
+
+
 PING_RESULT: dict = {}
 IMPORT_RESULT = {}
 STOP_EVENT = threading.Event()
-PING_REQUEST = threading.Event()
+PING_REQUEST = ThreadedAsyncEvent()
 MQTT_PING_REQUEST = threading.Event()
+MDNS_CONTAINER = MDNSContainer()
```
bdraco, line 1267: The goal it to eventually hang all of this off a single `ESPHomeDashboard` object (https://github.com/esphome/esphome/pull/5749#discussion_r1391848781) but its too big of a change to do right now so I'm doing it in steps

esphome/dashboard/dashboard.py
```diff
@@ -394,32 +399,54 @@ def on_close(self):
...
+            *args,
             config_file,
             "--device",
-            json_message["port"],
+            port,
         ]
 
 
+class EsphomeLogsHandler(EsphomePortCommandWebSocket):
+    async def build_command(self, json_message: dict[str, Any]) -> list[str]:
```
balloob, line 437: It used to return the command to run, now it is actually running the command ? Shouldn't build_command be a callback, and not await?

balloob (dismissed):

esphome/dashboard/dashboard.py
bdraco: I named it poorly `run_command` should be called something better as it doesn't actually run anything just returns as before.

I'll rename it `build_device_command`

bdraco: It still need to be a coro though since we need to await the mdns resolution if the cache is stale

balloob (approved):

PR #5725, opened 2023-11-10
Fix zeroconf name resolution refactoring error

bdraco:
When I implemented using the cache to connect in the dashboard in #5724 I noticed this refactoring error (the arguments were reversed).

`HostResolver` should get the `type` as the first arg instead of the `name` since its based on https://github.com/python-zeroconf/python-zeroconf/blob/9ca9a57470b17cf683e40f4e397c7e260730545b/src/zeroconf/_services/info.py#L118

regressed in #5681

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/zeroconf.py
+++ b/esphome/zeroconf.py
@@ -147,12 +147,13 @@ class DashboardImportDiscovery:
 
 
 class EsphomeZeroconf(Zeroconf):
-    def resolve_host(self, host: str, timeout=3.0):
+    def resolve_host(self, host: str, timeout: float = 3.0) -> str | None:
         """Resolve a host name to an IP address."""
         name = host.partition(".")[0]
-        info = HostResolver(f"{name}.{ESPHOME_SERVICE_TYPE}", ESPHOME_SERVICE_TYPE)
-        if (info.load_from_cache(self) or info.request(self, timeout * 1000)) and (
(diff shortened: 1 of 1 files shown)


esphome/zeroconf.py
```diff
@@ -147,12 +147,13 @@ def update_device_mdns(self, node_name: str, version: str):
...
         """Resolve a host name to an IP address."""
         name = host.partition(".")[0]
-        info = HostResolver(f"{name}.{ESPHOME_SERVICE_TYPE}", ESPHOME_SERVICE_TYPE)
-        if (info.load_from_cache(self) or info.request(self, timeout * 1000)) and (
-            addresses := info.ip_addresses_by_version(IPVersion.V4Only)
-        ):
+        info = HostResolver(ESPHOME_SERVICE_TYPE, f"{name}.{ESPHOME_SERVICE_TYPE}")
+        if (
+            info.load_from_cache(self)
+            or (timeout and info.request(self, timeout * 1000))
```
bdraco, line 156: if timeout is 0, don't make a request

jesserockz (approved):

PR #5736, opened 2023-11-12
Bump zeroconf to 0.123.0

bdraco:

changelog: https://github.com/python-zeroconf/python-zeroconf/compare/0.122.3...0.123.0

Before zeroconf 0.123.0, zeroconf would try to answer incoming questions even if the service registry was empty which will always be the case for the esphome cli since its only listening for incoming records. This also meant that it needless decoded all the known answers in each question packet that it would never have an answer for. If the sevice registry is empty it now skips the handle query path which significantly reduces background cpu when the esphome cli is running with a zeroconf instance created.

I never thought to do this before because HA always has entries in the service registry because its advertising itself.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ esptool==4.6.2
 click==8.1.7
 esphome-dashboard==20231107.0
 aioesphomeapi==18.4.0
-zeroconf==0.122.3
+zeroconf==0.123.0
 
 # esp-idf requires this, but doesn't bundle it by default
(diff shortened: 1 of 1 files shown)


jesserockz (approved):

bdraco:
thanks .... I can't believe I missed this for ... years 🤦

PR #5739, opened 2023-11-12
[Deep Sleep] Support wake from multiple pins (ESP32C3)

Xmister:

Changes the `wakeup_pin:` structure to accept multiple pins, each with its wakeup_pin_mode on supported devices.

Right now only ESP32C3 is supported, but I plan on adding deep sleep support for bk72xx which also supports it.

The PR also consolidates unnecessarily duplicated codes on deep sleep setup.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3416

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
deep_sleep:
  run_duration: 10s
  wakeup_pin:
    - pin: 3
    - pin: 4
      wakeup_pin_mode: INVERT_WAKEUP
```

## Checklist:


jesserockz (commented):
I'm not sure it makes sense to break the config for everyone when only people using C3 that want to use multiple pins get any gains from this.

Xmister:
@jesserockz Yeah, you're right. Updated the code to be backwards compatible with existing definitions.

esphome/components/deep_sleep/__init__.py
```diff
@@ -209,10 +221,61 @@ async def to_code(config):
...
+                raise cv.Invalid("Your board only supports wake from a single pin")
+            if CONF_WAKEUP_PIN_MODE in config:
+                raise cv.Invalid(
+                    "You need to remove the global wakeup_pin_mode and define it per pin"
+                )
+            for item in conf:
+                if CONF_PIN not in item.keys():
+                    raise cv.Invalid(
+                        "pin definition is required for each item in the list of wakeup_pin"
+                    )
```
jesserockz: Validations cannot be done in `to_code` they need to be added in validators on the schema itself.

jesserockz (changes requested):

esphome/components/deep_sleep/__init__.py
```diff
@@ -107,6 +109,9 @@
 
 def validate_pin_number(value):
     valid_pins = WAKEUP_PINS.get(get_esp32_variant(), WAKEUP_PINS[VARIANT_ESP32])
+    if isinstance(value, list):
+        value = value[0]
+    value = pins.internal_gpio_input_pin_schema(value)
```
Xmister: This had to be moved here from the definitions with this check, otherwise the validator considered each `- pin` item as a list itself.

Xmister:
@jesserockz Updated validators. If you are fine with the current code, I'll do another round of testing each combination on ESP32 and ESP32C3.

Xmister:
@jesserockz Updated the validators to be a bit cleaner, also added a bunch of pytests. Please let me know if there is anything else you would like to see.

esphome/components/deep_sleep/__init__.py
```diff
@@ -167,6 +194,18 @@ def validate_config(config):
     }
 )
 
+WakeupPinItem = deep_sleep_ns.struct("WakeupPinItem")
+WAKEUP_SINGLEPIN_SCHEMA = cv.All(pins.internal_gpio_input_pin_schema)
```
jesserockz: `cv.All` is not needed when there is only one validation
```suggestion
WAKEUP_SINGLEPIN_SCHEMA = pins.internal_gpio_input_pin_schema
```

esphome/components/deep_sleep/__init__.py
```diff
@@ -167,6 +194,18 @@ def validate_config(config):
...
+WAKEUP_MULTIPIN_SCHEMA = cv.Schema(
+    {
+        cv.Required(CONF_PIN): WAKEUP_SINGLEPIN_SCHEMA,
+        cv.Optional(CONF_WAKEUP_PIN_MODE): cv.All(
+            cv.enum(WAKEUP_PIN_MODES), upper=True
+        ),
```
jesserockz: Same here
```suggestion
        cv.Optional(CONF_WAKEUP_PIN_MODE): (
            cv.enum(WAKEUP_PIN_MODES, upper=True)
        ),
```

Xmister:
@jesserockz Can you please take a look? I'd like to follow this up with deep sleep support for bk72xx, which I think many people using them would like to see.

tests/test2.yaml
```diff
@@ -73,8 +73,9 @@ deep_sleep:
     gpio_wakeup_reason: 10s
     touch_wakeup_reason: 15s
   sleep_duration: 50s
-  wakeup_pin: GPIO2
-  wakeup_pin_mode: INVERT_WAKEUP
```
jesserockz: We need to keep these in a test yaml file to ensure we are not breaking for existing users

esphome/components/deep_sleep/deep_sleep_component.h
```diff
@@ -108,11 +111,11 @@ class DeepSleepComponent : public Component {
 
   optional<uint64_t> sleep_duration_;
 #ifdef USE_ESP32
-  InternalGPIOPin *wakeup_pin_;
-  WakeupPinMode wakeup_pin_mode_{WAKEUP_PIN_MODE_IGNORE};
+  std::vector<WakeupPinItem> wakeup_pins_;
   optional<Ext1Wakeup> ext1_wakeup_;
   optional<bool> touch_wakeup_;
   optional<WakeupCauseToRunDuration> wakeup_cause_to_run_duration_;
+  bool prepare_pin_(esphome::InternalGPIOPin * /*pin*/, WakeupPinMode /*pin_mode*/);
```
jesserockz: To keep it simpler, we just give the args their names inline
```suggestion
  bool prepare_pin_(esphome::InternalGPIOPin * pin, WakeupPinMode pin_mode);
```

esphome/components/deep_sleep/__init__.py
```diff
@@ -115,10 +125,27 @@ def validate_pin_number(value):
 
 
 def validate_config(config):
-    if get_esp32_variant() == VARIANT_ESP32C3 and CONF_ESP32_EXT1_WAKEUP in config:
-        raise cv.Invalid("ESP32-C3 does not support wakeup from touch.")
-    if get_esp32_variant() == VARIANT_ESP32C3 and CONF_TOUCH_WAKEUP in config:
-        raise cv.Invalid("ESP32-C3 does not support wakeup from ext1")
+    if CORE.target_platform in (PLATFORM_ESP32):
```
jesserockz: A little simpler =)
```suggestion
    if CORE.is_esp32:
```

jesserockz (changes requested):

tests/test2.yaml
Xmister: @jesserockz test1.yaml has it

Xmister:
@jesserockz Could you take another look please?

esphome/components/deep_sleep/__init__.py
```diff
@@ -115,10 +125,27 @@ def validate_pin_number(value):
...
 def validate_config(config):
-    if get_esp32_variant() == VARIANT_ESP32C3 and CONF_ESP32_EXT1_WAKEUP in config:
-        raise cv.Invalid("ESP32-C3 does not support wakeup from touch.")
-    if get_esp32_variant() == VARIANT_ESP32C3 and CONF_TOUCH_WAKEUP in config:
-        raise cv.Invalid("ESP32-C3 does not support wakeup from ext1")
+    if CORE.is_esp32:
+        if get_esp32_variant() == VARIANT_ESP32C3 and CONF_ESP32_EXT1_WAKEUP in config:
+            raise cv.Invalid("ESP32-C3 does not support wakeup from touch.")
+        if get_esp32_variant() == VARIANT_ESP32C3 and CONF_TOUCH_WAKEUP in config:
+            raise cv.Invalid("ESP32-C3 does not support wakeup from ext1")
```
voed: Isn't this messages should be swapped? 🤔

Xmister: You are right. Although this is not new code, I'll fix it.

esphome/components/deep_sleep/__init__.py
```diff
@@ -209,10 +251,25 @@ async def to_code(config):
...
         cg.add(var.set_sleep_duration(config[CONF_SLEEP_DURATION]))
     if CONF_WAKEUP_PIN in config:
-        pin = await cg.gpio_pin_expression(config[CONF_WAKEUP_PIN])
-        cg.add(var.set_wakeup_pin(pin))
-    if CONF_WAKEUP_PIN_MODE in config:
-        cg.add(var.set_wakeup_pin_mode(config[CONF_WAKEUP_PIN_MODE]))
+        conf = config[CONF_WAKEUP_PIN]
+        for item in conf:
+            pin_expr = item[CONF_PIN] if CONF_PIN in item else item
+            wakeup_pin_mode_parent = item if CONF_PIN in item else config
```
jesserockz: See how `wifi` "moves" the ssid and password into the `networks` list via a custom validator. I think this code should be doing the same here to simplify the `to_code` function

jesserockz (changes requested):

Xmister:
@jesserockz I made the suggested changes, can you take a look please?

Xmister:
@jesserockz Just a reminder...whenever you have some time to review.

Xmister:
@jesserockz Hi, Could you take look a please? It's been a few weeks already...

Xmister:
@jesserockz Here goes my weekly reminder :)

Xmister:
@jesserockz Hi! Can we get this sorted? I think I've addressed everything, and it's been a while now...

Xmister:
@jesserockz :smiling_face_with_tear:

Xmister:
@jesserockz Please review when you got a chance

swoboda1337 (changes requested):
Hi, there are conflicts, those need to be resolved before this can be reviewed. Thanks

PR #5749, opened 2023-11-14
dashboard: convert ping thread to use asyncio

bdraco:
~~needs https://github.com/esphome/esphome/pull/5724~~

Convert ping thread to use asyncio, gets rid of the need to use multiprocessing

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/dashboard/async_adapter.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+import asyncio
+
+
+class AsyncEvent:
```
bdraco, line 6: This is an intermediate step to keep the PR smaller.

I want to create a dashboard object that holds all the state but I it was too big to not blow up the PR

esphome/dashboard/dashboard.py
```diff
@@ -1020,64 +1040,67 @@ def on_update(dat: dict[str, bool | None]) -> None:
...
+            ]
+            for ping_group in chunked(to_ping, 16):
+                ping_group = cast(list[DashboardEntry], ping_group)
+                results = await asyncio.gather(
+                    *(_async_ping_host(entry.address) for entry in ping_group),
+                    return_exceptions=True,
+                )
+                for entry, result in zip(ping_group, results):
+                    if isinstance(result, BaseException):
+                        raise result
```
balloob: Wouldn't this kill the whole ping thread and never recover. Is that what we want? Also, should we clear `PING_RESULT` when that happens ?

bdraco: This is a bug for sure since `Exception` inherits from `BaseException` 🤦 . I switched it to check `Exception` first so it works correctly now.  We only want to raise on `BaseException` so we don't block cancelation on teardown.

balloob (dismissed):

jesserockz (approved):

bdraco:
thanks

bdraco:
retesed, all good

PR #5759, opened 2023-11-15
Pass the name to the log runner when available

bdraco:
~~needs https://github.com/esphome/aioesphomeapi/releases/tag/v18.4.1 via https://github.com/esphome/esphome/pull/5760~~

When connecting by IP, by passing the name, we can reconnect quicker if the device goes offline while running logs since we can use the zeroconf logic to watch for the device to come back online

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/api/client.py
```diff
@@ -18,17 +18,17 @@
 _LOGGER = logging.getLogger(__name__)
 
 
-async def async_run_logs(config, address):
+async def async_run_logs(config: dict[str, Any], address: str) -> None:
     """Run the logs command in the event loop."""
     conf = config["api"]
+    name = config["esphome"].get("name")
```
bdraco: I wasn't sure if this could actually be missing?

bdraco: If its not possible it would be better to do

```suggestion
    name = config["esphome"]["name"]
```

jesserockz (approved):

bdraco:
Thanks

PR #5775, opened 2023-11-16
dashboard: Add support for firing events

bdraco:

Add support for firing events

We now fire the following events:

Intended to be consumed by a `websocket` version of the `devices` endpoint.  These events are for the longer term, stored state of the dashboard entry.
- EVENT_ENTRY_ADDED - A new dashboard entry has been added
- EVENT_ENTRY_REMOVED - A dashboard entry has been removed
- EVENT_ENTRY_UPDATED - A dashboard entry has been updated

Intended to be consumed by a `websocket` version of the `ping` endpoint. These events are for the short term transient state of the dashboard entry. This event may change in the future to include more than the online/offline state if we need to know more about what the device is doing but currently this is no need to extend it.
- EVENT_ENTRY_STATE_CHANGED - The state of a dashboard entry has changed (previously ping result)

Current nothing listens for these events yet, there are no events for discovery yet (aka importable)

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/dashboard/entries.py
+++ b/esphome/dashboard/entries.py
@@ -155,14 +155,15 @@ class DashboardEntries:
         path_to_cache_key = await self._loop.run_in_executor(
             None, self._get_path_to_cache_key
         )
+        entries = self._entries
         added: dict[DashboardEntry, DashboardCacheKeyType] = {}
         updated: dict[DashboardEntry, DashboardCacheKeyType] = {}
         removed: set[DashboardEntry] = {
             entry
-            for filename, entry in self._entries.items()
+            for filename, entry in entries.items()
             if filename not in path_to_cache_key
         }
-        entries = self._entries
+
         for path, cache_key in path_to_cache_key.items():
             if entry := self._entries.get(path):
                 if entry.cache_key != cache_key:
@@ -181,8 +182,8 @@ class DashboardEntries:
             entries[entry.path] = entry
             bus.async_fire(EVENT_ENTRY_ADDED, {"entry": entry})
 
-        if entry in removed:
-            entries.pop(entry.path)
+        for entry in removed:
+            del entries[entry.path]
             bus.async_fire(EVENT_ENTRY_REMOVED, {"entry": entry})
 
         for entry in updated:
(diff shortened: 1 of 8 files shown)


esphome/dashboard/web_server.py
```diff
@@ -785,9 +796,6 @@ def post(self, configuration=None):
             if build_folder is not None:
                 shutil.rmtree(build_folder, os.path.join(trash_path, name))
 
-        # Remove the old ping result from the cache
-        DASHBOARD.ping_result.pop(configuration, None)
```
bdraco, line 789: The object gets deleted now so no need to remove it here

esphome/dashboard/entries.py
```diff
@@ -3,24 +3,78 @@
...
+
+def bool_to_entry_state(value: bool) -> EntryState:
+    """Convert a bool to an entry state."""
+    return _BOOL_TO_ENTRY_STATE[value]
+
+
+def entry_state_to_bool(value: EntryState) -> bool | None:
+    """Convert an entry state to a bool."""
+    return _ENTRY_STATE_TO_BOOL[value]
+
```
bdraco, line 58: The ping api expects null/true/false so we need to be able to convert

esphome/dashboard/status/mdns.py
```diff
@@ -22,16 +24,16 @@ def __init__(self) -> None:
...
         self.host_mdns_state: dict[str, bool | None] = {}
-        # This is the hostnames to filenames mapping
-        self.host_name_to_filename: dict[str, str] = {}
-        self.filename_to_host_name: dict[str, str] = {}
+        # This is the hostnames to path mapping
+        self.host_name_to_path: dict[str, str] = {}
+        self.path_to_host_name: dict[str, str] = {}
         # This is a set of host names to track (i.e no_mdns = false)
         self.host_name_with_mdns_enabled: set[set] = set()
         self._loop = asyncio.get_running_loop()
```
bdraco, line 32: All these indices should be moved into `DashboardEntries` in a future PR so we don't duplicate work

esphome/dashboard/status/mdns.py
```diff
@@ -22,16 +24,16 @@ def __init__(self) -> None:
...
 
-    def filename_to_host_name_thread_safe(self, filename: str) -> str | None:
-        """Resolve a filename to an address in a thread-safe manner."""
-        return self.filename_to_host_name.get(filename)
+    def get_path_to_host_name(self, path: str) -> str | None:
+        """Resolve a path to an address in a thread-safe manner."""
+        return self.path_to_host_name.get(path)
```
bdraco, line 36: This should move to `DashboardEntries` as well in a future PR

balloob (approved):

PR #5794, opened 2023-11-19
Fixed UARTDevice::write_array() for std::vector

szupi-ipuzs:
The UARTComponent classes do not have an appropriate write_array() for std::vector, so compilation fails if one wants to use UARDevice::write_array(). The issue was originally created as libretiny-eu/libretiny#208

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
uart:
  - id: tuya_uart
    rx_pin: RX1
    tx_pin: TX1
    baud_rate: 9600
    debug:
      direction: BOTH
      dummy_receiver: true
      sequence:
        - lambda: UARTDebug::log_hex(direction, bytes, ' ');
        - lambda: !lambda |-
            static const std::vector<uint8_t> WIFI_RESET_COMMAND = { 0x55, 0xAA, 0x00, 0x04, 0x00, 0x01, 0x01, 0x05 };
            static const std::vector<uint8_t> WIFI_RESET_RESPONSE = { 0x55, 0xAA, 0x00, 0x04, 0x00, 0x00, 0x03 };
            if ((direction == uart::UART_DIRECTION_RX) && (bytes == WIFI_RESET_COMMAND))
            {
               id(tuya_uart).write_array(WIFI_RESET_RESPONSE);
            }
```

## Checklist:

### DIFF
--- a/esphome/components/waveshare_epaper/display.py
+++ b/esphome/components/waveshare_epaper/display.py
@@ -135,7 +135,7 @@ CONFIG_SCHEMA = cv.All(
             cv.Required(CONF_MODEL): cv.one_of(*MODELS, lower=True),
             cv.Optional(CONF_RESET_PIN): pins.gpio_output_pin_schema,
             cv.Optional(CONF_BUSY_PIN): pins.gpio_input_pin_schema,
-            cv.Optional(CONF_FULL_UPDATE_EVERY): cv.uint32_t,
+            cv.Optional(CONF_FULL_UPDATE_EVERY): cv.int_range(min=1, max=4294967295),
             cv.Optional(CONF_RESET_DURATION): cv.All(
                 cv.positive_time_period_milliseconds,
                 cv.Range(max=core.TimePeriod(milliseconds=500)),


kroimon:
There is `void UARTComponent::write_array(const std::vector<uint8_t> &data)`?
Here: https://github.com/esphome/esphome/blob/89c4d9b0fd5ea4b3cac2588fd1de243fda44a01b/esphome/components/uart/uart_component.h#L34

szupi-ipuzs:
> There is `void UARTComponent::write_array(const std::vector<uint8_t> &data)`? Here:
> 
> https://github.com/esphome/esphome/blob/89c4d9b0fd5ea4b3cac2588fd1de243fda44a01b/esphome/components/uart/uart_component.h#L34

That is not enough, as this method is not immediately visible in the derived classes. See this matching code snippet on godbolt: https://godbolt.org/z/4aWYaMvPG

So the fix is either to use the other write_array, as in this PR, or **each** XXXUARTComponent class, that derives from UARTComponent, would have to do this:
```cpp
class LibreTinyUARTComponent : public UARTComponent, public Component {
 public:
  ...
   using UARTComponent::write_array;
  ...
```

jesserockz:
Ah ok, I see where the real issue is, `id(tuya_uart)` is a `LibreTinyUARTComponent`, and not just a `UARTComponent` when trying to use inside lambdas

szupi-ipuzs:
@jesserockz , yes, exactly. Do you know if it is this way intentionally? Maybe a real fix should be done in somewhere in code generator instead... But that would be way out of my league :)

kbx81 (dismissed):

kbx81:
Ah, missed the discussion -- hopefully we can decide on a path forward soon. 😄

kroimon:
> Ah ok, I see where the real issue is, `id(tuya_uart)` is a `LibreTinyUARTComponent`, and not just a `UARTComponent` when trying to use inside lambdas

I think it actually makes sense the way it is now, because a board-specific classes could offer functions that are not available in the parent class but could be useful in lambdas.

I still don't understand why the existing function in `UARTComponent` isn't visible via the `LibreTinyUARTComponent` reference, or how this PR fixes that without changing the function signature. Anyone care to explain?

Edit: I think I now understand why it's not visible (shadowed by the `LibreTinyUARTComponent::write_array()` function?), but still don't get how this fixes it...

jesserockz:
One thing I don't understand is how these changes will fix the example yaml that doesn't even use any `UARTDevice`?

jesserockz:
The correct solution I believe for what this PR is trying to fix is to add `using UARTComponent::write_array` to each of the UART implementations as mentioned in https://github.com/esphome/esphome/pull/5794#issuecomment-1817986787

PR #5800, opened 2023-11-20
Remove CONNECTED state from BLE

bdraco:

It looks like this is no longer used as we always set the state to ESTABLISHED

Noticed while investigating https://github.com/esphome/issues/issues/5119

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
+++ b/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
@@ -240,8 +240,7 @@ void BluetoothProxy::bluetooth_device_request(const api::BluetoothDeviceRequest 
         this->send_device_connection(msg.address, false);
         return;
       }
-      if (connection->state() == espbt::ClientState::CONNECTED ||
-          connection->state() == espbt::ClientState::ESTABLISHED) {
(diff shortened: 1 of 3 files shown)


jesserockz:
Ah, there is a reason for this state as shown if you follow the codepath that starts at 
```
        this->set_state(espbt::ClientState::CONNECTED);
        this->state_ = espbt::ClientState::ESTABLISHED;
```

The client itself is set manually to established, but each potential `BLEClientNode` is marked as CONNECTED and allowed to do its own establishment checks and those nodes will change their own state to ESTABLISHED once they are done.

Once the `BLEClientNode` state changes to ESTABLISHED it will do its gatt reads/writes etc.

bdraco:
I will close this and audit for both state instead 👍 

I'm still digging through all the code paths in https://github.com/esphome/issues/issues/5119 but I haven't found a bug in esphome yet.  I'm thinking we need esp-idf 4.4.6 but not being available on platformio is a blocker for sure.

PR #5805, opened 2023-11-20
feat: Add HTU31D Support

betterengineering:
This change adds support for the HTU31D sensor, which is the third generation temperature/humidity sensor from TE and available via a breakout board from [Adafruit](https://www.adafruit.com/product/4832). 

I wrote this component by using the [HTU21D (previous generation)](https://esphome.io/components/sensor/htu21d) component as a starting point and referencing the [adafrult/Adafruit_HTU31D](https://github.com/adafruit/Adafruit_HTU31D/tree/master) library for the updated commands. In addition, I used the [datasheet](https://www.tti.com/content/dam/ttiinc/manufacturers/te-connectivity/PDF/HTU31_Sensors_Datasheet.pdf) and the [reference driver](https://github.com/TEConnectivity/HTU31D_Generic_C_Driver) to extend the capabilities provided from the Adafruit library.

**One design choice I'd love feedback on**: While I wrote functionality into the CPP component to be able to inspect and toggle the heater, I removed configuration values for the end user. I feel it's a confusing feature for a user of this component, especially given it took me a bit of research to fully understand what it's purpose was :sweat_smile: .

From [Adafruit](https://www.adafruit.com/product/4832):
> The sensor also has an on-board heater that can be controlled on and off to help evaporate condensation on the sensing element. Of course, temperature sensing will not be accurate with the heater on. 

From the [previous version's reference](https://cdn-shop.adafruit.com/datasheets/1899_HTU21D.pdf):
> The heater is intended to be used for functionality diagnosis: relative humidity drops upon rising temperature. The heater consumes about 5.5mW and provides a temperature increase of about 0.5-1.5°C.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes #3485 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3397

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: htu31d
    temperature:
      name: "Temperature"
    humidity:
      name: "Humidity"

external_components:
  - source: github://pr#5805
    components: [ htu31d ]
```

## Checklist:


betterengineering:
Currently testing this component against a Govee H5074 and will report back with a time series once I have more data:
image

esphome/components/htu31d/htu31d.h
```diff
@@ -0,0 +1,61 @@
...
+#define HTU31D_HEATERON (0x04)
+
+/** Disable heater */
+#define HTU31D_HEATEROFF (0x02)
+
+/** Reset command. */
+#define HTU31D_RESET (0x1E)
+
+/** Diagnostics command. */
+#define HTU31D_DIAGNOSTICS (0x08)
```
jesserockz: Please use `static const uint8_t xxxx = xx;` for values like this.
They can also move to the `.cpp` file as they are not used in the header.

esphome/components/htu31d/htu31d.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  uint8_t crc = compute_crc_((uint32_t) raw_temp << 8);
+  if (crc != thdata[2]) {
+    this->status_set_warning();
+    ESP_LOGE(TAG, "Error validating temperature CRC");
+    return;
+  }
+
+  temperature = raw_temp;
+  temperature /= 65535.0;
```
jesserockz:
```suggestion
  temperature /= 65535.0f;
```

esphome/components/htu31d/htu31d.cpp
```diff
@@ -0,0 +1,254 @@
...
+  if (this->read_register(HTU31D_READTEMPHUM, thdata, 6) != i2c::ERROR_OK) {
+    this->status_set_warning();
+    ESP_LOGE(TAG, "Error reading temperature/humidty register");
+    return;
+  }
+
+  // Calculate temperature value.
+  uint16_t raw_temp = thdata[0];
+  raw_temp <<= 8;
+  raw_temp |= thdata[1];
```
jesserockz:
```suggestion
  uint16_t raw_temp = encode_uint16(thdata[0], thdata[1]);
```

esphome/components/htu31d/htu31d.cpp
```diff
@@ -0,0 +1,254 @@
...
+ * the BSD license:
+ *   Written by Limor Fried/Ladyada for Adafruit Industries.
+ *   BSD license, all text above must be included in any redistribution.
+ *
+ * Modifications made by Mark Spicer.
+ */
+
+#include "htu31d.h"
+#include "esphome/core/log.h"
+#include "esphome/core/hal.h"
```
jesserockz, line 11: Required for `encode_uint16`
```suggestion
#include "esphome/core/hal.h"
#include "esphome/core/helpers.h"
```

esphome/components/htu31d/htu31d.cpp
```diff
@@ -0,0 +1,254 @@
...
+  temperature -= 40;
+
+  if (this->temperature_ != nullptr) {
+    this->temperature_->publish_state(temperature);
+  }
+
+  // Calculate humidty value.
+  uint16_t raw_hum = thdata[3];
+  raw_hum <<= 8;
+  raw_hum |= thdata[4];
```
jesserockz:
```suggestion
  uint16_t raw_hum = encode_uint16(thdata[3], thdata[4]);
```

esphome/components/htu31d/htu31d.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  crc = compute_crc_((uint32_t) raw_hum << 8);
+  if (crc != thdata[5]) {
+    this->status_set_warning();
+    ESP_LOGE(TAG, "Error validating humidty CRC");
+    return;
+  }
+
+  humidity = raw_hum;
+  humidity /= 65535.0;
```
jesserockz:
```suggestion
  humidity /= 65535.0f;
```

esphome/components/htu31d/htu31d.cpp
```diff
@@ -0,0 +1,254 @@
...
+    ESP_LOGE(TAG, "Error reading device serial");
+    return 0;
+  }
+
+  serial = reply[0];
+  serial <<= 8;
+  serial |= reply[1];
+  serial <<= 8;
+  serial |= reply[2];
+  serial <<= 8;
```
jesserockz:
```suggestion
  serial = encode_uint24(reply[0], reply[1], reply[2]);
```

esphome/components/htu31d/htu31d.h
```diff
@@ -0,0 +1,61 @@
...
+
+class HTU31DComponent : public PollingComponent, public i2c::I2CDevice {
+ public:
+  void setup() override;        /// Setup (reset) the sensor and check connection.
+  void update() override;       /// Update the sensor values (temperature+humidity).
+  void dump_config() override;  /// Dumps the configuration values.
+
+  void set_temperature(sensor::Sensor *temperature) { temperature_ = temperature; }
+  void set_humidity(sensor::Sensor *humidity) { humidity_ = humidity; }
+  void set_heater(sensor::Sensor *heater) { heater_ = heater; }
```
jesserockz: Not used
```suggestion
```

esphome/components/htu31d/htu31d.h
```diff
@@ -0,0 +1,61 @@
...
+
+ private:
+  bool reset_();
+  uint32_t read_serial_num_();
+  uint8_t compute_crc_(uint32_t value);
+
+ protected:
+  sensor::Sensor *temperature_{nullptr};
+  sensor::Sensor *humidity_{nullptr};
+  sensor::Sensor *heater_{nullptr};
```
jesserockz: Not used
```suggestion
```

esphome/components/htu31d/sensor.py
```diff
@@ -0,0 +1,56 @@
...
+)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await i2c.register_i2c_device(var, config)
+
+    if CONF_TEMPERATURE in config:
+        sens = await sensor.new_sensor(config[CONF_TEMPERATURE])
```
jesserockz: This is the newest way to do this that the codebase is slowly switching to
```suggestion
    if temperature_config := config.get(CONF_TEMPERATURE):
        sens = await sensor.new_sensor(temperature_config)
```

jesserockz (changes requested):
Overall looks good. Just a few changes below please =)

betterengineering:
Thanks for all the feedback @jesserockz! :raised_hands: Let me know if I missed anything

betterengineering:
Some more stats for the nerds :nerd_face: The blue line is for the HTU31D. It looks like the temperature difference is about 1 degree Fahrenheit, but consistently so:
image

For humidity, the blue line is for the HTU31D and it appears to be a 2% difference, but consistently so:
image

dbaarda:
First, thanks for doing this. I haven't had time to play with it yet, but would definitely never have found time to implement it myself. Maybe I'll get a chance to soon.

 @betterengineering: In my experience with an ESP32C3 and a BME280 (documented at https://community.home-assistant.io/t/success-esphome-esp32-c3-bme280-mqtt-and-deep-sleep-with-homeassistant/432067), a constant temp offset like that can be caused by just having your sensor and micro close enough in the same enclosure and not using deep-sleep. The heat generated by the micro power consumption bumps the temp up a bit. A temp offset also directly translates into a humidity difference, since humidity % depends on air temp.

I think I 've also seen that the temp difference can depend on how good your wifi signal is. I think a weaker wifi connection requires it to be driven harder, chewing more power, generating more heat. I'm using a case with a fair bit of sensor/micro separation and pretty good passive ventilation grills, but they are not thermally isolated well.

I just noticed above that the heater feature says "5.5mW and provides a temperature increase of about 0.5-1.5°C" . A non-sleeping busy ESP32C3 board using wifi eats about 50mA, which at 3.3V  is 165mW and drops to about 2mA or 6.6mW for "light sleep", and down to 0.1mA or 0.33mW for "deep sleep". These power numbers, even for "light sleep" are more than the "heater mode", so it's only the extra distance from the sensor and ventilation that means you don't see even higher temp rises without deep-sleeping.

dbaarda:
Err... I want this... can we please not forget to merge this?

What else still needs to be done?

dbaarda:
If the original author is AWOL I'm willing to take this over and apply the suggested fixes... any way I can easily do that?

<edit> actually, it looks like all the outstanding comments/suggestions have been addressed... I assume this is ready to go just pending approval by the reviewer?</edit>

betterengineering:
Yes, I believe I'm just waiting on an approval. @jesserockz is this good to go?

jesserockz:
Im just making a few minor updates, but yes it will be merged soon

jesserockz (approved):

PR #5811, opened 2023-11-21
Add some additional VA triggers, part 2

kbx81:

Adds TTS start and end triggers for VA.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3399

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/voice_assistant/__init__.py
+++ b/esphome/components/voice_assistant/__init__.py
@@ -1,5 +1,6 @@
 import esphome.config_validation as cv
 import esphome.codegen as cg
+import esphome.final_validate as fv
 
 from esphome.const import (
     CONF_ID,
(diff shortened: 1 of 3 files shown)


esphome/components/voice_assistant/__init__.py
```diff
@@ -56,6 +59,17 @@
...
 
+
+def final_validate(config):
+    if CONF_SPEAKER not in fv.full_config.get() and (
+        CONF_ON_TTS_STREAM_START in config or CONF_ON_TTS_STREAM_END in config
+    ):
+        raise cv.Invalid(
+            f"{CONF_SPEAKER} is required when using {CONF_ON_TTS_STREAM_START} and/or {CONF_ON_TTS_STREAM_END}"
+        )
+    return config
```
jesserockz: This doesn't need to be a final validate to check the full schema, it can be a normal validation and only has to check for CONF_SPEAKER in the voice_assistant schema.

jesserockz (changes requested):

esphome/components/voice_assistant/__init__.py
kbx81: Ah! Great catch, I wasn't thinking about it that way. 😅 Thank you!

jesserockz (approved):

PR #5813, opened 2023-11-21
Pulse meter sensor timeout fix

hannesb:

The timeout check is actually not against the timestamp of the last valid edge, but against the one before. So if i.e. the timeout is set to 4 minutes, the reset to 0 pulses/min occurs already if the pulse length is longer than 2 minutes.

This bugfix checks against the double value of 'timeout', and changes the max allowed timeout from 70 minutes to 35 minutes.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/pulse_meter/pulse_meter_sensor.cpp
+++ b/esphome/components/pulse_meter/pulse_meter_sensor.cpp
@@ -71,7 +71,10 @@ void PulseMeterSensor::loop() {
       // Running and initial states can timeout
       case MeterState::INITIAL:
       case MeterState::RUNNING: {
-        if (time_since_valid_edge_us > this->timeout_us_) {
+        // time_since_valid_edge_us is the time not since the last edge, but since the one before
+        // This is because the data structs are swapped
+        // So we have to compare against the double timeout value
+        if (time_since_valid_edge_us > this->timeout_us_ * 2) {
           this->meter_state_ = MeterState::TIMED_OUT;
           ESP_LOGD(TAG, "No pulse detected for %" PRIu32 "s, assuming 0 pulses/min",
                    time_since_valid_edge_us / 1000000);
--- a/esphome/components/pulse_meter/sensor.py
+++ b/esphome/components/pulse_meter/sensor.py
@@ -43,8 +43,8 @@ def validate_internal_filter(value):
 
 def validate_timeout(value):
     value = cv.positive_time_period_microseconds(value)
-    if value.total_minutes > 70:
-        raise cv.Invalid("Maximum timeout is 70 minutes")
+    if value.total_minutes > 35:
+        raise cv.Invalid("Maximum timeout is 35 minutes")
     return value
 
 


TrentHouliston:
Hi!
I see the problem you're describing.

Problem

However I don't think this is the way to fix it as it's treating the symptom and not the cause.
There is really one problem at play here that needs to be solved and that is that the pulse is not being published soon enough.

The first option that could help and was discussed in https://github.com/esphome/issues/issues/4807 was that the value could be published on the first valid falling edge like this.

Falling Edge

However, after thinking about it this won't solve the problem for all cases, for example.

Falling Edge Problem

I think the real way to fix this up is that the value should be published filter_length after the last rising edge.

Solution

This will require a little more work to do though but I think it's possible.

The isr filter will need to store an additional value in the `set` struct, that being the time of the last rising edge.
So:

```cpp
struct State {
  uint32_t last_detected_edge_us_ = 0;
  uint32_t last_rising_edge_us_ = 0;
  uint32_t count_ = 0;
};
```

That should be set each time that a rising edge is detected that would set `in_pulse` to true.

Then in the main loop you can check 
`last_rising_edge_ != last_detected_edge_ && last_rising_edge_ - now >= filter_length`
and if so `count = count + 1` and `last_detected_edge_ =  last_rising_edge_`
You will also need to set a flag that you have "peeked" the next pulse which means the next time you get a non zero count value you must subtract 1 from that count
That will let you always publish the value as soon as possible.

For the edge filter, they should just always be set to the same value since the edge filter doesn't suffer this off-by-one problem.

hannesb:
After digging a bit more into the code and trying to understand what is does I think the solution for the delayed power reporting is quite easy. These lines here
      
      // Low pulse of filter length now rising (therefore last_intr_ was the falling edge)
      else if (sensor->in_pulse_ && !sensor->last_pin_val_) {
        sensor->set_->last_detected_edge_us_ = sensor->last_edge_candidate_us_;
        sensor->set_->count_++;
        sensor->in_pulse_ = false;
      }

will only be executed after an at least filter_us_ long low signal and an at least filter_us_ long high signal, which is already all that the noise filter does. So I don't see the point why to use an µs value that is two edges old, IMHO you just can use now instead:
      
      // Low pulse of filter length now rising (therefore last_intr_ was the falling edge)
      else if (sensor->in_pulse_ && !sensor->last_pin_val_) {
        sensor->set_->last_detected_edge_us_ = now;
        sensor->set_->count_++;
        sensor->in_pulse_ = false;
      }

Which would solve both problems, the power reporting laging behind and the 0 pulses/min detection as well.

hannesb:
And I have an idea for a filter that not only debounces the edges, but will also handle spikes i.e. in the middle of a pulse. Unfortunately I don't have the time right now to explain this in detail. One downside of this idea is that their will be a half pulse length delay of power reporting, which is inevitable if the signal has to be analysed a bit before publishing a new value.

I don't have the time right now to explain my idee in detail, I think I will find a time slot next week for this.

TrentHouliston:
You're correct, the power can be reported earlier and `last_edge_candidate_us_` is not needed at all. That was the 3rd scenario that I was describing, it makes things much better but can still have problems if the pulse is high for a significant portion of the total wavelength.

I wonder if your idea is similar to the windowing filter I have sitting in my codebase.
I haven't had a chance to test it and have not needed the style of filtering it provides (also I wanted to make the window width make more sense with the configuration parameters). The idea of that filter is that over the last 2 filter widths, the value was high/low for at least filter width of that
https://github.com/esphome/esphome/compare/dev...TrentHouliston:esphome:houliston/window_filter

That branch also had the report earlier fix I was describing for scenario 2 on it. However  I hadn't tested it yet.
However, does not the fix that gives scenario 4.

hannesb:
I now had the time to implement my idea and do some tests. My idea is similar to yours, but my implementation is quite different from yours:
[Pulse-Meter-Sensor-advanced-filtering](https://github.com/hannesb/esphome/blob/Pulse-Meter-Sensor-advanced-filtering/esphome/components/pulse_meter/pulse_meter_sensor.cpp)
I will test my sensor for a few days. But actually the signal of my sensor is quite good, for a stress test of the filtering I would have to add a noise source...

PR #5817, opened 2023-11-21
implement web_server local fallback plus enhance entity config

afarago:
1) improvements of web servers for a v2+ esphome-webserver: adding entity_category, icons-where missing
2) web_server component should have the right to work even if internet is not available for the the IoT network.
I have created a minimal (511 byte with gz) implementation of a display only server.

At the moment this PR is more for validating the idea.
Also no idea where and how to submit the .html source.
[source html/js of web_server_fallback](https://github.com/esphome/esphome/files/13431834/web_server_fallback.zip)

This PR is functional on its own, yet if accepted esphome-webserver PR will follow.
https://github.com/afarago/esphome-webserver/tree/ha-styling

## Types of changes

- [ ] Other

No documentation change.

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

No changes on any configuration - any default web_server component is sufficient.

```yaml
# Example config.yaml
web_server:

```

## Checklist:


jesserockz:
The webserver assets are downloaded directly in the browser from the hosted site. It does not matter if the device network has no internet access. 

For example, my devices are on a NoT network, but I visit the `webserver` from my laptop on the lan network and my browser fetches the js file from the internet.

If someone is putting their device on a network with no internet and needs to access from a computer/device with no internet, then they should use `local: true` to serve the javascript directly from the device.

afarago:
> The webserver assets are downloaded directly in the browser from the hosted site. It does not matter if the device network has no internet access.
> 
> For example, my devices are on a NoT network, but I visit the `webserver` from my laptop on the lan network and my browser fetches the js file from the internet.
> 
> If someone is putting their device on a network with no internet and needs to access from a computer/device with no internet, then they should use `local: true` to serve the javascript directly from the device.

I am more like targeting a situation where normally there is internet, however due to an outage a local temporary fallback comes handy. At that point one would still like to access and interact with the node.
Currently if my home WAN is down I am unable to check the web_server from my laptop.

What do you think?

esphome/components/web_server/__init__.py
```diff
@@ -117,6 +117,9 @@ def build_index_html(config) -> str:
     html += "<esp-app></esp-app>"
     if config[CONF_JS_URL]:
         html += f'<script src="{config[CONF_JS_URL]}"></script>'
+    html += (
+        "<script>if (!window.source) {window.location.search = '?fallback';}</script>"
```
jesserockz: How exactly does this work?
If I run `window.source` in the console of any webpage it comes back as `undefined`

jesserockz: :facepalm: as soon as I commented that I realised it was being set by the esphome webserver frontend code

jesserockz (changes requested):
I am fine with the idea, but I think we should only include it if configured by the user.
`fallback_webpage: true/false` maybe for example? Thoughts?

Also, please open a PR on the webserver repo

afarago:
> I am fine with the idea, but I think we should only include it if configured by the user. `fallback_webpage: true/false` maybe for example? Thoughts?
> 
> Also, please open a PR on the webserver repo

I have added fallback_webpage option with default turned off (to be checked, yet better safe than sorry).
Unfortunately came combining two separate topics in one PR, so apologies for that.

If you accept here is my suggestion: please review and comment - if you are Ok with the code I would still cancel this PR and create two separate ones.

1) web_server fallback_webpage option
Next steps: I will still need to add the documentation - have not done the infra steps for checkout/build/check steps yet.

2) improvements of web servers adding entity_category, icons-where missing
Next steps: parallel I will go on with the esphome-webserver PR

Is that OK or do you have any other suggestions? Apologies for my mistake again.

jesserockz:
I am happy with both ideas, but it will be much easier to review and provide if they were split PRs =)

I am a little confused by some of the json changes but I can look more into that once they are split

afarago:
Split the changes to appropriate PRs, closing this PR.

1) styling
Created PRs for esphome/esphome#5854, esphome/esphome-webserver#49

2) local fallback
Created PRs for esphome/esphome#5853, esphome/esphome-docs#3419
Changed the option to `local_fallback` for consistency reasons.

PR #5821, opened 2023-11-22
feat(packages): support removing components

### LINKED ISSUE #2163: Support removing components defined in packages
We can extend components defined in packages like this:

```yaml
packages:
  athom.rgbct-light: github://athom-tech/athom-configs/athom-rgbct-light.yaml

button:
  - id: !extend Reset
    name: Overriden name
```

But I'd like to use a package, _and_ be able to _remove_ a component. I imagine something like this:

```
button:
  - id: !remove Reset
```

I'd like to use the packages defined by the vendor of my device, in this case Athom Tech. But sometimes they define odd components that I'd like to not have, but I can't remove them. 

A workaround is to disable or hide the component. In this case, with button, I can mark it as internal. I think the best solution would be to completely remove it from the configuration, which potentially could mean that "button" wasn't even compiled in to the final firmware. At the minimum, it's not managed in memory.

```yaml
button:
  - id: !extend Reset
    internal: true
```

**Additional context**
N/A

TMaYaD:

Items from lists can be removed by `id`.
```yaml
sensor:
  - id: !remove uptime_sensor
```

Items from dicts can be removed directly:
```yaml
voice_assistant:
  media_player: media_player.kitchen_speaker_2
  speaker: !remove
```

Works even at top level:
```yaml
captive_portal: !remove
```

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [esphome/feature-requests#2163](https://github.com/esphome/feature-requests/issues/2163)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3406

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

_Included in description._

## Checklist:

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -37,6 +37,7 @@
           "!secret scalar",
           "!lambda scalar",
           "!extend scalar",
+          "!remove scalar",
           "!include_dir_named scalar",
           "!include_dir_list scalar",
           "!include_dir_merge_list scalar",
--- a/esphome/config.py
+++ b/esphome/config.py
@@ -25,7 +25,7 @@ from esphome.core import CORE, EsphomeError
 from esphome.helpers import indent
 from esphome.util import safe_print, OrderedDict
 
-from esphome.config_helpers import Extend
+from esphome.config_helpers import Extend, Remove
 from esphome.loader import get_component, get_platform, ComponentManifest
 from esphome.yaml_util import is_secret, ESPHomeDataBase, ESPForceValue
 from esphome.voluptuous_schema import ExtraKeysInvalid
@@ -343,6 +343,12 @@ class LoadValidationStep(ConfigValidationStep):
                         path + [CONF_ID],
                     )
                     continue
+                if isinstance(p_id, Remove):
+                    result.add_str_error(
+                        f"Source for removal of ID '{p_id.value}' was not found.",
+                        path + [CONF_ID],
+                    )
+                    continue
                 result.add_str_error("No platform specified! See 'platform' key.", path)
                 continue
             # Remove temp output path and construct new one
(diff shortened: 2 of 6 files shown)


nielsnl68:
Just wondering; what is the use case for this kind of feature?

TMaYaD:
There are some use cases mentioned in the feature-request.

Personally, I have a little over 200 devices on my network which are made up of 4 types of hardware and a mix of traits I've made into packages. As the old adage goes, there is an exception to every rule and my packages are no exception. Sometimes it's just easy (and better DRY) to remove a component from a package for a device than add it to all the other devices. 

One example is where most of my fans are fan controllers but there is that one fan which is just an on/off and runs off of a 4 gang switch board. It's better abstraction to configure the switch board and *remove* the one switch and redefine it as a fan rather than redefine all the other 3.

jesserockz (commented):
:heavy_check_mark: Tentative approval on the code side until I test it locally

TMaYaD:
@jesserockz any update on this?

jesserockz:
I just have not had the chance to test this yet as I have been doing other stuff. I hopefully try it today

jesserockz (changes requested):
Ok just gave this a quick test and the removal process works as expected when you write the correct id. This brought up an issue though, which is the fact that you can `!remove any_id` and it doesn't care if it exists or not which should not be allowed and should be an invalid config.

TMaYaD:
@jesserockz Added validations as requested. Turns out I mis-interpreted the missing ID test for !extend (after which I modelled this PR) 🤣

barndawgie:
Would be great to see this merged soon; we have some usecases that will be much simpler with it!

jesserockz (approved):
Thanks for making that final change!

PR #5828, opened 2023-11-24
Speed up writing protobuf strings/bytes

bdraco:

Speed up writing protobuf strings/bytes

I recently got an esp32 cam and noticed the performance was not so great over the api (https://community.home-assistant.io/t/getting-better-performance-from-an-espcam32/561615)

I tried a few different ways ... and this might be faster but not sure if its safe
```cpp
    this->buffer_->resize(old_size + len);
    std::memcpy(&(*this->buffer_)[old_size], data, len);
```

so I settled on
```cpp
this->buffer_->insert(this->buffer_->end(), data, data + len);
```

after reading https://stackoverflow.com/questions/2208293/what-is-the-most-efficient-way-to-append-one-stdvector-to-the-end-of-another , in theory it should be 2x the speed of the `push_back` but I currently have no way to validate that its actually faster on the ESP and not just locally.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/api/proto.h
+++ b/esphome/components/api/proto.h
@@ -160,8 +160,7 @@ class ProtoWriteBuffer {
     this->encode_field_raw(field_id, 2);
     this->encode_varint_raw(len);
     auto *data = reinterpret_cast<const uint8_t *>(string);
-    for (size_t i = 0; i < len; i++)
-      this->write(data[i]);
+    this->buffer_->insert(this->buffer_->end(), data, data + len);
   }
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Seems like a very sensible change

bdraco:
thanks

PR #5834, opened 2023-11-25
Add Chamberlain/HomEntry HE60R garage door opener

clydebarrow:

Add Chamberlain/HomEntry HE60R garage door opener

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3413

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
    uart:
      - id: door_uart
        baud_rate: 1200
        parity: even
        stop_bits: 1
        tx_pin:
          number: 23
          inverted: true
        rx_pin: 22

    cover:
      platform: he60r
      id: garage_door
      name: Garage Door
      open_duration: 14s
      close_duration: 14s

    binary_sensor:
      - platform: gpio
        pin:
          number: 14
          mode:
            input: true
            pullup: true
        name: "Garage Door Button"
        filters:
          - delayed_on: 50ms
          - delayed_off: 50ms
        on_press:
          then:
            - cover.toggle: garage_door

```

## Checklist:


esphome/components/he60r/cover.py
```diff
@@ -0,0 +1,52 @@
...
+    CONF_OPEN_DURATION,
+    CONF_MAX_DURATION,
+)
+
+he60r_ns = cg.esphome_ns.namespace("he60r")
+HE60rCover = he60r_ns.class_("HE60rCover", cover.Cover, cg.PollingComponent)
+
+CONFIG_SCHEMA = (
+    cover.COVER_SCHEMA.extend(uart.UART_DEVICE_SCHEMA)
+    .extend(cv.polling_component_schema("300ms"))
```
jesserockz: Should this be changable by the user or is it better to always be ~300ms?
If the latter, it should not be a polling component, but setup its own internal interval so the user cannot change the `update_interval`

esphome/components/he60r/cover.py
```diff
@@ -0,0 +1,52 @@
...
+            cv.GenerateID(): cv.declare_id(HE60rCover),
+            cv.Optional(
+                CONF_OPEN_DURATION, default="15s"
+            ): cv.positive_time_period_milliseconds,
+            cv.Optional(
+                CONF_CLOSE_DURATION, default="15s"
+            ): cv.positive_time_period_milliseconds,
+            cv.Optional(
+                CONF_MAX_DURATION, default="60s"
+            ): cv.positive_time_period_milliseconds,
```
jesserockz: Won't these be different for everyone that has a different height/weight door? If so, it should be a required field with no default

esphome/components/he60r/he60r.h
```diff
@@ -0,0 +1,49 @@
...
+  void start_direction_(cover::CoverOperation dir);
+  void update_operation_(cover::CoverOperation dir);
+  void endstop_reached_(cover::CoverOperation operation);
+  void recompute_position_();
+  void set_current_operation_(cover::CoverOperation operation);
+  void process_rx_(uint8_t data);
+
+  uint32_t open_duration_{0};
+  uint32_t close_duration_{0};
+  unsigned int toggles_needed_{0};
```
jesserockz:
```suggestion
  uint32_t toggles_needed_{0};
```

jesserockz (changes requested):

esphome/components/he60r/cover.py
clydebarrow: It would, but the only use of the open and close durations is to estimate the current position, to provide feedback, IOW it does not affect the operation. The max duration is not going to be invoked very often - maybe the default for it should be longer, but the door controller already has logic to deal with faults anyway, so it's not particularly important. I don't particularly care either way, but left up to me, I'd leave it as it is.

esphome/components/he60r/cover.py
clydebarrow: Good question, I think it does matter, let me look at it more.

esphome/components/he60r/cover.py
clydebarrow: I've removed the MAX duration since that's the responsibility of the door controller. The other two I think can remain as optional since they are for cosmetic effect only.

esphome/components/he60r/he60r.cpp
```diff
@@ -0,0 +1,263 @@
...
+  }
+  if (this->current_operation != COVER_OPERATION_IDLE) {
+    this->recompute_position_();
+
+    // if we initiated the move, check if we reached the target position
+    if (this->last_command_ != COVER_OPERATION_IDLE) {
+      if (this->is_at_target_()) {
+        this->start_direction_(COVER_OPERATION_IDLE);
+      }
+    }
```
jesserockz, line 147: According to this code the timeouts will affect the actual operation state and may toggle the door etc?

clydebarrow: If you use the slider on the door control to request e.g. 50% it will stop the door when it thinks it's at 50%. There is no feedback from the controller on the position other than the endstops.

jesserockz: Ah ok. I guess the main part I am worried about is that if the default is 15s, but someones garage takes 16 or 17 seconds to open (I have never timed my garage) then the code will stop it too early?

Thats kinda also why I was mentioning not having any defaults for the times and making them required.

clydebarrow: No, under normal operation (open or close request) it will run until the endstop. The "partial open" function which isn't a core feature is the only thing that depends on the timing.

jesserockz: Full open just sets the target position to 1.0, and full close is 0.0.
Partial is just somewhere in between.

This section of code runs `recompute_position()` which calculates `this->position` based off the time it has run and the configured duration.

Then it checks if `is_at_target_()` and if the calculated `this->position` is above or equal to 1.0 for example, then it will start the IDLE direction, setting `toggles_needed_ to 3`

At least as far as I can tell without actually running the code.
I could be way off, but that's just what I see when following the `update_()` flow.

clydebarrow: `recompute_position` never delivers 0 or 1.0. 

```
  float min_pos = COVER_CLOSED + 0.01f;
  float max_pos = COVER_OPEN - 0.01f;

...

    this->position = clamp(delta + this->position, min_pos, max_pos);
```

clydebarrow: If I'd started completely from scratch I might have done it differently, but it's based on an existing component and I believe the logic is sound.

jesserockz: Confusion cleared up now

jesserockz (approved):

PR #5836, opened 2023-11-25
bugfix: Change setup priority for PMSA003I

### LINKED ISSUE #2344: pmsa003i no sensor data
### The problem

Set up an esp8266 thing with no other components, flashes, boots, and connects to wifi no trouble. ESPHome logs shows I2C bus has a device at address 0x12 as expected. Default configuration used from documentation, added an update rate of 5s to speed up testing.

Logging set to VERY_VERBOSE only shows a ping request, no other errors or information returned by the sensor. HA shows no data values in the entities (which loaded fine).

Deleted and re-added device just to be sure.

Happen to have bought two of these, tried with both.

2021.8.0

Home Assistant Add-on

_No response_

ESP8266

### Board

thing

pmsa003i

### Example YAML snippet

```yaml
captive_portal:

# Enable logging
logger:
  level: VERY_VERBOSE

# Enable Home Assistant API
api:

ota:

i2c:

sensor:
  - platform: pmsa003i
    pm_1_0:
      name: "PM1.0"
    pm_2_5:
      name: "PM2.5"
    pm_10_0:
      name: "PM10.0"
    pmc_0_3:
      name: "PMC <0.3µm"
    pmc_0_5:
      name: "PMC <0.5µm"
    pmc_1_0:
      name: "PMC <1µm"
    pmc_2_5:
      name: "PMC <2.5µm"
    pmc_5_0:
      name: "PMC <5µm"
    pmc_10_0:
      name: "PMC <10µm"
    update_interval: 5s
```

```txt
`[15:54:33][C][i2c:028]: I2C Bus:
[15:54:33][C][i2c:029]:   SDA Pin: GPIO2
[15:54:33][C][i2c:030]:   SCL Pin: GPIO14
[15:54:33][C][i2c:031]:   Frequency: 50000 Hz
[15:54:33][I][i2c:033]: Scanning i2c bus for active devices...
[15:54:33][I][i2c:040]: Found i2c device at address 0x12
[15:54:33][C][logger:189]: Logger:
[15:54:33][C][logger:190]:   Level: VERY_VERBOSE
[15:54:33][C][logger:191]:   Log Baud Rate: 115200
[15:54:33][C][logger:192]:   Hardware UART: UART0
[15:54:33][C][pmsa003i:021]:   Address: 0x12
[15:54:33][C][captive_portal:148]: Captive Portal:
[15:54:33][C][ota:029]: Over-The-Air Updates:
[15:54:33][C][ota:030]:   Address: aqsensor.local:8266
[15:54:33][C][api:095]: API Server:
[15:54:33][C][api:096]:   Address: aqsensor.local:6053
[15:54:34][VV][api.service:327]: on_ping_request: PingRequest {}
[15:54:34][VV][api.service:043]: send_ping_response: PingResponse {}
[15:54:38][VV][api.service:327]: on_ping_request: PingRequest {}
[15:54:38][VV][api.service:043]: send_ping_response: PingResponse {}`
```

### Additional information

_No response_

betterengineering:
This change updates the setup priority for the PMSA003I to LATE to avoid a common failure mode of reading from the device before it's ready. I personally was experiencing issues setting up the sensor, and when I logged the error from `this->read` I was receiving `ERROR_TIMEOUT`. Setting this to LATE resolved my issues, which I tried after reading [this thread](https://community.home-assistant.io/t/is-there-some-way-to-delay-setting-up-a-sensor-is-that-setup-priority/481700). I am making this change at the component level so others do not experience this point of friction.

## Types of changes
- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/2344

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
sensor:
  - platform: pmsa003i
    pm_1_0:
      name: "PM1.0"
    pm_2_5:
      name: "PM2.5"
    pm_10_0:
      name: "PM10.0"
    pmc_0_3:
      name: "PMC >0.3µm"
    pmc_0_5:
      name: "PMC >0.5µm"
    pmc_1_0:
      name: "PMC >1µm"
    pmc_2_5:
      name: "PMC >2.5µm"
    pmc_5_0:
      name: "PMC >5µm"
    pmc_10_0:
      name: "PMC >10µm"
```

## Checklist:

### DIFF
--- a/esphome/components/pmsa003i/pmsa003i.h
+++ b/esphome/components/pmsa003i/pmsa003i.h
@@ -32,7 +32,7 @@ class PMSA003IComponent : public PollingComponent, public i2c::I2CDevice {
   void setup() override;
   void dump_config() override;
   void update() override;
-  float get_setup_priority() const override { return setup_priority::DATA; }
+  float get_setup_priority() const override { return setup_priority::LATE; }
 
   void set_standard_units(bool standard_units) { standard_units_ = standard_units; }
 


clydebarrow:
It looks to me like the change in priority is a work-around and thus not necessarily the right solution. If the problem is that the device needs more time to initialise, that should be dealt with explicitly, not by relying on a side-effect of changing the priority.

betterengineering:
Hey @clydebarrow! Appreciate the feedback. It looks like the Adafruit examples for their library simply [delay for 1 second for the sensor to boot](https://github.com/adafruit/Adafruit_PM25AQI/blob/ba3f10c1056ba88306cb22e6002396655bfa6e99/examples/PM25_test/PM25_test.ino#L21-L22). I could instead add a 1 second delay to the device `setup()` or potentially read/delay in a loop until the read is successful. Either case, we need to wait or retry the I2C connection until the sensor is available.

sjtrny:
Agreed, since setup priority does not guarantee sufficient time has passed. I'm not sure whether a single delay or multiple retries is preferred (nor how many retires or maximum time is allowed) by the core devs. You'll probably get the fastest answer to that by asking in the devs channel on Discord.

clydebarrow:
>  I could instead add a 1 second delay to the device `setup()` or potentially read/delay in a loop until the read is successful. 

Definitely don't busy-loop, and inline delays are not good practice (will fail CI checks if a delay is more than a few ms) but you can certainly use `set_timeout()` to get a callback after any interval you like. Given that there is another example to back up your experience I would consider that a decent solution.

jesserockz:
Yes `set_timeout()` is the best option to something at a later point in time while letting the rest of the application get on with what it needs to do.

PR #5839, opened 2023-11-26
Improved sensor readings in htu21d component.

matzman666:

Made sensor readings asyncronous using set_timeout() and deleted any delay() call. This removes the warning that "Component htu21d took a long time for an operation".

Furthermore I increased the delay when reading the temperature from the sensor because the original delay was 50ms but according to datasheet the sht21 sensor can take up to 85 ms when reading the temperature.

## Types of changes

- [ ] Other

- https://github.com/esphome/esphome/pull/4641
- https://github.com/esphome/issues/issues/3549

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/htu21d/htu21d.cpp
+++ b/esphome/components/htu21d/htu21d.cpp
@@ -39,45 +39,53 @@ void HTU21DComponent::dump_config() {
   LOG_SENSOR("  ", "Humidity", this->humidity_);
 }
 void HTU21DComponent::update() {
-  uint16_t raw_temperature;
   if (this->write(&HTU21D_REGISTER_TEMPERATURE, 1) != i2c::ERROR_OK) {
     this->status_set_warning();
     return;
   }
-  delay(50);  // NOLINT
-  if (this->read(reinterpret_cast<uint8_t *>(&raw_temperature), 2) != i2c::ERROR_OK) {
-    this->status_set_warning();
-    return;
-  }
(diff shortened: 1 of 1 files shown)


jesserockz (approved):

xxVDRxx:
after update to 2023.12.3 my htu-21 doesn`t read humidity 
image

matzman666:
Ok, this is strange. The code reading the humidity is exactly the same code as before with the only difference that its execution is now scheduled using set_timeout() instead of blocking the main loop. 

Another change is that I increased the wait time between starting the temperature measurement and reading the result, but why should that affect the humidity measurement?

The new code has been tested using a sht21 sensor and everything was working as it should. I couldn't test with a htu21d sensor since I don't have one, but in theory it should behave the same as a sht21 (except of having different measurement delays).
 
What does esphome's log say?

xxVDRxx:
htu-21 working ok, I don`t know what happened but I have a new entity for humidity. I use MQTT.  Don`t have time today to analize this but it`s look like that: 
image

sensor.sypialnia_htu21d_sypialnia_wilgotnosc is new entity... strange... i didn`t change config etc, only upgrade esphome and upgrade a node via ota.

PR #5849, opened 2023-11-28
Pass through additional arguments to create number

clydebarrow:

The helper functions `switch.new_switch()` and `sensor.new_sensor()` already allow additional arguments to be passed to the object constructor. This should also be allowed with `number.new_number()`.

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/number/__init__.py
+++ b/esphome/components/number/__init__.py
@@ -257,8 +257,8 @@ async def register_number(
     )
 
 
-async def new_number(config, *, min_value: float, max_value: float, step: float):
-    var = cg.new_Pvariable(config[CONF_ID])
+async def new_number(config, *args, min_value: float, max_value: float, step: float):
+    var = cg.new_Pvariable(config[CONF_ID], *args)
     await register_number(
         var, config, min_value=min_value, max_value=max_value, step=step
     )


jesserockz:
image
image

I would actually argue to remove the `*args` from them entirely. 

Only really `micronova` uses the args for most of the existing functions, and it should really be using `Parented` for consistency with the rest of the codebase.

For more consistency, setters should be used everywhere for config -> class instead of constructor arguments.

WDYT?

clydebarrow:
What's the rationale for preferring setters over constructor arguments for invariant properties?

clydebarrow:
One of the use cases for constructor arguments is to get this kind of thing out of the C++ code:

```
#define SUB_SENSOR(name) \
 protected: \
  sensor::Sensor *name##_sensor_{nullptr}; \
\
 public: \
  void set_##name##_sensor(sensor::Sensor *sensor) { this->name##_sensor_ = sensor; }


#ifdef USE_SENSOR
  SUB_SENSOR(voltage)
  SUB_SENSOR(current)
  SUB_SENSOR(battery_level)
  SUB_SENSOR(max_cell_voltage)
  SUB_SENSOR(max_cell_voltage_number)
  SUB_SENSOR(min_cell_voltage)
  SUB_SENSOR(min_cell_voltage_number)
  SUB_SENSOR(max_temperature)
  SUB_SENSOR(max_temperature_probe_number)
  SUB_SENSOR(min_temperature)
  SUB_SENSOR(min_temperature_probe_number)
  SUB_SENSOR(remaining_capacity)
  SUB_SENSOR(cells_number)
  SUB_SENSOR(temperature_1)
  SUB_SENSOR(temperature_2)
  SUB_SENSOR(cell_1_voltage)
  SUB_SENSOR(cell_2_voltage)

(ad nauseum)


...


            if (this->voltage_sensor_) {
              this->voltage_sensor_->publish_state((float) encode_uint16(it[4], it[5]) / 10);
            }
            if (this->current_sensor_) {
              this->current_sensor_->publish_state(((float) (encode_uint16(it[8], it[9]) - DALY_CURRENT_OFFSET) / 10));
            }
            if (this->battery_level_sensor_) {
              this->battery_level_sensor_->publish_state((float) encode_uint16(it[10], it[11]) / 10);


(rinse and repeat...)

```

And instead replace it with more like this:
```
class SpaValue {
 public:
  SpaValue(char row, size_t col) : row{row}, col{col} {}
  const char row;
  const size_t col;
  virtual void set_value(int value) {}
};

class SpanetSensor : public SpaValue, public sensor::Sensor, public Parented<spanet::Spanet> {
 public:
  SpanetSensor(char row, size_t col, float scale): SpaValue(row, col), scale_{scale} {}
  void set_value(int value) override {
    this->publish_state(value * this->scale_);
  }
 protected:
  const float scale_;
};


....

    for (auto val : values_) {
      if (val->row == row && fieldcnt > val->col) {
        val->set_value(decode_value_(fields[val->col], field_lengths[val->col]));
      }
    }


```

Where the necessary parameters to evaluate and scale the value are supplied to the constructor for the Sensor subclass. Yes, it would be possible to use setters instead of constructor arguments, but that is a lot more verbose, and exposes the properties as settable when they should be invariant.

jesserockz:
Your two use cases are entirely different though...
The first are basic sensors that only expose a value with no custom code.
The second is a custom sensor class that does some processing itself and would still require the sensor to be defined and have a setter in the header...

Unless I am missing something?

clydebarrow:
They're the same use case, implemented differently. The classic approach has a separate named property for each possible sensor, requiring a separate setter for each one, and the handling of each is on a case-by-case basis in the parent component.  My approach is to use a custom class (which is derived from `Sensor`, `Number` etc.) which either does processing itself, or exposes properties that the parent component uses to evaluate and publish, without knowing anything much about the sensor itself (in the best case.)

In either case the parent component needs a method to add a sensor - but one requires eleventeen different setters, the other a single `add_sensor()` or similar.

I'm not suggesting that my approach to adding sensors should be mandated (for a simple component with a couple of sensors it's overkill) but I do think it's a cleaner approach where there are many sensors for one parent component.

The relevance to this PR (which is _not_ about setters on the parent component) is that to implement the second approach you have to initialise properties in the `Sensor` subclass, and doing that via setters is sub-optimal, since it prevents them being immutable and is verbose.

This PR isn't essential in that the code in `new_number` can be replicated in the component itself, but IMHO it's better to keep the pattern in common code.

Which also means we've spent far too much time discussing this already :-)

jesserockz (approved):

jesserockz:
Ill merge this now for consistency. And we can revisit the ideas outside this PR

PR #5857, opened 2023-11-28
Added nano seconds function for time stamp resolution near to nano seconds

erapade:

We currently only have a timestamp function that measures down to micro seconds. In some situations there's a need to have time stamps with higher resolutions and hence I added the nanos() function.

I did this knowing that the resolution will not come down to nano seconds due to the system clock frequency but this will still be better than just using the already existing micros().

* Using a 80 MHz system clock the resolution will be 12.5 nano seconds
* using a 120 MHz system clock the resolution will be 6.25 nano seconds

This function will be used to increase the reliability of timing in the Dallas component

And yes, 12.5 nano seconds in resolution will not give us an accuracy of 12.5 nano seconds, but using relative time measures this will help.

There was already and existing arch_get_cpu_cycle_count() that will do the same thing but this function wasn't declared as IRAM_ATTR HOT and didn't follow the same naming convention as micros() and millis()

This is my first pull request of code, Please be nice to me ;-)

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/host/core.cpp
```diff
@@ -28,13 +30,27 @@ void IRAM_ATTR HOT delay(uint32_t ms) {
...
+
 uint32_t IRAM_ATTR HOT micros() {
   struct timespec spec;
   clock_gettime(CLOCK_MONOTONIC, &spec);
   time_t seconds = spec.tv_sec;
   uint32_t us = round(spec.tv_nsec / 1e3);
   return ((uint32_t) seconds) * 1000000U + us;
 }
+
+uint32_t IRAM_ATTR HOT nanos() {
```
jesserockz, line 42: This file is only used when building a native linux binary for example, not on esp32/esp8266 etc...
You will need to implement on all platforms in their respective components.

jesserockz (changes requested):

esphome/components/host/core.cpp
erapade: Ohhh, Thanks @jesserockz. Based on your comment I looked into that. Makes sense since I then can include those components as local in my yaml. But I really didn't understand how to implement it in ESP8266 and ESP32 components and the current implementation of micros() in these components looks very different.

Any chance you can point me to some directions. This is my first attempt to contribute to ESPHome

ssieb: I think microseconds is the smallest interval easily available.  You can try looking through the ESP docs to see if you can find something else.

PR #5869, opened 2023-12-01
core: Don't delay loop if already late

gaaf:

The application loop would delay a full loop_interval_ when the last loop run already took more than loop_interval_ time.

Instead, don't delay if we're already late and run the next loop as soon as possible.

This makes the code do what the docs for `set_loop_interval()` says:
> If the loop() method takes longer than the target interval, ESPHome won't sleep in loop()

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/core/application.cpp
+++ b/esphome/core/application.cpp
@@ -82,9 +82,7 @@ void Application::loop() {
   const uint32_t now = millis();
   const uint32_t elapsed = now - this->last_loop_;
 
-  if (HighFrequencyLoopRequester::is_high_frequency() || elapsed >= this->loop_interval_) {
-    yield();
-  } else {
+  if (!HighFrequencyLoopRequester::is_high_frequency() && elapsed < this->loop_interval_) {
     uint32_t delay_time = this->loop_interval_ - elapsed;
     uint32_t next_schedule = this->scheduler.next_schedule_in().value_or(delay_time);
     // next_schedule is max 0.5*delay_time
@@ -93,6 +91,7 @@ void Application::loop() {
     delay_time = std::min(next_schedule, delay_time);
     delay(delay_time);
   }
+
   this->last_loop_ = now;
 
   if (this->dump_config_at_ < this->components_.size()) {
--- a/esphome/core/application.cpp
+++ b/esphome/core/application.cpp
@@ -81,16 +81,17 @@ void Application::loop() {
 
   const uint32_t now = millis();
   const uint32_t elapsed = now - this->last_loop_;
+  uint32_t delay_time = 1;  // minimum delay due to tasks at prio 0
 
   if (!HighFrequencyLoopRequester::is_high_frequency() && elapsed < this->loop_interval_) {
-    uint32_t delay_time = this->loop_interval_ - elapsed;
+    delay_time = this->loop_interval_ - elapsed;
     uint32_t next_schedule = this->scheduler.next_schedule_in().value_or(delay_time);
     // next_schedule is max 0.5*delay_time
     // otherwise interval=0 schedules result in constant looping with almost no sleep
     next_schedule = std::max(next_schedule, delay_time / 2);
     delay_time = std::min(next_schedule, delay_time);
-    delay(delay_time);
   }
+  delay(delay_time);
 
   this->last_loop_ = now;
 
--- a/esphome/core/application.cpp
+++ b/esphome/core/application.cpp
@@ -79,8 +79,19 @@ void Application::loop() {
   }
   this->app_state_ = new_app_state;
 
-  const uint32_t now = millis();
-  const uint32_t elapsed = now - this->last_loop_;
+  if (this->dump_config_at_ < this->components_.size()) {
+    if (this->dump_config_at_ == 0) {
+      ESP_LOGI(TAG, "ESPHome version " ESPHOME_VERSION " compiled on %s", this->compilation_time_);
+#ifdef ESPHOME_PROJECT_NAME
+      ESP_LOGI(TAG, "Project " ESPHOME_PROJECT_NAME " version " ESPHOME_PROJECT_VERSION);
+#endif
+    }
+
+    this->components_[this->dump_config_at_]->call_dump_config();
+    this->dump_config_at_++;
+  }
+
+  const uint32_t elapsed = millis() - this->last_loop_;
   uint32_t delay_time = 1;  // minimum delay due to tasks at prio 0
 
   if (!HighFrequencyLoopRequester::is_high_frequency() && elapsed < this->loop_interval_) {
@@ -93,19 +104,7 @@ void Application::loop() {
   }
   delay(delay_time);
 
-  this->last_loop_ = now;
-
-  if (this->dump_config_at_ < this->components_.size()) {
-    if (this->dump_config_at_ == 0) {
-      ESP_LOGI(TAG, "ESPHome version " ESPHOME_VERSION " compiled on %s", this->compilation_time_);
-#ifdef ESPHOME_PROJECT_NAME
-      ESP_LOGI(TAG, "Project " ESPHOME_PROJECT_NAME " version " ESPHOME_PROJECT_VERSION);
-#endif
-    }
-
-    this->components_[this->dump_config_at_]->call_dump_config();
-    this->dump_config_at_++;
-  }
+  this->last_loop_ = millis();
 }
 
 void IRAM_ATTR HOT Application::feed_wdt() {
--- a/esphome/core/application.cpp
+++ b/esphome/core/application.cpp
@@ -80,14 +80,12 @@ void Application::loop() {
   this->app_state_ = new_app_state;
 
   const uint32_t now = millis();
+  const uint32_t elapsed = now - this->last_loop_;
 
-  if (HighFrequencyLoopRequester::is_high_frequency()) {
+  if (HighFrequencyLoopRequester::is_high_frequency() || elapsed >= this->loop_interval_) {
     yield();
   } else {
-    uint32_t delay_time = this->loop_interval_;
-    if (now - this->last_loop_ < this->loop_interval_)
-      delay_time = this->loop_interval_ - (now - this->last_loop_);
-
+    uint32_t delay_time = this->loop_interval_ - elapsed;
     uint32_t next_schedule = this->scheduler.next_schedule_in().value_or(delay_time);
     // next_schedule is max 0.5*delay_time
     // otherwise interval=0 schedules result in constant looping with almost no sleep


esphome/core/application.cpp
```diff
@@ -80,14 +80,12 @@ void Application::loop() {
   this->app_state_ = new_app_state;
 
   const uint32_t now = millis();
+  const uint32_t elapsed = now - this->last_loop_;
 
-  if (HighFrequencyLoopRequester::is_high_frequency()) {
+  if (HighFrequencyLoopRequester::is_high_frequency() || elapsed >= this->loop_interval_) {
```
clydebarrow: Can I suggest the following:

```
  const uint32_t elapsed = now - this->last_loop_;
  uint32_t delay_time = 0;
  if (!HighFrequencyLoopRequester::is_high_frequency() && elapsed < this->loop_interval_) {
    delay_time = this->loop_interval_ - elapsed;

    uint32_t next_schedule = this->scheduler.next_schedule_in().value_or(delay_time);
    // next_schedule is max 0.5*delay_time
    // otherwise interval=0 schedules result in constant looping with almost no sleep
    next_schedule = std::max(next_schedule, delay_time / 2);
    delay_time = std::min(next_schedule, delay_time);
  }
  delay(delay_time);
```

gaaf: I don't know what you're trying to accomplish with a commit that totally reverts my patch.

May I suggest that my fix is applied as-is first, so the bug is fixed and the code is logically correct and more readable. Then there is a solid base to do any refactoring like eliminating the yield.

Even if you insist on doing it in a single PR, there was really no need to revert my patch first, you could have done your proposal on top of that.

I don't think the delay(0) is necessary so near the end of the loop.

clydebarrow: If there was any actual commit to your PR, then that was accidental and I apologise - my intent was simply to make a suggestion.

`delay(0)` has the same effect as `yield()`.

jesserockz: I think the point of `delay(0)` and `yield()` is to allow the CPU to work on other RTOS tasks/threads that might be running?

clydebarrow: Yes, even if the loop has run over time, it would be wise to let the scheduler do its job.

gaaf: Control goes back to the "OS" at the end of the application loop. A yield/delay(0) just before returning makes no sense.

Just before this code there is a tight loop calling all components' loop() (via ->call()). This is the part of the application loop that takes the most time. If the OS would need more time, this would be the place to add a yield, not near the end of the application loop.
```diff
--- a/esphome/core/application.cpp
+++ b/esphome/core/application.cpp
@@ -71,16 +71,17 @@ void Application::loop() {
   for (Component *component : this->looping_components_) {
     {
       WarnIfComponentBlockingGuard guard{component};
       component->call();
     }
     new_app_state |= component->get_component_state();
     this->app_state_ |= new_app_state;
     this->feed_wdt();
+    yield();
   }
   this->app_state_ = new_app_state;
 
   const uint32_t now = millis();
   const uint32_t elapsed = now - this->last_loop_;
 
   if (!HighFrequencyLoopRequester::is_high_frequency() && elapsed < this->loop_interval_) {
     uint32_t delay_time = this->loop_interval_ - elapsed;
```

Well, what IMHO would be even nicer is moving the delay calculation all the way to the end:
```diff
--- a/esphome/core/application.cpp
+++ b/esphome/core/application.cpp
@@ -79,6 +79,18 @@ void Application::loop() {
   }
   this->app_state_ = new_app_state;
 
+  if (this->dump_config_at_ < this->components_.size()) {
+    if (this->dump_config_at_ == 0) {
+      ESP_LOGI(TAG, "ESPHome version " ESPHOME_VERSION " compiled on %s", this->compilation_time_);
+#ifdef ESPHOME_PROJECT_NAME
+      ESP_LOGI(TAG, "Project " ESPHOME_PROJECT_NAME " version " ESPHOME_PROJECT_VERSION);
+#endif
+    }
+
+    this->components_[this->dump_config_at_]->call_dump_config();
+    this->dump_config_at_++;
+  }
+
   const uint32_t now = millis();
   const uint32_t elapsed = now - this->last_loop_;
 
@@ -92,19 +104,7 @@ void Application::loop() {
     delay(delay_time);
   }
 
-  this->last_loop_ = now;
-
-  if (this->dump_config_at_ < this->components_.size()) {
-    if (this->dump_config_at_ == 0) {
-      ESP_LOGI(TAG, "ESPHome version " ESPHOME_VERSION " compiled on %s", this->compilation_time_);
-#ifdef ESPHOME_PROJECT_NAME
-      ESP_LOGI(TAG, "Project " ESPHOME_PROJECT_NAME " version " ESPHOME_PROJECT_VERSION);
-#endif
-    }
-
-    this->components_[this->dump_config_at_]->call_dump_config();
-    this->dump_config_at_++;
-  }
+  this->last_loop_ = millis();
 }
 
 void IRAM_ATTR HOT Application::feed_wdt() {
```

And set last_loop_ to the actual time the loop ended while at it.

clydebarrow: > If the OS would need more time, this would be the place to add a yield, not near the end of the application loop.

The OS doesn't "need more time", a `yield()` would just allow another task of the same priority to run. Which leads to an interesting question about the whole premise of your change - if there is a _lower_ priority task than the `loop` task, and the `loop` task never delays (blocks) due to one or more components taking excessively long, then the lower priority task will never run, `yield()` or no `yield()`.

So maybe instead of not delaying at all if the loop has run over time, it might be better to just delay by a smaller time than the default loop time - e.g. just one tick. That would then allow lower priority tasks to run at least for a short period.

gaaf: "more time" was maybe a bit badly worded.

My interpretation is the following: 

ESP32-IDF has a preemptive scheduler. The application loop is run at prio 1, which is the lowest prio above idle. Equal and higher prio tasks will get scheduled. Lower (idle) doesn't need to get scheduled if we need the cpu time.

On Arduino boards, the arduino core will "schedule" anything between loop() calls.

But, I'm not married to the proposal of eliminating the yield, its' just that I think it is weird to yield near the end of the application loop instead of within a tight loop within the application loop.

Any comments on the diffs from my previous comments? If not I can add them to the PR.

clydebarrow: Are you confident that no other tasks are created in ESPHome with priority lower than 1?

jesserockz: `esp32_camera` and `i2s_audio` speaker have tasks with 0 priorities

clydebarrow: I was hoping the OP would do his own checks :-)

One of those tasks is pinned to core 1 so should not be an issue, and I would question why the other is not at a higher priority than the loop task anyway, but this does conjure up the "law of unintended consequences." A seemingly minor change to improve something can have knock-on effects, especially when it's in the bowels of the core code.

If it were up to me, I would not change the basic logic, just clean up the code and reduce any excess delay to a (non-zero) minimum - I believe that would achieve most of the objective with little risk of breaking anything else.

gaaf: > I was hoping the OP would do his own checks :-)

Thanks for your kind words. OP however is not a fulltime esphome developer and has limited time available.

The reason I didn't check the prio's of other threads initially, is because it didn't seem necessary as the changes do not change the documented behavior of:
> If the loop() method takes longer than the target interval, **ESPHome won't sleep in loop()**.

I.e. Other code should already take into account that the main loop (at prio 1) will not sleep.


That being said, because it was specifically asked for in https://github.com/esphome/esphome/pull/5869#discussion_r1417119659, I would have checked it, time permitting.



> esp32_camera and i2s_audio speaker have tasks with 0 priorities

Back in the days code competing with the idle thread would be considered code that could have been left out entirely, as it's clearly not important enough to get executed.
I don't know what those tasks do, but I suspect a bug there as any non-idle task should get prio >= 1.

I'll update the PR with a new version. Could take a bit of time though.

clydebarrow: > Thanks for your kind words

I presume you're being sarcastic, but I'm not having a go at you - I'm trying to educate. Your basic idea is good, and I'm just trying to help you get it implemented without breaking anything.

> The reason I didn't check the prio's of other threads initially, is because it didn't seem necessary as the changes do not change the documented behavior

Yep, documentation is often wrong - I have no doubt whoever wrote that believed it to be true. Trust but verify. 

> I don't know what those tasks do, but I suspect a bug there as any non-idle task should get prio >= 1.

I agree. The question is whether you want to tackle finding and fixing any such code that relies on un-documented behaviour - scope creep is a terrible thing.

> I'll update the PR with a new version. Could take a bit of time though

No problem - better to take time, do it well, and not have to go into disaster recovery mode.

gaaf: I've looked at the prio 0 tasks.

i2s_audio: This task feeds the speaker from a buffer. This should be a higher prio task than the loop task.

esp32_camera (pinned to core 1): This task reads from the camera into a buffer. This should be a higher prio task than idle.

esp32_ble_beacon (pinned to core 1): This one is weird. It only does a setup, then loops over 1 sec delays. This task should probably get deleted after setup has completed. It also needs a prio>0 to guarantee execution.

esphome/core/application.cpp
```diff
@@ -108,6 +90,21 @@ void Application::loop() {
     this->components_[this->dump_config_at_]->call_dump_config();
     this->dump_config_at_++;
   }
+
+  const uint32_t elapsed = millis() - this->last_loop_;
+  uint32_t delay_time = 1;  // minimum delay due to tasks at prio 0
```
clydebarrow, line 95:
```suggestion
  uint32_t delay_time = portTICK_PERIOD_MS;  // minimum delay due to tasks at prio 0
```

FreeRTOS schedules in "ticks", and the minimum possible delay is one tick, requesting a delay less than one tick is the same as `delay(0)`. This is how `delay()` is implemented.

```
void IRAM_ATTR HOT delay(uint32_t ms) { vTaskDelay(ms / portTICK_PERIOD_MS); }
```

Now I believe that in ESPHome by default `portTICK_PERIOD_MS` _is_ equal to 1, but using the defined constant eliminates any dependence on that.

Otherwise looks all good.

gaaf: Except that the constant is only available on esp-idf.

Adding a high delay kind of defeats the purpose of the PR.

I'll look into another way to trigger a schedule to a lower prio task. Or just attempt fix the current prio 0 tasks first, eventually that needs to be fixed anyway.

clydebarrow: > Except that the constant is only available on esp-idf.

What makes you think that? `portTICK_PERIOD_MS` is a constant associated with FreeRTOS which is used by ESPHome on all platforms, it is not specific to ESP-IDF.

esphome/core/application.cpp
gaaf: I read the code for esp32_ble_beacon wrong. It is pinned to core 0 at prio 1. Should not pose a problem.

esphome/core/application.cpp
gaaf: Now that the prio 0 task in components have been eliminated, I'll remove this patch from the PR.

How is FreeRTOS involved on archs other than esp32-idf? I don't see any task created and only esp32-idf has an `app_main()` function.

I hope my initial assumption that on other archs the "OS" or arch code takes care of other tasks between running the app's `loop()` function. If not please holler.

clydebarrow: Like I said "FreeRTOS ... is used by ESPHome on all platforms, it is not specific to ESP-IDF." 

> I hope my initial assumption that on other archs the "OS" or arch code takes care of other tasks between running the app's `loop()` function. If not please holler.

Assumptions are often wrong. The best venue for questions like that is in the Discord channels, probably [#off-topic](https://discord.com/channels/429907082951524364/822162083452813312).

clydebarrow, line 95: Now that the lower-priority tasks have been eliminated by the other PRs, I would see no issue with  just setting this to 0.
```suggestion
  uint32_t delay_time = 0;
```

PR #5886, opened 2023-12-03
Waveshare: support red and black on EDP_7in5b_V2

### LINKED ISSUE #239: Support for  Waveshare Three-color E-Ink Displays

References #77. 
I've got Waveshare's 3-color e-ink display and it would be nice to have some support for them.

**Additional context**

twisterss:

- Add support of multiple colors to `WaveshareEPaper` class
- Create class `WaveshareEPaperPolled`, which inherits from `WaveshareEPaper` and uses a small state machine to avoid blocking waits (especially important for big screens that need more than 10s to refresh)
- Create class `WaveshareEPaper7In5BV2`, which inherits from `WaveshareEPaperPolled` and implements support for red and black on display `EDP_7in5b_V2` (7.5 inches with a v3 sticker on the back, based on the demo code here: https://www.waveshare.com/wiki/7.5inch_e-Paper_HAT_(B)_Manual#ESP32.2F8266 )

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/239
The issue seems to discuss different screen types, this PR adds support only for one type, not sure it will work for all.

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
display:
  - platform: waveshare_epaper
    id: epaper
    cs_pin: 15
    dc_pin: 27
    busy_pin:
      number: 25
      inverted: True
    reset_pin: 26
    reset_duration: 2ms
    model: 7.50in-bv2-rb
```

## Checklist:

### DIFF
--- a/esphome/components/waveshare_epaper/display.py
+++ b/esphome/components/waveshare_epaper/display.py
@@ -24,6 +24,9 @@ WaveshareEPaper = waveshare_epaper_ns.class_("WaveshareEPaper", WaveshareEPaperB
 WaveshareEPaperBWR = waveshare_epaper_ns.class_(
     "WaveshareEPaperBWR", WaveshareEPaperBase
 )
+WaveshareEPaperPolled = waveshare_epaper_ns.class_(
+    "WaveshareEPaperPolled", WaveshareEPaper
+)
 WaveshareEPaperTypeA = waveshare_epaper_ns.class_(
     "WaveshareEPaperTypeA", WaveshareEPaper
 )
@@ -98,6 +101,9 @@ WaveshareEPaper13P3InK = waveshare_epaper_ns.class_(
     "WaveshareEPaper13P3InK", WaveshareEPaper
 )
 GDEW0154M09 = waveshare_epaper_ns.class_("GDEW0154M09", WaveshareEPaper)
+WaveshareEPaper7In5BV2 = waveshare_epaper_ns.class_(
+    "WaveshareEPaper7In5BV2", WaveshareEPaperPolled
+)
 
 WaveshareEPaperTypeAModel = waveshare_epaper_ns.enum("WaveshareEPaperTypeAModel")
 WaveshareEPaperTypeBModel = waveshare_epaper_ns.enum("WaveshareEPaperTypeBModel")
@@ -128,6 +134,7 @@ MODELS = {
     "5.83inv2": ("b", WaveshareEPaper5P8InV2),
     "7.50in": ("b", WaveshareEPaper7P5In),
     "7.50in-bv2": ("b", WaveshareEPaper7P5InBV2),
+    "7.50in-bv2-rb": ("b", WaveshareEPaper7In5BV2),
     "7.50in-bv3": ("b", WaveshareEPaper7P5InBV3),
     "7.50in-bc": ("b", WaveshareEPaper7P5InBC),
     "7.50inv2": ("b", WaveshareEPaper7P5InV2),
--- a/esphome/components/waveshare_epaper/waveshare_epaper.cpp
+++ b/esphome/components/waveshare_epaper/waveshare_epaper.cpp
@@ -3034,51 +3034,51 @@ void WaveshareEPaper13P3InK::dump_config() {
 
 void WaveshareEPaperPolled::update() {
   this->do_update_();
-  if (this->state_ == State::sleeping) {
-    this->set_state_(State::update_requested);
+  if (this->state_ == State::SLEEPING) {
+    this->set_state_(State::UPDATE_REQUESTED);
   }
 }
 
 void WaveshareEPaperPolled::loop() {
   switch (this->state_) {
-    case State::sleeping:
+    case State::SLEEPING:
       break;
-    case State::update_requested:
+    case State::UPDATE_REQUESTED:
       this->reset_pin_->digital_write(false);
-      this->set_state_(State::resetting);
+      this->set_state_(State::RESETTING);
       break;
-    case State::resetting:
+    case State::RESETTING:
       if (millis() - this->last_state_change_ >= this->reset_duration_) {
         this->reset_pin_->digital_write(true);
-        this->set_state_(State::initializing);
+        this->set_state_(State::INITIALIZING);
       }
       break;
-    case State::initializing:
+    case State::INITIALIZING:
       if (millis() - this->last_state_change_ >= 200) {
         this->power_on();
-        this->set_state_(State::powering_on);
+        this->set_state_(State::POWERING_ON);
       }
       break;
-    case State::powering_on:
+    case State::POWERING_ON:
       if (millis() - this->last_state_change_ >= 100 && (!this->busy_pin_ || !this->busy_pin_->digital_read())) {
         this->configure();
-        this->set_state_(State::configuring);
+        this->set_state_(State::CONFIGURING);
       }
       break;
-    case State::configuring:
+    case State::CONFIGURING:
       this->display();
-      this->set_state_(State::displaying);
+      this->set_state_(State::DISPLAYING);
       break;
-    case State::displaying:
+    case State::DISPLAYING:
       if (millis() - this->last_state_change_ >= 200 && (!this->busy_pin_ || !this->busy_pin_->digital_read())) {
         this->power_off();
-        this->set_state_(State::powering_off);
+        this->set_state_(State::POWERING_OFF);
       }
       break;
-    case State::powering_off:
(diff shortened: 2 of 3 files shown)


esphome/components/waveshare_epaper/waveshare_epaper.cpp
```diff
@@ -129,26 +129,37 @@ void WaveshareEPaper::update() {
...
-  for (uint32_t i = 0; i < this->get_buffer_length_(); i++)
-    this->buffer_[i] = fill;
+  uint32_t offset = 0;
+  for (const auto &buf_color : this->get_supported_colors()) {
+    // A bit set to 1 means color is off
+    uint8_t value = 0x00;
+    if (color != buf_color) {
+      value = 0xFF;
+    }
+    for (uint32_t i = 0; i < (this->get_width_internal() * this->get_height_internal()) / 8u; i++) {
```
lougovsk, line 178: I wonder wouldn't it make sense to move the color mapping to the display function and rather fill the buffer_ array in a slightly different way. One could use bit-wise color coding here  meaning 
black: 1
red: 1<<1
-- potential further colors e.g. in 7color version---
green 1<<2
....
This trick would allow to keep size of buffer_ fixed (~xres*yres)
Of course this means that the display function would iterate over the same array n times (where n is number of supported colors) and apply mask and get actual color value there and send it to the display.

twisterss: Indeed it would be nice but it’s not possible with the current implementation because it is already optimal (it uses one bit per pixel, not one byte). So we need more space to store multiple colors. 

What could be done would be to have a different layout, but it would use the same array size. The implementation in this PR is:
B0 B1 B2 B3 R0 R1 R2 R3
we could use instead:
B0 R0 B1 R1 B2 R2 B3 R3
(B0 meaning color black on pixel 0)

It might allow some optimizations, but would use the exact same space.

lougovsk: Ah true. Sorry. After digging into the implementation I see what you mean. 
Regarding the color ordering I would say the current  implementation makes more sense.  Since it is easier to add new colors.

esphome/components/waveshare_epaper/waveshare_epaper.cpp
```diff
@@ -2129,5 +2147,164 @@ void WaveshareEPaper2P13InDKE::set_full_update_every(uint32_t full_update_every)
...
+  this->data(0x00);
+  this->data(0x00);
+  this->data(0x00);
+  this->data(0x00);
+}
+
+void HOT WaveshareEPaper7In5BV2::display() {
+  uint32_t buf_len = this->get_buffer_length_();
+  // COMMAND DATA START TRANSMISSION NEW DATA
+  this->command(0x10);
```
lougovsk, line 3169: I guess here you send COLOR_ON data in the first loop
Then change the color to red and send the second part of the buffer which corresponds to red pixels?
Is it so?

twisterss: Yes that's the idea. To be clear, I don't know exactly what each of the command does (0x10, 0x92, 0x13), but I know it's the right sequence to send pixels to the display because it's mimicked from the example code provided by WaveShare.

Also an important thing to note is that the pixel values for the red layer are flipped (`~this->buffer_[i]`)

lougovsk: So I've taken a look at [waveshare implementation](https://github.com/waveshareteam/e-Paper/blob/6db502f3dd3a8a5760ea127cf1d142195165f64c/RaspberryPi_JetsonNano/c/lib/e-Paper/EPD_7in5b_V2.c#L213) . They actually don't use 0x92 command, which seems to perform a partial update. Have you tried without it? 

this->command(0x12); does display refresh 
but according to https://github.com/waveshareteam/e-Paper/blob/6db502f3dd3a8a5760ea127cf1d142195165f64c/RaspberryPi_JetsonNano/c/lib/e-Paper/EPD_7in5b_V2.c#L92 we also it claims that one needs 
 DEV_Delay_ms(100);	        //!!!The delay here is necessary, 200uS at least!!!

twisterss: Indeed I realize I did not link to the right reference. I used the library provided by waveshare for the esp32, found here: https://files.waveshare.com/upload/5/50/E-Paper_ESP32_Driver_Board_Code.7z
The relevant file is esp32-waveshare-epd/src/utility/EPD_7in5b_V2.cpp . Here is the display function: 
```cpp
void EPD_7IN5B_V2_Display(const UBYTE *blackimage, const UBYTE *ryimage)
{
    UDOUBLE Width, Height;
    Width =(EPD_7IN5B_V2_WIDTH % 8 == 0)?(EPD_7IN5B_V2_WIDTH / 8 ):(EPD_7IN5B_V2_WIDTH / 8 + 1);
    Height = EPD_7IN5B_V2_HEIGHT;

 //send black data
    EPD_7IN5B_V2_SendCommand(0x10);
    for (UDOUBLE j = 0; j < Height; j++) {
        for (UDOUBLE i = 0; i < Width; i++) {
            EPD_7IN5B_V2_SendData(blackimage[i + j * Width]);
        }
    }
    EPD_7IN5B_V2_SendCommand(0x92); 

    //send red data
    EPD_7IN5B_V2_SendCommand(0x13);
    for (UDOUBLE j = 0; j < Height; j++) {
        for (UDOUBLE i = 0; i < Width; i++) {
            EPD_7IN5B_V2_SendData(~ryimage[i + j * Width]);
        }
    }
    EPD_7IN5B_V2_TurnOnDisplay();
}
```
Anyway I tried without the 0x92 command as you suggested and it works. I'll remove it.

About the delay after command 0x12: indeed it is required. And it is done in my code through the state machine here: 
```cpp
      this->display();
      this->set_state_(State::DISPLAYING);
      break;
    case State::DISPLAYING:
      if (millis() - this->last_state_change_ >= 200 && (!this->busy_pin_ || !this->busy_pin_->digital_read())) {
        this->power_off();
        this->set_state_(State::POWERING_OFF);
      }
      break;
```

I replaced all delays by asynchronous state changes to avoid long busy waits blocking the device (the refresh is slow, so that would trigger the watchdog)

ser:
we want this please merge :)

Freighter:
This Update is highly recomended and demanded by ePaper Users.
Hope the Issue with permanent powering the epaper will be solved soon, cause the bleeding in high light conditions can do serious damage to the displays. but first things first. This is a more important upstream. 

You all do a pretty good job on this. Keep up the awesome work !

clydebarrow:
if you want this reviewed, you'll need to resolve the conflicts.

twisterss:
Hi,
sorry I missed the requested changes on this PR.
I resolved the conflicts. It was a bit challenging because a change adding Black and Red support in a slightly different way was merged.

Here is what I did:
* I removed WaveshareEPaperBase because WaveshareEPaper just supports one or multiple colors directly depending on the value of get_supported_colors().
* WaveshareEPaperBWR simply overrides get_supported_colors() to support two colors

esphome/components/waveshare_epaper/display.py
```diff
@@ -98,6 +101,9 @@
     "WaveshareEPaper13P3InK", WaveshareEPaper
 )
 GDEW0154M09 = waveshare_epaper_ns.class_("GDEW0154M09", WaveshareEPaper)
+WaveshareEPaper7In5BV2 = waveshare_epaper_ns.class_(
+    "WaveshareEPaper7In5BV2", WaveshareEPaperPolled
+)
```
jesserockz, line 106: Please follow the naming scheme that is already used
```suggestion
WaveshareEPaper7P5InBV2RB = waveshare_epaper_ns.class_(
    "WaveshareEPaper7P5InBV2RB", WaveshareEPaperPolled
)
```

esphome/components/waveshare_epaper/waveshare_epaper.h
```diff
@@ -7,9 +7,9 @@
 namespace esphome {
 namespace waveshare_epaper {
 
-class WaveshareEPaperBase : public display::DisplayBuffer,
```
jesserockz, line 10: Such a big change should be made independently of adding new display models

jesserockz (changes requested):

swoboda1337:
Stale

PR #5909, opened 2023-12-12
UART change at runtime

edwardtfn:

This adds the method `load_settings()` to the UART component on ESP32 (`arduino` and `esp-idf`) allowing the interface to reload with the new settings used by methods `set_baud_rate`, `set_parity`, `set_data_bits`, etc. without the assignment of a new software UART.

This can be useful for connecting to devices which supports multiple baud rates or for devices which changes it's serial communication at runtime, but also allows some logic for a user's configurable baud rate to be set at `on_boot` after the UART component is loaded.

## Types of changes

- [ ] Other

Related Discord thread: https://discord.com/channels/429907082951524364/1180653963461275648

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3455

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
select:
  - id: change_baud_rate
    name: Baud rate
    platform: template
    options:
      - "2400"
      - "9600"
      - "38400"
      - "57600"
      - "115200"
      - "256000"
      - "512000"
      - "921600"
    initial_option: "115200"
    optimistic: true
    restore_value: True
    internal: false
    entity_category: config
    icon: mdi:swap-horizontal
    set_action:
      - lambda: |-
          tf_uart->flush();
          ESP_LOGD("change_baud_rate", "Set new baud rate to %i", stoi(x));
          tf_uart->set_baud_rate(stoi(x));
          tf_uart->load_settings();
```

## Checklist:


esphome/components/uart/uart_component_esp32_arduino.cpp
```diff
@@ -109,6 +109,11 @@ void ESP32ArduinoUARTComponent::setup() {
     this->number_ = next_uart_num;
     this->hw_serial_ = new HardwareSerial(next_uart_num++);  // NOLINT(cppcoreguidelines-owning-memory)
   }
+
+  load_settings(false);
```
jesserockz: Please ensure that you are prefixing member variables and function calls
```suggestion
  this->load_settings(false);
```

esphome/components/uart/uart_component_esp32_arduino.cpp
```diff
@@ -118,9 +123,15 @@ void ESP32ArduinoUARTComponent::setup() {
     invert = true;
   this->hw_serial_->setRxBufferSize(this->rx_buffer_size_);
   this->hw_serial_->begin(this->baud_rate_, get_config(), rx, tx, invert);
+  if (dump_config) {
+    ESP_LOGCONFIG(TAG, "UART %u was reloaded.", this->number_);
+    dump_config_();
```
jesserockz:
```suggestion
    this->dump_config_();
```

esphome/components/uart/uart_component_esp_idf.cpp
```diff
@@ -122,9 +122,23 @@ void IDFUARTComponent::setup() {
...
+void IDFUARTComponent::load_settings(bool dump_config) {
+  uart_config_t uart_config = this->get_config_();
+  esp_err_t err = uart_param_config(this->uart_num_, &uart_config);
+  if (err != ESP_OK) {
+    ESP_LOGW(TAG, "uart_param_config failed: %s", esp_err_to_name(err));
+    this->mark_failed();
+    return;
+  } else if (dump_config) {
+    ESP_LOGCONFIG(TAG, "UART %u was reloaded.", this->uart_num_);
+    dump_config_();
```
jesserockz:
```suggestion
    this->dump_config_();
```

esphome/components/uart/uart_component_esp32_arduino.cpp
```diff
@@ -118,9 +123,15 @@ void ESP32ArduinoUARTComponent::setup() {
...
+  }
 }
 
-void ESP32ArduinoUARTComponent::dump_config() {
+void ESP32ArduinoUARTComponent::dump_config() { this->dump_config_(); }
```
jesserockz: What is the purpose of having this intermediate function?

jesserockz (changes requested):

esphome/components/uart/uart_component_esp32_arduino.cpp
edwardtfn: The idea is to have it showing the new settings on the log when it changes. Some of the validations won't support a directly call to `dump_config()` and I don't wanna to duplicate it's code, so I've used a different name for the function and kept the original `dump_config()` with the reference to the other one as this is called when log starts.

edwardtfn: Well, look again at the Action's logs maybe it haven't failed due to calling `dump_config()`, but due to calling it directly (instead of `this->dump_config()`). I will try that again as it will make the code cleaner...

edwardtfn: Done!

jesserockz (approved):

nagyrobi:
@edwardtfn please add to the docs (next branch) a note containing the example above mentioning which uart parameters can be reconfigured at runtime.

edwardtfn:
I can do that, but as none of those pre-existing functions to change settings (`set_baud_rate`, `set_parity`, etc.) are not there, I think we should add a more extensive section describing those a bit.
I can take a look at this later.

nagyrobi:
Problem is that this PR was merged, functionality is in, but it's not documented for the users.

edwardtfn:
There is: https://github.com/esphome/esphome-docs/pull/3455

Please let me know your thoughts.

nagyrobi:
Nice, but I also see changes to the `components/display/nextion.rst` file. Those should be in a separate PR.

edwardtfn:
Hmmm
That shouldn't be there...
I will take a look soon.

edwardtfn:
Now it is correct at https://github.com/esphome/esphome-docs/pull/3456

edwardtfn:
I've also added some inline documentation at #5930.
Although that isn't changing any line of code, I will run some sanity check later and then request an approval.

PR #5910, opened 2023-12-12
Nextion upload TFT with a different baud rate

edwardtfn:
Attention: This have a dependance to #5909 which only supports ESP32 `arduino` and `esp-idf`. Waiting for that PR to be merged before we can run checks for this one, otherwise it will fail.

This enables the use of Nextion capability to upload TFT with a different baud rate than the used in the normal communication, which allows faster transfers with a higher baud rate or, in case of failures on the transfer, a lower baud rate could be used to reduce chances of errors (https://esphome.io/components/display/nextion.html#uploading-a-tft-file).
The baud rate won't change (current baud rate will be used) when not explicitly selected by the code. 

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [Pending] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
api:
  services:
    - service: update_nextion
      then:
        - lambda: 'id(nextion1)->upload_tft(9600);'
```

## Checklist:


edwardtfn:
I'm closing this as it is not stable on my tests.
As uart can be reset now, we can still change the communication baud rate before calling a TFT upload with similar results.
In the future I might come back to this.

PR #5915, opened 2023-12-12
Fix crash when handling pin_check error

clydebarrow:

The recently added check for multiple pin used threw an unhandled exception in certain cases when handling an error.
This turned out to be due to some incorrect code added back in 2021 - the `path` property of a `vol.Invalid` is read-only. Fixed here.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


clydebarrow:
Turns out this was actually triggered by a change in voluptuous:

```
Breaking change: Invalid.path and Invalid.error_message are no longer writable (properties).
```

jesserockz (approved):

PR #5921, opened 2023-12-13
Fix the initial run of lambda light effects

smithjacobj:
The timer used for `millis()` is a monotonic timer based on the last start time of the device. If, for some reason, you pick a long `update_interval` and try to apply it as soon as you start the device, nothing happens because the device hasn't been on for longer than the `update_interval`

Why would you want a long `update_interval`? Data-over-power lights tend to flicker with updates, so maybe you have a slower or static effect you want to apply immediately but run slowly.

Lambda light effects now apply immediately, regardless of `update_interval` or `millis()` uptime.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable): N/A

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
light:
  - ...
    effects:
      - addressable_lambda:
          name: Candy Cane
          update_interval: 1h
          lambda: |-
            for (int i = 0; i < it.size(); i++) {
              if (i & 1) {
                it[i] = Color(255, 0, 0);
              } else {
                it[i] = Color(255, 255, 255);
              }
            }
```

## Checklist:

### DIFF
--- a/esphome/components/light/addressable_light_effect.h
+++ b/esphome/components/light/addressable_light_effect.h
@@ -57,7 +57,7 @@ class AddressableLambdaLightEffect : public AddressableLightEffect {
   void start() override { this->initial_run_ = true; }
   void apply(AddressableLight &it, const Color &current_color) override {
     const uint32_t now = millis();
-    if (now - this->last_run_ >= this->update_interval_) {
+    if (now - this->last_run_ >= this->update_interval_ || this->initial_run_) {
       this->last_run_ = now;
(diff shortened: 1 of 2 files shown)


kroimon (approved):
Untested, but looks like a good bug fix to me 👍

kbx81 (approved):

PR #5924, opened 2023-12-14
Add ability to lock to set mode

ysmilda:

It allows locking the read out of the HLW8012 component to a single mode (e.g current or voltage). This is useful when you're not interested in one of the two values and want to continuously read back the other.

This allows the `change_mode_every` parameter to be set to `"never"` which locks the mode to the set `initial_mode`.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/3473

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

With the below config the `mode` will be locked to `VOLTAGE`.

```yaml
# Example config.yaml
  - platform: hlw8012
    change_mode_every: "never"
    initial_mode: VOLTAGE
```

## Checklist:

### DIFF
--- a/tests/test1.yaml
+++ b/tests/test1.yaml
@@ -794,7 +794,7 @@ sensor:
     update_interval: 15s
     current_resistor: 0.001 ohm
     voltage_divider: 2351
-    change_mode_every: 16
+    change_mode_every: "never"
     initial_mode: VOLTAGE
     model: hlw8012
   - platform: total_daily_energy
--- a/esphome/components/hlw8012/hlw8012.cpp
+++ b/esphome/components/hlw8012/hlw8012.cpp
@@ -96,7 +96,7 @@ void HLW8012Component::update() {
     this->energy_sensor_->publish_state(energy);
   }
 
-  if (this->change_mode_at_++ == this->change_mode_every_) {
+  if (this->change_mode_every_ != 0 && this->change_mode_at_++ == this->change_mode_every_) {
     this->current_mode_ = !this->current_mode_;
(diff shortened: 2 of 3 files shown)


jesserockz (changes requested):
This would be better to change `change_every` to allow `never` similar to `update_interval`s do

ysmilda:
I like that proposal, but I'm not sure how to implement that. The `update_interval`s take a time string, this is than checked to be `never` or parsed to an `uint32_t` in milliseconds. The `change_every` already takes an `uint32_t` and changing that to a `string` would break everyone's config. 

Ideally `change_every` would take either a `string` or an `int` but I'm unsure how that could be implemented.

jesserockz:
You can use something like `cv.Any(..., "never")` to allow the current time and the hardcoded string never. And then convert never into `0` and check for 0 in the cpp code since the minimum is currently 1

ysmilda:
Thanks for the insight, I've changed `change_every` to accept `"never"` and that locks the mode to the one set in `initial_mode`.

tests/test1.yaml
jesserockz, line 1: While it is better in the end, please undo the unrelated YAML changes here.

ysmilda: I've undone the changes, formatting on save is a blessing and a curse.

kroimon (approved):
LGTM

tests/test1.yaml
jesserockz: It is yes. Feel free to make a PR dedicated to just formatting the YAML files though =)

jesserockz (approved):

PR #5925, opened 2023-12-14
Add quad spi features

clydebarrow:

Add functionality to the SPI component (ESP-IDF only) to allow the use of quad SPI transactions. This will support
the addition of the LilyGo T-Display-S3 AMOLED board and other displays with quad SPI interfaces.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3454

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
spi:
  - id: spi_id_1
    clk_pin: GPIO7
    mosi_pin: GPIO6
    interface: any
  - id: quad_spi
    clk_pin: 47
    data_pins:
      - 40
      - 41
      - 42
      - 43
```

## Checklist:


esphome/components/spi/__init__.py
```diff
@@ -249,14 +251,26 @@ def get_spi_interface(index):
     return "new SPIClass(HSPI)"
 
 
+# Do not use a pin schema for the number, as that will trigger a pin reuse error due to duplication of the
+# clock pin in the standard and quad schemas.
```
jesserockz, line 255: Why would you reuse the pin?

esphome/components/spi/__init__.py
```diff
@@ -267,8 +281,34 @@ def get_spi_interface(index):
...
+SPI_QUAD_SCHEMA = cv.All(
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(QuadSPIComponent),
+            cv.Required(CONF_CLK_PIN): clk_pin_validator,
+            # Optional is used here so that the error messages when no pins are defined
+            # relates to the standard schema, not this.
+            cv.Optional(CONF_DATA_PINS): cv.All(
+                cv.ensure_list(pins.gpio_output_pin_schema), cv.Length(min=4, max=4)
+            ),
```
jesserockz: You can add a `msg` argument to `cv.Any` with a custom message to print if neither are matched.

esphome/components/spi/__init__.py
clydebarrow: You wouldn't, but the pin gets logged into the registry when the pin validation is evaluated, so if the first schema  fails (due e.g. to miso/mosi missing), the clock pin gets entered twice in the registry.

esphome/components/spi/__init__.py
```diff
@@ -267,8 +281,35 @@ def get_spi_interface(index):
     cv.only_on([PLATFORM_ESP32, PLATFORM_ESP8266, PLATFORM_RP2040]),
 )
 
+SPI_QUAD_SCHEMA = cv.All(
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(QuadSPIComponent),
+            cv.Required(CONF_CLK_PIN): clk_pin_validator,
+            # Optional is used here so that the error messages when no pins are defined
+            # relates to the standard schema, not this.
```
jesserockz: No longer relevant?
```suggestion
```

jesserockz (changes requested):

esphome/components/spi/__init__.py
```diff
@@ -267,8 +281,33 @@ def get_spi_interface(index):
...
 )
 
+SPI_QUAD_SCHEMA = cv.All(
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(QuadSPIComponent),
+            cv.Required(CONF_CLK_PIN): clk_pin_validator,
+            cv.Required(CONF_DATA_PINS): cv.All(
+                cv.ensure_list(pins.gpio_output_pin_schema), cv.Length(min=4, max=4)
+            ),
```
jesserockz: Does a full pin schema make sense here when they need to be the number only when configuring the bus and cant be inverted or external etc?

If not, then the below code that calls `set_data_pins` can be simplified to only set a vector of the pin numbers instead of the full objects...

jesserockz (dismissed):

esphome/components/spi/__init__.py
clydebarrow: I think the schema is fine, firstly it ensures that only output-capable pins are listed, secondly it allows for the (admittedly unlikely) multiple use of a pin. But it could well be that only the number need be used in the vector. I'll look at that again when I get back to a computer.

jesserockz: What I mean is, `pins.internal_gpio_output_pin_number` might be better. IIRC it still allows the multi use, but restricts to an internal pin number only

jesserockz (approved):

PR #5926, opened 2023-12-14
Update typing to modern syntax (see PEP 563)

bdraco:

Update typing to modern syntax (see PEP 563)

This PR is best reviewed via CLI since github web ui can't handle lots of one line changes

All changes were done with a small script to add `from __future__ import annotations` and than `ruff --fix` and `pyupgrade --py39-plus` to ensure there wasn't room for manual error.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:
Why is the import added to files that don't need it, e.g. `esphome/__init__.py` which is (was) an empty file? Since this is a work-around until 3.10 shouldn't it only be used where necessary (since it will subsequently be removed.)

bdraco:
It's a hint to anyone who makes changes to that file in the future that they can use the new typing.

It's 💯 a preference to add it or not as new typing is not required.  I'm generally in favor of adding it everywhere since it makes it clear it's allowed

clydebarrow:
I would like to think that we'll move to 3.1x in the not too distant future, and pyupgrade can automatically convert use of `Optional` to the new syntax anyway, so I would need more convincing that this is worth changing over 1000 files, which then will need updating again soon.

bdraco:
We can wait for py 3.10. There isn't enough typing to make it matter that much anyways

PR #5929, opened 2023-12-14
ESP32-S3 and ESP-IDF don't play well with USB_CDC and need USB_SERIAL_JTAG

clydebarrow:
instead.

Fixes compile errors when compiling for ESP-IDF on ESP32-S3 (and probably S2) due to change introduced with
#4853.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/logger/__init__.py
```diff
@@ -171,8 +171,8 @@ def validate_local_no_higher_than_global(value):
                 CONF_HARDWARE_UART,
                 esp8266=UART0,
                 esp32=UART0,
-                esp32_s2=USB_CDC,
-                esp32_s3=USB_CDC,
+                esp32_s2=USB_SERIAL_JTAG,
+                esp32_s3=USB_SERIAL_JTAG,
```
kbx81: In light of #5928, we can (and should) do this now:
```suggestion
                esp32_s2=USB_CDC,
                esp32_s3=USB_SERIAL_JTAG,
                esp32_s3_arduino=USB_CDC,
```
...the S2 does not support JTAG at all and CDC with the S3 won't work correctly with IDF.

kbx81 (dismissed):

esphome/components/logger/__init__.py
```diff
@@ -172,7 +172,8 @@ def validate_local_no_higher_than_global(value):
                 esp8266=UART0,
                 esp32=UART0,
                 esp32_s2=USB_CDC,
-                esp32_s3=USB_CDC,
+                esp32_s3=USB_SERIAL_JTAG,
```
kbx81: One more time 😅 
```suggestion
                esp32_s3_idf=USB_SERIAL_JTAG,
```

kbx81 (dismissed):

clydebarrow:
gonna add the C3 as well.

clydebarrow:
Tested on S3, C3 and plain ESP32. Also tested on RP2040.

clydebarrow:
Arduino on ESP32-C3 is still not right. The current setup results in a default of UART0, which means no output when using the USB_CDC port. If I tweak the config to allow USB_CDC for C3 then it fails at compile time because `Serial` is not defined by Arduino - it's `Serial0`. If I replace `Serial` with `Serial0` then another compile error in the platformio main.cpp blocks it. If I work-around that I can get it to build and flash, but then it bootloops :-(

So for now I think it's best left as is and note that USB serial is not supported on the C3 under Arduino.

kbx81 (approved):

kbx81:
> Arduino on ESP32-C3 is still not right. The current setup results in a default of UART0, which means no output when using the USB_CDC port. If I tweak the config to allow USB_CDC for C3 then it fails at compile time because `Serial` is not defined by Arduino - it's `Serial0`. If I replace `Serial` with `Serial0` then another compile error in the platformio main.cpp blocks it. If I work-around that I can get it to build and flash, but then it bootloops :-(

It behaves this way even [with this added](https://github.com/esphome/esphome/blob/0a188ad9d22a2b5a8060ec7ad3318cb9dba509e2/esphome/components/logger/__init__.py#L263-L265)? 🤔

clydebarrow:
> It behaves this way even [with this added](https://github.com/esphome/esphome/blob/0a188ad9d22a2b5a8060ec7ad3318cb9dba509e2/esphome/components/logger/__init__.py#L263-L265)? 🤔

Yes. That works fine with S3, but on C3 `Serial` (without a 0) is undefined, not only in ESPHome code but also in platformio code. And fixing that leads to the bootloop. I just don't think the Arduino support for the C3 is mature.


```
/Users/clyde/.platformio/packages/framework-arduinoespressif32@3.20005.220925/cores/esp32/main.cpp: In function 'void app_main()':
/Users/clyde/.platformio/packages/framework-arduinoespressif32@3.20005.220925/cores/esp32/main.cpp:58:5: error: 'Serial' was not declared in this scope
     Serial.begin();
     ^~~~~~
```

clydebarrow:
Oh, well this is weird, it's now working with Arduino on C3, with the default as UART0.

PR #5931, opened 2023-12-14
WiFi fast_connect: save/load BSSID and channel for faster connect from sleep

rguca:

The fast_connect option allows to skip the scan for networks. Still, a scan for BSSID and channel is performed, which results in a connect time around 4 seconds.
This PR allows to save BSSID and channel, so no scan has to be performed on wake. The connect time is reduced to 1 second -> significant battery life improvement for sleeping devices that only wake to push their data.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/wifi/wifi_component.cpp
+++ b/esphome/components/wifi/wifi_component.cpp
@@ -55,6 +55,9 @@ void WiFiComponent::start() {
   uint32_t hash = this->has_sta() ? fnv1_hash(App.get_compilation_time()) : 88491487UL;
 
   this->pref_ = global_preferences->make_preference<wifi::SavedWifiSettings>(hash, true);
+  if (this->fast_connect_) {
+    this->fast_connect_pref_ = global_preferences->make_preference<wifi::SavedWifiFastConnectSettings>(hash, false);
+  }
 
   SavedWifiSettings save{};
   if (this->pref_.load(&save)) {
@@ -78,6 +81,7 @@ void WiFiComponent::start() {
 
     if (this->fast_connect_) {
       this->selected_ap_ = this->sta_[0];
+      this->load_fast_connect_settings();
       this->start_connecting(this->selected_ap_, false);
     } else {
       this->start_scanning();
@@ -604,6 +608,11 @@ void WiFiComponent::check_connecting_finished() {
 
     this->state_ = WIFI_COMPONENT_STATE_STA_CONNECTED;
     this->num_retried_ = 0;
+
+    if (this->fast_connect_) {
+      this->save_fast_connect_settings();
+    }
+
     return;
   }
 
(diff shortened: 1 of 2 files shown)


jesserockz:
Why wouldn't you just add the bssid and channel into the yaml?

rguca:
If your network has multiple BSSIDs and the esp device is portable, it will do a scan every time the BSSID from the yaml is not in range.

jesserockz:
But setting the bssid in yaml vs it being loaded from the preferences in this PR is exactly the same...

rguca:
The difference is the preference will be updated, when the BSSID changes. This ensures a faster connection time for future connects. If you put the BSSID in the yaml, the connection time will be slow if the BSSID changes.

Unfocused:
> Why wouldn't you just add the bssid and channel into the yaml?

My WiFi network is 3 access points providing one network, and they automatically check congestion to switch to the least used channel every 2 weeks. So that's 1 SSID, 3 BSSIDs, and a channel that can randomly change.

It's not an unusual setup these days - consumer-level mesh WiFi systems are becoming more common, and a lot of them self-optimise. Then there are commercial/industrial environments, which will almost always have multiple access points.

As @rguca said, manually setting the BSSID and channel in YAML would only help when you've got only 1 access point. But it also only helps when the channel never changes - even for people with only 1 access point, that's not necessarily the case.

Saving them in preferences makes it a cache for the last known good connection, which is the most common case for reconnections for everyone. The times you wouldn't benefit would be (from most common to least common):
- Connecting to a different access point than last time (different BSSID, but same SSID)
- When the channel changes
- The first connection the device makes after being first flashed or factory reset (and maybe safe mode?)

Compared to saving in preferences, setting it in YAML would be only help with the last of those conditions - which should be the thing that happens the least often.

> But setting the bssid in yaml vs it being loaded from the preferences in this PR is exactly the same…

In addition to all the above, setting it in YAML requires that you know & manually specify the BSSID and channel - neither of which I would expect the average person to know how to find (doubly so because you'd almost never need to know them). Caching it in preferences makes it configuration-free, so everyone automatically benefits.

But I would suggest adding an optional `fast_connect_cache` config to allow disabling saving to preferences, for two reasons:
- If there's some worry about flash wear or running out of RTC memory
- The last known BSSID isn't necessarily the best BSSID. For some scenarios, connecting to the best BSSID could be more important than reconnecting as fast as possible.

jesserockz (approved):
Sorry for the delay. I just came back to this now and reading above, I think its fine to merge. This code is still "hidden" behind `fast_connect: true` so it should not affect most people that are not using that.

PR #5641, opened 2023-10-31
Support for CST 820 Touch screen controller

dirkjankrijnders:
# What does this implement/fix?

Added support for the CST860 Touch screen controller as found on 2432S024C boards. Implementation is based on other implementations as a datasheet for this controller is not readily available
## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
touchscreen:
  platform: cst860
  id: my_touchscreen
  i2c_id: bus_a
  transform:
    mirror_y: false
    swap_xy: false
  on_touch:
      - lambda: |-
            ESP_LOGI("cal", "x=%d, y=%d, x_raw=%d, y_raw=%0d",
                touches[0].x,
                touches[0].y,
                touches[0].x_raw,
                touches[0].y_raw
                );
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #5941, opened 2023-12-15
Add CST816 touchscreen driver

clydebarrow:

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3459

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


nielsnl68:
Is it not better to break this PR into two parts one for the SPI and the other for the touchscreen driver?

clydebarrow:
Oh, the spi stuff was not supposed to be in there - needed it for testing the board (could not do it as an external component), and thought I had avoided checking it in but obviously slipped up! Thanks for the heads-up.

esphome/components/cst816/touchscreen/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    "CST816Touchscreen",
+    touchscreen.Touchscreen,
+    i2c.I2CDevice,
+)
+
+CST816ButtonListener = cst816_ns.class_("CST816ButtonListener")
+CONFIG_SCHEMA = touchscreen.TOUCHSCREEN_SCHEMA.extend(
+    {
+        cv.GenerateID(): cv.declare_id(CST816Touchscreen),
+        cv.Optional(CONF_INTERRUPT_PIN): pins.internal_gpio_input_pin_schema,
```
nielsnl68, line 20: ~~Why do you allow to set the `CONF_INTERRUPT_PIN` pin while you dont use/setup it in your component? ~~
Never mind, I missed a part of the code.

esphome/components/cst816/touchscreen/cst816_touchscreen.h
```diff
@@ -0,0 +1,124 @@
...
+      this->reset_pin_->digital_write(true);
+      delay(5);
+      this->reset_pin_->digital_write(false);
+      delay(5);
+      this->reset_pin_->digital_write(true);
+      this->set_timeout(30, [this] { this->continue_setup_(); });
+    } else {
+      this->continue_setup_();
+    }
+  }
```
nielsnl68, line 51: Dont forget to set the
```c
    this->x_raw_min_ = 0;
    this->y_raw_min_ = 0;
    this->x_raw_max_ = [device_x_max];
    this->y_raw_max_ = [device_y_max];
```
This is needed to calculate the right display position.

nielsnl68 (changes requested):

esphome/components/cst816/touchscreen/cst816_touchscreen.h
```diff
@@ -0,0 +1,124 @@
...
+#include "esphome/core/component.h"
+#include "esphome/core/hal.h"
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace cst816 {
+
+static const char *const TAG = "cst816.touchscreen";
+
+static const uint8_t CST816_REG_STATUS = 0x00;
```
guillempages: Since all those constants are in the cst816 namespace anyway; is it really necessary to prefix all them with "CST816_"?
(Maybe it is done this way on all other components; just asking, since it looks redundant to me)

esphome/components/cst816/touchscreen/cst816_touchscreen.h
```diff
@@ -0,0 +1,124 @@
...
+    uint8_t num_of_touches = data[CST816_REG_TOUCH_NUM] & 3;
+    if (num_of_touches == 0) {
+      this->update_button_state_(false);
+      return;
+    }
+
+    uint16_t x = encode_uint16(data[CST816_REG_XPOS_HIGH] & 0xF, data[CST816_REG_XPOS_LOW]);
+    uint16_t y = encode_uint16(data[CST816_REG_YPOS_HIGH] & 0xF, data[CST816_REG_YPOS_LOW]);
+    esph_log_v(TAG, "Read touch %d/%d", x, y);
+    if (x > this->x_raw_max_) {
```
guillempages: This might not work if the display is rotated by 90/270 degrees; then x and y are swapped, aren't they?

esphome/components/cst816/touchscreen/cst816_touchscreen.h
```diff
@@ -0,0 +1,124 @@
...
+      this->reset_pin_->digital_write(false);
+      delay(5);
+      this->reset_pin_->digital_write(true);
+      this->set_timeout(30, [this] { this->continue_setup_(); });
+    } else {
+      this->continue_setup_();
+    }
+  }
+
+  void update_touches() override {
```
guillempages: Shouldn't the definition be in the cpp file?

esphome/components/cst816/touchscreen/cst816_touchscreen.h
```diff
@@ -0,0 +1,124 @@
...
+
+class CST816ButtonListener {
+ public:
+  virtual void update_button(bool state) = 0;
+};
+
+class CST816Touchscreen : public touchscreen::Touchscreen, public i2c::I2CDevice {
+ public:
+  void setup() override {
+    esph_log_config(TAG, "Setting up CST816 Touchscreen...");
```
guillempages: I thought the convention was to only put oneliners in the header file, but put more complex methods in the cpp file?

esphome/components/cst816/touchscreen/cst816_touchscreen.h
```diff
@@ -0,0 +1,124 @@
...
+
+  void register_button_listener(CST816ButtonListener *listener) { this->button_listeners_.push_back(listener); }
+  void dump_config() override;
+
+  void set_interrupt_pin(InternalGPIOPin *pin) { this->interrupt_pin_ = pin; }
+  void set_reset_pin(GPIOPin *pin) { this->reset_pin_ = pin; }
+
+ protected:
+  void continue_setup_() {
+    if (this->interrupt_pin_ != nullptr) {
```
guillempages: Shouldn't the definition be in the cpp file?

esphome/components/cst816/touchscreen/cst816_touchscreen.h
clydebarrow: Since these are not used outside the component, it's not a big deal either way IMHO.

esphome/components/cst816/touchscreen/cst816_touchscreen.h
clydebarrow: This is dealing with raw coordinates. Rotation is handled later.

esphome/components/cst816/touchscreen/cst816_touchscreen.h
clydebarrow: Not my convention. Keeping as much as possible in the .h file goes some way to alleviate the pain of C++. In the past this may have been undesirable due to implementation considerations, but not with C++17.

esphome/components/cst816/touchscreen/cst816_touchscreen.h
guillempages: Well, I'm dealing with the issue that if I set the *display* rotation to 90, the width and height are swapped, and I'm getting very funny results for the y coordinate. And since in the rotation case, the display width and height are also swapped, I thought this might be relevant here.

I'm still trying to find out where the actual problem is; just wanted to mention it here, in case this is actually causing the issue.

esphome/components/cst816/touchscreen/cst816_touchscreen.h
guillempages: I thought that was the ESPHome convention (at least I got some comments in that direction in a PR IIRC), but I'm not sure about that. I just wanted to point this out, but I don't have a saying on it anyway ;-)

esphome/components/cst816/touchscreen/cst816_touchscreen.h
guillempages: I also don't think it is a big deal; I just don't like redundancy, but it's your code 😉

esphome/components/cst816/touchscreen/cst816_touchscreen.h
clydebarrow: The `rotation` setting on the display is using software rotation but some displays also support hardware rotation using `transform`. The touchscreen also has `transform` available (though it's done in software!)  and that should be used to match the final display orientation. It may be that the display and touchscreen need different settings for those transforms depending on various factors.

It may also be that there is some bug in the new touchscreen code, but I was able to successfully test all orientations.

esphome/components/cst816/touchscreen/cst816_touchscreen.h
clydebarrow: It's a fair point. I won't be touching this again till the new year as I'm on holiday right now.

esphome/components/cst816/touchscreen/cst816_touchscreen.h
```diff
@@ -0,0 +1,124 @@
...
+      case CST816T_CHIP_ID:
+        break;
+      default:
+        this->mark_failed();
+        esph_log_e(TAG, "Unknown chip ID 0x%02X", this->chip_id_);
+        return;
+    }
+    this->write_byte(CST816_REG_IRQ_CTL, CST816_IRQ_EN_MOTION);
+    this->x_raw_max_ = this->display_->get_width();
+    this->y_raw_max_ = this->display_->get_height();
```
guillempages: I think I found the issue with the funny coordinate conversion on rotation:
When using software rotation, the width and height of the display are swapped (assuming 90° or 270° rotation).
This means that the "get_width()" and "get_height()" here also return the swapped coordinates. The raw x and y values returned by the touchscreen are not yet swapped though, so that in that case one of the coordinates gets very easily "out of range".

I think an example will make this clearer:
Standard rotation: width = 170, height = 320 (standard is portrait)
If using (software-) `rotation: 90`, `display->get_width()` returns 320 and `display->get_height()` returns 170 (as expected).
When touching the lower right coordinate (i.e. (320,170) in "display coordinates), the touchscreen chip reports (0, 320).
In the normalize function in the base touchscreen, the y coordinate is then capped to 170, since the reported 320 is greater than the reported display height (170).

I'm not sure how this should be handled... Should the "get_width" and "get_height" calls be swapped if the SW rotation is active? Should display offer additional `get_raw_height()` and `get_raw_width()` methods that are rotation-independent?

guillempages:
@clydebarrow I know you are on holiday (saw that on discord), and did not actually expect you to answer my comments so promptly. I just thought I'd write my thoughts in the PR.
Enjoy your holiday!

esphome/components/cst816/touchscreen/cst816_touchscreen.h
nielsnl68: > I think I found the issue with the funny coordinate conversion on rotation: When using software rotation, the width and height of the display are swapped (assuming 90° or 270° rotation). This means that the "get_width()" and "get_height()" here also return the swapped coordinates. The raw x and y values returned by the touchscreen are not yet swapped though, so that in that case one of the coordinates gets very easily "out of range".

The "out of range" should not be happening. The conversion formule should be handling this from my expectation of the formule.
When this is not the case then please ping my on discord and help me debug it again.

clydebarrow: > Should display offer additional `get_raw_height()` and `get_raw_width()`

Something like that. Other touchscreen chips have registers to read that provide the raw dimensions, but the registers in the CST816 datasheet did not return sensible values when I tested that, which is odd since the chip must have those values configured, but I could not make it work.

The other option (that I think I prefer) is to allow the dimensions to be specified explicitly in the touchscreen config. This also allows for the case (not sure how often it would occur) when the display and touchscreen have different dimensions.

nielsnl68: My preferred option is setting the dimensions in YAML, because they could not be the same as the display dimensions. We should introduce the calibration settings as used in the **XPT2024** component to all existing touchscreen drivers. When not set in YAML then there can be a failback set in the `setup()` when available. 
And I think we should flag touchscreen component as failed when the touchscreen calibration settings are not set before checking for touches. Something i need to implement. 
One future option could be to add a calibration page that show in sequence dot's in all four corners. And calculate the min and maximal touch dimensions.

guillempages: I think there should be the *possibility* to set the calibration options in yaml, but the normal use case should be IMHO that the touchscreen and the display match, so if it is possible to calibrate automatically, we should not force the user to do that in yaml.

nielsnl68: > I think there should be the _possibility_ to set the calibration options in yaml, but the normal use case should be IMHO that the touchscreen and the display match, so if it is possible to calibrate automatically, we should not force the user to do that in yaml.

i agree.

clydebarrow: I wasn't suggesting the dimensions be required but that it would work like this:

* If the driver can establish the raw coordinates itself (e.g. by reading the chip config registers) it does so;
* Otherwise the display is queried for the *raw* height and width and those are used;
* If the user configures specific raw min/max values these override the previous.

Currently the display driver doesn't expose the un-rotated height and width (`DisplayBuffer` does, but that needs to be pulled up to `Display` with a default implementation.)

dirkjankrijnders:
Hi, I'm working on a pull request for the cst820 #5641 which is very close to the cst816s. Maybe this can be one component?

clydebarrow:
> Hi, I'm working on a pull request for the cst820 #5641 which is very close to the cst816s. Maybe this can be one component?

Yes looks like one component would cover both. Will look at it more after Xmas.

bearpawmaxim:
Hello @clydebarrow !
Trying to use this driver for my CST816D touchscreen but it isn't working at all, while [this one](https://github.com/GadgetFactory/CST816S_touchscreen/tree/main) is working but is ugly...

Config:
```
touchscreen:
  platform: cst816
  interrupt_pin: GPIO0
  reset_pin: GPIO1 

binary_sensor:
  - platform: touchscreen
    name: Power on button
    id: power_on_button
    x_min: 20
    x_max: 20
    y_min: 220
    y_max: 220
    on_click: 
      ....
```

Log:
```
[14:04:04][I][i2c.arduino:069]: Results from i2c bus scan:
[14:04:04][I][i2c.arduino:075]: Found i2c device at address 0x15
.....
[14:07:51][C][cst816.touchscreen:007]: CST816 Touchscreen:
[14:07:51][C][cst816.touchscreen:008]:   Address: 0x15
[14:07:51][C][cst816.touchscreen:009]:   Interrupt Pin: GPIO0
[14:07:51][C][cst816.touchscreen:010]:   Reset Pin: GPIO1
[14:07:51][C][cst816.touchscreen:026]:   Chip type: Unknown
```

The board is [ESP32-2424S012](https://www.tindie.com/products/adz1122/esp32-c3-arduino-lvgl-128-inch-lcd-smart-display/)

nielsnl68:
could you add:
```yaml
touchscreen:
  platform: cst816
  interrupt_pin: GPIO0
  reset_pin: GPIO1 
  on_touch:
    - lambda: |-
              touch.x,
              touch.y,
              touch.x_raw,
              touch.y_raw
              );
```

And see of the touches are working as expected.

bearpawmaxim:
> And see

I tried this before asking and got nothing. Looks like an interrupt isn't working..
One more thing to note is the line in log about unknown chip type.
Thanks!

clydebarrow:
can you turn on VERBOSE logging and see what chip type it reports?

bearpawmaxim:
> can you turn on VERBOSE logging and see what chip type it reports?

No, but I've read the chip id manually with the lambda:
```
- lambda: |-
          uint8_t chip_id_;
          bool s = id(cst816d)->read_byte(0xA7, &chip_id_);
          ESP_LOGI("CST", "CST Chip id is %d (success %d)", chip_id_, s);
```

The chip id is 0xB6, which isn't supported by your driver... Will try to clone, add this id manually and test

bearpawmaxim:
> Will try to clone, add this id manually and test

Added CST_816D_CHIP_ID in my branch [here](https://github.com/bearpawmaxim/esphome/commit/45077a44a87df0bf03b3f9cdeb549ce5e1c235bf)

config:
```
external_components:
  - source: github://bearpawmaxim/esphome@pr5941fix
    components: [cst816]

touchscreen:
  id: cst816d
  platform: cst816
  interrupt_pin: GPIO0
  reset_pin: GPIO1
  on_touch:
    - lambda: |-
            touch.x,
            touch.y,
            touch.x_raw,
            touch.y_raw
        );

binary_sensor:
  - platform: touchscreen
    name: Power on button
    id: power_on_button
    x_min: 20
    x_max: 20
    y_min: 220
    y_max: 220
#     page_id: power_on_btn_page
    on_click:
      - lambda: |-
        ESP_LOGI("cal", "Binary touch sensor");
```

logs:
```
...
[13:39:10][I][cal:171]: x=116, y=83, x_raw=117, y_raw=84
[13:39:11][I][cal:171]: x=108, y=72, x_raw=109, y_raw=73
[13:39:11][I][cal:171]: x=95, y=76, x_raw=96, y_raw=77
[13:39:11][I][cal:171]: x=96, y=80, x_raw=97, y_raw=81
[13:39:11][I][cal:171]: x=102, y=83, x_raw=103, y_raw=84
[13:39:12][I][cal:171]: x=107, y=81, x_raw=108, y_raw=82
[13:39:12][I][cal:171]: x=105, y=84, x_raw=106, y_raw=85
[13:39:12][I][cal:171]: x=103, y=101, x_raw=104, y_raw=102
[13:39:12][I][cal:171]: x=97, y=89, x_raw=98, y_raw=90
[13:39:12][I][cal:171]: x=90, y=87, x_raw=91, y_raw=88
[13:39:13][I][cal:171]: x=81, y=88, x_raw=82, y_raw=89
[13:39:13][I][cal:171]: x=85, y=94, x_raw=86, y_raw=95
[13:39:13][I][cal:171]: x=92, y=80, x_raw=93, y_raw=81
[13:39:13][I][cal:171]: x=113, y=110, x_raw=114, y_raw=111
....
```

Touch now works, but binary sensor doesn't

nielsnl68:
@bearpawmaxim could you do me a favor and press all for corners and show the result?
thanks

bearpawmaxim:
> press all for corners and show the result?

Well, it is round, so touch sequence is the following (from 1 to 5):



```
[13:54:38][I][cal:140]: x=111, y=111, x_raw=112, y_raw=112
[13:54:39][I][cal:140]: x=138, y=218, x_raw=139, y_raw=219
[13:54:40][I][cal:140]: x=3, y=100, x_raw=4, y_raw=101
[13:54:42][I][cal:140]: x=152, y=5, x_raw=153, y_raw=6
[13:54:43][I][cal:140]: x=237, y=109, x_raw=238, y_raw=110
```
Update: looks like the touchscreen zero is located in the upper right corner.

P.S.: no rotation is applied to the display

clydebarrow:
> Touch now works, but binary sensor doesn't

Does that display have a binary sensor? Where would it be? IIRC the 816 implements it by reporting a touch outside the screen boundary.

alantreadway:
Hi - I found this PR whilst trying to get a board with the designation `ESP32-2432S024` up and working with ESPHome.

I followed the thread to try out this PR, and was hitting `Chip type: Unknown` - using the lambda provided by bearpawmaxim to log the chip id, I'm seeing: `0xB7` - not sure if this works (will try to fork and tweak now to confirm).

clydebarrow:
> to log the chip id, I'm seeing: `0xB7`

I have added that as CST820. Try it and see.

bearpawmaxim:
> Does that display have a binary sensor?

Ah, I now understand what the binary sensor means.. I mistakenly thought that binary sensor should trigger when you touch the screen area and that touch lies within the binary's sensor coordinates.
Now it is clear for me.

@clydebarrow one more question.
Because my "chinese miracle" have touchscreen and display zero coordinates mismatch (it is rotated 90° clockwise).. Can't you provide the rotation ability?

clydebarrow:
> Can't you provide the rotation ability?

https://deploy-preview-3487--esphome.netlify.app/components/touchscreen/

To rotate 90 or 270 you will need swap_xy and one of the mirror settings.

alantreadway:
> > to log the chip id, I'm seeing: `0xB7`
> 
> I have added that as CST820. Try it and see.

Thanks - I'm seeing `Chip type: CST816S` in the logs now, but that is the last message I see (tried with two different boards to rule out hardware failure) - not sure if I'm configuring things correctly, so it's unclear what is going on.

I don't really have time to dig into this further now, but when I do I can raise a PR if further changes are needed?

pixelwave:
I have the **ESP32-2432S024C** with **CST820** touch controller as well and trying to get the touch functionality working. I got the screen displaying stuff with this code:

```
color:
  - id: my_red
    red: 100%
    green: 3%
    blue: 5%

spi:
 clk_pin: GPIO14
 mosi_pin: GPIO13
 miso_pin: GPIO12

display:
  - platform: ili9xxx
    model: TFT 2.4
    cs_pin: GPIO15
    dc_pin: GPIO2
    rotation: 90
    id: tft_ha
    lambda: |-
      it.fill(my_red);

output:
  - platform: ledc
    pin: 27
    id: gpio_27_backlight_pwm

light:
  - platform: monochromatic
    output: gpio_27_backlight_pwm
    name: "Display Backlight"
    id: back_light
    restore_mode: ALWAYS_ON
``` 

How do I get the touch now working / what files do I have to include in my ESPHome build? Sorry I am still learning ESPHome ... quite new to the syntax. If somebody could give me support that would be awesome.

jesserockz (approved):

PR #5945, opened 2023-12-15
i2s_audio: Set player_task's prio to 1

gaaf:

Prio 0 is the lowest possible, competing with the idle task. All non-idle tasks should have a prio > 0. If there's another task at prio > 0  around (like the application's loop task), it is possible the framebuffere task never gets executed.

Perhaps the prio should even be above 1, to keep avoid missing audio frames.

This came to attention during review of eliminating the delay in the application's main loop in case the loop is already slow (https://github.com/esphome/esphome/pull/5869).

Code is untested due to lack of hardware.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
+++ b/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
@@ -29,7 +29,7 @@ void I2SAudioSpeaker::start_() {
   }
   this->state_ = speaker::STATE_RUNNING;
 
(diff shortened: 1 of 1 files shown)


jesserockz (approved):
Working fine on Atom Echo Voice Assistant firmware

PR #5947, opened 2023-12-16
Fix I2CBus::write() bug and add i2c documentation

DrCoolzic:
I fixed a bug where the buffer length was incorrectly specified as a uin8_t instead of a size_t.
I have also added a lot of documentation to the i2c library, as well as aliased ERROR_OK to NO_ERROR

File affected:
- esphome/components/i2c/i2c.h
- esphome/components/i2c/i2c_bus.h

- This PR fixes the problem of not being able to write buffer of 256 or more bytes when using the i2c bus (this affect the ESP-IDF framework as well as the Arduino framework.
- at the same time, I took the opportunity to document the I²C library usage
- and I have added (and not replaced so it does not break the I/F) an alias to ERROR_OK

The only code changes are:
```c++
ErrorCode write(const uint8_t *data, uint8_t len, bool stop = true) { return bus_->write(address_, data, len, stop); }
```
As been replaced by
```c++
ErrorCode write(const uint8_t *data, size_t len, bool stop = true) { return bus_->write(address_, data, len, stop); }
```
Plus the addition of the alias NO_ERROR in i2c_bus.h.

As far as the documentation is concerned, I've added information on classes, methods and attributes to help users of the I2C library. Please note that my native language is not English, so it's likely that the text contains errors. Please feel free to correct them.
Note also that I have used a syntax that allow detailed documentation in Doxygen output when looking the API, but that is hidden (only the details are hidden) when hovering in intellisense.

# What does this PR do not fix?
Unfortunately, this PR does not fix the problem of not being able to read messages with 256 or bytes when using the Arduino framework is used. Fixing this problem requires to modify the Wire.h file in the Arduino framework and I do not know how to do that (or even if it is possible). However, as a courtesy I provide the code to do so in case you know how to modify Arduino framework.

In Wire.h the following code:
```c++
uint8_t TwoWire::requestFrom(int address, int len) {
    return requestFrom(static_cast<uint16_t>(address), static_cast<size_t>(len), true);
}
uint8_t TwoWire::requestFrom(int address, int len, int sendStop) {
    return static_cast<uint8_t>(requestFrom(static_cast<uint16_t>(address), static_cast<size_t>(len), static_cast<bool>(sendStop)));
}
```

Needs to be replaced by
```c++
size_t TwoWire::requestFrom(int address, int len) {
    return requestFrom(static_cast<uint16_t>(address), static_cast<size_t>(len), true);
}
size_t TwoWire::requestFrom(int address, int len, int sendStop) {
    return requestFrom(static_cast<uint16_t>(address), static_cast<size_t>(len), static_cast<bool>(sendStop));
}
```

# PR Summary
When integrated this fix will allow to use buffers of 256 or more bytes in i2c library when using the ESP-IDF framework.
Unfortunately for the Arduino framework the buffer size will still be limited to 255 until the framework is also fixed.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: test-wk2132-i2c-arduino
  platformio_options:
    build_flags:
      - -DI2C_BUFFER_LENGTH=257
esp32:
  board: esp32dev
  framework:
    type: arduino
…
external_components:
  - source: github://pr#5932
    components: [uart]
    refresh: 0s
  - source: github://pr#5218
    components: [wk2132_i2c]
    refresh: 0s

i2c:
  sda: 21
  scl: 22
  scan: true
  id: bus_i2c
  frequency: 600kHz
```
## more info in this document
[Bus i2c problems and solutions en-us.pdf](https://github.com/esphome/esphome/files/13692916/Bus.i2c.problemes.et.solutions.en.pdf)

## Checklist:

### DIFF
--- a/esphome/components/i2c/i2c_bus.h
+++ b/esphome/components/i2c/i2c_bus.h
@@ -59,9 +59,9 @@ class I2CBus {
   /// @details This is a pure virtual method that must be implemented in a subclass.
   virtual ErrorCode readv(uint8_t address, ReadBuffer *buffers, size_t count) = 0;
 
-  // virtual ErrorCode write(uint8_t address, const uint8_t *buffer, size_t len) {
-  //   return write(address, buffer, len, true);
-  // }
+  virtual ErrorCode write(uint8_t address, const uint8_t *buffer, size_t len) {
+    return write(address, buffer, len, true);
+  }
 
   /// @brief Creates a WriteBuffer and calls the writev() method to send the bytes from this buffer
   /// @param address address of the I²C component on the i2c bus
@@ -70,16 +70,16 @@ class I2CBus {
   /// @param stop true or false: True will send a stop message, releasing the bus after
   /// transmission. False will send a restart, keeping the connection active.
   /// @return an i2c::ErrorCode
-  virtual ErrorCode write(uint8_t address, const uint8_t *buffer, size_t length, bool stop = true) {
+  virtual ErrorCode write(uint8_t address, const uint8_t *buffer, size_t length, bool stop) {
     WriteBuffer buf;
     buf.data = buffer;
     buf.len = length;
     return writev(address, &buf, 1, stop);
   }
 
-  // virtual ErrorCode writev(uint8_t address, WriteBuffer *buffers, size_t cnt) {
-  //   return writev(address, buffers, cnt, true);
-  // }
+  virtual ErrorCode writev(uint8_t address, WriteBuffer *buffers, size_t cnt) {
+    return writev(address, buffers, cnt, true);
+  }
 
   /// @brief This virtual method writes bytes to an I2CBus from an array of WriteBuffer.
   /// @param address address of the I²C component on the i2c bus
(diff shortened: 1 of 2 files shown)


esphome/components/i2c/i2c_bus.h
```diff
@@ -7,50 +7,93 @@
...
-  virtual ErrorCode write(uint8_t address, const uint8_t *buffer, size_t len, bool stop) {
+
+  /// @brief Creates a WriteBuffer and calls the writev() method to send the bytes from this buffer
+  /// @param address address of the I²C component on the i2c bus
+  /// @param buffer pointer to an array of bytes that contains the data to be sent
+  /// @param length length of the buffer = number of bytes to write
+  /// @param stop true or false: True will send a stop message, releasing the bus after
+  /// transmission. False will send a restart, keeping the connection active.
+  /// @return an i2c::ErrorCode
+  virtual ErrorCode write(uint8_t address, const uint8_t *buffer, size_t length, bool stop) {
```
jesserockz: For consistency, please change all `len` to `length` or leave all as `len`

jesserockz (changes requested):

esphome/components/i2c/i2c_bus.h
DrCoolzic: Chaged `length `to `len`. As it will be more consistent with read and write buffer that we can't change for compatibility reason.
Now len is used for i2c.h and i2c_bus.h

DrCoolzic:
Is there a way to fix the bug in Wire.h of the Arduino library?

jesserockz (approved):
Thanks @DrCoolzic

jesserockz:
> Is there a way to fix the bug in Wire.h of the Arduino library?

The Wire.h code is part of the Arduino Framework and would need to be changed there (If it is not already in a newer Arduino version).

PR #5952, opened 2023-12-17
Support for ST7567 display 128x64 (I2C, SPI)

latonita:

Add support for ST7567-based COG LCD displays.

## Types of changes

- [ ] Other

https://github.com/esphome/feature-requests/issues/2203

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3482

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
i2c:
  sda: D2
  scl: D1
  frequency: 400kHz
display:
  - platform: st7567_i2c
    address: 0x3f
    flip_x: false
    invert: false
    update_interval: 1s
    lambda: |-
      it.line(0,0,10,0); it.line(0,0,0,10); 
      it.line(127,63,117,63); it.line(127,63,127,53); 

```

## Checklist:

### DIFF
--- a/esphome/components/st7567_base/__init__.py
+++ b/esphome/components/st7567_base/__init__.py
@@ -14,14 +14,15 @@ st7567_base_ns = cg.esphome_ns.namespace("st7567_base")
 ST7567 = st7567_base_ns.class_("ST7567", cg.PollingComponent, display.DisplayBuffer)
 ST7567Model = st7567_base_ns.enum("ST7567Model")
 
-CONF_FLIP_X = "flip_x"
-CONF_FLIP_Y = "flip_y"
+# to reuse MIRROR constants from const.py when they are released
+CONF_MIRROR_X = "mirror_x"
+CONF_MIRROR_Y = "mirror_y"
 
 ST7567_SCHEMA = display.FULL_DISPLAY_SCHEMA.extend(
     {
         cv.Optional(CONF_RESET_PIN): pins.gpio_output_pin_schema,
-        cv.Optional(CONF_FLIP_X, default=False): cv.boolean,
-        cv.Optional(CONF_FLIP_Y, default=False): cv.boolean,
+        cv.Optional(CONF_MIRROR_X, default=False): cv.boolean,
+        cv.Optional(CONF_MIRROR_Y, default=False): cv.boolean,
         cv.Optional(CONF_INVERT, default=False): cv.boolean,
     }
 ).extend(cv.polling_component_schema("1s"))
@@ -33,10 +34,10 @@ async def setup_st7567(var, config):
     if CONF_RESET_PIN in config:
         reset = await cg.gpio_pin_expression(config[CONF_RESET_PIN])
         cg.add(var.set_reset_pin(reset))
-    if CONF_FLIP_X in config:
-        cg.add(var.init_flip_x(config[CONF_FLIP_X]))
-    if CONF_FLIP_Y in config:
-        cg.add(var.init_flip_y(config[CONF_FLIP_Y]))
+    if CONF_MIRROR_X in config:
+        cg.add(var.init_mirror_x(config[CONF_MIRROR_X]))
+    if CONF_MIRROR_Y in config:
+        cg.add(var.init_mirror_y(config[CONF_MIRROR_Y]))
     if CONF_INVERT in config:
         cg.add(var.init_invert(config[CONF_INVERT]))
     if CONF_LAMBDA in config:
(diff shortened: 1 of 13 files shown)


esphome/components/st7567_base/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+)
+
+CODEOWNERS = ["@latonita"]
+
+st7567_base_ns = cg.esphome_ns.namespace("st7567_base")
+ST7567 = st7567_base_ns.class_("ST7567", cg.PollingComponent, display.DisplayBuffer)
+ST7567Model = st7567_base_ns.enum("ST7567Model")
+
+CONF_FLIP_X = "flip_x"
+CONF_FLIP_Y = "flip_y"
```
guillempages: If I am not mistaken, other displays (at least the ILI9xxx which supports a lot of displays) use "mirror_x" and "mirror_y" instead (and have a "swap_x_y" as well, to support 90° rotation), and have those configurations inside a "transform:" part.
IMHO it would be nice if the config options are kept consistent between separate display components.
(I'm not claiming that they are consistent now; just saying it would be nice ;-)). Maybe @clydebarrow has additional input on that.

clydebarrow: Yes, `mirror_x` and y are used elsewhere, so should stay consistent. Those constants are in `core/const.py`.

latonita: I checked source code earlier - I took SSD1306 as a base - it has flip_x / flip_y
flip_x - used by ssd1306, max7219
mirror_x - used by ili9xxx, touchscreen

so the score is 2:2
It might be nice to select one approach and do a proper update.

latonita: @guillempages one more thing - ili9xxxx defines mirror_x in its .py file

@guillempages , @clydebarrow I actually think we might actually unify codebase and use `mirror`. on second though it might better describe the action.

latonita: > Yes, `mirror_x` and y are used elsewhere, so should stay consistent. Those constants are in `core/const.py`.

hm... It might be too late in the place where I live, but I can't find any mentions of "mirror" in any *.py files, except for manually defined for ili9xx

clydebarrow: Ah, I obviously spoke too soon. Those constants *will* be in core code Real Soon Now. (the CI checks insist on that when used in 4 places.)

latonita: @clydebarrow I changed FLIP to MIRROR, updated docs as well.
@guillempages SWAP_X_Y is hardware feature of ILI9xxx, not possible to do on this hardware. Software ROTATION: 90 works great.

clydebarrow:
For consistency with the other displays best to use the same schema for mirroring. And what does `invert:` do? If it inverts the colors then it should be named accordingly. E.g.

```
            cv.Optional(CONF_INVERT_COLORS, default=False): cv.boolean,
            cv.Optional(CONF_TRANSFORM): cv.Schema(
                {
                    cv.Optional(CONF_MIRROR_X, default=False): cv.boolean,
                    cv.Optional(CONF_MIRROR_Y, default=False): cv.boolean,
                }
            ),
```

Also, if a default is specified for an `Optional` value, then it is guaranteed to be present in the config in `to_code` so no need to test for the key in the dictionary.

Otherwise looking good.

latonita:
> For consistency with the other displays best to use the same schema for mirroring. And what does `invert:` do? If it inverts the colors then it should be named accordingly. E.g.
> 
> ```
>             cv.Optional(CONF_INVERT_COLORS, default=False): cv.boolean,
>             cv.Optional(CONF_TRANSFORM): cv.Schema(
>                 {
>                     cv.Optional(CONF_MIRROR_X, default=False): cv.boolean,
>                     cv.Optional(CONF_MIRROR_Y, default=False): cv.boolean,
>                 }
>             ),
> ```
> 
> Also, if a default is specified for an `Optional` value, then it is guaranteed to be present in the config in `to_code` so no need to test for the key in the dictionary.
> 
> Otherwise looking good.

Fixed

clydebarrow (approved):
I'm approving this as is, there's just one thing I noticed that would be worth attending to later;

Sending a command before the component has started `setup()` will fail, and may even cause a crash.  I see that you have taken care of this with your intiialisation, by using different methods (e.g. `init_invert_colors()` rather than `set_invert_colors()` but it's still possible for someone writing lambdas to call those methods before the display component has been set up (this has occurred with other displays.)

So a more robust strategy is for those methods to set the value, then test if the component has been setup (by calling `is_ready()` and only if true actually doing the command. If not ready, the command will be sent when `setup()` is called.

Since it only affects user code in lambdas, I don't consider it a big enough issue to prevent merging now, but it is probably worth addressing at some time.

PR #5968, opened 2023-12-19
Add RTTTL volume control.

nielsnl68:
Add Gain option to set the volume.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3644

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example
```yaml
   rtttl:
      speaker: my_speaker
      id: my_rtttl
      gain: 0.8
```


jesserockz (commented):
I have not tested this yet, but should this be a config option like volume between 0-10 etc that maps to bigger values?

nielsnl68:
> I have not tested this yet, but should this be a config option like volume between 0-10 etc that maps to bigger values?

Yes, but i need to figure out what the high and low values should be.

jesserockz:
> Yes, but i need to figure out what the high and low values should be.

True...haha

edwardtfn:
Could this be like a `volume_gain` argument on the rtttl settings?

Something like:

```yaml
rtttl:
  output: rtttl_out
  id: my_rtttl
  gain: 2
```
Or:
```yaml
rtttl:
  output: rtttl_out
  id: my_rtttl
  gain: 0.5
```
So you multiply 8192 by the gain, which will be float, with `1` as default or not used and limited by something like `0` to `10` (I haven't tested).

My point is more that naming `gain` makes more sense than `volume` in this case.

nielsnl68:
> Could this be like a `volume_gain` argument on the rtttl settings?
> 
> Something like:
> 
> ```yaml
> rtttl:
>   output: rtttl_out
>   id: my_rtttl
>   gain: 2
> ```
> 
> Or:
> 
> ```yaml
> rtttl:
>   output: rtttl_out
>   id: my_rtttl
>   gain: 0.5
> ```
> 
> So you multiply 8192 by the gain, which will be float, with `1` as default or not used and limited by something like `0` to `10` (I haven't tested).
> 
> My point is more that naming `gain` makes more sense than `volume` in this case.

I think this is the way to go.

nielsnl68:
Okay @edwardtfn, @jesserockz  i have added the `gain` option it will be a value between 0.1 to 1. And will be default set to 0.5.

esphome/components/rtttl/__init__.py
```diff
@@ -38,6 +39,7 @@
             cv.GenerateID(CONF_ID): cv.declare_id(Rtttl),
             cv.Optional(CONF_OUTPUT): cv.use_id(FloatOutput),
             cv.Optional(CONF_SPEAKER): cv.use_id(Speaker),
+            cv.Optional(CONF_GAIN, default="0.5"): cv.float_range(0.1, 1),
```
jesserockz: I feel like this would be better as a `cv.percentage` which is a little more user friendly?

Forcing a user to enter a 0.x number as a config doesnt really make sense.

esphome/components/rtttl/rtttl.cpp
```diff
@@ -278,18 +278,23 @@ void Rtttl::loop() {
...
-    } else {
-      this->samples_per_wave_ = 0;
-    }
-    if (need_note_gap) {
-      this->samples_gap_ = (this->sample_rate_ * DOUBLE_NOTE_GAP_MS) / I2S_SPEED;
-    } else {
-      this->samples_gap_ = 0;
+
+      // make sure there is enough samples to add a full last sinus.
+      uint16_t devision = ((this->samples_count_ << 10) / this->samples_per_wave_) + 1;
```
jesserockz:
```suggestion
      uint16_t division = ((this->samples_count_ << 10) / this->samples_per_wave_) + 1;
```

esphome/components/rtttl/rtttl.cpp
```diff
@@ -278,18 +278,23 @@ void Rtttl::loop() {
...
+      uint16_t x = this->samples_count_;
+      this->samples_count_ = (devision * this->samples_per_wave_);
+      ESP_LOGD(TAG, "play time old: %d div: %d new: %d %d", x, devision, this->samples_count_, this->samples_per_wave_);
```
jesserockz:
```suggestion
      this->samples_count_ = (division * this->samples_per_wave_);
      ESP_LOGD(TAG, "play time old: %d div: %d new: %d %d", x, division, this->samples_count_, this->samples_per_wave_);
```

esphome/components/rtttl/rtttl.h
```diff
@@ -60,6 +67,7 @@ class Rtttl : public Component {
   uint16_t note_duration_;
 
   uint32_t output_freq_;
+  double gain_{0.5};
```
jesserockz: We try to use floats instead of doubles
```suggestion
  float gain_{0.5f};
```

esphome/components/rtttl/rtttl.h
```diff
@@ -31,6 +31,13 @@ class Rtttl : public Component {
 #ifdef USE_SPEAKER
   void set_speaker(speaker::Speaker *speaker) { this->speaker_ = speaker; }
 #endif
+  void set_gain(double gain) {
+    if (gain < 0.1)
+      gain = 0.1;
+    if (gain > 1)
+      gain = 1;
+    this->gain_ = gain;
+  }
```
jesserockz: This could also use `clamp` from `helpers.h`
```suggestion
  void set_gain(float gain) {
    if (gain < 0.1f)
      gain = 0.1f;
    if (gain > 1.0f)
      gain = 1.0f;
    this->gain_ = gain;
  }
```

jesserockz (changes requested):

esphome/components/rtttl/__init__.py
nielsnl68:
```suggestion
            cv.Optional(CONF_GAIN, default="0.6"): cv.percentage,
```

esphome/components/rtttl/rtttl.cpp
```diff
@@ -254,7 +254,7 @@ void Rtttl::loop() {
     // Add small silence gap between same note
     this->output_freq_ = freq;
 
-    ESP_LOGVV(TAG, "playing note: %d for %dms", note, this->note_duration_);
+    ESP_LOGI(TAG, "playing note: %d for %dms", note, this->note_duration_);
```
jesserockz: Should this log level be reverted?

esphome/components/rtttl/__init__.py
```diff
@@ -38,6 +39,7 @@
             cv.GenerateID(CONF_ID): cv.declare_id(Rtttl),
             cv.Optional(CONF_OUTPUT): cv.use_id(FloatOutput),
             cv.Optional(CONF_SPEAKER): cv.use_id(Speaker),
+            cv.Optional(CONF_GAIN, default="0.6"): cv.percentage,
```
jesserockz, line 42: The default here should at east match what you have in the header file: https://github.com/esphome/esphome/pull/5968/files#diff-606929f046de6c81e59c42cfe6bdadfb235f8ea671634c1cec4090ec3dafed95R70

jesserockz (changes requested):
A couple of things below, and also can you please update the PR title and description and also add a docs PR for the new config?

Cheers
Jesse

esphome/components/rtttl/rtttl.h
```diff
@@ -60,6 +67,7 @@ class Rtttl : public Component {
   uint16_t note_duration_;
 
   uint32_t output_freq_;
+  float gain_{0.5f};
```
nielsnl68:
```suggestion
  float gain_{0.6f};
```

esphome/components/rtttl/__init__.py
nielsnl68: Done, by updating the header file

esphome/components/rtttl/rtttl.cpp
nielsnl68:
```suggestion
    ESP_LOGVV(TAG, "playing note: %d for %dms", note, this->note_duration_);
```

nielsnl68:
Thanks for the review @jesserockz. i resolved the issues you reported

nielsnl68:
> A couple of things below, and also can you please update the PR title and description and also add a docs PR for the new config?
> 
> Cheers Jesse

I have updated the documentation now.

esphome/components/rtttl/__init__.py
```diff
@@ -98,6 +100,9 @@ async def to_code(config):
         out = await cg.get_variable(config[CONF_SPEAKER])
         cg.add(var.set_speaker(out))
 
+    if CONF_GAIN in config:
+        cg.add(var.set_gain(config[CONF_GAIN]))
```
jesserockz:
```suggestion
    cg.add(var.set_gain(config[CONF_GAIN]))
```

jesserockz (changes requested):

jesserockz (approved):

PR #5970, opened 2023-12-20
Adding support for RDS-RGBW-02 data-over-power lights

smithjacobj:

This adds support for the RDS-RGBW-02 data-over-power lights and potentially further DOP implementations down the road.

This required significant re-architecting as the existing architecture was built under the assumption one would be using 1- or 2-wire strings that use simple RGB(W) arrays and shift subsequent data down the line. The RDS-RDBW-02 and other data-over-power schemes require in-band addressing and potentially other, more complex structuring, possibly application-specific.

To enable this, the transformation of the color buffer to `rmt_item32_t`s is facilitated by the `Generator` functor (e.g. lambda). This lets the component host swappable transformation code and can potentially allow a user to implement their own schema in a config-included header file if they so desire.

**MAINTAINER NOTE** I do not currently have any of the previously-supported addressable lights (e.g. WS2811/2), therefore I cannot properly test that I maintained backwards-compatibility. If I could have a volunteer test this code, that would be very helpful!

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here> **TODO**

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
platform: esp32_rmt_led_strip
chipset: RDS-RGBW-02
num_leds: 15
rgb_order: RGB
default_transition_length: 0s
effects:
  - addressable_rainbow:
      speed: 10
      width: 15
  - addressable_lambda:
      name: Candy Cane
      update_interval: 1h
      lambda: static_colors(it, {RED, WHITE});
  - addressable_lambda:
      name: Red Green
      update_interval: 1h
      lambda: static_colors(it, {RED, GREEN});
  - addressable_lambda:
      name: Five Color
      update_interval: 1h
      lambda: static_colors(it, {RED, GREEN, BLUE, YELLOW, PURPLE});
  - addressable_lambda:
      name: Seahawks
      update_interval: 1h
      lambda: static_colors(it, {GREEN, BLUE});
  - addressable_rainbow:
      name: ARE GEE BEE
      speed: 10
      width: 150
```

## Checklist:

### DIFF
--- a/esphome/components/esp32_rmt_led_strip/led_strip.cpp
+++ b/esphome/components/esp32_rmt_led_strip/led_strip.cpp
@@ -1,5 +1,6 @@
 #include <cinttypes>
 #include "led_strip.h"
+#include "rds_rgbw_02.h"
 
 #ifdef USE_ESP32
 
@@ -13,7 +14,12 @@ namespace esp32_rmt_led_strip {
 
 static const char *const TAG = "esp32_rmt_led_strip";
 
-static const uint8_t RMT_CLK_DIV = 2;
+static constexpr uint8_t RMT_CLK_DIV = 2;
+static constexpr float CLK_RATIO = (float) APB_CLK_FREQ / RMT_CLK_DIV / 1e09f;
+
+// 15 bits of duration per each high/low per item
+static constexpr uint32_t MAX_CYCLES_PER_DURATION = std::numeric_limits<uint16_t>::max() >> 1;
+static constexpr uint32_t MAX_CYCLES_PER_ITEM = MAX_CYCLES_PER_DURATION * 2;
 
 void ESP32RMTLEDStripLightOutput::setup() {
   ESP_LOGCONFIG(TAG, "Setting up ESP32 LED Strip...");
@@ -21,8 +27,12 @@ void ESP32RMTLEDStripLightOutput::setup() {
   size_t buffer_size = this->get_buffer_size_();
 
   ExternalRAMAllocator<uint8_t> allocator(ExternalRAMAllocator<uint8_t>::ALLOW_FAILURE);
-  this->buf_ = allocator.allocate(buffer_size);
-  if (this->buf_ == nullptr) {
+  this->buf_[0] = allocator.allocate(buffer_size);
+  if (this->allow_partial_updates_) {
+    // allocate a second buffer for swap->delta comparison
+    this->buf_[1] = allocator.allocate(buffer_size);
+  }
+  if (this->buf_[0] == nullptr || (this->allow_partial_updates_ && this->buf_[1] == nullptr)) {
     ESP_LOGE(TAG, "Cannot allocate LED buffer!");
     this->mark_failed();
     return;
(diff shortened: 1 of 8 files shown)


esphome/components/esp32_rmt_led_strip/led_strip.cpp
```diff
@@ -99,28 +111,42 @@ void ESP32RMTLEDStripLightOutput::write_state(light::LightState *state) {
...
-    psrc++;
+
+    if (this->intermission_ > 0) {
+      if (rmt_write_items(this->channel_, this->rmt_buf_, command_len, true) != ESP_OK) {
+        ESP_LOGE(TAG, "RMT TX error");
+        this->status_set_warning();
+        return;
+      }
+      dest = this->rmt_buf_;
+      delayMicroseconds(this->intermission_);
```
smithjacobj: I don't like this so I'm open to feedback on how this could be better implemented in the ESPHome architecture.

smithjacobj: After some discussion with @ssieb I'm thinking this is going to be the solution for now. Trying to let it do a full esphome loop while waiting for the next light is a 15+ millisecond delay. We want the refresh rate to be rather fast (e.g. 5ms is about optimal for my string). DoP lights tend to not have too many addresses (i.e. `num_leds`) and this is mostly for them, which need a "recovery period" between commands to prevent lights dropping out/flickering excessively.

The only other option I can think of is putting in a bunch of dummy RMT commands all set to low to fill out the space - 15+15 bits of rmt clock cycles doesn't get us milliseconds.

smithjacobj: On further thought this seemed sane enough so I've implemented the delay in low-only RMT items.

smithjacobj:
Ugh, accidentally clicked a merge button, then got a bad rebase, so it's added a bunch of people I didn't mean to. Pretty sure @jesserockz is supposed to be the only reviewer.

clydebarrow:
There's a heck of a lot of force-pushing going on here. Maybe time to review development processes? To paraphrase Oscar Wilde: "To force push once may be regarded as misfortune, to force-push twice looks like carelessness." :-)

kbx81:
I've also been meaning to take a look at this (since we talked about it _a lot_ on discord), too, but Jesse _did_ write the `esp32_rmt_led_strip` component, so he may have some thoughts...

smithjacobj:
> There's a heck of a lot of force-pushing going on here. Maybe time to review development processes? To paraphrase Oscar Wilde: "To force push once may be regarded as misfortune, to force-push twice looks like carelessness." :-)

I have adopted a philosophy of maintaining logical commits instead of a timeline of commits for feature branches since my last job, but unfortunately it's noisy on GitHub when responding to feedback and addressing CI failures.

Hopefully it makes reviewing easier too, if you review by commit instead of the monolithic PR.

My error was clicking the update button on GitHub without switching to rebase then accidentally rebasing the merge commit.

I'm sure this could be turned into a "vi vs emacs" debate but I don't think there's a need.

PR #5973, opened 2023-12-20
Fix pin reuse error with pin expanders

jesserockz:

If the same pin number was used but from two different i/o expanders in the same platform, the error was given. This changes to use the id of the expander as part of the key

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/pins.py
+++ b/esphome/pins.py
@@ -1,7 +1,7 @@
 import operator
 from functools import reduce
 import esphome.config_validation as cv
-from esphome.core import CORE, ID
+from esphome.core import CORE
 
 from esphome.const import (
     CONF_INPUT,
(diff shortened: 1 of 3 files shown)


kbx81 (dismissed):

clydebarrow:
Since the ID will be unique for each instance of a component, shouldn't it be sufficient to just use the changes around line 68 to use the correct value for the first part of the key, rather than adding an extra dimension to the key?

That section of code (lines 68 and 69) did bother me at the time since it appeared to work, but I was not 100% clear on why.

clydebarrow:
Or is it possible that a single component can have multiple registered pin schemas? In which case the extra key dimension is needed.

jesserockz:
> Since the ID will be unique for each instance of a component, shouldn't it be sufficient to just use the changes around line 68 to use the correct value for the first part of the key, rather than adding an extra dimension to the key?
> 
> That section of code (lines 68 and 69) did bother me at the time since it appeared to work, but I was not 100% clear on why.

I tried that, but the key is pulled out later on line 107 to get the final_validation function, so the non-id key is needed

clydebarrow:
Ah ok. Looks Good Enough to me. I'm sure it will get revisited over time :-)

esphome/pins.py
```diff
@@ -25,15 +25,15 @@ def __init__(self):
     def reset(self):
         self.pins_used = {}
 
-    def get_count(self, key, number):
+    def get_count(self, key, id, number):
         """
         Get the number of places a given pin is used.
         :param key: The ID of the defining component
```
clydebarrow:
```suggestion
        :param key: The key of the registered pin schema.
        :param id: The ID of the defining component
```

clydebarrow (changes requested):

clydebarrow (approved):

justlikeef:
Confirmed this fixed the issue for me in dev

PR #5976, opened 2023-12-20
Fix issue when setting cw/ww brightness via temperature

### LINKED ISSUE #5241: Setting color temperature on CW/WW breaks once CC/CW values have been set directly
### The problem

When using the "Cold White + Warm White Light" (`light` integration with the platform `cwww`), setting and publishing a CW/WW value breaks setting the values via temperature+brightness.

For example, a cwww light can not be set to max brightness once the `random` effect has been active (as that sets the CW/WW values directly).

This happens, because the method where the CW/WW values are calculated from the color temperature, the `white` value (which is used to calculate the brightness, basically) is set to the current max CW/WW value if `white` is not also being set:
https://github.com/esphome/esphome/blob/cd06dc77ee6693595356a72032731701c66c91ce/esphome/components/light/light_call.cpp#L352-L360

However, in the color mode `COLD_WARM_WHITE` this code targets (which is used for platform `cwww` and `rgbww` with `color_interlock: true`), a `white` value is never set (as the brightness is instead controlled by the `brightness` value).

So once both CW and WW are set below 1, it is impossible to get back to full brightness because the `white` value is never set.

2023.11.6

Home Assistant Add-on

_No response_

BK72XX

### Board

_No response_

light

### Example YAML snippet

```yaml
output:
  - platform: template
    id: output_cw
    type: float
    write_action:
      - logger.log:
              format: "CW: %.1f"
              args: [ 'state' ]
  - platform: template
    id: output_ww
    type: float
    write_action:
      - logger.log:
              format: "WW: %.1f"
              args: [ 'state' ]

light:
  - platform: cwww
    name: "Test"
    cold_white: output_cw
    warm_white: output_ww
    cold_white_color_temperature: 6536 K
    warm_white_color_temperature: 2000 K
    default_transition_length: 0s
    effects:
      - random:
```

_No response_

### Additional information

_No response_

patagonaa:

This fixes an issue when CW/WW lights are used via Home Assistant, after both CW and WW values have been set below 1 (which can happen with the `random` effect as that does `call.set_publish(true);` for some reason.

More details in the linked issue.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5241

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** -

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
output:
  - platform: template
    id: output_cw
    type: float
    write_action:
      - logger.log:
              format: "CW: %.1f"
              args: [ 'state' ]
  - platform: template
    id: output_ww
    type: float
    write_action:
      - logger.log:
              format: "WW: %.1f"
              args: [ 'state' ]

light:
  - platform: cwww
    name: "Test"
    cold_white: output_cw
    warm_white: output_ww
    cold_white_color_temperature: 6536 K
    warm_white_color_temperature: 2000 K
    default_transition_length: 0s
    gamma_correct: 1
    effects:
      - random:
          transition_length: 0s
```
After enabling and disabling the `random` effect once, it can be seen that there is no way (via Home Assistant) to go back to full brightness (before this PR).

## Checklist:
  - [ ] ~Tests have been added to verify that the new code works (under `tests/` folder).~ (there are currently no tests for the light component as far as I can tell)

  - [ ] ~Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).~

### DIFF
--- a/esphome/components/light/light_call.cpp
+++ b/esphome/components/light/light_call.cpp
@@ -337,9 +337,12 @@ LightColorValues LightCall::validate_() {
 void LightCall::transform_parameters_() {
   auto traits = this->parent_->get_traits();
 
-  // Allow CWWW modes to be set with a white value and/or color temperature. This is used by HA,
-  // which doesn't support CWWW modes (yet?), and for compatibility with the pre-colormode model,
-  // as CWWW and RGBWW lights used to represent their values as white + color temperature.
+  // Allow CWWW modes to be set with a white value and/or color temperature.
+  // This is used in three cases in HA:
+  // - CW/WW lights, which set the "brightness" and "color_temperature"
+  // - RGBWW lights with color_interlock=true, which also sets "brightness" and
+  //   "color_temperature" (without color_interlock, CW/WW are set directly)
+  // - Legacy Home Assistant (pre-colormode), which sets "white" and "color_temperature"
   if (((this->white_.has_value() && *this->white_ > 0.0f) || this->color_temperature_.has_value()) &&  //
       (*this->color_mode_ & ColorCapability::COLD_WARM_WHITE) &&                                       //
       !(*this->color_mode_ & ColorCapability::WHITE) &&                                                //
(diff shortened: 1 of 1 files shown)


jesserockz (commented):
Let me get this straight, you want to add new code to support an older HA version?

If you are happy to run an old HA version, why not just run an old ESPHome version to match?

patagonaa:
> Let me get this straight, you want to add new code to support an older HA version?

No, the code to support this old version is already there (and I'd be happy to remove this as it adds a bunch of additional complexity).

Currently (no matter the HA version), switching to the random effect on a CW/WW light and switching back causes the max brightness to be capped at the last value that was set by the effect.
This partly happens because the `white` value (which is never set in current HA versions) is calculated from the max of CW/WW (which should always be 1), and the random effect uses `call.set_publish(true);` (which is also not correct, in my opinion), which causes both CW/WW _and_ color temperature to be in `remote_values`.

Ideally, when switching off the effect, the last color temperature would also be applied, however that isn't as simple as it may sound and also wouldn't be necessary if the random effect didn't overwrite the `remote_values`.

Example (can be reproduced with my yaml snippet):
- Light is set to brightness 100% and center color temperature
  - -> CW: 1.0, WW: 1.0
- Set effect to `random`
  - random values are not only set, but also published as `remote_values`
  - -> CW: 0.2, WW: 0.5
- Set effect to `None`
  - -> CW: 0.2, WW: 0.5
- Set color Temperature back to center (`LightCall::transform_parameters_()` uses `max(CW, WW)` as `white` value)
  - -> CW: 0.5, WW: 0.5
- now there is no way from HA to get back to full brightness, as the `white` value is never set in current HA versions and it's instead calculated from the current max CW/WW value (which will never be above 0.5)

kbx81:
I just tested this with a CW/WW bulb I have and confirmed the original problem's existence (HA 2024.4.3 with current ESPHome `dev`) ; this PR resolves the issue.

kbx81 (approved):

PR #4596, opened 2023-03-20
Updating the touchscreen interface structure 

nielsnl68:
# What does this implement/fix?

**This breaking PR implements a consistent solution for the current available touchscreen drivers.**
- A `binary sensor` and `on_touch` event will now be triggered once until all touches are released. 
- Two new events are added:
   - `on_update` will be triggered when a touched position moved around or that an extra touch has been detected.
   - `on_release` will be triggered when all touches are release from the touchscreen.
- The raw touch position is added to the `on_touch` `touch` argument.
- The touchscreen drivers will from now on get the raw touch data on request of main touchscreen component.
  - the touchscreen component will convert the raw data and determine which triggers and events will be called.
 
**NOTE: This is a PR that will break all existing touchscreen components not in the esphome repository,** 

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3405
## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
touchscreen:
  - platform: xpt2046
    id: touch_1
    display: my_display
    ...
  ```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #3487, opened 2022-05-17
Implement Media Player and I2S Media player

jesserockz:
# What does this implement/fix?

Implement Media Players in ESPHome.

The included component - `i2s_audio` allows for playing a URL via the internal DAC on ESP32, or an external DAC via I2S. This is tested with a [CJMCU-1334 UDA1334A I2S Stereo Decoder](https://s.click.aliexpress.com/e/_Ac256f) (Affiliate link) ordered from AliExpress and I am waiting for my M5Stack Atom Speaker kit to arrive.

TODO:
- [ ] https://github.com/schreibfaul1/ESP32-audioI2S/pull/321
- [x] Test M5Stack Atom Echo
- [x] Test M5Stack Atom SPK Kit
- [ ] Test ESP-Muse-Luxe?
- [x] Documentation esphome/esphome-docs#2170
- [x] AIOESPHomeAPI code push and release https://github.com/esphome/aioesphomeapi/releases/tag/v10.9.0
- [x] HA Core code push https://github.com/home-assistant/core/pull/72047

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#2107

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
media_player:
  - platform: i2s_audio
  	name: ESP32 Media Player
    dac_type: external
    i2s_ws_pin: 26
    i2s_dout_pin: 25
    i2s_bclk_pin: 27
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #5986, opened 2023-12-21
[Touchscreen] Add expire of touch record.

Fabian-Schmidt:

I observed an issue while using touchscreen `lilygo_t5_47`. A touch event was fired once on the first touch, but no release or new touch events were fired.

The issue is when interrupt is used for touchscreen i2c handling. On touch release the `this->store_.touched` is no longer set inside the `loop()` function. Therefore `send_touches_` is never called to release the touch.

Added logic into the `loop()` that a touch event expires when not found by touchscreen for ~250ms.

Calling `update_touches` more frequently is not an option. As when requesting data via i2c without an event (indicated by interrupt) will result in no detected touches.

This is a regression of PR https://github.com/esphome/esphome/pull/4596.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
touchscreen:
- platform: lilygo_t5_47
  interrupt_pin: GPIO47
  address: 0x5A
  on_touch:
    then:
      - logger.log:
          format: "Touch x:%i y:%i"
          args: [ 'touch.x', 'touch.y' ]
```

## Checklist:

### DIFF
--- a/esphome/components/touchscreen/touchscreen.cpp
+++ b/esphome/components/touchscreen/touchscreen.cpp
@@ -47,29 +47,9 @@ void Touchscreen::loop() {
     } else {
       this->store_.touched = false;
       this->defer([this]() { this->send_touches_(); });
-    }
-  } else if (!this->touches_.empty()) {
-    // Found active touches without `touched` interrupt/event. Start expire/timeout process.
-    bool call_send_touches = false;
-    bool all_touches_state_released = true;
-    for (auto &tp : this->touches_) {
-      if (tp.second.state == STATE_PRESSED || tp.second.state == STATE_UPDATED) {
-        all_touches_state_released = false;
-        tp.second.release_counter++;
-      } else {
-        tp.second.state = STATE_RELEASED;
-        call_send_touches = true;
-      }
-      if (tp.second.release_counter == 5) {
-        // I made 5 loop (5*~50ms = >250ms) without `touched` active. Assume touch is expired.
-        tp.second.state = tp.second.state | STATE_RELEASING;
-        call_send_touches = true;
-      }
-    }
-    if (call_send_touches) {
-      this->need_update_ = false;
-      this->is_touched_ = !all_touches_state_released;
-      this->defer([this]() { this->send_touches_(); });
+      // Simulate a touch after 250ms. This will reset any existing timeout operation.
+      // This is to detect touch release.
+      this->set_timeout(TAG, 250, [this]() { this->store_.touched = true; });
     }
   }
 }
@@ -102,7 +82,6 @@ void Touchscreen::add_raw_touch_position_(uint8_t id, int16_t x_raw, int16_t y_r
     tp.x_org = tp.x;
     tp.y_org = tp.y;
   }
-  tp.release_counter = 0;
 
   this->touches_[id] = tp;
 
--- a/esphome/components/touchscreen/touchscreen.cpp
+++ b/esphome/components/touchscreen/touchscreen.cpp
@@ -48,7 +48,7 @@ void Touchscreen::loop() {
       this->store_.touched = false;
       this->defer([this]() { this->send_touches_(); });
     }
-  } else if (this->touches_.size() > 0) {
+  } else if (!this->touches_.empty()) {
     // Found active touches without `touched` interrupt/event. Start expire/timeout process.
     bool call_send_touches = false;
     bool all_touches_state_released = true;
--- a/esphome/components/touchscreen/touchscreen.cpp
+++ b/esphome/components/touchscreen/touchscreen.cpp
@@ -48,6 +48,29 @@ void Touchscreen::loop() {
       this->store_.touched = false;
       this->defer([this]() { this->send_touches_(); });
     }
+  } else if (this->touches_.size() > 0) {
+    // Found active touches without `touched` interrupt/event. Start expire/timeout process.
+    bool call_send_touches = false;
+    bool all_touches_state_released = true;
+    for (auto &tp : this->touches_) {
+      if (tp.second.state == STATE_PRESSED || tp.second.state == STATE_UPDATED) {
+        all_touches_state_released = false;
+        tp.second.release_counter++;
+      } else {
+        tp.second.state = STATE_RELEASED;
+        call_send_touches = true;
+      }
+      if (tp.second.release_counter == 5) {
+        // I made 5 loop (5*~50ms = >250ms) without `touched` active. Assume touch is expired.
+        tp.second.state = tp.second.state | STATE_RELEASING;
+        call_send_touches = true;
+      }
+    }
+    if (call_send_touches) {
+      this->need_update_ = false;
+      this->is_touched_ = !all_touches_state_released;
+      this->defer([this]() { this->send_touches_(); });
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


nielsnl68 (approved):

clydebarrow:
Isn't this an issue specific to the particular touch controller, and thus should be dealt with in that component, not the top-level touchscreen component?

nielsnl68:
> Isn't this an issue specific to the particular touch controller, and thus should be dealt with in that component, not the top-level touchscreen component?

No, imho, the particular touch controller should only delver the touches, all logic should be handled in the top-level touchscreen component.

nielsnl68:
That said, i suggest an other approach using a timeout timer. This timer will be (re)set every time touches are found (when `this->is_touched_ `  is set) and will set `this->store_.touched = true;`

clydebarrow:
> all logic should be handled in the top-level touchscreen component.

Only logic that is common to multiple touchscreens. This is a work-around for a specific chip, and should not be applied to all touchscreens, as it risks incorrectly announcing a release when it has not occurred. If it is to go in common code, it should only be activated if the particular touchscreen requires it.

Part of the problem is that the Lilygo touchscreen is not well documented. There may be other possible solutions, e.g. triggering on both edges of the interrupt signal. I believe I have some of these screens and could do some tests after Christmas.

Fabian-Schmidt:
> > all logic should be handled in the top-level touchscreen component.
> 
> Only logic that is common to multiple touchscreens. This is a work-around for a specific chip, and should not be applied to all touchscreens, as it risks incorrectly announcing a release when it has not occurred. If it is to go in common code, it should only be activated if the particular touchscreen requires it.
> 
> Part of the problem is that the Lilygo touchscreen is not well documented. There may be other possible solutions, e.g. triggering on both edges of the interrupt signal. I believe I have some of these screens and could do some tests after Christmas.

I agree with the statement that this should only include logic which is common to multiple touchscreens.

Therefore my question: How should a touchscreen driver announce to this component the touch release? A touch screen driver has only the method `add_raw_touch_position_`. So my understanding is it cannot do this. Also the touchscreen driver is only requested to submit touch if the interrupt is fired. Which given the touch was released is not the case.
Also based on what I can see in the other drivers it is common that they all only get events for when a touch happens, but never when a touch is released. Therefore I believe this common touchscreen component is the right place to implement this.

Please correct me if my assumptions are incorrect as I only have access to this one Touchscreen.
But in my testing it was very obvious that it did not work as the `on_touch` `logger` was fired once and no subsequent touch event were logged. Please validate this for another Touchscreen than `lilygo_t5_47`.

```yaml
  on_touch:
    then:
      - logger.log:
          format: "Touch x:%i y:%i"
          args: [ 'touch.x', 'touch.y' ]
```

clydebarrow:
> How should a touchscreen driver announce to this component the touch release?

When `update_touches()` is called, it calls `add_raw_touch()` zero or more times, and can also set a flag to say "can't update right now." If that flag is not set, and no touches are added, then release is assumed.

Fabian-Schmidt:
> > How should a touchscreen driver announce to this component the touch release?
> 
> When `update_touches()` is called, it either calls `add_raw_touch()` one or more times, or not, and can also set a flag to say "can't update right now." If that flag is not set, and no touches are added, then release is assumed.

Correct. But `update_touches` is not called because the interrupt is not fired. An interrupt is only fired when new touch data is present. If no touch (or release) occurs no interrupt is fired.

clydebarrow:
> But `update_touches` is not called because the interrupt is not fired

Sure, so that needs to be triggered one way or another. Either a timeout, or in `update()` perhaps. The previous code just tested the interrupt flag in `loop()`. An optional timeout that forced a call to `update_touches` is probably simplest. I.e. have a property `touch_timeout` or similar that if non-zero triggers a call to `update_touches` if that time has elapsed since the last call to `update_touches`. The default value for the property could be zero, have it overridden when needed. That would keep the logic in common code, add very little extra code, and make it activated only when needed. Make sense?

Fabian-Schmidt:
I changed the implementation to use a timeout to simulate a touch interrupt 250ms after the last real touch interrupt. This simulated touch is then using the existing code and will ask driver using the `update_touches` method.

Assumption is that every touch driver can fetch data every 250ms.

nielsnl68 (changes requested):
Perfect. Just on question:
Should we not clear the timeout when no touches where detected? 

We can do that by adding `this->cancel_timeout(TAG);` like:

```c++
  if (!this->is_touched_) {
    this->cancel_timeout(TAG);
    this->release_trigger_.trigger();
    for (auto *listener : this->touch_listeners_)
      listener->release();
    this->touches_.clear();
  } else {
```

clydebarrow:
> Assumption is that every touch driver can fetch data every 250ms.

Best not to assume - unless you can test this on every other touchscreen that it will impact, it would be unwise to change the logic for _all_ touchscreens.

I'd like to see a protected property for the delay as discussed above, with default value 0. This could either be set by code in the `lilygo_t5_47` setup, or configured in YAML with a default value for that component.

250ms seems a bit long to me, I suspect that will make button presses e.g. a bit laggy, but I don't know what the right value is, except that in the previous code it basically defaulted to the `loop()` time, which IIRC is something like 30ms. If it's configurable but with a sensible default, then it will work out of the box, but a user can tweak it to suit their preferences if required.

Otherwise heading in the right direction - much cleaner than the first pass.

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -47,6 +47,9 @@ void Touchscreen::loop() {
     } else {
       this->store_.touched = false;
       this->defer([this]() { this->send_touches_(); });
+      // Simulate a touch after 250ms. This will reset any existing timeout operation.
+      // This is to detect touch release.
+      this->set_timeout(TAG, 250, [this]() { this->store_.touched = true; });
```
clydebarrow: Is `TAG` the right choice here? What if there is more than one touchscreen configured?

Fabian-Schmidt: This is no issue. As the scheduler uses a combination of component and name to find items.

Fabian-Schmidt:
> 250ms seems a bit long to me, I suspect that will make button presses e.g. a bit laggy,

I don't think this is the case. The 250ms is only for the touch release detection. The binary sensors (and most other logic) is operating on the touch event. So a large timeout has very little impact on the perceived speed.

Actually having a 'long' timeout is beneficial against bouncing on release of the touch screen.

Fabian-Schmidt:
As requested I added an additional python configuration value `touch_timeout` which can be set by the touchscreen driver to activate the additional logic.

nielsnl68 (approved):
Seems to look good to me.

clydebarrow (approved):

nielsnl68:
Dont forget to add the new variable to the documentation.

Fabian-Schmidt:
> Dont forget to add the new variable to the documentation.

Sure. Do you mind if I wait until <https://github.com/esphome/esphome-docs/pull/3487> is merged? As this is an extensions to these changes / I want to avoid merge conflicts.

nielsnl68:
you can also send me a PR so i can add it to #3487 directly.

PR #5996, opened 2023-12-22
Add feature to reverse TM1637 display char order

### LINKED ISSUE #4375: For some TM1637 versions order of the chips is reversed 'reverse_enable' needed
### The problem

for some version the order of the chips is reversed so you’ll see “56781234” instead of “12345678”.
reverse_enable option is an invalid for that issue .
So if that is possible add reverse_enable option
i'll really apreciate some help on this !

2023.3.2

Home Assistant Add-on

2023.3.6

ESP8266

### Board

nodemcu

tm1637

### Example YAML snippet

```yaml
esphome:
  name: esp-test
  platform: ESP8266
  board: nodemcuv2
wifi:
  ssid: "TP 4M"
  password: "mojaliadato544"
  reboot_timeout: 15min
  power_save_mode: NONE
  fast_connect: true


captive_portal:

# Enable logging
logger:
  baud_rate: 0
# Enable Home Assistant API
api:

  - platform: status
    name: " esp-test-Status "


display:
    platform: tm1637
    id: tm1637_display
    clk_pin: D5
    dio_pin: D6
    inverted: True
    intensity: 6
    length: 6
    lambda: |-
     it.print("012345");
```

_No response_

### Additional information

IMG20230331013210

sellerie98:

This PR adds a feature to optionally reverse the character order.
This is helpful when driving TM1637 modules that have the segment-blocks reversed and display "210543" when "012345" is defined.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/4375

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
display:
    platform: tm1637
    id: tm1637_display
    clk_pin: 16
    dio_pin: 15
    reversed: True
    lambda: |-
      it.print("012345");

```

## Checklist:

### DIFF
--- a/esphome/components/tm1637/tm1637.cpp
+++ b/esphome/components/tm1637/tm1637.cpp
@@ -212,20 +212,20 @@ void TM1637Display::display() {
   this->send_byte_(TM1637_CMD_ADDR);
 
   if (this->reversed_) {
-    uint8_t revbuf_[6] = {0};
+    uint8_t revbuf[6] = {0};
 
     // Reverse character order first half
     for (int8_t i = 2; i >= 0; i--) {
-      revbuf_[2 - i] = this->buffer_[i];
+      revbuf[2 - i] = this->buffer_[i];
     }
 
     // Reverse character order second half
     for (int8_t i = 2; i >= 0; i--) {
-      revbuf_[5 - i] = this->buffer_[3 + i];
+      revbuf[5 - i] = this->buffer_[3 + i];
     }
 
     // Writeback
-    memcpy(this->buffer_, revbuf_, 6);
+    memcpy(this->buffer_, revbuf, 6);
   }
 
   // Write the data bytes
--- a/esphome/components/tm1637/tm1637.cpp
+++ b/esphome/components/tm1637/tm1637.cpp
@@ -215,13 +215,13 @@ void TM1637Display::display() {
     uint8_t revbuf_[6] = {0};
 
     // Reverse character order first half
-    for (int8_t i = 2; i >= 0;  i--) {
-    revbuf_[2-i] = this->buffer_[i];
+    for (int8_t i = 2; i >= 0; i--) {
+      revbuf_[2 - i] = this->buffer_[i];
     }
 
     // Reverse character order second half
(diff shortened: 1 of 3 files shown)


esphome/components/tm1637/tm1637.cpp
```diff
@@ -138,6 +138,7 @@ void TM1637Display::dump_config() {
   ESP_LOGCONFIG(TAG, "TM1637:");
   ESP_LOGCONFIG(TAG, "  Intensity: %d", this->intensity_);
   ESP_LOGCONFIG(TAG, "  Inverted: %d", this->inverted_);
+  ESP_LOGCONFIG(TAG, "  Reversed: %d", this->reversed_);
```
jesserockz, line 141: I know the line above is wrong, but %d is for numbers, not booleans. Look for the TRUE FALSE macro in the codebase

jesserockz (changes requested):

github-cyprien:
Thanks for this feature, working fine for me...
Why the merge is stopped ?

esphome/components/tm1637/tm1637.cpp
sanchosk: > I know the line above is wrong, but %d is for numbers, not booleans. Look for the TRUE FALSE macro in the codebase

I don't understand the complain here - the setter and getter require boolean as argument and set the value to 0 or 1 - this is a number. So the log will display the value 1 for true a 0 for false. Works the same for inverted as for the newly proposed reversed.
Can this PR move forward? It would really help with the 6 digit 7-segment displays...

sanchosk: Would it help if the line would be changed to 
`ESP_LOGCONFIG(TAG, "  Reversed: %s", this->reversed_ ? "true" : "false");`
It's really annoying to do the reversing in code for every display, having to always use lambda instead just standard yaml...

swoboda1337:
Stale, suggested changes not addressed and no docs

PR #5997, opened 2023-12-22
Touchscreen component and driver fixes

nielsnl68:

- Fixed ft6336 strange behavier reported by @nagyrobi.
- In some cases the touchscreen calibration was not setup correctly when display rotation was set. This introduces 2 fixes. 
   1. It will take the native dimensions from the display. 
   2. The option to set the calibration variables in the YAML file.
- The Display has now native width and height public methods. and the private internal width and height is moved to the display class.
- Making sure that per touchscreen driver the predefined calibration settings are not overriding the manual set values
- remove the need to call `touchscreen::get_width_();` and `touchscreen::get_heigth_();` and introducing two properties instead which are loaded on setup.
- Fixing issue where the `on_release` event's was repeatedly was fired after touch release. thanks to @clydebarrow 
- making sure that the `on_update` did give the right information about the state of the touches
   - state Bit 0 is set when there is a touch detect
   - state Bit 1 is set when touch is moving on the screen.
   - state Bit 2 is set when a touch release is detected
   - state Bit 7 is set when the calibration has not correctly set.
   - state returns 0 when there is no touch detected and the info is invalid.
- implanted most of the changes that @clydebarrow has shown in #6049 (thanks)
- making sure the display is running before  touchscreen is setup 
- The XPT2024 will break with this PR. The ``calibration_*_min`` and ``calibration_*_max`` are now moved under `calibration:`

- added a test special for touchscreen drivers.

``` yaml
touchscreen:
  - platform: XPT2024
    calibration:
      x_min: 280
      x_max: 3860
      y_min: 340
      y_max: 3860
```
 
## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes:
     https://github.com/esphome/issues/issues/5251
     https://github.com/esphome/issues/issues/5265
     https://github.com/esphome/issues/issues/5249
     https://github.com/esphome/issues/issues/5441

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3487

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

``` yaml
touchscreen:
  - platform: XPT2024
    calibration:
      x_min: 280
      x_max: 3860
      y_min: 340
      y_max: 3860
```

## Checklist:


esphome/components/display/display.h
```diff
@@ -175,9 +175,13 @@ class Display : public PollingComponent {
...
   /// Get the width of the image in pixels with rotation applied.
-  virtual int get_width() = 0;
+  virtual int get_width() { return this->get_width_internal(); }
+
   /// Get the height of the image in pixels with rotation applied.
-  virtual int get_height() = 0;
+  virtual int get_height() { return this->get_height_internal(); }
+
+  virtual int get_height_internal() = 0;
+  virtual int get_width_internal() = 0;
```
guillempages: If those are called "_internal", shouldn't they be better placed on the "protected:" section?

esphome/components/media_player/media_player.h
```diff
@@ -3,6 +3,10 @@
 #include "esphome/core/entity_base.h"
```
guillempages: I think the changes to the media_player.h file are unrelated to the touchscreen PR

esphome/components/touchscreen/__init__.py
```diff
@@ -30,6 +30,50 @@
...
+                - cv.int_(config[CONF_CALIBRATION][CONF_Y_MAX])
+            )
+            < 10
+        ):
+            raise cv.Invalid("Calibration Y values difference must be more then 10")
+
+    return config
+
+
+def calibration_schema(max):
```
guillempages: For consistency/clarity, you might want to split max in "max_x" and "max_y". I know it they are always used together to detect whether it has been set or not, but using "max" on both axes feels wrong.

Alternatively, you might want to rename the parameter to "default_value", or "unspecified_value" or similar.

esphome/components/touchscreen/__init__.py
```diff
@@ -30,6 +30,50 @@
...
+CONF_X_MIN = "x_min"
+CONF_X_MAX = "x_max"
+CONF_Y_MIN = "y_min"
+CONF_Y_MAX = "y_max"
+
+
+def validate_calibration(config):
+    if CONF_CALIBRATION in config:
+        if (
+            cv.int_(config[CONF_CALIBRATION][CONF_X_MIN]) != 0
```
guillempages: For simplicity/readability, once you know that `CONF_CALIBRATION` exists (i.e. inside the if), you could use a temporary variable `calibration_config` to avoid accessing the dict every time.

esphome/components/xpt2046/touchscreen/__init__.py
```diff
@@ -52,23 +29,31 @@ def validate_xpt2046(config):
...
+                cv.Optional(
+                    touchscreen.CONF_CALIBRATION
+                ): touchscreen.calibration_schema(4095),
+                cv.Optional(CONF_CALIBRATION_X_MIN): cv.invalid(
+                    "Deprecated: use the new Calibration variable"
                 ),
-                cv.Optional(CONF_CALIBRATION_X_MAX, default=4095): cv.int_range(
-                    min=0, max=4095
+                cv.Optional(CONF_CALIBRATION_X_MAX): cv.invalid(
+                    "Deprecated: use the new Calibration variable"
```
guillempages:
```suggestion
                    "Deprecated: use the new 'calibration' configuration variable"
```

esphome/components/xpt2046/touchscreen/__init__.py
```diff
@@ -52,23 +29,31 @@ def validate_xpt2046(config):
...
                 ),
-                cv.Optional(CONF_CALIBRATION_Y_MIN, default=0): cv.int_range(
-                    min=0, max=4095
+                cv.Optional(CONF_CALIBRATION_Y_MIN): cv.invalid(
+                    "Deprecated: use the new Calibration variable"
```
guillempages:
```suggestion
                    "Deprecated: use the new 'calibration:' configuration variable"
```

esphome/components/xpt2046/touchscreen/__init__.py
```diff
@@ -52,23 +29,31 @@ def validate_xpt2046(config):
...
                 ),
-                cv.Optional(CONF_CALIBRATION_Y_MAX, default=4095): cv.int_range(
-                    min=0, max=4095
+                cv.Optional(CONF_CALIBRATION_Y_MAX): cv.invalid(
+                    "Deprecated: use the new Calibration variable"
```
guillempages:
```suggestion
                    "Deprecated: use the new 'calibration' configuration variable"
```

esphome/components/xpt2046/touchscreen/__init__.py
```diff
@@ -52,23 +29,31 @@ def validate_xpt2046(config):
...
+                ),
+                cv.Optional(CONF_CALIBRATION_Y_MAX): cv.invalid(
+                    "Deprecated: use the new Calibration variable"
```
guillempages:
```suggestion
                    "Deprecated: use the new 'calibration' configuration variable"
```

esphome/components/xpt2046/touchscreen/__init__.py
```diff
@@ -52,23 +29,31 @@ def validate_xpt2046(config):
...
+                ),
+                cv.Optional("report_interval"): cv.invalid(
+                    "Deprecated: use the update_interval variable"
```
guillempages:
```suggestion
                    "Deprecated: use the 'update_interval' configuration variable"
```

esphome/components/xpt2046/touchscreen/__init__.py
```diff
@@ -52,23 +29,31 @@ def validate_xpt2046(config):
                 cv.Optional(CONF_INTERRUPT_PIN): cv.All(
                     pins.internal_gpio_input_pin_schema
                 ),
-                cv.Optional(CONF_CALIBRATION_X_MIN, default=0): cv.int_range(
-                    min=0, max=4095
+                cv.Optional(CONF_THRESHOLD, default=400): cv.int_range(min=0, max=4095),
```
guillempages, line 31: I have the feeling the indenting is wrong here

esphome/components/xpt2046/touchscreen/__init__.py
```diff
@@ -52,23 +29,31 @@ def validate_xpt2046(config):
...
+                cv.Optional(
```
guillempages, line 32: and here

esphome/components/xpt2046/touchscreen/xpt2046.cpp
```diff
@@ -77,8 +77,6 @@ void XPT2046Component::dump_config() {
   LOG_UPDATE_INTERVAL(this);
 }
```
guillempages, line 78: Is this really related to this PR?

esphome/components/display/display.h
nielsnl68: I made them public to allow access from the touchscreen drivers,

esphome/components/touchscreen/__init__.py
nielsnl68: I thought about adding *max_x* and *max_y* as you suggesting. But it is now only used for the XPT2046 and there both are set to 4095. In all other cases the the *max* values should be **0** at all times.

Renaming it to "default_max_values" is good idea.

esphome/components/media_player/media_player.h
nielsnl68: Yep you are right. Did not see it slipping into a commit

esphome/components/xpt2046/touchscreen/xpt2046.cpp
nielsnl68: to what do you refer to?

esphome/components/display/display.h
guillempages: Sure, but then maybe they should be renamed ;-)

nielsnl68 (commented):
thanks @guillempages for the quick review. There is one comment still open. When you give me a thumps-up then i can mark that one resolved as well.

esphome/components/xpt2046/touchscreen/xpt2046.cpp
guillempages: The rest of the PR is basically handling the new get_width_internal() changes, or the global calibration options.
Removing the `get_setup_priority()` method seems unrelated to the rest of the PR. IMHO this would be better handled in a separate PR, so that the reasons for that can be better explained, and (unlikely but...) if needed, it can be reverted separately.

(I normally like PRs that focus on once issue, and not generic PRs that happen to change many things in a component).

clydebarrow:
Looks like #5999 and this are doing the same thing.

nielsnl68:
> Looks like #5999 and this are doing the same thing.

There should be a difference, @guillempages made an attempt to allow to use the transform settings from the display and use that for the touchscreen.

And used this code as starting point, before i fixed a lot of issues. Which you are reporting.

clydebarrow:
Split the tt21100 fix into a separate PR and I can approve it straight away.

esphome/components/display/display.h
```diff
@@ -174,10 +174,15 @@ class Display : public PollingComponent {
...
-  virtual int get_width() = 0;
-  /// Get the height of the image in pixels with rotation applied.
-  virtual int get_height() = 0;
+  /// Get the calculated width of the display in pixels with rotation applied.
+  virtual int get_width() { return this->get_width_internal(); }
+  /// Get the calculated height of the display in pixels with rotation applied.
+  virtual int get_height() { return this->get_height_internal(); }
+
+  /// Get the native (original) width of the display in pixels.
+  int get_native_width() { return this->get_width_internal(); }
```
clydebarrow, line 184: Why introduce an extra method here, why not just rename `get_width_internal()` which is already public?

nielsnl68: Because then we need to change all display drivers as well. My intension was to use the internal version.  But that was not good enough. And you and @guillempages suggested to create an other set of methods to get the right values.
So i did that now. we have discussed it very well on discord.

clydebarrow: Yeah, but we don't want two sets of functions with different names doing the same thing (not public ones anyway) - that will just lead to confusion. Either stick with the original names, or replace them.

clydebarrow: So in your code, you're using `get_internal_width()` in some places, and `get_native_width()` in others, so confusion has already set in. I recommend removing the `native` versions and just sticking with the `internal` as before. The name is not critical, since they're not user-exposed anyway.

Other than that, it looks OK. If you fix up those functions I'll have a final look and it should be good to go.

nielsnl68: > So in your code, you're using `get_internal_width()` in some places, and `get_native_width()` in others, so confusion has already set in. I recommend removing the `native` versions and just sticking with the `internal` as before. The name is not critical, since they're not user-exposed anyway.

Good point going to fix that.
> 
> Other than that, it looks OK. If you fix up those functions I'll have a final look and it should be good to go.

🤩 let me know how it go's now. Sorry that i takeover your PR :D

nagyrobi:
Trying on a [Lanbon L8](https://deploy-preview-622--esphome-devices.netlify.app/devices/Lanbon-L8-LCD-Switch).
Intention is to use with LVGL (migrate from OpenHASP).

```
  - source: github://pr#6049
    refresh: 1440min
    components: [ touchscreen ]
  - source: github://pr#5997
    refresh: 1440min
    components: [ touchscreen, ft63x6 ]
  - source: github://pr#6055
    refresh: 1440min
    components: [ ft63x6 ]

touchscreen:
  - platform: ft63x6
    id: tft_touch
    display: tft_display
    calibration:
#      x_min: 0
      x_max: 240
#      y_min: 0
      y_max: 320
    on_touch:
      - lambda: |-
                touch.x,
                touch.y,
                touch.x_raw,
                touch.y_raw
                );
```

Results in:
```
src/esphome/components/touchscreen/touchscreen.cpp: In member function 'virtual void esphome::touchscreen::Touchscreen::call_setup()':
src/esphome/components/touchscreen/touchscreen.cpp:20:44: error: 'class esphome::display::Display' has no member named 'get_native_width'
   20 |     this->display_width_ = this->display_->get_native_width();
      |                                            ^~~~~~~~~~~~~~~~
src/esphome/components/touchscreen/touchscreen.cpp:21:45: error: 'class esphome::display::Display' has no member named 'get_native_height'
   21 |     this->display_height_ = this->display_->get_native_height();
      |                                             ^~~~~~~~~~~~~~~~~
```

Otherwise the device works fine, if I ignore the PRs related to ft63x6, remove calibration config and change straight in its code
```
  this->x_raw_max_ = 240;
  this->y_raw_max_ = 320;
 ```
 compiles and works correctly.

File: esphome/components/display/display.h (hunk 13 above)
Comment (guillempages) [reply esphome/components/display/display.h:184]: > So in your code, you're using `get_internal_width()` in some places, and `get_native_width()` in others, so confusion has already set in. I recommend removing the `native` versions and just sticking with the `internal` as before. The name is not critical, since they're not user-exposed anyway.

Why not keep the "internal" as protected, so that they can still be used by the components, but use the "native" for the touchscreen? Somehow I find the "internal" ones should not be used by "other" components or be made public.

Comment (nielsnl68) [reply esphome/components/display/display.h:184]: > Why not keep the "internal" as protected, so that they can still be used by the components, but use the "native" for the touchscreen? Somehow I find the "internal" ones should not be used by "other" components or be made public.

That is what it should have been, i just forgot those to changes.

**Comment by nielsnl68** [comment]
@nagyrobi to fix your issue you need to change the `external_component`:
```yaml
external_component:
  - source: github://pr#5997
    refresh: 1440min
    components: [ display, touchscreen, ft63x6 ]

File: esphome/components/display/display.h [hunk 14]
```diff
@@ -175,10 +175,15 @@ class Display : public PollingComponent {
...
```
clydebarrow, line 184: I'd suggest making this virtual so it can be overridden in subclasses.

esphome/components/display/display.h
```diff
@@ -532,6 +538,9 @@ class Display : public PollingComponent {
   void do_update_();
   void clear_clipping_();
 
+  virtual int get_height_internal() = 0;
```
clydebarrow, line 546: Since these are for internal use only, they should now be protected, and thus renamed to `get_width_internal_()`

clydebarrow (changes requested):

esphome/components/display/display.h
jesserockz: virtual protected functions do not use the `_` suffix

clydebarrow: Maybe that should be spelt out in https://esphome.io/guides/contributing.html#codebase-standards then.

esphome/components/display/display.h
nielsnl68: sorry i do not see the need for that. You can override the `get_width_internal()` already when you need.

esphome/components/display/display.h
jesserockz: clang will ping you for it if you add the suffix too I believe =)

esphome/components/display/display.h
clydebarrow: I'm still unhappy with having two sets of methods doing exactly the same thing with no obvious rationale for the duplication.

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -94,32 +105,37 @@ void Touchscreen::add_raw_touch_position_(uint8_t id, int16_t x_raw, int16_t y_r
 }
 
 void Touchscreen::send_touches_() {
+  TouchPoints_t touches;
+  for (auto tp : this->touches_) {
+    ESP_LOGV(TAG, "Touch status: %d/%d: raw:(%4d, %4d) calc:(%4d, %4d)", tp.second.id, tp.second.state, tp.second.x_raw,
+             tp.second.y_raw, tp.second.x, tp.second.y);
+    touches.push_back(tp.second);
```
clydebarrow, line 117: Touches that have been released should be filtered out here.

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -69,17 +78,19 @@ void Touchscreen::add_raw_touch_position_(uint8_t id, int16_t x_raw, int16_t y_r
   tp.x_raw = x_raw;
```
clydebarrow, line 83: Consider this scenario - two touches are registered, then one is lifted. Now another finger touches the screen in a different place and send with the same ID as the previously lifted touch. Since the touches map will still contain the original second touch, with state RELEASED, it will now be put in state UPDATED, even though it's actually a new touch and so should be in state PRESSED.

I think that once a touchpoint goes into state RELEASED, it should be removed from the touches map, so any new touch with the same ID will be recognised as a new touch, not a continuation of the previous one.

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -69,17 +78,19 @@ void Touchscreen::add_raw_touch_position_(uint8_t id, int16_t x_raw, int16_t y_r
...
+    if (this->swap_x_y_) {
+      std::swap(x, y);
+    }
 
-  if (this->swap_x_y_) {
-    std::swap(x, y);
+    tp.x = (uint16_t) ((int) x * this->display_width_ / 0x1000);
+    tp.y = (uint16_t) ((int) y * this->display_height_ / 0x1000);
+  } else {
+    tp.state |= STATE_CALIBRATE;
```
clydebarrow, line 97: What's the meaning of STATE_CALIBRATE and where is it used?

clydebarrow (changes requested):

esphome/components/display/display.h
nielsnl68: > I'm still unhappy with having two sets of methods doing exactly the same thing with no obvious rationale for the duplication.

How about updating the `get_height();` and `get_width();` by adding a default *boolean* parameter when set true it will return the **native** value otherwise it is as default behavier?

esphome/components/touchscreen/touchscreen.cpp
nielsnl68: The way I designed it is well thought out, by me. I have been thinking about the best solution for sometime now.
And I also have toughed about the solution you suggest, but then the user needs to check the touch ID's every time  an on_update is fired. Imho an lot of extra logic that is now prevented.

Say you created an zoom gesture that uses only 2 touches. at some point you touching 3 point on the screen (no zoom happens anymore). 
- In your solution, when you release point 2; the zoom gesture sees 2 points again and start zooming again.  Or the gesture needs to store the ID's that where uses when it started zooming. 

- In my solution the number of points stay's the same (3 Points) until all touches are removed. as long there are 2 touches the gesture needs only to validate the state =0 to enable the zoom function.

> I think that once a touchpoint goes into state RELEASED, it should be removed from the touches map, so any new touch with the same ID will be recognised as a new touch, not a continuation of the previous one.

The handling of retouching the screen is still an open issue. Until i know more how Touchscreen drivers handle this kind of touch sequence i cant yet handle this correctly. 

I belief you have a display where you can have multiple touches is it not? Please provide me with a log where i can see how it is handling retouching etc. 

Because there is no gesture's made i do not find it a big issue for now.

esphome/components/touchscreen/touchscreen.cpp
nielsnl68: This is a new flag that you can use in your yaml script to see of you need to calibrate the touchscreen or not.
And tells that the **x** and **y** are invalid.

nielsnl68:
- The red flag is because i made this PR useable for current version ESPHome users.
You should be able to fix most of the issues with you touchscreen with this PR.
Add the following part in your YAML file.

```yaml
external_components:
  - source: github://pr#5997
    components: [ display, touchscreen, YourTouchScreen]
    refresh: 0sec
``` 
> Replace "`YourTouchScreen`" with the name of the used touchscreen.

nagyrobi:
This PR works nicely for me on [Lanbon L8](https://deploy-preview-622--esphome-devices.netlify.app/devices/Lanbon-L8-LCD-Switch) with `ft63x6`.

nagyrobi (approved):

esphome/components/display/display.h
kbx81: Any decision/resolution on this? 🙂

nielsnl68: It will be resolved in a next PR. For the new version of the ili9xxx component. 
https://github.com/nielsnl68/esphome/tree/nvds-displaydriver imm this will become THE native platform component between all the graphical display drivers and the display component.

clydebarrow (changes requested):
The changes to `display` relating to anti-aliased fonts should be removed from this PR and left for #6198.

nielsnl68:
> The changes to `display` relating to anti-aliased fonts should be removed from this PR and left for #6198.

I have remove those.

esphome/components/xpt2046/test.esp32.yaml
jesserockz, line 1: Please dont add test files into the component folders.
If you want to add a test file for this, it needs to go into `/tests/components/xxxx/test.xxx.yaml`

esphome/components/tt21100/test.esp32.yaml
jesserockz, line 1: Same here

tests/test13.yaml
jesserockz, line 1: Given the new yaml test setup, we dont want any new test yaml files in here please

jesserockz (dismissed):

nielsnl68:
@clydebarrow Could you please approve this PR, thanks.

tests/components/ft63x6/test.esp32.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+esp32:
+  board: esp-wrover-kit
+  framework:
+    type: arduino
+
+psram:
+
+logger:
+  level: verbose
```
kbx81: In light of the new testing frameworks, please do not include any of this in the test yaml files.

Tests should now include **only** the component itself and components it depends on (such as `spi`, `display`, etc.). Do not include anything that isn't strictly required to compile the component the test is for -- it just makes the test take longer to complete.

tests/components/tt21100/test.esp32.yaml
kbx81, line 1: Same as above

tests/components/xpt2046/test.esp32.yaml
kbx81, line 1: Here, too

kbx81 (dismissed):
Comments about tests

esphome/components/touchscreen/__init__.py
```diff
@@ -33,6 +36,61 @@
 CONF_ON_UPDATE = "on_update"
 CONF_TOUCH_TIMEOUT = "touch_timeout"
 
+# added here so everone can use the PR.
+# CONF_TRANSFORM = "transform"
+# CONF_MIRROR_X = "mirror_x"
+# CONF_MIRROR_Y = "mirror_y"
+# CONF_SWAP_XY = "swap_xy"
```
guillempages: Do you still need those here?
```suggestion
```

esphome/components/touchscreen/__init__.py
```diff
@@ -33,6 +36,61 @@
...
+        if (
+            cv.int_([CONF_X_MIN]) != 0
+            and cv.int_(calibration_config[CONF_X_MAX]) != 0
+            and abs(
+                cv.int_(calibration_config[CONF_X_MIN])
+                - cv.int_(calibration_config[CONF_X_MAX])
+            )
+            < 10
+        ):
+            raise cv.Invalid("Calibration X values difference must be more then 10")
```
guillempages:
```suggestion
            raise cv.Invalid("Calibration X values difference must be more than 10")
```

esphome/components/touchscreen/__init__.py
```diff
@@ -33,6 +36,61 @@
...
+        if (
+            cv.int_(calibration_config[CONF_Y_MIN]) != 0
+            and cv.int_(calibration_config[CONF_Y_MAX]) != 0
+            and abs(
+                cv.int_(calibration_config[CONF_Y_MIN])
+                - cv.int_(calibration_config[CONF_Y_MAX])
+            )
+            < 10
+        ):
+            raise cv.Invalid("Calibration Y values difference must be more then 10")
```
guillempages:
```suggestion
            raise cv.Invalid("Calibration Y values difference must be more than 10")
```

esphome/components/touchscreen/__init__.py
```diff
@@ -33,6 +36,61 @@
...
+
+    return config
+
+
+def calibration_schema(default_max_values):
+    return cv.Schema(
+        {
+            cv.Optional(CONF_X_MIN, default=0): cv.int_range(min=0, max=4095),
```
guillempages, line 77: Is there a reason why the minimum cannot be greater than 4095? Are you sure that there are no strange devices out there that have weird values here?
(Comment valid for all 4 values)

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -13,26 +13,38 @@ void Touchscreen::attach_interrupt_(InternalGPIOPin *irq_pin, esphome::gpio::Int
...
+  }
+  PollingComponent::call_setup();
 }
 
 void Touchscreen::update() {
   if (!this->store_.init) {
     this->store_.touched = true;
   } else {
     // no need to poll if we have interrupts.
+    ESP_LOGW(TAG, "Touch Polling Stopped !!!");
```
guillempages: Is this really a warning? Wouldn't this be a "normal" use case, and thus just a LOGI would be enough?
Also, do you really need all three exclamation marks? ;-)

guillempages (commented):
Left some minor formatting/leg message improvement comments.

Otherwise, looks good to me

esphome/components/touchscreen/__init__.py
nielsnl68: I get where you saying and but for now i do not have seen a touchscreen driver that needs more then 4095 value.
As soon there is one we could change at that point is it not? The value is not set in stone.

esphome/components/touchscreen/touchscreen.cpp
nielsnl68:
```suggestion
    ESP_LOGW(TAG, "Touch Polling Stopped. You can safely remove the "update_interval:" variable from the YAML file.");
```
You could see this as a yaml configuration problem where the user should have remove the `update_interval:` variable.

nielsnl68 (commented):
Thanks @guillempages for reviewing and give your opinions. 
I changed them as suggested.

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -13,26 +13,38 @@ void Touchscreen::attach_interrupt_(InternalGPIOPin *irq_pin, esphome::gpio::Int
...
+  }
+  PollingComponent::call_setup();
 }
 
 void Touchscreen::update() {
   if (!this->store_.init) {
     this->store_.touched = true;
   } else {
     // no need to poll if we have interrupts.
+    ESP_LOGW(TAG, "Touch Polling Stopped. You can safely remove the "update_interval:" variable from the YAML file.");
```
nielsnl68:
```suggestion
    ESP_LOGW(TAG, "Touch Polling Stopped. You can safely remove the 'update_interval:' variable from the YAML file.");
```

clydebarrow (approved):

PR #6002, opened 2023-12-23
dashboard: refactor ping implementation to be more efficient

bdraco:

- Use icmplib to send the pings
- Cache the dns record for 120s (there were complaints about the number of dns requests that ping generates)
- Set reasonable event loop policies to avoid running out of executor threads (borrowed from HA)
- Use PidfdChildWatcher child watcher from python 3.12 when available (borrowed from HA)
- Log loop exceptions (borrowed from HA)

The subprocesses use a lot of cpu time since they have the overhead of the fork/exec/spawn
related https://github.com/esphome/issues/issues/5257

A future PR should be able to need to manually specify ping as its efficient enough that we can auto fallback

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/dashboard/status/ping.py
+++ b/esphome/dashboard/status/ping.py
@@ -6,10 +6,10 @@ from typing import cast
 
 from icmplib import Host, SocketPermissionError, async_ping
 
+from ..const import MAX_EXECUTOR_WORKERS
 from ..core import DASHBOARD
 from ..entries import DashboardEntry, EntryState, bool_to_entry_state
 from ..util.itertools import chunked
-from ..const import MAX_EXECUTOR_WORKERS
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -59,6 +59,9 @@ class PingStatus:
                         raise result
                     entries_with_addresses[entry] = result
 
+            _LOGGER.warning(
+                "Pinging entries with addresses: %s", entries_with_addresses
+            )
             # Ping all entries with valid addresses
             for ping_group in chunked(entries_with_addresses.items(), GROUP_SIZE):
                 entry_addresses = cast(tuple[DashboardEntry, list[str]], ping_group)
--- a/esphome/dashboard/status/ping.py
+++ b/esphome/dashboard/status/ping.py
@@ -36,17 +36,30 @@ class PingStatus:
             to_ping: list[DashboardEntry] = [
                 entry for entry in current_entries if entry.address is not None
             ]
-            # icmplib defaults to 50 pings at a time
-            for ping_group in chunked(to_ping, 50):
+            # Try to do 20 at a time
+            for ping_group in chunked(to_ping, 20):
                 ping_group = cast(list[DashboardEntry], ping_group)
+                dns_results = await asyncio.gather(
+                    *(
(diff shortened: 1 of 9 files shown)


esphome/dashboard/status/ping.py
```diff
@@ -27,6 +24,10 @@ async def async_run(self) -> None:
         """Run the ping status."""
         dashboard = DASHBOARD
         entries = dashboard.entries
+        privileged = await _can_use_icmp_lib_with_privilege()
+        if privileged is None:
+            _LOGGER.warning("Cannot use icmplib because privileges are insufficient")
+            return
```
bdraco, line 33: I'm not sure this would ever happen but it would be nice to get a warning if it does

esphome/dashboard/status/ping.py
```diff
@@ -35,15 +36,43 @@ async def async_run(self) -> None:
             to_ping: list[DashboardEntry] = [
                 entry for entry in current_entries if entry.address is not None
             ]
-            for ping_group in chunked(to_ping, 16):
+            # icmplib defaults to 50 pings at a time
+            for ping_group in chunked(to_ping, 50):
```
bdraco: 50 is too high since the default thread pool is much lower and it will spend all its time in the executor resolving

bdraco:
It also looks like we don't set up a reasonable thread pool limit and use the asyncio default which we know has problems since its based on the number of cpus which makes little sense for our case as its still running on a single cpu core.  We can copy the implementation from HA

rikroe:
Just my 2 cents: I'm not sure if the additional requirement of `icmplib` is needed.

With #6001 there is no noticeable CPU usage from esphome (at least when looking at htop, no further analysis done).
Also, even with an `icmplib` implementation, without #6001 we still would be flooding the DNS server with requests?

bdraco:
> Just my 2 cents: I'm not sure if the additional requirement of `icmplib` is needed.
> 
> 
> 
> With #6001 there is no noticeable CPU usage from esphome (at least when looking at htop, no further analysis done).
> 

At least on my test system the constant fork/exec of ping processes drove up the load quite a bit.

CPU wouldn't likely be attributed to the python process. It would be system and the ping processes since the overhead is not in user space.

Also if it's not a problem and nobody wants this change, I'm more than happy to close this PR as no reason to churn for a undesired performance improvement 

> Also, even with an `icmplib` implementation, without #6001 we still would be flooding the DNS server with requests?

Right

bdraco:
Another thing we could do is auto fallback to ping if is unreachable and the system ips are not on the subnet.  That should remove the need to do manual configuration

bdraco:
We could also implement the dns cache similar to how aiohttp does to avoid the executor churn

bdraco:
I think that would be really nice since we have people who turn off mdns when it actually works fine, what do you mean we could just eliminate the option because fallback would be automatic

jesserockz (approved):

bdraco:
Thanks. I'll hold off adding the auto fallback until this proves out in a beta + release

PR #6007, opened 2023-12-23
Add support for Sharp GP2Y1010AU0F PM2.5 sensor

zry98:

Add support for Sharp GP2Y1010AU0F dust sensor, which is a very cheap (read: inaccurate) optical (infrared LED and phototransistor) sensor suitable for detecting fine particulate matter like dust and smoke.

While the official [datasheet](https://global.sharp/products/device/lineup/data/pdf/datasheet/gp2y1010au_e.pdf) doesn't mention its sensing resolution (whether it can detect PM2.5 particle or not), some vendors (e.g., [Waveshare](https://www.waveshare.com/wiki/Dust_Sensor)) say it can detect particle larger than 0.8 μm in diameter. To make things easier, we'll assume it's capable of that and let the sensor use the PM2.5 device class, reporting in "µg/m³" with 0 accuracy decimals as default.

See also:
- [Datasheet](https://global.sharp/products/device/lineup/data/pdf/datasheet/gp2y1010au_e.pdf>)
- [Application note](https://global.sharp/products/device/lineup/data/pdf/datasheet/gp2y1010au_appl_e.pdf)
- [Research notes by Chris Nafis](https://www.howmuchsnow.com/arduino/airquality/)
- [Interfacing with Arduino](https://electropeak.com/learn/interfacing-gp2y1010au0f-optical-dust-sensor-module-with-arduino/)

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3492

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
output:
  - platform: gpio
    id: dust_sensor_led
    pin: 5
    inverted: true

sensor:
  - platform: adc
    id: adc_sensor
    pin: A0
    internal: true
    update_interval: never

  - platform: gp2y1010au0f
    name: 'Indoor PM2.5'
    sensor: adc_sensor
    adc_voltage_offset: 0.2
    adc_voltage_multiplier: 3.3
    output: dust_sensor_led
    update_interval: 10s
    filters:
          window_size: 3
          send_every: 3
```

## Checklist:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -152,6 +152,7 @@ esphome/components/ft63x6/* @gpambrozio
 esphome/components/gcja5/* @gcormier
 esphome/components/gdk101/* @Szewcson
 esphome/components/globals/* @esphome/core
+esphome/components/gp2y1010au0f/* @zry98
 esphome/components/gp8403/* @jesserockz
 esphome/components/gpio/* @esphome/core
 esphome/components/gpio/one_wire/* @ssieb
--- a/esphome/components/gp2y1010au0f/gp2y1010au0f.cpp
+++ b/esphome/components/gp2y1010au0f/gp2y1010au0f.cpp
@@ -1,6 +1,8 @@
 #include "gp2y1010au0f.h"
-#include "esphome/core/log.h"
 #include "esphome/core/hal.h"
+#include "esphome/core/log.h"
+
+#include <cinttypes>
 
 namespace esphome {
 namespace gp2y1010au0f {
@@ -11,7 +13,7 @@ static const float MAX_VOLTAGE = 4.0f;
 
 void GP2Y1010AU0FSensor::dump_config() {
   LOG_SENSOR("", "Sharp GP2Y1010AU0F PM2.5 Sensor", this);
-  ESP_LOGCONFIG(TAG, "  Sampling duration: %d ms", this->sample_duration_);
+  ESP_LOGCONFIG(TAG, "  Sampling duration: %" PRId32 " ms", this->sample_duration_);
   ESP_LOGCONFIG(TAG, "  ADC voltage multiplier: %.3f", this->voltage_multiplier_);
   LOG_UPDATE_INTERVAL(this);
 }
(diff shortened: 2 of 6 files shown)


optimusprimespace:
How hard would it be to implement this for the Sharp GP2Y1014AU0F, it's almost the same as the GP2Y1010AU0F, but it's improved. (https://github.com/sharpsensoruser/sharp-sensor-demos/wiki/Application-Guide-for-Sharp-GP2Y1014AU0F-Dust-Sensor)

zry98:
> How hard would it be to implement this for the Sharp GP2Y1014AU0F, it's almost the same as the GP2Y1010AU0F, but it's improved. (https://github.com/sharpsensoruser/sharp-sensor-demos/wiki/Application-Guide-for-Sharp-GP2Y1014AU0F-Dust-Sensor)

Looks like exactly the same, what do you think about adding a `model` option as in the [HLW8012](https://esphome.io/components/sensor/hlw8012.html?highlight=model) platform?

optimusprimespace:
I think it would be awesome to just choose the model you have and it would work automatically. Btw did you connect the sensor to the ESP with 150k ohm resistor & 220 µF capacitor? I think it needs to be added to the docs. Also the under zero dust condition voltage for the GP2Y1014AU0F is from 0.1-1.1. But in the real world it's output Voc is 0.6 Volts, so the offset can be adjusted proportionally. The GP2Y1010AU0F has a Voc from 0 to 1.5V noted here (https://github.com/sharpsensoruser/sharp-sensor-demos/blob/master/docs/Sharp%20GP2Y1010AU0F%20-%20Dust%20Density%20Conversion.pdf). The real world Voc is not noted here, but you can measure that if you put the sensor in a sealed off bag and let it sit for a while. You can also easily convert to Dust Density in units of µg/m3, but i see you opted for the ppm calculation.

zry98:
> I think it would be awesome to just choose the model you have and it would work automatically.

The problem is I found two of my 1010 produced different voltage ranges at the same time, so I think individual calibration using `adc_voltage_offset` is still essential (don't know if 1014 has better quality on this).

> Btw did you connect the sensor to the ESP with 150k ohm resistor & 220 µF capacitor?

I've used the capacitor but not the resistor since I used +3.3 V.

> You can also easily convert to Dust Density in units of µg/m3, but i see you opted for the ppm calculation.

I used the µg/m3 conversion approximate from [https://www.howmuchsnow.com/arduino/airquality/](https://www.howmuchsnow.com/arduino/airquality/).

jesserockz (approved):

PR #6008, opened 2023-12-23
tt21100: restore init read

nielsnl68:

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow (approved):

kroimon:
Just for my understanding: this got lost in #4596?

clydebarrow:
> Just for my understanding: this got lost in #4596?

Yes.

clydebarrow:
https://github.com/esphome/issues/issues/5274 refers.

kroimon:
Also, before #4596, there was a comment saying "Origin is top right, so mirror X by default". Do we also want to restore this default mirroring?

clydebarrow:
> Do we also want to restore this default mirroring?

IMHO no, since that's likely to be hardware-specific, and can be handled by the `transform:` option in yaml.

kroimon:
> IMHO no, since that's likely to be hardware-specific, and can be handled by the `transform:` option in yaml.

That's true, but should probably be part of the default configuration for this component, as the only hardware I know of that uses this touchscreen (ESP32-S3-BOX) has it mirrored and existing configurations will otherwise still be be broken.

PR #6014, opened 2023-12-24
Make `pulse_meter` PULSE filter report the pulse as soon as it can

TrentHouliston:

See https://github.com/esphome/esphome/pull/5813#issuecomment-1821803351

The pulse meter filter would only publish the pulses when the next pulse came through when in PULSE mode.
This caused some strange behaviour when the pulses were slow as features like timeout would happen at unexpected times as even though a pulse had happened since the next pulse hadn't yet it was not counted.

This change allows the main loop to "peek" the filter state to see if it can count the incomplete pulse and repay that pulse count later when the interrupt catches up.

This also resulted in some refactoring to the pulse filter to enable this which simplified it significantly in the process.
I've also added a markdown file to help describe the pulse filter so that myself and others are not as confused by it in the future.

Since almost all of my devices use higher frequencies and the EDGE filter mode, if someone who uses some of these slower pulse rate devices could test this that would be appreciated.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


TrentHouliston:
Should be a more complete fix for #5813 as described in that PR

esphome/components/pulse_meter/pulse_filter.md
jesserockz, line 1: While I appreciate you putting the effort into this file (It looks great btw), we do not want "documentation" throughout the esphome codebase in adhoc files. This is what https://esphome/esphome-docs is for.

esphome/components/pulse_meter/pulse_meter_sensor.h
```diff
@@ -6,6 +6,7 @@
 #include "esphome/core/helpers.h"
 
 #include <cinttypes>
+#include <variant>
```
jesserockz: Is this used?

jesserockz (changes requested):

esphome/components/pulse_meter/pulse_meter_sensor.h
TrentHouliston: It is not! I couldn't manage to get it to compile and work on the ESPs, I was trying to avoid having to have an object allocated for each filter when only one was used.

Looks like I left the header behind

jesserockz (approved):

PR #6022, opened 2023-12-26
Fix device not requesting Home Assistant time at the update interval

### LINKED ISSUE #4424: Home Assistant Time never polls after the initial request
### The problem

The [Home Assistant Time](https://esphome.io/components/time/homeassistant.html) component only updates when Home Assistant or ESPHome logging first connects to the ESP.

After some time (1-5 days), the local ESP clock has drifted significantly (sometimes a minute), which is really annoying when it is displaying the time.

I can get the time to re-sync by connecting to the logs with ESPHome, but that isn't a reasonable thing to need to do every couple days just to have an accurate clock displayed.

2023.3.2

Home Assistant Add-on

2023.4.2

ESP8266

### Board

D1 Mini

time/homeassistant

### Example YAML snippet

```yaml
esphome:
  name: ${name}

esp8266:
  board: d1_mini

logger:

api:

ota:

wifi:
  ssid: !secret wifi_ssid

globals:
  - id: GLOBAL_LAST_TIMESYNC
    type: char*
    restore_value: no
    initial_value: '"Never\0\0\0\0"'
  - id: GLOBAL_TIMESYNC_COUNT
    type: uint8_t
    restore_value: no
    initial_value: "0"

time:
  - platform: homeassistant
    id: ha_time
    update_interval: 30s
    on_time_sync:
      then:
        - lambda: |-
            std::strncpy(id(GLOBAL_LAST_TIMESYNC), id(ha_time).now().strftime("%H:%M:%S").c_str(), 8);
            id(GLOBAL_TIMESYNC_COUNT)++;
        - logger.log:
            format: "Synchronized system clock"
            level: WARN

display:
  - platform: ssd1306_i2c
    model: SSD1306 64x48
    address: 0x3C
    lambda: |-
      it.fill(COLOR_OFF);
      it.printf(32, 0, id(fnt9), COLOR_ON, TextAlign::TOP_CENTER, "%d %s", id(GLOBAL_TIMESYNC_COUNT), id(GLOBAL_LAST_TIMESYNC));
```

Notably, "Synchronized system clock" only appears when connecting an additional ESPHome log viewer, or (re)connecting Home Assistant.
It does not appear every 30 seconds as expected from the YAML. In fact, after running for 24+ hours, it never re-synchronizes.

### Additional information

I added some debug code to components/homeassistant/time/homeassistant_time.cpp files and verified that it does call `api::global_api_server->request_time();`. However, the time is not actually updated. Not sure if there is an API problem, or what.

bdraco:

fixes https://github.com/esphome/issues/issues/4424

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/api/api_server.cpp
+++ b/esphome/components/api/api_server.cpp
@@ -319,7 +319,7 @@ void APIServer::set_reboot_timeout(uint32_t reboot_timeout) { this->reboot_timeo
 #ifdef USE_HOMEASSISTANT_TIME
 void APIServer::request_time() {
   for (auto &client : this->clients_) {
-    if (!client->remove_ && client->connection_state_ == APIConnection::ConnectionState::CONNECTED)
+    if (!client->remove_ && client->is_authenticated())
       client->send_time_request();
   }
 }


esphome/components/api/api_server.cpp
```diff
@@ -319,7 +319,7 @@ void APIServer::set_reboot_timeout(uint32_t reboot_timeout) { this->reboot_timeo
 #ifdef USE_HOMEASSISTANT_TIME
 void APIServer::request_time() {
   for (auto &client : this->clients_) {
-    if (!client->remove_ && client->connection_state_ == APIConnection::ConnectionState::CONNECTED)
```
bdraco, line 322: This was effectively always going to be false since by the time it would ever get here it will be authenticated not connected unless something is very broken

clydebarrow (approved):
LGTM

esphome/components/api/api_server.cpp
clydebarrow: Confirmed by the code executed on successful "connect":
```
  if (correct) {
    ESP_LOGD(TAG, "%s: Connected successfully", this->client_combined_info_.c_str());
    this->connection_state_ = ConnectionState::AUTHENTICATED;
    this->parent_->get_client_connected_trigger()->trigger(this->client_info_, this->client_peername_);
#ifdef USE_HOMEASSISTANT_TIME
    if (homeassistant::global_homeassistant_time != nullptr) {
      this->send_time_request();
    }
#endif
```

bdraco:
Thanks

PR #6035, opened 2023-12-29
Add state listeners to `rotary_encoder`

clydebarrow:

Currently the `rotary_encoder` component maintains listeners for clockwise and anti-clockwise rotation events; this PR adds
listeners for computed state position changes.

This is intended for use by other components.

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


jesserockz:
It is a sensor, you can just use `add_on_state_callback` or `add_on_raw_state_callback`...

clydebarrow:
True, but that uses `float` so restricts the monotonic values reported to roughly +/- 2^23 after which it will start reporting changes at half the actual rate. While that's a corner case, reporting as `int` does not have that problem.

jesserockz (approved):

PR #6040, opened 2023-12-30
Support full (>460 char) dumps of Pronto IR commands

rpaskowitz:

Currently Pronto IR messages get truncated, this allows them to be fully logged

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):**  https://github.com/esphome/issues/issues/3557#issuecomment-1409604480

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/remote_base/pronto_protocol.cpp
+++ b/esphome/components/remote_base/pronto_protocol.cpp
@@ -227,16 +227,17 @@ optional<ProntoData> ProntoProtocol::decode(RemoteReceiveData src) {
 }
 
 void ProntoProtocol::dump(const ProntoData &data) {
-  std::string first, rest;
-  if (data.data.size() < 230) {
-    first = data.data;
-  } else {
-    first = data.data.substr(0, 229);
-    rest = data.data.substr(230);
-  }
-  ESP_LOGI(TAG, "Received Pronto: data=%s", first.c_str());
-  if (!rest.empty()) {
-    ESP_LOGI(TAG, "%s", rest.c_str());
+  std::string rest;
+
+  rest = data.data;
+  ESP_LOGI(TAG, "Received Pronto: data=");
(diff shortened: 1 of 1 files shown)


optimusprimespace:
Can you provide a way to test this? I followed this https://esphome.io/guides/faq.html#how-can-i-test-a-pull-request. I want to test the receiver and the transmitter part of this as well. But i don't know if it's a true external component and can be tested this way or i have to compile it myself. Thanks in advance. 😀

rpaskowitz:
I set up a full development environment, so I compiled and tested via that. I'm not familiar with external components, but the docs seem to indicate you an override even 'internal' components, so pulling in the `remote_base` component from my PR aught to work?

I think the yaml would look like:
```
external_components:
  - source: github://pr#6040
    components: [ remote_base ]
```


Note that this PR only touches _receiving_ code, and not transmitting. I did however notice when trying to transmit the received commands I still encountered issues. I traced this back to the received command containing one extra tailing set of bits. This caused it to fail the consistency check, resulting in the error `Inconsistent data` being logged and no command emitted. When I discovered that the length was wrong, I simply took the last set of bits off the received command and tried again, and that worked fine -- it was a guess that it was the final bits out of place.

I have not tracked down whether the remote is actually emitting those bits, or if there is a separate bug causing them to be printed. If they are actually emitted, then there may be a different bug in the consistency check, or the remote itself is emitting invalid (but functional commands).

I may update the 'Inconsistent data' error to provide more detail about in what way the command is failing the check. In any event, I wouldn't have reached that stage without first getting the untruncated command printed, as the commands were quite lengthy.

jesserockz (approved):

PR #6041, opened 2023-12-30
clang-format and clang-tidy scripts: More robust algorithm to find correct executable

clydebarrow:

The scripts `script/clang-format` and `script/clang-tidy` don't work well on a Mac with executables installed by
homebrew. This PR checks for either the plain executable name that reports the correct version or a versioned name, 

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,5 +23,3 @@ pyparsing >= 3.0
 
 # For autocompletion
 argcomplete>=2.0.0
-clang-tidy==14.0.6
-clang-format==13.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,3 +23,5 @@ pyparsing >= 3.0
 
 # For autocompletion
 argcomplete>=2.0.0
+clang-tidy==14.0.6
+clang-format=13.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,4 +24,4 @@ pyparsing >= 3.0
 # For autocompletion
 argcomplete>=2.0.0
 clang-tidy==14.0.6
-clang-format=13.0.1
+clang-format==13.0.1
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -13,3 +13,4 @@ asyncmock==0.4.2
 hypothesis==5.49.0
 
 clang-format==13.0.1 ; platform_machine != 'armv7l'
+clang-tidy==14.0.6 ; platform_machine != 'armv7l'
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -11,6 +11,3 @@ pytest-mock==3.12.0
 pytest-asyncio==0.23.2
 asyncmock==0.4.2
 hypothesis==5.49.0
-
-clang-format==13.0.1 ; platform_machine != 'armv7l'
-clang-tidy==14.0.6 ; platform_machine != 'armv7l'
--- a/script/clang-format
+++ b/script/clang-format
@@ -102,14 +102,14 @@ def main():
         task_queue.join()
 
     except FileNotFoundError as ex:
-        pass
+        return 1
     except KeyboardInterrupt:
         print()
         print("Ctrl-C detected, goodbye.")
-        os.kill(0, 9)
+        return 2
 
-    sys.exit(len(failed_files))
+    return len(failed_files)
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
--- a/script/clang-format
+++ b/script/clang-format
@@ -34,7 +34,7 @@ def get_binary(name, version):
         f"""
         Oops. It looks like {name} is not installed.
 
-        Please check you can run "{name} -version" or "{name}-{version} -version"
+        Please confirm you can run "{name} -version" or "{name}-{version} -version"
         in your terminal and install
         {name} (v{version}) if necessary.
 
--- a/script/clang-format
+++ b/script/clang-format
@@ -106,7 +106,10 @@ def main():
     except KeyboardInterrupt:
         print()
         print("Ctrl-C detected, goodbye.")
-        return 2
+        # Kill subprocesses (and ourselves!)
+        # No simple, clean alternative appears to be available.
+        os.kill(0, 9)
+        return 2    # Will not execute.
 
     return len(failed_files)
 
--- a/script/clang-format
+++ b/script/clang-format
@@ -13,11 +13,43 @@ import sys
 import threading
 
 
+
+def get_binary(name, version):
+    clang_binary = f"{name}-{version}"
+    try:
+        result = subprocess.check_output([clang_binary, '-version'])
+        if result.returncode == 0:
+            return clang_binary
+    except Exception:
+        pass
+    clang_binary = name
+    try:
+        result = subprocess.run([clang_binary, '-version'], text=True, capture_output=True)
+        if result.returncode == 0 and (f"version {version}") in result.stdout:
+            return clang_binary
+    except Exception as ex:
+        pass
+
+    print(
+        f"""
+        Oops. It looks like {name} is not installed.
+
+        Please check you can run "{name} -version" or "{name}-{version} -version"
+        in your terminal and install
+        {name} (v{version}) if necessary.
+
+        Note you can also upload your code as a pull request on GitHub and see the CI check
+        output to apply {name}
+        """
+    )
+    sys.exit()
+ 
+
 def run_format(args, queue, lock, failed_files):
     """Takes filenames out of queue and runs clang-format on them."""
     while True:
         path = queue.get()
-        invocation = ["clang-format-13"]
+        invocation = [get_binary("clang-format", 13)]
         if args.inplace:
             invocation.append("-i")
         else:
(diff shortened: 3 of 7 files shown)


script/clang-format
```diff
@@ -13,11 +13,43 @@ import sys
...
+        if result.returncode == 0 and (f"version {version}") in result.stdout:
+            return clang_binary
+    except Exception as ex:
+        pass
+
+    print(
+        f"""
+        Oops. It looks like {name} is not installed.
+
+        Please check you can run "{name} -version" or "{name}-{version} -version"
```
kbx81: Nitpicking -- this should read either:

```suggestion
        Please check if you can run "{name} -version" or "{name}-{version} -version"
```

kbx81: ...or (and this would be my preference):

```suggestion
        Please confirm you can run "{name} -version" or "{name}-{version} -version"
```

script/clang-tidy
```diff
@@ -25,6 +25,39 @@ import sys
...
+        print(f"did not find version {version} in output: {result.stdout}")
+    except Exception as ex:
+        print(ex)
+        pass
+
+    print(
+        f"""
+        Oops. It looks like {name} is not installed.
+
+        Please check you can run "{name} -version" or "{name}-{version} -version"
```
kbx81: As above

script/clang-tidy
```diff
@@ -25,6 +25,39 @@ import sys
 import tempfile
 import threading
 
+def get_binary(name, version):
```
bdraco: It would be nice to move this into a module since its duplicate code. Maybe a `util.py`.

script/clang-format
```diff
@@ -13,11 +13,43 @@ import sys
 import threading
 
 
+
+def get_binary(name, version):
```
bdraco:
```suggestion
def get_binary(name: str, version: str) -> str:
```

script/clang-format
```diff
@@ -13,11 +13,43 @@ import sys
...
+
+        Please confirm you can run "{name} -version" or "{name}-{version} -version"
+        in your terminal and install
+        {name} (v{version}) if necessary.
+
+        Note you can also upload your code as a pull request on GitHub and see the CI check
+        output to apply {name}
+        """
+    )
+    sys.exit()
```
bdraco: Ideally functions other than `main` don't ever call `sys.exit()` as it makes tracing execution flow more difficult. It would be better to let this throw and catch it

nielsnl68: Also please add how to install. Just for dummy's like me, who get puzzled how and where something need to be done.

clydebarrow: > Ideally functions other than `main` don't ever call `sys.exit()` as it makes tracing execution flow more difficult. It would be better to let this throw and catch it

I didn't change that. Not trying to refactor the whole thing, just wanted to make it run.

clydebarrow: > Also please add how to install. Just for dummy's like me, who get puzzled how and where something need to be done.

It's installed in `venv/bin` by `script/setup` and then you should have done `source venv/bin/activate` which will put that in your PATH.

script/clang-tidy
clydebarrow: There was already a helpers.py. Moved there.

script/clang-format
clydebarrow: Well, it *wasn't* being installed. It is now.

requirements.txt
```diff
@@ -23,3 +23,5 @@ pyparsing >= 3.0
 
 # For autocompletion
 argcomplete>=2.0.0
+clang-tidy==14.0.6
+clang-format==13.0.1
```
bdraco: I think these are dev only requirements

script/clang-format
```diff
@@ -109,6 +101,8 @@ def main():
         # Wait for all threads to be done.
         task_queue.join()
 
+    except FileNotFoundError as ex:
+        pass
```
bdraco: I think this was originally intended to exit it code 1 on failure with the `return 1`

It looks like at some point it was refactored to not work though

maybe make `main()` return the exit code instead?

```python
if __name__ == "__main__":
    sys.exit(main())
```

script/helpers.py
```diff
@@ -153,3 +153,40 @@ def load_idedata(environment):
...
+
+        Please confirm you can run "{name} -version" or "{name}-{version} -version"
+        in your terminal and install
+        {name} (v{version}) if necessary.
+
+        Note you can also upload your code as a pull request on GitHub and see the CI check
+        output to apply {name}
+        """
+    )
+    raise FileNotFoundError(f"{name} not found")
```
bdraco:
```suggestion
    except FileNotFoundError:
        print(
            f"""
            Oops. It looks like {name} is not installed. It should be available under venv/bin
            and in PATH after running in turn:
            script/setup
            source venv/bin/activate.

            Please confirm you can run "{name} -version" or "{name}-{version} -version"
            in your terminal and install
            {name} (v{version}) if necessary.

            Note you can also upload your code as a pull request on GitHub and see the CI check
            output to apply {name}
            """
        )
        raise
```

script/helpers.py
```diff
@@ -153,3 +153,40 @@ def load_idedata(environment):
...
+    except Exception:
+        pass
+    binary_file = name
+    try:
+        result = subprocess.run(
+            [binary_file, "-version"], text=True, capture_output=True
+        )
+        if result.returncode == 0 and (f"version {version}") in result.stdout:
+            return binary_file
+        print(f"did not find version {version} in output: {result.stdout}")
```
bdraco: Is this debug code?

script/clang-format
```diff
@@ -109,13 +101,15 @@ def main():
         # Wait for all threads to be done.
         task_queue.join()
 
+    except FileNotFoundError as ex:
+        return 1
     except KeyboardInterrupt:
         print()
         print("Ctrl-C detected, goodbye.")
-        os.kill(0, 9)
+        return 2
```
bdraco: I think we still want to kill the subprocess

bdraco:
Looks good.  One remaining comment and I think its good to go 👍

script/clang-format
clydebarrow: Yes - sys.exit() wil stop threads, but I overlooked the fact that there are actually subprocesses spawned. I could not readily find a cleaner way of killing the subprocesses, other than what the original author did.

clydebarrow: There are actually two subprocesses spawned per thread - the clang-tidy executable in venv/ invokes the python module, which runs the LLVM executable.

bdraco:
Looks good.  Will test shortly

script/setup
```diff
@@ -22,3 +22,7 @@ pip3 install --no-use-pep517 -e .
 pre-commit install
 
 script/platformio_install_deps.py platformio.ini --libraries --tools --platforms
+
+echo
+echo
+echo "Virtual environment created; source venv/bin/activate to use it"
```
bdraco, line 28: I don't think this incantation will work in windows... but not sure that matters

bdraco (dismissed):

requirements_test.txt
```diff
@@ -13,3 +13,4 @@ asyncmock==0.4.2
 hypothesis==5.49.0
 
 clang-format==13.0.1 ; platform_machine != 'armv7l'
+clang-tidy==14.0.6 ; platform_machine != 'armv7l'
```
bdraco: We need to skip on aarch64 as well since building the wheel will take hours
```
#15 53.11   Building wheel for clang-tidy (pyproject.toml): started
#15 115.9   Building wheel for clang-tidy (pyproject.toml): still running...
#15 177.1   Building wheel for clang-tidy (pyproject.toml): still running...
#15 238.7   Building wheel for clang-tidy (pyproject.toml): still running...
#15 300.4   Building wheel for clang-tidy (pyproject.toml): still running...
#15 362.7   Building wheel for clang-tidy (pyproject.toml): still running...
```
...

bdraco: https://pypi.org/project/clang-tidy/14.0.6/#files

clydebarrow: Well, firstly it would need to be 'arm64' as that's what machine_platform is set to, and secondly that would defeat the purpose, since I am using an aarch64 machine (M1) so that stops script/setup installing it for me anyway. 

These are the potential solutions I can think of:

1. Is there a way to have a pre-built wheel and install it instead of building on the fly? (I know zilch about Python wheels.)
2. Could the docker images start with a base that has stuff pre-installed? (Would add workload to maintain those.)
3. Move the requirement into a different file that does not get considered by the ci-docker build, but still runs from script/setup.

I will try 3.

clydebarrow: The Dockerfile for the `lint` images installs clang-tidy and clang-format using apt-get, which is obviously much faster, so installing with pip is redundant anyway.

script/setup
clydebarrow: The same instruction is given in https://esphome-docs.web.app/guides/contributing.html#setting-up-development-environment

It's really there to just to remind forgetful devs what's needed. Once we have proper dev docs then this message could just point there.

requirements_test.txt
clydebarrow: Is there any reason that this:
```
    pip3 install \
    --break-system-packages --no-cache-dir -r /requirements.txt -r /requirements_optional.txt \
    && /platformio_install_deps.py /platformio.ini --libraries
```
could not have `--prefer-binary` added - that would speed things up!

bdraco: `--prefer-binary` isn't going to help as there are no aarch64 wheels provided by the distribution.

https://pypi.org/project/clang-tidy/14.0.6/#files

I don't think we need to install it anyways since we aren't using the `pypi` package

bdraco: Since we are calling these as subprocess, it's much better to install via apt, apk, or brew since we know we will get binaries for the system

clydebarrow: In the containers that's ideal (and how it is done) but for a local desktop install pip is good because it's platform independent. Anyway, I think it's good to go now. Gotta call a halt somewhere.

kbx81 (approved):
Tested on my M1 MacBook and confirmed working as expected.

Thanks @clydebarrow for working on this and @bdraco for your reviews & input! 🍻

PR #6042, opened 2023-12-30
GT911 touchscreen: Fix bug causing touch button release to fail

clydebarrow:

Recent changes to the touchscreen components introduced a bug in the GT911 that resulted in a release of a touch button (e.g. the red button on an S3BOX-3) to not be notified.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/gt911/touchscreen/gt911_touchscreen.cpp
+++ b/esphome/components/gt911/touchscreen/gt911_touchscreen.cpp
@@ -14,6 +14,7 @@ static const uint8_t GET_TOUCHES[2] = {0x81, 0x4F};
 static const uint8_t GET_SWITCHES[2] = {0x80, 0x4D};
 static const uint8_t GET_MAX_VALUES[2] = {0x80, 0x48};
 static const size_t MAX_TOUCHES = 5;  // max number of possible touches reported
+static const size_t MAX_BUTTONS = 4;  // max number of buttons scanned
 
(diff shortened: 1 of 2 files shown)


esphome/components/gt911/touchscreen/gt911_touchscreen.cpp
```diff
@@ -79,9 +79,6 @@ void GT911Touchscreen::update_touches() {
     return;
   }
 
-  if (num_of_touches == 0)
-    return;
-
```
nielsnl68, line 84: Perfect.
I really need a test bed with all kind of displays. 

About the button section inside `update_touches()` is it correct that this method sends the button state everytime it runs? 
when so, should we not add a buffer variable that stores the last button statuses and compare the current value with this buffered status and only send the new statuses when they are different?

clydebarrow: It's likely that the listener is doing that anyway, e.g. the binary sensor that is probably the only listener already does caching. Caching here would save a few cycles.

nielsnl68 (approved):

kbx81 (approved):

PR #6044, opened 2024-01-01
analog_threshold: Publish state only when available

### LINKED ISSUE #5325: Analog threshold binary sensor cannot reliably publish initial state
### The problem

In `setup()`, there is a call to `this->sensor_->get_state();`. It is later checked for NaN, and if true, then the initial value is published as `FALSE`.

Code:
https://github.com/esphome/esphome/blob/773cd0f414bf917d150606b2b7b0166d60d98563/esphome/components/analog_threshold/analog_threshold_binary_sensor.cpp#L10

Because only `binary_sensor` implements `publish_initial_state()` (it is not implemented by `sensor`), the value read during `setup()` is almost always going to be NaN. Later, upon the first `update()`, the first real value is collected from the sensor and the state is for the first time valid.

This results in spurious triggers (exactly what `publish_initial_state()` is meant to avoid).

What do you think about not publishing the initial state from `setup()` but instead doing so the first time the underlying sensor has a value? This will avoid the spurious triggers and is a reasonable way to do this kind of threshold given that `sensor` won't have a valid state in `setup()`.

As an alternative, we could implement something like `publish_initial_state()` for `sensor`, but that is a huge project.

2023.12.5

Home Assistant Add-on

2023.12.4

ESP32-IDF

### Board

_No response_

analog_threshold

### Example YAML snippet

```yaml
esphome:
  name: water-meter
  friendly_name: Water meter

esp32:
  board: adafruit_qtpy_esp32s2
  framework:
    type: esp-idf
    platform_version: 6.4.0
    version: 5.1.2
    sdkconfig_options:
      CONFIG_ESP_CONSOLE_USB_CDC: n

time:
  - platform: homeassistant
    id: homeassistant_time
    on_time:
      - seconds: 0
        minutes: 0
        hours: 0
        then:
          - globals.set:
              id: pulse_count
              value: '0'

globals:
  - id: pulse_count
    type: int

i2c:
  sda: 41
  scl: 40
  scan: true
  frequency: 400kHz

sensor:
  - platform: mmc5983
    field_strength_z:
      id: magnet_z
    update_interval: 50ms

  - platform: copy
    source_id: magnet_z
    name: "Magnetic field strength"
    filters:
      - throttle: 1s
    entity_category: DIAGNOSTIC

  - platform: template
    id: total_daily_water
    name: "Total daily water consumption"
    unit_of_measurement: gal
    device_class: water
    state_class: total_increasing
    accuracy_decimals: 3
    update_interval: 1s
    lambda: "return id(pulse_count) / 64.0;"

binary_sensor:
  - platform: analog_threshold
    id: magnet_threshold
    sensor_id: magnet_z
    threshold:
      lower: -0.0137
      upper: 0.0063
    on_state:
      then:
        - lambda: "id(pulse_count)++;"

# Enable logging
logger:

# Enable Home Assistant API
api:

wifi:
  ssid: !secret wifi_ssid
```

_No response_

### Additional information

_No response_

agoode:
This avoids spurious triggers since most sensors won't have state during setup().

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5325

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esphome:
  name: water-meter
  friendly_name: Water meter

esp32:
  board: adafruit_qtpy_esp32s2
  framework:
    type: esp-idf
    platform_version: 6.4.0
    version: 5.1.2
    sdkconfig_options:
      CONFIG_ESP_CONSOLE_USB_CDC: n

time:
  - platform: homeassistant
    id: homeassistant_time
    on_time:
      - seconds: 0
        minutes: 0
        hours: 0
        then:
          - globals.set:
              id: pulse_count
              value: '0'

globals:
  - id: pulse_count
    type: int

i2c:
  sda: 41
  scl: 40
  scan: true
  frequency: 400kHz

sensor:
  - platform: mmc5983
    field_strength_z:
      id: magnet_z
    update_interval: 50ms

  - platform: copy
    source_id: magnet_z
    name: "Magnetic field strength"
    filters:
      - throttle: 1s
    entity_category: DIAGNOSTIC

  - platform: template
    id: total_daily_water
    name: "Total daily water consumption"
    unit_of_measurement: gal
    device_class: water
    state_class: total_increasing
    accuracy_decimals: 3
    update_interval: 1s
    lambda: "return id(pulse_count) / 64.0;"

binary_sensor:
  - platform: analog_threshold
    id: magnet_threshold
    sensor_id: magnet_z
    threshold:
      lower: -0.0137
      upper: 0.0063
    on_state:
      then:
        - lambda: "id(pulse_count)++;"

# Enable logging
logger:

# Enable Home Assistant API
api:

wifi:
  ssid: !secret wifi_ssid

```

## Checklist:

### DIFF
--- a/esphome/components/analog_threshold/analog_threshold_binary_sensor.cpp
+++ b/esphome/components/analog_threshold/analog_threshold_binary_sensor.cpp
@@ -7,29 +7,30 @@ namespace analog_threshold {
 static const char *const TAG = "analog_threshold.binary_sensor";
 
 void AnalogThresholdBinarySensor::setup() {
-  float sensor_value = this->sensor_->get_state();
-
-  // TRUE state is defined to be when sensor is >= threshold
-  // so when undefined sensor value initialize to FALSE
-  if (std::isnan(sensor_value)) {
-    this->publish_initial_state(false);
-  } else {
-    this->publish_initial_state(sensor_value >=
-                                (this->lower_threshold_.value() + this->upper_threshold_.value()) / 2.0f);
-  }
-}
-
-void AnalogThresholdBinarySensor::set_sensor(sensor::Sensor *analog_sensor) {
-  this->sensor_ = analog_sensor;
-
   this->sensor_->add_on_state_callback([this](float sensor_value) {
-    // if there is an invalid sensor reading, ignore the change and keep the current state
-    if (!std::isnan(sensor_value)) {
-      this->publish_state(sensor_value >=
-                          (this->state ? this->lower_threshold_.value() : this->upper_threshold_.value()));
+    if (std::isnan(sensor_value)) {
+      // If there is an invalid sensor reading, ignore the change and keep the current state.
+      // As an alternative, we could call this->invalidate_state() here, but that would be
+      // a breaking change without a config option to enable it.
+      return;
     }
+
+    float threshold;
+    if (!this->has_state()) {
+      // No prior state, use the midpoint of the thresholds.
+      threshold = (this->lower_threshold_.value() + this->upper_threshold_.value()) / 2.0f;
+    } else if (this->state) {
+      // Currently TRUE, use lower_threshold for comparison.
+      threshold = this->lower_threshold_.value();
+    } else {
+      // Currently FALSE, use upper_threshold for comparison.
+      threshold = this->upper_threshold_.value();
+    }
+    this->publish_state(sensor_value >= threshold);
   });
 }
+
+void AnalogThresholdBinarySensor::set_sensor(sensor::Sensor *analog_sensor) { this->sensor_ = analog_sensor; }
 
 void AnalogThresholdBinarySensor::dump_config() {
   LOG_BINARY_SENSOR("", "Analog Threshold Binary Sensor", this);


ssieb:
How is this different from just setting initial publish to false?

agoode:
`publish_initial_state` is already false by default, but the current behavior (prior to this PR) sometimes behaves as if `publish_initial_state` were set to true (depending on the first value coming from the underlying sensor).

Currently:

- In `setup()`, this component will try to read the state of the underling sensor.
- The sensor likely doesn't have a state, since `update()` has not yet been called on the sensor.
- This component treats NaN as `FALSE` and publishes initial state with that value (not triggering anything, correctly).
- The first real sensor data comes in via `update()` and the callback, and if that value evaluates to `TRUE`, triggers are called incorrectly. If the value evaluates to `FALSE`, then no triggers are fired since the component is already primed with a `FALSE` value.
- In my water metering setup, half the time this causes an incorrect trigger at startup, depending on the physical position of the measurement wheel. If the wheel is in a position that evaluates to `TRUE`, the component will report it as if it made a half-rotation, instead of storing it as its initial value.

This PR:

- Always primes the component (without firing triggers) with the actual result of the first sensor reading (`TRUE` or `FALSE`), instead of priming the component with `FALSE` in most cases.
- When the underlying sensor is first read, that will be set as the initial value, so that the next actual change will result in a trigger.

ssieb:
Please move the whole `add_on_state_callback` part to `setup` and I'll merge it.

nagyrobi:
Another reason to have normal support for unknown/unavailable states.

agoode:
I've update the PR but haven't tested yet (I'm away from the hardware for about 2 weeks).

agoode:
I've tested the latest PR on an esp8266 and it has the expected behavior. I think it is good to merge. Thanks!

Triodes (commented):
So I have looked at this proposed solution a bit, and it only solves half the problem. I'll put some tables below with different scenario's for when the esp device boots and the problems that arise with both the current implementation and the solution proposed in this PR.

I'll be using ✅ for TRUE and ❌ for FALSE
All of these examples assume that the sensor value stays the same during the reboot.

### Current Situation

| State before reboot | current sensor value | initial state | next state | problem |
| --- | --- | --- | --- | --- |
| ❌ | ⬇️ below lower threshold | ❌ | ❌ | No problems
| ❌ | ➡️ between thresholds | ❌ | ❌ | No problems
| ✅ | ➡️ between thresholds | ❌ | ❌ | the state change from ✅ to ❌ that would have occurred on the next drop below the lower threshold is skipped |
| ✅ | ⬆️ above upper threshold | ❌ | ✅ | an extra state change from ❌ to ✅ is triggered on boot |

### Proposed solution

Now in the proposed solution the initial state that is published is based of the first actual sensor value. The resulting TRUE of FALSE is based on a comparison with the `this->upper_threshold_` because `this->state` is always initialized to `false` on boot. 

https://github.com/esphome/esphome/blob/9aa1ed1dd2fbfefd0d296a306aeda85dff841bf4/esphome/components/analog_threshold/analog_threshold_binary_sensor.cpp#L13-L19

Therefore we now get the following table:

| State before reboot | current sensor value | initial state | next state | problem |
| --- | --- | --- | --- | --- |
| ❌ | ⬇️ below lower threshold | ❌ | ❌ | No problems
| ❌ | ➡️ between thresholds | ❌ | ❌ | No problems
| ✅ | ➡️ between thresholds | ❌ | ❌ | the state change from ✅ to ❌ that would have occurred on the next drop below the lower threshold is skipped |
| ✅ | ⬆️ above upper threshold | ✅ | ✅ | No problems |

### Context of sensor setup

Now I only care about the state change from ❌ to ✅, since that is the moment I increment my pulse counter for a full rotation. So this PR solves that problem. 

@agoode will still have a problem if his device reboots when the sensor value is between the thresholds when it was moving towards the lower threshold. In this case a half rotation of his water meter will be skipped.

### Fixing both problems

I think fixing both problems can only be done by sending an initial state of i.e. `UNKNOWN` when the sensor value is **between** the thresholds on boot. Then when either of the thresholds is crossed the corresponding state change to ❌ or ✅ can be reported correctly.

Another thought on this is that you will still have false state changes if the sensor value is jittering around one of the thresholds on boot. Which is much less likely, but still possible. The hysteresis meant for inherently noisy signals after all...

### Final thoughts

Even though this PR only solves part of the full problem, it would solve mine. So if this could be merged that would be great!

P.S. @agoode Thanks for taking the time to make a fix for this!

agoode:
Hmm, this might need to be updated given the changes in #8961. I'll have to think about it, unless anyone else has an idea.

agoode:
I think the callback should be something like this instead:

```
    if (std::isnan(sensor_value)) {
      this->invalidate_state();
    } else {
      this->publish_state(sensor_value >= (this->state ? this->lower_threshold_.value() : this->upper_threshold_.value()));
    }
```

agoode:
Ok, I am pretty happy with the current code. It's much more straightforward now thanks to the improvements from #8961.

esphome/components/analog_threshold/analog_threshold_binary_sensor.cpp
```diff
@@ -7,30 +7,31 @@ namespace analog_threshold {
...
+      // If there is an invalid sensor reading, ignore the change and keep the current state.
+      // As an alternative, we could call this->invalidate_state() here, but that would be
+      // a breaking change without a config option to enable it.
+      return;
     }
+
+    float threshold;
+    if (!this->has_state()) {
+      // No prior state, use the midpoint of the thresholds.
+      threshold = (this->lower_threshold_.value() + this->upper_threshold_.value()) / 2.0f;
```
agoode, line 21: Note this currently replicates the existing behavior of the code. @Triodes suggests only setting a state if it is outside both thresholds, which I could do here but would maybe prefer to do in a subsequent PR.

PR #6049, opened 2024-01-01
Touchscreen: Fix multiple release callback bug

clydebarrow:

The touchscreen component was sending continuous release callbacks when nothing was pressed.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


nielsnl68 (approved):

clydebarrow:
Found another bug (would only affect multiple touch situation.)

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -110,7 +110,8 @@ void Touchscreen::send_touches_() {
   } else {
     TouchPoints_t touches;
     for (auto tp : this->touches_) {
-      touches.push_back(tp.second);
+      if (tp.second.state == STATE_PRESSED || tp.second.state == STATE_UPDATED)
```
nielsnl68, line 116: Imho, we need to send the release action as well, the using function need to monitor the `state` flag it self to see of the button is still pressed or not. so it can handle the right action.
That said; maybe we need to call `on_update` also when no more touches are registered.

nielsnl68 (changes requested):

esphome/components/touchscreen/touchscreen.cpp
clydebarrow: > maybe we need to call `on_update` also when no more touches are registered

This^

clydebarrow:
Hmm, Github is not reflecting my latest commits.

esphome/components/touchscreen/touchscreen.cpp
```diff
@@ -94,18 +96,25 @@ void Touchscreen::add_raw_touch_position_(uint8_t id, int16_t x_raw, int16_t y_r
...
-      listener->release();
-    this->touches_.clear();
+    if (this->was_touched_) {
+      this->release_trigger_.trigger();
+      for (auto *listener : this->touch_listeners_)
+        listener->release();
+      this->touches_.clear();
+      for (auto *listener : this->touch_listeners_) {
+        listener->update(touches);
+      }
```
nielsnl68, line 111: After reading the method again i think we should change a bit more. 
```c++
  TouchPoints_t touches;
  for (auto tp : this->touches_) {
      touches.push_back(tp.second);
  }
  if (this->need_update_ || ( !this->is_touched_ && this->was_touched_)) {
    this->update_trigger_.trigger(touches);
    for (auto *listener : this->touch_listeners_) {
      listener->update(touches);
    }
  }
  if (!this->is_touched_) {
    if (this->was_touched_) {
      if (this->touch_timeout_ > 0) {
        this->cancel_timeout(TAG);
      }
      this->release_trigger_.trigger();
      for (auto *listener : this->touch_listeners_)
        listener->release();
      this->touches_.clear();
      this->was_touched_ = false;
    }
  } else {
    if (this->first_touch_) {
      TouchPoint tp = this->touches_.begin()->second;
      this->touch_trigger_.trigger(tp, touches);
      for (auto *listener : this->touch_listeners_) {
        listener->touch(tp);
      }
    }
  }
}
```
i'm not 100% happy with this function yet. But it should work, i think.

nielsnl68 (changes requested):

esphome/components/touchscreen/touchscreen.cpp
clydebarrow: The problem is that nobody is currently using the multi-touch feature AFAIK, nor do we have a clear specification of how it is expected to work, so randomly tweaking the behaviour is unlikely to be useful when someone does come to use it. I'm not particularly happy with the code either, not least because it's hard to understand what it's doing, and what the various state bits mean (some of them are set but never tested.)

Right now the major problem is that continual release events get sent, which is clearly wrong. Maybe we should just fix that (my first commit) and leave the rest until we have a plan for how it should work.

nielsnl68: I;m already working on mote multi touch support. It is still on paper but it will be there soon.
There are just a couple of projects that i need to finish first. But when done it would be cool.

clydebarrow: Sounds good - having it on paper to start helps a lot. I'll revert to the simple fix.

nielsnl68: Please use the method as i shown above. that should fix the issue and my own issue as well.

clydebarrow: The problem with that is that `update()` continues to report touches that have been released. So if there are two touches, and one is lifted, the next update reports one with state UPDATED and one with RELEASED, but continues to do so until the remaining touch is released. Just like with the released callback, a release for a touch should only be delivered once.

Also on some occasions a touch is reported with both UPDATED and RELEASING set. What does this mean? For the touch consumer having to interpret these states that have no description as to what they mean is difficult, whereas if only touches that were still ongoing were reported, it can easily keep track of which have been released.

nielsnl68: then there is something else wrong. 
I will look of i can reproduce what you are telling me now.

clydebarrow: I think the update() call should just return the touches that are in PRESSED or UPDATED state only. And send an empty list on last release.

nielsnl68:
This PR is being included into #5997

nagyrobi:
Indeed, with https://github.com/esphome/esphome/pull/5997 this is not needed anymore.

PR #6052, opened 2024-01-03
Add an optional binary_sensor to digital signals components

dianlight:

This PR add the option to use a "Digital Signal" component also as binary sensor.

The need arises from the automation of a BFT gate where the SCA (Gate Open Indicator) signal is open (HIGH) if the gate is open, closed (LOW) if the gate is full close and blink (intermittent LOW/HIGH) if the gate is closing.

Ref: [BFT PHEBE TECNICAL MANUAL](https://www.atecnica.it/file/manuale-istruzioni-bft-vega-centrale-scheda-elettronica-di-comando-bft-phebe-basculante-i700062-10001.pdf)

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3514

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [x] BK72xx
- [x] RTL87xx

```yaml
- platform: pulse_counter
    pin:
      number: D7
      mode: INPUT_PULLUP
      inverted: True
    id: sca_pulse
    name: "Pulse Counter"
    update_interval: 1s
    binary_sensor:
        id: cover_open
        internal: True    
    filters:
      - filter_out: nan
      - throttle_average: 2s
```

## Checklist:

### DIFF
--- a/esphome/components/pulse_counter/pulse_counter_sensor.cpp
+++ b/esphome/components/pulse_counter/pulse_counter_sensor.cpp
@@ -180,6 +180,13 @@ void PulseCounterSensor::update() {
     ESP_LOGD(TAG, "'%s': Total : %" PRIu32 " pulses", this->get_name().c_str(), current_total_);
     this->total_sensor_->publish_state(current_total_);
   }
+
+  if (this->binary_sensor_ != nullptr) {
+    bool level = this->storage_.pin->digital_read();
+    ESP_LOGD(TAG, "'%s': Sending state %s", this->get_name().c_str(), ONOFF(level));
+    this->binary_sensor_->publish_state(level);
+  }
+
   this->last_time_ = now;
 }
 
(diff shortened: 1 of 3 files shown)


jesserockz (changes requested):
Thanks for your contribution. I see where you are going with this, but I am not sure the `pulse_counter` is the right platform for your use case. 

But as for my review of the actual changes:

The main thing I see wrong here is that it's not even a binary sensor, its a regular sensor...

If it should be a represent the pin state, it should be in its own `binary_sensor` platform file with a link to the main parent sensor to register itself to be published to.

In the end, the YAML should look something like this:
```yaml
...
sensor:
  - platform: pulse_counter
    id: my_pc
    ...
    
binary_sensor:
  - platform: pulse_counter
    pulse_counter_id: my_pc
    name: PC Pin State
```

dianlight:
@jesserockz 

In this way, however, I would have `binary_sensors` which are called `pulse_counter` because they depend on the `pulse_counter` in Digital Signals.

Isn't it better at this point to make a `binary_component` digital_signal that accepts any sensor among the digital signals?

like this:
 
 ```yaml
 ...
 sensor:
   - platform: pulse_counter
     id: my_pc
...
   - platform: pulse_meter
     id: my_pm
     ...
     
 binary_sensor:
   - platform: digital_signal
     digital_signal_id: my_pc
     name: PC Pin State
...
   - platform: digital_signal
     digital_signal_id: my_pm
     name: PM Pin State
 ```

swoboda1337:
Stale

PR #3171, opened 2022-02-07
Adding RC522 Gain support

egroen:
# What does this implement/fix? 
Adds RC522 Gain support

Quick description and explanation of changes
RC522.h
Added variable
Added setter

RC522.cpp
Push gain on initialize
Print gain value on config_dump

__init__.py
Added gain optional, might need a check, keeps complaining 'invalid option'

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>
https://github.com/esphome/feature-requests/issues/1576

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266

## Example entry for `config.yaml`:
```yaml
rc522_spi:
  ...
  gain: 6
```

## Checklist:
  - [X] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).
  
If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6063, opened 2024-01-08
[rc522] Allow setting gain from config

### LINKED ISSUE #1576: RC522 Gain control configuration
Adding a configuration option for the RC522 to change the gain (`RF_CFG_REG`)

When the default gain isn't viable, it can be increased for better penetration

I've manually added this locally to increase range through thin wooden wall for a small receiver tag, lost code due to update
But in fact nothing more than a `pcd_write_register(RF_CFG_REG, [ConfigValue]);` on the proper location

**Additional context**
It's a small addition, where default value is 0x48
So possible config value 0-7, default 4. Register = ((reg&0x07) << 4 ) | 0x08
Register layout:
```
bit: Desc        Default
  7: Reserved    [b0]
6-4: RxGain      [b100]
3-0: Reserved    [b1000]
```

selamanse:

Adds RC522 Gain support.

Basically took changes from https://github.com/esphome/esphome/pull/3171 and applied dev change requests.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [#1576](https://github.com/esphome/feature-requests/issues/1576)

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
spi:
  - id: spi_rc522
    clk_pin: 16
    mosi_pin: 17
    miso_pin: 15

rc522_spi:
  id: rc522_nfcc
  cs_pin: 12
  gain: 48dB

binary_sensor:
  - platform: rc522
    rc522_id: rc522_nfcc
    name: PN532 NFC Tag
    uid: 74-10-37-94

```

## Checklist:


jesserockz (changes requested):
There is no correlation between the configured value (0-7) and the dB valued that is logged. 
Ideally, the gain config variable should be an `enum` with the actual dB values that the user can choose from.

Also please look at how other code use LogString for logging the enums.

selamanse:
@jesserockz thanks for pointing this out. I have now updated the code to be more transparent in the requested manner. Please bear with me since I'm not so firm with c++ language but I tried my best by aligning with existing implementations of enum and LogString usage. 

I have also tested this version on an esp32 dev board with rc522 attached

selamanse:
@jesserockz just a friendly reminder that I have fixed the issues you mentioned.

selamanse:
@glmnet @jesserockz friendly reminder.

gabriel403:
@glmnet @jesserockz I'm very interested in this change too, what can we do to get this merged?

.gitignore
kbx81, line 1: Please remove these changes from this PR.

esphome/components/rc522/__init__.py
```diff
@@ -34,6 +57,7 @@
                 cv.GenerateID(CONF_TRIGGER_ID): cv.declare_id(RC522Trigger),
             }
         ),
+        cv.Optional(CONF_GAIN, default="38dB"): validate_gain,
```
kbx81: Couldn't this just use `cv.decibel`, instead?

tests/test1.yaml
kbx81, line 1: Please update the tests in the `/tests/components` directory, instead

esphome/components/rc522/__init__.py
```diff
@@ -7,6 +7,7 @@
     CONF_ON_TAG_REMOVED,
     CONF_TRIGGER_ID,
     CONF_RESET_PIN,
+    CONF_GAIN,
```
kbx81: Please alphabetize the constants

esphome/components/rc522/__init__.py
selamanse: sure thing

kbx81:
Please also make sure all CI/tests are passing. 🙂

esphome/components/rc522/__init__.py
```diff
@@ -34,6 +47,7 @@
                 cv.GenerateID(CONF_TRIGGER_ID): cv.declare_id(RC522Trigger),
             }
         ),
+        cv.Optional(CONF_GAIN, default="38dB"): cv.enum(GAIN, string=True),
```
selamanse, line 50: I can't use cv.decibel here since a pure float based decibel value can never be used here (without the mapping logic). the rc522 only allows the custom fixed values that GAIN enum exposes (according to datasheet). so I decided to use cv.enum without any number validating logic anymore.

kbx81: You can use more than one validator

kbx81: Something like [this](https://github.com/esphome/esphome/blob/dceab6ce29493e67caaa8cdadf8cf7f4df063635/esphome/components/http_request/__init__.py#L108-L112)

selamanse: @kbx81 can you please elaborate what your suggestion is?

jesserockz: I think this is fine as is given it also has two special dBA values

esphome/components/rc522/rc522.cpp
```diff
@@ -37,6 +37,27 @@ std::string format_uid(std::vector<uint8_t> &uid) {
...
+    case RC522Gain::RC522_GAIN_38DB:
+      return LOG_STR("38 decibels");
+    case RC522Gain::RC522_GAIN_43DB:
+      return LOG_STR("43 decibels");
+    case RC522Gain::RC522_GAIN_48DB:
+      return LOG_STR("48 decibels");
+    default:
+      return LOG_STR("UNKNOWN");
+  }
+}
```
jesserockz: This really should show the value as close to the config entered as possible.
```suggestion
const LogString *rc522_gain_to_string(RC522Gain gain) {
  switch (gain) {
    case RC522Gain::RC522_GAIN_18DB:
      return LOG_STR("18dB");
    case RC522Gain::RC522_GAIN_18DBA:
      return LOG_STR("18dB_a");
    case RC522Gain::RC522_GAIN_23DB:
      return LOG_STR("23dB");
    case RC522Gain::RC522_GAIN_23DBA:
      return LOG_STR("23dB_a");
    case RC522Gain::RC522_GAIN_33DB:
      return LOG_STR("33dB");
    case RC522Gain::RC522_GAIN_38DB:
      return LOG_STR("38dB");
    case RC522Gain::RC522_GAIN_43DB:
      return LOG_STR("43dB");
    case RC522Gain::RC522_GAIN_48DB:
      return LOG_STR("48dB");
    default:
      return LOG_STR("UNKNOWN");
  }
}
```

jesserockz (changes requested):

selamanse:
@kbx81 fyi: I added docs here https://github.com/esphome/esphome-docs/pull/3979

selamanse:
superseeded by #7539

PR #6067, opened 2024-01-09
VEML7700 and VEML6030 light sensors

latonita:

Implements component for ambient light sensors from Vishay - VEML7700 and VEML6030

## Types of changes

- [ ] Other

https://github.com/esphome/feature-requests/issues/2493
https://github.com/esphome/feature-requests/issues/218
https://github.com/esphome/feature-requests/issues/1915

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3525

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esphome:
  name: veml7700-test

esp32:
  board: esp32dev
  framework:
    type: arduino

# esp8266:
#   board: nodemcuv2

logger:
  level: DEBUG

i2c:

sensor:
  - platform: veml7700
    address: 0x10
    update_interval: 30s
    # lux_compensation: false
    # glass_attenuation_factor: 1.10
    # auto_mode: false
    # gain: 1x
    # integration_time: 100ms
    ambient_light:
      name: "Ambient light"
    full_spectrum:
      name: "Full spectrum"
    infrared:
      name: "Calculated Near-IR"
    actual_integration_time:
      name: "Actual integration time"
    actual_gain:
      name: "Actual gain"
    # ambient_light_counts:
    #   name: "Ambient light counts"
    # full_spectrum_counts:
    #   name: "Full spectrum counts"

```

## Checklist:


esphome/components/veml7700/sensor.py
```diff
@@ -0,0 +1,177 @@
...
+    ICON_TIMER,
+    DEVICE_CLASS_ILLUMINANCE,
+    STATE_CLASS_MEASUREMENT,
+)
+
+CODEOWNERS = ["@latonita"]
+DEPENDENCIES = ["i2c"]
+
+UNIT_COUNTS = "#"
+ICON_GAIN = "mdi:multiplication"
```
jesserockz: Please name consts with what actually represents the value
```suggestion
ICON_MULTIPLICATION = "mdi:multiplication"
```

esphome/components/veml7700/sensor.py
```diff
@@ -0,0 +1,177 @@
...
+    .extend(i2c.i2c_device_schema(0x10)),
+    cv.has_at_least_one_key(
+        CONF_AMBIENT_LIGHT,
+        CONF_FULL_SPECTRUM,
+        CONF_INFRARED,
+        CONF_AMBIENT_LIGHT_COUNTS,
+        CONF_FULL_SPECTRUM_COUNTS,
+        CONF_ACTUAL_GAIN,
+        CONF_ACTUAL_INTEGRATION_TIME,
+    ),
```
jesserockz: I know this might seem strange, but we do not enforce the use of any sensors etc in ESPHome
```suggestion
```

esphome/components/veml7700/veml7700.cpp
```diff
@@ -0,0 +1,364 @@
...
+    default:
+      ms = 100;
+  }
+  return ms;
+}
+
+void wait_for_device(IntegrationTime time, uint16_t mult) {
+  delay(mult * get_itime_ms(time));
+  App.feed_wdt();
+}
```
jesserockz: This is a big no. Components should not delay for any significant amount. 
The component should be reworked to use `this->set_timeout(fn, milliseconds)` which will call the function sometime after the milliseconds has elapsed without blocking the rest of the application.

jesserockz (changes requested):
Thank you for your contribution.

I got as far as `wait_for_device` and decided to stop reviewing there as it will need quite a bit of reworking to be accepted in.

Please take a look at my comments below.

esphome/components/veml7700/sensor.py
kbx81: Marked as resolved, but it seems this change was not made?

esphome/components/veml7700/sensor.py
kbx81: As above, this was marked as resolved, but it seems this change was not made?

esphome/components/veml7700/sensor.py
jesserockz: Indeed

esphome/components/veml7700/sensor.py
jesserockz: Correct

esphome/components/veml7700/veml7700.h
```diff
@@ -0,0 +1,198 @@
...
+
+enum PSM : uint8_t {
+  PSM_MODE_1 = 0,
+  PSM_MODE_2 = 1,
+  PSM_MODE_3 = 2,
+  PSM_MODE_4 = 3,
+};
+
+#pragma pack(push)
+#pragma pack(1)
```
jesserockz: Please change to use `__attribute__((packed));`

jesserockz (dismissed):

esphome/components/veml7700/sensor.py
latonita: apologies, merge glitch. fixed now.

esphome/components/veml7700/sensor.py
latonita: apologies, merge glitch. fixed now.

esphome/components/veml7700/veml7700.h
latonita: Done.

mrtoy-me:
Hi @latonita
I have tested the component on veml6030 seems to work fine ... but I will do some more testing overnight and let you know if I pick up anything.

Two suggestions I have 
1.
I was getting some compile warnings around the packed ConfigurationRegister for IntegrationTime ALS_IT relating to gcc4.4
From what I have read, it related to the compiler complaining about the packed structure crossing a type boundary with uint8_t.
In the .h file, I changed enum IntegrationTime : uint16_t rather than uint8_t and no more warnings :) - makes no difference to packed structure anyway.
2.
There are plenty of  ESP_LOGD messages in the component and I can understand why they are there when you were debugging but now it is working it makes the logs pretty noisy - I changed all the ESP_LOGD to ESP_LOGV so you can get the messages as verbose if you want and in debug there are enough messages on what values are being sent.

They are minor changes so I didn't do a PR to you, but I think they are worth you considering.

Anyway thanks a lot for developing the component - great work.

latonita:
@mrtoy-me thanks for taking a look and testing
1. I don't believe its a good thing, it just a hack which works due to specific bit positions in the struct. Stupid compilation note from the GCC drives people craze since 2010... Moreover another GCC bug doesn't let to suppress this with `#pragma GCC diagnostic ignored "-Wxxxxx"` :) It was fixed in 2022 in GCC 13, but extensa-gcc is based on gcc8 i believe.
What works properly is `#pragma pack(push), #pragma pack(1) ..... #pragma pack(pop)` but this was removed by request from @jesserockz .
2. Good catch, thanks! I actually forgot to change them to LOGV. Fixed.

esphome/components/veml7700/veml7700.cpp
latonita: fixed, FSM implemented

kbx81 (approved):
LGTM -- thanks for working on this!

latonita:
> LGTM -- thanks for working on this!

Thanks. There are several more PRs to review! :)

PR #6071, opened 2024-01-10
Add serial EEPROM component

pebblebed-tech:

Adds a component that enables the use of serial EEPROM's such 24LC32. The component can be used by lambda's . Its supports reading and writing the following types
 * unit8, unit16, unit32
* Float, Double
* Objects
* Strings upto 254 chars in length.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3535

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
external_eeprom:
  id: ext_eeprom_component_1
  address: 0x57
  ee_memory_type: 24XX32
  i2c_buffer_size: 128
  i2c_id: bus_1
```
## Example usage:
```uint8_t myValue1 = 42;
ext_eeprom_component_1->write8(15, myValue1);
uint8_t myRdValue1;
myRdValue1 = ext_eeprom_component_1->read8(15);
ESP_LOGD("Eeprom", "I read: %d",myRdValue1 );
```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs#3535).


mrkvo:
Hi, I have just tested this feature and its working like a charm. 
Tested with 24C02, writing and reading back int32 and floats. 

Many thanks for this!!!

pebblebed-tech:
> Hi, I have just tested this feature and its working like a charm. Tested with 24C02, writing and reading back int32 and floats.
> 
> Many thanks for this!!!

Hi you are very welcome. Thank you for posting a comment 😊

esphome/components/external_eeprom/external_eeprom.cpp
```diff
@@ -0,0 +1,443 @@
...
+}
+/// @brief Bulk read from the device
+/// @note breaking up read amt into 32 byte chunks (can be overriden with setI2Cbuffer_size)
+/// @note Handles a read that straddles the 512kbit barrier
+/// @param memaddr is the starting location to read
+/// @param buff is the pointer to an array of bytes that will be used to store the data received
+/// @param buffer_size is the size of the buffer and also the number of bytes to be read
+void ExtEepromComponent::read(uint32_t memaddr, uint8_t *buff, uint16_t buffer_size) {
+  ESP_LOGVV(TAG, "Read %d bytes from address %d", buffer_size, memaddr);
+  uint32_t size = buffer_size;
```
kbx81: Should this be a `uint16_t` so as to be consistent with the `buffer_size` parameter?

esphome/components/external_eeprom/external_eeprom.cpp
```diff
@@ -0,0 +1,443 @@
...
+  uint8_t *p = buff;
+  i2c::ErrorCode ret;
+  while (size >= 1) {
+    // Limit the amount to read to a page size
+    uint16_t amt_to_read = size;
```
kbx81, line 96: `buffer_size` is a `uint16_t`, which gets assigned to `uint32_t size` above, but here we're back to a `uint16_t`. haha
This assignment _could_ be risky, although I don't think it would actually cause an issue given what you've got going on here in general.

esphome/components/external_eeprom/external_eeprom.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+  uint16_t max_write_size = this->memory_page_size_bytes_;
+  if (max_write_size > this->i2c_buffer_size_)
+    max_write_size = this->i2c_buffer_size_;
+
+  /// Break the buffer into page sized chunks
+
+  while (size >= 1) {
+    /// Limit the amount to write to either the page size or the I2C limit
+    uint16_t amt_to_write = size;
```
kbx81, line 222: Same issue I mentioned above

kbx81 (commented):
A couple little comments, but otherwise LGTM.

Would you mind changing the tests over to the new format? I have a number of PRs open if you need an example -- [here is one](https://github.com/esphome/esphome/pull/6176/files).

esphome/components/external_eeprom/external_eeprom.cpp
pebblebed-tech: updated to ensure consistancy

esphome/components/external_eeprom/external_eeprom.cpp
pebblebed-tech: Updated size to unit16_t

esphome/components/external_eeprom/external_eeprom.cpp
pebblebed-tech: Updated to ensure consistancy

pebblebed-tech:
@kbx81 All corrections applied and new testing framework tests added, do I remove the old tests or leave them be for now?

kbx81:
> @kbx81 All corrections applied and new testing framework tests added, do I remove the old tests or leave them be for now?

Yes, it's ok to remove them from the original test files now.

pebblebed-tech:
@kbx81 All updates as requested have been completed. Please could this now be merged in dev. Reason for keenness is this component is a pre req for the rtc scheduler I'm working on, but I can't create a PR until this in dev otherwise the eeprom would be in 2 PR's. Thanks

esphome/components/external_eeprom/external_eeprom.cpp
```diff
@@ -0,0 +1,443 @@
...
+/// @brief Read a std::string from the device
+/// @note It write the string with an extra byte containing the size at the memaddr
+/// @note It is limited to reading a max length of 254 bytes
+/// @param memaddr is the starting location to read
+/// @param str_to_read will hold the bytes read from the device on return of the fuction
+
+uint32_t ExtEepromComponent::read_string_from_eeprom(uint32_t memaddr, std::string &str_to_read) {
+  uint8_t new_str_len = read8(memaddr);
+  uint8_t data[new_str_len + 1];
+  read(memaddr + 1, (uint8_t *) data, new_str_len);
```
kbx81:
```suggestion
  this->read(memaddr + 1, (uint8_t *) data, new_str_len);
```
I see a few more of these below -- please be sure to include the `this->` prefix.
Sorry for not catching this earlier!

kbx81:
Sorry for the delay -- what I've been wondering about is accessing the eeprom and data stored in it...if there are perhaps ways we could make this easier and/or doable via yaml alone. 🤔

I don't want to suggest making this PR bigger, but as it is it seems like it only provides a way to store a single object (or, perhaps an array of them?) into the eeprom, which feels a bit limiting. 😇 I realize there is a lot of complexity in storing multiple objects (perhaps getting into some kind of mini-filesystem territory) and doing so is arguably outside of the scope of the initial hardware support this PR adds.

I'm just curious, what is your intended use case?

mrkvo:
For me its storing integration sensor value.

kbx81:
@mrkvo could you elaborate on that? I'm just curious why an external eeprom is necessary for this -- it seems to me that you could store this value as a [global variable](https://esphome.io/guides/automations#global-variables) with `restore_value` set to `true`, eliminating the need for the extra hardware component.

I'm not questioning the existence of this component -- I'm more interested in learning about its use cases so I can better visualize ways we might make it easier to use and/or better integrated with ESPHome.

pebblebed-tech:
Hi Keith,
My use case is to store schedules for a  rtc scheduler component. It will
eventually allow a user to set a schedule using HA or via the device, that
would be stored in eeprom. So that after a power fail the schedule is
retained regardless of the state of HA or network. This to ensure the
schedule events will always run on time as long as power is present. Eg
keeping it local and removing complexity.

I can see others using it to store series data over a period of time in
remote locations with intermittent conectivity.
Another use case would be storing blob data for displays or sound files.
This would be particularly pertinent to esp8266 devices with a small amount
of flash.
Thanks
Paul

esphome/components/external_eeprom/external_eeprom.cpp
pebblebed-tech: Have updated the code to resolve this and the issues below. Thank You

RoganDawes:
I have a Sherlotronics ComplexRX Keeloq receiver (running on an STM32 using my ESPHome port) that has an EEPROM for storing enrolled transmitters, and updating sequence numbers to prevent replay attacks. This would be perfect for that, as well as storing logs. I do need to find the time to port the I2C drivers to STM32, though, before I can test this.

Just another use case. Thanks for creating this!

pebblebed-tech:
Hi @kbx81 any chance is completing the review and merging with dev so I can create my other PR for the real time clock based scheduler. Thanks Paul

kbx81:
Hi @pebblebed-tech...sorry for the delay. 😇

I've discussed & brainstormed a bit with Jesse and we'd like to make this a platform under a more generic "storage" component. In general, something like this:

```yaml
storage:
  - platform: eeprom  # this PR
    ...
  - platform: sd_card  # future platform for SD card use
    ...
  - platform: rtc_storage  # future platform for storing data on RTCs that also have an EEPROM
    ...
```

Doing so allows a bit more flexibility and it's easier to avoid breaking changes later down the line -- especially since there has been some recent discussion on the Discord server about implementing support for storage on SD cards.

In an effort to not make this PR any bigger, let's get a separate PR going to add the generic `storage` component and then this PR can be refactored (shouldn't take much!) to be made into a platform that'll fall under `storage`.

Later, we can build out an API as well as some ESPHome automations (actions/triggers/etc.) to make it all easier to use, ideally without lambdas or any C code in general. Hopefully this makes sense? Let us know if you have any questions. 😄

RoganDawes:
Still interested in this PR, but it turns out that my target has SPI EEPROM, not i2c. Would be great to consider that as an alternative transport bus in the design of this component.

A bit unsure about including SD cards as an equivalent storage target. EEPROM and the rtc have similar fixed capacity raw storage, while more often, SD cards have a filesystem, typically FAT, and files of various names, which can grow as needed. The order of magnitude size difference is also worth considering, I think.

PR #6083, opened 2024-01-12
fix negative temperature for pmsx003

### LINKED ISSUE #3814: Incorrect readout for negative values for PMS5003T
### The problem

When the temperature dips below 0.0 °C the readout inverts and goes to 6553,5 °C. This equals the value 65535 without rounding.

When investigating the [source code](https://github.com/esphome/esphome/blob/d8746266625c5a4c9127577297876e445cffe859/esphome/components/pmsx003/pmsx003.cpp#L198), it can be seen that an unsigned 16-bit integer is used, which is *not* compatible with getting negative values.

So the solution would be reading out the temperature as a signed integer.

2022.11.1

Home Assistant Add-on

2022.11.3

ESP32

### Board

az-delivery-devkit-v4

pmsx003

### Example YAML snippet

_No response_

_No response_

### Additional information

I have attached an example temperature readout below: 
grafik

This issue has also been previously reported here:  #3013

ssieb:

The current component doesn't handle negative temperatures.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/3814

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/pmsx003/pmsx003.cpp
+++ b/esphome/components/pmsx003/pmsx003.cpp
@@ -195,7 +195,7 @@ void PMSX003Component::send_command_(uint8_t cmd, uint16_t data) {
 void PMSX003Component::parse_data_() {
   switch (this->type_) {
     case PMSX003_TYPE_5003ST: {
-      float temperature = this->get_16_bit_uint_(30) / 10.0f;
+      float temperature = (this->get_16_bit_uint_(30) / 10.0f;
       float humidity = this->get_16_bit_uint_(32) / 10.0f;
 
       ESP_LOGD(TAG, "Got Temperature: %.1f°C, Humidity: %.1f%%", temperature, humidity);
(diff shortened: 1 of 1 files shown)


esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -195,7 +195,7 @@ void PMSX003Component::send_command_(uint8_t cmd, uint16_t data) {
 void PMSX003Component::parse_data_() {
   switch (this->type_) {
     case PMSX003_TYPE_5003ST: {
-      float temperature = this->get_16_bit_uint_(30) / 10.0f;
+      float temperature = (this->get_16_bit_uint_(30) / 10.0f;
```
ssieb:
```suggestion
      float temperature = (int16_t) this->get_16_bit_uint_(30) / 10.0f;
```

kbx81 (approved):

PR #6085, opened 2024-01-12
add support for samples per second via data_rate option to ADS1115

re-ynd:

Add samples per second feature to ADS 1115. Before this change, there was no way to change the data rate (samples per second) and ADS 1115 was always configured to use 860 samples per second (SPS). With this change, user can configure the SPS to all the available options in ADS 1115 by setting `data_rate` in yaml. I set the default value to 860 SPS so older projects will not be affected.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
ads1115:
  - address: 0x48
sensor:
  - platform: my_ads1115
    multiplexer: 'A0_GND'
    gain: 6.144
    data_rate: 'ADS1115_8_SPS'
```

## Checklist:

### DIFF
--- a/esphome/components/ads1115/sensor.py
+++ b/esphome/components/ads1115/sensor.py
@@ -55,6 +55,7 @@ DATA_RATE = {
     "ADS1115_860_SPS": ADS1115DataRate.ADS1115_860_SPS,
 }
 
+
 def validate_gain(value):
     if isinstance(value, float):
         value = f"{value:0.03f}"
--- a/esphome/components/ads1115/sensor.py
+++ b/esphome/components/ads1115/sensor.py
@@ -5,6 +5,7 @@ from esphome.const import (
     CONF_GAIN,
     CONF_MULTIPLEXER,
     CONF_RESOLUTION,
+    CONF_DATA_RATE,
     DEVICE_CLASS_VOLTAGE,
     STATE_CLASS_MEASUREMENT,
     UNIT_VOLT,
@@ -42,6 +43,17 @@ RESOLUTION = {
     "12_BITS": ADS1115Resolution.ADS1015_12_BITS,
 }
 
+ADS1115DataRate = ads1115_ns.enum("ADS1115DataRate")
+DATA_RATE = {
+    "ADS1115_8_SPS": ADS1115DataRate.ADS1115_8_SPS,
+    "ADS1115_16_SPS": ADS1115DataRate.ADS1115_16_SPS,
+    "ADS1115_32_SPS": ADS1115DataRate.ADS1115_32_SPS,
+    "ADS1115_64_SPS": ADS1115DataRate.ADS1115_64_SPS,
+    "ADS1115_128_SPS": ADS1115DataRate.ADS1115_128_SPS,
+    "ADS1115_250_SPS": ADS1115DataRate.ADS1115_250_SPS,
+    "ADS1115_475_SPS": ADS1115DataRate.ADS1115_475_SPS,
+    "ADS1115_860_SPS": ADS1115DataRate.ADS1115_860_SPS,
+}
 
 def validate_gain(value):
     if isinstance(value, float):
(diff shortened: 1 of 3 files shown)


esphome/components/ads1115/sensor.py
```diff
@@ -42,6 +43,18 @@
...
+DATA_RATE = {
+    "ADS1115_8_SPS": ADS1115DataRate.ADS1115_8_SPS,
+    "ADS1115_16_SPS": ADS1115DataRate.ADS1115_16_SPS,
+    "ADS1115_32_SPS": ADS1115DataRate.ADS1115_32_SPS,
+    "ADS1115_64_SPS": ADS1115DataRate.ADS1115_64_SPS,
+    "ADS1115_128_SPS": ADS1115DataRate.ADS1115_128_SPS,
+    "ADS1115_250_SPS": ADS1115DataRate.ADS1115_250_SPS,
+    "ADS1115_475_SPS": ADS1115DataRate.ADS1115_475_SPS,
+    "ADS1115_860_SPS": ADS1115DataRate.ADS1115_860_SPS,
+}
```
jesserockz, line 56: The rate used in yaml should be more user friendly, eg `data_rate: 860sps` / `data_rate: 860 sps`
```suggestion
DATA_RATE = {
    "8_SPS": ADS1115DataRate.ADS1115_8_SPS,
    "16_SPS": ADS1115DataRate.ADS1115_16_SPS,
    "32_SPS": ADS1115DataRate.ADS1115_32_SPS,
    "64_SPS": ADS1115DataRate.ADS1115_64_SPS,
    "128_SPS": ADS1115DataRate.ADS1115_128_SPS,
    "250_SPS": ADS1115DataRate.ADS1115_250_SPS,
    "475_SPS": ADS1115DataRate.ADS1115_475_SPS,
    "860_SPS": ADS1115DataRate.ADS1115_860_SPS,
}
```

esphome/components/ads1115/sensor.py
```diff
@@ -73,6 +86,9 @@ def validate_gain(value):
             cv.Optional(CONF_RESOLUTION, default="16_BITS"): cv.enum(
                 RESOLUTION, upper=True, space="_"
             ),
+            cv.Optional(CONF_DATA_RATE, default="ADS1115_860_SPS"): cv.enum(
```
jesserockz, line 89:
```suggestion
            cv.Optional(CONF_DATA_RATE, default="860_SPS"): cv.enum(
```

jesserockz (changes requested):

esphome/components/ads1115/sensor.py
re-ynd: Thanks for the review. The reason I prefixed the enums with ADS1115 is because while the code is under esphome/components/**ads1115** it works on both ADS1115  and ADS1015. The constants I added are only applicable for ADS1115. I could take your suggestions and add a comment in the documentation if you prefer that.

jesserockz: Ok, so it looks like the only difference between the two (that we can use in config) is the resolution, which is 16 for 1115 and 12 for 1015.

If this is true, then we can make a validation on the config to only allow `data_rate` to be configured if resolution is set to 16 bits.

re-ynd: Actually the data rate can be configured for both. I just added ADS 1115 enums because that is the only device I have to test. I could add ADS 1015 enums too (like below), but I have no way to test.

```
#define RATE_ADS1015_128SPS (0x0000)  ///< 128 samples per second
#define RATE_ADS1015_250SPS (0x0020)  ///< 250 samples per second
#define RATE_ADS1015_490SPS (0x0040)  ///< 490 samples per second
#define RATE_ADS1015_920SPS (0x0060)  ///< 920 samples per second
#define RATE_ADS1015_1600SPS (0x0080) ///< 1600 samples per second
#define RATE_ADS1015_2400SPS (0x00A0) ///< 2400 samples per second
#define RATE_ADS1015_3300SPS (0x00C0) ///< 3300 samples per second
```

re-ynd: Friendly ping

PR #6087, opened 2024-01-12
update script/setup so it works fine on windows 

nielsnl68:

It looks like that Python 3.12 changed the location where the *venv* files are stored on windows systems,
instead if `venv\bin` it is now in `venv\Script`.
This PR checks if `activate` file exist in `venv\Script` or `venv\bin`  and call it from that location.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/script/setup
+++ b/script/setup
@@ -4,14 +4,13 @@
 set -e
 
 cd "$(dirname "$0")/.."
-
+location = "venv/bin/activate"
 if [ ! -n "$DEVCONTAINER" ] && [ ! -n "$VIRTUAL_ENV" ] && [ ! "$ESPHOME_NO_VENV" ]; then
   python3 -m venv venv
   if [ -f venv/Scripts/activate ]; then
-    source venv/Scripts/activate;
-  else
-    source venv/bin/activate;
+    location = "venv/Scripts/activate"
   fi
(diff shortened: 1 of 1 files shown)


script/setup
```diff
@@ -25,4 +29,5 @@ script/platformio_install_deps.py platformio.ini --libraries --tools --platforms
 
 echo
 echo
-echo "Virtual environment created; source venv/bin/activate to use it"
+echo "Virtual environment created. Depending on your Phyton version and OS run;"
+echo "'source venv/Scripts/activate' or 'source venv/bin/activate' to use it."
```
jesserockz: You should be able to determine the version and OS within the script and print the correct path. Or use the above check to set a variable and print the path based on that

jesserockz (changes requested):

script/setup
nielsnl68: done!

jesserockz (approved):
Thanks

PR #6112, opened 2024-01-17
Improve fan preset mode implementation

### LINKED ISSUE #1661: Add support for fan preset modes

mill1000:

Addresses some of the issues reported with the recently introduced fan preset modes (#5694).

* Add `preset` option to fan.turn_on Action (https://github.com/esphome/feature-requests/issues/1661#issuecomment-1879336947).

WIP
* Setting speed from a lambda clears the preset (https://github.com/esphome/feature-requests/issues/1661#issuecomment-1879250623).

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/1661

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#TBD

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
# TBD
```

## Checklist:


mill1000:
Creating this PR because I believe there needs to be some discussion around solving the 2nd bullet.

> Setting speed from a lambda clears the preset (https://github.com/esphome/feature-requests/issues/1661#issuecomment-1879250623).

The only means to set speed, direction and oscillation from a template appears to be by creating a `FanCall` with the desired settings. This has the side effect of clearing the fan's preset mode, unless the preset mode is explicitly added to the call. This feels pretty hacky to me.
e.g.
```yaml
sensor:
  - platform: adc
    name: Test Sensor
    # ...snip...
    on_value:
      then:
        lambda: |-
          if (id(test_fan).preset_mode == "Auto") {
            // Set fan speed based on sensor value
            auto call = id(test_fan).make_call();
            call.set_preset_mode(id(test_fan).preset_mode); // Copy preset mode into call so it won't get cleared
            call.set_speed(x/10);
            call.perform();
          }
```

A possible alternative would be to expose a public `update_state()` or `write_state()` method to apply the current fan state (speed, direction, etc) to it's outputs without using a FanCall.

e.g. Proposed
```yaml
sensor:
  - platform: adc
    name: Test Sensor
    # ...snip...
    on_value:
      then:
        lambda: |-
          if (id(test_fan).preset_mode == "Auto") {
            // Set fan speed based on sensor value
            auto fan = id(test_fan);
            fan.speed = x / 10;
            fan.update_state();
          }
```

Thoughts? Other ideas?

esphome/components/fan/automation.h
```diff
@@ -14,6 +14,7 @@ template<typename... Ts> class TurnOnAction : public Action<Ts...> {
   TEMPLATABLE_VALUE(bool, oscillating)
   TEMPLATABLE_VALUE(int, speed)
   TEMPLATABLE_VALUE(FanDirection, direction)
+  TEMPLATABLE_VALUE(std::string, preset_mode)
```
filmkorn, line 17: I'd like to turn off the fan when not needed while maintaining the preset_mode.

So I'm wondering if this argument could be added to TurnOffAction as well? The reason is that you can't turn off the fan using `turn_on` with speed=0. The fan will remain on with speed=1 (due to the config validation as far as I can tell).

filmkorn (commented):
Personally I tend to try and keep lambdas to a minimum and avoid calls.

In my use case the script would look something like this (not sure if this input helps):

```yaml
script:
  - id: set_fan_seed
    then:
      - if:
          # Auto mode
          condition:
            - lambda: 'return id(air_purifier).preset_mode == "auto";'
          then:
          - if:
              condition:
                lambda: 'return id(pm25).state < 1;'
              then:
                - fan.turn_off:
                    id: air_purifier
                    preset: !lambda "return id(air_purifier).preset_mode"
              else:
                 - fan.turn_on:
                     id: air_purifier
                     # Using a fan speed_count = 3 to get designated presets low, med, high in lovelace.
                     # Remapping the pm25 sensor from 0-1000 to those 3 speeds.
                     speed: !lambda "return remap(id(pm25).state, 0.0f, 1000.0f , 0.0f, 3.0f);"
                     # Keep the preset mode.
                     preset: !lambda "return id(air_purifier).preset_mode"
```

I'm wondering if having a mix of `preset` and `preset_mode` might cause confusion?

mill1000:
> I'm wondering if having a mix of preset and preset_mode might cause confusion?

You're right. It would. I thought I was mimicking the climate implementation, but climate never uses a "_mode" suffix for presets.

esphome/components/fan/automation.h
mill1000: I don't think specifying a `preset_mode` in the TurnOffAction would be necessary if preset_mode wasn't changed on every call()

mill1000:
I don't personally have an interest in finishing this work. Not sure if it's still applicable either.

PR #6113, opened 2024-01-17
Light component: add a way to disable restore_mode

### LINKED ISSUE #2547: Add restore_mode DISABLED  option in Light component

Add restore_mode DISABLED  option in Light component

I have an ESP32 with ESPHOME, it works well but it restarts every now and then. I have some Light components configured on it that send RF codes. My problem is that every time it restarts, it tries to recover the state, but since those lights can also be manually turned on, it would be preferable if it didn’t try to recover any state.
ALTERNATIVES I HAVE TRIED 
Use restore_from_flash to avoid restoring it but it is not for ESP32 so it does not work 

**Additional context**
The SWITCH component does have this functionality as described in the restore_mode section of https://esphome.io/components/switch/index.html

I could try to replicate the functionality of the Switch in the light component but my C++ skills are between scarce and none.

Thank you, and sorry for my english

t3chguy:

Adds a `restore_mode: DISABLED` enum value to the light component

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2547

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
remote_transmitter:
  pin: GPIO26
  carrier_duty_percent: 100%

output:
  - platform: template
    type: binary
    id: rf_magnifying_light
    write_action:
      - if:
          condition:
            light.is_on: _rf_magnifying_light
          then:
            remote_transmitter.transmit_rc_switch_raw:
              code: '001110000100011100000001'
              protocol: 1
              repeat:
                  times: 5
                  wait_time: 0s
          else:
            remote_transmitter.transmit_rc_switch_raw:
              code: '001110000100011100000011'
              protocol: 1
              repeat:
                  times: 5
                  wait_time: 0s

light:
  - platform: binary
    name: "Magnifying Light"
    output: rf_magnifying_light
    id: _rf_magnifying_light
    restore_mode: DISABLED
```

## Checklist:


jesserockz (dismissed):
LGTM

t3chguy:
@jesserockz do you think the alternative (overengineered) approach summarised [here](https://github.com/esphome/feature-requests/issues/2547#issuecomment-1897452759) would be preferable? Would be good to avoid duplication/breaking changes and getting it right off the bat if you think being able to restore the initial state without writing it out to external devices (e.g. RF) would be better.

cc @ssieb

will fast follow this with any consensus on approach, the approach here was sufficient for my own usage

t3chguy:
@jesserockz let me know if you're happy with this approach, works just as well as the original one for my use case. Will make the accompanying docs PR at that point

ssieb:
assumed state is very different from just not setting a value at boot.  What you're doing is only disabling restore, which seems better to be added as a restore mode like other components.

jesserockz (changes requested):
If it is set to disabled then you should also not save the state back to flash/rtc.

PR #6114, opened 2024-01-18
Enable networking and some other components on host platform

clydebarrow:

Allows building ESPHome executables with working networking on the `host:` platform (uses platformio native platform.)

Specifically useful for testing non-hardware related components without having to flash to ROM.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/host/__init__.py
```diff
@@ -34,6 +34,11 @@ def set_core_data(config):
 
 
 async def to_code(config):
-    cg.add_build_flag("-DUSE_HOST")
+    cg.add_build_flag("-DUSE_HOST=1")
```
jesserockz: This should stay with no value but be checked with `ifdef` as per all of our other platform checks
```suggestion
    cg.add_build_flag("-DUSE_HOST")
```

esphome/components/host/__init__.py
```diff
@@ -35,5 +35,8 @@ def set_core_data(config):
 
 async def to_code(config):
     cg.add_build_flag("-DUSE_HOST")
+    cg.add_build_flag("-std=c++17")
+    cg.add_build_flag("-lsodium")
+    cg.add_build_flag("-L/opt/homebrew/lib")
```
jesserockz: This is specific for macos right?
If so, you should check the host os and only add if it is macos

jesserockz (approved):

PR #6116, opened 2024-01-18
[Fingerprint_grow] Implements Sleep Mode feature

alexborro:

This feature enable the use of one GPIO pin to toggle the sensor power off when idling for a period of time.
It will be turned on again when there is a request for a communication to the sensor.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** Implements https://github.com/esphome/feature-requests/issues/1672

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3558

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
uart:
  rx_pin: GPIO16
  tx_pin: GPIO17
  baud_rate: 57600
    
fingerprint_grow:
  sensing_pin: GPIO4
  sensor_power_pin:
      number: GPIO18
      inverted: true
  idle_period_to_sleep: 5s
```

## Checklist:

### DIFF
--- a/esphome/components/fingerprint_grow/fingerprint_grow.cpp
+++ b/esphome/components/fingerprint_grow/fingerprint_grow.cpp
@@ -16,9 +16,14 @@ void FingerprintGrowComponent::update() {
   }
 
   if (this->has_sensing_pin_) {
+    // A finger touch results in a low level (digital_read() == false)
     if (this->sensing_pin_->digital_read()) {
       ESP_LOGV(TAG, "No touch sensing");
       this->waiting_removal_ = false;
+      if ((this->enrollment_image_ == 0) &&  // Not in enrolment process
+          (millis() - this->last_transfer_ms_ > this->idle_period_to_sleep_ms_)) {
+        this->sensor_sleep_();
+      }
       return;
     } else if (!this->waiting_removal_) {
       this->finger_scan_start_callback_.call();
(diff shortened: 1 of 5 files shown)


esphome/components/fingerprint_grow/fingerprint_grow.cpp
```diff
@@ -344,20 +371,20 @@ uint8_t FingerprintGrowComponent::send_command_() {
...
+  for (auto data : *p_data_buffer) {
     this->write(data);
     sum += data;
   }
 
   this->write((uint8_t) (sum >> 8));
   this->write((uint8_t) (sum & 0xFF));
 
-  this->data_.clear();
+  p_data_buffer->clear();
```
OnFreund, line 389: I'm not sure clearing out a client supplied buffer is the right approach...

alexborro: Well, I kind of agree with you, but the original code was already using the member `data_` to receive and return data. 

And the `transfer_()`  argument is not `const` so the caller cannot expect to pass a pointer and have no changes on it (otherwise the argument should have the `const` keyword). This whole function can be refactored to fix this original bug but it would require a major refactoring. 

I have a plan to change this whole communication to an asynchronous process, so it no longer keeps the component executing for more than one second without returning.

OnFreund: Yeah, but the original code operated under the assumption that there was only one message at a time, so reusing one buffer was more efficient. It's a bit different now.
In any case, if you plan on getting this communication overhaul done soon I wouldn't sweat it now.

alexborro: Yeap, that is the idea. I didn't want to refactor the whole component communication procedure because it is huge and should be done as a separate task - that is my plan. This is just a new feature some fellows are asking for - including me.

alexborro: Hi @OnFreund 
If you agree on this idea (refactoring this send command function as a next step), could you please approve this PR? Otherwise It is hard for me keeping rebasing my development branches on top of this one (including that ongoing refactoring).
Thanks.

OnFreund: This looks good to me, but it's @loongyh who really knows this code, not me...

loongyh (approved):
Hmm interesting, I have always wondered why the touch pin needs to be seperately powered with `3.3VT` and what's the use case with it. Although I'm currently unable to test it myself, the code LGTM.

 Thanks for your contribution!

esphome/components/fingerprint_grow/fingerprint_grow.cpp
loongyh: We need a maintainer who has write access to approve. @jesserockz can you proceed to review and merge?

alexborro:
Hi @jesserockz , could you please check if this PR is good to be merged ? It is lingering here for a few weeks. And it also fixes some reported issues (https://github.com/esphome/issues/issues/5502)

alexborro:
Hi @jesserockz , could you please check if this PR is good enough for the next release ? It is lingering here for weeks. It also fixes some issues so people don't need to keep using external component (issues mentioned in comments).
Cheers.
Alex.

jesserockz (approved):

PR #6128, opened 2024-01-21
Nextion TFT upload IDF memory optimization

edwardtfn:

This optimizes the memory in use for TFT upload when using `esp-idf` framework.

Basically, the engine establishes 3 connections to the the http/https server:
1. Fetch the file size (used to manage chunks and file size)
2. Transfer the 1st chunk (when it evaluates Nextion response to define either to continue from that point or to another point in the file)
3. Transfer the remaining data.

Until now, connection 1 was kept open during the whole process taking approx.. 40kb of heap in a esp32dev (NSPanel in my tests) and the same amount of memory was needed to the 2nd and 3rd connections (which never competes to each other). With this change, each connection is closed and released before opening the next one with a significant reduction on the required heap needed for this transfer.

This can still be improved to use a persistent connection, but I will look at this in the future, so it is not part of this change.

In addition to the better connection management, I've added quite a lot of log (mostly at VERBOSE or VERY_VERBOSE levels), which was used for troubleshooting here. I was unsure about removing this. As it can be useful for others, I decided to keep it, but I will be fine about removing it if this is now in line with ESPHome best practices.

## Types of changes

- [ ] Other

I couldn't find any issue registered about this, but it's fixing an unnecessary memory allocation which could easily drives to a bug on systems with low memory.

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

> [!NOTE]
> Tested (successfully) also with ESP v5:
> ```yaml
> esp32:
>   framework:
>     type: esp-idf
>     version: 5.0.2
>     platform_version: 6.3.2
> ```
N/A

## Checklist:

### DIFF
--- a/esphome/components/nextion/nextion_upload_idf.cpp
+++ b/esphome/components/nextion/nextion_upload_idf.cpp
@@ -222,7 +222,6 @@ bool Nextion::upload_tft() {
   ESP_LOGVV(TAG, "Connection closed");
   ESP_LOGVV(TAG, "Available heap: %" PRIu32, esp_get_free_heap_size());
 
-
   if (tft_file_size < 4096) {
     ESP_LOGE(TAG, "File size check failed. Size: %zu", tft_file_size);
     return this->upload_end(false);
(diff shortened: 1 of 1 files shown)


kbx81 (approved):
Tested locally, LGTM -- thanks again! 🍻

PR #6131, opened 2024-01-23
SN74HC595 support for inverted & restore modes without relay pulsing

gdgib:

The SN74HC595 component currently pulses relays when they're active low (e.g. `inverted: true`) due to the way the reset logic handles inversion. In particular, because inversion is a property of the pin class, and not known to the parent class, the setup logic can't properly use inversion values.  This means the restore modes are opposite what you want, with effects ranging from annoying to security issues if the pins control gates/latches/etc.

This change moves the storage of the pin inverted state to the parent class, so that the setup logic can use it.  Basically now instead of the pin class/objects storing their "inverted" value, the shift register does.  Thus `restore_mode` and `inverted` play together nicely at boot/programming/etc times.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** I don't believe a documentation change is in order for this.

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

The key to testing this is to look at `inverted` shift register outputs at boot time either using a logic analyzer, or just connecting them to an LED or relay.  If you connect all 8, and all 8 are inverted:

* Without this fix you'll see them change one by one (quickly, but visibly) as the set_pin/digital_write code is executed.  
* With this fix they'll all be set to the correct value without visible delay at boot time.

```yaml
sn74hc595:
  - id: 'sn74hc595_hub'
    latch_pin: D7
    oe_pin: D6
    sr_count: 2

switch:
  - platform: gpio
    name: "SN74HC595 Pin #0"
    pin:
      sn74hc595: sn74hc595_hub
      number: 0
      inverted: true
```

## Checklist:
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).  - note from @gdgib I'm open to adding tests, but there don't appear to be any for the component to start with, so I'd need some pointers


gdgib:
If anyone wants this functionality before this is merged, I created https://github.com/gdgib/esphome_sn74hc595.  Note the "i" on the end of the component name that you'll need to use in your config.

For anyone reading this, that's also how I've been testing & using these changes, of course.

gdgib:
I wouldn't mind at all if someone can help me with the build errors, here.  They're not the most self-explanatory, and since this works for me in production it'll be a while before deciphering them becomes a priority.  They seem to be mostly about formatting, I think?

gdgib:
This PR is still relevant and the bug it fixes is still a problem, and I still use my own component as a work around.  I spent a few minutes resolving conflicts, but honestly, I'm not even going to bother re-testing this unless I see some sign that the esphome project might be interested in this since it's been over a year.

PR #6133, opened 2024-01-23
Add workflow for changed components

clydebarrow:

Adds a twice-daily workflow that lists changed components between the last release version and the current dev
version, and publishes this as a json file to Github pages.

The intention is that this can then be used by the ESPHome CLI and Dashboard to alert users that there are upcoming
changes to components used in their YAML configs, and suggest testing against the dev branch.

The alerts will require further work, this is the first step.

A sample output can be seen here: https://clydebarrow.github.io/esphome/changed_components.json

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


.github/workflows/changed_components.yml
```diff
@@ -0,0 +1,69 @@
+# Simple workflow for deploying static content to GitHub Pages
+name: List upcoming component changes
+
+on:
```
jesserockz, line 5: Unfortunately this is the most glorious way to fix this specific lint issue (I don't want to disable the rule entirely)
```suggestion
# yamllint disable-line rule:truthy
on:
```

jesserockz:
I wonder if this better fits in its own repo? (within the org still)
Mostly so that we don't have to enable gh pages on the main code repo for this one file, but also in case we do decide to use pages at some point, it might be difficult to manage/migrate later.

clydebarrow:
> I wonder if this better fits in its own repo? (within the org still)

Ok, done. Repo is esphome/changed_components, output in GH Pages is:

`https://esphome.github.io/changed_components/changed_components.json`

Ideally this would have a dedicated subdomain of .esphome.io assigned e.g. `data.esphome.io` and be proxied through Cloudflare for CDN purposes.

PR #6134, opened 2024-01-23
Nextion - Remove soft reset at start

edwardtfn:

This introduces a user configurable attribute to Nextion display which allows skipping the soft reset on displays during the component startup.
I don't know really why this reset was introduced from the beginning, so I've left as a configurable attribute to avoid a breaking change, but on my tests it always work without this command and it makes the startup process a bit faster (my main goal).

This is not a breaking change, as the previous behavior is the default.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** solves esphome/feature-requests#2074

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3566

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
display:
  - id: disp1
    platform: nextion
    soft_reset_on_start: false
```

## Checklist:

### DIFF
--- a/esphome/components/nextion/display.py
+++ b/esphome/components/nextion/display.py
@@ -23,6 +23,7 @@ from .base_component import (
     CONF_START_UP_PAGE,
     CONF_AUTO_WAKE_ON_TOUCH,
     CONF_EXIT_REPARSE_ON_START,
+    CONF_SOFT_RESET_ON_START,
 )
 
 CODEOWNERS = ["@senexcrenshaw"]
@@ -72,6 +73,7 @@ CONFIG_SCHEMA = (
             cv.Optional(CONF_START_UP_PAGE): cv.positive_int,
             cv.Optional(CONF_AUTO_WAKE_ON_TOUCH, default=True): cv.boolean,
             cv.Optional(CONF_EXIT_REPARSE_ON_START, default=False): cv.boolean,
+            cv.Optional(CONF_SOFT_RESET_ON_START, default=True): cv.boolean,
         }
     )
     .extend(cv.polling_component_schema("5s"))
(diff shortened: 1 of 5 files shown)


kbx81 (approved):
LGTM -- thanks!

esphome/components/nextion/display.py
```diff
@@ -72,6 +73,7 @@
             cv.Optional(CONF_START_UP_PAGE): cv.positive_int,
             cv.Optional(CONF_AUTO_WAKE_ON_TOUCH, default=True): cv.boolean,
             cv.Optional(CONF_EXIT_REPARSE_ON_START, default=False): cv.boolean,
+            cv.Optional(CONF_SOFT_RESET_ON_START, default=True): cv.boolean,
```
kbx81, line 76: Hmm, I don't see where this is set in `to_code`

kbx81 (changes requested):
Just noticed something...

esphome/components/nextion/display.py
edwardtfn: I'm not sure I got your question.

kbx81: In the yaml, if I define `soft_reset_on_start: false`, how/where does this value get passed to/set in the underlying C code?

edwardtfn: Now I understand why I haven't noticed any difference with this. 😆 

I've just fixed with https://github.com/esphome/esphome/pull/6134/commits/3557439f68c550d998213abd76f158ce22e132d9, but now I will have to run tests again before resubmit this PR.
Thanks for your catch!

edwardtfn:
I'm dropping this PR as the display requires this `rest` to announce itself to the system. It does work without it, but then it requires a manual set to "Ready" status, but then the queue gets unsync.
Maybe I come back to this in the future.

PR #6137, opened 2024-01-23
host platform: improvements and bugfixes

clydebarrow:

Tweaks to make the `host:` platform more usable.

The mac address was not being initialised on the host: platform, so would change randomly. This allows it to be configured, not perhaps as good as getting an actual one, but since there are likely to be multiple hardware interfaces, that's not really practical.

The sntp component just gets the current system time on host:

The text_sensor component had some incorrect C++ declaring virtual methods;

Replace `sprintf` with `snprintf` in helpers (NB there are still plenty of places in the codebase where this needs doing.)

Add timestamps to logger output (this is done by the logging program for non-host platforms.)

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml
host:
  mac_address: "62:23:45:AF:B3:DD"

```

## Checklist:


esphome/components/host/__init__.py
```diff
@@ -28,13 +30,19 @@ def set_core_data(config):
 
 
 CONFIG_SCHEMA = cv.All(
-    cv.Schema({}),
+    cv.Schema(
+        {
+            cv.Optional(CONF_MAC_ADDR, default="98:35:69:175:179:221]"): cv.mac_address,
```
Unfocused: Even without the extra `]` at the end, that default isn't a valid MAC address

esphome/components/host/__init__.py
```diff
@@ -28,13 +30,19 @@ def set_core_data(config):
 
 
 CONFIG_SCHEMA = cv.All(
-    cv.Schema({}),
+    cv.Schema(
+        {
+            cv.Optional(CONF_MAC_ADDR, default="98:35:69:175:179:221"): cv.mac_address,
```
Unfocused: (This'll teach me for getting distracted by lunch while I'm in the middle of updating my comment...)

Even with the fixed typo, that default isn't a valid MAC address - should be 6 groups of 2 hexadecimal digits.

esphome/components/host/__init__.py
```diff
@@ -18,6 +18,8 @@
 CODEOWNERS = ["@esphome/core"]
 AUTO_LOAD = ["network"]
 
+CONF_MAC_ADDR = "mac_addr"
```
Unfocused: const.py already has a `CONF_MAC_ADDRESS`

esphome/components/host/__init__.py
clydebarrow: ~Actually, what was there worked, but~ I agree it looks wrong. Needs more work.

esphome/components/host/__init__.py
clydebarrow: It does. Strange that the CI didn't pick that up.

clydebarrow: Ah, MAC_ADDR vs MAC_ADDRESS

esphome/core/helpers.cpp
```diff
@@ -551,6 +557,10 @@ void HighFrequencyLoopRequester::stop() {
 bool HighFrequencyLoopRequester::is_high_frequency() { return num_requests > 0; }
 
 void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parameter)
+#ifdef USE_HOST
+  static const uint8_t esphome_host_mac_addr[6] = {ESPHOME_HOST_MAC_ADDR};
```
Unfocused: While I'm here & heckling from the cheap seats... this isn't all that important, but it really irked me to see.

Enclosing `ESPHOME_HOST_MAC_ADDR` in curly brackets like this feels wrong. The usual convention for defined macros is to try have them be something the compiler sees as valid by itself where possible/relevant - that way it's less likely to be misused (when in doubt, don't give yourself a footgun). So I'd expect to see this look like:

```suggestion
  static const uint8_t esphome_host_mac_addr[6] = ESPHOME_HOST_MAC_ADDR;
```

And the curly brackets added when the macro is originally defined (in the call to `add_build_flag` in the Python code)

esphome/components/host/__init__.py
```diff
@@ -28,13 +29,19 @@ def set_core_data(config):
...
+        }
+    ),
     set_core_data,
 )
 
 
 async def to_code(config):
     cg.add_build_flag("-DUSE_HOST")
+    mac_addr = "0x" + ",0x".join(f"{config[CONF_MAC_ADDRESS]}".split(":"))
+    cg.add_build_flag(f"-DESPHOME_HOST_MAC_ADDRESS={mac_addr}")
```
Unfocused: Isn't this what `cg.add_define` is for?

esphome/core/helpers.cpp
clydebarrow: It needs enclosing in braces to make it a valid initialiser. Passing braces via the command line with -D is problematic because of the required escaping and therefore itself prone to misuse.

> The usual convention for defined macros is to try have them be something the compiler sees as valid by itself where possible/relevant

I honestly don't know what that means. A macro is just a textual substitution, its validity is always going to be context dependent.

Would it make you happier if I just left spaces around the braces?

esphome/components/host/__init__.py
clydebarrow: You would think so wouldn't you? But it doesn't work, as it surrounds the bytes with quotes. 

> Since it's only going to be used here, not globally

Both methods result in a global definition.

esphome/core/helpers.cpp
Unfocused: I meant in the sense that usually you can put a macro anywhere, without needing to worry about adding anything else to make it a valid token or valid code. Usually. Like you wouldn't define a macro as `hello world"` and expect to need to add a `"` before it every time you use it.

And yea, escaping makes that less nice.... but see my comment about `cg.add_define`.

But as I said, it was just something I saw that irked me (a code smell), not something that's all that important.

clydebarrow: On further reflection, not only are there implementation difficulties in including the braces in the define, I think that would be wrong. The braces are syntax, not data. Syntax should be fixed and obvious, and not dependent on macro expansion as far as possible.

esphome/components/host/__init__.py
Unfocused: ... huh. I'll have to remember that quirk

esphome/core/helpers.cpp
Unfocused: Fair enough

esphome/core/helpers.cpp
```diff
@@ -551,6 +557,10 @@ void HighFrequencyLoopRequester::stop() {
 bool HighFrequencyLoopRequester::is_high_frequency() { return num_requests > 0; }
 
 void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parameter)
+#ifdef USE_HOST
```
Unfocused: Shouldn't this be part of the existing `#if defined(...)` and `#elif defined(...)` for the various other host platforms?

```cpp
#if defined(USE_HOST)
...
#elif defined(USE_ESP32)
...
```

esphome/core/helpers.cpp
```diff
@@ -415,7 +421,7 @@ std::string value_accuracy_to_string(float value, int8_t accuracy_decimals) {
 int8_t step_to_accuracy_decimals(float step) {
   // use printf %g to find number of digits based on temperature step
   char buf[32];
-  sprintf(buf, "%.5g", step);
+  snprintf(buf, sizeof buf, "%.5g", step);
```
Unfocused, line 424: This prompted me to file:
- https://github.com/esphome/issues/issues/5444

nielsnl68:
Clyde could you show me your YAML file where you use this and the VLC client. So i can us it myself as well?

clydebarrow:
> Clyde could you show me your YAML file where you use this and the VLC client. So i can us it myself as well?

Sure. All the necessary stuff is merged into the lvgl branch. Sample host.yaml:

```
substitutions:
  name: host-lvgl
  friendly_name: Host lvgl test

esphome:
  name: ${name}
  name_add_mac_suffix: false
  project:
    name: esphome.host
    version: "1.1"

api:
  reboot_timeout: 0s

external_components:
  - source: github://clydebarrow/esphome@lvgl
    refresh: 10min
    components: [ lvgl, font, host, vnc, sntp, time]

packages:
  lvgl: !include lvgl-package.yaml

time:
  - platform: sntp
    id: esptime
    timezone: Australia/Sydney

logger:
  level: VERBOSE
  logs:
    lvgl: DEBUG
    display: DEBUG
    sensor: DEBUG
    vnc: DEBUG

host:
  mac_addr: "62:23:45:AF:B3:DD"

touchscreen:
  - platform: vnc
    id: vnc_touchscreen
    display: vnc_display

display:
  - platform: vnc
    id: vnc_display
    update_interval: never
    auto_clear_enabled: false
    dimensions:
      width: 450
      height: 600

lvgl:
  touchscreens:
    - touchscreen_id: vnc_touchscreen
      long_press_time: 800ms
      long_press_repeat_time: 400ms
  color_depth: 16
  byte_order: big_endian
  displays:
    - display_id: vnc_display
```

esphome/components/host/__init__.py
jesserockz: I have done this in a commit added to this PR

jesserockz (approved):

PR #6146, opened 2024-01-26
Time based cover with tilt function

klaudiusz223:

Time based cover implementation with position and tilt control that are driven by one motor and do not have any position feedback. Such a solution can be found in the external venetian blinds.
Has an automatic recalibration function at extreme positions (fully open or fully closed). Takes into account the inertia time and the interlocking time when changing direction.
It can also be used to control covers without a tilt function. In such case, the rotation time must be set to 0. 

## Types of changes

- [ ] Other

https://github.com/esphome/feature-requests/issues/664
https://github.com/esphome/feature-requests/issues/2537
https://github.com/esphome/feature-requests/issues/2087
https://github.com/esphome/feature-requests/issues/1329
https://github.com/esphome/feature-requests/issues/778
https://github.com/esphome/feature-requests/issues/739

   esphome/esphome-docs#3574
Preview: https://deploy-preview-3574--esphome.netlify.app/components/cover/time_based_tilt/

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example configuration entry
cover:
  - platform: time_based_tilt
    name: "Time-Based Tilt Cover"
    open_duration: 58700ms
    close_duration: 57700ms
    inertia_open_time: 300ms
    tilt_open_duration: 930ms
    inertia_close_time: 250ms
    tilt_close_duration: 900ms
    recalibration_close_time: 6500ms
    recalibration_open_time: 4500ms
    actuator_activation_open_time: 200ms
    actuator_activation_close_time: 250ms
    close_sets_tilt: false
    open_sets_tilt: false

    open_action:
      - switch.turn_on: open_cover_switch

    close_action:
      - switch.turn_on: close_cover_switch

    stop_action:
      - switch.turn_off: open_cover_switch
      - switch.turn_off: close_cover_switch

```

## Checklist:


DaniEll-AT:
I would love to see this sophisticated component (inertia, interlock and recalibration :star_struck:) as a first class citizen in esphome.

MichalLytek:
I'm also eager to see this in action 😍 
My Shelly 2 PM are waiting for this 😄 

@klaudiusz223 Would it be hard to use the power meter sensors like in the `current_based` cover platform to distinguish between tilt and move/cover? 🤔

DaniEll-AT:
After testing some more, I found problems.

Currently I am using [bruxy70/Venetian-Blinds-Control](https://github.com/bruxy70/Venetian-Blinds-Control).
With exactly the same configuration/timing, this implementation does not work when tilt is changed in small steps. My educated guess is, there is a lack of precision/a rounding problem.

Example:
```
  - platform: time_based_tilt
    name: "Raffstore-1"
    id: cover1
    # Open
    open_action:
      - switch.turn_on: relay_1
    inertia_open_time: 200ms
    tilt_open_duration: 900ms
    open_duration: 43300ms
    # Close
    close_action:
      - switch.turn_on: relay_2
    inertia_close_time: 200ms
    tilt_close_duration: 900ms
    close_duration: 42500ms
    # Common
    interlock_wait_time: 1000ms
    recalibration_time: 2500ms
    # Stop
    stop_action:
      - switch.turn_off: relay_1
      - switch.turn_off: relay_2
```
When the cover is closed, opening tilt in small steps does not work when done repeatedly.

klaudiusz223:
Hi @DaniEll-AT 
Thanks for the feedback.
please describe in more details what does it mean doesn't work and how do You test small steps and what is the small step.
Doesn't move at all? Or move to the wrong position?
Do You control it localy eg by button or do You send command by Home Assistant. And what command? Do You take into accout inertia?
And Yes, there is a rounding to 1%, for both  tilt and postion.
But most important is what is Your loop time in the esphome.
e.g  (for 900 ms tilt time )   20 ms means that the smallest step will be greater than 2% 
 30 ms means more than 3%
but if You want to tilt by 5% it will be  probably reported as tilt by not less than 7% (3x 2.22% - and rounding).
I thought also about implementing tilting using HighFrequencyLoopRequester

klaudiusz223:
Hi @MichalLytek 

> @klaudiusz223 Would it be hard to use the power meter sensors like in the `current_based` cover platform to distinguish between tilt and move/cover? 🤔

I do not have current sensors in my device so I couldn't  implement it. 
What is exactly Your case scenario. Why do You need to distinguish tilting from moving? is the power consuption different for moving and tilting?
If I understand correctly, current based covers also calculate position based on time.
You could define some other functions e.g. obstacle detection as an external automation that sends a stop signal

MichalLytek:
> What is exactly Your case scenario. Why do You need to distinguish tilting from moving? is the power consuption different for moving and tilting?

Yes, there's even a different noise of the engine while tilting and moving. I'm afraid of the tricky time measuring inertia and tilting time, would be nice to use current reading to have auto calibration done this way.

But first let me try your code on my Shelly's 😉

klaudiusz223:
@DaniEll-AT 

> When the cover is closed, opening tilt in small steps does not work when done repeatedly.

I think, I know what You mean.
if the position of the cover is fully closed You have to send tilt command. 
In my implementation You can send both target postion and  target tilt at once.
I think I will improve a little bit logic of steering but i need to test it first

DaniEll-AT:
I am controlling the cover with Home Assistant. To me it looks like tilting does not consider inertia-time. I would have expected it to. What do you think?

The datasheet of my drive (Somfy J4 WT) states `Continuous orders of at least 200 ms must be sent to the drive to ensure proper execution.`
When tilting small steps (e.g. 2%) the relais is turned on for way to short, so nothing moves.
(This is what is called `actuator_activation_duration` at [bruxy70/Venetian-Blinds-Control ESPHome Custom Component](https://github.com/bruxy70/Venetian-Blinds-Control/tree/master/components/venetian_blinds))

klaudiusz223:
> Continuous orders of at least 200 ms must be sent to the drive to ensure proper execution.

Now it's clear.
The inertia I have implemented is something else.
I use Geiger Gj56 motors for blinds
when changing direction (and only when changing direction) there is some inertia before the lamellas rotate.
I think it has to do with the design of the motor or maybe the blinds.
So inertia is like the mechanism's transition from closing to opening, and the path to transition is always the same. It can be completed in one sequence or in several.
If 50% of the transition is traveled in one direction, only 50% of the way in the other direction have to be traveled when changing direction. And this travel distance is not "reset" after stop

I will consider whether it would be easy to implement what You need because I wanted to simplify my implementation a bit

Bascht74:
> I will consider whether it would be easy to implement what You need because I wanted to simplify my implementation a bit

Hi, @klaudiusz223,

anything new about simplifying or adding that feature? 
I would love to see that features on my Shelly devices so I am hoping you can finish the work soon so that it can be merged (milestone was changed)

maromme:
@klaudiusz223 thank's for your great work. it works for me flawlessly. Didn't find any problemes. 
Maybe one improvement: It would be nice if there was a option to choose weather the cover_recalibration_time is added at the top or bottom endstop. I've got a blind that should stop 1s bevore the bottom endstop and still I'd like the recalibration time at the top endstop.

klaudiusz223:
I've made some improvements to the code.
I added actuator activation time – separately for closing and opening. @DaniEll-AT @Bascht74 
I split the calibration process for the top and bottom positions. @maromme Is that what you meant?

Bascht74:
Not quite. 

The calibration I want (To set manually) is that motors get power but doesn’t start right away or slowing down at the end positions. 
So the movement time of the cover doesn’t reflect the exact position. So what is needed is a linear correction of that. You should be able to say that e.g. 95% of the opening time is fully open 95–>100), and after 50% of the opening time it is allready at a 65% position (50->65) and at 10% of the opening time is is at a 5% position (10->5) and after 1% of the opening time it really starts opening (1->0). It would be the same that tasmota implemented here:

https://youtu.be/Z-grrvnu2bU?si=73ERqEmoKLBMrtQ5

https://tasmota.github.io/docs/Blinds-and-Shutters
(Look after Increasing Calibration Granularity)

That would help a lot of people with strange shutters (like the German ones behind the windows) and in other situations wheee the running time doesn’t mechanically reflect the position of the cover.

Sebastian

sleepymaxx (approved):
Thank´s for your Work really apreciate it!

CODEOWNERS
```diff
@@ -423,6 +423,7 @@ esphome/components/template/fan/* @ssieb
 esphome/components/text/* @mauritskorse
 esphome/components/thermostat/* @kbx81
 esphome/components/time/* @OttoWinter
+esphome/components/time_based_tilt/* @klaudiusz223
```
sleepymaxx, line 507: great work

sleepymaxx (approved):

sleepymaxx:
Hi Klaudiusz, again me ... sorry for your inconveniece... got a little warning about feature updates of the source: 

`INFO ESPHome 2025.5.2
INFO Reading configuration /config/esphome/rollaeden-jalousien-esp.yaml...
WARNING Using `cover.COVER_SCHEMA` is deprecated and will be removed in ESPHome 2025.11.0. Please use `cover.cover_schema(...)` instead. If you are seeing this, report an issue to the external_component author and ask them to update it. https://developers.esphome.io/blog/2025/05/14/_schema-deprecations/. Component using this schema: time_based_tilt
WARNING Using `cover.COVER_SCHEMA` is deprecated and will be removed in ESPHome 2025.11.0. Please use `cover.cover_schema(...)` instead. If you are seeing this, report an issue to the external_component author and ask them to update it. https://developers.esphome.io/blog/2025/05/14/_schema-deprecations/. Component using this schema: time_based_tilt
INFO Generating C++ source...`

best greetings and wishes furthermore

XilviuX:
@klaudiusz223 not sure you're still maintaing your code but, as reported above, it gives a warning now:

```
WARNING Using `cover.COVER_SCHEMA` is deprecated and will be removed in ESPHome 2025.11.0. Please use `cover.cover_schema(...)` instead. If you are seeing this, report an issue to the external_component author and ask them to update it. 
```

Can you please update this?

klaudiusz223:
Implemented new options:
  * close_sets_tilt — if set to true: automatically set tilt to 0.0 when position=0.0 and no tilt is provided
  * open_sets_tilt — if set to true: automatically set tilt to 1.0 when position=1.0 and no tilt is provided
  The documentation has been completely rewritten. Preview: https://deploy-preview-3574--esphome.netlify.app/components/cover/time_based_tilt/

timotheeandres:
Any update on this? It'd be really cool :)

tests/components/time_based_tilt/test.esp32-ard.yaml
```diff
@@ -0,0 +1 @@
+<<: !include common.yaml
```
bdraco, line 1: Please remove duplicate tests for ard as there is no ard specific code in this PR

tests/components/time_based_tilt/test.esp32-c3-ard.yaml
```diff
@@ -0,0 +1 @@
+<<: !include common.yaml
```
bdraco, line 1: Please remove duplicate tests for c3 as there is no c3 specific code in this PR

esphome/components/time_based_tilt/time_based_tilt_cover.h
```diff
@@ -0,0 +1,93 @@
...
+
+ protected:
+  void control(const cover::CoverCall &call) override;
+  bool is_at_target_position_() const;
+  bool is_at_target_tilt_() const;
+  bool is_at_extreme_position_() const;
+
+  Trigger<> *open_trigger_{new Trigger<>()};
+  Trigger<> *close_trigger_{new Trigger<>()};
+  Trigger<> *stop_trigger_{new Trigger<>()};
```
bdraco, line 56: Please avoid heap allocations for Triggers

esphome/components/time_based_tilt/time_based_tilt_cover.h
```diff
@@ -0,0 +1,93 @@
...
+  float target_tilt_{TARGET_NONE};
+  float inertia_{0.0f};
+  bool has_built_in_endstop_{false};
+  bool assumed_state_{false};
+  cover::CoverOperation last_operation_{cover::COVER_OPERATION_OPENING};
+  State fsm_state_{STATE_IDLE};
+  cover::CoverOperation interlocked_direction_{cover::COVER_OPERATION_IDLE};
+  uint32_t interlocked_time_{0};
+  bool close_sets_tilt_{false};
+  bool open_sets_tilt_{false};
```
bdraco, line 90: Please align members from largest to smallest

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
```diff
@@ -0,0 +1,356 @@
...
+bool TimeBasedTiltCover::is_at_extreme_position_() const {
+  return (this->position == COVER_CLOSED && (tilt_close_duration_ == 0 || this->tilt == COVER_CLOSED)) ||
+         (this->position == COVER_OPEN && (tilt_open_duration_ == 0 || this->tilt == COVER_OPEN));
+}
+
+void TimeBasedTiltCover::loop() {
+  if (this->fsm_state_ == STATE_IDLE && this->target_position_ == TARGET_NONE && this->target_tilt_ == TARGET_NONE)
+    return;
+
+  const uint32_t now = millis();
```
bdraco, line 77:
```suggestion
  const uint32_t now = App.get_loop_component_start_time();
```

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+  // recalibrating in extreme postions
+  if (this->fsm_state_ == STATE_CALIBRATING) {
```
bdraco, line 80: Looks like this could be written as a switch on `this->fsm_state_`

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
```diff
@@ -0,0 +1,356 @@
...
+  ESP_LOGCONFIG(TAG, "  Inertia close time: %.3fs", this->inertia_close_time_ / 1e3f);
+  ESP_LOGCONFIG(TAG, "  Inertia open time: %.3fs", this->inertia_open_time_ / 1e3f);
+  ESP_LOGCONFIG(TAG, "  Recalibration close time: %.3fs", this->recalibration_close_time_ / 1e3f);
+  ESP_LOGCONFIG(TAG, "  Recalibration open time: %.3fs", this->recalibration_open_time_ / 1e3f);
+  ESP_LOGCONFIG(TAG, "  Actuator activation close time: %.3fs", this->actuator_activation_close_time_ / 1e3f);
+  ESP_LOGCONFIG(TAG, "  Actuator activation open time: %.3fs", this->actuator_activation_open_time_ / 1e3f);
+  ESP_LOGCONFIG(TAG, "  Close sets tilt: %s", this->close_sets_tilt_ ? "YES" : "NO");
+  ESP_LOGCONFIG(TAG, "  Open sets tilt: %s", this->open_sets_tilt_ ? "YES" : "NO");
+  ESP_LOGCONFIG(TAG, "  Current position: %.4f", this->position);
+  ESP_LOGCONFIG(TAG, "  Current tilt: %.4f", this->tilt);
```
bdraco, line 29: Please combine ESP_LOGCONFIG statements

https://developers.esphome.io/architecture/logging/

esphome/components/time_based_tilt/time_based_tilt_cover.h
```diff
@@ -0,0 +1,93 @@
...
+
+  const static float TARGET_NONE;
+  enum State : uint8_t { STATE_IDLE, STATE_MOVING, STATE_STOPPING, STATE_CALIBRATING };
+
+  uint32_t last_recompute_time_{0};
+  uint32_t last_publish_time_{0};
+  float target_position_{TARGET_NONE};
+  float target_tilt_{TARGET_NONE};
+  float inertia_{0.0f};
+  bool has_built_in_endstop_{false};
```
bdraco, line 83: never used

esphome/components/time_based_tilt/time_based_tilt_cover.h
```diff
@@ -0,0 +1,93 @@
...
+  void set_inertia_close_time(uint32_t inertia_time) { this->inertia_close_time_ = inertia_time; }
+  void set_actuator_activation_open_time(uint32_t activation_time) {
+    this->actuator_activation_open_time_ = activation_time;
+  }
+  void set_actuator_activation_close_time(uint32_t activation_time) {
+    this->actuator_activation_close_time_ = activation_time;
+  }
+  void set_close_sets_tilt(bool value) { this->close_sets_tilt_ = value; }
+  void set_open_sets_tilt(bool value) { this->open_sets_tilt_ = value; }
+  cover::CoverOperation compute_direction(float target, float current) {
```
bdraco, line 38: should be const

esphome/components/time_based_tilt/time_based_tilt_cover.h
```diff
@@ -0,0 +1,93 @@
...
+    if (target == current) {
+      return cover::COVER_OPERATION_IDLE;
+    }
+    return target < current ? cover::COVER_OPERATION_CLOSING : cover::COVER_OPERATION_OPENING;
+  };
+  float round_position(float pos) { return round(100 * pos) / 100; };
```
bdraco, line 44: should be const

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
```diff
@@ -0,0 +1,356 @@
...
+#include "esphome/core/log.h"
+#include "esphome/core/hal.h"
+
+namespace esphome::time_based_tilt {
+
+static const char *const TAG = "time_based_tilt.cover";
+
+const float TimeBasedTiltCover::TARGET_NONE = -1;
+
+using namespace esphome::cover;
```
bdraco, line 11:
```suggestion
```

Please qualify namespaces instead

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
```diff
@@ -0,0 +1,356 @@
...
+  if (this->tilt_close_duration_ == 0 || this->tilt_open_duration_ == 0) {
+    this->tilt_close_duration_ = 0;
+    this->tilt_open_duration_ = 0;
+  }
+  auto restore = this->restore_state_();
+  if (restore.has_value()) {
+    restore->apply(this);
+  } else {
+    this->position = 0.5f;
+    this->tilt = 0.5f;
```
bdraco, line 41: please use a constexpr for magic numbers
INERTIA_CENTER = 0.5f

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
```diff
@@ -0,0 +1,356 @@
...
+      this->position = this->round_position(this->position);
+
+      // If the cover is in extreme positions, perform recalibration.
+      if (this->current_recalibration_time_ > 0 && this->is_at_extreme_position_()) {
+        this->fsm_state_ = STATE_CALIBRATING;
+        this->publish_state(false);
+        ESP_LOGD(TAG, "Transition to the calibration state");
+      } else {
+        this->fsm_state_ = STATE_STOPPING;
+        ESP_LOGD(TAG, "Transition to the stopping state");
```
bdraco, line 248: This pattern appears many times, please make a helper

```
  void TimeBasedTiltCover::transition_to_stopping_() {                                                                                                                                                                                                  
    this->fsm_state_ = STATE_STOPPING;                                                                                                                                                                                                                  
    ESP_LOGD(TAG, "Transition to the stopping state");                                                                                                                                                                                                  
  }            
```

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
bdraco: For readability, I'd break each branch into its own function

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    if (now - this->last_publish_time_ > 1000) {
+      this->publish_state(false);
+      this->last_publish_time_ = now;
+    }
+  }
+}
+
+float TimeBasedTiltCover::get_setup_priority() const { return setup_priority::DATA; }
```
bdraco, line 259:
```suggestion
```

DATA is default

maromme:
I'm not sure where to ask you @klaudiusz223:

I'm using the component since a while, but with updating today I run into troubles using that code in a lamda:

`id(mycover).close()`

That is not compling anymore. Is there a "new" way of calling th clos /stop etc function from within a lambda code?

Thanks a lot, Markus

Btw: great work! Really hope it's integrated in esphome soon


**EDIT: here I got my answer. nothing todo with the pull request...
https://github.com/esphome/esphome-docs/issues/5153**

klaudiusz223:
> There hasn't been any activity on this pull request recently. 
Please keep it open

esphome/components/time_based_tilt/time_based_tilt_cover.h
esphbot: Confirmed. Sibling component shows the pattern: `time_based/cover/time_based_cover.h:35-39` holds `Trigger<> open_trigger_;` as value members and returns `&this->open_trigger_`. Note the getters must drop `const` — returning `Trigger<>*` from a const method won't compile against a non-mutable member.

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
esphbot: Worth noting: `time_based/cover/time_based_cover.cpp:10` uses the identical `using namespace esphome::cover;` at file scope. Not disputing the request — just flagging the sibling component would need the same treatment for consistency.

esphome/components/time_based_tilt/time_based_tilt_cover.h
esphbot: Confirmed dead. `has_built_in_endstop_` has no setter, no reader, and no schema key. Delete the field.

esphome/components/time_based_tilt/time_based_tilt_cover.cpp
esphbot: Confirmed. `esphome/core/component.cpp:82` — `Component::get_setup_priority()` already returns `setup_priority::DATA`. Drop both the override and the header declaration.

tests/components/time_based_tilt/test.esp32-ard.yaml
esphbot: Agreed. No platform-specific code here. `tests/components/time_based` ships exactly three: `esp32-idf`, `esp8266-ard`, `rp2040-ard`. Match that set.

esphbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - `TARGET_NONE` sentinel leaks into numeric comparisons — motor can run forever
> - Toggle sets tilt targets without checking that tilt is supported
> - Schema accepts incoherent tilt configurations silently

PR #6147, opened 2024-01-26
ILI9XXX: Restore offset usage in set_addr_window

clydebarrow:

this fixes a bug introduced in the recent changes for the Pico Res Touch 3.5 support. The x and y offsets to the screen
position were lost. Affects the T-Embed.
## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/ili9xxx/ili9xxx_display.cpp
+++ b/esphome/components/ili9xxx/ili9xxx_display.cpp
@@ -366,6 +366,10 @@ void ILI9XXXDisplay::init_lcd_() {
 
 // Tell the display controller where we want to draw pixels.
 void ILI9XXXDisplay::set_addr_window_(uint16_t x1, uint16_t y1, uint16_t x2, uint16_t y2) {
+  x1 += this->offset_x_;
+  x2 += this->offset_x_;
+  y1 += this->offset_y_;
+  y2 += this->offset_y_;
   this->command(ILI9XXX_CASET);
   this->data(x1 >> 8);
   this->data(x1 & 0xFF);


kbx81 (approved):

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -366,6 +366,10 @@ void ILI9XXXDisplay::init_lcd_() {
 
 // Tell the display controller where we want to draw pixels.
 void ILI9XXXDisplay::set_addr_window_(uint16_t x1, uint16_t y1, uint16_t x2, uint16_t y2) {
+  x1 += this->offset_x_;
+  x2 += this->offset_x_;
+  y1 += this->offset_y_;
+  y2 += this->offset_y_;
```
nielsnl68, line 372: Should that not be: 
```suggestion
  x1 += this->offset_x_;
  x2 -= this->offset_x_;
  y1 += this->offset_y_;
  y2 -= this->offset_y_;
```
Now you only move the display window and go over the display width and height.

clydebarrow: No, what I have is correct. It's a fixed offset, it has to be applied to all positions.

nielsnl68: As it is i disapprove this merge, 

There should have be some checks that the address window does not go before 0 and behind width and height.

clydebarrow:
Those checks are already done before set_addr_window is called. This is restoring functionality that was there before my recent PR where I inadvertently removed it. It is required for use on boards that do not use the full display area, such as the Lilygo T-Embed.

nielsnl68:
> Those checks are already done before set_addr_window is called. This is restoring functionality that was there before my recent PR where I inadvertently removed it. It is required for use on boards that do not use the full display area, such as the Lilygo T-Embed.

Okay, i still think it should not be here in the first place, considering it was there before your last PR and i did not saw any issues popping up, i cant say much any more. 
Let us talk to see  of we can find a better way to use this offset, in a future version (the display_buffer).

PR #6148, opened 2024-01-26
Fixes sml parser to process extended length lists with a number of items that is dividable by 16

irgendwienet:

My smartmeter type ZPA crashes the esp when receiving each sml message. 
By chance this meter sends a list of 16 items and the sml parser had a bug when the number of items is dividable by 16. 

## Types of changes

- [ ] Other

Maybe some of the problems of https://github.com/esphome/issues/issues/4642 are fixed.

no docs needed

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

See default yaml from docs: https://esphome.io/components/sml.html?highlight=sml

The original error is only observed with certain smart meters like ZPA. If needed I could provide a binary dump of the message that crashes esphime.

## Checklist:

Currently there are no tests for the sml parser

## Details of fix

The code change is very small. If the extended length field is used and the low part of the 2nd byte is 0x0 then the end-of-message-if was triggered. That was wrong. A list with e.g. 16 items is valid AND has a 2nd byte with 0x00. This is followed by the Data and is not the end of message.

Furthermore. The calculation `parse_length = length - 1;` was wrong since. 0x10 is 16 items not 15. for 15 0x0F is used and no extended-length is needed.

### DIFF
--- a/esphome/components/sml/sml_parser.cpp
+++ b/esphome/components/sml/sml_parser.cpp
@@ -27,7 +27,7 @@ bool SmlFile::setup_node(SmlNode *node) {
   uint8_t parse_length = length;
   if (has_extended_length) {
     length = (length << 4) + (this->buffer_[this->pos_ + 1] & 0x0f);
-    parse_length = length - 1;
+    parse_length = length;
     this->pos_ += 1;
   }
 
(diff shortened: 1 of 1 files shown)


irgendwienet:
This is a binary file I recorded from my smart meter that crashes the old version and is fixed with my PR. 
One could probably feed this into the serial port instead of the smart meter ir.
[z broken.zip](https://github.com/esphome/esphome/files/14068313/z.broken.zip)

passionsfrucht:
Hi all,

I was stumbling into similar issues with the parser and have been testing this fix for some time now, always rebased/cherry-picked against the newest release when I felt like it.
The stability has much improved, the patches applied without any problem.

I track the current state of rebase under https://github.com/passionsfrucht/esphome/tree/current-with-sml-fix.

Hope this helps someone or get this merged one day.

Many thanks for the patch!

jesserockz (approved):

PR #6156, opened 2024-01-30
Improve error message when client uses encryption when not enabled

bdraco:

Improve error message when client uses encryption when not enabled
Also improves the error message when the client uses plaintext when noise is required

`Bad indicator byte` is cryptic to most users and it generates questions that can be avoided with a better error message

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/api/api_frame_helper.cpp
+++ b/esphome/components/api/api_frame_helper.cpp
@@ -211,7 +211,11 @@ APIError APINoiseFrameHelper::try_read_frame_(ParsedFrame *frame) {
   uint8_t indicator = rx_header_buf_[0];
   if (indicator != 0x01) {
     state_ = State::FAILED;
-    HELPER_LOG("Bad indicator byte %u", indicator);
+    if (indicator == 0x00) {
+      HELPER_LOG("Client requested plaintext, but only noise encryption is required");
+    } else {
+      HELPER_LOG("Bad indicator byte %u", indicator);
+    }
     return APIError::BAD_INDICATOR;
   }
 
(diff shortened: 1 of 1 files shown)


bdraco:
sadly out of time to test it but I didn't want to forget to open it as I just spent way too long helping someone with this issue

esphome/components/api/api_frame_helper.cpp
```diff
@@ -211,7 +211,11 @@ APIError APINoiseFrameHelper::try_read_frame_(ParsedFrame *frame) {
   uint8_t indicator = rx_header_buf_[0];
   if (indicator != 0x01) {
     state_ = State::FAILED;
-    HELPER_LOG("Bad indicator byte %u", indicator);
+    if (indicator == 0x00) {
+      HELPER_LOG("Client requested plaintext, but only noise encryption is required");
```
Unfocused, line 215: The wording of "but only noise encryption is required" hurts my head.

Also, ESPHome's client/server structure is counter-intuitive to a lot of people. When using Home Assistant most people think of Home Assistant as the server, and the ESPHome device as the client. But with ESPHome it's the reverse - with the tiny microcontroller device being the server. I've seen this led to all sorts of confusion.

Maybe wording it more like "Remote client requested plaintext, but ESPHome device is set to require noise encryption" would help? Including "is set to" also hints that this can be fixed by altering the device config.

Ditto with the other message.

bdraco:
Not worth the trade-off of improving this as it will use more bytes on flash. The error message on the client side is enough. The user was just ignoring it and I had spent way too much time digging in their logs to help them only to find they hadn't read the error in HA which is not cryptic at all.

Unfocused:
Error messages are a user interface - they can be the cause of hours of confusion and frustration to the point where they can ruin someone's day, or they can be the source of an immediate solution. That seems worth a couple of bytes to me.

And ESPHome has numerous clients - there's no guarantee they'll all have useful error messages.

bdraco:
> And ESPHome has numerous clients - there's no guarantee they'll all have useful error messages.

The client side errors messages are in aioesphomeapi. If someone is using something else, its up to that library to provide good error messages

PR #6157, opened 2024-01-31
Add "transformer_active" flag for use in effects.

TikiBill:

This adds a simple boolean flag to indicate if a transformer is active such that effects may easily know when to return immediately in the apply() method. As such, it simplifies effect development.

This is in the Kauf bulb source and I used it with great success when developing a new effect. I went to bring the code over to ESPHome to contribute and realized there is no simple (that I could find) alternative.

## Types of changes

- New feature (non-breaking change which adds functionality)

This is internal only and does not affect any configuration files nor add any configurations.


esphome/components/light/light_state.h
```diff
@@ -144,6 +144,19 @@ class LightState : public EntityBase, public Component {
...
+   *
+   * if (this->state_->transformer_active) {
+   *   // Something is already running.
+   *   return;
+   * }
+   *
+   * This property is read-only for users. Any changes to it will be ignored.
+   */
+  bool transformer_active = false;
+
```
jesserockz: Can we change this to a protected member and add `bool is_transformer_active()`?

jesserockz (changes requested):

esphome/components/light/light_state.h
TikiBill: Thank you for the feedback. Your suggestion has been implemented.

jesserockz (approved):

jesserockz:
Thanks @TikiBill

PR #6161, opened 2024-02-01
LTR390 - Multiple bugfixes

### LINKED ISSUE #4380: ltr390 light sensor reports very low UVI values and resolution settings not working
### The problem

Issue 1: When using the ltr390 sensor component, it reports very low UVI values. For example, when under direct midday sun, when UVI of 7 or greater is expected, it would report UVI of less than 1.

Issue 2: When changing the resolution to something other than default, the raw outputs, als or uvs, do not scale with the resolution. Instead the calculated light lux (lx) output would change by factors of 2 even though it should stay more or less the same under the same lighting conditions.

Looking into the source file, I think I have found and fixed the issues. Attached is the modified .cpp file that seems to give reasonable values. I have no light measurement tool to verify.

[ltr390.cpp.txt](https://github.com/esphome/issues/files/11130942/ltr390.cpp.txt)

The cause of issue 1 seems to be the interpretation of the ambiguous datasheet. The UVI Conversion Formula in the datasheet has no terms to compensate for gain and resolution settings. While on Figure 4.6, it implies that the Sensitivity of 2300 counts/UVI is tied to Gain 18x and Resolution 20-bit.

In my experiments, even under direct midday sun, the UV ADC output values are still a number of bits short of the max afforded by the ADC register at Gain 18x. So it is best to use Gain 18x for UV measurement all the time. In the .cpp file attached, it uses the user config gain setting for the ambient light measurement and uses 18x for UV measurement. The calculation of UVI is changed to include a factor ( 4 / INT ) to account for the different resolution settings.

The cause of issue 2 is that the user config resolution setting is never applied to the ltr390 chip. There are a few lines in the original source in setup() with a comment of "Set resolution". But the few lines end up not really changing anything. In the attached .cpp, it has been modified to actually write the user config resolution setting to the ltr390 chip.

2022.12.8 is what I am using

pip

n/a, debug using esphome log

ESP8266

### Board

d1 mini

ltr390

### Example YAML snippet

```yaml
i2c:
  scl: 5
  sda: 4

sensor:
  - platform: ltr390
    address: 0x53
    uv_index:
      name: ${esp_name} uv index
    uv:
      name: ${esp_name} uvs
    light:
      name: ${esp_name} alight
    ambient_light:
      name: ${esp_name} als
    resolution: 20
    update_interval: 10s
```

_No response_

### Additional information

_No response_

sjtrny:

- fix https://github.com/esphome/issues/issues/4380
  - Set resolution parameter on the device, which was not set previously.
  - Added correction to UV index calculation to account for different gain and integration time combinations. 
  - Set default gain and resolution/integration time to the value which is used as a reference point in the data sheet and provides the most accurate values. All other UVI calculations are computed from this reference point.

- https://github.com/home-assistant/core/issues/88278
  - Fix incorrect device class

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** See above

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/3585

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
sensor:
  - platform: ltr390
    update_interval: 5s
    gain: "X3"
    resolution: 18
    uv_index:
      name: "UV Index"
    uv:
      name: "UV Counts"
```

## Checklist:

### DIFF
--- a/esphome/components/ltr390/ltr390.h
+++ b/esphome/components/ltr390/ltr390.h
@@ -17,12 +17,6 @@ enum LTR390CTRL {
 };
 
 // enums from https://github.com/adafruit/Adafruit_LTR390/
-static const uint8_t LTR390_MAIN_CTRL = 0x00;
-static const uint8_t LTR390_MEAS_RATE = 0x04;
-static const uint8_t LTR390_GAIN = 0x05;
-static const uint8_t LTR390_PART_ID = 0x06;
-static const uint8_t LTR390_MAIN_STATUS = 0x07;
-
 // Sensing modes
 enum LTR390MODE {
   LTR390_MODE_ALS,
--- a/esphome/components/ltr390/ltr390.h
+++ b/esphome/components/ltr390/ltr390.h
@@ -17,13 +17,11 @@ enum LTR390CTRL {
 };
 
 // enums from https://github.com/adafruit/Adafruit_LTR390/
-
 static const uint8_t LTR390_MAIN_CTRL = 0x00;
 static const uint8_t LTR390_MEAS_RATE = 0x04;
 static const uint8_t LTR390_GAIN = 0x05;
 static const uint8_t LTR390_PART_ID = 0x06;
 static const uint8_t LTR390_MAIN_STATUS = 0x07;
-static const float LTR390_SENSITIVITY = 2300.0;
 
 // Sensing modes
 enum LTR390MODE {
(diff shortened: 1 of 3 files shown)


esphome/components/ltr390/ltr390.h
```diff
@@ -50,6 +48,17 @@ enum LTR390RESOLUTION {
...
+static const float RESOLUTIONVALUE[6] = {4.0, 2.0, 1.0, 0.5, 0.25, 0.125};
+
+// Request fastest measurement rate - will be slowed by device if conversion rate is slower.
+static const float RESOLUTION_SETTING[6] = {0x00, 0x10, 0x20, 0x30, 0x40, 0x50};
+static const uint32_t MODEADDRESSES[2] = {0x0D, 0x10};
+
+static const float SENSITIVITY_MAX = 2300;
+static const float INTG_MAX = RESOLUTIONVALUE[0] * 100;
+static const int GAIN_MAX = GAINVALUES[4];
+
```
jesserockz: We dont put consts in the header unless they are required there, but these should stay in the cpp file please

jesserockz (changes requested):

esphome/components/ltr390/ltr390.h
sjtrny: Thanks, I've moved them into the cpp file along with other consts defined above that.

jesserockz (approved):

PR #6165, opened 2024-02-02
Float Output new feature: min_turn_on_power

Alex1s:

I am building a lightstrip with a LED driver.
The LED driver needs a minimum brightness to turn on. Once it is turned on though, it is possible to dim it down to a brightness less than was needed to turn it on.
To deal with this a added two optional configuration options to the ledc output component:
- **min_duty_turn_on** This is the minimum duty cycle needed to switch from a state where it is off to a state where it emits some light
- **min_duty** The minimum duty cycle for which some light is emitted

**min_duty_turn_on** works as follows:  When the light is currently off and a brightness is requested which is below **min_duty_turn_on**, the ledc component will set the duty cycle to **min_duty_turn_on** so that the LED turns on. Then it will dimm the LED down to the user requested brightness.

**min_duty_turn_on** will map a requested brightness / duty cycle which is in the interval [0,1] to the interval [min_brightness,1] lineary.

I did not find any tests for the ledc component and thus did not add anything to the `tests` folder.

## Types of changes

- [ ] Other

Not applicable

I don't have a PR yet, but the updated documentation can be found here: [https://github.com/Alex1s/esphome-docs/blob/lightstrip/components/output/ledc.rst](https://github.com/Alex1s/esphome-docs/blob/lightstrip/components/output/ledc.rst)

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```
esphome:
  name: lightstrip-esp32-c3-super-mini
  friendly_name: Lightstrip ESP32 C3 Super Mini
  platformio_options:
    board_build.flash_mode: dio  # QIO does not seem to work

esp32:
  board: esp32-c3-devkitm-1  # is actually a c3_super_mini
  variant: esp32c3
  framework:
    type: esp-idf  # "This is an alternative base framework for ESP32 chips, and recommended for variants of the ESP32 like ESP32S2, ESP32S3, ESP32C3 and single-core ESP32 chips."

# Enable logging
logger:
  level: VERY_VERBOSE

# Enable Home Assistant API
api:
  encryption:
    key: !secret api_encryption_key
  reboot_timeout: 0s

ota:

wifi:
  ssid: !secret wifi_ssid

  ap:
    ssid: "Lightstrip Fallback Hotspot"
    password: !secret wifi_ap_password

captive_portal:

binary_sensor:
  - platform: gpio
    id: up
    pin:
      number: 0
      mode:
        input: true
        pulldown: true
    on_press:
      then:
        light.dim_relative:
          id: lightstrip
          relative_brightness: 1%
  - platform: gpio
    id: down
    pin:
      number: 1
      mode:
        input: true
        pulldown: true
    on_press:
      then:
        light.dim_relative:
          id: lightstrip
          relative_brightness: -1%
  - platform: gpio
    id: power
    pin:
      number: 3
      mode:
        input: true
        pulldown: true
    on_press:
      then:
        - logger.log: "power is pressed"
        - if:
            condition:
              light.is_off: lightstrip
            then: 
              light.turn_on:
                id: lightstrip
                brightness: 100%
            else:
              light.turn_off:
                id: lightstrip

output:
  - platform: ledc
    pin: 10
    id: ledc_id
    frequency: 1000Hz  # highest frequency supported by the LED driver
    min_duty: 0.15%
    min_duty_turn_on: 5%

light:
  - platform: monochromatic
    output: ledc_id
    id: lightstrip
    name: "Lightstrip"
    default_transition_length: 0s
#    gamma_correct: 1
```yaml
# Example config.yaml

```

## Checklist:


ssieb:
A change like this should be done on the base float output component like the other similar options.

kbx81:
Yes, as @ssieb mentioned, making this unique to `ledc` means it can't be used with other types of PWM-generating hardware such as the [PCA9685](https://esphome.io/components/output/pca9685).

If you want this feature merged, please make it a part of the float `output` component.

ssieb:
This is a feature that has been requested before and it's not specific to pwm either.  However, the `min_duty` option is already covered by output `min_power` option.  The `min_duty_turn_on` needs to be renamed to `min_turn_on_power` and should be an absolute value, not scaled according to the `min_power` value.  It should be checked against that in the config though.  Also, the time to scale back after turn on needs to be configurable as well.

Alex1s:
Thank you for you feedback!

You are right: `min_duty` can be achieved by using `min_power` and `zero_means_zero`. Thus i removed this feature.

Additionally I moved all my changes to the float output. I renamed `min_duty_turn_on` to `min_turn_on_power`. The value of `min_turn_on_power` is not scaled. The delay can be configured using `min_turn_on_power_delay`. Also the documentation at [https://github.com/Alex1s/esphome-docs/tree/lightstrip](https://github.com/Alex1s/esphome-docs/tree/lightstrip) is updated accordingly.

Looking forward for any additional feedback.

esphome/components/output/__init__.py
```diff
@@ -58,6 +64,10 @@ async def setup_output_platform_(obj, config):
         cg.add(obj.set_min_power(config[CONF_MIN_POWER]))
     if CONF_ZERO_MEANS_ZERO in config:
         cg.add(obj.set_zero_means_zero(config[CONF_ZERO_MEANS_ZERO]))
+    if CONF_MIN_TURN_ON_POWER in config:
+        cg.add(obj.set_min_turn_on_power(config[CONF_MIN_TURN_ON_POWER]))
+    if CONF_MIN_TURN_ON_POWER_DELAY in config:
+        cg.add(obj.set_min_turn_on_power_delay(config[CONF_MIN_TURN_ON_POWER_DELAY]))
```
jesserockz: Having defaults means they will always be in config
```suggestion
    cg.add(obj.set_min_turn_on_power(config[CONF_MIN_TURN_ON_POWER]))
    cg.add(obj.set_min_turn_on_power_delay(config[CONF_MIN_TURN_ON_POWER_DELAY]))
```

esphome/components/output/float_output.cpp
```diff
@@ -1,6 +1,7 @@
 #include "float_output.h"
 #include "esphome/core/log.h"
 #include "esphome/core/helpers.h"
+#include "esphome/core/hal.h"  // delay
```
jesserockz:
```suggestion
#include "esphome/core/hal.h"
```

esphome/components/output/float_output.cpp
```diff
@@ -37,7 +54,23 @@ void FloatOutput::set_level(float state) {
 
   if (this->is_inverted())
     state = 1.0f - state;
+
+  if (!this->is_on_() && 0 < state && state < this->min_turn_on_power_) {
+    ESP_LOGV(TAG, "FloatOutput is turned on and set to a low power (%f < %f), thus we set the min power (%f) first",
+             state, this->min_turn_on_power_, this->min_turn_on_power_);
+    this->write_state(this->is_inverted() ? 1.0f - this->min_turn_on_power_ : this->min_turn_on_power_);
+    ESP_LOGV(TAG, "Staying at minimum power for %ums", this->min_turn_on_power_delay_);
+    delay(this->min_turn_on_power_delay_);
```
jesserockz, line 63: This has potential to cause issues if it is a long delay and should be written to use `set_timeout` instead

jesserockz (changes requested):

esphome/components/output/float_output.cpp
Alex1s: Currently I can not work on this as I do not have the hardware. How about we simply add a warning to the documetation?

esphome/components/output/__init__.py
Alex1s: Are you sure about this? When I commit your suggestion the tests fail with `KeyError: 'min_turn_on_power'`.

PR #6167, opened 2024-02-02
Separate logger implementations for each hardware platform into different files

tomaszduda23:

Make logger code a little bit more readable.

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] ESP32-S2

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/logger/logger_rp2040.cpp
```diff
@@ -0,0 +1,40 @@
+#ifdef USE_RP2040
+#ifndef USE_ARDUINO
+#error "Only ARDUINO is supported"
+#endif
```
jesserockz: Its not possible in ESPHome to not use Arduino at this point in time so this is not needed.
```suggestion
```

esphome/components/logger/logger_esp8266.cpp
```diff
@@ -0,0 +1,46 @@
+#ifdef USE_ESP8266
+#ifndef USE_ARDUINO
+#error "Only ARDUINO is supported"
+#endif
```
jesserockz: Same here
```suggestion
```

esphome/components/logger/__init__.py
```diff
@@ -274,6 +276,16 @@ async def to_code(config):
             add_idf_sdkconfig_option("CONFIG_ESP_CONSOLE_USB_CDC", True)
         elif config[CONF_HARDWARE_UART] == USB_SERIAL_JTAG:
             add_idf_sdkconfig_option("CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG", True)
+    try:
+        uart_selection(USB_SERIAL_JTAG)
+        cg.add_build_flag("-DUSE_USB_SERIAL_JTAG")
```
jesserockz:
```suggestion
        cg.add_define("-DUSE_LOGGER_USB_SERIAL_JTAG")
```

jesserockz (dismissed):

esphome/components/logger/__init__.py
```diff
@@ -127,6 +127,8 @@ def uart_selection(value):
         if CORE.using_arduino and value.upper() in ESP_ARDUINO_UNSUPPORTED_USB_UARTS:
             raise cv.Invalid(f"Arduino framework does not support {value}.")
         variant = get_esp32_variant()
+        if CORE.using_esp_idf and variant == VARIANT_ESP32C3 and value == USB_CDC:
+            raise cv.Invalid(f"esp idf variant {variant} does not support {value}.")
```
kbx81:
```suggestion
            raise cv.Invalid(f"{value} is not supported for variant {variant} when using ESP-IDF .")
```

kbx81 (commented):
I've tested this extensively on 8266, ESP32, S2, S3 and C3 with both USB CDC/JTAG (where supported) and serial adapter IC. Everything looks good. Just one little nitpick to make the error message a bit clearer.

kbx81 (dismissed):

esphome/components/logger/__init__.py
```diff
@@ -127,6 +127,8 @@ def uart_selection(value):
         if CORE.using_arduino and value.upper() in ESP_ARDUINO_UNSUPPORTED_USB_UARTS:
             raise cv.Invalid(f"Arduino framework does not support {value}.")
         variant = get_esp32_variant()
+        if CORE.using_esp_idf and variant == VARIANT_ESP32C3 and value == USB_CDC:
+            raise cv.Invalid(f"{value} is not supported for variant {variant} when using ESP-IDF .")
```
kbx81:
```suggestion
            raise cv.Invalid(
                f"{value} is not supported for variant {variant} when using ESP-IDF."
            )
```

kbx81 (dismissed):

esphome/components/logger/logger.cpp
```diff
@@ -211,6 +119,12 @@ void HOT Logger::log_message_(int level, const char *tag, int offset) {
   if (xPortGetFreeHeapSize() < 2048)
     return;
 #endif
+
+  this->log_callback_.call(level, tag, msg);
+}
+
+#if defined(USE_HOST) || defined(USE_ARDUINO)
```
jesserockz: I feel like since you are splitting them per platform, you should just split this out into each platform file to make it really clear.

jesserockz (changes requested):

esphome/components/logger/logger.cpp
tomaszduda23: My goal here was to get rid of complex `#ifdef` since those are hard to read. I was thinking about logger_host.cpp as well. I didn't want to do that since more code would be duplicated that way. With curent structure it is not possible to get rid of platform specific things from common files completely. You would have to also do logger_esp32_arduino etc. It seems to be nice balance.

jesserockz: I understand the code would be duplicated, but I would rather have it split as much as possible if we are gonna do it at all.

The reason I noticed this was because when I was looking at the logger_esp32.cpp file, I was trying to find where Arduino logging was and couldn't find it which makes it more confusing I think.

I don't think you would need `logger_esp32_arduino.cpp` as the `pre_setup` code inside `logger_esp32.cpp` already does stuff for Arduino etc.

jesserockz (approved):
Thanks for making that final change. LGTM now

PR #6184, opened 2024-02-06
WLED Sync fix and BK72XX support

ChuckMash:
moved to #6190


ChuckMash:
I need to tidy up and resubmit.

PR #6192, opened 2024-02-07
Nextion new unified TFT upload engine

edwardtfn:

This unifies the TFT upload engine between `arduino` and `esp-idf`, which increases consistency in the code and brings the advantages previously existing in one code to the other.

These are the changes included in this PR:

1. ESP-IDF now supports the persistent connection to http server, aligning with the previous Arduino implementation.
2. Arduino now transfers with a single stream, reducing the number of requests to the http server. In fact, this is similar to increasing the previously used chunk size to the file size, so all the data is requested to the http server in a single request, but then transferred to the UART on 4096 bytes chunks (as required by Nextion). If the display requests a different position, then the next request is the new http request (still in a persistent connection).
3. The `upload_tft` function now accept two parameters to define a target baud rate for the transfer (as Nextion supports the transfer in an alternative baud rate) and if an exit reparse command should be sent before starting the transfer, as the Nextion Upload Protocol defines the device should leave reparse mode before the transfer. Those two parameters are optional, so this is not breaking existing implementations.
4. The `upload_tft` function now returns an ENUM containing the result of the transfer, which can be either an `OK` or specific error related information. **This is a breaking change** as this function previously (since a couple of releases ago) was returning a `bool`.

The inline doc says all the details:
```cpp
  /**
   * @brief Uploads the TFT file to the Nextion display.
   *
   * This function initiates the upload of a TFT file to the Nextion display. Users can specify a target baud rate for
   * the transfer. If the provided baud rate is not supported by Nextion, the function defaults to using the current
   * baud rate set for the display. If no baud rate is specified (or if 0 is passed), the current baud rate is used.
   *
   * Supported baud rates are: 2400, 4800, 9600, 19200, 31250, 38400, 57600, 115200, 230400, 250000, 256000, 512000
   * and 921600. Selecting a baud rate supported by both the Nextion display and the host hardware is essential for
   * ensuring a successful upload process.
   *
   * @param baud_rate The desired baud rate for the TFT file transfer, specified as an unsigned 32-bit integer. 
   * If the specified baud rate is not supported, or if 0 is passed, the function will use the current baud rate.
   * The default value is 0, which implies using the current baud rate.
   * @param exit_reparse If true, the function exits reparse mode before uploading the TFT file. This parameter
   * defaults to true, ensuring that the display is ready to receive and apply the new TFT file without needing
   * to manually reset or reconfigure. Exiting reparse mode is recommended for most upload scenarios to ensure
   * the display properly processes the uploaded file command.
   * @return Nextion::TFTUploadResult Indicates the outcome of the transfer. Returns Nextion::TFTUploadResult::OK for
   * success, otherwise, indicates failure with additional details on the operation status. These additional details
   * can include specific error codes or messages that help identify the cause of failure, making troubleshooting more
   * straightforward.
   */
  Nextion::TFTUploadResult upload_tft(uint32_t baud_rate = 0, bool exit_reparse = true);
```
To support that return from `upload_tft`, there is a friendly message for each of the `TFTUploadResult` values, which can be retrieved with the following:
```cpp
  /**
   * @brief Converts TFTUploadResult enum values to their corresponding string representations.
   *
   * This function is used to obtain human-readable strings for logging or displaying
   * the outcomes of TFT upload operations to a Nextion display. It maps each enum value
   * defined in TFTUploadResult to a descriptive text message.
   *
   * @param result The TFTUploadResult enum value to be converted to a string.
   * @return A const char* pointing to the string representation of the given result.
   */
  static const char *tft_upload_result_to_string(TFTUploadResult result);
```

> [!IMPORTANT]
> This is not a small change, as it changes substantially the Upload TFT engine.
> I would appreciate if anyone could test it and give me feedback.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3598

## Test Environment

- [X] ESP32
- [X] ESP32 IDF (both v4 and v5)
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
button:
  - name: Upload TFT
    platform: template
    on_press:
      lambda:
        id(disp1).upload_tft(0, false);

  - name: Exit Reparse and Upload TFT
    platform: template
    on_press:
      lambda: id(disp1).upload_tft(0);

  # This will exit reparse only if display isn't setup (so probably in reparse mode)
  # and then upload at 92160 bps, showing the result message in the log.
  - name: Upload TFT at 921600bps
    platform: template
    on_press:
      - lambda: |-
          static const char *const TAG = "addon_upload_tft.script.nextion_upload";
          ESP_LOGD(TAG, "Starting TFT upload...");
          esphome::nextion::Nextion::TFTUploadResult upload_tft_result = disp1->upload_tft(921600, !disp1->is_setup());
          ESP_LOGD(TAG, "TFT upload: %s", esphome::nextion::Nextion::tft_upload_result_to_string(upload_tft_result));
```

## Checklist:

## Pending actions:
  - [X] Find someone who can test on ESP8266. Although I've avoided touching on that code, it could be affected.
  - [X] ~Review if needed to document (include config example) new response from upload_tft(), as this is a breaking change for the ones who implemented the bool response (#5667 only for `esp-idf`) - Wait for feedback in the feature first.~ Described here only.
  - [X] The [text about HTTPS](https://esphome.io/components/display/nextion.html#uploading-a-tft-file) should be updated. This is using unsafe connection over HTTPS (and it is like this since a while for IDF) and it works pretty fine with an ESP32. But some changes on that page should be changed in another PR.


edwardtfn:
I'm dropping this PR for now as I'm finding some issues in my tests and cannot find time to work at his right now. I will create a new one when it is more in a ready state.

edwardtfn:
I believe I got all issues fixed. 😀

edwardtfn:
Hey @kbx81, is there anything pending to have this PR moving? Or is there anything in the PR against ESPHome best practices that is blocking it?
Please don't take me wrong, I'm not trying to make any pressure, just understand if something isn't following the best practices for a smooth merge. I will be happy to change things in the PR if that will align more with ESPHome best practices.

kbx81:
No worries @edwardtfn -- I haven't dug into it just yet. As this is a pretty big PR, it would be good to get some additional eyes on it, too. I'll try to get through it as soon as I can -- I did have a look at the docs PR and just pushed an update to it, as well. (Our Nextion docs need _a lot_ of help...)

edwardtfn:
> Our Nextion docs need a lot of help...

Yeah, I have to say I'm not a big fan of that page, but I think we can slowly improve that. I would be happy to help.

I think also we could have some of the commands from `display` component working with Nextion. Many of the display methods could be easily translated to a Nextion command and that will make this library a bit easier for the ones with previous experience in other displays. And this can all be done without breaking changes.
I'm about to setup a testing environment with an ESP8266 connected to a Nextion display, which will give me more confidence to try things without breaking for legacy users.

So, back to this PR, please let me know if I can help with something. I'm open to do changes to make it closer to ESPHome best practices if needed.

And no rush. I'm using it anyways, with a reference to the PR. Just wanna go back to the main branch to have a more reliable solution from my side. 😉

esphome/components/nextion/nextion_upload.cpp
```diff
@@ -0,0 +1,565 @@
...
+    } else {
+      ESP_LOGE(TAG, "Failed to read from HTTP client, error code: %d", read_len);
+      break;  // Exit the loop on error
+    }
+  }
+  range_start = range_end + 1;
+  return Nextion::TFTUploadResult::OK;
+}
+
+Nextion::TFTUploadResult Nextion::upload_tft(uint32_t baud_rate, bool exit_reparse) {
```
kbx81: Would it be better to keep this broken out into each of the two `.cpp` files for Arduino/IDF? I realize there will be some duplicated code, but the better part of this function has bits wrapped in `ifdef` and it seems like it'd be easier to follow if it were split out that way. The macros don't improve readability. 🙂

esphome/components/nextion/nextion.h
```diff
@@ -913,17 +926,211 @@ class Nextion : public NextionBase, public PollingComponent, public uart::UARTDe
...
+     * @brief Failed to open a connection to the HTTP server.
+     * Check the network connection and server URL.
+     */
+    HTTP_ERROR_FAILED_TO_OPEN_CONNECTION,
+
+    /**
+     * @brief Failed to get content length from the HTTP server.
+     * Verify the server's response headers.
+     */
+    HTTP_ERROR_FAILED_TO_GET_CONTENT_LENGHT,
```
kbx81: Spelling 😄

```suggestion
    HTTP_ERROR_FAILED_TO_GET_CONTENT_LENGTH,
```

esphome/components/nextion/nextion_upload.cpp
edwardtfn: I don't have any problem on splitting in two files, as my goal is to unify the engine. The macros will still be there for ESP8266, but certainly will be much easier to read.
I will split that.

edwardtfn: I've split into different files, but still pending some tests. Hopefully I can run those tomorrow.

edwardtfn: I've tested with ESP32 both arduino and idf and all looks good.

edwardtfn:
I've finally tested this with ESP8266. An issue was found, but it was already fixed.

edwardtfn:
I will work more at this to improve memory consumption during TFT update.

esphome/components/nextion/nextion_upload.cpp
edwardtfn: After maintaining this with quite a few changes, I can clear see that splitting in multiple files make this harder, as every change must be done in two places.
I'm not saying we should merge again, just providing some inputs to the discussion. 😉

edwardtfn:
The core of this PR still the same, but I've done so many other changes not related to the title that I'm unsure if I should rephrase that.

kbx81:
> The core of this PR still the same, but I've done so many other changes not related to the title that I'm unsure if I should rephrase that.

Multiple, smaller PRs are easier to review. If you can break this one up in a logical way, I think we would prefer that. 😇

edwardtfn:
I will break down this in smaller pieces (where possible).

kbx81:
@edwardtfn would you also mind adding yourself to the CODEOWNERS for the Nextion component(s)? @SenexCrenshaw no longer seems to be active here. 😇

PR #6194, opened 2024-02-07
fix rp2040_pio_led flicker and proper multi-strip support

### LINKED ISSUE #5158: RP2040 PIO LED Strip library causes device to fail when using more than 2 strips
### The problem

I am trying to hook up 5 rgb led strips to a pi pico w and I got it working no problem for a single led strip however once I got past two strips it then either fails to connect to wifi or if it does eventually connect to wifi it fails to connect to the api socket. The rp2040 chip also gets very hot when I tried to connect all 5 (thought I killed my pico w until I uploaded a different sketch and confirmed it still works). It looks like its making the wifi connection drop periodically but I'm not sure. I've tried different combinations of using pio 0 and 1 but that doesn't seem to make a difference in getting me past 2 led strips even though its suppose to be able to handle 4 per pio

2023.11.4

Home Assistant Add-on

2023.11.3

RP2040

### Board

Pi Pico W

RP2040 PIO LED Strip

### Example YAML snippet

```yaml
light:
  - platform: partition
    name: "LEDs"
    segments:
      - id: strip0
        from: 0
        to: 49
      - id: strip1
        from: 0
        to: 49
      # - id: strip2
        # from: 0
        # to: 49
      # - id: strip3
      #   from: 0
      #   to: 49
      # - id: strip4
      #   from: 0
      #   to: 49
  - platform: rp2040_pio_led_strip
    id: strip0
    chipset: ws2812b
    pin: GPIO18
    num_leds: 50
    rgb_order: RGB
    pio: 1
    internal: true
  - platform: rp2040_pio_led_strip
    id: strip1
    chipset: ws2812b
    pin: GPIO19
    num_leds: 50
    rgb_order: RGB
    pio: 1
    internal: true
  - platform: rp2040_pio_led_strip
    id: strip2
    chipset: ws2812b
    pin: GPIO20
    num_leds: 50
    rgb_order: RGB
    pio: 1
    internal: true
  # - platform: rp2040_pio_led_strip
  #   id: strip3
  #   chipset: ws2812b
  #   pin: GPIO21
  #   num_leds: 50
  #   rgb_order: RGB
  #   pio: 0
  #   internal: true
  # - platform: rp2040_pio_led_strip
  #   id: strip4
  #   chipset: ws2812b
  #   pin: GPIO22
  #   num_leds: 50
  #   rgb_order: RGB
  #   pio: 1
  #   internal: true
```

_No response_

### Additional information

_No response_

### LINKED ISSUE #5450: LED Strip flickering because of interrupts while filling the PIO queue
@Papa-DMan tagging you, because you are listed as the code owner.

I noticed flickering of longer LED strips, when using the `rp2040_pio_led_strip` component on my Pico W. 
output_320p
(video is at half speed)

I tracked it down to the PIO state machine stalling on a low bit, which is interpreted by the LEDs as the reset signal. 

This stall only occurs occasionally, when an interrupt is fired while filling the PIO queue in `write_state`. 

https://github.com/esphome/esphome/blob/92798751c2a0860dc5f85cf9f34395372cca72d6/esphome/components/rp2040_pio_led_strip/led_strip.cpp#L72-L80

This problem gets worse the longer the LED strip is, and might not be that noticeable on shorter strips. Especially if the entire data fits into the PIO queue. (≤10 RGB LEDs)

I improvised a fix by turning interrupts off during the loop and then re-enable them. No more flickering. 🎉 But I don't know what side effects this may have. 😕

```cpp
#include <hardware/sync.h>

uint32_t irqs = save_and_disable_interrupts();
// loop
restore_interrupts(irqs);
```

Additionally, this problem could be solved in different ways: 

1. Disable interrupts for the loop. (I don't know what side effects this may have.)
2. Run the fill loop on the second core. (← could be the best solution)
3. Modify the PIO code so that it stalls while the output is high. But then it needs to be able to discern a stall from the end of the transmission. 

----

This is my first time with C++, so I decided against a pull request for now.

I have tested other improvements for `rp2040_pio_led_strip` as well, like shaving down the fill loop and the PIO code. 

2023.12.9

Home Assistant Add-on

RP2040

### Board

Pico W

rp2040_pio_led_strip

Papa-DMan:

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5450 and fixes https://github.com/esphome/issues/issues/5158

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
light:
  - platform: rp2040_pio_led_strip
    name: LED test
    id: led_strip
    pin: GPIO28
    num_leds: 30
    pio: 0
    rgb_order: GRB
    chipset: WS2812B
    effects:
      - addressable_rainbow:
```

## Checklist:


esphome/components/rp2040_pio_led_strip/led_strip.cpp
```diff
@@ -34,24 +35,71 @@ void RP2040PIOLEDStripLightOutput::setup() {
...
 
+  // Initialize the PIO program
+
   // Select PIO instance to use (0 or 1)
-  this->pio_ = pio0;
   if (this->pio_ == nullptr) {
     ESP_LOGE(TAG, "Failed to claim PIO instance");
     this->mark_failed();
     return;
   }
```
jesserockz, line 54: Is this not going to be nullptr all the time now?

Papa-DMan: No when the program gets compiled down it's my understanding that the set_pio(int) function in the header [file](https://github.com/esphome/esphome/blob/dev/esphome/components/rp2040_pio_led_strip/led_strip.h#L68) will get generated and  called based on the yaml before the final setup function runs.

jesserockz: AH my bad, that was outside of the diff so I didnt see it.

esphome/components/rp2040_pio_led_strip/led_strip.cpp
```diff
@@ -132,6 +172,20 @@ void RP2040PIOLEDStripLightOutput::dump_config() {
...
+  } else if (strncmp(chipset.c_str(), "WS2812B", 7) == 0) {
+    this->chipset_ = 1;
+  } else if (strncmp(chipset.c_str(), "SK6812", 6) == 0) {
+    this->chipset_ = 2;
+  } else if (strncmp(chipset.c_str(), "SM16703", 7) == 0) {
+    this->chipset_ = 3;
+  } else {
+    this->chipset_ = 4;
+  }
+}
```
jesserockz: This would be much nicer if it was just an `enum`, and have CUSTOM as `0x00` or `0xFF`

jesserockz (changes requested):

jesserockz (approved):

PR #4579, opened 2023-03-17
Eink weact 2.13/2.9 support

alinelena:
# What does this implement/fix?

<!-- Quick description and explanation of changes -->

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>
https://github.com/esphome/feature-requests/issues/2101

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040

## Example entry for `config.yaml`:
<!--
  Supplying a configuration snippet, makes it easier for a maintainer to test
  your PR. Furthermore, for new integrations, it gives an impression of how
  the configuration would look like.
  Note: Remove this section if this PR does not have an example entry.
-->

```yaml
- platform: waveshare_epaper                                                                                                                                                                                                                        
  model: "2.90inBS" 
  model: "2.13inBN" 

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6217, opened 2024-02-13
Eink 2.9 inch weact(probably others) support

alinelena:

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

this is practically this pr https://github.com/esphome/esphome/pull/4579 which github closed in eagerness.

### DIFF
--- a/esphome/components/waveshare_epaper/waveshare_213v3.cpp
+++ b/esphome/components/waveshare_epaper/waveshare_213v3.cpp
@@ -52,7 +52,8 @@ static const uint8_t BORDER_PART[] = {0x3C, 0x80};  // border waveform
 static const uint8_t BORDER_FULL[] = {0x3C, 0x05};  // border waveform
 static const uint8_t CMD1[] = {0x3F, 0x22};
 static const uint8_t RAM_X_START[] = {0x44, 0x00, 121 / 8};           // set ram_x_address_start_end
-static const uint8_t RAM_Y_START[] = {0x45, 0x00, 0x00, 250 - 1, 0};  // set ram_y_address_start_end
+// static const uint8_t RAM_Y_START[] = {0x45, 0x00, 0x00, 250 - 1, 0};  // set ram_y_address_start_end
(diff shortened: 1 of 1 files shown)


alinelena:
at this stage I would like feedabck of the most acceptable way to integrate these changes in the esphome. changes are really minimal.

alinelena:
@jbergler reopened in here... I will need to check the 3colours i wonder if your change is relevant for bw only screens too.

clydebarrow (changes requested):
These changes modify an existing display model, which will break support for that. If you are adding support for a new model you need create a new one, copying relevant code and modifying as required. That will need changes in the CPP code and Python code.

alinelena:
> These changes modify an existing display model, which will break support for that. If you are adding support for a new model you need create a new one, copying relevant code and modifying as required. That will need changes in the CPP code and Python code.

sorry I was not very clear, I put only to show what are the changes, I am not gonna modify that screen.. just wanted to know what is the best way to add support for 2.9 given the changes are minimal. I would like not to create an entire 2.9 file that just duplicates 99% of 2.13

clydebarrow:
> I would like not to create an entire 2.9 file that just duplicates 99% of 2.13

That's the quick, easy and somewhat clunky option. The other would be to refactor the code to have a single class with suitable parameters for the different models. That involves more work, will take longer to review, and if done properly the scope could get quite broad.

Refactoring _should_ result in a better end result, but unless you have the time and skills to do it well, it's probably best to go for the the quick and easy option right now.

See https://github.com/esphome/esphome/pull/6226 for an example.

alinelena:
I have the 2.13 version of this and I can test,I remember from playing with it on another system, changes are minimal between the two vesions, similar with the bw version. 
do you want to add support for both given I test the 2.13?

PR #6222, opened 2024-02-13
Adjust HeatpumpIR dependency

ivankravets:
You can remove implicit dependency (`crankyoldgit/IRremoteESP8266@~2.8.4`) later when PR is merged https://github.com/ToniA/arduino-heatpumpir/pull/167


jesserockz (changes requested):
This needs to match also in the codegen
https://github.com/esphome/esphome/blob/558588ee8ae3e0b8f95d9a8f53b23dd92031a047/esphome/components/heatpumpir/climate.py#L122

ivankravets:
@jesserockz , thanks, updated!

jesserockz (approved):

PR #6168, opened 2024-02-03
ld2420: Firmware v1.5.4+ bug workaround

descipher:
# What does this implement/fix?

This is a workaround to a firmware bug that started in v1.5.4 where if we send write registers updates
faster than 22 per second the module will not respond correctly and will ignore further commands.
The symptoms are constant detection at the 35cm range and no communication with it until a power cycle is done..

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5115

## Test Environment

- [x] ESP32

## Example entry for `config.yaml`:
N/A

## Checklist:
  - [x] The code change is tested and works locally.

PR #6225, opened 2024-02-14
ld2420: fix energy mode documentation

andresv:

LD2420 energy output mode documentation was incorrect. In energy output mode it returns:
```rust
presence: bool
distance: u16
energies: [u16; 16]
```
Also length was wrong in the example. Correct is `23 00` instead of `00 23`.

Here can be seen that parser also already expects energies right after distance field: https://github.com/esphome/esphome/blob/dev/esphome/components/ld2420/ld2420.cpp#L351-L354.

## Types of changes

- [x] Other


descipher:
Thanks for catching the error, I have a PR in to do a workaround at the moment #6168 If you could do review comment there we can get it done faster that way.

jesserockz (approved):

PR #6255, opened 2024-02-21
Check permissions

### LINKED ISSUE #2604: Better error message when flashing with improper permissions
On Linux, when attempting to flash a device, depending on the tty permissions, it might fail to connect and give the error: `ERROR Running command failed: Could not open /dev/ttyUSB0, the port doesn't exist`. The tty required the dialout group, and my user was not a member of the dialout group.
I would like this error message to be more specific, as it doesn't indicate that the port could exist but I do not have the correct permissions to see it. A better error message would tell me to check the user privileges and give the command to add the user to dialout.

**Additional context**

OdileVidrine:
An implementation of  https://github.com/esphome/feature-requests/issues/2604. In Linux, when attempting to install onto a device and the user does not have permission to access the port, the error message will now print a message explaining how to add the user to the dialout group. The old version of the error message said that the port does not exist, which is incorrect.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2604

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


jesserockz:
I was able to test the first use case where port is not available, but I tried removing myself from the `plugdev` and `uucp` groups (Arch Linux) and I could still open the serial port :cry:

jesserockz:
Ok ignore that, got it working. I must have set my system up for 666 permissions on usb devices :see_no_evil:

jesserockz (approved):

PR #6258, opened 2024-02-21
Add setting to customise the reported manufacturer

sybrenstuvel:
Allow setting the device manufacturer from the YAML config.

This makes it possible to have a custom "manufacturer" reported to API consumers like HomeAssistant. For the devices that I've made myself, I like to have my own name in there, and not "Espressif".

## Types of changes

- [ ] Other

Will update documentation once this PR has been approved, at least on the "yeah this is a good change" level.

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: my_device
  friendly_name: My Device
  manufacturer: My Name
```

If the `manufacturer` key is not given or empty, it will default to the current behaviour of a hardware-dependent string (Espressif, Raspberry Pi, etc.)

## Checklist:


Cossid:
This might be partially duplicate of the `project` property, and might conflict.

Additionally, if you do `manufacturer` you might as well do `model` as well.

The conflict would need to be mitigated in the metadata builder in the API, if both project and mfg/model are defined, which currently has case conditions for when to use it.  There are limitations about which characters can be used in `project` because of the parsing involved, I think you can't use spaces or periods currently, so maybe project could be deprecated in favor of clearer manufacturer/model that don't need special parsing or character consideration.

sybrenstuvel:
Thanks for the info. I'll abandon this PR, I can reach what I need with non-breaking spaces and non-ASCII-dot punctuation marks in the project name string like this:

```yaml
  project:
    name: "dr․ Sybren.VentilatorLEDs"
    version: "v1.0"
```

PR #6180, opened 2024-02-05
CSE7766: fix power and current measurements at low loads

twasilczyk:
# What does this implement/fix?

CSE7766 (Sonoff S31) with the current implementation reports bad values at low loads. This PR fixes this along with some cleanup.

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Other

**Related issue or feature (if applicable):** fixes a bug that hasn't been officially reported outside of this PR

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
esphome:
  name: s31
  friendly_name: S31

esp8266:
  board: esp01_1m

# Enable logging
logger:
  baud_rate: 0 # (UART logging interferes with cse7766)
  level: VERY_VERBOSE
  logs:
    sensor: VERBOSE
    api: INFO
    api.connection: INFO
    api.service: INFO
    scheduler: INFO
    mdns: INFO
    wifi: INFO

api:

ota:

wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_password

captive_portal:

uart:
  rx_pin: RX
  baud_rate: 4800

binary_sensor:
  - platform: gpio
    pin:
      number: GPIO0
      mode: INPUT_PULLUP
      inverted: True
    name: "Sonoff S31 Button"
    on_press:
      - switch.toggle: relay
  - platform: status
    name: "Sonoff S31 Status"

sensor:
  - platform: wifi_signal
    name: "Sonoff S31 WiFi Signal"
    update_interval: 60s
  - platform: cse7766
    #update_interval: 1s # XXX: this works with release branch, but not dev!
    current:
      name: "Sonoff S31 Current"
      accuracy_decimals: 4
    voltage:
      name: "Sonoff S31 Voltage"
      accuracy_decimals: 1
    power:
      name: "Sonoff S31 Power"
      accuracy_decimals: 2

switch:
  - platform: gpio
    name: "Sonoff S31 Relay"
    pin: GPIO12
    id: relay
    restore_mode: ALWAYS_ON

status_led:
  pin:
    number: GPIO13
    inverted: True
```

## Checklist:
  - [X] The code change is tested and works locally.
[...]

PR #6261, opened 2024-02-21
Always publish energy to address esphome/issues#5501

### LINKED ISSUE #5501: CSE7766 hard to use since ESPHome 2024.2.0
### The problem

image

</details>

2024.2.0

Home Assistant Add-on

2024.2.2

ESP8266

### Board

Sonoff S31

cse7766

### Example YAML snippet

```yaml
esphome:
  name: s31
  friendly_name: S31

esp8266:
  board: esp01_1m

logger:
  baud_rate: 0

api:

ota:

wifi:
  ap:

captive_portal:

uart:
  rx_pin: RX
  baud_rate: 4800

switch:
  - platform: gpio
    name: None
    id: relay
    pin: GPIO12

binary_sensor:
  - platform: gpio
    id: power_button
    entity_category: diagnostic
    pin:
      number: GPIO0
      mode: INPUT_PULLUP
      inverted: true
    on_press:
      - switch.toggle: relay

sensor:
  - platform: cse7766
    current:
      name: Current
      id: current
    voltage:
      name: Voltage
      id: voltage
    power:
      name: Power
      id: power
    energy:
      name: Energy

status_led:
  pin: GPIO13
```

_No response_

### Additional information

_No response_

DAVe3283:

Energy would not be published after startup/reboot. All other sensors of the CSE7766 publish at \~10 Hz, so publish energy at the same time and let the user handle it the same way. It would publish at \~10 Hz when power is being consumed anyway, so they needed a filter regardless.

**HOWEVER** I believe reverting the original breaking change change (esphome/esphome#6095) is the correct solution here, not this band-aid: #6265 

The way the component works, the individual outputs drift out of sync with each other, which makes calculating things like power factor or apparent power impossible, since they have to be measured over the same time period.

The original implementation intentionally averaged everything together over a _single_ time period, which enabled calculating downstream values (power factor, apparent power) that had meaning. Notice how after only 2 hours of uptime, the sensors are now reporting over a 15-second window, instead of simultaneously as they should:

```
[12:05:09][D][sensor:093]: 'Uptime': Sending state 10122.79492 s with 0 decimals of accuracy
[12:05:22][D][sensor:093]: 'Energy': Sending state 0.53937 Wh with 0 decimals of accuracy
[12:05:25][D][sensor:093]: 'Voltage': Sending state 120.19772 V with 1 decimals of accuracy
[12:05:32][D][sensor:093]: 'Current': Sending state 0.20806 A with 2 decimals of accuracy
[12:05:37][D][sensor:093]: 'Power': Sending state 23.99924 W with 1 decimals of accuracy
```

I don't know how to solve this with filters, and think the original solution of building the averaging into the component itself was the right answer.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/issues#5501

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3620

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
external_components:
  - source: github://pr#6261
    components: [ cse7766 ]

sensor:
  - platform: cse7766
    current:
      name: Current
      id: current
      filters:
        - throttle_average: 60s
    voltage:
      name: Voltage
      id: voltage
      filters:
        - throttle_average: 60s
    power:
      name: Power
      id: power
      filters:
        - throttle_average: 60s
    energy:
      name: Energy
      filters:
        - throttle: 60s
```

## Checklist:

### DIFF
--- a/esphome/components/cse7766/cse7766.cpp
+++ b/esphome/components/cse7766/cse7766.cpp
@@ -161,11 +161,9 @@ void CSE7766Component::parse_data_() {
 
     energy = cf_diff * float(power_coeff) / 1000000.0f / 3600.0f;
     this->energy_total_ += energy;
-    if (this->energy_sensor_ != nullptr)
-      this->energy_sensor_->publish_state(this->energy_total_);
-  } else if ((this->energy_sensor_ != nullptr) && !this->energy_sensor_->has_state()) {
-    this->energy_sensor_->publish_state(0);
   }
+  if (this->energy_sensor_ != nullptr)
+    this->energy_sensor_->publish_state(this->energy_total_);
 
   float current = 0.0f;
   float calculated_current = 0.0f;


DAVe3283:
My preference would be to merge #6265 instead of this band-aid PR.

ssieb:
I'm curious how you can calculate power factor with averaged values.  That doesn't make sense to me.
If the throttle average outputs are drifting out of sync, I think that's a bug in the filter.

DAVe3283:
> I'm curious how you can calculate power factor with averaged values. That doesn't make sense to me.

It does require the power factor be approximately constant over the averaging period, but since most loads have a static power factor over the short-term (10-60 seconds), it works very well in practice.

For example, this is a car battery charger as it brings a dead battery up to the float voltage over the course of a couple weeks:
image

> If the throttle average outputs are drifting out of sync, I think that's a bug in the filter.

It looks like they start out of sync right after a reboot. Using the YAML in this PR, this is what I get right out of the gate:

```
[15:42:16][D][sensor:093]: 'Uptime': Sending state 63.01600 s with 0 decimals of accuracy
[15:42:23][D][sensor:093]: 'Energy': Sending state 0.00000 Wh with 3 decimals of accuracy
[15:42:39][D][sensor:093]: 'Power': Sending state 0.00000 W with 1 decimals of accuracy
[15:42:39][D][sensor:093]: 'Voltage': Sending state 122.94889 V with 1 decimals of accuracy
[15:42:43][D][sensor:093]: 'Current': Sending state 0.00000 A with 2 decimals of accuracy
```

So immediately after reboot the Power & Voltage readings come in 4 seconds before the Current reading does.

And while I suspect power factor and apparent power can still be calculated off this in the long-term (e.g. weeks as my example above) it will be much less responsive to changes on the order of the averaging period.

ssieb:
The offset will be consistent, but different on every boot.  It's a side effect of how the throttle filter does the timing.

DAVe3283:
Yeah, but that is a problem the old built-in averaging filter solved, so it still feels like a regression to me.

But if the consensus is the new way is preferred, I can live with the slightly less responsive power factor and active power calculations. Though this MR is still needed to prevent energy from showing "Unknown" at startup.

ssieb:
I would like to fix that offset issue in some way.
The energy issue appears to be caused by #6180 not #6095.

DAVe3283:
> I would like to fix that offset issue in some way. Is the energy issue caused by #6095 or #6180?

It was caused by an oversight in #6095, specifically an incorrect assumption that `this->energy_sensor_->has_state()` would return true only once a valid state had been published.

That is why my PR #6265 only reverts #6095 but keeps the changes from #6180.

ssieb:
How is that an incorrect assumption?  That's literally what that method is for.

DAVe3283:
~~Regardless of the intention of the function / code written there, as seen in the [issue I opened](https://github.com/esphome/issues/issues/5501), it doesn't actually report 0 on startup.~~

~~My read of the code ([esphome/components/sensor/sensor.cpp:90](https://github.com/esphome/esphome/blob/76a3ffc8a9961bab029a33f91b833a92b014c274/esphome/components/sensor/sensor.cpp#L90C1-L97C60)) sure looks like as soon as any state (including the initial NAN) get sent to the frontend, it will report that it has a state. But I could be reading it wrong. Either way, it doesn't report 0 on startup and this PR fixes that.~~

If there is a better way, let me know.

EDIT: I stand corrected, this appears to have been a filter I applied to try and tame the update rate.
But without reverting to the old behavior, after a reboot it doesn't report values consistently with a throttle_average filter:

image

```
[19:49:45][D][api.connection:1121]: Home Assistant 2024.1.5 (10.0.1.19): Connected successfully
[19:49:45][D][time:048]: Synchronized time: 2024-02-21 19:49:45
[19:49:49][D][sensor:093]: 'Power': Sending state nan W with 1 decimals of accuracy
[19:49:57][D][sensor:093]: 'Current': Sending state nan A with 2 decimals of accuracy
[19:49:59][D][sensor:093]: 'Voltage': Sending state nan V with 1 decimals of accuracy
[19:50:19][D][sensor:093]: 'Power': Sending state nan W with 1 decimals of accuracy
[19:50:21][D][sensor:093]: 'Uptime': Sending state 58.57100 s with 0 decimals of accuracy
[19:50:27][D][sensor:093]: 'Current': Sending state 0.00000 A with 2 decimals of accuracy
[19:50:29][D][sensor:093]: 'Voltage': Sending state 117.08681 V with 1 decimals of accuracy
[19:50:49][D][sensor:093]: 'Power': Sending state 0.00000 W with 1 decimals of accuracy
[19:50:57][D][sensor:093]: 'Current': Sending state nan A with 2 decimals of accuracy
[19:50:59][D][sensor:093]: 'Voltage': Sending state nan V with 1 decimals of accuracy
[19:51:19][D][sensor:093]: 'Power': Sending state nan W with 1 decimals of accuracy
```

DAVe3283:
After more experimentation, this is not the solution to the problems seen with the updated component. Closing this PR.

PR #6265, opened 2024-02-21
Revert #6095 but retain improvements from #6180

### LINKED ISSUE #5501: CSE7766 hard to use since ESPHome 2024.2.0

### LINKED ISSUE #5504: [update_interval] is an invalid option for [sensor.cse7766]. Please check the indentation.
### The problem

After receiving the latest update to the ESPHome add-on for Home assistant I get the following error when I try to update an athom-smart-plug-v2 device. The error is:

[update_interval] is an invalid option for [sensor.cse7766]. Please check the indentation.

2024.2.0

Home Assistant Add-on

2024.2.2

ESP8266

### Board

athom-smart-plug-v2

_No response_

### Example YAML snippet

_No response_

```txt
INFO ESPHome 2024.2.0
INFO Reading configuration /config/esphome/athom-smart-plug-v2-f1971b.yaml...
INFO Detected timezone 'Europe/Amsterdam'
Failed config

sensor.cse7766: [source /data/packages/43f53298/athom-smart-plug-v2.yaml:86]
  platform: cse7766
  
  [update_interval] is an invalid option for [sensor.cse7766]. Please check the indentation.
  update_interval: 10s
  current: 
    name: Current
    filters: 
      - lambda: if (x < 0.060) return 0.0; else return x;
  voltage: 
    name: Voltage
  power: 
    name: Power
    id: power_sensor
    filters: 
      - lambda: if (x < 3.0) return 0.0; else return x;
  energy: 
    name: Energy
    id: energy
    unit_of_measurement: kWh
    filters: 
      - multiply: 0.001
    on_value: 
      then: 
        - lambda: |-
            static float previous_energy_value = 0.0;
            float current_energy_value = id(energy).state;
            id(total_energy) += current_energy_value - previous_energy_value;
            previous_energy_value = current_energy_value;
```

### Additional information

_No response_

DAVe3283:

#6095 was a breaking change that converted CSE7766 to non-polling, but in the process introduced bugs around how energy was handled, and more subtle bugs around the time period that measurements were taken over.

The subtle change in the timing of measurements made certain downstream calculations (specifically power factor and apparent power) impossible to do accurately.

And since the CSE7766 updates at almost 10 Hz, any sane use of it will require a filter to throttle that down to something Home Assistant won't be unhappy about, which means the benefits of moving away from averaging over an arbitrary period were entirely theoretical.

As such, this reverts #6095 to restore the old behavior that could be used for downstream calculations.
I retained the improvements made in #6180 though, so it is not merely rolling the code back to how it was.

I am marking this as a breaking change because #6095 was a breaking change and undoing that is also a breaking change, but for people coming from ESPHome prior to 2024.2.0 this is not actually a breaking change.

## Types of changes

- [ ] Other

- fixes esphome/issues#5501
- fixes esphome/issues#5504

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3622

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
external_components:
  - source: github://pr#6265
    components: [ cse7766 ]

sensor:
  - platform: cse7766
    update_interval: 30s
    current:
      name: Current
      id: current
    voltage:
      name: Voltage
      id: voltage
    power:
      name: Power
      id: power
      on_value:
        then:
          - component.update: apparent_power
          - component.update: pf
    energy:
      name: Energy
```

## Checklist:

### DIFF
--- a/esphome/components/cse7766/cse7766.cpp
+++ b/esphome/components/cse7766/cse7766.cpp
@@ -40,8 +40,8 @@ void CSE7766Component::loop() {
 float CSE7766Component::get_setup_priority() const { return setup_priority::DATA; }
 
 bool CSE7766Component::check_byte_() {
-  uint8_t index = this->raw_data_index_;
-  uint8_t byte = this->raw_data_[index];
+  const uint8_t index = this->raw_data_index_;
+  const uint8_t byte = this->raw_data_[index];
   if (index == 0) {
     return !((byte != 0x55) && ((byte & 0xF0) != 0xF0) && (byte != 0xAA));
   }
@@ -82,7 +82,7 @@ void CSE7766Component::parse_data_() {
 #endif
 
   // Parse header
-  uint8_t header1 = this->raw_data_[0];
+  const uint8_t header1 = this->raw_data_[0];
 
   if (header1 == 0xAA) {
     ESP_LOGE(TAG, "CSE7766 not calibrated!");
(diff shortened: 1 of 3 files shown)


z3liff:
FWIW I agree with reasoning behind this, and have noticed some of the same issues on my own setup at home.  Reverting the change but keeping the improvements from #6180 seems like the best overall route.

dmitkam:
Thank you. Works.

NearlCrews:
This fixed the issue for me too. I'm hoping it gets merged soon.

puuu:
Also, I am not happy with the breaking changes, but overall the changes in [#6095](https://github.com/esphome/esphome/pull/6095) are the correct way, similar to other components in esphome. CSE7766 is not a polling sensor and should not be treated as such.

If you are interested in power factor and apparent power, why not add the calculation in `parse_data_()` where you can check for `have_power`, `have_current` and `have_voltage` and calculate power factor and apparent power? Then provide the data as additional sensors of the cse7766 device in the same way as power, voltage and current in the current non-polling version.

DAVe3283:
Adding power factor and apparent power directly to the component is a good idea, and I may open a PR to do that when I get the chance. Though I think it will have the same trouble dealing with power and current not necessarily being published simultaneously under this new architecture causing missed or non-representative readings.

But even with the fix to the `throttle_average` filter (#6275) this component doesn't work correctly publishing values on variable intervals. The main remaining issue for me is energy reports Unknown with either a `throttle` or `heartbeat` filter until some energy is actually consumed, when it should report 0 after startup. And running without a filter leads to an insane number of published states back to Home Assistant.

puuu:
The energy calculation is and was broken anyway. See my pull request #6286.

> Though I think it will have the same trouble dealing with power and current not necessarily being published simultaneously under this new architecture causing missed or non-representative readings.

Yes, you must check that all three values (power, voltage, current) are sent together, otherwise the calculation is not correct. Using averages as before does not seem to be correct either.

DAVe3283:
Replaced / handled by #6286 and #6292

PR #6273, opened 2024-02-23
Injecting last 3 bytes of MAC to MQTT topic + introducing <topic_prefix> placeholder in mqtt topics

### LINKED ISSUE #2504: Adding the last 3 bytes of the MAC address to the device name in the MQTT topic_prefix.

**Additional context**

After adding the last 3 bytes of the MAC address to the device name, the device host name is formed from the device name and the last 3 bytes of the MAC address.
However, in MQTT, topic_prefix still consists only of the device name by default.

Please add the possibility that if name_add_mac_suffix is used, then the last 3 bytes of the MAC address should also be added to the topic_prefix MQTT.

latonita:

PROBLEM 1
When making many similar sensors it might be cumbersome to make individual configs for each.
Existing `name_add_mac_suffix` option only affects device name, but doesn't affect mqtt topics.

This change adds last 3 bytes of mac address to `topic_prefix` in case it wasn't customized by user.

Since `topic_prefix` is now dynamic some code added to inject into base topics:
   - last will topic
   - birth message topic
   - shutdown message topic
   - log message topic
   - availability topic
 

PROBLEM 2
Subscription via `on_message` only works with fully hardcoded topic. Impossible to subscribe to messages sent directly to our device in case of adding mac suffix or using `topic_prefix`

Solution proposal: add a placeholder <topic_prefix> so user can even use it to customize topics
```
esphome:
  name: esp-sensor
  name_add_mac_suffix: true
mqtt:
#  topic_prefix: custom_prefix
  on_json_message:
    topic: my/custom/topic            # fully hardcoded, subscribing to "my/custom/topic"
  on_json_message:
    topic: <topic_prefix>/some/topic  # <topic_prefix> gets replaced either by "esp-sensor-DDEEFF" 
                                      # or by user-customized topic_prefix "custom_prefix"
 ```

PROBLEM 3
Now users want to subscribe to a topic in lambda using node's dynamic name. 
`subscribe` methods updated to check and replace <topic_prefix> placeholder

PROBLEM 4
Users want to publish messages not to a hardcoded topic, but also use it's own topic_prefix which might be dynamic
`publish` methods updated to check and replace <topic_prefix> placeholder

Here is an example of a device logs (see config below) + receiving a message sent to topic `esp-sensor-551495/on_message_topic`
```
[C][mqtt:169]: MQTT:
[C][mqtt:170]:   Server Address:
[C][mqtt:172]:   Username: 
[C][mqtt:173]:   Client ID: 'esp-sensor-551495-48e729551495'
[C][mqtt:178]:   Topic Prefix: 'esp-sensor-551495'
[C][mqtt:180]:   Log Topic: 'esp-sensor-551495/debug'
[C][mqtt:183]:   Availability: 'esp-sensor-551495/my_availability'
[C][api:139]: API Server:
[C][api:140]:   Address: esp-sensor.local:6053
[C][api:144]:   Using noise encryption: NO
[C][mqtt:710]: MQTT Message Trigger:
[C][mqtt:711]:   Topic: 'esp-sensor-551495/on_message_topic'
[C][mqtt:712]:   QoS: 0
[C][mqtt:728]: MQTT JSON Message Trigger:
[C][mqtt:729]:   Topic: 'esp-sensor-551495/on_json_message_topic'
[C][mqtt:730]:   QoS: 0
[D][main:248]: Received on_message_topic
[D][main:266]: Received on_json_message_topic

```

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2504

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3634

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esphome:
  name: esp-sensor
  name_add_mac_suffix: true

mqtt:
  broker: !secret mqtt_broker
  discovery: False
  #topic_prefix: custom_prefix
  birth_message:
    topic: <topic_prefix>/my_availability
    payload: online
  will_message:
    topic: <topic_prefix>/my_availability
    payload: offline
  on_message:
     - topic: <topic_prefix>/on_message_topic
       then:
        - logger.log: Received on_message_topic
        - mqtt.publish:
            topic: <topic_prefix>/on_message_topic_confirmation
            payload: "Received on_message_topic"
        - mqtt.publish:
            topic: <topic_prefix>/on_json_message_topic
            payload: "{ \"Received\": \"on_message_topic\" }"
  on_json_message:
     - topic: <topic_prefix>/on_json_message_topic
       then:
        - logger.log: Received on_json_message_topic
```

## Checklist:


arthurkomatsu:
Nice change, I actually need this on one of my projects, what I ended up doing was adding:
```yaml
esphome:
  on_boot:
    then:
      - lambda: |-
          // Get the name of this device with its mac.
          std::string name_with_mac = App.get_name() + "-" + get_mac_address();

          // Set ap again.
          wifi::WiFiAP wifi_wifiap = wifi::WiFiAP();
          wifi_wifiap.set_ssid(name_with_mac);
          wifi_wifiap.set_password(WIFI_AP_PASSWORD);
          wifi_wificomponent->set_ap(wifi_wifiap);

          // Mqtt prefix.
          mqtt_client->set_client_id(name_with_mac.c_str());
          mqtt_client->set_topic_prefix(name_with_mac);
          mqtt_client->set_birth_message(mqtt::MQTTMessage{
              .topic = name_with_mac + "/status",
              .payload = "online",
              .qos = 0,
              .retain = true,
          });
          mqtt_client->set_last_will(mqtt::MQTTMessage{
              .topic = name_with_mac + "/status",
              .payload = "offline",
              .qos = 0,
              .retain = true,
          });
          mqtt_client->set_shutdown_message(mqtt::MQTTMessage{
              .topic = name_with_mac + "/status",
              .payload = "offline",
              .qos = 0,
              .retain = true,
          });
          mqtt_client->set_log_message_template(mqtt::MQTTMessage{
              .topic = name_with_mac + "/debug",
              .payload = "",
              .qos = 0,
              .retain = true,
          });
```

Also, for consistency's sake, consider using the full Mac address (like the client ID does) instead of the last 3 bytes (the whole `get_mac_address()` function).

latonita:
> Also, for consistency's sake, consider using the full Mac address (like the client ID does) instead of the last 3 bytes (the whole `get_mac_address()` function).

This was done to be unified with existing functionality. And I know it was also done from the early days of esp8266 as well.
So didn't want to change this to full mac.
Do you think 3 bytes of mac will not be enough for you?

arthurkomatsu:
> > Also, for consistency's sake, consider using the full Mac address (like the client ID does) instead of the last 3 bytes (the whole `get_mac_address()` function).
> 
> This was done to be unified with existing functionality. And I know it was also done from the early days of esp8266 as well. So didn't want to change this to full mac. Do you think 3 bytes of mac will not be enough for you?

I see your point, you are thinking about the old `ESP.getChipId` API which gives an integer of the last 3 bytes of its mac address. I would rather work with the full mac address for the following reasons:

- It is easier to identify a device by looking at its full mac address rather than its last 3 bytes. Usually, devices have a sticker with them with their full mac addresses.
- For consistency sake, the client ID uses the full mac address (see [here](https://github.com/esphome/esphome/blob/19022ace12595295fadcc3fbb36c69a8d5c846f4/esphome/components/mqtt/mqtt_client.cpp#L32)), so if you happen to implement the MQTT backend server, you can easily get the prefix by looking at its client ID on the connection routine. Also, maybe you can simplify the code by using the already mounted variable `this->credentials_.client_id`?
- Collisions can happen (same ID for different devices). The number of different Chip IDs are `2^24=16777216`, which is way less than the actual number of manufactured ESPs. For a quick example, if you get 5000 random ESP devices, the probability of collision (by the [Birthday Problem](https://en.wikipedia.org/wiki/Birthday_problem#:~:text=In%20probability%20theory%2C%20the%20birthday,that%20probability%20to%20exceed%2050%25.)) is ~52% (see [here](https://www.wolframalpha.com/input?i=1-%2816777216%21%2F%2816777216-5000%29%21%29%2F16777216%5E5000))

jesserockz:
> When making many similar sensors it might be cumbersome to make individual configs for each.
Existing name_add_mac_suffix option only affects device name, but doesn't affect mqtt topics.

End users should not be using `name_add_mac_suffix`. You should have a YAML file per every device and use `packages` to cut down on duplication of config.

If you add the mac address yourself to the `esphome`->`name` then it will be part of the `topic_prefix` automatically.

latonita:
Closing PR as 'against the design idea` for now.

PR #6276, opened 2024-02-24
Allow to specify global build directory

werwolfby:
> [!NOTE]
> Reopen after rebase of #5387

Default implementation use `.esphome/build` folder to store build artifacts. This folder can be relatively big. I run esphome container in kubernetes and all `/config` folder is mounted as remote storage. This is fine for small pregenerated json and `idedata`. But this is not best option for build temporary files. Instead I can use local storage for `.esphome/build` folders.

This PR allows to specify `ESPHOME_BUILD_PATH` environment variable that will be used as default build folder path. This build path still can be overridden by individual components over `build_path` option (as described in [doc](https://esphome.io/components/esphome.html)), but introduce default folder that can be specified.

## Types of changes

- [ ] Other

Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable): https://github.com/esphome/esphome-docs/pull/3183

## Checklist:

### DIFF
--- a/esphome/core/config.py
+++ b/esphome/core/config.py
@@ -190,11 +190,8 @@ def preload_core_config(config, result):
     CORE.data[KEY_CORE] = {}
 
     if CONF_BUILD_PATH not in conf:
-        if get_str_env("ESPHOME_BUILD_PATH", None) is not None:
-            conf[CONF_BUILD_PATH] = get_str_env("ESPHOME_BUILD_PATH", None)
-        # If the env var is not set, use the default
-        else:
-            conf[CONF_BUILD_PATH] = f"build/{CORE.name}"
+        build_path = get_str_env("ESPHOME_BUILD_PATH", "build")
+        conf[CONF_BUILD_PATH] = os.path.join(build_path, CORE.name)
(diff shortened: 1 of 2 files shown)


esphome/core/config.py
```diff
@@ -190,7 +190,11 @@ def preload_core_config(config, result):
     CORE.data[KEY_CORE] = {}
 
     if CONF_BUILD_PATH not in conf:
-        conf[CONF_BUILD_PATH] = f"build/{CORE.name}"
+        if get_str_env("ESPHOME_BUILD_PATH", None) is not None:
+            conf[CONF_BUILD_PATH] = get_str_env("ESPHOME_BUILD_PATH", None)
```
jesserockz: This will put all build in the same directory. It will still need to be suffixed with the `CORE.name` I believe

jesserockz (changes requested):

esphome/core/config.py
werwolfby: Yep, I’ve realized it just now, will fix soon

jesserockz (approved):

PR #6286, opened 2024-02-26
CSE7766: Fix energy calculation

puuu:
Adding small float values to a large float value is not a good idea. Due to the lack of accuracy, the small values may be completely ignored.

This commit uses an unsigned integer to sum the cf pulses and calculate the total energy from the total number of cf pulses.

After some time the accumulation of energy does not fit to measured power.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: cse7766
    current:
      name: Current
      filters:
        - throttle_average: 30s
    voltage:
      name: Voltage
      filters:
        - filter_out: 0.0
        - throttle_average: 60s
    power:
      name: Power
      filters:
        - throttle_average: 30s
    energy:
      name: Energy
      filters:
        - throttle: 30s

```

## Checklist:

### DIFF
--- a/esphome/components/cse7766/cse7766.cpp
+++ b/esphome/components/cse7766/cse7766.cpp
@@ -118,7 +118,7 @@ void CSE7766Component::parse_data_() {
   uint32_t power_coeff = this->get_24_bit_uint_(14);
   uint32_t power_cycle = this->get_24_bit_uint_(17);
   uint8_t adj = this->raw_data_[20];
-  uint32_t cf_pulses = (this->raw_data_[21] << 8) + this->raw_data_[22];
+  uint16_t cf_pulses = (this->raw_data_[21] << 8) + this->raw_data_[22];
 
   bool have_power = adj & 0x10;
   bool have_current = adj & 0x20;
@@ -132,8 +132,24 @@ void CSE7766Component::parse_data_() {
     }
   }
 
+  float energy = 0.0;
+  if (this->energy_sensor_ != nullptr) {
+    if (this->cf_pulses_last_ == 0 && !this->energy_sensor_->has_state()) {
+      this->cf_pulses_last_ = cf_pulses;
+    }
+    uint16_t cf_diff = cf_pulses - this->cf_pulses_last_;
+    this->cf_pulses_total_ += cf_diff;
+    this->cf_pulses_last_ = cf_pulses;
+    // Calculate CF pulses into total energy only if we have Power coefficient to multiply by
+    if (have_power) {
+      energy = this->cf_pulses_total_ * float(power_coeff) / 1000000.0f / 3600.0f;
+      this->energy_sensor_->publish_state(energy);
+    } else if (!this->energy_sensor_->has_state()) {
+      this->energy_sensor_->publish_state(0);
+    }
+  }
+
   float power = 0.0f;
-  float energy = 0.0f;
   if (power_cycle_exceeds_range) {
     // Datasheet: power cycle exceeding range means active power is 0
     if (this->power_sensor_ != nullptr) {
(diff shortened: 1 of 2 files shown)


esphome/components/cse7766/cse7766.cpp
```diff
@@ -132,8 +132,24 @@ void CSE7766Component::parse_data_() {
...
+  if (this->energy_sensor_ != nullptr) {
+    if (this->cf_pulses_last_ == 0 && !this->energy_sensor_->has_state()) {
+      this->cf_pulses_last_ = cf_pulses;
+    }
+    uint16_t cf_diff = cf_pulses - this->cf_pulses_last_;
+    this->cf_pulses_total_ += cf_diff;
+    this->cf_pulses_last_ = cf_pulses;
+    // Calculate CF pulses into total energy only if we have Power coefficient to multiply by
+    if (have_power) {
+      energy = this->cf_pulses_total_ * float(power_coeff) / 1000000.0f / 3600.0f;
```
DAVe3283: Since the value of a cf_pulse depends on the power_coeff for that measurement cycle, won't this potentially allow total energy to go backwards if power_coeff goes down for one cycle vs. the previous?

You would also need to only accumulate cf_pulses inside the `if (have_power)` block to avoid adding bogus pulses when there is no valid power_coeff.

I think you will have to store the total energy not in terms of pulses (since a pulse appears to be a variable amount of energy), but in fixed-point decimal in Watt-hours.

puuu: A
According to the CSE7766 datasheet, the pulses at the CF pin are proportional to the total energy. The value of cf_pulse is a count of the cf output pin. The energy calculation in the datasheet is based on the "power coefficient" or a theoretical value. To me, the datasheet reads like the "power coefficient" is a kind of calibration value and is always constant for a specific chip. In fact, I have not seen any changes in the "power coefficient".

What made you think it could change?

The datasheet also says nothing about pogus pulses, which should be ignored. How should I do it if I only use the cf pin?

DAVe3283: The cse7766 doesn't have the clearest datasheet IMO, but you are right, I don't see anything stating it can change at runtime, and I haven't noticed it change either.

Since power_coeff never changes, and appears to be populated with the calibrated value even when it doesn't have a power reading (on my unit at least), then why not remove the `if (have_power)` and just use `power_coeff` to publish any time? Simplify things a bit.

esphome/components/cse7766/cse7766.h
```diff
@@ -30,8 +30,8 @@ class CSE7766Component : public Component, public uart::UARTDevice {
   sensor::Sensor *current_sensor_{nullptr};
   sensor::Sensor *power_sensor_{nullptr};
   sensor::Sensor *energy_sensor_{nullptr};
-  float energy_total_{0.0f};
-  uint32_t cf_pulses_last_{0};
+  uint32_t cf_pulses_total_{0};
```
DAVe3283, line 33: Not sure how real of a concern this is, but a uint32_t could theoretically overflow.

The max value (0xFFFFFFFF = 4294967295) with the power_coeff I see on my unit (0x4AA178 = 4891000) would overflow after 5835.190 kWh. That is a TON of power, but if it is measuring a high-load (1 kW) device, that would overflow in 240 days.

In practice, most (all?) people will update their device or lose power before that happens, so might not be worth handling that theoretical extreme?

esphome/components/cse7766/cse7766.cpp
```diff
@@ -132,8 +132,24 @@ void CSE7766Component::parse_data_() {
...
+      this->energy_sensor_->publish_state(energy);
+    } else if (!this->energy_sensor_->has_state()) {
+      this->energy_sensor_->publish_state(0);
+    }
```
DAVe3283: Since it appears the power_coeff and cf_pulses values are valid every time (even when power is 0), it would simplify things to publish every time. A filter will be needed to throttle energy regardless.

```suggestion
    energy = this->cf_pulses_total_ * float(power_coeff) / 1000000.0f / 3600.0f;
    this->energy_sensor_->publish_state(energy);
```

Edit: this works great on my Sonoff and Athom outlets.

puuu: Yes you are right, I did not check `power_coeff` for `!have_power`. I will check this on my devices first.

esphome/components/cse7766/cse7766.h
puuu: Yes, the `uint32_t` value will overflow at about 1 month at maximum load.

Home Assistant (long term statistics and utility meter) will correctly deal with energy resets to small values or 0 (as will happen with an integer overflow). We may lose a few pulses between the last value and the maximum of the integer value. Since the update interval is so fast, this error is IMHO negligible compared to the ±1% error of the CSE7766.

jesserockz (approved):
LGTM. Thanks

PR #6291, opened 2024-02-27
nau7802: new component

cujomalainey:

Adds support for NAU7802 24bit ADC. See [adafruit board](https://www.adafruit.com/product/4538).

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4092

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esphome:
  name: test
  friendly_name: tester

esp32:
  board: featheresp32

logger:

i2c:
  - sda: SDA
    scl: SCL
    scan: true
    id: i2c_bus

sensor:
  - platform: nau7802
    id: weight
    i2c_id: i2c_bus
```

## Checklist:

### DIFF
--- a/esphome/components/nau7802/nau7802.cpp
+++ b/esphome/components/nau7802/nau7802.cpp
@@ -1,0 +1,323 @@
+#include "nau7802.h"
+#include "esphome/core/log.h"
+#include "esphome/core/hal.h"
+
+namespace esphome {
+namespace nau7802 {
+
+static const char *const TAG = "nau7802";
+
+// Only define what we need
+
+static const uint8_t READ_BIT = 0x01;
+
+static const uint8_t PU_CTRL_REG = 0x00;
+static const uint8_t PU_CTRL_REGISTER_RESET = 0x01;
+static const uint8_t PU_CTRL_POWERUP_DIGITAL = 0x02;
+static const uint8_t PU_CTRL_POWERUP_ANALOG = 0x04;
+static const uint8_t PU_CTRL_POWERUP_READY = 0x08;
+static const uint8_t PU_CTRL_CYCLE_START = 0x10;
+static const uint8_t PU_CTRL_CYCLE_READY = 0x20;
+static const uint8_t PU_CTRL_AVDD_EXTERNAL = 0x80;
+
+static const uint8_t CTRL1_REG = 0x01;
+static const uint8_t CTRL1_LDO_SHIFT = 3;
+static const uint8_t CTRL1_LDO_MASK = (0x7 << CTRL1_LDO_SHIFT);
+static const uint8_t CTRL1_GAIN_MASK = 0x7;
+
+static const uint8_t CTRL2_REG = 0x02;
+static const uint8_t CTRL2_CRS_SHIFT = 4;
+static const uint8_t CTRL2_CRS_MASK = (0x7 << CTRL2_CRS_SHIFT);
+static const uint8_t CTRL2_CALS = 0x04;
+static const uint8_t CTRL2_CAL_ERR = 0x08;
+static const uint8_t CTRL2_GAIN_CALIBRATION = 0x03;
+static const uint8_t CTRL2_CONFIG_MASK = 0xF0;
+
+static const uint8_t OCAL1_B2_REG = 0x03;
+static const uint8_t GCAL1_B3_REG = 0x06;
+static const uint8_t GCAL1_FRACTIONAL = 23;
+
+// only need the first data register for sequential read method
+static const uint8_t ADCO_B2_REG = 0x12;
+
+static const uint8_t ADC_REG = 0x15;
+static const uint8_t ADC_CHPS_DISABLE = 0x30;
+
+static const uint8_t PGA_REG = 0x1B;
+static const uint8_t PGA_LDOMODE_ESR = 0x40;
+
+static const uint8_t POWER_REG = 0x1C;
+static const uint8_t POWER_PGA_CAP_EN = 0x80;
+
+static const uint8_t DEVICE_REV = 0x1F;
+
+void NAU7802Sensor::setup() {
+  i2c::I2CRegister pu_ctrl = this->reg(PU_CTRL_REG);
+  ESP_LOGCONFIG(TAG, "Setting up NAU7802 '%s'...", this->name_.c_str());
+  uint8_t rev;
+
+  if (this->read_register(DEVICE_REV | READ_BIT, &rev, 1)) {
+    ESP_LOGE(TAG, "Failed I2C read during setup()");
+    this->mark_failed();
+    return;
+  }
+  ESP_LOGI(TAG, "Setting up NAU7802 Rev %d", rev);
+
+  // reset
+  pu_ctrl |= PU_CTRL_REGISTER_RESET;
+  delay(10);
+  pu_ctrl &= ~PU_CTRL_REGISTER_RESET;
+
+  // power up digital hw
+  pu_ctrl |= PU_CTRL_POWERUP_DIGITAL;
+
+  delay(1);
+  if (!(pu_ctrl.get() & PU_CTRL_POWERUP_READY)) {
+    ESP_LOGE(TAG, "Failed to reset sensor during setup()");
+    this->mark_failed();
+    return;
+  }
+
+  uint32_t gcal = (uint32_t) (round(this->gain_calibration_ * (1 << GCAL1_FRACTIONAL)));
+  this->write_value_(OCAL1_B2_REG, 3, this->offset_calibration_);
+  this->write_value_(GCAL1_B3_REG, 4, gcal);
+
+  // turn on AFE
+  pu_ctrl |= PU_CTRL_POWERUP_ANALOG;
+  auto f = std::bind(&NAU7802Sensor::complete_setup_, this);
+  this->set_timeout(600, f);
+}
+
+void NAU7802Sensor::complete_setup_() {
+  i2c::I2CRegister pu_ctrl = this->reg(PU_CTRL_REG);
+  i2c::I2CRegister ctrl1 = this->reg(CTRL1_REG);
+  i2c::I2CRegister ctrl2 = this->reg(CTRL2_REG);
+  pu_ctrl |= PU_CTRL_CYCLE_START;
+
+  // set gain
+  ctrl1 &= ~CTRL1_GAIN_MASK;
+  ctrl1 |= this->gain_;
+
+  // enable internal LDO
+  if (this->ldo_ != NAU7802_LDO_EXTERNAL) {
+    pu_ctrl |= PU_CTRL_AVDD_EXTERNAL;
+    ctrl1 &= ~CTRL1_LDO_MASK;
+    ctrl1 |= this->ldo_ << CTRL1_LDO_SHIFT;
+  }
+
+  // set sps
+  ctrl2 &= ~CTRL2_CRS_MASK;
+  ctrl2 |= this->sps_ << CTRL2_CRS_SHIFT;
+
+  // disable ADC chopper clock
+  i2c::I2CRegister adc_reg = this->reg(ADC_REG);
+  adc_reg |= ADC_CHPS_DISABLE;
+
+  // use low ESR caps
+  i2c::I2CRegister pga_reg = this->reg(PGA_REG);
+  pga_reg &= ~PGA_LDOMODE_ESR;
+
+  // PGA stabilizer cap on output
+  i2c::I2CRegister pwr_reg = this->reg(POWER_REG);
+  pwr_reg |= POWER_PGA_CAP_EN;
+
+  this->setup_complete_ = true;
+}
+
+void NAU7802Sensor::dump_config() {
+  LOG_SENSOR("", "NAU7802", this);
+  LOG_I2C_DEVICE(this);
+
+  if (this->is_failed()) {
+    ESP_LOGE(TAG, "Communication with NAU7802 failed earlier, during setup");
+    return;
+  }
+  // Note these may differ from the values on the device if calbration has been run
+  ESP_LOGCONFIG(TAG, "  Offset Calibration: %s", to_string(this->offset_calibration_).c_str());
+  ESP_LOGCONFIG(TAG, "  Gain Calibration: %f", this->gain_calibration_);
+
+  std::string voltage = "unknown";
+  switch (this->ldo_) {
+    case NAU7802_LDO_2V4:
+      voltage = "2.4V";
+      break;
+    case NAU7802_LDO_2V7:
+      voltage = "2.7V";
+      break;
+    case NAU7802_LDO_3V0:
+      voltage = "3.0V";
+      break;
+    case NAU7802_LDO_3V3:
+      voltage = "3.3V";
+      break;
+    case NAU7802_LDO_3V6:
+      voltage = "3.6V";
+      break;
+    case NAU7802_LDO_3V9:
+      voltage = "3.9V";
+      break;
+    case NAU7802_LDO_4V2:
+      voltage = "4.2V";
+      break;
+    case NAU7802_LDO_4V5:
+      voltage = "4.5V";
+      break;
+    case NAU7802_LDO_EXTERNAL:
+      voltage = "External";
+      break;
+  }
+  ESP_LOGCONFIG(TAG, "  LDO Voltage: %s", voltage.c_str());
+  int gain = 0;
+  switch (this->gain_) {
+    case NAU7802_GAIN_128:
+      gain = 128;
+      break;
+    case NAU7802_GAIN_64:
+      gain = 64;
+      break;
+    case NAU7802_GAIN_32:
+      gain = 32;
+      break;
+    case NAU7802_GAIN_16:
+      gain = 16;
+      break;
+    case NAU7802_GAIN_8:
+      gain = 8;
+      break;
+    case NAU7802_GAIN_4:
+      gain = 4;
+      break;
+    case NAU7802_GAIN_2:
+      gain = 2;
+      break;
+    case NAU7802_GAIN_1:
+      gain = 1;
+      break;
+  }
+  ESP_LOGCONFIG(TAG, "  Gain: %dx", gain);
+  int sps = 0;
+  switch (this->sps_) {
+    case NAU7802_SPS_320:
+      sps = 320;
+      break;
+    case NAU7802_SPS_80:
+      sps = 80;
+      break;
+    case NAU7802_SPS_40:
+      sps = 40;
+      break;
+    case NAU7802_SPS_20:
+      sps = 20;
+      break;
+    case NAU7802_SPS_10:
+      sps = 10;
+      break;
+  }
+  ESP_LOGCONFIG(TAG, "  Samples Per Second: %d", sps);
+  LOG_UPDATE_INTERVAL(this);
+}
+
+void NAU7802Sensor::write_value_(uint8_t start_reg, size_t size, int32_t value) {
+  uint8_t data[4];
+  for (int i = 0; i < size; i++) {
+    data[i] = 0xFF & (value >> (size - 1 - i) * 8);
+  }
+  this->write_register(start_reg, data, size);
+}
+
+int32_t NAU7802Sensor::read_value_(uint8_t start_reg, size_t size) {
+  uint8_t data[4];
+  this->read_register(start_reg, data, size);
+  int32_t result = 0;
+  for (int i = 0; i < size; i++) {
+    result |= data[i] << (size - 1 - i) * 8;
+  }
+  // extend sign bit
+  if (result & 0x800000 && size == 3) {
+    result |= 0xFF000000;
+  }
+  return result;
+}
+
+bool NAU7802Sensor::calibrate_(enum NAU7802CalibrationModes mode) {
+  // check if already calbrating
+  if (this->state_ != CalibrationState::INACTIVE) {
+    ESP_LOGW(TAG, "Calibration already in progress");
+    return false;
+  }
+
+  this->state_ = mode == NAU7802_CALIBRATE_GAIN ? CalibrationState::GAIN : CalibrationState::OFFSET;
+
+  i2c::I2CRegister ctrl2 = this->reg(CTRL2_REG);
+  // clear calibraye registers
+  ctrl2 &= CTRL2_CONFIG_MASK;
+  // Calibrate
+  ctrl2 |= mode;
+  ctrl2 |= CTRL2_CALS;
+  return true;
+}
+
+void NAU7802Sensor::set_calibration_failure_(bool failed) {
+  switch (this->state_) {
+    case CalibrationState::GAIN:
+      this->gain_calibration_failed_ = failed;
+      break;
+    case CalibrationState::OFFSET:
+      this->offset_calibration_failed_ = failed;
+      break;
+    case CalibrationState::INACTIVE:
+      // shouldn't happen
+      break;
+  }
+}
+
+void NAU7802Sensor::loop() {
+  i2c::I2CRegister ctrl2 = this->reg(CTRL2_REG);
+
+  if (this->state_ != CalibrationState::INACTIVE && !(ctrl2.get() & CTRL2_CALS)) {
+    if (ctrl2.get() & CTRL2_CAL_ERR) {
+      this->set_calibration_failure_(true);
+      this->state_ = CalibrationState::INACTIVE;
+      ESP_LOGE(TAG, "Failed to calibrate sensor");
+      this->status_set_error("Calibration Failed");
+      return;
+    }
+
+    this->set_calibration_failure_(false);
+    this->state_ = CalibrationState::INACTIVE;
+
+    if (!this->offset_calibration_failed_ && !this->gain_calibration_failed_)
+      this->status_clear_error();
+
+    int32_t ocal = this->read_value_(OCAL1_B2_REG, 3);
+    ESP_LOGI(TAG, "New Offset: %s", to_string(ocal).c_str());
+    uint32_t gcal = this->read_value_(GCAL1_B3_REG, 4);
+    float gcal_f = ((float) gcal / (float) (1 << GCAL1_FRACTIONAL));
+    ESP_LOGI(TAG, "New Gain: %f", gcal_f);
+  }
+}
+
+float NAU7802Sensor::get_setup_priority() const { return setup_priority::DATA; }
+
+void NAU7802Sensor::update() {
+  if (!this->is_data_ready_()) {
+    ESP_LOGW(TAG, "No measurements ready!");
+    this->status_set_warning();
+    return;
+  }
+
+  this->status_clear_warning();
+
+  // Get the most recent sample to publish
+  int32_t result = this->read_value_(ADCO_B2_REG, 3);
+
+  ESP_LOGD(TAG, "'%s': Got value %" PRId32, this->name_.c_str(), result);
+  this->publish_state(result);
+}
+
+bool NAU7802Sensor::is_data_ready_() { return this->reg(PU_CTRL_REG).get() & PU_CTRL_CYCLE_READY; }
+
+bool NAU7802Sensor::can_proceed() { return this->setup_complete_; }
+
+}  // namespace nau7802
+}  // namespace esphome
--- a/esphome/components/nau7802/nau7802.h
+++ b/esphome/components/nau7802/nau7802.h
@@ -50,14 +50,14 @@ enum NAU7802CalibrationModes {
 
 class NAU7802Sensor : public sensor::Sensor, public PollingComponent, public i2c::I2CDevice {
  public:
-  void set_samples_per_second(NAU7802SPS sps) { sps_ = sps; }
-  void set_ldo_voltage(NAU7802LDO ldo) { ldo_ = ldo; }
-  void set_gain(NAU7802Gain gain) { gain_ = gain; }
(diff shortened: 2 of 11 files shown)


BrunoB81HK:
@cujomalainey Is this ready for integration? It would be nice to be able to use this component.

cujomalainey:
> @cujomalainey Is this ready for integration? It would be nice to be able to use this component.

@BrunoB81HK still a WIP. Just been working on this in my limited spare time. If you can offer help testing it out that would be appreciated. I recently got the calibration code migrated to actions but haven't had a chance to test it. Once that is done and I have completed the docs it will be ready for review.

BrunoB81HK:
That's great! Thanks for the great work! I'm not that familiar with the inner working of ESPHome, but I can't guarantee anything, but I'll try to find some time to help you here.

gnsbglora:
Hello, thank you for developing this integration.  I gave it a try. I used a standard ESP32 38 pin, a sparkfun quick nau7802 and a Adafruit nau7802. The sketch compiles but it crashes/loops in the setup. The ESP32 can see the nau7802 with its address without the integration.

```
esphome:
  name: esp32-test
  friendly_name: ESP32-Test

preferences:
  flash_write_interval: 1440min

esp32:
  board: esp32dev
  framework:
    type: arduino

logger:

api:
  encryption:
    key: ""

ota:
  - platform: esphome
    password: "" 

wifi:
  ssid: !secret wifi_ssid
  
i2c:
  - id: bus_a
    sda: 21
    scl: 22
    scan: True

external_components:
  source: github://pr#6291
  components: nau7802    

sensor:
  - platform: nau7802
    name: "Scale"
    id: scale
    i2c_id: bus_a
    unit_of_measurement: "kg" 

```
```
[11:02:01]ets Jun  8 2016 00:22:57
[11:02:01]
[11:02:01]rst:0x8 (TG1WDT_SYS_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
[11:02:01]configsip: 0, SPIWP:0xee
[11:02:01]clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[11:02:01]mode:DIO, clock div:2
[11:02:01]load:0x3fff0030,len:1184
[11:02:01]load:0x40078000,len:13132
[11:02:01]load:0x40080400,len:3036
[11:02:01]entry 0x400805e4
[11:02:02][I][logger:156]: Log initialized
[11:02:02][C][safe_mode:079]: There have been 3 suspected unsuccessful boot attempts
[11:02:02][D][esp32.preferences:114]: Saving 1 preferences to flash...
[11:02:02][D][esp32.preferences:143]: Saving 1 preferences to flash: 0 cached, 1 written, 0 failed
[11:02:02][I][app:029]: Running through setup()...
[11:02:02][I][i2c.arduino:218]: Performing I2C bus recovery
[11:02:02][C][nau7802:057]: Setting up NAU7802 'Scale'...
[11:02:02][I][nau7802:065]: Setting up NAU7802 Rev 15
```

cujomalainey:
> Hello, thank you for developing this integration. I gave it a try. I used a standard ESP32 38 pin, a sparkfun quick nau7802 and a Adafruit nau7802. The sketch compiles but it crashes/loops in the setup. The ESP32 can see the nau7802 with its address without the integration.
> 
> ```
> esphome:
>   name: esp32-test
>   friendly_name: ESP32-Test
> 
> preferences:
>   flash_write_interval: 1440min
> 
> esp32:
>   board: esp32dev
>   framework:
>     type: arduino
> 
> logger:
> 
> api:
>   encryption:
>     key: ""
> 
> ota:
>   - platform: esphome
>     password: "" 
> 
> wifi:
>   ssid: !secret wifi_ssid
>   password: !secret wifi_password
>   
> i2c:
>   - id: bus_a
>     sda: 21
>     scl: 22
>     scan: True
> 
> external_components:
>   source: github://pr#6291
>   components: nau7802    
> 
> sensor:
>   - platform: nau7802
>     name: "Scale"
>     id: scale
>     i2c_id: bus_a
>     unit_of_measurement: "kg" 
> ```
> 
> ```
> [11:02:01]ets Jun  8 2016 00:22:57
> [11:02:01]
> [11:02:01]rst:0x8 (TG1WDT_SYS_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
> [11:02:01]configsip: 0, SPIWP:0xee
> [11:02:01]clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
> [11:02:01]mode:DIO, clock div:2
> [11:02:01]load:0x3fff0030,len:1184
> [11:02:01]load:0x40078000,len:13132
> [11:02:01]load:0x40080400,len:3036
> [11:02:01]entry 0x400805e4
> [11:02:02][I][logger:156]: Log initialized
> [11:02:02][C][safe_mode:079]: There have been 3 suspected unsuccessful boot attempts
> [11:02:02][D][esp32.preferences:114]: Saving 1 preferences to flash...
> [11:02:02][D][esp32.preferences:143]: Saving 1 preferences to flash: 0 cached, 1 written, 0 failed
> [11:02:02][I][app:029]: Running through setup()...
> [11:02:02][I][i2c.arduino:218]: Performing I2C bus recovery
> [11:02:02][C][nau7802:057]: Setting up NAU7802 'Scale'...
> [11:02:02][I][nau7802:065]: Setting up NAU7802 Rev 15
> ```

Thanks, ill take a look at it this week, most of my recent iterations i did while commuting so i was only able to compile test and verify with CI. My guess I mucked up my string handling somewhere.

cujomalainey:
found the bug, I was passing in the args flipped into my write helper. Boots fine now.

cujomalainey:
Fixed some other stuff in this push

- external calibration action was actually running internal calibration
- offset calibration from yaml wasn't being written to the device
- fixed strings in calibration function
- added logging for new calibration values after calibration actions

Note I was not able to test gain calibration as the device reported an error and the datasheet does not explain how one is supposed to execute the workflow so if someone else wants to test that then great, otherwise I am leaving as is.

Please test this out, otherwise I will focus on docs for the next bit.

gnsbglora:
Thank you for the fix. It is working now, but I didn't manage to get the calibration to work. In order to use a negative offset of -820 I had to use "offset_calibration: -8387800". Is that intentional? The offset calibration I calculated didn't make sense. I'm using a small 1Kg strain cell that is quite sensitive. With the esphome sensor filters it is working.

```
sensor:
  - platform: nau7802
    name: "Scale"
    id: scale
    i2c_id: bus_a
    unit_of_measurement: "g"
    update_interval: 1s
    gain: 1
    #gain_calibration: 15.39
    #offset_calibration: -8387800
    #samples_per_second: 320
    filters:
      - offset: -820
      - calibrate_linear:
         method: least_squares
         datapoints:
          - 0.0 -> 1.0
          - 9000.0 -> 539
      - clamp:
          min_value: 0
```

cujomalainey:
> Thank you for the fix. It is working now, but I didn't manage to get the calibration to work. In order to use a negative offset of -820 I had to use "offset_calibration: -8387800". Is that intentional? The offset calibration I calculated didn't make sense. I'm using a small 1Kg strain cell that is quite sensitive. With the esphome sensor filters it is working.
> 
> ```
> sensor:
>   - platform: nau7802
>     name: "Scale"
>     id: scale
>     i2c_id: bus_a
>     unit_of_measurement: "g"
>     update_interval: 1s
>     gain: 1
>     #gain_calibration: 15.39
>     #offset_calibration: -8387800
>     #samples_per_second: 320
>     filters:
>       - offset: -820
>       - calibrate_linear:
>          method: least_squares
>          datapoints:
>           - 0.0 -> 1.0
>           - 9000.0 -> 539
>       - clamp:
>           min_value: 0
> ```

offset calibration should be the value of whatever your sensor is reading with no load. There is the possibility i have a sign extension bug in the offset or value window given the offset is only 3 bytes. I'll compare my values against the adafruit library and see what is going on.

cujomalainey:
> offset calibration should be the value of whatever your sensor is reading with no load. There is the possibility i have a sign extension bug in the offset or value window given the offset is only 3 bytes. I'll compare my values against the adafruit library and see what is going on.

I confirmed I get the same readings out with the adafruit library with the same configuration on my 5kg load cell. My values were also, when uncalibrated, in the negative -100000s. I think this is WAI.

gnsbglora:
I did some experimenting, I'm using this [nau7802 module](https://www.sparkfun.com/products/15242) and this [Strain load cell](https://www.amazon.com/Wishiot-Precision-Miniature-Electronic-Portable/dp/B0C3QHT4TT/ref=asc_df_B0C3QH7V8M/?tag=hyprod-20&linkCode=df0&hvadid=693071814688&hvpos=&hvnetw=g&hvrand=704602480246115246&hvpone=&hvptwo=&hvqmt=&hvdev=c&hvdvcmdl=&hvlocint=&hvlocphy=9012270&hvtargid=pla-2189119876076&mcid=e787aff36fc5353ea847dd1ccaf2af37&gad_source=1&th=1) but I also tried the Adafruit nau7802 and some other [S-shaped load cell](https://www.amazon.com/CALT-DYLY-103-Compression-Tension-Pressure/dp/B0BYSLN766?th=1 ) . 
My load cells deliver a positive value and according to this [website](https://learn.sparkfun.com/tutorials/qwiic-scale-hookup-guide/all) they are supposed to. According to the tutorial a negative weight is the result of switched A+/A- wires.

with: 
```
sensor:
  - platform: nau7802
    name: "Scale"
    id: scale
    i2c_id: bus_a
    unit_of_measurement: "g"
    update_interval: 1s
    gain: 1
    offset_calibration: 585 
    gain_calibration: 16.13
```

the calibration_offset would double the value of 585 and a value of -585 results in something like -8387371.0000 for the empty scale. If I set the offset_calibration to -8388608 plus the offset it zeros out the empty scale but then the gain_calibration of 16.13 (8065 / 500 Gramm) results in wrong weight values. 

When I try higher gain values like the default value of 128, the gain_calibration would require a value of 2081 what is more than the max of 511.9999998807907 from the sensor.py.

I also tried to switch the load cells A+/A- wires to get negative values for the empty scale. Then I can zero the scale but the gain_calibration results in wrong values.

Am I supposed to zero the scale first with the gain_offset and then to calculate the gain_calibration by "measured value divided by known mass" or the other way around?

So if I try to just use the raw value from the nau7802 and do the calibration in esphome like this it is working, The gain setting could probably dropped too:
```

sensor:
  - platform: nau7802
    name: "Scale"
    id: scale
    i2c_id: bus_a
    unit_of_measurement: "g"
    update_interval: 1s
    gain: 1
    filters:
      - offset: -875
      - calibrate_linear:
         method: least_squares
         datapoints:
          - 0.0 -> 1.0
          - 8065.0 -> 500
      - clamp:
          min_value: 0
      - delta: 1
      - throttle_average: 1s

```

cujomalainey:
> I did some experimenting, I'm using this [nau7802 module](https://www.sparkfun.com/products/15242) and this [Strain load cell](https://www.amazon.com/Wishiot-Precision-Miniature-Electronic-Portable/dp/B0C3QHT4TT/ref=asc_df_B0C3QH7V8M/?tag=hyprod-20&linkCode=df0&hvadid=693071814688&hvpos=&hvnetw=g&hvrand=704602480246115246&hvpone=&hvptwo=&hvqmt=&hvdev=c&hvdvcmdl=&hvlocint=&hvlocphy=9012270&hvtargid=pla-2189119876076&mcid=e787aff36fc5353ea847dd1ccaf2af37&gad_source=1&th=1) but I also tried the Adafruit nau7802 and some other [S-shaped load cell](https://www.amazon.com/CALT-DYLY-103-Compression-Tension-Pressure/dp/B0BYSLN766?th=1) . My load cells deliver a positive value and according to this [website](https://learn.sparkfun.com/tutorials/qwiic-scale-hookup-guide/all) they are supposed to. According to the tutorial a negative weight is the result of switched A+/A- wires.
> 
> with:
> 
> ```
> sensor:
>   - platform: nau7802
>     name: "Scale"
>     id: scale
>     i2c_id: bus_a
>     unit_of_measurement: "g"
>     update_interval: 1s
>     gain: 1
>     offset_calibration: 585 
>     gain_calibration: 16.13
> ```
> 
> the calibration_offset would double the value of 585 and a value of -585 results in something like -8387371.0000 for the empty scale. If I set the offset_calibration to -8388608 plus the offset it zeros out the empty scale but then the gain_calibration of 16.13 (8065 / 500 Gramm) results in wrong weight values.
> 
> When I try higher gain values like the default value of 128, the gain_calibration would require a value of 2081 what is more than the max of 511.9999998807907 from the sensor.py.

Yes, this maps to the gain register which highest bit is 2^8 in significance. Are you trying to maximize the the dynamic range of your load cell? What level of precision do you need here?

> 
> I also tried to switch the load cells A+/A- wires to get negative values for the empty scale. Then I can zero the scale but the gain_calibration results in wrong values.

Wrong or does it error out?

> 
> Am I supposed to zero the scale first with the gain_offset and then to calculate the gain_calibration by "measured value divided by known mass" or the other way around?

Personally I have not played with the gain offset as much as my use case was not that high precision. My assumption would be that you would do external offset calibration, and assuming valid values attempt gain calibration. I would think there are scenarios where the range available is larger than the gain can cover and you would end up not able to utilize the fullness of the bandwidth.

> 
> So if I try to just use the raw value from the nau7802 and do the calibration in esphome like this it is working, The gain setting could probably dropped too:
> 

I intend to leave it as it is a hardware control for the device which might be needed in high precision use cases, but both are valid :) Also note the gain enum, is a analog control where as the offset and gain calibration are digital.

> ```
> 
> sensor:
>   - platform: nau7802
>     name: "Scale"
>     id: scale
>     i2c_id: bus_a
>     unit_of_measurement: "g"
>     update_interval: 1s
>     gain: 1
>     filters:
>       - offset: -875
>       - calibrate_linear:
>          method: least_squares
>          datapoints:
>           - 0.0 -> 1.0
>           - 8065.0 -> 500
>       - clamp:
>           min_value: 0
>       - delta: 1
>       - throttle_average: 1s
> ```

Also I am not aware, but if esphome has the concept of interrupt driven components vs polling, it would be great to enable it here since the device does have the ability to just drive samples directly from a pin.

gnsbglora:
I'm ok with the gain: 1, it results in smaller numbers.

In my project the load cell is used to determine the fuel level of a bottle, so zeroing it when it's replaced is useful. Right now the accuracy is about 2-3 Gram (0.3%). I thought that's pretty good. I'ts not very temperature stable. The nau7802 and the load cell is supposed to have a temperature sensor/compensation. 

Switching the A+/A- wires results in a negative value when empty and in values going more negative when a load is put on. No error and just wrong weight.  One could use the load cell upside-down but I'm not sure if that has hardware consequences. 

Not sure if the interrupt would help since the signal is pretty noisy.

cujomalainey:
> Not sure if the interrupt would help since the signal is pretty noisy.

Interrupt has no affect on the signal noise, it would just make sure you are getting a new value every time the device actually takes a sample, important when `poll_rate ~ sample_rate`

esphome/components/nau7802/__init__.py
```diff
@@ -0,0 +1,2 @@
+CODEOWNERS = ["@cujomalainey"]
+DEPENDENCIES = ["i2c"]
```
jesserockz:
```suggestion
```

esphome/components/nau7802/sensor.py
```diff
@@ -0,0 +1,131 @@
...
+from esphome import automation
+from esphome.automation import maybe_simple_id
+from esphome.components import sensor, i2c
+from esphome.const import (
+    CONF_GAIN,
+    CONF_ID,
+    ICON_SCALE,
+    STATE_CLASS_MEASUREMENT,
+)
+
```
jesserockz, line 10:
```suggestion

CODEOWNERS = ["@cujomalainey"]
DEPENDENCIES = ["i2c"]

```

esphome/components/nau7802/nau7802.h
```diff
@@ -0,0 +1,111 @@
...
+
+template<typename... Ts> class NAU7802CalbrateExternalOffsetAction : public Action<Ts...> {
+ public:
+  NAU7802CalbrateExternalOffsetAction(NAU7802Sensor *nau7802) : nau7802_(nau7802) {}
+
+  void play(Ts... x) override { this->nau7802_->calibrate_external_offset(); }
+
+ protected:
+  NAU7802Sensor *nau7802_;
+};
```
jesserockz: Please also apply to the others
```suggestion
template<typename... Ts> class NAU7802CalbrateExternalOffsetAction : public Action<Ts...>, public Parented<NAU7802Sensor> {
 public:
  void play(Ts... x) override { this->parent_->calibrate_external_offset(); }
};
```

esphome/components/nau7802/sensor.py
```diff
@@ -0,0 +1,131 @@
...
+CONF_OFFSET_CALIBRATION = "offset_calibration"
+CONF_LDO_VOLTAGE = "ldo_voltage"
+CONF_SAMPLES_PER_SECOND = "samples_per_second"
+
+nau7802_ns = cg.esphome_ns.namespace("nau7802")
+NAU7802Sensor = nau7802_ns.class_(
+    "NAU7802Sensor", sensor.Sensor, cg.PollingComponent, i2c.I2CDevice
+)
+NAU7802CalbrateExternalOffsetAction = nau7802_ns.class_(
+    "NAU7802CalbrateExternalOffsetAction", automation.Action
```
jesserockz:
```suggestion
    "NAU7802CalbrateExternalOffsetAction", automation.Action, cg.Parented.template(NAU7802Sensor)
```

esphome/components/nau7802/sensor.py
```diff
@@ -0,0 +1,131 @@
...
+    NAU7802_CALIBRATE_SCHEMA,
+)
+@automation.register_action(
+    "nau7802.calibrate_gain",
+    NAU7802CalbrateGainAction,
+    NAU7802_CALIBRATE_SCHEMA,
+)
+async def nau7802_calibrate_to_code(config, action_id, template_arg, args):
+    paren = await cg.get_variable(config[CONF_ID])
+    return cg.new_Pvariable(action_id, template_arg, paren)
```
jesserockz:
```suggestion
    var = cg.new_Pvariable(action_id, template_arg)
    await cg.register_parented(var, config[CONF_ID])
    return var
```

esphome/components/nau7802/nau7802.cpp
```diff
@@ -0,0 +1,285 @@
...
+    return;
+  }
+
+  this->status_clear_warning();
+
+  // Get the most recent sample to publish
+  int32_t result = 0;
+  do {
+    result = this->read_value_(ADCO_B2_REG, 3);
+  } while (is_data_ready_());
```
jesserockz: How long could this potentially run for on a 60s updated sensor?

esphome/components/nau7802/nau7802.cpp
```diff
@@ -0,0 +1,285 @@
...
+    return;
+  }
+
+  uint32_t gcal = (uint32_t) (round(this->gain_calibration_ * (1 << GCAL1_FRACTIONAL)));
+  this->write_value_(OCAL1_B2_REG, 3, this->offset_calibration_);
+  this->write_value_(GCAL1_B3_REG, 4, gcal);
+
+  // turn on AFE
+  pu_ctrl |= PU_CTRL_POWERUP_ANALOG;
+  delay(600);  // NOLINT
```
jesserockz: If you need to delay the rest of setup until a later time, you need to use `this->set_timeout(600, [this]{...});` to do in a scheduled task after 600ms
```suggestion
```

esphome/components/nau7802/nau7802.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+void NAU7802Sensor::dump_config() {
+  LOG_SENSOR("", "NAU7802", this);
+  LOG_I2C_DEVICE(this);
+
+  if (this->is_failed()) {
+    ESP_LOGE(TAG, "Communication with NAU7802 failed earlier, during setup");
+    return;
+  }
+  int32_t ocal = this->read_value_(OCAL1_B2_REG, 3);
```
jesserockz: `dump_config` should not be doing data operations. It should only be dumping the configuration that the user has set, or important information that was already read during `setup` (firmware versions for example are in some other components)

esphome/components/nau7802/nau7802.cpp
```diff
@@ -0,0 +1,285 @@
...
+}
+
+bool NAU7802Sensor::calibrate_(enum NAU7802CalibrationModes mode) {
+  i2c::I2CRegister ctrl2 = this->reg(CTRL2_REG);
+  // Calibrate
+  ctrl2 |= mode;
+  ctrl2 |= CTRL2_CALS;
+  while (ctrl2.get() & CTRL2_CALS) {
+    delay(10);
+  }
```
jesserockz: This needs to be moved to use a state machine and be changed to check this once per `loop` function.

esphome/components/nau7802/nau7802.cpp
```diff
@@ -0,0 +1,285 @@
...
+  ESP_LOGI(TAG, "New Gain: %f", gcal_f);
+  return true;
+}
+
+float NAU7802Sensor::get_setup_priority() const { return setup_priority::DATA; }
+
+void NAU7802Sensor::update() {
+  if (this->is_failed()) {
+    return;
+  }
```
jesserockz: `update` will not be called if it is marked as failed
```suggestion
```

esphome/components/nau7802/nau7802.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+  if (!is_data_ready_()) {
```
jesserockz: > All uses of class members and member functions should be prefixed with `this->` to distinguish them from global functions in code review.
https://esphome.io/guides/contributing#codebase-standards

```suggestion
  if (!this->is_data_ready_()) {
```

jesserockz (changes requested):

esphome/components/nau7802/nau7802.cpp
cujomalainey: Ack, i will work that in, the AFE has a long power on time.

esphome/components/nau7802/nau7802.cpp
cujomalainey: @jesserockz this is a bit of a special case where actions can change the values users set, so the applied config can change. this is beneficial for applying a constant config later. Is this fine?

esphome/components/nau7802/nau7802.cpp
cujomalainey: Ack, will move to non-blocking

esphome/components/nau7802/nau7802.cpp
cujomalainey: thanks for pointing this out, I mean't to delete this as the datasheet suggests in its wording it only has a single sample buffer.

One question I do have, are there any sensor examples that can be polling or driven by an interrupt pin? Would it be better to have two classes or just a polling component that no-ops on update if a pin is passed in?

esphome/components/nau7802/nau7802.cpp
cujomalainey: That is intentional. If calibration fails then the calibration registers are invalid, and either calibration needs to be performed again or the device needs to be reset. I can mark the respective calibrations as passed or not passed and block if one of them fails until it passes later, would that be acceptable?

cujomalainey:
@jesserockz sorry to bug you but what is the ETA? This has been pending for a month for a re-review.

esphome/components/nau7802/nau7802.cpp
jesserockz: What I mean is, the function `update()` will not be called by the underlying ESPHome code if the component was previously marked as failed with `this->mark_failed()` so you don't need to protect here.

cujomalainey: sorry i deleted it once i went back and realized what you meant, it should be fine now.

esphome/components/nau7802/nau7802.cpp
cujomalainey: also fixed this to just dump on parameter change so dump_config is only the passed in data

jesserockz:
Please do not force push a PR after a review has started. I now cannot see easily the changes you made after the review.

cujomalainey:
> Please do not force push a PR after a review has started. I now cannot see easily the changes you made after the review.

Apologies, I come from the kernel world where that is expected

jesserockz:
We don't care for specific commit history inside a branch/PR as PRs are always squashed with the PR title itself as the commit message. 
Would rather have the progress of the PR easier to follow =)

esphome/components/nau7802/sensor.py
```diff
@@ -0,0 +1,134 @@
...
+    cg.add(var.set_samples_per_second(config[CONF_SAMPLES_PER_SECOND]))
+    cg.add(var.set_ldo_voltage(config[CONF_LDO_VOLTAGE]))
+    cg.add(var.set_gain(config[CONF_GAIN]))
+    cg.add(var.set_gain_calibration(config[CONF_GAIN_CALIBRATION]))
+    cg.add(var.set_offset_calibration(config[CONF_OFFSET_CALIBRATION]))
+
+
+NAU7802_CALIBRATE_SCHEMA = maybe_simple_id(
+    {
+        cv.Required(CONF_ID): cv.use_id(NAU7802Sensor),
```
jesserockz: This will allow the user to not have to specify the id if there is only one instance
```suggestion
        cv.GenerateID(CONF_ID): cv.use_id(NAU7802Sensor),
```

esphome/components/nau7802/nau7802.h
```diff
@@ -0,0 +1,121 @@
...
+class NAU7802Sensor : public sensor::Sensor, public PollingComponent, public i2c::I2CDevice {
+ public:
+  void set_samples_per_second(NAU7802SPS sps) { sps_ = sps; }
+  void set_ldo_voltage(NAU7802LDO ldo) { ldo_ = ldo; }
+  void set_gain(NAU7802Gain gain) { gain_ = gain; }
+  void set_gain_calibration(float gain_calibration) { gain_calibration_ = gain_calibration; }
+  void set_offset_calibration(int32_t offset_calibration) { offset_calibration_ = offset_calibration; }
+  bool calibrate_external_offset() { return calibrate_(NAU7802_CALIBRATE_EXTERNAL_OFFSET); }
+  bool calibrate_internal_offset() { return calibrate_(NAU7802_CALIBRATE_INTERNAL_OFFSET); }
+  bool calibrate_gain() { return calibrate_(NAU7802_CALIBRATE_GAIN); }
```
jesserockz:
```suggestion
  bool calibrate_external_offset() { return this->calibrate_(NAU7802_CALIBRATE_EXTERNAL_OFFSET); }
  bool calibrate_internal_offset() { return this->calibrate_(NAU7802_CALIBRATE_INTERNAL_OFFSET); }
  bool calibrate_gain() { return this->calibrate_(NAU7802_CALIBRATE_GAIN); }
```

esphome/components/nau7802/nau7802.h
```diff
@@ -0,0 +1,121 @@
...
+  NAU7802_CALIBRATE_GAIN = 0b11,
+};
+
+class NAU7802Sensor : public sensor::Sensor, public PollingComponent, public i2c::I2CDevice {
+ public:
+  void set_samples_per_second(NAU7802SPS sps) { sps_ = sps; }
+  void set_ldo_voltage(NAU7802LDO ldo) { ldo_ = ldo; }
+  void set_gain(NAU7802Gain gain) { gain_ = gain; }
+  void set_gain_calibration(float gain_calibration) { gain_calibration_ = gain_calibration; }
+  void set_offset_calibration(int32_t offset_calibration) { offset_calibration_ = offset_calibration; }
```
jesserockz:
```suggestion
  void set_gain_calibration(float gain_calibration) { this->gain_calibration_ = gain_calibration; }
  void set_offset_calibration(int32_t offset_calibration) { this->offset_calibration_ = offset_calibration; }
```

esphome/components/nau7802/nau7802.h
```diff
@@ -0,0 +1,121 @@
...
+  NAU7802_CALIBRATE_EXTERNAL_OFFSET = 0b10,
+  NAU7802_CALIBRATE_INTERNAL_OFFSET = 0b00,
+  NAU7802_CALIBRATE_GAIN = 0b11,
+};
+
+class NAU7802Sensor : public sensor::Sensor, public PollingComponent, public i2c::I2CDevice {
+ public:
+  void set_samples_per_second(NAU7802SPS sps) { sps_ = sps; }
+  void set_ldo_voltage(NAU7802LDO ldo) { ldo_ = ldo; }
+  void set_gain(NAU7802Gain gain) { gain_ = gain; }
```
jesserockz:
```suggestion
  void set_samples_per_second(NAU7802SPS sps) { this->sps_ = sps; }
  void set_ldo_voltage(NAU7802LDO ldo) { this->ldo_ = ldo; }
  void set_gain(NAU7802Gain gain) { this->gain_ = gain; }
```

jesserockz (changes requested):

cujomalainey:
@jesserockz sorry friendly ping again

jesserockz (approved):

PR #6292, opened 2024-02-27
CSE7766 Apparent Power & Power Factor calculations

### LINKED ISSUE #5501: CSE7766 hard to use since ESPHome 2024.2.0

DAVe3283:

Since moving the averaging out of cse7766, calculating power factor & apparent power is much easier and more accurate when done inside the component. This adds optional Power Factor and Apparent Power sensors.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/issues#5501

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3643

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Assuming a Sonoff S31, Athom PG03V2, or similar

substitutions:
  update_interval: 30s

external_components:
  - source: "github://pr#6292"
    components: [ cse7766 ]

sensor:
  - platform: cse7766
    current:
      name: Current
      id: current
      filters:
        - throttle_average: ${update_interval}
    voltage:
      name: Voltage
      id: voltage
      filters:
        - throttle_average: ${update_interval}
    power:
      name: Power
      id: power
      filters:
        - throttle_average: ${update_interval}
    energy:
      name: Energy
      filters:
        - throttle: ${update_interval}
    apparent_power:
      name: Apparent Power
      filters:
        - throttle_average: ${update_interval}
    power_factor:
      name: Power Factor
      filters:
        - throttle_average: ${update_interval}
```

## Checklist:

### DIFF
--- a/esphome/components/cse7766/cse7766.cpp
+++ b/esphome/components/cse7766/cse7766.cpp
@@ -173,6 +173,32 @@ void CSE7766Component::parse_data_() {
     }
   }
 
+  if (have_voltage && have_current) {
+    const float apparent_power = voltage * current;
+    if (this->apparent_power_sensor_ != nullptr) {
+      this->apparent_power_sensor_->publish_state(apparent_power);
+    }
+    if (this->power_factor_sensor_ != nullptr && (have_power || power_cycle_exceeds_range)) {
+      float pf = NAN;
+      if (apparent_power > 0) {
+        pf = power / apparent_power;
+        if (pf < 0 || pf > 1) {
+          ESP_LOGD(TAG, "Impossible power factor: %.4f not in interval [0, 1]", pf);
+          pf = NAN;
+        }
+      } else if (apparent_power == 0 && power == 0) {
+        // No load, report ideal power factor
+        pf = 1.0f;
+      } else if (current == 0 && calculated_current <= 0.05f) {
+        // Datasheet: minimum measured current is 50mA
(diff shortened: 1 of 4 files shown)


puuu:
I tested this PR on S31 and it seems to work fine. Nice to have optional extra information.

jesserockz (approved):

PR #6297, opened 2024-02-27
Add support for AT581x component

X-Ryl669:

This adds support for the AT581x 5.8GHz radar that's found in ESP32S3-BOX-3 device.
It's a very power efficient human detection device that configurable via I2C.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** ~~fixes~~

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3646

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
external_components:
  - source: github://pr#6297
    components: at581x
    refresh: 0s

at581x:
  id: "Waveradar"
  i2c_id: bus_a
  detection_pin: 21

binary_sensor:
  - platform: at581x
    at581x_id: "Waveradar"
    motion_detected:
      name: "Radar motion"

switch:
  - platform: at581x
    at581x_id: "Waveradar"
    rf:
      name: "Enable Radar"

i2c:
  - id: bus_a
    sda: GPIO41
    scl: GPIO40
    sda_pullup_enabled: false
    scl_pullup_enabled: false
    scan: false
    frequency: 400kHz

esphome:
  [...]
  on_boot:
    priority: 600
    then:
      - at581x.settings:
          id: "Waveradar"
          factory_reset: false
          frequency: 5800  # In MHz
          sensing_distance: 200 # 0-1023
          poweron_selfcheck_time: 2000 # in ms
          protect_time: 1000 # in ms
          trigger_base: 500  # in ms
          trigger_keep: 10000 # in ms
          stage_gain: 3 # 0-12 the higher the value the smaller the gain
          power_consumption: 70 # in µA
```

## Checklist:


X-Ryl669:
The failing CI test is completely random (and unrelated to the code). I think it's a CI failure

esphome/components/at581x/switch/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+DEPENDENCIES = ["at581x"]
+
+RFSwitch = at581x_ns.class_("RFSwitch", switch.Switch)
+
+CONF_RF = "rf"
+
+CONFIG_SCHEMA = {
+    cv.GenerateID(CONF_AT581X_ID): cv.use_id(AT581XComponent),
+    cv.Optional(CONF_RF): switch.switch_schema(
```
jesserockz: I feel like this should have a longer, better descriptive name. Maybe `RF_POWER` or `SENSOR_POWER` etc?

esphome/components/at581x/__init__.py
```diff
@@ -0,0 +1,186 @@
...
+                5784,
+                5800,
+                5819,
+                5836,
+                5851,
+                5869,
+                5888,
+                int=True,
+            )
+        ),
```
jesserockz: Please move these values to a const list higher up in the file, and also validate with `cv.frequency`. We don't like having "random" number config items that don't have units

esphome/components/at581x/__init__.py
```diff
@@ -0,0 +1,186 @@
...
+                44,
+                47,
+                51,
+                54,
+                61,
+                68,
+                78,
+                int=True,
+            )
+        ),
```
jesserockz: Same here

esphome/components/at581x/__init__.py
```diff
@@ -0,0 +1,186 @@
...
+                5888,
+                int=True,
+            )
+        ),
+        cv.Optional(CONF_SENSING_DISTANCE, default=823): cv.templatable(
+            cv.int_range(min=0, max=1023)
+        ),
+        cv.Optional(CONF_POWERON_SELFCHECK_TIME, default=2000): cv.templatable(
+            cv.int_range(min=0, max=65535)
+        ),
```
jesserockz: These should have units too

esphome/components/at581x/at581x.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+namespace esphome {
+namespace at581x {
+
+static const char *const TAG = "at581x";
+
+AT581XComponent::AT581XComponent() {}
+
+bool AT581XComponent::i2c_write_reg(uint8_t addr, uint8_t data) {
+  return write_register(addr, &data, 1) == esphome::i2c::NO_ERROR;
```
jesserockz: All member function calls and variables should be prefixed with `this->` please
```suggestion
  return this->write_register(addr, &data, 1) == esphome::i2c::NO_ERROR;
```

esphome/components/at581x/at581x.cpp
```diff
@@ -0,0 +1,213 @@
...
+  // The main operation is to detect the human presence
+  bool state = this->detection_pin_->digital_read();
+  if (this->motion_binary_sensor_ != nullptr) {
+    this->motion_binary_sensor_->publish_state(state);
+  }
+}
+void AT581XComponent::dump_config() {
+  if (this->motion_binary_sensor_ != nullptr) {
+    LOG_BINARY_SENSOR("", "AT581X", this->motion_binary_sensor_);
+  }
```
jesserockz: Not required as the macro does this already
```suggestion
  LOG_BINARY_SENSOR("", "AT581X", this->motion_binary_sensor_);
```

esphome/components/at581x/at581x.cpp
```diff
@@ -0,0 +1,213 @@
...
+  LOG_PIN("  Pin: ", this->detection_pin_);
+  LOG_I2C_DEVICE(this);
+}
+#define ArrSz(X) (sizeof(X) / sizeof((X)[0]))
```
jesserockz: It is preferred to use long/descriptive/meaningful names rather than short ones.
```suggestion
#define ARRAY_SIZE(X) (sizeof(X) / sizeof((X)[0]))
```

esphome/components/at581x/at581x.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// float AT581XComponent::get_setup_priority() const { return 0; }
+bool AT581XComponent::set_factory_reset() {
+  if (!i2c_write_reg(RESET_ADDR, (uint8_t) 0) || !i2c_write_reg(RESET_ADDR, (uint8_t) 1)) {
+    ESP_LOGE(TAG, "Failed to reset AT581X component");
+    return false;
+  }
+  return true;
+}
```
jesserockz: I don't think that this function should be called `factory_reset` considering its not actually factory resetting, but simply restarting?

esphome/components/at581x/at581x.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
+
+#include <utility>
+
+#include "esphome/core/component.h"
+#include "esphome/core/hal.h"
+#include "esphome/components/binary_sensor/binary_sensor.h"
+#include "esphome/components/sensor/sensor.h"
+#include "esphome/components/switch/switch.h"
```
jesserockz: All of these and any other references to the classes and variables need to be protected with `#ifdef USE_BINARY_SENSOR` etc.
If someone configures this device with no other switches for example in YAML, this will fail compilation.

esphome/components/at581x/automation.h
```diff
@@ -0,0 +1,71 @@
...
+      int v = this->stage_gain_.value(x...);
+      this->parent_->set_stage_gain(v);
+    }
+
+    // This actually perform all the modification on the system
+    this->parent_->i2c_write_config();
+
+    if (this->factory_reset_.has_value() && this->factory_reset_.value(x...) == true) {
+      this->parent_->set_factory_reset();
+    }
```
jesserockz: The last line of `i2c_write_config` is `set_factory_reset`, so is this really needed?

esphome/components/at581x/binary_sensor.py
```diff
@@ -0,0 +1,28 @@
...
+
+from . import CONF_AT581X_ID, AT581XComponent
+
+DEPENDENCIES = ["at581x"]
+
+CONF_MOTION_DETECTED = "motion_detected"
+
+CONFIG_SCHEMA = {
+    cv.GenerateID(CONF_AT581X_ID): cv.use_id(AT581XComponent),
+    cv.Optional(CONF_MOTION_DETECTED): binary_sensor.binary_sensor_schema(
```
jesserockz: Can we please change this to be in line with the other motion/radar/mmwave sensors?
```suggestion
    cv.Optional(CONF_HAS_TARGET): binary_sensor.binary_sensor_schema(
```

jesserockz: The alternative is to not make this a child sensor of the config at all, but just directly on the CONFIG_SCHEMA since it is the only one.

tests/test1.yaml
jesserockz, line 1: Can we move the tests into the new test file format please under `tests/components/at581x`?

jesserockz (changes requested):

esphome/components/at581x/__init__.py
X-Ryl669: The sensing distance is unitless (or it's not specified in the datasheet). 
Ok for the time stuff, they are all milliseconds.

esphome/components/at581x/at581x.cpp
X-Ryl669: Agree. It's the same name as in Espressif's driver but it doesn't make sense. It's more like a RF's hardware reset.

esphome/components/at581x/automation.h
X-Ryl669: Yes, see above

esphome/components/at581x/binary_sensor.py
X-Ryl669: I'm not 100% sure I get it here. This radar detects motion not static target (unlike 24GHz or 60GHz? IIUC). If you stop moving in front of the radar, the detection signal will go down. 

In the datasheet, there's also a light detection sensor, but I don't know if it's an error in the Chinese to English translation (in that case, light would mean RF emission, I'm not sure), or if it's actually here. 

What's sure is that on Box S3, the radar is under a opaque plastic cover so it can't detect light.

What do you mean about not making a child sensor, do you have an example I can look at?

X-Ryl669:
I'm waiting to test on the real hardware before committing the last change that'll modify `esphome/const.py` for all other components (but will make the PR unusable as an `external_components`). Maybe these kinds of changes should be in another PR that should be merged at the same time as this one, what do you think?

tests/test1.yaml
X-Ryl669: This was a huge PITA. Since it's using I2C I had to guess what pins I could use for each supported platform (C3, C6, ESP32, ESP8266, etc...) since each have their own constraint on strapping pin/flash pins. I wasn't able to figure out how to generate the complete YAML from the test YAML on the command line, so I relied on trial and error here with the CI. 

Maybe a better documentation on how to run those tests locally is required.

X-Ryl669:
Tested on my Box S3 and it's working as expected.

esphome/components/at581x/binary_sensor.py
```diff
@@ -0,0 +1,27 @@
...
+
+DEPENDENCIES = ["at581x"]
+
+CONFIG_SCHEMA = {
+    cv.GenerateID(CONF_AT581X_ID): cv.use_id(AT581XComponent),
+    cv.Optional(CONF_HAS_TARGET): binary_sensor.binary_sensor_schema(
+        device_class=DEVICE_CLASS_MOTION,
+        icon=ICON_MOTION_SENSOR,
+    ),
+}
```
jesserockz: > I'm not 100% sure I get it here. This radar detects motion not static target (unlike 24GHz or 60GHz? IIUC). If you stop moving in front of the radar, the detection signal will go down.
> 
> In the datasheet, there's also a light detection sensor, but I don't know if it's an error in the Chinese to English translation (in that case, light would mean RF emission, I'm not sure), or if it's actually here.
>
> What's sure is that on Box S3, the radar is under a opaque plastic cover so it can't detect light.
>
> What do you mean about not making a child sensor, do you have an example I can look at?

So because this is the only binary sensor, there is no real need to have it as a deeper object in the config. See the below suggestion.

```suggestion
CONFIG_SCHEMA = (
    binary_sensor.binary_sensor_schema(
        device_class=DEVICE_CLASS_MOTION,
        icon=ICON_MOTION_SENSOR,
    ).extend(cv.Schema(
        {
            cv.GenerateID(CONF_AT581X_ID): cv.use_id(AT581XComponent),
        }
    )
)

```

This would allow config like:
```yaml
binary_sensor:
  - platform: at581x
    name: Motion Detected
    at581x_id: ...
```

X-Ryl669:
I think everything should be ok now.

esphome/components/at581x/at581x.cpp
```diff
@@ -0,0 +1,215 @@
...
+const uint8_t TRIGGER_KEEP_TIME_ADDR = 0x42;  // 4 bytes, so up to 0x45
+const uint8_t TIME41_VALUE = 1;
+const uint8_t SELF_CHECK_TIME_ADDR = 0x38;  // 2 bytes, up to 0x39
+
+namespace esphome {
+namespace at581x {
+
+static const char *const TAG = "at581x";
+
+AT581XComponent::AT581XComponent() {}
```
jesserockz: No need to create an empty constructor
```suggestion
```

jesserockz (changes requested):
After reviewing the code and how it actually works, I don't think this entire component needs its own binary_sensor platform and the `detection_pin` config item. 

Setting up the main component, and settings and power switch is fine, but for detecting motion, the user can just set up a `gpio` binary sensor on that pin.

jesserockz (approved):
Thanks for making that change.

PR #6303, opened 2024-02-29
aht10: Added new CMD and renamed existing CMD to match datasheet

cptskippy:

Sometimes the AHT10 setup fails following a firmware update of the micro controller and a full power cycle of the device is necessary in order to get it working again. The [AHT10](https://www.maritex.com.pl/product/attachment/147096/c5093eda6658ef654b651f3d5705c2ef) and [AHT20](https://cdn.sparkfun.com/assets/d/2/b/e/d/AHT20.pdf) datasheets define a Soft Reset command:

> 5.4 Soft Reset
> This command (see Table 9) is used to restart the sensor system without having to turn off and turn on the power again. After receiving this command, the sensor system starts to reinitialize and restore the default settings. Soft reset takes no more than 20 milliseconds.

By adding the Soft Reset command during setup we can avoid the need to power cycle the device to get the AHT10 working again.

In addition to adding the Soft Reset command, I renamed the Calibrate command to Initialize as that is the correct name in the datasheets. The AHT10 and AHT20 are calibrated at the factory, not by the Initialize command.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
    external_components:
      - source:
          url: https://github.com/cptskippy/esphome.aht10
          type: git
          ref: main
        components: aht10
        refresh: 0s
    
    sensor:
      - platform: aht10
        update_interval: 20s

    i2c:
      sda: GPIO0
      scl: GPIO2
      scan: true
```

## Checklist:


esphome/components/aht10/aht10.cpp
```diff
@@ -21,30 +21,39 @@ namespace esphome {
...
+static const uint8_t AHT10_INIT_ATTEMPTS = 10;
+
 static const uint8_t AHT10_STATUS_BUSY = 0x80;
 
 void AHT10Component::setup() {
-  const uint8_t *calibrate_cmd;
+  if (this->write(AHT10_SOFTRESET_CMD, sizeof(AHT10_SOFTRESET_CMD)) != i2c::ERROR_OK) {
+    ESP_LOGE(TAG, "Reset AHT10 failed!");
+  }
+  delay(AHT10_SOFTRESET_DELAY);
```
cptskippy, line 42: This is the only functional change.  I've added a soft reset call as the first part of setup to put the component into a known good state of operation.

cptskippy (commented):
Aside from lines 39-42, all other changes are just cosmetic changes to the code.

The cosmetic changes include renaming some constants and log messages to align with the terminology used in the manufacturer's datasheet, as well as rearranging the order of declarations to group them by type (e.g. commands, delays, counters).

jesserockz (approved):

PR #6307, opened 2024-02-29
Optimize display rendering

Mat931:

Significantly improves rendering performance on 1-bit-per-pixel displays by writing bytes instead of individual pixels to the buffer.

Currently only implemented for image rendering. Other stuff like fonts, lines, rectangles etc can be modified later to use the new features. Already works with clipping and display rotation.

As a demonstration the ssd1306 component was modified to provide this new feature, but all of these display components can be easily modified later:

- inkplate6
- max7219digit
- pcd8544
- ssd1306_base
- st7920
- waveshare_epaper
- st7567_base

Benchmark results for drawing a 128x104 ESPHome logo on a 128x64 display:
`24.1ms` - before (old code)
`4.0ms` - if the image y position is divisible by 8 (depending on display type and rotation this can apply to the x position instead)
`6.8ms` - if the y position is not divisible by 8

Depending on the display model and rotation the format of the image data must be set correctly in the config file. If set incorrectly a warning appears in the logs and the image is rendered using the old method.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
i2c:
  sda: GPIO5
  scl: GPIO4
  frequency: 800kHz

display:
  - platform: ssd1306_i2c
    model: "SSD1306 128x64"
    address: 0x3C
    update_interval: 0.3s
    rotation: 0°
    lambda: |-
      it.image(0, 0, id(esphome_logo));
      //it.image(1, 1, id(esphome_logo));

image:
  - file: https://esphome.io/_images/logo.png
    id: esphome_logo
    resize: 128x128
    type: BINARY_VERTICAL

```

## Checklist:


nielsnl68:
Hi, @Mat931, i'm working on a display driver component that should be as interface between the display and driver components. As you have been working streamlining the drivers above, are you willing to help out to make this driver interface working with the displays you made better?
Check my work at: 

https://github.com/nielsnl68/esphome/tree/nvds-displaydriver/esphome/components/display_driver
~~https://github.com/nielsnl68/esphome/tree/nvds-displaydriver/esphome/components/ili9xxx/display_driver~~

Mat931:
@nielsnl68 Sure! Can you explain what your driver does?

nielsnl68:
> @nielsnl68 Sure! Can you explain what your driver does?

The display_driver class is an continuity from the ili9xxx platform component.
The main features that i want to solve with this component is as follows:

- a simplified and standardized interface to talk to the displays.
- support for all 24bit, 18bit, 16bit, 8bit. 8bit palette/indexed, 4bit, 2bit and 1bit screens
- support for different way's to store the color information based on the display requirements.
- allow multiple methods to talk to a display: I2C, SPI, parallel, RGB(eso32s3) etc. 
- Use the buffer only when needed and try to setup the size as perfect for the display. 
- using the fastest way to send content to the display. 
- prepare for the use of the LVGL component.

Most of the features are already implemented but i have not yet implemented other displays then those that where available in the ili9xxx platform class.

Mat931:
So you want your driver component to replace all the different display components or move some functionality from them to the driver component? For example I noticed that all the draw_vertical_pixel_group_internal functions mostly do the same thing with maybe minimal differences. They could be merged into a driver component for example.

>     support for different way's to store the color information based on the display requirements.

How do you want to solve this on the Python side? This would be interesting for this PR too because right now the user has to configure the right format for the display and orientation. (vertical or horizontal pixel groups) Are you planning to auto-detect the format based on the display type somehow? And what if there are multiple different displays?

>     Use the buffer only when needed and try to setup the size as perfect for the display.

For displays with multiple pixels in one byte I think the buffer is always needed. This PR relies on it to handle things like transparency, clipping, drawing images at arbitrary y-positions etc.

Mat931:
Here's what I found out about the data format of all the 1bpp displays:
This will be needed when making them compatible with this PR
```
max7219digit:
 Normal colors: color_on = 1, color_off = 0
 Vertical pixel groups
 Bit 01 is at the top of the pixel group
 Custom buffer

pcd8544:
 Normal colors: color_on = 1, color_off = 0
 Vertical pixel groups
 Bit 01 is at the top of the pixel group
 Byte order in buffer: Left to right, then top to bottom

ssd1306_base:
 Normal colors: color_on = 1, color_off = 0
 Vertical pixel groups
 Bit 01 is at the top of the pixel group
 Byte order in buffer: Left to right, then top to bottom

st7567_base:
 Normal colors: color_on = 1, color_off = 0
 Vertical pixel groups
 Bit 01 is at the top of the pixel group
 Byte order in buffer: Left to right, then top to bottom

inkplate6:
 Inverted colors: color_on = 0, color_off = 1
 Horizontal pixel groups
 Bit 01 is on the left of the pixel group
 Byte order in buffer: Top to bottom, then left to right

waveshare_epaper:
 Inverted colors: color_on = 0, color_off = 1
 Horizontal pixel groups
 Bit 01 is on the right of the pixel group
 Byte order in buffer: Left to right, then top to bottom

st7920:
 Normal colors: color_on = 1, color_off = 0
 Horizontal pixel groups
 Bit 01 is on the right of the pixel group
 Byte order in buffer: Left to right, then top to bottom
```

nielsnl68:
Just to be sure; what do you mean with "pixel groups"? Are you talking about the bytes wherein the pixels are stored?

Mat931:
Exactly

nielsnl68:
> For example I noticed that all the draw_vertical_pixel_group_internal functions mostly do the same thing with maybe minimal differences. They could be merged into a driver component for example

Yes, that is exactly what i want to do indeed.

> How do you want to solve this on the Python side?

I have enhanced the `colorutil::ColorBitness` which now stores all kind of display data like number of bits per color, how many bytes are used and how the pixel is stored. This value could be added to the python part. 
 
> Are you planning to auto-detect the format based on the display type somehow? And what if there are multiple different displays?

No plans for this right now. It is all the users responsibility to give the compiler the right information.
> For displays with multiple pixels in one byte I think the buffer is always needed. This PR relies on it to handle things like transparency, clipping, drawing images at arbitrary y-positions etc.

That would be nice to be integrated in the *display_component*

> about the `draw_vertical_pixel_group` etc.

Sorry to say, the name *group* for what you mean does not sounds fine to me. 
Also i am not sure of we should point that out as well, because not every display stores multiple pixels in a byte (group). 
Having the vertical and horizontal line functions virtual to be override when there is a faster way to draw that is something of interest to me.

Mat931:
I'm open to renaming anything, the names are just placeholders. I think BINARY_VERTICAL, TRANSPARENT_BINARY_VERTICAL should be renamed to something that makes a bit more sense too.

PR #6308, opened 2024-02-29
Make substitutions work for any YAML type, not only strings

fakuivan:

Substitutions only allow variables to be strings, this makes sense since they are implemented as string interpolation, however this is rather limiting for cases where YAML anchors cannot be used, like in anchors defined across files (as seen in [issues/1011](https://github.com/esphome/issues/issues/1011)). This PR allows substitutions of any other YAML types besides strings when the substitution takes over the entire field, meaning that no string interpolation is intended, just a straight replacement.

Here's an example of what is valid before this PR:

```yaml
substitutions:
  my_pin: D8

binary_sensor:
- platform: gpio
  name: My pin sensor
  pin: $my_pin
```

Now, say for example that instead we need our pin to be in pull up mode, or to be inverted, without this PR we would need to do something like this:

```yaml
substitutions:
  my_pin_number: D8
  my_pin_inverted: "True"

binary_sensor:
- platform: gpio
  name: My pin sensor
  pin:
    number: $my_pin_number
    inverted: $my_pin_inverted
```

This is because we can't directly substitute an entire pin schema, as it's a YAML record and not a plain string. With this patch, `my_pin` will be able to hold a pin schema like so:

```yaml
substitutions:
  my_pin: 
    number: D8
    inverted: True

binary_sensor:
- platform: gpio
  name: My pin sensor
  pin: $my_pin
```

Notice that the binary sensor definition is the same as in the first example, meaning that `my_pin` can now either be a pin number as we originally intended, or it can be a pin schema, without having to modify the sensor definition. However, if the definition of the sensor does string interpolation with `my_pin`, verification will fail since only strings can be interpolated.

```yaml
substitutions:
  my_pin: 
    number: D8
    inverted: True

binary_sensor:
- platform: gpio
  # this interpolation will break validation
  name: My pin sensor number $my_pin
  pin: $my_pin
```

In this case, we can be more specific while still using substitutions:

```yaml
substitutions:
  my_pin_number: D8
  my_pin:
    number: $my_pin_number
    inverted: True

binary_sensor:
- platform: gpio
  # works just fine, although it'd be nice to have a way to access fields like `${my_pin_number.number}`
  name: My pin sensor number $my_pin_number
  pin: $my_pin
```

Notice the nested substitution in `my_pin`, my understanding is that this "just works" without doing any changes to the other parts of the substitution code since it will substitute stuff until the result of the substitution is the same as what we started with, checking all the nested elements (this is what python does by default with `__eq__` for `OrderedDict`s, `list`s and `set`s).

I was thinking maybe adding some stringification or basic evaluation of algebraic expressions for numeric types, or merging for dicts, but that felt like it could derail into a jinja2 clone. I feel like this is a sensible middleground between implementation complexity and usefulness.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3659

## Test Environment

Not device specific

```yaml
host: {}
# Example config.yaml
substitutions:
  pin_data:
    number: D8
    inverted: True
binary_sensor:
  - platform: gpio
    name: Parametric pin
    pin: $pin_data
```

## Checklist:

### DIFF
--- a/esphome/components/substitutions/__init__.py
+++ b/esphome/components/substitutions/__init__.py
@@ -62,7 +62,7 @@ def _find_tokens(value):
         if name.startswith("{") and name.endswith("}"):
             name = name[1:-1]
 
-        yield match.group(1), (start - last_end, end - start)
+        yield name, (start - last_end, end - start)
         last_end = end
 
 
--- a/esphome/components/substitutions/__init__.py
+++ b/esphome/components/substitutions/__init__.py
@@ -66,7 +66,7 @@ def _find_tokens(value):
         last_end = end
 
 
-def _expand_substitutions(substitutions, value, path, ignore_missing):
+def _expand_substitutions(substitutions, value, path, ignore_missing, is_key=False):
     substituted = ""
     start_from = 0
     for name, (ignored_chars, length) in _find_tokens(value):
@@ -89,6 +89,13 @@ def _expand_substitutions(substitutions, value, path, ignore_missing):
             start_from += ignored_chars + length
             continue
 
+        if is_key:
+            raise cv.Invalid(
+                "Key substitution is only allowed for string types, "
+                f"however {name!r} (used in {'->'.join(str(x) for x in path)}) "
+                f"is of type {type(sub)}"
+            )
+
         if length != len(value):
             raise cv.Invalid(
                 "String interpolation is only allowed for substitutions with "
@@ -110,7 +117,7 @@ def _expand_substitutions(substitutions, value, path, ignore_missing):
     return substituted
 
 
-def _substitute_item(substitutions, item, path, ignore_missing):
+def _substitute_item(substitutions, item, path, ignore_missing, is_key=False):
     if isinstance(item, list):
         for i, it in enumerate(item):
             sub = _substitute_item(substitutions, it, path + [i], ignore_missing)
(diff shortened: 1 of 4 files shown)


jesserockz:
Have you heard of YAML anchors?

```yaml
.pin_data: &pin_data
    number: D8
    inverted: True
    
binary_sensor:
  name: Sensor for pin
  pin: *pin_data
```

fakuivan:
@jesserockz YAML anchors don't work across files. If you want an imported `package:` to have the pin definitions and use them throughout the main config, anchors won't work

fakuivan:
I think this would also need to be modified, as `!include { vars: ... }` is related to this

https://github.com/esphome/esphome/blob/357ac3b85f6cff3c93ec8a3c29daca1e945a4fda/esphome/yaml_util.py#L286-L326

I'll turn it into a draft until I take care of the `!include` case

nielsnl68 (approved):
looks good to me.

fakuivan:
Fixed edge cases and added the PR for the docs. Ready for review.

jpeletier:
Hello! Enabling substitutions to work for any YAML type, including in package vars, is part of the renewed Jinja PR #8955 so that substitution variables can be used in jinja expressions. Let's push for this together!

jpeletier:
@fakuivan I suggest you close this since it is solved by https://github.com/esphome/esphome/pull/8955 
Thanks!

PR #6311, opened 2024-03-01
ADE7953: Add the ability to use accumulating energy registers, more precise power reporting

isorin:

Use ADE7935 accumulated energy registers instead of instant power registers. These registers store the accumulated energy since the last read and should provide better accuracy for power calculation.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/3656

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
  - platform: ade7953_i2c
    use_accumulated_energy_registers: true
```

## Checklist:


esphome/components/ade7953_base/ade7953_base.cpp
```diff
@@ -69,6 +71,9 @@ void ADE7953::dump_config() {
   }
 #define ADE_PUBLISH(name, val, factor) ADE_PUBLISH_(name, val, factor)
 
+#define ADE_POWER_FACTOR 154.0f
+#define ADE_WATTSEC_POWER_FACTOR (ADE_POWER_FACTOR * ADE_POWER_FACTOR / 3600)
```
jesserockz: Please use static consts for these

jesserockz (changes requested):

esphome/components/ade7953_base/ade7953_base.cpp
isorin: Done, thanks for reviewing.

jesserockz (approved):

PR #6320, opened 2024-03-04
fix tmp102 negative calculation

ssieb:

Corrects the change in https://github.com/esphome/esphome/pull/6316.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/tmp102/tmp102.cpp
+++ b/esphome/components/tmp102/tmp102.cpp
@@ -39,9 +39,6 @@ void TMP102Component::update() {
     return;
   }
   raw_temperature = i2c::i2ctohs(raw_temperature);
-  if (raw_temperature & 0x8000) {
-    raw_temperature |= 0xF000;
-  }
   raw_temperature = raw_temperature >> 4;
   float temperature = raw_temperature * TMP102_CONVERSION_FACTOR;
   ESP_LOGD(TAG, "Got Temperature=%.1f°C", temperature);


jesserockz (approved):

esphome/components/tmp102/tmp102.cpp
```diff
@@ -39,9 +39,6 @@ void TMP102Component::update() {
     return;
   }
   raw_temperature = i2c::i2ctohs(raw_temperature);
-  if (raw_temperature & 0x8000) {
```
timsavage, line 42: What is the reason for removing this code here?

timsavage: This is required to move the high bit to allow for negative numbers.

timsavage: I suspect a proper fix might be to look at the example in the SparkFun library [here](https://github.com/sparkfun/SparkFun_TMP102_Arduino_Library/blob/master/src/SparkFunTMP102.cpp).

Check if the device is in 12 or 13bit mode and apply the correct bit manipulation.

ssieb: That code isn't moving anything.  The high bit is already set, it's a 2s-complement number and the right shift will work correctly.   The only change that was required was the data type change.  The code that I removed actually modified the value.  It would only matter if it was a very negative temperature, but it was incorrect.

PR #6335, opened 2024-03-08
Add IRK support to allow tracking of devices with random MAC addresses

chbmuc:

Most modern smartphones use use a privacy feature called Resolvable Private Addresses to avoid BLE tracking. However if you know the devices "Identity Resolving Key" (IRK), you can check if the generated private MAC address belongs to the device with the IRK.

This PR adds an option to the `ble_presence` sensor to configure the IRK.

There is no support to obtain the key. You will have to use one of the options described here: https://espresense.com/beacons

Once you have the IRK, configuration is similar to the other methods for the `ble_presence` binary sensor.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** feature (as requested in esphome/feature-requests#2285)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation:** esphome/esphome-docs#3666

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example configuration entry
esp32_ble_tracker:

binary_sensor:
  # Presence based on IRK
  - platform: ble_presence
    irk: 1234567890abcdef1234567890abcdef
    name: "ESP32 BLE Tracker with Identity Resolving Key"
```

## Checklist:


esphome/const.py
```diff
@@ -377,6 +377,7 @@
 CONF_INVERT_COLORS = "invert_colors"
 CONF_INVERTED = "inverted"
 CONF_IP_ADDRESS = "ip_address"
+CONF_IRK = "irk"
```
clydebarrow: Given that this constant is not used, nor likely to be used, anywhere other than the BLE code, I'd suggest it be defined locally, not in the global const.py.

chbmuc: Sure! I just looked where all these definitions come from and did it the same way.

clydebarrow: Yes, there are some definitions in the global const.py that probably should not be there. The CI checks have a threshold of 4 separate definitions before it enforces a constant to be globally defined. One problem with modifying core code in a PR is that it then makes the PR unusable as an external component, since that does not pull in core changes. Anyway, looking good now.

clydebarrow (approved):

PR #6341, opened 2024-03-10
Default ILI9xxx SPI speed now works with ESP-IDF

HeMan:

Default SPI speed (40 MHz) for ILI9xxx gav the following message:
```
E (296) spi_hal: spi_hal_cal_clock_conf(102): When work in full-duplex mode at frequency > 26.7MHz, device cannot read correct data.
[08:40:58]Try to use IOMUX pins to increase the frequency limit, or use the half duplex mode.
[08:40:58]Please note the SPI master can only work at divisors of 80MHz, and the driver always tries to find the closest frequency to your configuration.
[08:40:58]Specify ``SPI_DEVICE_NO_DUMMY`` to ignore this checking. Then you can output data at higher speed, or read data at you
[08:40:58][D][esp-idf:000]: E (335) spi_master: spi_bus_add_device(362): assigned clock speed not supported
```

Changed to 20MHz
## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: ilidisplay

esp32:
  board: esp32dev
  framework:
    type: esp-idf

logger:

spi:
  clk_pin: 18
  miso_pin: 12
  mosi_pin: 23
  interface: any


display:
  - platform: ili9xxx
    model: TFT_2.4
    cs_pin: 27
    dc_pin: 32
    reset_pin: 14

```

## Checklist:

### DIFF
--- a/esphome/components/ili9xxx/display.py
+++ b/esphome/components/ili9xxx/display.py
@@ -152,7 +152,7 @@ CONFIG_SCHEMA = cv.All(
         }
     )
     .extend(cv.polling_component_schema("1s"))
-    .extend(spi.spi_device_schema(False, "40MHz")),
+    .extend(spi.spi_device_schema(False, "20MHz")),
     cv.has_at_most_one_key(CONF_PAGES, CONF_LAMBDA),
     _validate,
 )


clydebarrow (changes requested):
The 40MHz rate is fine with ESP-IDF, unless it's on an ESP32 and the pins used are not the default for the particular SPI bus.

There are other reasons why 40MHz might not work, e.g. the display driver chip might not support that speed, or the hardware interface is bad (e.g. DuPont wires) but the same is true for any arbitrary default speed. Setting it to 20MHz when it could have been 40 will make the display slower.


After considering all that I am of the opinion that (for this component at least) the SPI data rate should be a required config option, with no default. This could be achieved by replacing the entry in the schema with `cv.Required(CONF_DATA_RATE): ...`

This would be a breaking change of course - but so is the original proposal and at least making it required breaks the code at build time, not just introduces a subtle runtime change in behaviour.

nielsnl68:
As what @clydebarrow said, for now you can set the **data_rate** as follows:

```yaml
display:
  - platform: ili9xxx
    model: TFT_2.4
    data_rate: 20Mhz
```

nagyrobi:
Closing because an alternate approach is being condidered to mitigate this.

PR #6349, opened 2024-03-10
[bh1745] Colour Sensor

### LINKED ISSUE #2639: Please add support for the BH1745NUC Colour Sensor
The BH1745NUC colour sensor is not yet supported, only the BH1750 is currently.

I have searched for a solution but cannot find anything. I have even tried to adapt the BH1750 code from the esphome repository and  tried make a custom component but I do not have the knowledge to make it work.

**Additional context**

latonita:

RGBC sensor support + lux and CCT calculations

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2639

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3680

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

external_components:
  - source: github://pr#6349
    refresh: 10s
    components:
      - bh1745

i2c:
#...

sensor:
  - platform: bh1745
    id: bh1745_sensor1
    address: 0x38
    gain: 1X  # 1X, 2X, 16X
    integration_time: 160ms  # 160ms, 320ms, 640ms, 1280ms, 2560ms, 5120ms
    red_channel: Red channel
    green_channel: Green channel
    blue_channel: Blue channel
    clear_channel: Clear channel
    illuminance: Illuminance
    color_temperature: Color temperature
    update_interval: 60s

# Optional: the INT pin can be driven as an open drain output, which is how the
# Pimoroni breakout wires up its illumination LEDs.
output:
  - platform: bh1745
    id: bh1745_leds
    bh1745_id: bh1745_sensor1

light:
  - platform: binary
    name: BH1745 LEDs
    output: bh1745_leds
```

## Checklist:


jesserockz:
@latonita I hope you dont mind. I restructured this slightly. Instead of having `pimoroni_led_switch` exposing a switch entity, there is now an `output` platform that can be attached to a `binary` light for example, or anything you want to connect, should you be building a circuit using this chip.

latonita:
@jesserockz i'm absolutely okay. 
thanks

esphbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - Channel compensation multiplies floats back into `uint16_t` — overflows in ordinary light
> - Lux/CCT formulas are applied to compensated counts, making the `clear/green < 0.160` branch unreachable
> - `read_bytes()` return value ignored while `buffer` is uninitialized

latonita:
Merged up to latest dev and addressed the review.

The channel compensation is gone. It was multiplied back into the `uint16_t` fields, so the clear channel wrapped above 6553 raw counts, and it also fed scaled values into the ROHM lux and CCT formulas, which are defined on raw counts. The `*_channel` sensors now publish raw counts, which is what the `#` unit says they are, and the formulas get the values they expect. The factors were the Pimoroni breakout's optical correction rather than a device characteristic, and `glass_attenuation_factor` already covers board specific attenuation.

Also changed:

- `read_bytes()` result is checked, so an I2C error sets a warning instead of publishing an uninitialized buffer. Same for the threshold writes in `setup()`.
- The data ready wait is bounded by elapsed time (one fifth of the integration time) instead of a loop iteration count.
- Thresholds are written LSB first as explicit bytes, which also drops the temporary allocation inside `write_bytes_16()`.
- Lux and CCT maths is single precision now, so `__ieee754_exp` is no longer pulled in.
- `MeasurementTime` and `AdcGain` are `enum class` with prefixed enumerator names, `Bh1745Registers` renamed to `BH1745Registers`, `u_int8_t` replaced, `Readings` members initialized and passed by const reference.
- `dump_config()` logs the illuminance and colour temperature sensors, and the integration time uses the right format specifier.
- Fixed the `BH1745SComponent` typo, added type hints, used the walrus operator in `to_code()`, and moved the test fixtures to dict style packages.

On the two older inline suggestions: the threshold registers were never swapped, the suggested replacement was identical to the code already there, and the read buffer had already been changed to a local array.

Verified with config validation on all four platforms, a compile for esp32-idf, clang-tidy and prek.

The docs PR still describes the old switch entity and needs a rewrite for the output platform.

esphome/components/bh1745/bh1745.cpp
```diff
@@ -0,0 +1,288 @@
...
+    case AdcGain::ADC_GAIN_16X:
+      return 16;
+    default:
+      return 1;
+  }
+}
+
+void BH1745Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up BH1745");
+
```
jesserockz: We have been removing these, the code that calls setup will log it
```suggestion
```

esphome/components/bh1745/bh1745.cpp
```diff
@@ -0,0 +1,288 @@
...
+  uint8_t manuf_id = this->reg((uint8_t) BH1745Registers::MANUFACTURER_ID).get();
+  if (manuf_id != BH1745_MANUFACTURER_ID) {
+    ESP_LOGW(TAG, "Manufacturer ID of BH1745 is not correct! Got 0x%02X, expected 0x%02X", manuf_id,
```
jesserockz: Dont need the component name, its in the tag already
```suggestion
    ESP_LOGW(TAG, "Manufacturer ID not correct! Got 0x%02X, expected 0x%02X", manuf_id,
```

esphome/components/bh1745/bh1745.cpp
```diff
@@ -0,0 +1,288 @@
...
+             BH1745_MANUFACTURER_ID);
+    this->mark_failed();
+    return;
+  }
+
+  SystemControlRegister sys_ctrl;
+  sys_ctrl.raw = this->reg((uint8_t) BH1745Registers::SYSTEM_CONTROL).get();
+  if (sys_ctrl.part_id != BH1745_DEVICE_ID) {
+    ESP_LOGW(TAG, "Device ID of BH1745 is not correct! Got 0x%02X, expected 0x%02X", sys_ctrl.part_id,
```
jesserockz:
```suggestion
    ESP_LOGW(TAG, "Device ID not correct! Got 0x%02X, expected 0x%02X", sys_ctrl.part_id,
```

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Interrupt thresholds are set so the INT pin can never assert — the `output` platform is inert

PR #6337, opened 2024-03-08
Add support for Waveshare 2.13" V2 display

manuelkasper:
# What does this implement/fix?

Adds support for Waveshare e-Paper 2.13" V2 displays (also referred to as B72 in some libraries) as a new `2.13inv2` model. It is based on the [EPD_2in13_V2 code from Waveshare](https://files.waveshare.com/upload/5/50/E-Paper_ESP32_Driver_Board_Code.7z). I have tested this change with the [2.13" e-Paper Cloud Module](https://www.waveshare.com/wiki/2.13inch_e-Paper_Cloud_Module), which incorporates this display along with an ESP32.

While this display type already sort-of worked with the `2.13in-ttgo` model configuration (which uses the same LUT), partial updates were unusable, and the image faded when using the Deep Sleep component (see also https://github.com/esphome/feature-requests/issues/1779).

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/1779

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3669

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example with pins configured for Waveshare 2.13" e-Paper Cloud Module
spi:
  clk_pin: 13
  mosi_pin: 14

display:
  - platform: waveshare_epaper
    model: 2.13inv2
    id: epaper_display
    cs_pin: 15
    dc_pin: 27
    busy_pin: 25
    reset_pin: 26
```

## Checklist:
  - [X] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6357, opened 2024-03-12
Require reset_pin for certain waveshare_epaper models in YAML validation

manuelkasper:

Following up on https://github.com/esphome/esphome/pull/6337#discussion_r1520263776, this adds a YAML validation check that `reset_pin` is set on the two models where this is required.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** https://github.com/esphome/esphome/pull/6337#discussion_r1520263776

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/waveshare_epaper/display.py
+++ b/esphome/components/waveshare_epaper/display.py
@@ -131,6 +131,8 @@ MODELS = {
     "1.54in-m5coreink-m09": ("c", GDEW0154M09),
 }
 
+RESET_PIN_REQUIRED_MODELS = ("2.13inv2", "2.13in-ttgo-b74")
+
 
 def validate_full_update_every_only_types_ac(value):
     if CONF_FULL_UPDATE_EVERY not in value:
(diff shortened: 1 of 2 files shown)


manuelkasper:
@clydebarrow As requested – it indeed makes sense to catch this on the YAML validation level. Hopefully the PR is acceptable; I'm not really a Pythonista, and discovered ESPHome only a few days ago (but totally love it!).

clydebarrow (approved):
I have to say the use of literal strings in the `...REQUIRED_MODELS` list is not ideal, and it would be cleaner to do it more like the full update check - but - I also don't like the use of tuples as the value in the MODELS dict (which are indexed by numbers)- they should be either dicts or classes. And "b" etc should be defined constants, not literals. So there would be quite a bit of refactoring to clean it up.

So I'll approve this now because it implements a needed fix, but if you have the motivation to do a big clean up refactoring that would be a nice follow-up.

manuelkasper:
> I have to say the use of literal strings in the `...REQUIRED_MODELS` list is not ideal, and it would be cleaner to do it more like the full update check

I didn't like it either, but my motivation to touch everything, including a lot of code that I didn't write, plus my very limited Python experience didn't allow for a cleaner solution at the moment. But I'll keep it in mind!

PR #6365, opened 2024-03-13
WireGuard for esp8266

droscy:
# What does this implement?

Add WireGuard support for esp8266 boards.

There are no breaking changes regarding the usage of the component because the configuration schema is still the same, but something changed in the underlying `esp_wireguard` library even for esp32 platform. So I would ask everyone here to do tests also on esp32 boards.

## Types of changes

- [x] Other (see note above)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3708

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
external_components:                                                                                                                                                                                                                          
  - source: github://pr#6365                                                                                                                                                                                                                  
    components: [wireguard]

esp8266:                                                                                                                                                                                                                                      
  board: ...

wireguard:
  # configure as usual
```

## Checklist:


thomas0bernard:
Tested and working on esp32 and esp8266. Sorry for the delay. Amazing work @droscy :raised_hands:

esphome/components/wireguard/wireguard.cpp
```diff
@@ -257,20 +249,13 @@ void Wireguard::start_connection_() {
...
 
   ESP_LOGD(TAG, "starting WireGuard connection...");
-
-  /*
-   * The function esp_wireguard_connect() contains a DNS resolution
-   * that could trigger the watchdog, so before it we suspend (or
-   * increase the time, it depends on the platform) the wdt and
-   * then we resume the normal timeout.
-   */
-  suspend_wdt();
```
thomas0bernard, line 267: I see that you removed the wdt suspension. Any reason for that? I can test on the offending network that ran into this problem later today.

droscy: I removed the watchdog suspension because in `esp_wireguard` 0.4 the DNS resolution is handled asynchronously, see https://github.com/esphome/esphome/pull/6365/commits/5c40618663e8ab8cdc993f59a9334b5c59d9adea and https://github.com/droscy/esp_wireguard/commit/3b4a1b5585f4aaec8d4d1440b60341e5949258d3, but please test the offending network to be sure it works correctly.

thomas0bernard: Wonderful, I figured you might have improved something in `esp_wireguard` to fix that. I'll have access to that network in a few hours and I'll keep you posted, thank you!

thomas0bernard (approved):
Confirmed working on the network with long DNS resolution. Everything looks good to me!

jesserockz (changes requested):
Nice work.

Can you please move the tests to the new test file format that should go in `tests/components/wireguard/` You can delete any wireguard code from the `tests/testX.yaml` files at the same time

jesserockz (approved):

PR #6378, opened 2024-03-15
Fix Nextion set_component_picture call

edwardtfn:

This fixes the call to the Nextion display to change the pic id from a component. It was previously changing the attribute `val`, which is related to something else. In addition, I've changed the parameter for picture_id to be uint_8, as Nextion requires an integer from 0 to 255 on this attribute.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

API inline doc updated.

## Checklist:


kbx81 (approved):
LGTM -- thanks!

PR #6380, opened 2024-03-15
Add support for Waveshare 7.3" ACeP 7-Color display

NeilSCGH:

Add support for Waveshare 7.3" ACeP 7-Color e-paper display.

The display size is 480x800 pixels, and support 7 colors (black, white, red, green, blue, orange and yellow).

I've splitted the buffer in chunks to avoid problems due to heap fragmentation.

I've also optimized the allocated ram so it requires only 144,2kB of ram (each pixel requires 3 bits of ram)

Note: Also works with the Pimoroni Inky Frame 7.3 display (tested by @aeroniemi).

demo

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3806

## Test Environment

- [x] ESP32
- [x] ESP-WROOM-32 (tested by me)
- [x] ESP32 Lolin S2 mini (tested by @francisbesset)
- [x] Pi Pico W (tested by @aeroniemi)
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
font:
  - file: 'epaper/myfont.ttf'
    id: font1
    size: 40

spi:
  clk_pin: 18
  mosi_pin: 23

display:
  - platform: waveshare_epaper
    id: eink_display
    cs_pin: 5
    dc_pin: 26
    busy_pin: 
      number: 27
      inverted: True
    reset_pin: 25
    model: 7.30in-f
    update_interval: 900s
    lambda: |- 
      const auto BLACK   = Color(0,   0,   0,   0);
      const auto RED     = Color(255, 0,   0,   0);
      const auto GREEN   = Color(0,   255, 0,   0);
      const auto BLUE    = Color(0,   0,   255, 0);
      const auto YELLOW  = Color(255, 255, 0,   0);
      const auto ORANGE  = Color(255, 127, 0,   0);
      const auto WHITE   = Color(255, 255, 255, 0);

      it.print(0, 0, id(font1), BLACK, "Hello world!");
      it.print(80, 80, id(font1), BLUE, "Hello world!");
      it.print(160, 160, id(font1), GREEN, "Hello world!");
      it.print(240, 240, id(font1), YELLOW, "Hello world!");
      it.print(320, 320, id(font1), ORANGE, "Hello world!");
      it.print(400, 400, id(font1), RED, "Hello world!");
```

You can already use (and test) this screen by adding this to the esphome code:
```yaml
external_components:
  - source: github://pr#6380
    components: [ waveshare_epaper, display ]
```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs): [esphome/esphome-docs#3805](https://github.com/esphome/esphome-docs/pull/3806)

### DIFF
--- a/esphome/components/display/display_buffer.h
+++ b/esphome/components/display/display_buffer.h
@@ -25,7 +25,7 @@ class DisplayBuffer : public Display {
  protected:
   virtual void draw_absolute_pixel_internal(int x, int y, Color color) = 0;
 
-  virtual void init_internal_(uint32_t buffer_length);
+  void init_internal_(uint32_t buffer_length);
 
   uint8_t *buffer_{nullptr};
 };
--- a/esphome/components/display/display_buffer.h
+++ b/esphome/components/display/display_buffer.h
@@ -25,7 +25,7 @@ class DisplayBuffer : public Display {
  protected:
   virtual void draw_absolute_pixel_internal(int x, int y, Color color) = 0;
 
-  void init_internal_(uint32_t buffer_length);
+  virtual void init_internal_(uint32_t buffer_length);
 
   uint8_t *buffer_{nullptr};
 };
--- a/esphome/components/waveshare_epaper/display.py
+++ b/esphome/components/waveshare_epaper/display.py
@@ -24,9 +24,7 @@ WaveshareEPaper = waveshare_epaper_ns.class_("WaveshareEPaper", WaveshareEPaperB
 WaveshareEPaperBWR = waveshare_epaper_ns.class_(
     "WaveshareEPaperBWR", WaveshareEPaperBase
 )
-WaveshareEPaper7C = waveshare_epaper_ns.class_(
-    "WaveshareEPaper7C", WaveshareEPaperBase
-)
+WaveshareEPaper7C = waveshare_epaper_ns.class_("WaveshareEPaper7C", WaveshareEPaperBase)
 WaveshareEPaperTypeA = waveshare_epaper_ns.class_(
     "WaveshareEPaperTypeA", WaveshareEPaper
 )
--- a/esphome/components/waveshare_epaper/display.py
+++ b/esphome/components/waveshare_epaper/display.py
@@ -24,6 +24,9 @@ WaveshareEPaper = waveshare_epaper_ns.class_("WaveshareEPaper", WaveshareEPaperB
 WaveshareEPaperBWR = waveshare_epaper_ns.class_(
     "WaveshareEPaperBWR", WaveshareEPaperBase
 )
+WaveshareEPaper7C = waveshare_epaper_ns.class_(
+    "WaveshareEPaper7C", WaveshareEPaperBase
+)
 WaveshareEPaperTypeA = waveshare_epaper_ns.class_(
     "WaveshareEPaperTypeA", WaveshareEPaper
 )
@@ -63,6 +66,9 @@ WaveshareEPaper5P8In = waveshare_epaper_ns.class_(
 )
 WaveshareEPaper5P8InV2 = waveshare_epaper_ns.class_(
     "WaveshareEPaper5P8InV2", WaveshareEPaper
+)
+WaveshareEPaper7P3InF = waveshare_epaper_ns.class_(
+    "WaveshareEPaper7P3InF", WaveshareEPaper7C
 )
 WaveshareEPaper7P5In = waveshare_epaper_ns.class_(
     "WaveshareEPaper7P5In", WaveshareEPaper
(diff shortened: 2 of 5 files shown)


francisbesset:
Tested on my Waveshare 7.3" 7 colors and an ESP32 Lolin S2 mini,  it works fine (tested with colored text and colored image).

Thank you very much @NeilSCGH. ❤️

jesserockz (dismissed):
Instead of altering the internal code (and every display using it), just don't use `init_internal_` for this specific display...

The better option is to actually refactor the init_internal_ to potentially use split buffers for all displays, but that should be done it its own PR unrelated to this display though.

NeilSCGH:
Hello @jesserockz,
The reason I renamed the `DisplayBuffer::init_internal_()` function is the following: 
Each class that comes from `class WaveshareEPaper` (`WaveshareEPaperBWR`, `WaveshareEPaperTypeA` and now `WaveshareEPaper7C`) are running `WaveshareEPaperBase::setup_pins_()` during the setup, which run `DisplayBuffer::init_internal_()`
But for WaveshareEPaper7C I need a new init_internal function (that splits the buffer), so I preferred to override the `init_internal_()` function, so the original `WaveshareEPaperBase::setup_pins_()` will run my own `WaveshareEPaper7C::init_internal_()`.
But to override it I needed to make it virtual, and clang doesn't allow virtual function to end with `_`.

The other option was to duplicate the `WaveshareEPaperBase::setup()` function, as it was made for `WaveshareEPaper2P13InV3::setup`, but here the only difference to the original is the first line (the call to init_internal()).

I agree that refactoring the original `DisplayBuffer::init_internal_()` is the best option, but this will still change many things because each line with `this->buffer_[pos]` will not work anymore as there will be many small buffers instead of one big buffer. To make it as transparent as possible, I could also implement a new function to read the buffer so `this->buffer_[pos]` will become `this->read_buffer_(pos)`, or just `this->buffer_(pos)`, that will convert this pos to the right position in the right buffer, what do you think of that idea? I will also have to implement a write_buffer function;
If you validate this idea (`this->buffer_[pos]` to `this->read_buffer_(pos)`) I will make another pull request to create this.


Edit : I just checked and 15 components are using this buffer, so the impact would be very wide, if you have a solution other than read_buffer_ and write_buffer_() to access the splitted buffer I am interested.
The impacted components are ili9xxx, inkplate6, pcd8544, qspi_amoled, ssd1306_base, ssd1322_base, ssd1325_base, ssd1327_base, ssd1331_base, ssd1351_base, st7567_base, st7735, st7789v, st7920 and waveshare_epaper.

NeilSCGH:
Hello @jesserockz,
What do you think of my last message ? I still think the simpliest is to rename `DisplayBuffer::init_internal_()`, maybe I'm wrong about the fact that refactoring the function will impact 15 components, if you have an other solution I am interested.

clydebarrow:
The real problem here is that the waveshare code uses `setup_pins()` to do more than just setup pins. There are three calls in `setup_pins()` that should be moved out to `setup()` (which is already virtual so can be overridden where required.)

Move these lines to `setup()` in `waveshare_epaper.h` (and potentially move the function body to the .cpp file.) Then you can have a custom `setup()` that doesn't call `init_internal_()`. That should mean no changes to unrelated components, just a change to `WaveshareEPaper2P13InV3::setup` which already has a custom `setup()`.

```
  this->init_internal_(this->get_buffer_length_());
  this->spi_setup();
  this->reset_();
```

So `setup()` now becomes:
```
  void setup() override {
    this->init_internal_(this->get_buffer_length_());
    this->setup_pins_();
    this->spi_setup();
    this->reset_();
    this->initialize();
  }
```

Lupla:
Hey, thank you very much!!!
It works:
IMG_9168
Greetings,
Bastian

aeroniemi:
I've also tested this, and it is working correctly with a Pi Pico W and the Pimoroni Inky Frame 7.3

NeilSCGH:
Hello,

I've combined @clydebarrow and @jesserockz advices, to move `WaveshareEPaperBase::init_internal_()` from `WaveshareEPaperBase::setup_pins()` to `WaveshareEPaperBase::setup()`, which is a virtual function. 
Then I've overrided this setup function with `WaveshareEPaper7C::setup()` that calls a custom `WaveshareEPaper7C::init_internal_7c_()` function.

I've also updated `WaveshareEPaper2P13InV3::setup()` that was using the original setup_pins().

sjetzer:
Thank's a lot !!!
Tested with Waveshare e-Paper ESP32 Driver Board 



Greetings.
Seb

NeilSCGH:
Hello @jesserockz,

Do you validate the changes I've done on the code?

clydebarrow (approved):

PR #6383, opened 2024-03-16
Make SPI compile with IDF >= 5.0

HeMan:

SPI compiles with IDF >= 5.0

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/spi_device/spi_device.cpp
```diff
@@ -18,9 +19,9 @@ void SPIDeviceComponent::dump_config() {
   LOG_PIN("  CS pin: ", this->cs_);
   ESP_LOGCONFIG(TAG, "  Mode: %d", this->mode_);
   if (this->data_rate_ < 1000000) {
-    ESP_LOGCONFIG(TAG, "  Data rate: %dkHz", this->data_rate_ / 1000);
+    ESP_LOGCONFIG(TAG, "  Data rate: %" PRId32 "kHz", this->data_rate_ / 1000);
```
clydebarrow, line 22: The PRId32 stuff is ugly and fragile. Why not simply:
```suggestion
    ESP_LOGCONFIG(TAG, "  Data rate: %dkHz", (int)(this->data_rate_ / 1000));
```

esphome/components/spi_device/spi_device.cpp
```diff
@@ -18,9 +19,9 @@ void SPIDeviceComponent::dump_config() {
...
   } else {
-    ESP_LOGCONFIG(TAG, "  Data rate: %dMHz", this->data_rate_ / 1000000);
+    ESP_LOGCONFIG(TAG, "  Data rate: %" PRId32 "MHz", this->data_rate_ / 1000000);
```
clydebarrow, line 24:
```suggestion
    ESP_LOGCONFIG(TAG, "  Data rate: %dMHz", (int)(this->data_rate_ / 1000000));
```

clydebarrow (changes requested):

esphome/components/spi_device/spi_device.cpp
HeMan: That is how it is done in other PRs, https://github.com/esphome/esphome/pull/6382, https://github.com/esphome/esphome/pull/5451, https://github.com/esphome/esphome/pull/5614 to name a few

kbx81: This is the pattern we've been using in a lot of places -- I agree it's quite ugly, but if there's a cleaner way to handle this that keeps IDF 4, IDF 5 and Arduino all happy, I'm happy to use it. 😄

clydebarrow: In this instance the calculated numbers are small integers, printing as decimal. Casting the result of the division to int is portable, and I just don't see the point of reducing the code readability when there's a portable, correct alternative.

In a different context, e.g. where the `uint32_t` value represents an address potentially using the full 32 bits (and usually to be printed as hex), this would not be the right solution, so using `PRIu32` there would be a necessary compromise.

kbx81: Makes sense -- fine by me! 😄

HeMan: Isn't the cast done at runtime while the macro is done at compile time?

clydebarrow: > Isn't the cast done at runtime while the macro is done at compile time?

There is absolutely no change to the generated code, because `int` and `uint32_t` are the same width (32 bits.) The sole effect of the cast in this case is to suppress the error at compile time. Technically it should be cast to `unsigned int` and the format should be `%u` because the result is never negative.

HeMan: With `(int)` it got one op longer but `(unsigned int)` was the same. How ever, there seem to be some changes in between SDK versions since `uint_32` was `unsigned int` on Xtensa and `unsigned long` on RISC-V, but is `unsigned long` on both after ESP-IDF 5.0, https://docs.espressif.com/projects/esp-idf/en/v5.0/esp32/migration-guides/release-5.x/gcc.html#espressif-toolchain-changes

clydebarrow: Given the code is only executed once, speed is hardly a consideration. Code readability and thus maintainability is far more valuable in the long run.

But in any case this is fiddling at the margins - I made a suggestion, you can choose to take it on board or not, I won't care, and the Sun will still rise tomorrow.

clydebarrow (approved):

PR #6388, opened 2024-03-17
Shows component operation time in `ms`

edwardtfn:

This aligns the message with the following one with refence values. Also, we shouldn't be using "seconds" to measure time in this context. If precision is a concern, then we should add that to the message, and that should be the case regardless if using `ms` or `s`,

## Types of changes

- [X] Other

**Related issue or feature (if applicable):** I could find just a mention about this unit mismatch: https://github.com/esphome/esphome/issues/9517

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:

### DIFF
--- a/esphome/core/component.cpp
+++ b/esphome/core/component.cpp
@@ -5,6 +5,7 @@
 #include "esphome/core/helpers.h"
 #include "esphome/core/log.h"
 #include <utility>
+#include <inttypes.h>
 
 namespace esphome {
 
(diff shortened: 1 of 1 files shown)


esphome/core/component.cpp
```diff
@@ -5,6 +5,7 @@
 #include "esphome/core/helpers.h"
 #include "esphome/core/log.h"
 #include <utility>
+#include <inttypes.h>
```
jesserockz:
```suggestion
#include <cinttypes>
```

jesserockz (approved):

PR #5769, opened 2023-11-15
Add Uponor Smatrix component

kroimon:
# What does this implement/fix?

This adds a component to read and control an Uponor Smatrix Base Pulse underfloor heating control system.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3381

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:
```yaml
# Example config.yaml

time:
  - platform: homeassistant
    id: ha_time

uart:
  baud_rate: 19200
  tx_pin: GPIO17
  rx_pin: GPIO16
  debug:

uponor_smatrix:
  address: 0x110B
  time_id: ha_time
  time_device_address: 0xDE13

climate:
  - platform: uponor_smatrix
    address: 0xDE13
    name: "Thermostat Living Room"

sensor:
  - platform: uponor_smatrix
    address: 0xDE13
    humidity:
      name: "Thermostat Humidity Living Room"
    temperature:
      name: "Thermostat Temperature Living Room"
    external_temperature:
      name: "Thermostat Floor Temperature Living Room"
```

## Checklist:
  - [X] The code change is tested and works locally.
  - [X] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [X] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6392, opened 2024-03-18
Fix sending packets to uponor_smatrix devices

kroimon:

While testing the beta, I just noticed a lot of unnecessary zero bytes being sent on the bus before the actual packets.
It turns out that the `vector(size_type count)` constructor does not only reserve a number of elements, but "Constructs the container with `count` default-inserted instances of T."
I therefore replaced this initialization with a call to `reserve` instead.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/uponor_smatrix/uponor_smatrix.cpp
+++ b/esphome/components/uponor_smatrix/uponor_smatrix.cpp
@@ -173,7 +173,9 @@ bool UponorSmatrixComponent::send(uint16_t device_address, const UponorSmatrixDa
     return false;
 
   // Assemble packet for send queue. All fields are big-endian except for the little-endian checksum.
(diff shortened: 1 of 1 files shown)


kroimon:
@jesserockz As I can not comment on https://github.com/esphome/esphome/pull/5769#discussion_r1458292176 anymore:
While this constructor works for reserving storage, it would require using the index operator to fill the vector afterwards instead of just pushing elements to the back. I opted to keep the `push_back` calls for better readability and reserved storage with `reserve` instead.

jesserockz (approved):

PR #6399, opened 2024-03-19
Add support for time entities

jesserockz:

Adds support for `time` entities. This is pretty much the same as `date` entities with the obvious difference.

Related PRs:
- https://github.com/esphome/aioesphomeapi/pull/850
- https://github.com/home-assistant/core/pull/113852
- https://github.com/home-assistant/core/pull/113854

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3748

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

datetime:
  - platform: template
    type: time
    optimistic: true
    name: Time
    id: my_time
    restore_value: true
    on_value:
      - logger.log:
          format: "The time is now: %02d:%02d:%02d"
          args:
            - x.hour
            - x.minute
            - x.second
```

## Checklist:


gabest11:
Can you take a look at #6424 ? I see you also used strptime. This function eliminates 1M devices from ESPHome because it uses std::regex... Smart plugs, etc. With OTA there is max 500k space and even a barebone config is nearly that big.

gabest11:
Then you need to call DateTimeBase::strptime instead, Which does not exists until this is merged. Can't you move it there in your PR? And this can be cancelled then.

gabest11:
Ignore my last comment, putting ifdefs in time.h. strptime will stay there.

jesserockz:
Converting to draft as I need to add the `on_time` trigger

jesserockz:
Depends on #6468

esphome/components/datetime/__init__.py
```diff
@@ -65,7 +77,18 @@ def date_schema(class_: MockObjClass) -> cv.Schema:
 def time_schema(class_: MockObjClass) -> cv.Schema:
     schema = {
         cv.GenerateID(): cv.declare_id(class_),
+        cv.OnlyWith(CONF_MQTT_ID, "mqtt"): cv.declare_id(mqtt.MQTTTimeComponent),
         cv.Optional(CONF_TYPE, default="TIME"): cv.one_of("TIME", upper=True),
+        cv.Inclusive(
+            CONF_ON_TIME,
+            CONF_ON_TIME,
```
kbx81: Is this right? 👀

esphome/components/datetime/time_entity.cpp
```diff
@@ -0,0 +1,156 @@
...
+    this->has_state_ = false;
+    ESP_LOGE(TAG, "Hour must be between 0 and 23");
+    return;
+  }
+  if (this->minute_ < 0 || this->minute_ > 59) {
+    this->has_state_ = false;
+    ESP_LOGE(TAG, "Minute must be between 0 and 59");
+    return;
+  }
+  if (this->second_ < 0 || this->second_ > 59) {
```
kbx81: `hour_`, `minute_` and `second_` are unsigned -- could probably trim out the `< 0` checks.

esphome/components/datetime/time_entity.cpp
```diff
@@ -0,0 +1,156 @@
...
+void TimeCall::validate_() {
+  if (this->hour_.has_value() && (this->hour_ < 0 || this->hour_ > 23)) {
+    ESP_LOGE(TAG, "Hour must be between 0 and 23");
+    this->hour_.reset();
+  }
+  if (this->minute_.has_value() && (this->minute_ < 0 || this->minute_ > 59)) {
+    ESP_LOGE(TAG, "Minute must be between 0 and 59");
+    this->minute_.reset();
+  }
+  if (this->second_.has_value() && (this->second_ < 0 || this->second_ > 59)) {
```
kbx81: Same here

esphome/components/datetime/__init__.py
jesserockz: Yes this is correct, the first one is the config key, and the second is the inclusive key, I just reused the const instead of hard coding another string in there (twice, one for each Inclusive)

jesserockz: I will add the arg name to be clear

esphome/components/datetime/time_entity.cpp
jesserockz: Interesting clang did not pull that up. It usually throws an error for this kind of thing.

kbx81: I thought so, too! haha

kbx81 (approved):

PR #6406, opened 2024-03-20
Add Three New On-States (restore_mode) with Specific Color Temperatures.

TikiBill:
The ALWAYS_ON restore_mode sets a Kauf bulb to a very blue color when it is powered on, which I - along with other members of my household - find displeasing.  This pull request adds three new restore_mode settings which allow the user to be specific about warm, bright, and cool white. This change allows me to have a bulb connected to a regular circuit breaking switch and come on with an acceptable color temperature. 

Good people can disagree about the exact names for each color temperature... I picked "cool" for 6500K since we often fake black-body radiation by adding more blue for the higher color temperatures.

This was tested in the kauf tree and code copied over. A pull request for the documentation will be forthcoming and the change can be found here https://github.com/TikiBill/esphome-docs/blob/new-restore-modes-doc/components/light/index.rst in the restore_mode section.

Thanks.


TikiBill:
The related doc pull request: https://github.com/esphome/esphome-docs/pull/3694

nagyrobi:
You shouldn't need this with `RESTORE_AND_ON`, where you can set your desired color temperature from HA first, and it will be restored (after written to flash, ofcourse) when powering on the bulb.

TikiBill:
> You shouldn't need this with `RESTORE_AND_ON`, where you can set your desired color temperature from HA first, and it will be restored (after written to flash, ofcourse) when powering on the bulb.

An example where this does not work well is when I set all the bulbs to a dim amber/orange color for mood lighting and then someone comes by and turns the lights off with a physical switch. The next day -- when someone wants regular light -- they turn the physical switch back on and the lights are still dim and amber. This change allows me to have a more desirable color temperature and full brightness regardless of the last state of the bulb.

nagyrobi:
Still, this is a niche use case.
For your example a more universal solution is to just set up the light with `ALWAYS_OFF` restore mode, and use the `on_boot` node automation to turn on the light using `light.turn_on` parametrised with the `color_temperature` of your desire, not only the three hardcoded values you chosen.
Moreover, you could make it even more pro by adding a template `number` (with `restore_value` enabled) as a configurable preset for your specific color temperature, which you can use as a lambda in your `light.turn_on`. This way the color temperature at which the light will turn on at physical switch toggling can be entirely customised from the user interface for your family members, considering the specifics of the room too.

kbx81:
As @nagyrobi has sort of alluded to, the behavior you desire isn't intended to be solved with `restore_mode`.  If you want a specific state at boot, you should set that with the `on_boot` automation. What you're implementing here leads us down the path of having a `restore_mode` for every possible color (temperature) which just isn't tenable -- this is where automations like `on_boot` come into play. In addition, you get more flexibility this way -- for example, what if you want a transition time so that the light fades on as opposed to snapping on?

In short, `restore_mode` defines the restore behavior (on/off/previous state) at boot, not the complete state of the light itself.

TikiBill:
Thanks for the feedback and direction. I had thought about the "where does it end?" aspect of this implementation. I'll play around with the the `on_boot` automation, and probably follow that up with a doc pull request outlining what was discussed here.

Thanks!

PR #6401, opened 2024-03-19
AHT10: Use state machine to avoid blocking delay

clydebarrow:
# What does this implement/fix?

The `aht10` sensor uses delays in `update()` to wait for the chip to become ready to read. This causes unacceptable blocking of the `loop()` thread.

This PR replaces the in-line delays with scheduled callbacks to eliminate the thread blocking.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
sensor:
  - platform: aht10
    variant: aht20
    update_interval: 15s
    humidity:
      name: "Humidity"
    temperature:
      name: "Temperature"
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6409, opened 2024-03-20
AHT10: Fix bug

clydebarrow:

#6303 has a subtle error: 

```
- if (this->write(calibrate_cmd, SIZE_CALIBRATE_CMD) != i2c::ERROR_OK) {
+ if (this->write(init_cmd, sizeof(init_cmd)) != i2c::ERROR_OK) {
```

Where `init_cmd` is a pointer, not one of the initialisation data arrays. So it writes 4 bytes instead of 3. That may be affecting some chips.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:
Tested and confirmed as working by 2 others - see https://github.com/esphome/esphome/pull/6401#issuecomment-2010585107

esphome/components/aht10/aht10.cpp
```diff
@@ -52,7 +52,7 @@ void AHT10Component::setup() {
       ESP_LOGCONFIG(TAG, "Setting up AHT10");
   }
 
-  if (this->write(init_cmd, sizeof(init_cmd)) != i2c::ERROR_OK) {
+  if (this->write(init_cmd, sizeof(AHT10_INITIALIZE_CMD)) != i2c::ERROR_OK) {
```
jesserockz: I know it is the same length, but in the past, I requested that this not just use the AHT10 size, but set a new constant that suits both command sizes.
https://github.com/esphome/esphome/pull/5198#discussion_r1286436228

Lemme know if this is just me being too strict...haha

clydebarrow: Well, it's probably better not to assume they are the same length at all.

jesserockz (approved):
Much better =)
Thanks

PR #6298, opened 2024-02-28
Add `on_update` trigger for Project versions

jesserockz:
# What does this implement/fix?

This allows project creators to have a trigger when an end-user has updated from one version of their firmware to a more recent one.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3647

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
esphome:
  name: project-update-test
  project:
    name: jesserockz.test-project
    version: "1.0.0.5"
    on_update:
      - logger.log:
          format: "Project updated from %s to %s"
          args:
            - "version.c_str()"
            - "ESPHOME_PROJECT_VERSION"
```

```
[11:40:03][D][main:058]: Project updated from 1.0.0.4 to 1.0.0.5
```

## Checklist:
  - [ ] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6420, opened 2024-03-22
Use hashing for  `ESPHOME_PROJECT_VERSION`

### LINKED ISSUE #5629: 2024.3.0 breaks existing configs with addition of project version update
### The problem

Starting with 2024.3.0 esphome won't compile most of my firmware that has been working fine for years. Seems like the implementation of adding `on_update` trigger to project version introduced a regression by limiting `PROJECT_VERSION` to 30 characters (29?).

2024.3.0

Home Assistant Add-on

20240.3.3 - N/A

ESP32

### Board

esp32doit-devkit-v1

core

### Example YAML snippet

```yaml
substitutions:
  devicename: big-tank
  devicename_upper: Big Tank
  platform: esp32
  board: esp32doit-devkit-v1
  project_name: excalbian.big-tank
  project_version: "v1.0.1 (${smart_tank_version} ${wifi_version} ${project_info_version} ${text_version} ${restart_version})"
```

```txt
In file included from src/esphome/components/api/api_frame_helper.h:7,
                 from src/esphome/components/api/api_connection.h:3,
                 from src/esphome.h:3,
                 from src/main.cpp:3:
src/esphome/core/base_automation.h: In member function 'virtual void esphome::ProjectUpdateTrigger::setup()':
src/esphome/core/defines.h:5:33: error: initializer-string for array of chars is too long [-fpermissive]
 #define ESPHOME_PROJECT_VERSION "v1.0.1 (tank-0.5.3 wifi-1.1.0 info-1.0.1 text-1.0.1 restart-1.0.1)"
                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/esphome/core/defines.h:5:33: note: in definition of macro 'ESPHOME_PROJECT_VERSION'
 #define ESPHOME_PROJECT_VERSION "v1.0.1 (tank-0.5.3 wifi-1.1.0 info-1.0.1 text-1.0.1 restart-1.0.1)"
                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
*** [.pioenvs/big-tank/src/main.cpp.o] Error 1
========================= [FAILED] Took 122.91 seconds =========================
```

### Additional information

I've been making use for years now of shared YAML (packages) to modularize my esphome configurations across 24+ devices. Where those ${*_version} come from the included packages so it's easy to tell what version of each package was being used in a given firmware.

With the  addition of the [on update trigger for project versions](https://github.com/esphome/esphome/pull/6298) it seems the code assumes / enforces a limit of 30 characters to the version string, causing my firmware to not be able to compile.

nagyrobi:
In https://github.com/esphome/esphome/pull/6298 `ESPHOME_PROJECT_VERSION` got limited to 30 characters.
30 chars get used up very quickly when the user uses some special characters in the version string. 

Eg. `Gázkazán 1-2-4 PID vezérlés` requires double amount of chars: `G\303\241zkaz\303\241n 1-2-4 PID vez\303\251rl\303\251s`.

Each special letter needs 8(!) chars...

This PR deals with this by using a hash of fixed length to store the current project version. 

It also implements preferences saved to storage for the `host` platform, which enabled this PR to be tested readily.

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5629

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml

```

## Checklist:


esphome/core/base_automation.h
```diff
@@ -132,9 +132,9 @@ class ProjectUpdateTrigger : public Trigger<std::string>, public Component {
  public:
   void setup() override {
     uint32_t hash = fnv1_hash(ESPHOME_PROJECT_NAME);
-    ESPPreferenceObject pref = global_preferences->make_preference<char[30]>(hash, true);
-    char previous_version[30];
-    char current_version[30] = ESPHOME_PROJECT_VERSION;
+    ESPPreferenceObject pref = global_preferences->make_preference<char[120]>(hash, true);
```
clydebarrow: 1. The constant 120 should be defined, not repeated in the code;
2. `current_version` should be initialised with `strncpy` to ensure it's null-terminated.
3. Using any fixed size will still be wrong sometimes.
4. Has this been tested on ESP8266? A quick look at the code suggests the *total* space available for preferences is only 128 (or even 64) bytes.

clydebarrow:
Further thoughts:

The only reason to store the project version in flash is as far as I can tell is to compare with the current version and fire the `esphome::project::on_update` trigger. This could be done with a fixed-length hash e.g. SHA-1 which only needs 20 bytes. SHA-256 needs 32 bytes and is probably overkill for this application.

This would change the behaviour slightly - currently the trigger is fired only if the new version is lexically greater than the old (the documentation does not mention this, ~~and also says it depends on the name matching, which I don't see in the code~~.) Using a hash would fire the trigger if the version changed at all.

The hash would be computed at build time so there would be no run-time impact.

nagyrobi:
Yes indeed, using SHA-1 (or even simpler just CRC-32, which only requires 8 hex chars) would suit the purpose well, without loosing the original functionality.

clydebarrow:
> or even simpler just CRC-32

I'd be concerned about too little entropy with that. Reducing the space used from 30 to 20 bytes is already a win.

jesserockz (dismissed):
Changes look fine to me. But please move the `host` preference stuff to its own PR

esphome/core/base_automation.h
```diff
@@ -131,15 +131,14 @@ class LoopTrigger : public Trigger<>, public Component {
...
+    ESPPreferenceObject pref =
+        global_preferences->make_preference<char[VERSION_LENGTH]>(ESPHOME_PROJECT_NAME_HASH, true);
+    char previous_version[VERSION_LENGTH];
     if (pref.load(&previous_version)) {
-      int cmp = strcmp(previous_version, current_version);
-      if (cmp < 0) {
-        this->trigger(previous_version);
-      }
+      if (memcmp(previous_version, current_version, VERSION_LENGTH) != 0)
+        this->trigger(std::string(ESPHOME_PROJECT_VERSION));
```
jesserockz: This is going to trigger with the current version now, instead of the previous version that was stored in flash.

clydebarrow: I don't think so.  Since the size of the stored pref is different it will be ignored.

clydebarrow: Which means it will fail to trigger if the first update after this is merged also has a version number change. Given that the trigger only got introduced last month it's unlikely to affect too many installs, but should strictly be labelled a breaking change.

jesserockz: No, what I mean is the value sent to the `trigger()` function, it is currently the previous firmware version string, pulled from prefs, but with this change, it is going to be the current installed version string (`ESPHOME_PROJECT_VERSION`)

clydebarrow: Well, yes, but:
1. To send the previous version would require an upper bound on the length of the version string, which is a breaking change from prior to the introduction of this trigger, and what this PR is meant to fix, and
2. Why would you send the previous version anyway? I would have thought the current version would be more relevant. Having said that I'm only speculating what the use-case for this trigger is.

jesserockz: You can always get the current version with the ESPHOME_PROJECT_VERSION macro in a lambda action anyway, which is why I added the previous version string as the trigger variable.

I'm happy to remove the trigger variable altogether as it becomes useless without previous version.

But if I might be so honest, who uses a version field with such a long string :speak_no_evil:. Its meant to be a firmware version, not a full description of the device, that is what the project->name is for.

The original description has `Gázkazán 1-2-4 PID vezérlés` as the version, but it really should only be `1.2.4`

jesserockz: The use case is to run some automation when a user updates from an older version to the current firmware. I thought it might be useful to know where you came from to do certain things.

jesserockz: The use case we made this feature for was nabu casa generated firmware to know if a user has a fresh install, or upgraded so we could set some defaults/initial states and not override states that a user previously set. But that doesnt require the old version. It was just extra.

nagyrobi: It's not only my use case: https://github.com/esphome/issues/issues/5629

excalbian: > Well, yes, but:
> 
> 1. To send the previous version would require an upper bound on the length of the version string, which is a breaking change from prior to the introduction of this trigger, and what this PR is meant to fix, and
> 2. Why would you send the previous version anyway? I would have thought the current version would be more relevant. Having said that I'm only speculating what the use-case for this trigger is.

Looking at the responses, I'm reading:
1. The old version number isn't required for the intended use case ([comment](https://github.com/esphome/esphome/pull/6420#discussion_r1538559247))
2. As the author of the linked issue, yes, I confirm that fixing this (reverting to the behavior where project version isn't assumed to be within some upper bound) is valuable 👍 ([comment](https://github.com/esphome/esphome/pull/6420#discussion_r1538624236))

I do see how (strictly speaking) this is a breaking change on 2024.3.0 so yes, I'd be in favor of labeling it as such. I'd also suggest we create a docs PR to better document how this trigger (project version) works.

excalbian:
```suggestion
        this->trigger(std::string(previous_version));
```

I started to [update the documentation](https://github.com/esphome/esphome-docs/pull/3712) and realized that while the current version doesn't make any sense to be passed (the firmware is already 'aware' of it's version 😉  and as @jesserockz pointed out there is the `ESPHOME_PROJECT_VERSION` macro if you want it) there is a use case for knowing what version the device was last running:

* v1 - defaults to all relays off
* v2 - adds a third relay that has inverted output (default is now on for this relay)
* v3 - switches to inverted relay outputs  on all relays (because!)

In v3 (and subsequent versions) I'd like to be able to determine if I'm updating from v1 (where the values stored in flash should be inverted for relays one and two and a default for relay three should be set) or v2 (where only relays one and two should be inverted) or v3+ where the flash values aren't changed. 

I believe we solve this by returning the hash of the `previous_version` - if it's output in the build (or documented in a way easy enough to compute) then anyone aiming for this use case can just as easily compare the hash in their automation (`if (previous_version_hash) == [0x9c, 0xc2, 0xf8, 0x2f, 0xbb, 0x85, 0xdc, 0xa5, 0x3d, 0x92, 0x1c, 0x91, 0x67, 0x93, 0x53, 0x33, 0xa6, 0xf7, 0x2d, 0x37]`).

excalbian:
There is a second use case (and I'm not sure, since it's what I read in @jesserockz's comment as their use case): how can a firmware run some code (eg to set default configuration) when the previous version is not set or if the project name doesn't match? It seems to me that we'd have to add a second trigger for that case, if I'm not mistaken.

As I read the current code - it can only be used when an automation should run if the version hash has changed (and note that currently the proposed change doesn't take into account the project name).

excalbian:
What can we do to get this fix merged in? I did make a suggestion above but if that's not the direction we want to go in, I'm happy to drop it. If we are undecided, I'm also happy to open a new issue / PR and work through it there. As it stands I (and others) can't compile our firmwares with the current ESPHome because of this bug.

jesserockz:
Apologies @excalbian. I will try to at least fix the underlying compilation bug that is stopping you from compiling without changing this features functionality.

Edit: https://github.com/esphome/esphome/pull/6535

excalbian:
Thank you @jesserockz. Unfortunately, I tried with 2024.4.0 and it's still failing:
```
In file included from src/esphome/components/api/api_frame_helper.h:7,
                 from src/esphome/components/api/api_connection.h:3,
                 from src/esphome.h:3,
                 from src/main.cpp:3:
src/esphome/core/base_automation.h: In member function 'virtual void esphome::ProjectUpdateTrigger::setup()':
src/esphome/core/defines.h:6:36: error: initializer-string for array of chars is too long [-fpermissive]
 #define ESPHOME_PROJECT_VERSION_30 "v1.0.1 (tank-0.5.3 wifi-1.1.0 "
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/esphome/core/base_automation.h:137:32: note: in expansion of macro 'ESPHOME_PROJECT_VERSION_30'
     char current_version[30] = ESPHOME_PROJECT_VERSION_30;
                                ^~~~~~~~~~~~~~~~~~~~~~~~~~
Archiving .pioenvs/big-tank/lib2fc/libAsyncTCP-esphome.a
Compiling .pioenvs/big-tank/lib64d/WiFi/WiFiMulti.cpp.o
Compiling .pioenvs/big-tank/lib64d/WiFi/WiFiSTA.cpp.o
*** [.pioenvs/big-tank/src/main.cpp.o] Error 1
========================= [FAILED] Took 193.11 seconds =========================
```

If I'm not mistaken, the fix needs to change [this line](https://github.com/esphome/esphome/blob/655dbc48b5f7e53a4ceee83f4aa159e276d8b864/esphome/core/config.py#L397) to:
```
cg.add_define("ESPHOME_PROJECT_VERSION_30", project_conf[CONF_VERSION][:29]) // null-terminated strings
```

PR #6422, opened 2024-03-23
Add IRK support to ble_rssi

chbmuc:

Most modern smartphones use use a privacy feature called Resolvable Private Addresses to avoid BLE tracking. However if you know the devices "Identity Resolving Key" (IRK), you can check if the generated private MAC address belongs to the device with the IRK.

This PR adds an option to the `ble_rssi` sensor to configure the IRK.

There is no support to obtain the key. You will have to use one of the options described here: https://espresense.com/beacons

Once you have the IRK, configuration is similar to the other methods for the `ble_rssi` binary sensor.

Implementation is almost the same as in `ble_presence` (#6335).

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** feature (as requested in https://github.com/esphome/feature-requests/issues/2285)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation:** esphome/esphome-docs#3707

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example configuration entry
esp32_ble_tracker:

sensor:
  # RSSI based on IRK
  - platform: ble_rssi
    irk: 1234567890abcdef1234567890abcdef
    name: "BLE Tracker with Identity Resolving Key"

```

## Checklist:

### DIFF
--- a/esphome/components/ble_rssi/ble_rssi_sensor.h
+++ b/esphome/components/ble_rssi/ble_rssi_sensor.h
@@ -15,6 +15,10 @@ class BLERSSISensor : public sensor::Sensor, public esp32_ble_tracker::ESPBTDevi
     this->match_by_ = MATCH_BY_MAC_ADDRESS;
     this->address_ = address;
   }
+  void set_irk(uint8_t *irk) {
+    this->match_by_ = MATCH_BY_IRK;
+    this->irk_ = irk;
+  }
   void set_service_uuid16(uint16_t uuid) {
     this->match_by_ = MATCH_BY_SERVICE_UUID;
     this->uuid_ = esp32_ble_tracker::ESPBTUUID::from_uint16(uuid);
@@ -48,6 +52,13 @@ class BLERSSISensor : public sensor::Sensor, public esp32_ble_tracker::ESPBTDevi
     switch (this->match_by_) {
       case MATCH_BY_MAC_ADDRESS:
         if (device.address_uint64() == this->address_) {
+          this->publish_state(device.get_rssi());
+          this->found_ = true;
+          return true;
+        }
+        break;
+      case MATCH_BY_IRK:
+        if (device.resolve_irk(this->irk_)) {
           this->publish_state(device.get_rssi());
           this->found_ = true;
           return true;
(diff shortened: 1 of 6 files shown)


esphome/components/ble_rssi/ble_rssi_sensor.h
```diff
@@ -106,6 +128,43 @@ class BLERSSISensor : public sensor::Sensor, public esp32_ble_tracker::ESPBTDevi
 
   bool check_ibeacon_major_;
   bool check_ibeacon_minor_;
+
+  bool resolve_irk_(uint64_t addr64, const uint8_t *irk) {
```
jesserockz: Since this is exactly the same as `ble_presence`, lets move it to the base `esp32_ble` component for reuse.

jesserockz (changes requested):

esphome/components/ble_rssi/ble_rssi_sensor.h
chbmuc: I decided to move it to `esp32_ble_tracker`. Hope that's okay.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -692,6 +699,45 @@ void ESP32BLETracker::print_bt_device_info(const ESPBTDevice &device) {
...
+    mbedtls_aes_free(&ctx);
+    return false;
+  }
+
+  if (mbedtls_aes_crypt_ecb(&ctx,
+#ifdef USE_ARDUINO
+                            MBEDTLS_AES_ENCRYPT,
+#elif defined(USE_ESP_IDF)
+                            ESP_AES_ENCRYPT,
+#endif
```
jesserockz: I probably should have asked this in the initial PR, but why do you need to differentiate here between arduino and idf? All of the idf values are available when using arduino still...

chbmuc: I didn't know that, but you are right. I've removed it.

jesserockz (approved):

PR #6191, opened 2024-02-06
Add datetime date entities

RFDarter:
# What does this implement/fix?

Adds a datime component. You will be able to pick a datetime, a date, or a time over the web_server frontend and Trigger automations based on the time provided by the user after compile time

Related PRs:
- https://github.com/home-assistant/core/pull/112475
- https://github.com/esphome/aioesphomeapi/pull/824
- https://github.com/home-assistant/core/pull/112560

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3595

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
esphome:
  name: on-time-example

esp32:
  board: esp32dev
  framework:
    type: arduino

logger:
  level: DEBUG

wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_password

web_server:
  js_include: "www.js"
  js_url: ""

time:
  - platform: sntp
    id: my_time
    on_time_sync:
      then:
        - datetime.set:
            id: turn_on_time_datetime
            value: !lambda |-
              return ESPTime::from_epoch_local(id(my_time).now().timestamp + 30).strftime("%H:%M:%S");
        - datetime.set:
            id: turn_off_time_datetime
            value: !lambda |-
              return ESPTime::from_epoch_local(id(my_time).now().timestamp + 40).strftime("%H:%M:%S");

button:
  - platform: template
    name: Set the time inputs to local time + 10 seconds
    on_press: 
      then:
        - datetime.set:
            id: turn_on_time_datetime
            value: !lambda |-
              return ESPTime::from_epoch_local(id(my_time).now().timestamp + 10).strftime("%H:%M:%S");
        - datetime.set:
            id: turn_off_time_datetime
            value: !lambda |-
              return ESPTime::from_epoch_local(id(my_time).now().timestamp + 20).strftime("%H:%M:%S");

switch:
  - platform: template
    name: "Template Switch"
    id: my_switch
    optimistic: yes

datetime:
[...]

PR #6424, opened 2024-03-24
Don't compile strptime unless its required

### LINKED ISSUE #5630: #6191 increased firmware size by 180k
### The problem

I tried every commit until I found the problem ([#6191](https://github.com/esphome/esphome/pull/6191)). It's ESPTime::strptime and std::regex. It pulls in an insane amount of code, and std::regex isn't even used anywhere else. 

If you are testing docker images, it first appeared in 2024.3.0-dev20240311.

It's quite a problem, since 180k is a significant amount for 1M devices.

dev

Home Assistant Add-on

_No response_

ESP32-IDF

### Board

_No response_

_No response_

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

gabest11:

std::regex adds 180k to the firmware even if the datetime component is not included in the build. Ideally, strptime could be implemented with basic string functions, but I'm not sure if the original author (@RFDarter) would like that.

Introduced in #6191

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5630

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [x] BK72xx
- [x] RTL87xx

## Checklist:

### DIFF
--- a/esphome/core/time.cpp
+++ b/esphome/core/time.cpp
@@ -1,3 +1,7 @@
+#ifdef USE_DATETIME_DATE
+#include <regex>
+#endif
+
 #include "helpers.h"
 #include "time.h"  // NOLINT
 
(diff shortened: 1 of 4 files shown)


jesserockz (changes requested):
I am fine with moving this code, but please move it into `datetime_base.h/cpp` so #6399 can also use it without moving it again.

RFDarter:
Why not wrap the methode and the include in ```#ifdef USE_DATETIME ... #endif``` ?

gabest11:
That was my first idea, it also worked. But it seemed out of place in a core file.

esphome/core/time.cpp
```diff
@@ -1,4 +1,6 @@
+#if defined(USE_DATETIME_DATE) || defined(USE_DATETIME_TIME)
```
jesserockz: This can be simplified, otherwise I will need to add USE_DATETIME_DATETIME later when that entity is added too
```suggestion
#ifdef USE_DATETIME
```

gabest11: Okay, I was not sure about that.

jesserockz (approved):

PR #6430, opened 2024-03-26
Add get_size method to QR Code header

deisterhold:

Add get_size method to QR Code so we can position it on screen accordingly.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3711

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [X] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
external_components:
  - source: github://deisterhold/esphome@patch-1
    components: [ qr_code ]
    refresh: 30s

qr_code:
  - id: google
    value: https://google.com/

display:
  - platform: ...
    lambda: |-
      auto scale = 4;
      auto size = id(google).get_size() * scale;
      auto x = (it.get_width() / 2) - (size / 2);
      auto y = (it.get_height() / 2) - (size / 2);
      it.qr_code(x, y, id(google), Color(255,255,255), scale);
```

## Checklist:


esphome/components/qr_code/qr_code.cpp
```diff
@@ -51,5 +51,17 @@ void QrCode::draw(display::Display *buff, uint16_t x_offset, uint16_t y_offset,
...
   }
 }
+
+uint8_t QrCode::get_size() {
+  if (this->needs_update_) {
+    this->generate_qr_code();
+    this->needs_update_ = false;
+  }
+
+  uint8_t size = qrcodegen_getSize(this->qr_);
```
clydebarrow, line 61: Why only 8 bits for the size? What does the library function return?

jesserockz:
```c
int qrcodegen_getSize(const uint8_t qrcode[]) {
	assert(qrcode != NULL);
	int result = qrcode[0];
	assert((qrcodegen_VERSION_MIN * 4 + 17) <= result
		&& result <= (qrcodegen_VERSION_MAX * 4 + 17));
	return result;
}
```

jesserockz: uint8_t would be fine, as its only a single byte from the source data, and the documentation of this function says it will be between 21 and 177

jesserockz (dismissed):

jesserockz (approved):
Unrelated CI failure on yamllint

PR #6431, opened 2024-03-26
Fix editor live validation

### LINKED ISSUE #5533: Editor Code Check not 'Live' anymore [Edit: Being worked on]
### The problem

When editing code, the editor no longer updates live with errors. 
 - You have to save, close and re-open for them to appear.
 

2024.2.0

Home Assistant Add-on

_No response_

ESP8266

### Board

_No response_

_No response_

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

bdraco:

This broke in 25ab6f0297fcf3a21e07644854db5083ba5a0eed because `read_config_file` was no longer being called. It was missed that `read_config_file` will check the `CORE.vscode` and `CORE.ace` globals, and in some cases instead of reading a config file from disk, it will print an event to stdout, and than decode json from stdin to get the contents of the yaml.

```python
def read_config_file(path: str) -> str:
    if CORE.vscode and (
        not CORE.ace or os.path.abspath(path) == os.path.abspath(CORE.config_path)
    ):
```

This PR removes `read_config_file` since its no longer used, which removes more of the global `CORE.ace` and `CORE.vscode` checks.

There are still two places where these global checks remain that should be removed, but they are a bit harder to remove and not contemplated in this PR
```python
esphome/config_validation.py:        not CORE.ace or os.path.abspath(path) == os.path.abspath(CORE.config_path)
esphome/config_validation.py:        not CORE.ace or os.path.abspath(path) == os.path.abspath(CORE.config_path)
```

I added some more typing since it was very hard to follow what was going on as the variables names were not obvious.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5533

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/vscode.py
+++ b/esphome/vscode.py
@@ -1,20 +1,22 @@
+from __future__ import annotations
 import json
 import os
+from io import StringIO
+from typing import Any
 
-from typing import Optional
-
-from esphome.config import load_config, _format_vol_invalid, Config
+from esphome.yaml_util import parse_yaml
+from esphome.config import validate_config, _format_vol_invalid, Config
 from esphome.core import CORE, DocumentRange
 import esphome.config_validation as cv
 
 
-def _get_invalid_range(res: Config, invalid: cv.Invalid) -> Optional[DocumentRange]:
+def _get_invalid_range(res: Config, invalid: cv.Invalid) -> DocumentRange | None:
     return res.get_deepest_document_range_for_path(
         invalid.path, invalid.error_message == "extra keys not allowed"
     )
 
 
-def _dump_range(range: Optional[DocumentRange]) -> Optional[dict]:
+def _dump_range(range: DocumentRange | None) -> dict | None:
     if range is None:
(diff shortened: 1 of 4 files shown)


esphome/vscode.py
```diff
@@ -56,6 +58,25 @@ def add_validation_error(self, range_, message):
...
+        json.dumps(
+            {
+                "type": "read_file",
+                "path": path,
+            }
+        )
+    )
+
+
 def read_config(args):
```
bdraco, line 80: It would be nice to add tests for this, but in its current form its not very testable, and I didn't want to refactor it here.

jesserockz (approved):

PR #6445, opened 2024-03-29
[web_server] add entity sorting for v3

RFDarter:

Adds the possibility to sort the entity displayed on the websever via the config

PR on `webserver`
https://github.com/esphome/esphome-webserver/pull/58

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3741

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: "sorting-test"

esp32:
  board: wemos_d1_mini32

wifi:
  ssid: !secret wifi_ssid

logger:
  level: DEBUG

web_server:

output:
  - platform: template
    id: output_component1
    type: float
    write_action:
      logger.log: "Nothing"

text_sensor:
  - platform: template
    name: "Template Text Sensor"
    lambda: |-
      return {"Hello World"};
    web_server_sorting_weight: -10

text:
  - platform: template
    name: "My Text"
    mode: text
    optimistic: True
    web_server_sorting_weight: -9

select:
  - platform: template
    name: "My select"
    optimistic: true
    options:
      - one
      - two
      - three
    initial_option: two
    web_server_sorting_weight: -5

lock:
  - platform: template
    name: "My Lock"
    optimistic: True
    web_server_sorting_weight: -4

light:
  - platform: monochromatic
    name: "Kitchen Lights"
    output: output_component1
    web_server_sorting_weight: -3

fan:
  - platform: template
    name: "My Fan"
    web_server_sorting_weight: -2

switch:
  - platform: template
    name: "My Switch 1"
    id: template_switch1
    optimistic: true
    web_server_sorting_weight: -7
  - platform: template
    name: "My Switch 2"
    id: template_switch2
    optimistic: true
    web_server_sorting_weight: -8

sensor:
  - platform: template
    name: "My Sensor"
    id: template_sensor1
    lambda: |-
      if (millis() > 10000) {
        return 42.0;
      } else {
        return 0.0;
      }
    update_interval: 60s
    web_server_sorting_weight: -6

climate:
  - platform: bang_bang
    name: Bang Bang Climate
    sensor: template_sensor1
    humidity_sensor: template_sensor1
    default_target_temperature_low: 18°C
    default_target_temperature_high: 24°C
    idle_action:
      - switch.turn_on: template_switch1
    cool_action:
      - switch.turn_on: template_switch2
    heat_action:
      - switch.turn_on: template_switch1
    away_config:
      default_target_temperature_low: 16°C
      default_target_temperature_high: 20°C
    web_server_sorting_weight: 0

cover:
  - platform: template
    name: "My Cover"
    web_server_sorting_weight: -1

button:
  - platform: template
    name: My Button
    web_server_sorting_weight: 4

binary_sensor:
  - platform: template
    name: "My Binary Sensor"
    web_server_sorting_weight: 3

alarm_control_panel:
  - platform: template
    name: "My Alarm Control Panel"
    web_server_sorting_weight: 2

number:
  - platform: template
    name: "My Number 1"
    min_value: 0
    max_value: 100
    step: 1
    mode: box
    optimistic: True
    initial_value: 2
    web_server_sorting_weight: 1

```

## Checklist:


esphome/components/web_server/web_server.cpp
```diff
@@ -1455,6 +1461,17 @@ void WebServer::handleRequest(AsyncWebServerRequest *request) {
 
 bool WebServer::isRequestHandlerTrivial() { return false; }
 
+void WebServer::add_entity_to_sorting_list(EntityBase *entity, float weight, std::string group) {
+  for (auto sorting_entity : this->sorting_entitys_) {
```
jesserockz: Considering the way esphome works and calls this, I don't think you need to check it exists already.

esphome/components/web_server/web_server.h
```diff
@@ -285,12 +291,15 @@ class WebServer : public Controller, public Component, public AsyncWebHandler {
...
 
+  void add_entity_to_sorting_list(EntityBase *entity, float weight, std::string group);
+
  protected:
   void schedule_(std::function<void()> &&f);
   friend ListEntitiesIterator;
   web_server_base::WebServerBase *base_;
   AsyncEventSource events_{"/events"};
   ListEntitiesIterator entities_iterator_;
+  std::vector<SortingEntity *> sorting_entitys_;
```
jesserockz: Instead of storing a vector that needs to be iterated, maybe a std::map with the pointer to the entity as the key and the group/weight as the value. 
This should make the detail_all lookup a bit better too and you don't need the object hash.

kroimon:
Are we sure we want to introduce more top-level properties to all those components?
I wonder if something like the introduction of a `meta` property would make sense that could store properties for other components to use, including the already existing MQTT properties. This would reduce clutter, especially if similar additional properties for other components are necessary in the future.
An example would be:

```yaml
button:
  - platform: template
    name: My Button
    meta:
      web_server:
        sorting_weight: 4
      mqtt:
        qos: 1
        retain: true
        availability:
          topic: mybutton/status
          payload_available: online
          payload_not_available: offline
        state_topic: mybutton/custom_state_topic
        command_topic: mybutton/custom_command_topic
        command_retain: false
```

I understand this would be the scope of a separate PR and would introduce a breaking change for existing MQTT properties, but it might be worth considering before merging this.

RFDarter:
> Are we sure we want to introduce more top-level properties to all those components?
> I wonder if something like the introduction of a `meta` property would make sense that could store properties for other components to use, including the already existing MQTT properties. This would reduce clutter, especially if similar additional properties for other components are necessary in the future.
> An example would be:
> 
> ```yaml
> button:
>   - platform: template
>     name: My Button
>     meta:
>       web_server:
>         sorting_weight: 4
>       mqtt:
>         qos: 1
>         retain: true
>         availability:
>           topic: mybutton/status
>           payload_available: online
>           payload_not_available: offline
>         state_topic: mybutton/custom_state_topic
>         command_topic: mybutton/custom_command_topic
>         command_retain: false
> ```
> 
> I understand this would be the scope of a separate PR and would introduce a breaking change for existing MQTT properties, but it might be worth considering before merging this.

I like that idea!

esphome/components/binary_sensor/__init__.py
```diff
@@ -385,70 +386,79 @@ def validate_click_timing(value):
...
+    .extend(cv.MQTT_COMPONENT_SCHEMA)
+    .extend(
+        {
+            cv.GenerateID(): cv.declare_id(BinarySensor),
+            cv.OnlyWith(CONF_MQTT_ID, "mqtt"): cv.declare_id(
+                mqtt.MQTTBinarySensorComponent
+            ),
+            cv.OnlyWith(CONF_WEB_SERVER_ID, "web_server"): cv.use_id(
+                web_server.WebServer
+            ),
```
jesserockz: This can go inside `WEBSERVER_SORTING_SCHEMA`. The only reason the mqtt ones are in here is because they all have different rhs with different classes.

jesserockz (changes requested):

esphome/components/binary_sensor/__init__.py
RFDarter: Allright, i might find the time tomorrow to change that.

jesserockz (approved):

jesserockz (changes requested):
I realised we need to validate that sorting_weight is only specified when using webserver v3

RFDarter:
> I realised we need to validate that sorting_weight is only specified when using webserver v3

@jesserockz Any hint how we could accomplish that? I was looking into it for some time now and there seems no obvios way of doing this

jesserockz:
CI will fail until #6716 is merged and updated into this PR

jesserockz (approved):

sodabrew:
<3 when I go looking for how to solve a problem and someone has just merged it earlier today! Thank you @RFDarter! (and #6833 is the other half of what I was also hoping to see, nice)

PR #6455, opened 2024-03-31
[homeassistant] Native number entity import and control

landonr:
- new Home Assistant Number which reads the state, min, max and step from Home Assistant. Syncs updates between ESPHome and Home Assistant.
- new Home Assistant Component which makes reading and writing to Home Assistant easier. I use it for covers, lights, numbers, and switches here https://github.com/landonr/esphome-components/tree/main/components/homeassistant_component

I'm adding the number component which I tested using a graphical display menu config

### Home Assistant Component

- next_api_publish_: prevents publishing the next value to the API, used to prevent publishing duplicate values back to the api when the entity changes from the api
- can_update_from_api: updates from the api are ignored for a second after the device publishes to the API, this is used so the updates don't recursively continue
- ignore_api_updates_with_seconds: ignore updates before publishing to update `can_update_from_api`

all of this is used in combination to keep the device's entity in sync with the API as close as possible. This works for most devices like switches, numbers, covers, but it can be a bit buggy for more complicated entities like Lights. 

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3942

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
number:
  - platform: homeassistant
    id: my_number
    name: "My Number"
    entity_id: "number.bedroom_bass"
    on_value:
      then:
        lambda: 'ESP_LOGI("number", "value: %f", x);'

graphical_display_menu:
  id: my_graphical_display_menu
...
  items:
    - type: number
      text: 'My Number'
      format: '%.2f'
      number: my_number
```

## Checklist:


nagyrobi:
I think that the attributes like `min_value`, `max_value` and `step` should be automatically retrieved from HA, and not re-configured here. Someone may configure them differently on the sides which will likely cause malfunctions.

landonr:
> I think that the attributes like `min_value`, `max_value` and `step` should be automatically retrieved from HA, and not re-configured here. Someone may configure them differently on the sides which will likely cause malfunctions.

I think I agree, I'll make the changes and see how it works. thanks!

landonr:
> I think that the attributes like `min_value`, `max_value` and `step` should be automatically retrieved from HA, and not re-configured here. Someone may configure them differently on the sides which will likely cause malfunctions.

works great, nice suggestion thanks

landonr:
@nagyrobi can you take another look

jesserockz:
Requires https://github.com/esphome/esphome/pull/7258 to be merged first

jesserockz:
I removed all of the timeout stuff and also added a simple `value == state` check in the state callback to not do anything if it is the same state.

We can add more protections back in later if needed, but in my basic number testing, it was fine.

jesserockz (dismissed):

jesserockz (approved):

PR #6075, opened 2024-01-10
basic support for nrf52840

tomaszduda23:
# What does this implement/fix?

It is possible to buy cheap [nrf52840 module](https://github.com/joric/nrfmicro/wiki). It would be very convenience to program it using yaml files.
It is proof of concept for support NRF52 on esphome.

It requires:
- https://github.com/esphome/esphome/pull/7049

## How to upload
There are few options:
1. with stlink `python -m esphome  upload tests/test12.2.yaml --device PYOCD`
2. if board have adafruit bootloader image has to build with
```yaml
nrf52:
  board: adafruit_itsybitsy_nrf52840
```
to upload `python -m esphome  upload tests/test12.3.yaml`

3. if board support UF2
```bash
cp ./tests/.esphome/build/nrf52-test-nrf-adafruit/.pioenvs/nrf52-test-nrf-adafruit/zephyr/zephyr.uf2 /media/0321-5432/
``` 

4. BLE OTA with mcuboot `python -m esphome upload tests/test12.2.yaml` to specify mac address of BLE device `python -m esphome upload tests/test12.2.yaml --device AA:BB:CC:11:22:33`

## TODO

**Arduino**
moved to https://github.com/esphome/esphome/pull/6228

**Zephyr**
- [x] GPIO
- [x] CDC DFU (0.8.2 is more stable) - adafruit
- [x] CDC logger
- [x] nrf-sdk based
- [x] UART0 logger
- [ ] UART0 logger tested
- [x] BLE OTA - mcuboot
- [ ] ~CDC OTA - mcuboot recover~ (`FLASH' overflowed by 11320 bytes)
- [x] CDC OTA - mcuboot
- [ ] serial OTA - mcuboot recovery
- [x] watchdog
- [x] BLE logger
- [x] preferences for restore mode
- [x] preferences for adafruit
- [ ] disable update of preferences during OTA - why is it done for other platforms?
- [x] fix arm builds
- [x] add pyocd to upload
- [x] move to main branch for smp
- [x] move to main branch for smpclient
- [x] move debug config to separate component
- [ ] validate prj.conf after compilation to make sure that config was applied correctly
- [ ] GPIO interrupts
- [ ] what to do if image cannot be confirmed during OTA?
- [x] implement get mac
- [x] change name for dfu?
- [x] adc
- [ ] pull up does not work for P0.17 and P0.20
- [ ] change voltage (it works only twice without erase)
- [ ] delete files if app.overlay change
[...]

PR #5586, opened 2023-10-22
[ota] http_request update platform

oarcher:
# What does this implement/fix?

ota in client (http) mode: This allow to do a remote ota when the device is far away, behind a firewall, and only reachable with mqtt. (For ex in another house).

As the ota might be triggered by mqtt, a long time deep sleep can be defined. 

original component from https://github.com/oarcher/piotech/tree/main/components/ota_http

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <https://github.com/esphome/feature-requests/issues/801>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3291

## Test Environment

- [x] ESP32
- [x] ESP32 IDF (v4 & v5)
- [x] ESP8266
- [x] RP2040

### Http/Https support matrix

|   | ESP32  | ESP32 IDF   | ESP8266  | RP2040  |
|---|---|---|---|---|
|  HTTP | Yes   | Yes  | Yes  | Yes  |
|  HTTPS | Yes | Yes  |  Yes|  Yes |

## Example entry for `config.yaml`:

# Example config.yaml
```yaml
ota_http:


button:
  - platform: template
    name: "Firmware update"
    on_press:
      then:
        - ota_http.flash:
            url: http://example.com/firmware.bin
        - logger.log: "This message should be not displayed(reboot)"
```

## Checklist:
  - [X] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6459, opened 2024-04-02
Separate `OTABackend` from OTA component

kbx81:

Splits the OTABackend out from the OTA component so that it can be more easily used by other (future) OTA platforms.

#6475 should be merged _before_ this PR or validation won't enforce the presence of a platform under `ota:`, potentially rendering devices un-OTA-able!

#6623 should also be merged before this PR as it permits more user-friendly error messages should the required `esphome` platform not be present when components which depend on it (`safe_mode` `button` or `switch`) are present.

## Breaking Change Notes

This change only impacts the OTA configuration as it exists in YAML; it does not affect ESPHome's original OTA functionality. Existing configuration variables placed under `ota:` must be moved into the `esphome` OTA platform. For example, if your original configuration looked like this:
```yaml
ota:
  password: "xxxxxxxxxxxxxxxx"
  num_attempts: 3
  safe_mode: no
```
...you now need to arrange it like this:
```yaml
ota:
  - platform: esphome
    password: "xxxxxxxxxxxxxxxx"
    num_attempts: 3
    safe_mode: no
```
No other changes are required to continue using OTA and you will be able to update your devices in the same manner as before.

## Types of changes

- [x] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable): esphome/esphome-docs#3782**

## Test Environment

- [x] ESP32
- [x] ESP32 IDF 4 + 5
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs). **N/A**


oarcher:
It would be nice to make the function `std::unique_ptr<ota::OTABackend> make_ota_backend()`  become a public method of `OTAESPHomeComponent`, so other platforms can get the backend easily.

in https://github.com/esphome/esphome/pull/5586, I had to copy/paste the whole function to get the backend...

kbx81:
> It would be nice to make the function `std::unique_ptr<ota::OTABackend> make_ota_backend()` become a public method of `OTAESPHomeComponent`, so other platforms can get the backend easily.
> 
> in #5586, I had to copy/paste the whole function to get the backend...

I had considered doing that -- I can add that change here. 😄

oarcher:
I've tried to turn `ota_http` in a platform component, and it worked ! 

My yaml was:
```yaml
ota:
   - platform: esphome
     password: "xxxxxxxxxxxxxxxx"
     num_attempts: 3
     safe_mode: no
   - platform: ota_http
     safe_mode: false
     exclude_certificate_bundle: true
     watchdog_timeout: 20s
     timeout: 4min
     max_url_length: 128
     force_update: true
 ```
 
 This was just a local quick and dirty test, I've not refactored it yet to `http_request`.

**Comment by kbx81** [comment]
Thank you for confirming, @oarcher! 🍻

**Comment by kbx81** [comment]
@oarcher I've moved `make_ota_backend` into...the OTA backend. 😅 Should help eliminate duplicate code. Feel free to try it out!

**Comment by oarcher** [comment]
ok, I've been able to remove `make_ota_backend` implementation from `ota_http`: works well, with less code!

File: esphome/components/esphome/ota/__init__.py [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+    CONF_TRIGGER_ID,
+    CONF_VERSION,
+    KEY_PAST_SAFE_MODE,
+)
+from esphome.core import CORE, coroutine_with_priority
+
+
+CODEOWNERS = ["@esphome/core"]
+AUTO_LOAD = ["md5", "socket"]
+DEPENDENCIES = ["network", "ota"]
```
jesserockz: Not required as this is an ota platform so it will always be loaded
```suggestion
DEPENDENCIES = ["network"]
```

esphome/components/esphome/ota/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+                cv.GenerateID(CONF_TRIGGER_ID): cv.declare_id(OTAESPHomeErrorTrigger),
+            }
+        ),
+        cv.Optional(CONF_ON_PROGRESS): automation.validate_automation(
+            {
+                cv.GenerateID(CONF_TRIGGER_ID): cv.declare_id(
+                    OTAESPHomeProgressTrigger
+                ),
+            }
+        ),
```
jesserockz: I feel like these could all stay in the base `ota` and be in a BASE_OTA_SCHEMA and a base c++ class that all platforms should implement.

esphome/components/esphome/ota/ota_esphome.h
```diff
@@ -2,48 +2,19 @@
...
-  OTA_RESPONSE_ERROR_WRONG_CURRENT_FLASH_CONFIG = 0x86,
-  OTA_RESPONSE_ERROR_WRONG_NEW_FLASH_CONFIG = 0x87,
-  OTA_RESPONSE_ERROR_ESP8266_NOT_ENOUGH_SPACE = 0x88,
-  OTA_RESPONSE_ERROR_ESP32_NOT_ENOUGH_SPACE = 0x89,
-  OTA_RESPONSE_ERROR_NO_UPDATE_PARTITION = 0x8A,
-  OTA_RESPONSE_ERROR_MD5_MISMATCH = 0x8B,
-  OTA_RESPONSE_ERROR_RP2040_NOT_ENOUGH_SPACE = 0x8C,
-  OTA_RESPONSE_ERROR_UNKNOWN = 0xFF,
-};
+namespace ota_esphome {
```
jesserockz: Namespaces need to match the platform
```suggestion
namespace esphome {
```

esphome/components/esphome/ota/ota_esphome.h
```diff
@@ -2,48 +2,19 @@
...
 
-enum OTAState { OTA_COMPLETED = 0, OTA_STARTED, OTA_IN_PROGRESS, OTA_ERROR };
+enum OTAESPHomeState { OTA_COMPLETED = 0, OTA_STARTED, OTA_IN_PROGRESS, OTA_ERROR };
 
-/// OTAComponent provides a simple way to integrate Over-the-Air updates into your app using ArduinoOTA.
-class OTAComponent : public Component {
+/// OTAESPHomeComponent provides a simple way to integrate Over-the-Air updates into your app using ArduinoOTA.
+class OTAESPHomeComponent : public Component {
```
jesserockz: This one is optional, but I think having them this way reads better
```suggestion
class ESPHomeOTAComponent : public Component {
```

jesserockz (changes requested):
A couple of points below I thought I would submit before going through the rest

esphome/components/esphome/ota/__init__.py
kbx81: 🤦🏻‍♂️

esphome/components/esphome/ota/__init__.py
kbx81: Solid agree. I'll rework this.

esphome/components/esphome/ota/ota_esphome.h
kbx81: Works for me!

esphome/components/esphome/ota/__init__.py
kbx81: Did some rearranging in [`8426084` (#6459)](https://github.com/esphome/esphome/pull/6459/commits/8426084ebfe523066855c20dcd636b0dfb04bc5d) and it's at least working...not 100% sure it's what you had in mind but have a look! 😄

kbx81: An outstanding question I have is about `USE_OTA_STATE_CALLBACK` -- it seems it's always set when `defines.h` is included...which would seem to make it somewhat...useless? There may be some history there that I'm not aware of and/or maybe I just don't understand its intended use. Regardless, just something I wanted to call out.

oarcher: If using `esp32_ble_tracker`, `USE_OTA_STATE_CALLBACK` is forced: https://github.com/esphome/esphome/blob/19fdef7bca4b165eb69ce9f8633d421aaf78e74a/esphome/components/esp32_ble_tracker/__init__.py#L275 

So the tracker is able to add a callback when ota starts, to disable the scan.

If `esp32_ble_tracker` is not used, `USE_OTA_STATE_CALLBACK` should not be in `define.h`.

Does the `esp32_ble_tracker` do that to release memory for the OTA ? If so, https://github.com/esphome/esphome/pull/5586 should raise an `ota::OTA_STARTED` event *before* the http request, because this one take much memory.

kbx81: More broadly, it _appears_ that `USE_OTA_STATE_CALLBACK` is not to be defined unless one or more of the callbacks (actions) is defined; `esp32_ble_tracker` just hooks into this.

But it's defined in `defines.h`, so it ends up being defined wherever that gets included...which I think may be an error.

esphome/components/ota/ota_backend.h
```diff
@@ -15,5 +52,80 @@ class OTABackend {
...
+};
+
+std::unique_ptr<ota::OTABackend> make_ota_backend();
+
+///
+/// Automations
+///
+
+#ifdef USE_OTA_STATE_CALLBACK
+class OTAStateChangeTrigger : public Trigger<OTAState> {
```
jesserockz: Why did these move out of the automation.h file?

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -58,7 +58,7 @@ void ESP32BLETracker::setup() {
   this->scanner_idle_ = true;
 
 #ifdef USE_OTA
-  ota::global_ota_component->add_on_state_callback([this](ota::OTAState state, float progress, uint8_t error) {
+  esphome::global_ota_component->add_on_state_callback([this](ota::OTAState state, float progress, uint8_t error) {
```
jesserockz: This should not be specific to the esphome ota component.

esphome/components/ota/__init__.py
```diff
@@ -1,61 +1,45 @@
...
-    KEY_PAST_SAFE_MODE,
-    CONF_VERSION,
-)
-from esphome.core import CORE, coroutine_with_priority
+
+from esphome.const import CONF_ESPHOME, CONF_OTA, CONF_PLATFORM, CONF_TRIGGER_ID
 
 CODEOWNERS = ["@esphome/core"]
+AUTO_LOAD = ["md5"]
 DEPENDENCIES = ["network"]
```
jesserockz: Does the backend actually still have a dependency on network still?

jesserockz (changes requested):

esphome/components/ota/ota_backend.h
kbx81: That's just where I pasted them. 😅 We can move them back into `automation.h`; it shouldn't cause any issues.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
kbx81: Ah! I think this will need a little refactoring. I'm on it. 😄

esphome/components/ota/__init__.py
kbx81: Looks like it does not. I'll remove this

esphome/components/ota/ota_backend.h
kbx81: Done

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
kbx81: Ok, I think [`4f8dd25`](https://github.com/esphome/esphome/pull/6459/commits/4f8dd25478df50150e3c01e8a77195387a8dd5c4) should take care of this.

esphome/components/esphome/ota/__init__.py
kbx81: ..I should've really read the comment at the top of `defines.h`...

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
kbx81: Done more better in [`2da0bca`](https://github.com/esphome/esphome/pull/6459/commits/2da0bcadc70093b9fc234baad5ee88ef104596b3)...I think 😅

kbx81: Done even more better in [`8dd03a3`](https://github.com/esphome/esphome/pull/6459/commits/8dd03a3dac480e8eda0c8d8e31d1bf62ae23efae)

tomaszduda23:
I just merged the change to https://github.com/esphome/esphome/pull/6075. It simplified the implementation a lot :+1:

jesserockz (approved):

PR #6463, opened 2024-04-02
Add Several I2S Configuration Options and VolumeSetAction for Speakers

kahrendt:

Adds several features and simplifies some of the ``i2s_audio`` code:

- Ability to select primary or secondary mode for an I2S audio parent config (default matches current behavior as the primary controller)
- Uses ``i2s_set_clk`` function to automatically handle splitting the expected mono channel audio into stereo audio instead of doing so manually
- Removes the commented out ``stop_`` function in the i2s_speaker header file
- Adds ``bits_per_sample`` configuration option for I2S speakers. The code uses ``i2s_write_expand`` function to automatically convert the expected 16 bits per sample audio into the appropriate format.
- Adds a volume attribute to the parent ``speaker`` component along with a corresponding ``speaker.volume_set`` action. The ``i2s_speaker`` uses this attribute to scale the audio volume.
- Defines the TDM parameters in ``i2s_driver_config_t`` on supported variants, eliminating several compilation warnings.
- Adds component tests for ``i2s_audio``.

The default behavior shouldn't be different with these modifications, so it isn't a breaking change. I tested these changes on an ATOM Echo without modifying the YAML with no issues.

The volume attribute and set action could be moved into the ``i2s_speaker`` component, as that is where the volume scaling actually happens. I moved the ``bits_per_sample`` codegen into the parent component. The ``i2s_microphone`` and ``i2s_speaker`` components import it to avoid code duplication.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3739

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
i2s_audio:
  - id: i2s_output
    i2s_lrclk_pin: GPIO7
    i2s_bclk_pin: GPIO8
    i2s_mclk_pin: GPIO9
    i2s_mode: secondary

speaker:
  - platform: i2s_audio
    id: xiao_speaker
    i2s_audio_id: i2s_output
    dac_type: external
    i2s_dout_pin: GPIO43
    mode: stereo
    bits_per_sample: 32bit

wifi:
  on_connect:
    - speaker.volume_set: 0.1



```

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/__init__.py
+++ b/esphome/components/i2s_audio/__init__.py
@@ -39,8 +39,8 @@ I2SAudioOut = i2s_audio_ns.class_(
 )
 i2s_mode_t = cg.global_ns.enum("i2s_mode_t")
 I2S_MODE_OPTIONS = {
-    CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,  // NOLINT
-    CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,  // NOLINT
+    CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,  # NOLINT
+    CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,  # NOLINT
 }
 i2s_bits_per_sample_t = cg.global_ns.enum("i2s_bits_per_sample_t")
 BITS_PER_SAMPLE = {
--- a/esphome/components/i2s_audio/__init__.py
+++ b/esphome/components/i2s_audio/__init__.py
@@ -22,6 +22,10 @@ CONF_I2S_MCLK_PIN = "i2s_mclk_pin"
 CONF_I2S_BCLK_PIN = "i2s_bclk_pin"
 CONF_I2S_LRCLK_PIN = "i2s_lrclk_pin"
 
+CONF_I2S_MODE = "i2s_mode"
+CONF_PRIMARY = "primary"
+CONF_SECONDARY = "secondary"
+
 CONF_I2S_AUDIO = "i2s_audio"
 CONF_I2S_AUDIO_ID = "i2s_audio_id"
 
@@ -31,6 +35,11 @@ I2SAudioIn = i2s_audio_ns.class_("I2SAudioIn", cg.Parented.template(I2SAudioComp
 I2SAudioOut = i2s_audio_ns.class_(
     "I2SAudioOut", cg.Parented.template(I2SAudioComponent)
 )
+i2s_mode_t = cg.global_ns.enum("i2s_mode_t")
+I2S_MODE_OPTIONS = {
+    CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,
+    CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,
+}
 
 # https://github.com/espressif/esp-idf/blob/master/components/soc/{variant}/include/soc/soc_caps.h
 I2S_PORTS = {
@@ -46,6 +55,9 @@ CONFIG_SCHEMA = cv.Schema(
         cv.Required(CONF_I2S_LRCLK_PIN): pins.internal_gpio_output_pin_number,
         cv.Optional(CONF_I2S_BCLK_PIN): pins.internal_gpio_output_pin_number,
         cv.Optional(CONF_I2S_MCLK_PIN): pins.internal_gpio_output_pin_number,
+        cv.Optional(CONF_I2S_MODE, default=CONF_PRIMARY): cv.enum(
+            I2S_MODE_OPTIONS, lower=True
+        ),
     }
 )
 
@@ -68,6 +80,7 @@ async def to_code(config):
     var = cg.new_Pvariable(config[CONF_ID])
     await cg.register_component(var, config)
 
+    cg.add(var.set_i2s_mode(config[CONF_I2S_MODE]))
     cg.add(var.set_lrclk_pin(config[CONF_I2S_LRCLK_PIN]))
     if CONF_I2S_BCLK_PIN in config:
         cg.add(var.set_bclk_pin(config[CONF_I2S_BCLK_PIN]))
--- a/esphome/components/i2s_audio/__init__.py
+++ b/esphome/components/i2s_audio/__init__.py
@@ -22,6 +22,8 @@ CONF_I2S_MCLK_PIN = "i2s_mclk_pin"
 CONF_I2S_BCLK_PIN = "i2s_bclk_pin"
 CONF_I2S_LRCLK_PIN = "i2s_lrclk_pin"
 
+CONF_BITS_PER_SAMPLE = "bits_per_sample"
+
 CONF_I2S_MODE = "i2s_mode"
 CONF_PRIMARY = "primary"
 CONF_SECONDARY = "secondary"
@@ -40,6 +42,13 @@ I2S_MODE_OPTIONS = {
     CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,
     CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,
 }
+i2s_bits_per_sample_t = cg.global_ns.enum("i2s_bits_per_sample_t")
+BITS_PER_SAMPLE = {
+    16: i2s_bits_per_sample_t.I2S_BITS_PER_SAMPLE_16BIT,
+    32: i2s_bits_per_sample_t.I2S_BITS_PER_SAMPLE_32BIT,
+}
+
+_validate_bits = cv.float_with_unit("bits", "bit")
 
 # https://github.com/espressif/esp-idf/blob/master/components/soc/{variant}/include/soc/soc_caps.h
 I2S_PORTS = {
--- a/esphome/components/i2s_audio/__init__.py
+++ b/esphome/components/i2s_audio/__init__.py
@@ -39,8 +39,8 @@ I2SAudioOut = i2s_audio_ns.class_(
 )
 i2s_mode_t = cg.global_ns.enum("i2s_mode_t")
 I2S_MODE_OPTIONS = {
-    CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,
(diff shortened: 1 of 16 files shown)


esphome/components/i2s_audio/__init__.py
```diff
@@ -31,6 +37,18 @@
 I2SAudioOut = i2s_audio_ns.class_(
     "I2SAudioOut", cg.Parented.template(I2SAudioComponent)
 )
+i2s_mode_t = cg.global_ns.enum("i2s_mode_t")
+I2S_MODE_OPTIONS = {
+    CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,
+    CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,
```
jesserockz:
```suggestion
    CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,  // NOLINT
    CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,  // NOLINT
```

jesserockz:
CI updated in #6464

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -40,30 +40,47 @@ void I2SAudioSpeaker::player_task(void *params) {
...
-      .dma_buf_len = 128,
-      .use_apll = false,
-      .tx_desc_auto_clear = true,
-      .fixed_mclk = I2S_PIN_NO_CHANGE,
-      .mclk_multiple = I2S_MCLK_MULTIPLE_DEFAULT,
-      .bits_per_chan = I2S_BITS_PER_CHAN_DEFAULT,
+    .mode = (i2s_mode_t) (this_speaker->parent_->get_i2s_mode() | I2S_MODE_TX),
+    .sample_rate = 16000,
+    .bits_per_sample = this_speaker->bits_per_sample_,
+    .channel_format = I2S_CHANNEL_FMT_RIGHT_LEFT,
```
gnumpi:
```suggestion
    .channel_format = I2S_CHANNEL_FMT_ONLY_RIGHT,
```
If you set the channel_format here there is no need of calling i2s_set_clk

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -40,30 +40,47 @@ void I2SAudioSpeaker::player_task(void *params) {
...
 
-  esp_err_t err = i2s_driver_install(this_speaker->parent_->get_port(), &config, 0, nullptr);
-  if (err != ESP_OK) {
+  esp_err_t err_driver = i2s_driver_install(this_speaker->parent_->get_port(), &config, 0, nullptr);
+
+  // Specify incoming audio stream is mono channel so it gets converted automatically when written to DMA buffer
+  uint32_t bits_cfg = (this_speaker->bits_per_sample_ << 16) | this_speaker->bits_per_sample_;
+  esp_err_t err_clk = i2s_set_clk(this_speaker->parent_->get_port(), 16000, bits_cfg, I2S_CHANNEL_MONO);
+
+  if ((err_driver != ESP_OK) || (err_clk != ESP_OK)) {
```
gnumpi:
```suggestion
  esp_err_t err_driver = i2s_driver_install(this_speaker->parent_->get_port(), &config, 0, nullptr);
  if (err_driver != ESP_OK) {
```

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -40,30 +40,47 @@ void I2SAudioSpeaker::player_task(void *params) {
...
     event.type = TaskEventType::WARNING;
-    event.err = err;
+    if (err_driver != ESP_OK) {
+      event.err = err_driver;
+    } else {
+      event.err = err_clk;
+    }
```
gnumpi:
```suggestion
    event.err = err_driver;
```

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -107,11 +124,24 @@ void I2SAudioSpeaker::player_task(void *params) {
...
+
+      esp_err_t err;
+      if (this_speaker->bits_per_sample_ == I2S_BITS_PER_SAMPLE_16BIT) {
+        err = i2s_write(this_speaker->parent_->get_port(), &volume_adjusted_sample, sizeof(volume_adjusted_sample),
+                        &bytes_written, (10 / portTICK_PERIOD_MS));
+      } else {
+        err = i2s_write_expand(this_speaker->parent_->get_port(), &volume_adjusted_sample,
+                               sizeof(volume_adjusted_sample), I2S_BITS_PER_SAMPLE_16BIT,
+                               this_speaker->bits_per_sample_, &bytes_written, (10 / portTICK_PERIOD_MS));
+      }
```
gnumpi, line 136: I suggest writing the adjusted samples to a buffer first and avoid calling i2s_write for each sample. Maybe even include an if condition to skip the sample changes if no volume change is required.

gnumpi (commented):
Nice improvements, just a few suggestions...

gnumpi:
What do you think of also moving the volume multiplier to the microphone component? For 32-bit samples, integrating the multiplier within the quantization process could prevent amplification of quantization errors, which occur if the multiplier is applied afterward.

gnumpi:
Not directly related to this PR but, relevant for the microphone component.
Right now the microphone component won't work with the current version of micro_wake_word. The component waits 100ms for receiving its requested 1024 bytes. But, in its current implementation  mWW only processes one frame per loop, so it consumes only 320 bytes in each loop step and the buffer fills up.  To make it compatible with the current version of mWW the maximal waiting time can be 10ms, I suggest changing the following line to a waiting time to max 1ms. https://github.com/esphome/esphome/blob/950b059e99df35e51b99d19e369edf61d933492e/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp#L116

Sorry, just realised that this PR targets the speaker component only and not the microphone. I will create a new PR for the mentioned issue.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: Thanks!

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: This was my initial approach, but it caused the audio output to stutter. I was attempting to write the entire buffer in one command, but I ?think? the problem is that if there isn't enough room in the DMA buffer, then the write operation times out causing the stutter. Writing it one sample at a time seems to avoid this issue, but there is probably a better approach. I did notice that the ESP32-audioI2S library seems to only [write one sample at a time], (https://github.com/schreibfaul1/ESP32-audioI2S/blob/fc5bba029bb8a87fa60b7f545f95b49b6799225e/src/Audio.cpp#L5025), so this approach is used elsewhere.

Edit: Do you have any specific implementation ideas on how to better handle writing more than one sample at a time that still avoids stuttering? I didn't try anything between writing one sample and writing all the samples.

kahrendt:
> What do you think of also moving the volume multiplier to the microphone component? For 32-bit samples, integrating the multiplier within the quantization process could prevent amplification of quantization errors, which occur if the multiplier is applied afterward.

This seems like it could be very beneficial, but it would need a lot of experimentation to ensure it is actually improving quality, or at the very least, not making anything worse. I won't have time to test this out for awhile, so I'd prefer to leave it out of this PR.

kahrendt:
> Not directly related to this PR but, relevant for the microphone component. Right now the microphone component won't work with the current version of micro_wake_word. The component waits 100ms for receiving its requested 1024 bytes. But, in its current implementation mWW only processes one frame per loop, so it consumes only 320 bytes in each loop step and the buffer fills up. To make it compatible with the current version of mWW the maximal waiting time can be 10ms, I suggest changing the following line to a waiting time to max 1ms.
> 
> https://github.com/esphome/esphome/blob/950b059e99df35e51b99d19e369edf61d933492e/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp#L116
> 
> Sorry, just realised that this PR targets the speaker component only and not the microphone. I will create a new PR for the mentioned issue.

The 100 ms is the maximum wait time. If the I2S DMA buffer has 1024 samples already, then this command takes at most 1 ms. If it doesn't have that many samples, it will wait until it's full (up to a max of 100 ms) before completing. Since the ``micro_wake_word`` component has it's own ring buffer, it will process that 32 ms of audio (plus any left over samples in the buffer that it hasn't processed) in 20 ms strides, once per loop. 

I did log the timing of this command specifically, and if I am running two of the slower model architectures at the same time, then reading takes at most 1 ms (usually 0 ms). If I only run one model using the new, faster architecture, then the read command waits around 6 ms each time. Either way, the entire ``micro_wake_word`` component keeps up and processes 1000 ms of audio in approximately 1000 ms.

gnumpi:
Apologies once more for bringing up the microphone-related concerns in this PR. Let's continue our discussion on this topic e.g. on discord.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
gnumpi: Which buffer length did you choose? Probably it shouldn't be larger than dma_buf_len, so in the current settings 256 x 16bit samples. But ok, the current implementation also uses a per sample approach for filling up the two-channel samples, so maybe its best to leave it as you suggested for now.

kahrendt: I used ``data_event.len`` as the length, which should at most be 512x16 bit samples, so that was probably the issue. I'll mark this as resolved, but I do believe a future PR could handle this better after thorough testing.

kahrendt:
I'm doing a deeper rework of the speaker output, so I'm closing this PR to avoid potentially introducing breaking changes in the future.

PR #6477, opened 2024-04-04
md5 checksum of firmware, and 'esphome publish' command

oarcher:

This PR compute the md5 of the firmware, and add an `esphome publish config.yaml --directory /path/to` that moves the firmware files to `/path/to/device_name`.

Using Home Assistant esphome addon, `/path/to` can be `/config/ww`, so the firmware file can be made available by the integrated http server.

This work is related to https://github.com/esphome/esphome/pull/5586

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/__main__.py
+++ b/esphome/__main__.py
@@ -6,6 +6,8 @@ import os
 import re
 import sys
 import time
+import hashlib
+import gzip
 from datetime import datetime
 
 import argcomplete
@@ -37,7 +39,7 @@ from esphome.const import (
     SECRETS_FILES,
 )
 from esphome.core import CORE, EsphomeError, coroutine
-from esphome.helpers import indent, is_ip_address
+from esphome.helpers import indent, is_ip_address, copy_file_if_changed
 from esphome.util import (
     run_external_command,
     run_external_process,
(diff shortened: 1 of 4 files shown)


oarcher:
Currently, I've added `CORE.firmware_md5`, `CORE.firmware_gz`, `CORE.firmware_gz_md5`, but it might not be a good idea.

I've done that because `espota2.py` compute md5 by his own, and I've refactored it to use the md5 files and compressed file computed before, with a modified `esphome compile`.

Those changes can be reverted, and at the most simple, I can just add `esphome publish` that copy the firmware, and create the md5 file and the compressed file in the **destination** directory.

balloob:
We should not add this feature. It's a very niche feature. 

It's already possible to push OTA from ESPHome Dashboard in Home Assistant to ESPHome devices. It is also possible to upload OTA via the ESPHome web server dashboard.

dougiteixeira:
#5586 has been merged, does this implementation have a chance of being accepted?

This would be a very useful feature for using the OTA's http_request platform.

BugsBunny1403:
Hello, I'm here to support / relaunch this request to use the ota_http_request plugin. Is it possible to have a visual option in the ESPHOME add-on on HA to simplify the multiple update process?

ximex:
i would recommend sha1/sha256 instead or additionally to md5

PR #6478, opened 2024-04-04
Update sm2135.cpp / fixed lights flickering

matika77:

fixed sm2135_set_high_ which caused flicker while color changes.

the pin was set to mode output and high (strong drive) before going to open drain / pull up. this caused transmission errors (maybe just on certain hardware with open drain configuration).

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
sm2135:
  data_pin: GPIO12
  clock_pin: GPIO14
  rgb_current: 20mA
  cw_current: 55mA
  
output:
  - platform: sm2135
    id: output_red
    channel: 0
  - platform: sm2135
    id: output_green
    channel: 1
  - platform: sm2135
    id: output_blue
    channel: 2
  - platform: sm2135
    id: output_white
    channel: 3
  - platform: sm2135
    id: output_warmwhite
    channel: 4
```

## Checklist:

### DIFF
--- a/esphome/components/sm2135/sm2135.cpp
+++ b/esphome/components/sm2135/sm2135.cpp
@@ -144,7 +144,6 @@ void SM2135::sm2135_set_low_(GPIOPin *pin) {
 }
 
 void SM2135::sm2135_set_high_(GPIOPin *pin) {
-  pin->digital_write(true);
   pin->pin_mode(gpio::FLAG_PULLUP);
 }
 


Cossid:
@matika77 could this not lead to failures to actually write to the data channel.  I'm assuming this is meant to fix clock/timing.

Perhaps a better solution would be to mimic the separation of write_bit and write_byte with the timings those functions contained in each?  Similar to [sm10bit_base](https://github.com/esphome/esphome/blob/dev/esphome/components/sm10bit_base/sm10bit_base.cpp)?

I'm saying this without looking back to the datasheet, this is based on pure speculation.

matika77:
Looking at the datasheet, which is only available in chinese or autotranslated to my knownledge, the suggested wiring diagram shows a direct connection between the MCU and the CLK/DAT Pins without any pull-up resistors that would be used on a standard i2c bus. There is no mention about if there is significant current needed to drive the clk/dat pin high, so i think it's voltage driven and an internal pull-up of the esp8266/32 is enough.

Using only the internal pullups works for my sm2135 lamp. but what does NOT work is to set the output to high (providing the possibility to drain some mA) and THEN switch to the internal pullup. This causes my lamp to flicker.

Does anybody experience something else when using only the pull-ups ?  @BoukeHaarsma23  @dd32 

@Cossid can you explain what possible timing problems you're refering to ?

Cossid:
set_high is being called by both data_pin and clock_pin and with delays only at specific stops instead of between each pin type change.  It seems odd that changing only set_high and not set_low is the correct thing to do.

I recall some timing issues only working on one platform or another (esp8266 vs beken for example), and seeing differences on each platform.  On top of that, there are 3 different SM2135 chips (SM2135, SM2135E, and SM2135EH), and I don't fully know the differences between them.

I think it would be best to first try adding delays between the data & clock on each change, like I mentioned is being done in the sm10bit_base driver, as that seems to make things more consistent across platforms, and I think that is the way (at least that specific) spec is written.

Having a write_bit with delays for data & clock seemed to make things more stable.

matika77:
@Cossid I don't understand the relationship between the pin drive-mode and the possible timing issues. my commit is only related to the drive-mode. 

but i agree that this chip is very sensitive to timing. for me the delayMicroseconds(x) timings in this code works fine.

Cossid:
I'd be curious if we even need pin_mode at all, or as much as it's currently used.  I think it's just hiding the timing issues.  A digital_write already should be doing the mode change at the platform level.  sm10bit_base for example only uses pin_mode to change between input and output.

Did you try doing it the other way with leaving the digital_write and removing the pin_mode?

Doesn't appear I have an esp8266 or esp32 sm2135 bulb to test.  I'll try to test your proposed change on a beken based one and see how that goes.

matika77:
the pin_mode() is used to actually change the pin high/low state. the PULLUP mode is used to drive the pin high and changing it to output is used to drive the pin LOW. this is a common way to do it on an open-drain bus - but commonly with a floating high-z pin mode (on a i2c bus with external pull-up resistors) which is not available on the ESP to my knownledge. 

so the pinMode IS rather important.

Cossid:
But we're just bit-banging here via software.  digital_write(true) and pin_mode(internal_pullup) are essentially the same thing, so having both is presenting the potential timing issue, were the LED driver interpretation is time/signal-based.

That's why I find it odd you only removed digital_write from set_high and not set_low.

That being said, this change doesn't appear to break SM2135 on my Beken test bulb.

matika77:
No, digital_write(HIGH) is NOT the same as pin_mode(internal_pullup):

- digital_write(HIGH) leaves the pinmode to output which is capable to output some mA and has a low internal resistance. depending on the input-pin of the sm2135 this can lead (which i don't know the internals) to a current-spike (and possible voltage instabilities)

- pin_mode(internal_pullup) configures the pin to have a high-impedance input but adds an additional resistor between VCC and the configured pin. in this mode the current-spike is not possible.

- the digitalWrite(LOW) is needed to set the output to low BEFORE changing the mode to output. otherwise this can lead to an unwanted behaviour

the difference between these two modes made the data transmission to become corrupted on my lamp.

Cossid:
Thanks for explaining that, I think I follow what you're saying.  I'm finding the SM2135 and SM10bit (SM2235,SM2335) have a bit different timings where I was expecting them to be more similar.

As I said, I have now tested this on Beken SM2135 and do not find your change to have any ill side effects there.

matika77:
thank you for testing this. i hope someone else can confirm this and merge the pull request.

matika77:
@Cossid just out of curiosity: are there pull up resistors on the clk/dat lines on your beken device ? do you have the schematics or a high-res photo of your device ? or a link to that ?

Cossid:
Unfortunately, it doesn't look like I took photos for this device, and I am unsure if I ever physically opened any of these, as they were exploited via CloudCutter.  Their pin configuration does not require changing from the default pin assignment for mode.

Cossid:
But since you mention that, I do wonder if that is where we are seeing things differently.  I think the pullup should be part of the pin configuration, and not part of the driver, and the driver should only be changing high/low & input/output (bit-banging only).  Which is why I was expecting digital_write to always be doing the correct thing, but since this driver seems to be opinionated on the pullup via setup and high change, that probably isn't the case.

Cossid:
@matika77 here is an example of what I was recommending; https://github.com/Cossid/esphome/blob/sm2135_refactor_bitbang_only/esphome/components/sm2135/sm2135.cpp
All the timings are taken as closely as I could from the current driver, as I don't feel like trying to retranslate the existing docuemnts again.

Can you give

```yaml
external_components:
  - source: github://cossid/esphome@sm2135_refactor_bitbang_only
    refresh: 10min
    components: [sm2135]
```
a try and see if it works correctly and if there is any flickering?  You may need to change the pin assignment in your config to include `INPUT_PULLUP`

Edit: I did make one modification, instead of stop/starting and changing channel pointer to cw, I just passed empty color to avoid the start/stop, should not make a difference to anything.

matika77:
Hey @Cossid. I don't exactly understand how i should modify your source to include "INPUT_PULLUP", since your example only uses the OUTPUT-mode and digitalWrite. digitalWrite should only be used when a pin is in OUTPUT mode. so i shouldn't modify the pins pinMode() to INPUT_PULLUP.

But i tested your example and it does flicker, yes. my lamp doesn't like an high-current HIGH. and I really recommend to not use digitalWrite(HIGH) because it can cause problems in an open-drain bus configuration. the sm2135 uses an (non-standard) i2c bus which should be an open-drain. this requires the use of digitalWrite(LOW) && pinMode(OUTPUT) as LOW and pinMode(INPUT_PULLUP) as HIGH. This is the way standard i2c is implemented on ESP and also (with direct register writing) on AVR.

Cossid:
@matika77 You would use pin configuration https://esphome.io/guides/configuration-types.html#pin-schema mode in longer form instead of just GPIO to set your pullup, which you can do separately from input/output.  I'm saying this shouldn't be a concern of the driver, but rather the configuration.

PR #6485, opened 2024-04-05
Added HTU21D models

MRemy2:

Added models to the HTU21D because the HTU21D sensor does not support heater_level

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3742
## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (changes requested):
Please do not combine fixes across different components into a single PR.

MRemy2:
Removed the fix for ble. I will include it in another pr.

esphome/core/component.cpp
jesserockz, line 1: As core changes, these should go into their own PR as well. Also you defined the message, but never used it.

MRemy2: The core changes are used in the htu21d.cpp files. I do not see why the commit should be split. The chages to the htu21d component depend on the core changes. I used  the 'status_clear_warning' with a message in the component files. Also the never used message is because of https://github.com/esphome/esphome/pull/6485/commits/fd59a0fc42a2cf26a0a944163faa5b044493f049. I did use the message in the first place. 
Thanks.

jesserockz: Ah my bad, there were conflicts when I did a merge and I guess it got lost. But in any case, small core changes like this should be in their own PRs so its easier to follow the chain of commits. It doesnt make sense if the PR/commit that adds a new core feature to be named `Added HTU21D models`.
This PR should only focus on that specific feature of adding new models.

MRemy2: created a new pull as my branches got messed up really bad and I found it easier to create a new PR 
https://github.com/esphome/esphome/pull/6511

MRemy2:
Closing pull request. tracked by https://github.com/esphome/esphome/pull/6511

PR #6511, opened 2024-04-10
Added Htu21d model option

MRemy2:

Added models to the HTU21D because the HTU21D sensor does not support heater_level
## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3742

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/htu21d/htu21d.cpp
+++ b/esphome/components/htu21d/htu21d.cpp
@@ -83,14 +83,20 @@ void HTU21DComponent::update() {
 
       int8_t heater_level;
 
-      // HTU21D does not have heater level
+      // HTU21D does have a heater module but does not have heater level
+      // Setting heater level to 1 in case the heater is ON
       if (this->sensor_model_ == HTU21D_SENSOR_MODEL_HTU21D) {
-        heater_level = 0;
+        if(is_heater_enabled) {
+          heater_level = 1;
+        } else {
+          heater_level = 0;
+        }
       } else {
         heater_level = this->get_heater_level();
-        ESP_LOGD(TAG, "Heater Level=%d", heater_level);
       }
 
+      ESP_LOGD(TAG, "Heater Level=%d", heater_level);
+      
       if (this->heater_ != nullptr)
         this->heater_->publish_state(heater_level);
       this->status_clear_warning();
--- a/esphome/components/htu21d/htu21d.cpp
+++ b/esphome/components/htu21d/htu21d.cpp
@@ -76,12 +76,21 @@ void HTU21DComponent::update() {
 
       float humidity = (float(raw_humidity & 0xFFFC)) * 125.0f / 65536.0f - 6.0f;
 
-      int8_t heater_level = this->get_heater_level();
-
-      ESP_LOGD(TAG, "Got Humidity=%.1f%% Heater Level=%d", humidity, heater_level);
+      ESP_LOGD(TAG, "Got Humidity=%.1f%%", humidity);
 
       if (this->humidity_ != nullptr)
         this->humidity_->publish_state(humidity);
+
+      int8_t heater_level;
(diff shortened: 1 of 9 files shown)


esphome/components/htu21d/htu21d.cpp
```diff
@@ -76,12 +76,21 @@ void HTU21DComponent::update() {
...
-
-      ESP_LOGD(TAG, "Got Humidity=%.1f%% Heater Level=%d", humidity, heater_level);
+      ESP_LOGD(TAG, "Got Humidity=%.1f%%", humidity);
 
       if (this->humidity_ != nullptr)
         this->humidity_->publish_state(humidity);
+
+      int8_t heater_level;
+
+      // HTU21D does not have heater level
```
jesserockz: If this is true, then `heater` should also be blocked with a custom validation from being added in config if using the htu21d model type

jesserockz (changes requested):
Along with the below comment, please also add the model to `tests/components/htu21d/...`

esphome/components/htu21d/htu21d.cpp
MRemy2: @jesserockz 
I really do not want to sound rude or anything like that but there are some things I need to mention:

- I do not know python and do not really understand how the python files work (I took inspiration from the hlw8012 for the model option).
- I do now how to read a datasheet and have some C/C++ experience but I would not say expert level.
- the component HTU21D (after checking all the datasheets of the different models of sensors suported) is very flawed
- all sensor models have a heater that can be turned ON/OFF but only the SI7021 has power levels for the heater.
- these 2 registers are only available on the SI7021 HTU21D_WRITEHEATER_REG_CMD = 0x51, HTU21D_READHEATER_REG_CMD = 0x11 so why they have HTU21D in the name beats my logic.
- after an update of the esphome (as I am using the HTU21D sensor in a project) I noticed a lot of set and cleared warnings in the logs, I saw that the 'heater' functionality added quite recent is the cause, and I did try to fix it to some extent (add models and for HTU21D ignore the heater level)
- I do not have the rest of the sensors to fully fix and test the component and also my time is quite limited (full time job and a 3 year old to fill the rest of the time)

Also I am not really sure why "this->read_register(HTU21D_REGISTER_STATUS, reinterpret_cast<uint8_t *>(&raw_heater), 2)" is reading 2 bytes as the user registers R/W are 1 byte long ... (yes the temp and humidity being 14 bits resolution need a 2 byte read)

Until recently I fixed issues I found in different projects I came across on a local basis and didn't think to much about giving back (PRs etc) but because I do like the esphome project so much I thought it would not cost me that much to do a PR and give back to others... boy I was wrong. A week of back and forth for a simple model add and a quick fix for ble ... as all needs to be separate, etc. (To be fair I have not managed a huge repo like this with all the PRs it gets etc, to know the implications)

I wish that esphome will become greater and greater as in my opinion is a great asset for the home automation community.

Regarding the PR I did add a condition that sets the heater level to 1 in case the heater inside the chip is on for HTU21D.
If this is not enough ... then with all regret I give up as I do not have the necesary time. Feel free to make any changes needed.

jesserockz: I will take a look, but I understood from your inline comment there was no heater, my mistake there.

PR reviews are also about teaching developers how to learn the python and c++ side and I thank you for sticking around this long =)

jesserockz (approved):
Thanks for making this change. Indeed using the `is_heater_enabled` to set the value is a good workaround here.

PR #6512, opened 2024-04-11
Disallow variant/family override for known boards

### LINKED ISSUE #5664: RPI_DPI_RGB display will not compile
### The problem

I am attempting to configure the RPI_DPI_RGB display driver for the ESP32-8048S070 development board. This has an ESP32-S3, 7 inch RGB display, GT911 touch screen and a plethora of other handy features.

**The problem**
When compiling a basic YAML configuration I get the following error.

```
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp: In member function 'virtual void esphome::rpi_dpi_rgb::RpiDpiRgb::setup()':
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:10:3: error: 'esp_lcd_rgb_panel_config_t' was not declared in this scope
   esp_lcd_rgb_panel_config_t config{};
   ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:10:3: note: suggested alternative: 'esp_lcd_panel_init'
   esp_lcd_rgb_panel_config_t config{};
   ^~~~~~~~~~~~~~~~~~~~~~~~~~
   esp_lcd_panel_init
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:11:3: error: 'config' was not declared in this scope
   config.flags.fb_in_psram = 1;
   ^~~~~~
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:11:3: note: suggested alternative: 'confstr'
   config.flags.fb_in_psram = 1;
   ^~~~~~
   confstr
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:35:19: error: 'esp_lcd_new_rgb_panel' was not declared in this scope
   esp_err_t err = esp_lcd_new_rgb_panel(&config, &this->handle_);
                   ^~~~~~~~~~~~~~~~~~~~~
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:35:19: note: suggested alternative: 'esp_lcd_panel_set_gap'
   esp_err_t err = esp_lcd_new_rgb_panel(&config, &this->handle_);
                   ^~~~~~~~~~~~~~~~~~~~~
                   esp_lcd_panel_set_gap
*** [.pioenvs\displaytest\src\esphome\components\rpi_dpi_rgb\rpi_dpi_rgb.o] Error 1
Compiling .pioenvs\displaytest\src\esphome\core\helpers.o
```

latest and dev. Clean install of python, ESPHome etc did not change the issue

pip

NA

ESP32-IDF

### Board

ESP32-8048S070

RPI_DPI_RGB display driver

### Example YAML snippet

```yaml
substitutions:
  esphome_name: displaytest

esphome:
  name: "${esphome_name}"
  
esp32:
  board: esp32dev
  variant: esp32s3
  #  board: ESP32-8048S070
  framework:
    type: esp-idf
    version: recommended

psram:
  mode: octal
  speed: 80MHz

# Enable logging
logger:

display:
  - platform: rpi_dpi_rgb
    id: S070
    data_pins:
        red:
            - 45        #r1
            - 48        #r2
            - 47        #r3
            - 21        #r4
            - 14        #r5
        green:
            - 9         #g0
            - 46        #g1
            - 3         #g2
            - 8         #g3
            - 16        #g4
            - 1         #g5
        blue:
            - 15        #b1
            - 7         #b2
            - 6         #b3
            - 5         #b4
            - 4         #b5
    de_pin:
        number: 41
    pclk_pin: 42
    hsync_pin: 39
    vsync_pin: 40
    hsync_pulse_width: 30
    hsync_front_porch: 210
    hsync_back_porch: 16
    vsync_pulse_width: 13
    vsync_front_porch: 22
    vsync_back_porch: 10
    update_interval: never
    auto_clear_enabled: false
    color_order: RGB
    dimensions:
        width: 800
        height: 480
    pclk_frequency: 16MHz
    pclk_inverted: true
    rotation: 90
```

```txt
C:\Users\simon\esphome>esphome compile DisplayTest1.yml
INFO ESPHome 2024.3.1
INFO Reading configuration DisplayTest1.yml...
INFO Generating C++ source...
INFO Compiling app...
Processing displaytest (board: esp32dev; framework: espidf; platform: platformio/espressif32@5.4.0)
-------------------------------------------------------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/espressif32/esp32dev.html
PLATFORM: Espressif 32 (5.4.0) > Espressif ESP32 Dev Module
HARDWARE: ESP32 240MHz, 320KB RAM, 4MB Flash
DEBUG: Current (cmsis-dap) External (cmsis-dap, esp-bridge, esp-prog, iot-bus-jtag, jlink, minimodule, olimex-arm-usb-ocd, olimex-arm-usb-ocd-h, olimex-arm-usb-tiny-h, olimex-jtag-tiny, tumpa)
PACKAGES:
 - framework-espidf @ 3.40406.240122 (4.4.6)
 - tool-cmake @ 3.16.4
 - tool-esptoolpy @ 1.40400.0 (4.4.0)
 - tool-idf @ 1.0.1
 - tool-mconf @ 1.4060000.20190628 (406.0.0)
 - tool-ninja @ 1.9.0
 - toolchain-esp32ulp @ 2.35.0-20220830
 - toolchain-xtensa-esp32 @ 8.4.0+2021r2-patch5
Creating a new virtual environment for IDF Python dependencies
Installing ESP-IDF's Python dependencies
Collecting urllib3<2
  Downloading urllib3-1.26.18-py2.py3-none-any.whl.metadata (48 kB)
     -------------------------------------- 48.9/48.9 kB 415.8 kB/s eta 0:00:00
Collecting cryptography<35.0.0,>=2.1.4
  Downloading cryptography-3.4.8-cp36-abi3-win_amd64.whl.metadata (5.1 kB)
Collecting future>=0.18.3
  Downloading future-1.0.0-py3-none-any.whl.metadata (4.0 kB)
Collecting pyparsing<2.4.0,>=2.0.3
  Downloading pyparsing-2.3.1-py2.py3-none-any.whl.metadata (1.0 kB)
Collecting kconfiglib~=13.7.1
  Using cached kconfiglib-13.7.1-py2.py3-none-any.whl.metadata (39 kB)
Collecting idf-component-manager~=1.0
  Downloading idf_component_manager-1.5.2-py2.py3-none-any.whl.metadata (12 kB)
Collecting cffi>=1.12 (from cryptography<35.0.0,>=2.1.4)
  Using cached cffi-1.16.0-cp312-cp312-win_amd64.whl.metadata (1.5 kB)
Collecting packaging (from idf-component-manager~=1.0)
  Using cached packaging-24.0-py3-none-any.whl.metadata (3.2 kB)
Collecting requests<3 (from idf-component-manager~=1.0)
  Using cached requests-2.31.0-py3-none-any.whl.metadata (4.6 kB)
Collecting requests-file<2 (from idf-component-manager~=1.0)
  Downloading requests_file-1.5.1-py2.py3-none-any.whl.metadata (596 bytes)
Collecting requests-toolbelt (from idf-component-manager~=1.0)
  Downloading requests_toolbelt-1.0.0-py2.py3-none-any.whl.metadata (14 kB)
Collecting schema (from idf-component-manager~=1.0)
  Downloading schema-0.7.5-py2.py3-none-any.whl.metadata (34 kB)
Collecting six (from idf-component-manager~=1.0)
  Using cached six-1.16.0-py2.py3-none-any.whl.metadata (1.8 kB)
Collecting tqdm<5 (from idf-component-manager~=1.0)
  Downloading tqdm-4.66.2-py3-none-any.whl.metadata (57 kB)
     ---------------------------------------- 57.6/57.6 kB 1.5 MB/s eta 0:00:00
Collecting colorama (from idf-component-manager~=1.0)
  Using cached colorama-0.4.6-py2.py3-none-any.whl.metadata (17 kB)
Collecting pyyaml>5.2 (from idf-component-manager~=1.0)
  Using cached PyYAML-6.0.1-cp312-cp312-win_amd64.whl.metadata (2.1 kB)
Collecting cachecontrol>0.12.6 (from cachecontrol[filecache]>0.12.6; python_version >= "3.6"->idf-component-manager~=1.0)
  Downloading cachecontrol-0.14.0-py3-none-any.whl.metadata (3.1 kB)
Collecting contextlib2>0.6.0 (from idf-component-manager~=1.0)
  Downloading contextlib2-21.6.0-py2.py3-none-any.whl.metadata (4.1 kB)
Collecting click (from idf-component-manager~=1.0)
  Using cached click-8.1.7-py3-none-any.whl.metadata (3.0 kB)
Collecting msgpack<2.0.0,>=0.5.2 (from cachecontrol>0.12.6->cachecontrol[filecache]>0.12.6; python_version >= "3.6"->idf-component-manager~=1.0)
  Downloading msgpack-1.0.8-cp312-cp312-win_amd64.whl.metadata (9.4 kB)
Collecting filelock>=3.8.0 (from cachecontrol[filecache]>0.12.6; python_version >= "3.6"->idf-component-manager~=1.0)
  Downloading filelock-3.13.3-py3-none-any.whl.metadata (2.8 kB)
Collecting pycparser (from cffi>=1.12->cryptography<35.0.0,>=2.1.4)
  Using cached pycparser-2.22-py3-none-any.whl.metadata (943 bytes)
Collecting charset-normalizer<4,>=2 (from requests<3->idf-component-manager~=1.0)
  Using cached charset_normalizer-3.3.2-cp312-cp312-win_amd64.whl.metadata (34 kB)
Collecting idna<4,>=2.5 (from requests<3->idf-component-manager~=1.0)
  Using cached idna-3.6-py3-none-any.whl.metadata (9.9 kB)
Collecting certifi>=2017.4.17 (from requests<3->idf-component-manager~=1.0)
  Using cached certifi-2024.2.2-py3-none-any.whl.metadata (2.2 kB)
Downloading urllib3-1.26.18-py2.py3-none-any.whl (143 kB)
   ---------------------------------------- 143.8/143.8 kB 1.2 MB/s eta 0:00:00
Downloading cryptography-3.4.8-cp36-abi3-win_amd64.whl (1.6 MB)
   ---------------------------------------- 1.6/1.6 MB 1.9 MB/s eta 0:00:00
Downloading future-1.0.0-py3-none-any.whl (491 kB)
   ---------------------------------------- 491.3/491.3 kB 2.8 MB/s eta 0:00:00
Downloading pyparsing-2.3.1-py2.py3-none-any.whl (61 kB)
   ---------------------------------------- 61.8/61.8 kB 3.4 MB/s eta 0:00:00
Using cached kconfiglib-13.7.1-py2.py3-none-any.whl (145 kB)
Downloading idf_component_manager-1.5.2-py2.py3-none-any.whl (160 kB)
   ---------------------------------------- 160.3/160.3 kB 3.2 MB/s eta 0:00:00
Downloading cachecontrol-0.14.0-py3-none-any.whl (22 kB)
Using cached cffi-1.16.0-cp312-cp312-win_amd64.whl (181 kB)
Downloading contextlib2-21.6.0-py2.py3-none-any.whl (13 kB)
Using cached PyYAML-6.0.1-cp312-cp312-win_amd64.whl (138 kB)
Using cached requests-2.31.0-py3-none-any.whl (62 kB)
Downloading requests_file-1.5.1-py2.py3-none-any.whl (3.7 kB)
Downloading tqdm-4.66.2-py3-none-any.whl (78 kB)
   ---------------------------------------- 78.3/78.3 kB 2.1 MB/s eta 0:00:00
Using cached click-8.1.7-py3-none-any.whl (97 kB)
Using cached colorama-0.4.6-py2.py3-none-any.whl (25 kB)
Using cached packaging-24.0-py3-none-any.whl (53 kB)
Downloading requests_toolbelt-1.0.0-py2.py3-none-any.whl (54 kB)
   ---------------------------------------- 54.5/54.5 kB 1.4 MB/s eta 0:00:00
Downloading schema-0.7.5-py2.py3-none-any.whl (17 kB)
Using cached six-1.16.0-py2.py3-none-any.whl (11 kB)
Using cached certifi-2024.2.2-py3-none-any.whl (163 kB)
Using cached charset_normalizer-3.3.2-cp312-cp312-win_amd64.whl (100 kB)
Downloading filelock-3.13.3-py3-none-any.whl (11 kB)
Using cached idna-3.6-py3-none-any.whl (61 kB)
Downloading msgpack-1.0.8-cp312-cp312-win_amd64.whl (75 kB)
   ---------------------------------------- 75.3/75.3 kB 4.1 MB/s eta 0:00:00
Using cached pycparser-2.22-py3-none-any.whl (117 kB)
Installing collected packages: pyparsing, urllib3, six, pyyaml, pycparser, packaging, msgpack, kconfiglib, idna, future, filelock, contextlib2, colorama, charset-normalizer, certifi, tqdm, schema, requests, click, cffi, requests-toolbelt, requests-file, cryptography, cachecontrol, idf-component-manager
Successfully installed cachecontrol-0.14.0 certifi-2024.2.2 cffi-1.16.0 charset-normalizer-3.3.2 click-8.1.7 colorama-0.4.6 contextlib2-21.6.0 cryptography-3.4.8 filelock-3.13.3 future-1.0.0 idf-component-manager-1.5.2 idna-3.6 kconfiglib-13.7.1 msgpack-1.0.8 packaging-24.0 pycparser-2.22 pyparsing-2.3.1 pyyaml-6.0.1 requests-2.31.0 requests-file-1.5.1 requests-toolbelt-1.0.0 schema-0.7.5 six-1.16.0 tqdm-4.66.2 urllib3-1.26.18
Installing windows-curses package
Collecting windows-curses
  Downloading windows_curses-2.3.2-cp312-cp312-win_amd64.whl.metadata (2.7 kB)
Downloading windows_curses-2.3.2-cp312-cp312-win_amd64.whl (80 kB)
   ---------------------------------------- 80.5/80.5 kB 562.4 kB/s eta 0:00:00
Installing collected packages: windows-curses
Successfully installed windows-curses-2.3.2
Installing windows-curses package
Processing c:\users\simon\.platformio\packages\framework-espidf\tools\kconfig_new\esp-windows-curses
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Requirement already satisfied: windows-curses in c:\users\simon\.platformio\penv\.espidf-4.4.6\lib\site-packages (from esp-windows-curses==0.1) (2.3.2)
Building wheels for collected packages: esp-windows-curses
  Building wheel for esp-windows-curses (pyproject.toml): started
  Building wheel for esp-windows-curses (pyproject.toml): finished with status 'done'
  Created wheel for esp-windows-curses: filename=esp_windows_curses-0.1-py3-none-any.whl size=1167 sha256=57bc264f68a456a3c5d059b9d8173938c7f67a48a75d1d857e04ce408f895a48
  Stored in directory: c:\users\simon\appdata\local\pip\cache\wheels\1b\d0\81\36fe1ca212effe933ad1e6540447a1d8ba214c3aeadacd3299
Successfully built esp-windows-curses
Installing collected packages: esp-windows-curses
Successfully installed esp-windows-curses-0.1
Reading CMake configuration...
Generating assembly for certificate bundle...
LDF: Library Dependency Finder -> https://bit.ly/configure-pio-ldf
LDF Modes: Finder ~ off, Compatibility ~ soft
Found 0 compatible libraries
Scanning dependencies...
No dependencies
Building in release mode
Compiling .pioenvs\displaytest\src\esphome\components\display\display.o
Compiling .pioenvs\displaytest\src\esphome\components\display\display_buffer.o
Compiling .pioenvs\displaytest\src\esphome\components\display\rect.o
Compiling .pioenvs\displaytest\src\esphome\components\esp32\core.o
Compiling .pioenvs\displaytest\src\esphome\components\esp32\gpio.o
Compiling .pioenvs\displaytest\src\esphome\components\esp32\preferences.o
Compiling .pioenvs\displaytest\src\esphome\components\logger\logger.o
Compiling .pioenvs\displaytest\src\esphome\components\logger\logger_esp32.o
Compiling .pioenvs\displaytest\src\esphome\components\logger\logger_esp8266.o
Compiling .pioenvs\displaytest\src\esphome\components\logger\logger_host.o
Compiling .pioenvs\displaytest\src\esphome\components\logger\logger_libretiny.o
Compiling .pioenvs\displaytest\src\esphome\components\logger\logger_rp2040.o
Compiling .pioenvs\displaytest\src\esphome\components\psram\psram.o
Compiling .pioenvs\displaytest\src\esphome\components\rpi_dpi_rgb\rpi_dpi_rgb.o
Compiling .pioenvs\displaytest\src\esphome\core\application.o
Compiling .pioenvs\displaytest\src\esphome\core\color.o
Compiling .pioenvs\displaytest\src\esphome\core\component.o
Compiling .pioenvs\displaytest\src\esphome\core\component_iterator.o
Compiling .pioenvs\displaytest\src\esphome\core\controller.o
Compiling .pioenvs\displaytest\src\esphome\core\entity_base.o
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp: In member function 'virtual void esphome::rpi_dpi_rgb::RpiDpiRgb::setup()':
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:10:3: error: 'esp_lcd_rgb_panel_config_t' was not declared in this scope
   esp_lcd_rgb_panel_config_t config{};
   ^~~~~~~~~~~~~~~~~~~~~~~~~~
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:10:3: note: suggested alternative: 'esp_lcd_panel_init'
   esp_lcd_rgb_panel_config_t config{};
   ^~~~~~~~~~~~~~~~~~~~~~~~~~
   esp_lcd_panel_init
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:11:3: error: 'config' was not declared in this scope
   config.flags.fb_in_psram = 1;
   ^~~~~~
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:11:3: note: suggested alternative: 'confstr'
   config.flags.fb_in_psram = 1;
   ^~~~~~
   confstr
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:35:19: error: 'esp_lcd_new_rgb_panel' was not declared in this scope
   esp_err_t err = esp_lcd_new_rgb_panel(&config, &this->handle_);
                   ^~~~~~~~~~~~~~~~~~~~~
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:35:19: note: suggested alternative: 'esp_lcd_panel_set_gap'
   esp_err_t err = esp_lcd_new_rgb_panel(&config, &this->handle_);
                   ^~~~~~~~~~~~~~~~~~~~~
                   esp_lcd_panel_set_gap
*** [.pioenvs\displaytest\src\esphome\components\rpi_dpi_rgb\rpi_dpi_rgb.o] Error 1
Compiling .pioenvs\displaytest\src\esphome\core\helpers.o
================================================================ [FAILED] Took 77.70 seconds ================================================================
```

### Additional information

_No response_

clydebarrow:

## Types of changes

Currently setting `variant:` (`family:` for LibreTiny) in the esp32 component makes ESPHome think it's compiling for a variant that may be different to the default one from `board:`. This option should not be allowed if the board type is known, since the user may infer it actually changes something. 

~This is a breaking change since any yamls currently including `variant:` will now fail to compile.~

Now the variant will be allowed without any warning or error message if it matches the board.

Tests have not been added since the ESPHome testing structure currently has no provision for tests that must fail.

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5664

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3754
## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [x] RTL87xx

```yaml
# Example config.yaml

rtl87xx:
  board: generic-rtl8710bn-2mb-xxxk
  variant: rtl8720cf
```

## Checklist:


nielsnl68 (approved):

kuba2k2:
Hi,
I think these changes might actually be unnecessary for the LibreTiny platform - it doesn't use the `build.mcu` field for compilation. The compilation process only depends on the `board` JSON file, which contains the information about chip family.

The `family:` field was added to ESPHome to allow using new boards supported by LibreTiny, which haven't been added to ESPHome yet. It does not change the compilation parameters, it only tells ESPHome to adjust its own `USE_*` macros to that family. LibreTiny will not work without a board JSON. Changing its family directly is also not possible (different partition layouts, sometimes even encryption keys).

Since the term "variant" is not used internally in LibreTiny (instead "family" is used), I'd propose to keep the original naming (maybe even replace the `USE_LIBRETINY_VARIANT_*` with `USE_LIBRETINY_FAMILY_*`.

Also, adding each chip type to ESPHome might make maintaining it more involving - each new chip type supported by LibreTiny will need to be added and users will not be able to just choose a family of a newly supported board.

The following test YAML:
```yaml
rtl87xx:
  board: generic-rtl8710bn-2mb-788k
  variant: rtl8720cf
```
will compile the firmware for the RTL8710BN chip, not the RTL8720CF. These belong to different families (AmebaZ vs AmebaZ2).

I hope this will make it easier to understand how LibreTiny handles the compilation for different chips.

clydebarrow:
Thanks for the response.

> I think these changes might actually be unnecessary for the LibreTiny platform - it doesn't use the build.mcu field for compilation. The compilation process only depends on the board JSON file, which contains the information about chip family.

And currently that (the PIO board file) is where PIO gets the `mcu` type from since ESPHome doesn't pass on the `variant` selection.

Let me backtrack a bit to explain the reason for going down this rabbit hole:

The original problem (which came up for ESP32) was that if a board is specified but a variant also specified, internally ESPHome uses the variant, but does not pass that to Platformio, which leads to C++ compile-time failures if e.g. pins or libraries dependent on ESP32S3 are compiled for ESP32. E.g:
```
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp: In member function 'virtual void esphome::rpi_dpi_rgb::RpiDpiRgb::setup()':
src/esphome/components/rpi_dpi_rgb/rpi_dpi_rgb.cpp:10:3: error: 'esp_lcd_rgb_panel_config_t' was not declared in this scope
   esp_lcd_rgb_panel_config_t config{};
```

That's clearly undesirable - config errors should be caught at ESPHome build time - one of the strong points of ESPHome is that the average user should never see a compiler error, only YAML validation errors.

So the logical solution is to pass the `variant` on to PIO for compilation - it makes sense that if you have said this chip is an XYZ123 then PIO should compile for that, not the default for the board. So my original (simple) change was just to add the `build.mcu` PIO flag, and that works just fine for ESP32 (RP2040 and ESP8266 don't have variants, so it doesn't apply to them.) And I thought "Great! Job done!"


But, tests for Libretiny failed because for *some* of the chips, the `family` being derived from the board files was not an mcu type recognised by PIO. So I ended up in the rabbit hole changing that.

There is the question of what purpose the `variant` and `family` options serve in the first place, and if there's no good justification for being able to override the variant they could simply be removed. 

But if they are going to remain, they should work properly.

```
rtl87xx:
  board: generic-rtl8710bn-2mb-788k
  variant: rtl8720cf
```
> will compile the firmware for the RTL8710BN chip, not the RTL8720CF. These belong to different families (AmebaZ vs AmebaZ2).

No, that will (after this PR) compile for rtl8720cf. The current implementation would look like this:

```
rtl87xx:
  board: generic-rtl8710bn-2mb-788k
  family: rtl8720cf
```

And that (currently) fails to pass YAML verification:
```
rtl87xx: [source rtl87.yaml:7]
  board: generic-rtl8710bn-2mb-788k

  Unknown value 'RTL8720CF', did you mean 'RTL8720C', 'RTL8710B'?.
  family: rtl8720cf
```
If you use one of the suggested options, it now builds, but as you pointed out, for rtl8710bn. That's an unexpected result since something different was requested.

So basically the `family`/`variant` option is currently broken - it does not do what it says on the tin. So it should either be fixed or removed. I went to the trouble of fixing it, since I can see some utility in it being available, but if there's a consensus that it should be removed, I'll have no issue with that.

clydebarrow:
> Also, adding each chip type to ESPHome might make maintaining it more involving - each new chip type supported by LibreTiny will need to be added

There's a script that does that so the effort is small. If there is a board file in Platformio then ESPHome supports it. New variants would also automagically become available via the same process.

> Since the term "variant" is not used internally in LibreTiny (instead "family" is used), I'd propose to keep the original naming (maybe even replace the USE_LIBRETINY_VARIANT_* with USE_LIBRETINY_FAMILY_*.

I'd argue that the typical user doesn't need to know what happens internally in LibreTiny, and `variant` is an established term in ESPHome, so it makes sense to keep consistency. Internal constants like `USE_LIBRETINY_VARIANT_` are only seen by developers so that's of lesser importance (but VARIANT was already in use.)

kuba2k2:
> And currently that (the PIO board file) is where PIO gets the mcu type from since ESPHome doesn't pass on the variant selection.

Unfortunately, that's not the case in LibreTiny. It [uses `build.family`](https://github.com/libretiny-eu/libretiny/blob/master/builder/utils/env.py#L61) to determine the chip family, based on this the appropriate code [building script is used](https://github.com/libretiny-eu/libretiny/blob/master/builder/frameworks/base.py#L72) (`builder/family/*.py`). The `board.mcu` property is used 1) to [name the output .UF2 file](https://github.com/libretiny-eu/libretiny/blob/master/builder/utils/ltchiptool-util.py#L31) accordingly, 2) as a [`#define` for C code](https://github.com/libretiny-eu/libretiny/blob/master/builder/frameworks/base.py#L129), which only uses it in [`lt_cpu_get_model()` function](https://github.com/libretiny-eu/libretiny/blob/master/cores/common/base/api/lt_cpu.c#L19).

> That's clearly undesirable - config errors should be caught at ESPHome build time - one of the strong points of ESPHome is that the average user should never see a compiler error, only YAML validation errors.

I do 100% agree with that - I was trying to find a solution for LibreTiny too.

> But, tests for Libretiny failed because for some of the chips, the family being derived from the board files was not an mcu type recognised by PIO. So I ended up in the rabbit hole changing that.

Because `variant` is different than `family`. Family is for... well, a family of chips - MCUs that are similar, but not identical, and code may be directly compatible between them. One thing known for sure, is that code from a certain family will *not* boot on a different one.

> There is the question of what purpose the variant and family options serve in the first place, and if there's no good justification for being able to override the variant they could simply be removed.

You're right, there's no reason to override the `variant` (MCU type). However, the `family` override was added to make it possible to use new boards in ESPHome without waiting for a release with updated `boards.py`. 

> There's a script that does that so the effort is small. If there is a board file in Platformio then ESPHome supports it. 

Continuing: there is a script that generates the `boards.py`. But if a board gets support in LibreTiny, a PR for ESPHome will have to be submitted, merged and released as a new ESPHome minor version. Only then can the user compile for that board. The `family` field was added to "ignore what ESPHome knows" and basically say `hey ESPHome, use this board, it really exists and belongs to <xyz> family`.

---

> So the logical solution is to pass the variant on to PIO for compilation - it makes sense that if you have said this chip is an XYZ123 then PIO should compile for that, not the default for the board. 

Unfortunately, it is not that simple. Let's say you want to compile for the `cbu` board (BK7231N) and you change the family to BK7231T. If LibreTiny was to respect that change (by setting `build.family`, not `build.mcu`) it would compile for the BK7231T chip using settings appropriate for BK7231N chip - like the partition layout. The resulting firmware would probably not even be able to connect to Wi-Fi (if it boots at all).

> > will compile the firmware for the RTL8710BN chip, not the RTL8720CF. These belong to different families (AmebaZ vs AmebaZ2).
> 
> No, that will (after this PR) compile for rtl8720cf. The current implementation would look like this:

As mentioned in the 1st paragraph of this comment, LibreTiny will ignore that change. It will still compile for RTL8710BN. If you change `build.family` instead, it will try to compile for RTL8720CF and fail, because the board JSON is missing information that are required for that family (partition layout, bootloader keys).

---

> So basically the family/variant option is currently broken - it does not do what it says on the tin. So it should either be fixed or removed. I went to the trouble of fixing it, since I can see some utility in it being available, but if there's a consensus that it should be removed, I'll have no issue with that.

In conclusion: there's no reason to change the chip family in LibreTiny - it is not even possible. The `family:` field was only introduced to let ESPHome know about the family type of (yet) unknown boards. And you're correct that it is indeed half broken because of that

- If we remove the `family:` field, ESPHome will not allow using unknown (new) boards.
- If we keep that field, ESPHome will not be able to validate if the family type even exists.

I think the right solution would be to make the `family:` field only applicable for unknown boards, reserve it for advanced users only. This way, it's their responsibility to make sure the family is correct. Known boards wouldn't allow setting that property, so the validation and compilation would always pass without problems.

Maybe it would help to rename it, e.g. `board_family:` or `custom_family:` or `unknown_board_family:`. Then we can keep the existing `USE_LIBRETINY_VARIANT_*` macros with family names. Using chip names here is redundant, since the chips within a single family offer the same peripherals.

clydebarrow:
Ok, so I understand more about what's going on, and the motivation for the option. It does seem to work differently to using PIO with esp-idf.

> Maybe it would help to rename it, e.g. board_family: or custom_family: or unknown_board_family:.

I think this is probably the best idea right now. And it should be disallowed if the `board:` setting *is* known to ESPHome. Probably not even necessary to rename it, but have something like this:

```
rtl87xx:
  board: generic-rtl8710bn-2mb-1024k
  allow_unknown_board: true
  family: rtl8710b
```

It seems only effect of specifying the `family` on ESPHome is to set macros like:
`-DUSE_LIBRETINY_FAMILY_RTL8720CF` which (as far as I can tell) has no effect on any ESPHome code right now. But it needs a placeholder since it can't get the family from the board definition. So another question is whether there should be any validation on the argument, since if the board is unknown to ESPHome the family may be also.

How does that sound? And I can't see any reason not to use the same approach with ESP32, since the motivation for the `variant` option is presumably the same (but for ESP32 the variant does affect ESPHome since it has conditional code for different variants.)

Not important - but what's the difference between RTL8710B (a family) and RTL8710BN (or X) - which are mcu types within that family. The BK series apparently has no such oddities.

kuba2k2:
> I think this is probably the best idea right now. And it should be disallowed if the `board:` setting _is_ known to ESPHome. Probably not even necessary to rename it, but have something like this:
> 
> ```
> rtl87xx:
>   board: generic-rtl8710bn-2mb-1024k
>   allow_unknown_board: true
>   family: rtl8710b
> ```

Yes, this solution seems to fit really well.

> It seems only effect of specifying the `family` on ESPHome is to set macros like:
> `-DUSE_LIBRETINY_FAMILY_RTL8720CF` which (as far as I can tell) has no effect on any ESPHome code right now. But it needs a placeholder since it can't get the family from the board definition. So another question is whether there should be any validation on the argument, since if the board is unknown to ESPHome the family may be also.

Correct, the macros are only supposed to enable/disable certain features available on a certain family only, however as you have noticed they aren't used so far. Up to this point, the family type validation required it to be known to ESPHome - after all, there's no benefit of specifying the family name unknown to ESPHome, as there will be no macros using that name anyway. Removing the macros would allow to remove the `family:` field entirely, but I think it's better to keep them for any future needs.

In the current codebase the family name for *known* boards is in `boards.py`, so retrieving the name for macros is not a problem. For *unknown* boards, you can just use whatever family is used in `family:`.

I can't speak for the ESP32 variant field, because I'm not familiar enough with the differences that it introduces.

> Not important - but what's the difference between RTL8710B (a family) and RTL8710BN (or X) - which are mcu types within that family. The BK series apparently has no such oddities.

Good question, it seems to be a bit confusing at first. LibreTiny configures most compilation options based on the family - SDK to use, files to compile, #defines to enable, etc. The MCU type is less important and not generally used. In short, the family divides chips by their similarity and compatibility.
The BK series could all belong to the `BK72XX` or `BK7231` family - however, the T and N variants (as well as BK7252) do have some differences in peripherals (like Wi-Fi, PWM, interrupts) which makes the code pretty much incompatible, so they have been split into several families.
That being said, if I come up with a better solution to reorganize the families, I probably will do that in the next refactor.

esphome/components/esp32/__init__.py
```diff
@@ -316,17 +316,23 @@ def _parse_platform_version(value):
...
 
 def _detect_variant(value):
-    if CONF_VARIANT not in value:
-        board = value[CONF_BOARD]
-        if board not in BOARDS:
+    board = value[CONF_BOARD]
+    if board in BOARDS:
+        variant = BOARDS[board][KEY_VARIANT]
+        if CONF_VARIANT in value and variant != value[CONF_VARIANT]:
+            raise cv.Invalid(f"Option '{CONF_VARIANT}' does not match selected board.")
```
kuba2k2: I might be wrong, but shouldn't `path=[CONF_VARIANT]` be added here, so that the validation error shows what's wrong?

kuba2k2 (approved):

jesserockz (approved):

PR #6515, opened 2024-04-11
Add beken_spi_led_strip component

Mat931:

Allows using addressable LEDs like the WS2812 on pin P16 of the BK7231N. This is a temporary solution until LibreTiny provides the necessary SPI functions. Other chips (BK7231T and BK7251) are untested. For accurate timings the SPI DMA interface of the Beken chip is used.

[Documentation](https://deploy-preview-3755--esphome.netlify.app/components/light/beken_spi_led_strip)

[Device using this component](https://devices.esphome.io/devices/Oyajia-LED-Ceiling-light-AS105)

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3755

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
bk72xx:
  board: generic-bk7231n-qfn32-tuya

external_components:
  - source: github://pr#6515
    components: [beken_spi_led_strip]
    refresh: always

light:
  - platform: beken_spi_led_strip
    name: "Color Light"
    pin: P16
    chipset: WS2812
    num_leds: 60
    rgb_order: GRB
```

## Checklist:


esphome/components/beken_spi_led_strip/light.py
```diff
@@ -0,0 +1,133 @@
...
+    light.ADDRESSABLE_LIGHT_SCHEMA.extend(
+        {
+            cv.GenerateID(CONF_OUTPUT_ID): cv.declare_id(BekenSPILEDStripLightOutput),
+            cv.Required(CONF_PIN): cv.All(
+                pins.internal_gpio_output_pin_number, _validate_pin
+            ),
+            cv.Required(CONF_NUM_LEDS): cv.positive_not_null_int,
+            cv.Required(CONF_RGB_ORDER): cv.enum(RGB_ORDERS, upper=True),
+            cv.Optional(CONF_MAX_REFRESH_RATE): cv.positive_time_period_microseconds,
+            cv.Required(CONF_CHIPSET): cv.one_of(*CHIPSETS, upper=True),
```
jesserockz, line 103: No option for custom timings like `esp32_rmt_led_strip` and `rp2040_pio_led_strip` support?

Mat931: How do you suggest implementing this? On the Beken chips we don't have nearly as much control over the timings as on the ESP32 for example. There are only a few values for the frequency depending on the clock divider. To change the duty cycle of each LED strip bit we have 8 SPI bit positions we can change plus a 9th bit at the end which is always low. To be sure that the timings are compatible with a given LED strip we need to know the acceptable timing range from the datasheet. Should we just let the user figure all that out and expose bit0, bit1 and spi_frequency in YAML? Or should we try to calculate some (probably inaccurate) values for the timings in Python?

jesserockz: So in the other led_strips, the chipset just defines the know bit0/1 high/low values for a specific led chipset. 
But in yaml, a user can define these values directly. 

I have not done a full proper review yet so i'm not sure exactly how the bk chips work in this PR.

esphome/components/beken_spi_led_strip/led_strip.cpp
```diff
@@ -0,0 +1,385 @@
...
+#include "led_strip.h"
+
+#ifdef USE_BK72XX
+
+#include "esphome/core/helpers.h"
+#include "esphome/core/log.h"
+
+extern "C" {
+#include "../../beken378/os/include/rtos_pub.h"
+#include "../../beken378/driver/spi/spi.h"
```
jesserockz: There should not be relative paths to external included files...

jesserockz (changes requested):

esphome/components/beken_spi_led_strip/light.py
```diff
@@ -0,0 +1,133 @@
...
+        supported_pin_info = ", ".join(f"{x}" for x in SUPPORTED_PINS[family])
+        raise cv.Invalid(
+            f"Pin {value} is not supported on the {family}. Supported pins: {supported_pin_info}"
+        )
+    return value
+
+
+def _validate_num_leds(value):
+    max_num_leds = 165  # 170
+    if value.get(CONF_IS_RGBW, False) or value.get(CONF_IS_WRGB, False):
```
jesserockz: Since there is a default of False for both of these, you can just access it directly
```suggestion
    if value[CONF_IS_RGBW] or value[CONF_IS_WRGB]:
```

esphome/components/beken_spi_led_strip/light.py
```diff
@@ -0,0 +1,133 @@
...
+        max_num_leds = 123  # 127
+    if value.get(CONF_NUM_LEDS) > max_num_leds:
```
jesserockz:
```suggestion
    if value[CONF_NUM_LEDS] > max_num_leds:
```

esphome/components/beken_spi_led_strip/light.py
```diff
@@ -0,0 +1,133 @@
...
+        raise cv.Invalid(
+            f"The maximum number of LEDs for this configuration is {max_num_leds}."
```
jesserockz:
```suggestion
            f"The maximum number of LEDs for this configuration is {max_num_leds}.",
            path=CONF_NUM_LEDS,
```

jesserockz (changes requested):
Minor changes below, otherwise good to go I think.

jesserockz (dismissed):

jesserockz (approved):

PR #6526, opened 2024-04-12
Define `USE_PSRAM`

edwardtfn:

This is just an enabler for other (future) changed in other components.
Currently, some components (`esp32_ble_tracker`) use `CONFIG_SPIRAM` definition, however this will depend on the framework version to work properly and therefore is less reliable. It was even a discussion about which flag to use on #4486.
Other components try to allocate PSRAM without checking for the component installed, which is not a problem, but with this `USE_PSRAM` defined it will make easier to adapt the code accordingly.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:

### DIFF
--- a/esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
+++ b/esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
@@ -248,11 +248,11 @@ class ESP32BLETracker : public Component,
   SemaphoreHandle_t scan_result_lock_;
   SemaphoreHandle_t scan_end_lock_;
   size_t scan_result_index_{0};
-#if CONFIG_SPIRAM
+#if USE_PSRAM
   const static u_int8_t SCAN_RESULT_BUFFER_SIZE = 32;
 #else
   const static u_int8_t SCAN_RESULT_BUFFER_SIZE = 16;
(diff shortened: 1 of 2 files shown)


esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
```diff
@@ -248,11 +248,11 @@ class ESP32BLETracker : public Component,
   SemaphoreHandle_t scan_result_lock_;
   SemaphoreHandle_t scan_end_lock_;
   size_t scan_result_index_{0};
-#if CONFIG_SPIRAM
+#ifdef USE_PSRAM
```
jesserockz, line 252: This will also need to then include the defines.h file at the top of this file

jesserockz (changes requested):

esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
edwardtfn: Done!

jesserockz (approved):

PR #6363, opened 2024-03-12
LVGL implementation for ESPHome

clydebarrow:
# What does this implement/fix?

Implements [LVGL](https://docs.lvgl.io/8.3/index.html) in ESPHome in a native way - configured with YAML.

Note: this PR will not be merged, it is replaced by #7184.

It is recommended that you use that PR as an external component in preference to this, which will not be further updated.

To use this PR prior to being merged, add this to your yaml:

```
external_components:
  - source: github://pr#6363
    refresh: 10min
    components: [lvgl]
```

And read the docs as below:

Doc preview is here: https://deploy-preview-3678--esphome.netlify.app/components/lvgl.html

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3678
## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6537, opened 2024-04-15
Implement i80 bus for ili9xxx displays

clydebarrow:

## NOTE

This PR has been changed to draft status, since it is going to be replaced by 3 separate PRs. In the meantime you can use this PR as an external component, e.g. for the W32-SC01-Plus display:

```
external_components:
 - source: github://pr#6537
   components: [ i80, io_bus, ili9xxx, spi ]

i80:
  dc_pin:
    number: 0
  data_pins:
    - 9
    - ignore_strapping_warning: true
      number: 46
    - ignore_strapping_warning: true
      number: 3
    - 8
    - 18
    - 17
    - 16
    - 15
  wr_pin: 47

display:
  - platform: ili9xxx
    bus_type: i80
    id: w32_disp
    model: st7796
    dimensions:
      height: 320
      width: 480
    transform:
      mirror_y: false
      mirror_x: false
      swap_xy: true
    reset_pin: 4
    data_rate: 4MHz
    color_order: bgr
    invert_colors: true
```

## Types of changes

Add a new bus parallel type - `i80` - used for interfacing  LCD display controllers. This is used in a similar way to the `spi` component, and is utilised by the `ili9xxx` display driver, enabling support for the Seeed [W32-SC01-Plus](https://www.seeedstudio.com/WT32-3-5-Inch-Display-p-5542.html) and the [LilyGo T-Display S3](https://www.lilygo.cc/products/t-display-s3?variant=42589373268149).

The changes have been tested on the above as well as numerous SPI interfaced displays previously supported by `ili9xxx` to ensure that nothing has been broken in the process.

- [ ] Other

- https://github.com/esphome/feature-requests/issues/2387
- https://github.com/esphome/feature-requests/issues/901
- 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3761

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
i80:
  dc_pin: 7
  wr_pin: 8
  rd_pin: 9
  data_pins:
    - 39
    - 40
    - 41
    - 42
    -
      number: 45
    -
      number: 46
    - 47
    - 48
    
display:
  - platform: ili9xxx
    bus_type: i80
    cs_pin: 6
    reset_pin: 5
    model: st7789v
```

## Checklist:


Unfocused:
In case this ends up being mentioned in documentation: Seeed is only a distributor for the WT32-SC01 and WT32-SC01 Plus boards. They're developed & made by [Wireless-Tag](https://en.wireless-tag.com/), which makes a few things that'd be compatible with ESPHome.

Unfocused:
Also on the topic of documentation, to avoid confusion the docs should probably mention that i80 is also commonly called 8080 (or more rarely Intel 8080, or Intel Parallel, or some combination of those words). And some specs call it DBI Type B, just for fun.

And worth mentioning this is only only supporting 8bit bus width for now. Someone will inevitably end up trying to use a display that only supports 16bit.

clydebarrow:
The name is discussed in the docs, and it does specifically say it requires 8 data lines, so one would hope that no-one will try it with 16 (which will fail config validation anyway.)

esphome/components/byte_bus/byte_bus.h
```diff
@@ -0,0 +1,71 @@
...
+#include "esphome/core/log.h"
+#include "esphome/core/gpio.h"
+
+namespace esphome {
+namespace byte_bus {
+
+/**
+ * A pin to replace those that don't exist.
+ */
+class NullPin : public GPIOPin {
```
Unfocused: Worth moving the 2 copies of this class to `core/gpio.h` now before there's half a dozen of the things running about? Aside from removing code duplication, it would also avoid having multiple instances of a class meant to represent nothing. One instance of nothing seems adequate.

esphome/components/byte_bus/byte_bus.h
```diff
@@ -0,0 +1,71 @@
...
+};
+
+// https://bugs.llvm.org/show_bug.cgi?id=48040
+static GPIOPin *const NULL_PIN = new NullPin();  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
+
+/**
+ * A class that can write a byte-oriented bus interface with CS and DC controls, typically used for
+ * LCD display controller interfacing.
+ */
+class ByteBus {
```
nielsnl68:
```suggestion
class DataBus {
```
I suggest to make it more widely usable like for 16 bit mode etc.

esphome/components/i80/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+        cv.Schema(
+            {
+                cv.GenerateID(): cv.declare_id(I80Component),
+                cv.Required(CONF_DATA_PINS): cv.All(
+                    cv.ensure_list(pins.internal_gpio_output_pin_number),
+                    cv.Length(min=8, max=8),
+                ),
+                cv.Required(CONF_WR_PIN): pins.internal_gpio_output_pin_schema,
+                cv.Optional(CONF_RD_PIN): pins.internal_gpio_output_pin_schema,
+                cv.Required(CONF_DC_PIN): pins.internal_gpio_output_pin_schema,
```
nielsnl68, line 37: To make this more consistent with the SPI bus i would suggest to move this to the user side of the component (like display).

esphome/components/byte_bus/byte_bus.h
nielsnl68, line 1: The appoach you took is better imho, then that from me. 
I will add my IObus code implementation here below as well. So we have a consistent structure with what i already have done.

esphome/components/byte_bus/byte_bus.h
```diff
@@ -0,0 +1,71 @@
...
+  virtual void bus_setup() = 0;
+
+  virtual void bus_teardown() = 0;
+
+  /**
+   * Write a data array. Will not control dc or cs pins.
+   * @param data
+   * @param length
+   */
+  virtual void write_array(const uint8_t *data, size_t length) = 0;
```
nielsnl68, line 43:
```suggestion
```
Move this to the *protected* section.

esphome/components/byte_bus/byte_bus.h
nielsnl68, line 1: I really like this approach. The only thing i would suggest is to rename it to "*data_bud*" or as i named it "*io_bus*".
To synchronize my work and that from you i added the differences below.

esphome/components/byte_bus/byte_bus.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Write a command followed by data. CS and DC will be controlled automatically.
+   * On return, the DC pin will be in DATA mode.
+   * @param cmd
+   * @param data
+   * @param length
+   */
+  virtual void write_cmd_data(int cmd, const uint8_t *data, size_t length) = 0;
```
nielsnl68, line 52: Move those two to de `protected:` section.

esphome/components/byte_bus/byte_bus.h
```diff
@@ -0,0 +1,71 @@
...
+
```
nielsnl68, line 53:
```suggestion

  inline void send_command(uint8_t command) { this->send_command(command, nullptr, 0); }
  inline void send_command(uint8_t command, uint8_t data) { this->send_command(command, &data, 1); }
  void send_command(uint8_t command, const uint8_t *data, size_t length = 0) {
    this->begin_transaction();
    this->write_command(command);  // Send the command byte
    if (length => 0) {
       this->write_data(data, length);
    }
    this->end_transaction();
  }

  inline void send_byte(const uint8_t data) { this->send_data(&data, 1); }
  inline void send_word(const uint16_t data) { this->send_data(&data, 2); }
  void send_data(const uint8_t *data, size_t length = 1) {
    this->begin_transaction();
    this->write_data(data, length);
    this->end_transaction();     
  }

  virtual void begin_pixels() { this->begin_transaction(); }
  virtual void end_pixels() { this->end_transaction(); }

  virtual void send_pixels(const uint8_t *data, size_t length, display::Rect window = display::Rect() ) { 
    this->send_data(data, length); 
  }

  virtual void dump_config(){};
```

esphome/components/byte_bus/byte_bus.h
```diff
@@ -0,0 +1,71 @@
...
+  void set_dc_pin(GPIOPin *dc_pin) { this->dc_pin_ = dc_pin; }
+
+  /**
+   * Select data (true) or control mode.
+   * @param data True for data mode, false for control
+   */
+  void set_dc_data(bool data) { this->dc_pin_->digital_write(data); }
+
+ protected:
+  GPIOPin *dc_pin_{byte_bus::NULL_PIN};
```
nielsnl68, line 68:
```suggestion
  virtual void start_command(){};
  virtual void end_command(){};
  virtual void start_data(){};
  virtual void end_data(){};
 
  virtual void write_command(uint8_t value) {
    this->start_command();
    this->send_byte(value);
    this->end_command();
  }
  virtual void write_data(const uint8_t *value, size_t length){
    this->start_data();
    if (length == 1) {
      this->send_byte(*value);
    } else {
      this->send_array(value, length);
    }
    this->end_data();
  }
  /**
   * Write a data array. Will not control dc or cs pins.
   * @param data
   * @param length
   */
  virtual void write_array(const uint8_t *data, size_t length) = 0;

  /**
   * Write a command followed by data. CS and DC will be controlled automatically.
   * On return, the DC pin will be in DATA mode.
   * @param cmd
   * @param data
   * @param length
   */
  virtual void write_cmd_data(int cmd, const uint8_t *data, size_t length) = 0;
  GPIOPin *dc_pin_{byte_bus::NULL_PIN};
  bool always_begin_transaction_{false};
  int in_transaction{0};
```

nielsnl68 (changes requested):
This is part of my suggestions. 

Sorry need to clean the below. I did go something wrong.

esphome/components/i80/__init__.py
clydebarrow: Can't do that, they have to be specified when setting up the bus, and must be the same for all displays if there is more than one on the same bus (like CLK and DATA for SPI). Only the CS pin is per-device. The SPI bus doesn't know about the DC pin so it can be per-device for SPI, but not for i80.

esphome/components/byte_bus/byte_bus.h
clydebarrow: No problem with that.

esphome/components/byte_bus/byte_bus.h
clydebarrow: It needs to be public to be callable from the device code (used to send pixel data.)

esphome/components/byte_bus/byte_bus.h
clydebarrow: I'm not sure exactly what the suggestions achieve. The i80 bus sends a command and parameters in a single transaction, automatically switching the DC line as required. While it can send only a command or only data, that will slow things down if it's required be broken up. SPI needs separate transactions since it doesn't directly control DC, but as implemented that is hidden from the display code. 

The whole point of the `ByteBus` interface (or `IOBus`) is that it is very simple. There are just 6 methods required:

```
  virtual void bus_setup();
  virtual void bus_teardown();
  virtual void begin_transaction();
  virtual void end_transaction();
  virtual void write_cmd_data(int cmd, const uint8_t *data, size_t length);
  virtual void write_array(const uint8_t *data, size_t length);

```
`bus_setup()` initialises the bus and `bus_teardown()` releases it (which is actually never done in ESPHome.)

`begin_` and `end_transaction` basically drive the CS pin and surround calls to the remaining two:

`write_cmd_data()` sends commands optionally followed by data parameters of a relatively small number of bytes.

`write_array()` sends large amounts of data (i.e. pixels.)

Direct control of the DC pin is also exposed via `void set_dc_data(bool data);` but the _only_ reason for that is to enable driving the Waveshare Pico-Res-Touch-LCD-3.5 which requires non-standard SPI handling. (I would actually be quite happy to see that display split out into its own component since it does not behave like all the other ilixxx series.)

So I don't see what benefit there is to adding `send_byte`, `send_word` etc. They aren't needed. The SPI bus has all that, but it's used for lots of other things besides these displays (and IMHO its interface should be trimmed down anyway, but that's not going to happen.)

nielsnl68:
> The name is discussed in the docs, and it does specifically say it requires 8 data lines, so one would hope that no-one will try it with 16 (which will fail config validation anyway.)

We can always change the doc's so it could also be used with 16 bit data busses.

esphome/components/byte_bus/byte_bus.h
nielsnl68: Ha,  i use the send_* methods for that. the write_* methods are talking to the interface it self,

esphome/components/byte_bus/byte_bus.h
nielsnl68: The idea behind it is that imho, is that the display driver talks only to the *IObus* (*) and does not care of it talks to a I2C, SPI or I80 bus interface. The bus interface handles all the needed communication. For this the *IObus* class needs to be as flexible as possible. That is why i added this extra methods.

For example the **Pico-Res-Touch-LCD-3.5** should have a specific *IObus* handler that handles the communication specific for the for this display. 

(*) not specific chosen this name here It could also named "DataBus" as above suggested and you liked.

esphome/components/i80/__init__.py
nielsnl68: To be clear, i was only talking about the DC pin here not the others.

esphome/components/ili9xxx/ili9xxx_display.h
```diff
@@ -66,13 +63,13 @@ class ILI9XXXDisplay : public display::DisplayBuffer,
     this->offset_y_ = offset_y;
   }
   void invert_colors(bool invert);
-  virtual void command(uint8_t value);
-  virtual void data(uint8_t value);
-  void send_command(uint8_t command_byte, const uint8_t *data_bytes, uint8_t num_data_bytes);
+  virtual void send_command(uint8_t command_byte, const uint8_t *data_bytes, uint8_t length);
+  void send_command(uint8_t command_byte) { this->send_command(command_byte, nullptr, 0); }
```
nielsnl68: In my imho, this should be handled by the *DataBus*.

esphome/components/ili9xxx/ili9xxx_display.h
```diff
@@ -195,40 +188,51 @@ class ILI9XXXILI9488 : public ILI9XXXDisplay {
...
+        dfun[1] = 0x62;
       }
     }
-    this->command(ILI9XXX_DFUNCTR);
-    this->data(0);
-    this->data(dfun);
-    this->command(ILI9XXX_MADCTL);
-    this->data(mad);
+    this->send_command(ILI9XXX_DFUNCTR, dfun, sizeof dfun);
+    this->send_command(ILI9XXX_MADCTL, &mad, 1);
```
nielsnl68:
```suggestion
    this->bus->send_command(ILI9XXX_DFUNCTR, dfun, sizeof dfun);
    this->bus->send_command(ILI9XXX_MADCTL, &mad, 1);
```

esphome/components/spi/spi.h
```diff
@@ -477,18 +410,77 @@ class SPIDevice : public SPIClient {
...
+
+  void write_cmd_data(int cmd, const uint8_t *data, size_t length) override;
+
+  void dump_config() override;
+  void set_dc_pin(GPIOPin *dc_pin) { this->dc_pin_ = dc_pin; }
+
+ protected:
+  SPIClient *client_;
+  GPIOPin *dc_pin_{byte_bus::NULL_PIN};
+};
```
nielsnl68, line 482:
```suggestion
};

class SPI16DByteBus : public SPIByteBus  {
 public:
 void send_command(uint8_t command_byte, const uint8_t *data_bytes, uint8_t length) override {
    this->begin_transaction();
    this->set_dc_data(false);
    this->write_array(&command_byte, 1);
    this->set_dc_data(true);
    uint8_t buf[2]{};
    if (length != 0) {
      for (size_t i = 0; i != length; i++) {
        buf[1] = *data_bytes++;
        this->write_array(buf, 2);
      }
    }
    this->end_transaction();
  }
}; 
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -25,19 +25,23 @@ void ILI9XXXDisplay::set_madctl() {
...
+  this->bus_->bus_setup();
+  this->reset_pin_->digital_write(false);
+  delay(20);
+  this->reset_pin_->digital_write(true);
+  delay(20);
   this->init_lcd_();
-
   this->set_madctl();
-  this->command(this->pre_invertcolors_ ? ILI9XXX_INVON : ILI9XXX_INVOFF);
+  this->send_command(this->pre_invertcolors_ ? ILI9XXX_INVON : ILI9XXX_INVOFF);
```
nielsnl68:
```suggestion
  this->bus->send_command(this->pre_invertcolors_ ? ILI9XXX_INVON : ILI9XXX_INVOFF);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -228,7 +215,7 @@ void ILI9XXXDisplay::display_() {
     // 16 bit mode maps directly to display format
     ESP_LOGV(TAG, "Doing single write of %zu bytes", this->width_ * h * 2);
     set_addr_window_(0, this->y_low_, this->width_ - 1, this->y_high_);
-    this->write_array(this->buffer_ + this->y_low_ * this->width_ * 2, h * this->width_ * 2);
+    this->bus_->write_array(this->buffer_ + this->y_low_ * this->width_ * 2, h * this->width_ * 2);
```
nielsnl68, line 271:
```suggestion
    this->bus_->send_pixels(this->buffer_ + this->y_low_ * this->width_ * 2, h * this->width_ * 2);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -260,7 +247,7 @@ void ILI9XXXDisplay::display_() {
         idx += 2;
       }
       if (idx == ILI9XXX_TRANSFER_BUFFER_SIZE) {
-        this->write_array(transfer_buffer, idx);
+        this->bus_->write_array(transfer_buffer, idx);
```
nielsnl68, line 304:
```suggestion
        this->bus_->send_pixels(transfer_buffer, idx);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -272,10 +259,10 @@ void ILI9XXXDisplay::display_() {
     }
     // flush any balance.
     if (idx != 0) {
-      this->write_array(transfer_buffer, idx);
+      this->bus_->write_array(transfer_buffer, idx);
```
nielsnl68, line 316:
```suggestion
      this->bus_->send_pixels(transfer_buffer, idx);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -272,10 +259,10 @@ void ILI9XXXDisplay::display_() {
...
     }
   }
-  this->end_data_();
+  this->bus_->end_transaction();
```
nielsnl68, line 319:
```suggestion
  this->bus_->end_pixels();
```

nielsnl68: I'm missing the *start_transaction* above but that should be *this->bus->start_pixels();*  ;)

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -302,58 +289,22 @@ void ILI9XXXDisplay::draw_pixels_at(int x_start, int y_start, int w, int h, cons
   // x_ and y_offset are offsets into the source buffer, unrelated to our own offsets into the display.
   if (x_offset == 0 && x_pad == 0 && y_offset == 0) {
     // we could deal here with a non-zero y_offset, but if x_offset is zero, y_offset probably will be so don't bother
-    this->write_array(ptr, w * h * 2);
+    this->bus_->write_array(ptr, w * h * 2);
```
nielsnl68:
```suggestion
    this->bus_->send_pixels(ptr, w * h * 2);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -302,58 +289,22 @@ void ILI9XXXDisplay::draw_pixels_at(int x_start, int y_start, int w, int h, cons
...
   } else {
     auto stride = x_offset + w + x_pad;
     for (size_t y = 0; y != h; y++) {
-      this->write_array(ptr + (y + y_offset) * stride + x_offset, w * 2);
+      this->bus_->write_array(ptr + (y + y_offset) * stride + x_offset, w * 2);
```
nielsnl68:
```suggestion
      this->bus_->send_pixels(ptr + (y + y_offset) * stride + x_offset, w * 2);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -302,58 +289,22 @@ void ILI9XXXDisplay::draw_pixels_at(int x_start, int y_start, int w, int h, cons
...
     }
   }
-  this->end_data_();
+  this->bus_->end_transaction();
```
nielsnl68, line 383:
```suggestion
  this->bus_->end_pixels();
```
I'm missing the start_transaction has well above but that should be this->bus->start_pixels(); ;)

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -302,58 +289,22 @@ void ILI9XXXDisplay::draw_pixels_at(int x_start, int y_start, int w, int h, cons
...
-
-void ILI9XXXDisplay::reset_() {
-  if (this->reset_pin_ != nullptr) {
-    this->reset_pin_->digital_write(false);
-    delay(20);
-    this->reset_pin_->digital_write(true);
-    delay(20);
-  }
+void ILI9XXXDisplay::send_command(uint8_t command_byte, const uint8_t *data_bytes, uint8_t length) {
+  this->bus_->write_cmd_data(command_byte, data_bytes, length);  // Send the command byte
```
nielsnl68:
```suggestion
```
imho should be handled by the databus

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -375,24 +326,25 @@ void ILI9XXXDisplay::set_addr_window_(uint16_t x1, uint16_t y1, uint16_t x2, uin
...
-  this->data(y2 >> 8);
-  this->data(y2 & 0xFF);
-  this->command(ILI9XXX_RAMWR);  // Write to RAM
-  this->start_data_();
+  uint8_t buf[4];
+  buf[0] = x1 >> 8;
+  buf[1] = x1;
+  buf[2] = x2 >> 8;
+  buf[3] = x2;
+  this->send_command(ILI9XXX_CASET, buf, sizeof buf);
```
nielsnl68:
```suggestion
  this->bus_->send_command(ILI9XXX_CASET, buf, sizeof buf);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -375,24 +326,25 @@ void ILI9XXXDisplay::set_addr_window_(uint16_t x1, uint16_t y1, uint16_t x2, uin
...
+  buf[0] = y1 >> 8;
+  buf[1] = y1;
+  buf[2] = y2 >> 8;
+  buf[3] = y2;
+  this->send_command(ILI9XXX_PASET, buf, sizeof buf);
+  this->send_command(ILI9XXX_RAMWR);
```
nielsnl68:
```suggestion
  this->buf_->send_command(ILI9XXX_PASET, buf, sizeof buf);
  this->buf_->send_command(ILI9XXX_RAMWR);
```

esphome/components/ili9xxx/ili9xxx_display.cpp
```diff
@@ -375,24 +326,25 @@ void ILI9XXXDisplay::set_addr_window_(uint16_t x1, uint16_t y1, uint16_t x2, uin
...
+  this->bus_->begin_transaction();
 }
 
 void ILI9XXXDisplay::invert_colors(bool invert) {
   this->pre_invertcolors_ = invert;
   if (is_ready()) {
-    this->command(invert ? ILI9XXX_INVON : ILI9XXX_INVOFF);
+    this->send_command(invert ? ILI9XXX_INVON : ILI9XXX_INVOFF);
```
nielsnl68:
```suggestion
    this->buf_->send_command(invert ? ILI9XXX_INVON : ILI9XXX_INVOFF);
```

nielsnl68 (changes requested):
this is what i have so far

esphome/components/byte_bus/byte_bus.h
Unfocused: +1 for `data_bud`, or alternatively `data_buddy` or `data_friend`. If there's enough NZ/Australian support, perhaps even `data_mate`

esphome/components/byte_bus/byte_bus.h
Unfocused: +1 for a flexible universal bus abstraction interface with convenience methods centralised to that layer. There are plenty of other display drivers that support multiple buses, with different needs. (As an aside, it would be nice to have 6800 and MIPI DSI eventually too - the upcoming ESP32-P4 has a MIPI DSI peripheral.)

But displays aren't the only possible use-case for this - more and more sensors are supporting support multiple buses too. Admittedly, a lot of the sensors I've come across use different data structures/features for each bus, frustratingly requiring a completely different driver for each bus. But there are some that use the same data structures for different buses (I'd expect this to be becoming more common), which could use a generic abstraction layer to support multiple buses with just a configuration change. And for those, the added convenience methods would help a LOT.

Otherwise, drivers are going be implementing those things themselves - independently, and each a little differently (while also raising the barrier for people adding new drivers). If those convenience methods are going to exist anyway, it's far better to have them centralised to the higher level abstraction layer.

Arguably the convenience methods that the SPI bus interface has should be pushed out to the higher level abstraction layer... but for the SPI API, I think that ship has sailed, wrecked, and sunk to the bottom of the ocean already.

esphome/components/byte_bus/byte_bus.h
clydebarrow: Yes, but that would be s separate PR. Core changes in a new component PR make it unusable as an external component.

clydebarrow:
> Arguably the convenience methods that the SPI bus interface has should be pushed out to the higher level abstraction layer..

I assume you mean things like single-byte writes? There is a case for that, but but beyond single bytes, and a data-less command operation (which should be default arguments, but C++ won't allow that on a virtual function) anything more I think is best left to be implemented when and where it's actually needed. I'm looking at you, `write_array16()`.

esphome/components/byte_bus/byte_bus.h
clydebarrow: https://github.com/esphome/esphome/pull/6611

esphome/components/byte_bus/byte_bus.h
clydebarrow: I've gone with io_bus::IOBus - it's for IO and definitely a bus (can support multiple devices each with a CS pin.)

jesserockz (changes requested):
Please split up this PR into PRs for the separate component changes?

I know its all a big dependency chain, but its much easier to review and merge them individually. :pray:

clydebarrow:
> Please split up this PR into PRs for the separate component changes?

Started - I have a PR for the i80 bus. Will have to wait for that to be merged before generating the others since they will fail against the dev branch until then. So have left the original PR in place as a draft, to support early adopters using it via external components, given this could be a drawn-out process.

bearpawmaxim:
Hello @clydebarrow ! Thank you for your work!
I finally got my WT32-SC01 Plus display working with esphome with the following config:

```
external_components:
  - source: github://pr#6537
    components: [ io_bus, i80, spi, ili9xxx ]

i80:
  - id: i80bus
    dc_pin: GPIO0
    wr_pin: GPIO47
    data_pins:
      - GPIO9
      - GPIO46
      - GPIO3
      - GPIO8
      - GPIO18
      - GPIO17
      - GPIO16
      - GPIO15

display:
  - platform: ili9xxx
    model: ST7789V
    bus_type: i80
    reset_pin: GPIO4
    data_rate: 20MHZ
    pixel_mode: 18bit
    color_order: bgr
    invert_colors: true
    dimensions:
      width: 480
      height: 320
    transform:
      swap_xy: true
    auto_clear_enabled: false
```

But I am a little confused about the fact that it can't work with 40mHz data rate (as with ArduinoGFX and LovyanGFX) and the 250ms full redraw time (when doing `strftime`with `auto_clear_enabled: true`)..
Is there a way to somehow speed up the drawing?

clydebarrow:
Use LVGL for fast drawing: https://github.com/esphome/esphome/pull/6363

And I'm pretty sure LoyanGFX doesn't actually drive it at 40MHz, I believe the ESP32 maxes out at 20MHz I/O clock for parallel transfers. Just because someone wrote 40MHz in the setup doesn't mean it actually achieves that.

bearpawmaxim:
> Use LVGL for fast drawing

Actually, I'm now drawing the UI with it. Thanks!

nagyrobi:
Wow

Cenedd:
@clydebarrow Apologies if this is a daft question - I'm not at a stage yet where I know the difference yet!
I've managed to make a Lilygo T-Display-S3 Touch work with your code - thanks for saving me after buying it assuming it would just work!
I'm trying to get I²S Audio Media Player working on this unit (idea is a sort of media-player alarm clock) but that requires the Arduino framework (docs don't mention it but editor and compiler do).  Conversely your i80 bus requires IDF framework.  Is there anything I can do here or do I need to find a different base unit that has a more supported screen?

Many thanks,
Gareth

clydebarrow:
General support questions are best directed to the Discord channel, but do you need the media player? Can you use the i2s speaker or rtttl buzzer for audio output?

Cenedd:
> General support questions are best directed to the Discord channel, but do you need the media player? Can you use the i2s speaker or rtttl buzzer for audio output?

Thanks Clyde.  I think I've found the Discord channel now so I'll take this there.  I'm not sure I can use i2s speaker....I'm not entirely sure I can use i2s media player to be honest. I'd like to be able to announce to it and also use it as a Spotify Connect. I'm not sure the latter is possible currently but local streaming seemed at least a step closer!

PR #6550, opened 2024-04-17
web_server: Add error and workaround for web_server v3 when local: true

pzich:

Currently using `web_server` with `version: 3` and `local: true` does not work https://github.com/esphome/issues/issues/5692 (you get version 2 instead).

I think the way `local` works should be reworked and decoupled from the static inline build https://github.com/esphome/issues/issues/5704 but in the interim, the described workaround lets you get the latest v3 locally.

This adds a `cv.Invalid()` when `CONF_LOCAL in config and config[CONF_VERSION] == 3` with the following message:

local v3 error

## Types of changes

- [ ] Other

https://github.com/esphome/issues/issues/5692
https://github.com/esphome/issues/issues/5704

Should the workaround go in the docs instead and be linked to from the `cv.Invalid()`? 🤔

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
web_server:
  version: 3
  local: true
```

## Checklist:

### DIFF
--- a/esphome/components/web_server/__init__.py
+++ b/esphome/components/web_server/__init__.py
@@ -55,6 +55,21 @@ def default_url(config):
 def validate_local(config):
     if CONF_LOCAL in config and config[CONF_VERSION] == 1:
         raise cv.Invalid("'local' is not supported in version 1")
+    if CONF_LOCAL in config and config[CONF_VERSION] == 3:
+        raise cv.Invalid(
+            "'local' is not (yet) supported in version 3\n"
+            "\n"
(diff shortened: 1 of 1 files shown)


pzich:
I took a stab at adding support for local v3 in #6563, which seems to work, but `server_index_v3.h` will need to be updated as v3 development continues.

jesserockz:
Closing in favour of #6563

PR #6563, opened 2024-04-17
web_server: Add support for v3 local server_index

### LINKED ISSUE #5692: Web Server version 3 doesn't work with local enabled
### The problem

When using the new "version 3" of the web_server and using "local: true" it only loads the v2 of the web_server component. Removing "local: true" and re-compiling actually provides the v3 of the web_server component.

2024.3.2

Docker

_No response_

ESP32

### Board

_No response_

_No response_

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

pzich:

This adds support for the `local: true` configuration for `web_server` when using `version: 3`. Currently the compilation succeeds, but `v2` is what is actually used.

To accomplish this, I ran the `make_header.sh` script from https://github.com/esphome/esphome-webserver/pull/73 via `npm run deploy` and copied the resulting headers in here. I then added conditional logic to include the headers based on `USE_WEBSERVER_VERSION`

```
#ifdef USE_WEBSERVER_LOCAL
#if USE_WEBSERVER_VERSION == 2
#include "server_index_v2.h"
#elif USE_WEBSERVER_VERSION == 3
#include "server_index_v3.h"
#endif
#endif
```

However, this conditional include alone did not seem sufficient, as the compilation would fail saying:

```
In file included from src/esphome.h:63,
                 from src/main.cpp:3:
src/esphome/components/web_server/server_index_v3.h:8:15: error: redefinition of 'const uint8_t esphome::web_server::INDEX_GZ []'
    8 | const uint8_t INDEX_GZ[] PROGMEM = {
      |               ^~~~~~~~
In file included from src/esphome.h:62,
                 from src/main.cpp:3:
src/esphome/components/web_server/server_index_v2.h:8:15: note: 'const uint8_t esphome::web_server::INDEX_GZ [11586]' previously defined here
    8 | const uint8_t INDEX_GZ[] PROGMEM = {
      |               ^~~~~~~~
```

To mitigate this, I added `#if USE_WEBSERVER_VERSION == ...` blocks to `server_index_v2.h` and `server_index_v3.h` directly. https://github.com/esphome/esphome-webserver/pull/73 modifies the `make_header.sh` script to add these automatically.

## Types of changes

- [ ] Other

fixes https://github.com/esphome/issues/issues/5692
https://github.com/esphome/esphome-webserver/pull/73

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
web_server:
  version: 2
  local: true
```

```yaml
web_server:
  version: 3
  local: true
```

## Checklist:


jesserockz (changes requested):
Please update this PR to use the files from the new release https://github.com/esphome/esphome-webserver/releases/tag/20240418-041055

jesserockz:
> However, this conditional include alone did not seem sufficient, as the compilation would fail saying:

This is because all header files are automatically includes into a generated file named `esphome.h` in the project root. So even if you skip it in the webserver file, it still gets loaded.

pzich:
> Please update this PR to use the files from the new release https://github.com/esphome/esphome-webserver/releases/tag/20240418-041055

Done!

> This is because all header files are automatically includes into a generated file named esphome.h in the project root.

Aah okay I figured that was what was happening, but didn't see where that is setup.

Does that mean that `INDEX_GZ` was always being included even when `local: false`? Hopefully that means some space savings for those not using local!

pzich:
Let me know if there are any other changes you'd like or if this is good to merge.

jesserockz:
> Does that mean that `INDEX_GZ` was always being included even when `local: false`? Hopefully that means some space savings for those not using local!

I would expect that the compiler should be able to remove the INDEX_GZ since it was not being used anywhere, but you never know really.

It could pay to also add `#ifdef USE_WEBSERVER_LOCAL` into the generated headers which would solve this problem entirely.

kevireilly:
Long shot here, but is this potentially related to the `INDEX_GZ` changes we dropped from the v3 revision? I know that I personally was not familiar with what they were doing. Just remembering this potential keyword correlation.
- https://github.com/esphome/esphome-webserver/pull/54/commits/f9c3b93f6f5c60f080a62b95e0c6c57e61fb3156
- https://github.com/esphome/esphome-webserver/pull/54/commits/e5e96efa0da60e0432c5ba91f6146e239eb75e5d

pzich:
> Long shot here, but is this potentially related to the `INDEX_GZ` changes we dropped from the v3 revision?

I think it's sort of related. Essentially currently (before this PR) the `INDEX_GZ` only has one version, and it's based on v2. This utilizes https://github.com/esphome/esphome-webserver/pull/73 to allow different `INDEX_GZ`s to be included based on the `web_server` version chosen.

pzich:
> It could pay to also add `#ifdef USE_WEBSERVER_LOCAL` into the generated headers which would solve this problem entirely.

I just added this here https://github.com/esphome/esphome-webserver/pull/89

Let me know if that looks good, and I can update the `INDEX_GZ` files from a new release.

pzich:
Actually I just realized you're referring to the generated `esphome.h`…

I'm not sure if `USE_WEBSERVER_VERSION` still ends up with a value if `USE_WEBSERVER_LOCAL` is not set, but might be worth having both of these checks anyway.

pzich:
Aha, based on

https://github.com/esphome/esphome/blob/2fa58468939b1319b072f7fc3f123dcc1f916605/esphome/components/web_server/__init__.py#L72

and

https://github.com/esphome/esphome/blob/2fa58468939b1319b072f7fc3f123dcc1f916605/esphome/components/web_server/__init__.py#L147-L159

it looks like `USE_WEBSERVER_VERSION` always gets added, defaulting to 2.

jesserockz:
https://github.com/esphome/esphome-webserver/releases/tag/20240425-205343

pzich:
Updated!

pzich:
Rebased this on latest `dev`. Let me know if there are any other changes needed

jesserockz (approved):

PR #6574, opened 2024-04-18
Fix or filter

swoboda1337:

Docs for the "or" filter say: "Pass forward a value with the first child filter that returns. Below example will only pass forward values that are either at least 1s old or are if the absolute difference is at least 5.0."

The actual implementation would pass forward a value for every filter that returned. That resulted in duplicating the value and sending it up multiple times. This PR fixes that bug. 

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/sensor/filter.cpp
+++ b/esphome/components/sensor/filter.cpp
@@ -360,12 +360,17 @@ OrFilter::PhiNode::PhiNode(OrFilter *or_parent) : or_parent_(or_parent) {}
 
 optional<float> OrFilter::PhiNode::new_value(float value) {
   this->or_parent_->output(value);
-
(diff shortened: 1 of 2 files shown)


ssieb:
I think all the filters should receive the value, but only the first returned value gets published.

swoboda1337:
> I think all the filters should receive the value, but only the first returned value gets published.

Yeah I think that makes sense, updated.

jesserockz (approved):

PR #6576, opened 2024-04-18
[nextion] Use persistent http connection for TFT upload (ESP-IDF)

edwardtfn:

TFT upload now transfers with a single stream, reducing the number of requests to the http server. In fact, this is similar to increasing the previously used chunk size to the file size, so all the data is requested to the http server in a single request, but then transferred to the UART on 4096 bytes chunks (as required by Nextion). If the display requests a different position, then the next request is the new http request (still in a persistent connection).
This will avoid a new http connection per chunk on Nextion when uploading TFT file, improving memory management and reliability of the whole process.
This is only for esp-idf. I will push a similar change to Arduino.
This is part of the breaking down of #6192

## Types of changes

- [X] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:


esphome/components/nextion/nextion_upload_idf.cpp
```diff
@@ -20,152 +19,152 @@ static const char *const TAG = "nextion.upload.idf";
...
-        }
+    ESP_LOGD(TAG, "Successfully allocated upload buffer in PSRAM");
+  }
+#else
+  buffer = (uint8_t *) malloc(4096);
+#endif
+  if (!buffer) {
+    ESP_LOGE(TAG, "Failed to allocate upload buffer");
+    return -1;
+  }
```
jesserockz: Please see ExternalRAMAllocator which does all of this for you already

esphome/components/nextion/nextion_upload_idf.cpp
```diff
@@ -249,68 +252,91 @@ bool Nextion::upload_tft() {
...
-  ESP_LOGD(TAG, "Restarting Nextion");
-  this->soft_reset();
-  vTaskDelay(pdMS_TO_TICKS(1500));  // NOLINT
+  this->ignore_is_setup_ = false;
+
   if (successful) {
     ESP_LOGD(TAG, "Restarting ESPHome");
-    esp_restart();  // NOLINT(readability-static-accessed-through-instance)
+    vTaskDelay(pdMS_TO_TICKS(1500));  // NOLINT
+    esp_restart();                    // NOLINT(readability-static-accessed-through-instance)
```
jesserockz:
```suggestion
    delay(1500);  // NOLINT
    arch_restart();
```

jesserockz (changes requested):
Please also move/duplicate `#define USE_NEXTION_TFT_UPLOAD` in `esphome/core/defines.h` so that the idf upload code is run by clang correctly. Currently it is not.

esphome/components/nextion/nextion_upload_idf.cpp
edwardtfn: That makes things much easier. Thanks for pointing it.

edwardtfn:
> Please also move/duplicate #define USE_NEXTION_TFT_UPLOAD in esphome/core/defines.h so that the idf upload code is run by clang correctly. Currently it is not.

I've changed that in PR #6661 as this wasn't duplicated by any of my previous PRs, so I wanna track this on the side.

jesserockz (approved):

PR #6579, opened 2024-04-18
Add component qesan_ble_remote

Mat931:

Allows receiving QESAN BLE remotes that are included with many tuya lamps, for example [this one](https://github.com/esphome/esphome-devices/pull/689).

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3769

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
external_components:
  - source: github://pr#6579
    components: [qesan_ble_remote]
    refresh: always

esp32_ble_tracker:
  scan_parameters:
    interval: 100ms
    window: 20ms
    active: false

qesan_ble_remote:
  - id: my_remote
    mac_address: 51:45:53:AA:BB:CC
    address: 0x1234

binary_sensor:
  - platform: qesan_ble_remote
    id: my_button
    remote_id: my_remote
    code: 0x7e
```

## Checklist:


esphome/components/qesan_ble_remote/qesan_ble_remote.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include <esp_bt_defs.h>
+
+namespace esphome {
+namespace qesan_ble_remote {
+
+static const char *const TAG = "qesan_ble_remote";
+
+#ifdef USE_BINARY_SENSOR
+bool QesanBinarySensor::on_update_received(uint8_t button_code, bool pressed) {
+  bool match = (button_code == this->button_code_);
```
jesserockz, line 16: This should just `return` if this callback is not for the current button.

jesserockz (changes requested):
I wrote my below comment, but also had a thought;

I see that the `action` is provided by the data, but is it important? This could be the first component (outside of template) to provide native `Event` entities for the buttons. WDYT?

esphome/components/qesan_ble_remote/qesan_ble_remote.cpp
Mat931: With this remote you can only press one button at a time. While holding a button it repeatedly sends the "pressed" message. So if we miss the "released" message and a different button was pressed, we know that this one should be released. If we return early we can't do that check.

Mat931:
@jesserockz You're right, the `action` isn't that important (except maybe for detecting long double presses.) When doing a short press only the "released" action is sent anyway. Implementing Event entities for this component would make sense but I don't really know how to do that yet.

PR #6582, opened 2024-04-18
[nextion] Use persistent http connection for TFT upload (Arduino)

edwardtfn:

TFT upload now transfers with a single stream, reducing the number of requests to the http server. In fact, this is similar to increasing the previously used chunk size to the file size, so all the data is requested to the http server in a single request, but then transferred to the UART on 4096 bytes chunks (as required by Nextion). If the display requests a different position, then the next request is the new http request (still in a persistent connection).
This will avoid a new http connection per chunk on Nextion when uploading TFT file, improving memory management and reliability of the whole process.
This is only for Arduino.  The ESP-IDF equivalent is on #6576.
This is part of the breaking down of #6192.
## Types of changes

- [X] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:


esphome/components/nextion/nextion_upload_arduino.cpp
```diff
@@ -20,124 +21,159 @@ static const char *const TAG = "nextion.upload.arduino";
...
-      continue;
-    }
-
-    http->addHeader("Range", range_header);
+inline uint32_t Nextion::get_free_heap_() {
+#ifdef ESP32
+  return ESP.getFreeHeap();
+#elif defined(USE_ESP8266)
+  return EspClass::getFreeHeap();
+#endif  // ESP32 vs USE_ESP8266
```
jesserockz:
```suggestion
#if defined(USE_ESP32)
  return heap_caps_get_free_size(MALLOC_CAP_INTERNAL);
#elif defined(USE_ESP8266)
  return EspClass::getFreeHeap();
#endif  // USE_ESP32 vs USE_ESP8266
```

esphome/components/nextion/nextion_upload_arduino.cpp
```diff
@@ -203,135 +237,118 @@ bool Nextion::upload_tft() {
...
-  ESP_LOGD(TAG, "Restarting Nextion");
-  this->soft_reset();
+  this->ignore_is_setup_ = false;
+
   if (successful) {
-    delay(1500);  // NOLINT
-    ESP_LOGD(TAG, "Restarting esphome");
+    ESP_LOGD(TAG, "Restarting ESPHome");
+    delay(1500);    // NOLINT
     ESP.restart();  // NOLINT(readability-static-accessed-through-instance)
```
jesserockz: This is abstracted out in "esphome/core/hal.h"
```suggestion
    arch_restart();
```

jesserockz (changes requested):

esphome/components/nextion/nextion_upload_arduino.cpp
```diff
@@ -20,124 +21,159 @@ static const char *const TAG = "nextion.upload.arduino";
...
   }
-  http->end();
-  ESP_LOGN(TAG, "Fetched %d of %d bytes", fetched, this->content_length_);
-
-  // upload fetched segments to the display in 4KB chunks
-  int write_len;
-  for (int i = 0; i < range; i += 4096) {
+#else
+  std::vector<uint8_t> buffer(4096);  // Initialize buffer
+#endif  // ESP32 & USE_PSRAM
```
jesserockz: See ExternalRAMAllocator for here too

jesserockz (approved):

PR #6594, opened 2024-04-21
Limit Rx wait loop time to 3 seconds.

descipher:

Addresses a TODO
Limits an LD2420 send cmd wait time to 3 seconds.
Previously it used a count which could exceed a WDT.  

## Types of changes

Discord

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/ld2420/ld2420.cpp
+++ b/esphome/components/ld2420/ld2420.cpp
@@ -493,19 +493,16 @@ void LD2420Component::handle_ack_data_(uint8_t *buffer, int len) {
 }
 
 int LD2420Component::send_cmd_from_array(CmdFrameT frame) {
+  uint32_t start_millis = millis();
   uint8_t error = 0;
   uint8_t ack_buffer[64];
   uint8_t cmd_buffer[64];
-  uint16_t loop_count;
   this->cmd_reply_.ack = false;
   if (frame.command != CMD_RESTART)
     this->set_cmd_active_(true);  // Restart does not reply, thus no ack state required.
   uint8_t retry = 3;
   while (retry) {
-    // TODO setup a dynamic method e.g. millis time count etc. to tune for non ESP32 240Mhz devices
-    // this is ok for now since the module firmware is changing like the weather atm
     frame.length = 0;
-    loop_count = 1250;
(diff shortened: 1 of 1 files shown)


ssieb:
Since you're modifying the code, can you remove the wait entirely?

jesserockz (dismissed):

descipher:
> Since you're modifying the code, can you remove the wait entirely?

Unfortunately no, its there for a module restart.

ssieb:
I mean remove that hard wait loop and make it non-blocking.

ssieb:
If that change is too difficult, we can just use the simpler change for now.  Let us know if you want to try.

descipher:
> If that change is too difficult, we can just use the simpler change for now. Let us know if you want to try.

Doing a callback method here is significantly complex. It’s not obvious but the general service loop does the serial send. During the development I found that it does not process the Rx immediately in that loop. To solve that issue, I used the main loop to process it. This works very well in reducing blocking time when the 2420 module works as designed. Right now I have too much going on to refactor it and I am not certain what we can gain for it yet so I am inclined to just do the simple tick count for now.

esphome/components/ld2420/ld2420.cpp
```diff
@@ -538,12 +535,12 @@ int LD2420Component::send_cmd_from_array(CmdFrameT frame) {
         this->readline_(read(), ack_buffer, sizeof(ack_buffer));
       }
       delay_microseconds_safe(1450);
-      if (loop_count <= 0) {
+      // Wait on an Rx from the LD2420 for up to 3 1 second loops, otherwise it could trigger a WDT.
```
jesserockz, line 538: unless you reset the start_millis in here it will instantly be true after the first retry?

descipher: Right, While its still in the loop it never clears. Adjusting it now.

GelidusResearch: Done.

jesserockz (approved):

PR #6608, opened 2024-04-23
Display: add diagnostic test_card option

nielsnl68:

When you start using a display, you want to know if it is showing correct etc. 
This PR will show a testcard with the RGB colors, an indicator what the top left corner is and text (R,G,B) to validate how the display oriented.

The main purpose to help people that have issues with there display and want to show what is going on with there display.

Just add the option "`show_testcard: true`" to the display component or add `it.test_card();` to your lambda code.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3796

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
display:
  - platform: ili9xxx
    model: ili9342
    cs_pin: GPIO15
    dc_pin: GPIO2
    show_testcard: true
```

## Checklist:


jesserockz:
As with all ESPHome code, we prefer not to have "magical" things happen like this.

It would make more sense if it was a function to call inside the lambda or a page:
```yaml
lambda: |-
  it.test_card();
```

esphome/components/display/display.cpp
```diff
@@ -511,6 +511,8 @@ void Display::do_update_() {
     this->page_->get_writer()(*this);
   } else if (this->writer_.has_value()) {
     (*this->writer_)(*this);
+  } else {
```
clydebarrow: When LVGL is in use, this would get triggered since there is no lambda. `auto_clear_enabled` is on by default, but should be false when LVGL is used.

```suggestion
  } else if (this->auto_clear_enabled_) {
```

clydebarrow:
Nice idea! A couple of suggestions - as well as the `fade_to_{white,black}` areas (which don't need to be repeated 7 times IMHO) something equivalent to this lambda is useful to diagnose mis-wired RGB interfaces, since it allows identification of individual bits for each color.

```
      it.filled_rectangle(0, 0, 40, 40, Color(0x80, 0, 0));
      it.filled_rectangle(0, 60, 40, 40, Color(0x40, 0, 0));
      it.filled_rectangle(0, 120, 40, 40, Color(0x20, 0, 0));
      it.filled_rectangle(0, 180, 40, 40, Color(0x10, 0, 0));
      it.filled_rectangle(0, 240, 40, 40, Color(0x8, 0, 0));

      it.filled_rectangle(40, 0, 40, 40, Color(0, 0x80, 0));
      it.filled_rectangle(40, 60, 40, 40, Color(0, 0x40, 0));
      it.filled_rectangle(40, 120, 40, 40, Color(0, 0x20, 0));
      it.filled_rectangle(40, 180, 40, 40, Color(0, 0x10, 0));
      it.filled_rectangle(40, 240, 40, 40, Color(0, 0x8, 0));
      it.filled_rectangle(40, 320, 40, 40, Color(0, 0x4, 0));

      it.filled_rectangle(80, 0, 40, 40, Color(0, 0, 0x80));
      it.filled_rectangle(80, 60, 40, 40, Color(0, 0, 0x40));
      it.filled_rectangle(80, 120, 40, 40, Color(0, 0, 0x20));
      it.filled_rectangle(80, 180, 40, 40, Color(0, 0, 0x10));
      it.filled_rectangle(80, 240, 40, 40, Color(0, 0, 0x8));
```
Screenshot 2024-04-23 at 12 23 25 pm

esphome/components/display/display.cpp
clydebarrow: Or what Jesse said.

nielsnl68: Good point, i will follow Jesses suggestion. And make it a callable method.

nielsnl68:
> Nice idea! A couple of suggestions - as well as the `fade_to_{white,black}` areas (which don't need to be repeated 7 times IMHO) something equivalent to this lambda is useful to diagnose mis-wired RGB interfaces, since it allows identification of individual bits for each color.
> 
the reason i used the fade_to is to see of the display driver runs in 8 or 16 bit mode. Having just a couple of blocks does not show that 100%. And it cost not that much time extra. The testcard is just called once after that the poller is stopt.

nielsnl68:
> As with all ESPHome code, we prefer not to have "magical" things happen like this.
> 
> It would make more sense if it was a function to call inside the lambda or a page:
> 
> ```yaml
> lambda: |-
>   it.test_card();
> ```

Yes, i will do this.

clydebarrow:
> the reason i used the fade_to is to see of the display driver runs in 8 or 16 bit mode. Having just a couple of blocks does not show that 100%.

Each repetition is the same, I'm suggesting keeping that, with say 2 repetitions over half the area, add the single-bit colours in the other half.  This number of copies of the same gradients seems unnecessary:
Screenshot 2024-04-23 at 12 38 56 pm

nielsnl68:
> > the reason i used the fade_to is to see of the display driver runs in 8 or 16 bit mode. Having just a couple of blocks does not show that 100%.
> 
> Each repetition is the same, I'm suggesting keeping that, with say 2 repetitions over half the area, add the single-bit colours in the other half. This number of copies of the same gradients seems unnecessary.

Is this how you see it on your display?

esphome/components/display/display.cpp
```diff
@@ -608,6 +608,66 @@ bool Display::clamp_y_(int y, int h, int &min_y, int &max_y) {
...
+  int w = get_width(), h = get_height(), image_w, image_h;
+  if (this->get_display_type() == DISPLAY_TYPE_COLOR) {
+    Color R(255, 0, 0), G(0, 255, 0), B(0, 0, 255);
+    if (w > h) {
+      image_w = w > 330 ? 310 : w - 20;
+      image_h = h > 280 ? 255 : h - 20;
+    } else {
+      image_h = h > 330 ? 310 : h - 20;
+      image_w = w > 280 ? 255 : w - 20;
+    }
```
clydebarrow: When image_h is > 255 the value passed to esp_scale (which takes 8 bit arguments) is truncated.
```suggestion
    image_w = std::min(w - 20, 310);
    image_h = std::min(h - 20, 255);
```

clydebarrow:
> Is this how you see it on your display?

Yes. Found the reason - https://github.com/esphome/esphome/pull/6608/files#r1575898035

nielsnl68:
> > Is this how you see it on your display?
> 
> Yes. Found the reason - https://github.com/esphome/esphome/pull/6608/files#r1575898035

Okay thanks. your suggestion is added.

nielsnl68:
Thinking about it a bit more i decided to make it much easier to popup the testcard.
Now you can just add `show_testcard: true` to the display component configuration and it shows up at the first moment the display is updated.

esphome/components/display/__init__.py
```diff
@@ -38,6 +38,7 @@
 )
 
 CONF_ON_PAGE_CHANGE = "on_page_change"
+CONF_SHOW_TESTCART = "show_testcart"
```
clydebarrow: Should be `CONF_SHOW_TESTCARD`

esphome/components/display/__init__.py
```diff
@@ -113,6 +115,8 @@ async def setup_display_core_(var, config):
         await automation.build_automation(
             trigger, [(DisplayPagePtr, "from"), (DisplayPagePtr, "to")], conf
         )
+    if (CONF_SHOW_TESTCART in config) and (config[CONF_PAGES] is True):
```
clydebarrow: Why only if PAGES is true? 

This should probably only be permitted if neither PAGES nor LAMBDA are set.

And since it's a boolean, if set to False the card should not be shown. As written, `show_testcart: false` will still show the card.

clydebarrow (changes requested):

esphome/components/display/__init__.py
nielsnl68: Oeps, [dutch] foutje bedankt [/dutch] that should have been `CONF_SHOW_TESTCART`

nielsnl68:
> > Is this how you see it on your display?
> 
> Yes. Found the reason - https://github.com/esphome/esphome/pull/6608/files#r1575898035

Clyde could you hummer me, why your fix works better then what i did? I just cant understand. And i want to learn from my mistake. :D 
while i was testing it the **c** value nicely from 0 to 255. But when i see your image it looks as if it did do the same loop multiple times.

clydebarrow:
> Clyde could you humour me, why your fix works better then what i did?

In your original code you had this:
```
    if (w > h) {
      image_w = w > 330 ? 310 : w - 20;
      image_h = h > 280 ? 255 : h - 20;
    } else {
      image_h = h > 330 ? 310 : h - 20;
      image_w = w > 280 ? 255 : w - 20;
    }
```
So `image_h` could be (and was in my test cases) greater than 255. But, you then passed that value to `esp_scale()`:
```
      int c = esp_scale(i, image_h);

```
and the arguments to `esp_scale()` are defined as `uint8_t`:
```
inline static uint8_t esp_scale(uint8_t i, uint8_t scale, uint8_t max_value = 255) { return (max_value * i / scale); }
```

So e.g. a value for `image_h` of 280 gets truncated to 8 bits and would be seen by `esp_scale()` as 24, so as you go around the loop 280 times, the colour value cycles within a range of only 24 distinct values, so the pattern gets repeated numerous times.

Any portrait mode with height greater than 255 would fail to display as expected (as would landscape mode between 256 and 280.) So I assume you did not test in portrait mode, or you used a really small display.

nielsnl68:
thanks for the explanation. I was looking at the wrong place. and could not figure out what was wrong.

esphome/components/display/__init__.py
```diff
@@ -113,6 +115,8 @@ async def setup_display_core_(var, config):
         await automation.build_automation(
             trigger, [(DisplayPagePtr, "from"), (DisplayPagePtr, "to")], conf
         )
+    if (CONF_SHOW_TEST_CARD in config) and (config[CONF_SHOW_TEST_CARD] is True):
```
clydebarrow:
```suggestion
    if config.get(CONF_SHOW_TEST_CARD):
```

esphome/components/display/__init__.py
```diff
@@ -82,6 +83,7 @@ def validate_rotation(value):
             }
         ),
         cv.Optional(CONF_AUTO_CLEAR_ENABLED, default=True): cv.boolean,
+        cv.Optional(CONF_SHOW_TEST_CARD): cv.boolean,
```
clydebarrow, line 86: There should really be a check here that `show_test_card` is not true if a lambda has been provided.

clydebarrow (changes requested):

esphome/components/display/__init__.py
nielsnl68: The show_test_card flag overrules the lambda and pages. So both can exist.

clydebarrow (approved):
Working fine here.

nielsnl68:
> Working fine here.

Could you make a picture to add to the Display doc?

PR #6611, opened 2024-04-23
Add null GPIO pin 

clydebarrow:

This is for use in places where no pin has been specified, and avoids having to check for null pointers all the time.

So allows this code:

```
  if (this->sleep_pin_ != nullptr) {
    this->sleep_pin_->setup();
    this->sleep_pin_->digital_write(false);
    this->sleep_pin_state_ = false;
  }

...

protected:
  GPIOPin *sleep_pin_{nullptr};

```

To be refined to:

```
  this->sleep_pin_->setup();
  this->sleep_pin_->digital_write(false);
  this->sleep_pin_state_ = false;

...

protected:
  GPIOPin *sleep_pin_{NULL_PIN};
```

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/cst226/touchscreen/cst226_touchscreen.cpp
+++ b/esphome/components/cst226/touchscreen/cst226_touchscreen.cpp
@@ -5,17 +5,13 @@ namespace cst226 {
 
 void CST226Touchscreen::setup() {
   esph_log_config(TAG, "Setting up CST226 Touchscreen...");
-  if (this->reset_pin_ != nullptr) {
-    this->reset_pin_->setup();
-    this->reset_pin_->digital_write(true);
-    delay(5);
-    this->reset_pin_->digital_write(false);
-    delay(5);
-    this->reset_pin_->digital_write(true);
-    this->set_timeout(30, [this] { this->continue_setup_(); });
-  } else {
-    this->continue_setup_();
-  }
+  this->reset_pin_->setup();
+  this->reset_pin_->digital_write(true);
+  delay(5);
+  this->reset_pin_->digital_write(false);
+  delay(5);
+  this->reset_pin_->digital_write(true);
+  this->set_timeout(30, [this] { this->continue_setup_(); });
 }
 
 void CST226Touchscreen::update_touches() {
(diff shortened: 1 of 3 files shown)


esphome/components/cst226/touchscreen/cst226_touchscreen.cpp
```diff
@@ -5,17 +5,13 @@ namespace cst226 {
...
-  } else {
-    this->continue_setup_();
-  }
+  this->reset_pin_->setup();
+  this->reset_pin_->digital_write(true);
+  delay(5);
+  this->reset_pin_->digital_write(false);
+  delay(5);
+  this->reset_pin_->digital_write(true);
+  this->set_timeout(30, [this] { this->continue_setup_(); });
```
jesserockz, line 14: So in order to avoid a simple nullptr check (Which already has been coded btw), you would rather the setup be delayed by ~40ms regardless because of the delays and timeout?

The majority of the time it is more likely to be more useful to know when the pin is not actually set.

Correct me if I am wrong there though...

clydebarrow: Yes, the 40ms delay is of little consequence since it only occurs once, and is required if there *is* a pin anyway. I could even argue there's a benefit in having consistent behaviour with or without.

There's a significant benefit in code reliability and readability in general by avoiding the possibility of null pointer dereferences. If you've spent any time coding in a language that enforces null-safety you will appreciate the benefits of knowing that a pointer will never be null. C++ doesn't natively support null-safety, but the principles are still applicable and the benefits the same.

> The majority of the time it is more likely to be more useful to know when the pin is not actually set.

I'd disagree with that. Most of the time all you need to know is that the pin does its job.

I'm not suggesting that there should be a wholesale rewrite of existing code to make use of this, just want it to be available where it's useful. The CST226 usage is there as an example.

jesserockz (approved):

ssieb:
I really don't like this change and I don't see the benefit.  In most cases, there is different behaviour if the pin is null vs just ignoring it.  We use this `nullptr` pattern everywhere, it's C++.

PR #6618, opened 2024-04-24
Fix for #6614- use background_color, improve anti-aliasing

clydebarrow:

For anti-aliased fonts, there _is_ a background colour available in the menu code, so use it instead of inverting the foreground colour - should have been in #6614.

Also always draw a background for a menu item.

Improve rendering of anti-aliased fonts.

The net effect is to change this:
Screenshot 2024-04-24 at 11 16 31 am

to this:

Screenshot 2024-04-24 at 11 16 02 am

## Types of changes

- [] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] Host

```yaml
# Example config.yaml
color:
  - id: light_red
    hex: FF3340
  - id: light_blue
    hex: 3340FF

graphical_display_menu:
  id: my_graphical_display_menu
  background_color: light_blue
  foreground_color: light_red
  display: vnc_display
  on_redraw:
    then:
      component.update: vnc_display
  active: false
  font: roboto
  mode: rotary
  items:
    - type: Label
      text: Menu Item 1
    - type: Label
      text: Menu Item 2
```

## Checklist:


esphome/components/graphical_display_menu/graphical_display_menu.cpp
```diff
@@ -219,9 +219,7 @@ inline void GraphicalDisplayMenu::draw_item(display::Display *display, const dis
   // int background_width = std::max(bounds->width, available_width);
   int background_width = bounds->w;
 
-  if (selected) {
-    display->filled_rectangle(bounds->x, bounds->y, background_width, bounds->h, background_color);
-  }
+  display->filled_rectangle(bounds->x, bounds->y, background_width, bounds->h, background_color);
```
MrMDavidson, line 225: I've not tried this with multi-colour displays, only with eInk, but was this required? I can't recall the specific timing but found that painting the non-selected item's background was a bit of a performance hit. Are we better to do a single `display->filled_rectangle(.... this->background_color_)` in `draw_menu_internal_` ?

clydebarrow: Well, the problem is that if you don't fill the background, then the background is whatever happens to be underneath it, IOW the background of the menu item is transparent. If the underlying background colour happens to be the same as the menu item's foreground colour, the text disappears.

The way the original code is written it appears that it was intended that foreground and background be swapped for selected and non-selected items. I assumed that was intentional though the docs are silent on the matter.
```
  const auto background_color = selected ? this->foreground_color_ : this->background_color_;
  const auto foreground_color = selected ? this->background_color_ : this->foreground_color_;
```

As the code stood it drew a background (using the foreground colour) only for the selected item.

If the intention is to use the same background for all items, then there would need to be a different foreground colour for selected and non-selected items.

MrMDavidson: > Well, the problem is that if you don't fill the background, then the background is whatever happens to be underneath it, IOW the background of the menu item is transparent. If the underlying background colour happens to be the same as the menu item's foreground colour, the text disappears.

So what I'm thinking to solve that is (pseudocode)
```
draw_menu_internal_(...) {
  display->filled_rectangle(0, 0, full_menu_bounds.w, full_menu_bounds.h, this->background_color);
  foreach (auto item : this->menu_items_) {
    const auto background_color = item->selected ? this->foreground_color_ : this->background_color_;
    const auto foreground_color = item->selected ? this->background_color_ : this->foreground_color_;

    if (selected) {
      display->filled_rectangle(bounds->x, bounds->y, background_width, bounds->h, background_color);
    }
  }
}
```

This is just my recollection that this did have a performance impact. I can test locally over the weekend.

> The way the original code is written it appears that it was intended that foreground and background be swapped for selected and non-selected items. I assumed that was intentional though the docs are silent on the matter.

Yeah, that's accurate. Basically inverting the menu item, in the same way the LCD Matrix menu did. It may be worth extending this to have distinction on how selected items are drawn.

esphome/components/graphical_display_menu/graphical_display_menu.cpp
```diff
@@ -166,7 +168,8 @@ void GraphicalDisplayMenu::draw_menu_internal_(display::Display *display, const
   // Render the items into the view port
   display->start_clipping(*bounds);
 
-  int y_offset = bounds->y;
+  display->filled_rectangle(bounds->x, bounds->y, max_width, total_height, this->background_color_);
```
MrMDavidson, line 171: Should this be `max_width` or `bounds->w` ? `max_width` exceeding `bounds->w` is fine because of the call to `start_clipping` but thinking of a scenario where a user scrolls the menu, or a menu item changes to shorten its label, so that `max_width` is narrower than the entire width the menu has and doesn't appropriately repaint the prior background colour?

clydebarrow: I have no idea what `bounds` is meant to represent or where it comes from - in my test code it's the entire screen. The existing code only updates the background of the selected item, and only to the declared width of that item. These changes simply make sure that the background of the unselected items is also drawn, to the same width. If `bounds` is smaller, then it gets clipped. If it's larger, anything inside `bounds` but outside the menu is untouched, just like it always was.

I think I've gone as far as I have the time or inclination at this point - I don't use the component, was just asked to make anti-aliased fonts draw nicely in it.

MrMDavidson: `bounds` will be different depending on if its in "Pop Up" and "Advanced" drawing mode. I can do a test case in the next day or so.

MrMDavidson: Okay, I've setup a test case and at least on a black and white screen I'm not seeing the issue I was concerned with.

clydebarrow: So we're good to go?

MrMDavidson: Yup, all good on my end.

jesserockz (approved):

PR #6621, opened 2024-04-24
Revert "Add null GPIO pin "

jesserockz:
Reverts esphome/esphome#6611

### DIFF
--- a/esphome/components/cst226/touchscreen/cst226_touchscreen.cpp
+++ b/esphome/components/cst226/touchscreen/cst226_touchscreen.cpp
@@ -5,13 +5,17 @@ namespace cst226 {
 
 void CST226Touchscreen::setup() {
   esph_log_config(TAG, "Setting up CST226 Touchscreen...");
-  this->reset_pin_->setup();
-  this->reset_pin_->digital_write(true);
-  delay(5);
-  this->reset_pin_->digital_write(false);
-  delay(5);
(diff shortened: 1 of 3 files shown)


clydebarrow:
Any particular reason for reverting this? I can't imagine anything broke.

I note the objection from @ssieb though it appears to boil down to "that's not how we do it round here". I'd emphasise again that nothing changes unless someone chooses to use this facility, but I _do_ see the benefit.

The inherent difficulty of proper unit or integration testing in ESPHome suggests to me that any programming style that is inherently less prone to errors should at least be considered. C++ has some major design flaws that can't be fixed, but can be worked around.

Greater minds than mine don't like null pointers; from https://en.wikipedia.org/wiki/Void_safety

> In a 2009 talk, Tony Hoare traced the invention of the null pointer to his design of the ALGOL W language and called it a "mistake":
> 
> I call it my billion-dollar mistake. It was the invention of the null reference in 1965. At that time, I was designing the first comprehensive type system for references in an object oriented language (ALGOL W). My goal was to ensure that all use of references should be absolutely safe, with checking performed automatically by the compiler. But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement. This has led to innumerable errors, vulnerabilities, and system crashes, which have probably caused a billion dollars of pain and damage in the last forty years.[1]
>

PR #6623, opened 2024-04-24
Allow platform dependencies

kbx81:

Small PR that enables the ability to specify platform dependencies. For example:

```python
DEPENDENCIES = ["ota.esphome"]
```

...would require the platform `esphome` be present within the `ota` component configuration:

```yaml
ota:
  - platform: esphome
```

This enables more user-friendly validation errors should a required platform not be present.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/config.py
+++ b/esphome/config.py
@@ -450,23 +450,32 @@ class MetadataValidationStep(ConfigValidationStep):
         success = True
         for dependency in self.comp.dependencies:
             dependency_parts = dependency.split(".")
-            component_dep = dependency_parts[-1]
-            platform_dep = dependency_parts[0]
+            if len(dependency_parts) > 2:
+                result.add_str_error(
+                    "Dependencies must be specified as a single component or in component.platform format only",
+                    self.path,
+                )
+                return
+            component_dep = dependency_parts[0]
+            platform_dep = dependency_parts[-1]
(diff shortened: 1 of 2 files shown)


nielsnl68:
Nice idea @kbx81, is there a reason why you choose "esphome.ota" and not "ota.esphome"? To me the last version makes more sense to me. But maybe you have something else in mind that makes it easier to extend this later on?

kbx81:
> Nice idea @kbx81, is there a reason why you choose "esphome.ota" and not "ota.esphome"? To me the last version makes more sense to me. But maybe you have something else in mind that makes it easier to extend this later on?

It was sort of arbitrary -- but it seems we use this arrangement for the logger tags, so I just copied that. It's trivial to change, so if we agree we should swap platform.component for component.platform, I've got no issues with that. 😄

kbx81:
I have switched it to use `component.platform` syntax as I generally think it makes more sense here. Also added some additional logic to better handle various types of config errors.

jesserockz (approved):

PR #6631, opened 2024-04-25
proceed if AP mode is set up

### LINKED ISSUE #3956: Web Server not working on AP mode when a WiFi network is configured
### The problem

The web server does not seem to be started on AP mode when there is a WiFi network configured but out of reach. When not configuring a WiFi network and only AP mode the server is normally started.

My use case is a portable heater that can be used with home assistant when in reach of the network but can still be used with the build in web server if not. This does not seem possible right now and I would have to choose to either only allow home assistant access or only access via the build in AP.

As the web server does indeed work over the AP when no network is configured this seems like a bug and not like an intended limitation?

 v2022.12.3

Docker

2022.12.8

ESP32

### Board

_No response_

Web Server

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

ssieb:

If wifi is configured but can't connect and the AP comes up, the web server and other services are not available.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/3956

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/wifi/wifi_component.cpp
+++ b/esphome/components/wifi/wifi_component.cpp
@@ -694,7 +694,7 @@ void WiFiComponent::retry_connect() {
 }
 
 bool WiFiComponent::can_proceed() {
-  if (!this->has_sta() || this->state_ == WIFI_COMPONENT_STATE_DISABLED) {
+  if (!this->has_sta() || this->state_ == WIFI_COMPONENT_STATE_DISABLED || this->state_ == WIFI_COMPONENT_STATE_AP) {
     return true;
   }
   return this->is_connected();
--- a/esphome/components/wifi/wifi_component.cpp
+++ b/esphome/components/wifi/wifi_component.cpp
@@ -694,7 +694,7 @@ void WiFiComponent::retry_connect() {
 }
 
 bool WiFiComponent::can_proceed() {
-  if (!this->has_sta() || this->state_ == WIFI_COMPONENT_STATE_DISABLED || this->state_ == WIFI_COMPONENT_STATE_AP) {
+  if (!this->has_sta() || this->state_ == WIFI_COMPONENT_STATE_DISABLED || this->ap_setup_) {
     return true;
   }
   return this->is_connected();


Hiekkaharju:
No success. See attached. 
No sign of web server starting, even when AP is connected.
I believe starting of Captive portal is done the way it is for a reason.
[test2.yaml.txt](https://github.com/esphome/esphome/files/15107202/test2.yaml.txt)
[test2_logs.txt](https://github.com/esphome/esphome/files/15107210/test2_logs.txt)

ssieb:
Try again.  I've tested this one and it works.

Hiekkaharju:
Thanks for yor effort. This version indeed seems to start the web server properly with or without STA-wifi presence.

BUT strangely enough I found one sequence that does not now work:
-reset the device when STA-wifi is not available - AP and webserver will start after ap_timeout
-make STA wifi available - device connects to STA-wifi and stops AP
-make STA wifi unavailable - device starts to scan for wifi. BUT it does not start AP after the ap_timeout time. Web server likely would be still running, but it can't be accessed as the AP is not. 

The strangest part is that if the device is started with STA-wifi available, i.e AP is not started before STA-wifi is connected, it will start AP when STA-wifi is lost.

I tested with esphome 2024.4.1 + pr#6631, but I can't see anything in 2024.4.2 changelog that could affect this.

ssieb:
What happens in that scenario without this change?

Hiekkaharju:
As you apparently suspected, the same happens also with pure 2024.4.1. (exept of course web server starts only when STA-wifi is connected)
I created a new issue for the the ap restart
https://github.com/esphome/issues/issues/5759
Thus I think that this PR would be ready for review, merge and release.
Thanks !

jesserockz (approved):

PR #6633, opened 2024-04-25
BedJet: expose the outlet temperature on the climate and as a sensor

javawizard:

- The `bedjet` climate platform now takes a `temperature_source` option that can be used to tell it whether to use the temperature of the air being discharged by the BedJet or the ambient temperature of the room it's in as the source of the climate entity's current temperature
- Both temperatures can now be exposed via `sensor` platform entities
- The `bedjet_id:` option is now optional on the climate entity (and the newly-introduced sensor entities) when only one `bedjet:` hub is defined

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/3799

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

bedjet:
  # ...

climate:
  - platform: bedjet
    name: My BedJet
    temperature_source: outlet

sensor:
  - platform: bedjet
    ambient_temperature:
      name: "My BedJet Ambient Temperature"
    outlet_temperature:
      name: "My BedJet Outlet Temperature"

```

## Checklist:

### DIFF
--- a/esphome/components/bedjet/__init__.py
+++ b/esphome/components/bedjet/__init__.py
@@ -31,7 +31,7 @@ CONFIG_SCHEMA = (
 
 BEDJET_CLIENT_SCHEMA = cv.Schema(
     {
-        cv.Required(CONF_BEDJET_ID): cv.use_id(BedJetHub),
+        cv.GenerateID(CONF_BEDJET_ID): cv.use_id(BedJetHub),
     }
 )
 
--- a/esphome/components/bedjet/bedjet_const.h
+++ b/esphome/components/bedjet/bedjet_const.h
@@ -40,6 +40,14 @@ enum BedjetHeatMode {
   HEAT_MODE_EXTENDED,
 };
 
+// Which temperature to use as the climate entity's current temperature reading
+enum BedjetTemperatureSource {
+  // Use the temperature of the air the BedJet is putting out
+  TEMPERATURE_SOURCE_OUTLET,
+  // Use the ambient temperature of the room the BedJet is in
+  TEMPERATURE_SOURCE_AMBIENT
+};
+
 enum BedjetButton : uint8_t {
   /// Turn BedJet off
   BTN_OFF = 0x1,
--- a/esphome/components/bedjet/climate/bedjet_climate.cpp
+++ b/esphome/components/bedjet/climate/bedjet_climate.cpp
@@ -8,12 +8,6 @@ namespace bedjet {
 
 using namespace esphome::climate;
 
-/// Converts a BedJet temp step into degrees Celsius.
-float bedjet_temp_to_c(const uint8_t temp) {
-  // BedJet temp is "C*2"; to get C, divide by 2.
-  return temp / 2.0f;
-}
-
 static const std::string *bedjet_fan_step_to_fan_mode(const uint8_t fan_step) {
   if (fan_step < BEDJET_FAN_SPEED_COUNT)
     return &BEDJET_FAN_STEP_NAME_STRINGS[fan_step];
@@ -236,9 +230,14 @@ void BedJetClimate::on_status(const BedjetStatusPacket *data) {
   if (converted_temp > 0)
     this->target_temperature = converted_temp;
 
-  converted_temp = bedjet_temp_to_c(data->ambient_temp_step);
-  if (converted_temp > 0)
+  if (this->temperature_source_ == TEMPERATURE_SOURCE_OUTLET) {
+    converted_temp = bedjet_temp_to_c(data->actual_temp_step);
+  } else {
+    converted_temp = bedjet_temp_to_c(data->ambient_temp_step);
+  }
+  if (converted_temp > 0) {
     this->current_temperature = converted_temp;
+  }
 
   const auto *fan_mode_name = bedjet_fan_step_to_fan_mode(data->fan_step);
   if (fan_mode_name != nullptr) {
--- a/esphome/components/bedjet/sensor/__init__.py
+++ b/esphome/components/bedjet/sensor/__init__.py
@@ -1,0 +1,55 @@
+import logging
+
+import esphome.codegen as cg
+import esphome.config_validation as cv
+from esphome.components import sensor
+from esphome.const import (
+    CONF_ID,
+    DEVICE_CLASS_TEMPERATURE,
+    STATE_CLASS_MEASUREMENT,
+    UNIT_CELSIUS,
+)
+from .. import (
+    BEDJET_CLIENT_SCHEMA,
+    bedjet_ns,
+    register_bedjet_child,
+)
+
+_LOGGER = logging.getLogger(__name__)
+CODEOWNERS = ["@jhansche", "@javawizard"]
+DEPENDENCIES = ["bedjet"]
+
+CONF_OUTLET_TEMPERATURE = "outlet_temperature"
+CONF_AMBIENT_TEMPERATURE = "ambient_temperature"
+
+BedjetSensor = bedjet_ns.class_("BedjetSensor", cg.Component)
+
+CONFIG_SCHEMA = cv.Schema(
+    {
+        cv.GenerateID(): cv.declare_id(BedjetSensor),
+        cv.Optional(CONF_OUTLET_TEMPERATURE): sensor.sensor_schema(
+            unit_of_measurement=UNIT_CELSIUS,
+            device_class=DEVICE_CLASS_TEMPERATURE,
+            state_class=STATE_CLASS_MEASUREMENT,
+        ),
+        cv.Optional(CONF_AMBIENT_TEMPERATURE): sensor.sensor_schema(
+            unit_of_measurement=UNIT_CELSIUS,
+            device_class=DEVICE_CLASS_TEMPERATURE,
+            state_class=STATE_CLASS_MEASUREMENT,
+        ),
+    }
+).extend(BEDJET_CLIENT_SCHEMA)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await register_bedjet_child(var, config)
+
+    if outlet_temperature_sensor := config.get(CONF_OUTLET_TEMPERATURE):
+        sensor_var = await sensor.new_sensor(outlet_temperature_sensor)
+        cg.add(var.set_outlet_temperature_sensor(sensor_var))
+
+    if ambient_temperature_sensor := config.get(CONF_AMBIENT_TEMPERATURE):
+        sensor_var = await sensor.new_sensor(ambient_temperature_sensor)
+        cg.add(var.set_ambient_temperature_sensor(sensor_var))
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -51,6 +51,7 @@ esphome/components/bang_bang/* @OttoWinter
 esphome/components/bedjet/* @jhansche
 esphome/components/bedjet/climate/* @jhansche
 esphome/components/bedjet/fan/* @jhansche
+esphome/components/bedjet/sensor/* @javawizard @jhansche
 esphome/components/bh1750/* @OttoWinter
 esphome/components/binary_sensor/* @esphome/core
 esphome/components/bk72xx/* @kuba2k2
--- a/esphome/components/bedjet/bedjet_codec.cpp
+++ b/esphome/components/bedjet/bedjet_codec.cpp
@@ -157,5 +157,11 @@ bool BedjetCodec::compare(const uint8_t *data, uint16_t length) {
   return explicit_fields_changed;
 }
 
+/// Converts a BedJet temp step into degrees Celsius.
+float bedjet_temp_to_c(uint8_t temp) {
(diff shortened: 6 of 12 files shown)


javawizard:
This is still a draft - I accidentally hit the button early. Sorry for the noise!

esphome/components/bedjet/__init__.py
```diff
@@ -31,7 +33,7 @@
 
 BEDJET_CLIENT_SCHEMA = cv.Schema(
     {
-        cv.Required(CONF_BEDJET_ID): cv.use_id(BedJetHub),
+        cv.GenerateID(CONF_BEDJET_ID): cv.use_id(BedJetHub),
```
javawizard, line 34: @jhansche - this makes the `bedjet_id:` config field optional when there's only one `bedjet:` hub configured.

javawizard:
Alright this is ready for review!

esphome/components/bedjet/bedjet_const.h
```diff
@@ -40,6 +40,14 @@ enum BedjetHeatMode {
   HEAT_MODE_EXTENDED,
 };
 
+// Which temperature to use as the climate entity's current temperature reading
+enum BedjetTemperatureSource {
+  // Use the temperature of the air the BedJet is putting out
+  TEMPERATURE_SOURCE_OUTLET,
```
jhansche, line 46: There have been other discussions about this topic in the past, and I guess my main holdup is that typically a climate entity has the goal of altering the temperature, until the current temp matches the target temp, at which point its job is more or less "done" for now.

When using the outlet temp, it will (or should) 99% of the time match the target temp -- if it doesn't, then that means you have a problem with your BedJet, because it has a fine grained regulator to keep the outlet air pushing a constant temp that matches the target. And with that in mind, it does not just stop blowing air once the target temp is reached.

Personally I don't find that useful, because the two numbers will almost always be aligned.

At the same time, a typical climate entity _will_ influence the ambient air temp, until they are eventually aligned - so in that perspective, showing a climate entity where the ambient temp will never actually move in the direction of the target, is also awkward and not quite consistent with how a normal climate entity would work.

That said, I appreciate that it is a configuration option, and I recognize that my opinion is not the only one out there! 🙌  So thanks for making it configurable 👍 

One thing to keep in mind is that this outlet temp will now be included in the average temp of the room where the ESPHome climate entity is - so if you have the BedJet in a bedroom Room, and an ambient air temp sensor in the same Room, a Room card will display the average of those temps, which will be incorrect when the outlet temp source is chosen. It's a pretty minor detail overall, but might be something we want to call out in the [documentation](https://github.com/esphome/esphome-docs/pull/3799#discussion_r1580043268).

esphome/components/bedjet/climate/bedjet_climate.cpp
```diff
@@ -236,9 +230,18 @@ void BedJetClimate::on_status(const BedjetStatusPacket *data) {
...
+    if (converted_temp > 0) {
+      this->current_temperature = converted_temp;
+    }
+
+  } else if (this->temperature_source_ == TEMPERATURE_SOURCE_AMBIENT) {
+    converted_temp = bedjet_temp_to_c(data->ambient_temp_step);
+    if (converted_temp > 0) {
+      this->current_temperature = converted_temp;
+    }
+  }
```
jhansche: Given the nearly identical code, it might be easier to read with a ternary (or two)? 🤔 
```cpp
auto source_temp = // or reuse converted_temp
  this->temperature_source_ == TEMPERATURE_SOURCE_OUTLET ? data->actual_temp_step
    : this->temperature_source_ == TEMPERATURE_SOURCE_AMBIENT ? data->ambient_temp_step
      : 0;
if (source_temp > 0) {
  this->current_temperature = bedjet_temp_to_c(source_temp)
}
```

esphome/components/bedjet/sensor/__init__.py
jhansche, line 1: minor: I would generally prefer to split these into 2 separate PRs: one for the climate entity temp source, and a 2nd for the new sensor platform.

That said, the sensor addition is surprisingly simple! And it's exactly what I had in mind when refactoring to add the "hub".

esphome/components/bedjet/__init__.py
jhansche: Awesome - great idea. I assume this is automatic in ESPHome, when there is only one instance (and will raise a validation error if there's more than one)?

jhansche (approved):
Thank you for this! The code looks good - with a minor suggestion for reducing code duplication.

esphome/components/bedjet/bedjet_const.h
jhansche: Also note, I could be mistaken about the way a Room's temp gets averaged 🤔  Have you tested that, by chance? If it doesn't influence the room's average temp, then we can ignore that

esphome/components/bedjet/__init__.py
javawizard: Exactly.

esphome/components/bedjet/climate/bedjet_climate.cpp
javawizard: Good call 👍

esphome/components/bedjet/sensor/__init__.py
javawizard: Noted, and totally fair 🙂 

Agree! I love the hub approach, totally feels like the right way to do it.

esphome/components/bedjet/bedjet_const.h
javawizard: > One thing to keep in mind is that this outlet temp will now be included in the average temp of the room where the ESPHome climate entity is

Ok I think I'm totally missing something - I'm not seeing a Room card in my local HA instance's card browser, nor am I finding any other way of viewing an average temperature for a room. Is that a third-party card type? I see https://github.com/marcokreeft87/room-card but that seems to require entities to be listed out manually so I imagine that's not what you're talking about.

Let me know if there's somewhere specific I should be looking in the UI.

(Fun side note: I have temperature sensors in tons of things that are not at ambient temp - like my refrigerator and freezer, my 3D printer, and my laser cutter's fume extractor - so average room temperature readings would likely be totally off for me 🙂 )

> There have been other discussions about this topic in the past, and I guess my main holdup is that typically a climate entity has the goal of altering the temperature, until the current temp matches the target temp, at which point its job is more or less "done" for now.
> 
> When using the outlet temp, it will (or should) 99% of the time match the target temp -- if it doesn't, then that means you have a problem with your BedJet, because it has a fine grained regulator to keep the outlet air pushing a constant temp that matches the target. And with that in mind, it does not just stop blowing air once the target temp is reached.
> 
> Personally I don't find that useful, because the two numbers will almost always be aligned.
> 
> At the same time, a typical climate entity will influence the ambient air temp, until they are eventually aligned - so in that perspective, showing a climate entity where the ambient temp will never actually move in the direction of the target, is also awkward and not quite consistent with how a normal climate entity would work.
> 
> That said, I appreciate that it is a configuration option, and I recognize that my opinion is not the only one out there! 🙌 So thanks for making it configurable 👍

For sure. It's definitely a tricky one: neither temperature source seems to perfectly fit the way a climate entity would work. The argument against using the ambient temperature would of course be that a BedJet doesn't influence the ambient room temp _that much_, or at least that's not its primary purpose, and if for example I put it on cool at the lowest temperature I still want it to put out air even if the ambient temperature drops below the minimum setpoint of 66 degrees.

The reason I like seeing the outlet temperature is that I adjust the temperature on my BedJet quite frequently and often I'll think "hm, I don't feel like I turned it up/down enough, has it reached the target temp yet or is it still adjusting?" and then I'll look and see that it's still adjusting up or down and wait a few moments before making adjustments. It seems to take _just_ long enough to respond that I'm always second guessing it without knowing that - last I checked it takes ~2 minutes for the temperature to drop from 104 degrees down to ambient with the airflow at 20%, presumably because of all the residual heat in the heating block that has to be dissipated.

Anyway. I'm definitely in the same boat as you in recognizing that that's my preference and by no means everyone else's, so configurable it is! 🙂

javawizard:
@jhansche - responded and pushed changes, let me know if you have any more questions!

esphome/components/bedjet/bedjet_const.h
jhansche: > Ok I think I'm totally missing something - I'm not seeing a Room card in my local HA instance's card browser

My bad, it's the Area card, not Room 🤦‍♂️ Example(s)
Screenshot_20240426-102332

jhansche (approved):
Nice, and thanks for the PR!
Just FYI, I can't approve it for merge, but looks good

javawizard:
No worries @jhansche, I assume @jesserockz will be along soon enough to click the button.

Thanks for taking a look at this! (And thanks for writing this awesome component in the first place 😄)

esphome/components/bedjet/bedjet_const.h
javawizard: Ah, got it, no worries!

I just checked the area card and it looks like thermostat current temperatures are _not_ incorporated into an area's average temperature; only temperature sensors are.

esphome/components/bedjet/climate/bedjet_climate.cpp
```diff
@@ -236,9 +230,12 @@ void BedJetClimate::on_status(const BedjetStatusPacket *data) {
   if (converted_temp > 0)
     this->target_temperature = converted_temp;
 
-  converted_temp = bedjet_temp_to_c(data->ambient_temp_step);
-  if (converted_temp > 0)
+  converted_temp = bedjet_temp_to_c(this->temperature_source_ == TEMPERATURE_SOURCE_OUTLET    ? data->actual_temp_step
+                                    : this->temperature_source_ == TEMPERATURE_SOURCE_AMBIENT ? data->ambient_temp_step
+                                                                                              : 0);
```
jesserockz: Multi level ternaries can get messy...
```suggestion
  if (this->temperature_source_ == TEMPERATURE_SOURCE_OUTLET) {
    converted_temp = bedjet_temp_to_c(data->actual_temp_step);
  } else {
    converted_temp = bedjet_temp_to_c(data->ambient_temp_step);
  }
```

jesserockz (changes requested):

esphome/components/bedjet/climate/bedjet_climate.cpp
jhansche: If we don't have the 3rd value, then it doesn't have to be multi-level:
```
  converted_temp = bedjet_temp_to_c(
    this->temperature_source_ == TEMPERATURE_SOURCE_OUTLET
      ? data->actual_temp_step
      : data->ambient_temp_step)
```
Since the source has to be one or the other, we don't have to fall back to 0.

But the if/else also works. My main motivation in the original [suggestion](https://github.com/esphome/esphome/pull/6633#discussion_r1580033366) was to reduce duplication. The if/else version is also easy to read

jesserockz: If it has to even be multiline, it shouldnt be a ternary. The compiler will optimise it in the end, so I would rather it be easier to read.

javawizard: Ha. No strong feelings myself either way!

Change accepted, this is ready for review again.

jhansche (approved):
LGTM 
@javawizard not sure if Jesse filters his view, but it might help to dismiss/re-request review to show that you've made the changes

esphome/components/bedjet/sensor/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+).extend(BEDJET_CLIENT_SCHEMA)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await register_bedjet_child(var, config)
+
+    if CONF_OUTLET_TEMPERATURE in config:
+        sensor_var = await sensor.new_sensor(config[CONF_OUTLET_TEMPERATURE])
```
jesserockz:
```suggestion
    if outlet_temperature_sensor := config.get(CONF_OUTLET_TEMPERATURE):
        sensor_var = await sensor.new_sensor(outlet_temperature_sensor)
```

esphome/components/bedjet/sensor/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+        cg.add(var.set_outlet_temperature_sensor(sensor_var))
+
+    if CONF_AMBIENT_TEMPERATURE in config:
+        sensor_var = await sensor.new_sensor(config[CONF_AMBIENT_TEMPERATURE])
```
jesserockz:
```suggestion
    if ambient_temperature_sensor := config.get(CONF_AMBIENT_TEMPERATURE):
        sensor_var = await sensor.new_sensor(ambient_temperature_sensor)
```

jesserockz (changes requested):

jhansche (approved):

jesserockz (approved):

PR #6645, opened 2024-04-26
Add OpenTherm component (part 1: communication layer and hub)

olegtarasov:

This PR adds a new component that allows ESPHome to communicate with OpenTherm devices via a hardware bridge.

OpenTherm (OT) is a standard communications protocol used in central heating systems for the communication between a central heating appliances and a thermostatic controller.

EDIT: Right after finishing all the work of fixing linter errors, I saw a small “integration: opentherm“ label and discovered another open pull request: https://github.com/esphome/esphome/pull/3921 😂  

I would still argue for my implementation (which is based on Arthur Rump's work here: https://github.com/arthurrump/esphome-opentherm) for the following reasons:

1. I replaced Ihor Melnik's underlying OpenTherm communication library with code based on Jiří Praus' implementation (https://github.com/jpraus/arduino-opentherm). I would argue that it's conceptually more robust, since it uses hardware timers to sample OpenTherm signal levels, and thus doesn't rely on measuring the amount of milliseconds passed from a previous event. It's also more stable because some timer samples can be safely lost and message would still be correctly decoded, as opposed to Ihor Melnik's implementation, where loss of a single interrupt signal leads to OpenTherm message getting decoded incorrectly. This proved to be an issue when sharing a single ESP device between OpenTherm component and Dallas temperature sensors. And as a final argument, my implementation allows fine-grained protocol debugging, where faulty frames can be logged with precision of a single incorrect bit.
2. Arthur Rump's implementation generates C++ code dynamically, so if some OpenTherm sensors are not used (and most of them are not used, since most boilers don't support the full set of sensors), the code for their handling is simply not generated.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3800

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml

esphome:

esp8266:
  board: d1_mini

opentherm:
  in_pin: 1
  out_pin: 2

number:
  - platform: opentherm
    t_set:

```

## Checklist:

# WARNING! This is a partial PR!

I'm following the advice in https://github.com/esphome/esphome/discussions/7170#discussioncomment-10268270 and splitting this PR into parts. The first part contains communication code only, no sensors.

If you want to try this component out, please use it as an external component from my repository: https://github.com/olegtarasov/esphome-opentherm.


olegtarasov:
@jesserockz Hi! Can you please take a look at this pull request? It adds a new OpenTherm component to control various boilers through a hardware bridge. There is also another open pull request that implements a smaller subset of this functionality: https://github.com/esphome/esphome/pull/3921. I would argue that my PR should be accepted instead, for the following reasons:

1. I use completely async communication by default, where both conversation phases — send and receive — don't block the loop. The other implementation blocks the loop while sending data, while their receive is async.
2. This implementation uses more robust OpenTherm protocol handling which doesn't rely on measuring elapsed timespans by hand. Instead I use a timer to sample IO levels at regular intervals, so the code is more resistant to low-level protocol fluctuations and doesn't suffer when other components block the loop for too long. The other implementation will break if such components are enabled along with it.
3. This implementation is better structured to diagnose low-level protocol failures. You can debug protocol errors with 1-bit precision, which is not the case with the other PR.
4. Arthur Rump's component, on which this PR is based, supports defining control signals as a `numer`, `output` or as an input sensor, whereas the other PR supports only `number` outputs, requiring additional work to connect them to Climate components. So this implementation leads to cleaner config files.
5. This component generates C++ code dynamically, based on what sensors and control signals are used in config. If sensors are not used (and most of them are not on account of many boilers not supporting large portions of OpenTherm standard), their corresponding code is not generated.

jesserockz:
I know literally nothing about OpenTherm myself.

I can only provide a review in terms of ESPHome functionality and code safety/style.

I cannot simply choose one PR over another (apart from saying first in first served), and really need all PR authors to come to a consensus together. The better thing to do would have been to base your work off of the PR that was created first to add extra functionality or at least talk to the author of https://github.com/esphome/esphome/pull/3921 which I see none of.

Please work together with other contributors to make ESPHome better for all.

There is also now #7169 which does OpenTherm in another way, although as per my comment on that PR, I do not like the use of a restrictive Arduino library.

Please use https://github.com/esphome/esphome/discussions/7170 to discuss.

esphome/components/opentherm/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        {
+            cv.GenerateID(): cv.declare_id(OpenthermHub),
+            cv.Required("in_pin"): pins.internal_gpio_input_pin_schema,
+            cv.Required("out_pin"): pins.internal_gpio_output_pin_schema,
+            cv.Optional("ch_enable", True): cv.boolean,
+            cv.Optional("dhw_enable", True): cv.boolean,
+            cv.Optional("cooling_enable", False): cv.boolean,
+            cv.Optional("otc_active", False): cv.boolean,
+            cv.Optional("ch2_active", False): cv.boolean,
+            cv.Optional("sync_mode", False): cv.boolean,
```
jesserockz: All config items should be using consts for the keys: `CONF_CH_ENABLE`

Maybe even better to use existing consts for the pins, `CONF_TX_PIN` / `CONF_RX_PIN`?

esphome/components/opentherm/hub.h
```diff
@@ -0,0 +1,110 @@
...
+  void add_initial_message(MessageId message_id) { this->initial_messages_.insert(message_id); }
+  // Add a request to the set of repeating requests. Note that a large number of repeating
+  // requests will slow down communication with the boiler. Each request may take up to 1 second,
+  // so with all sensors enabled, it may take about half a minute before a change in setpoint
+  // will be processed.
+  void add_repeating_message(MessageId message_id) { this->repeating_messages_.insert(message_id); }
+
+  // There are five status variables, which can either be set as a simple variable,
+  // or using a switch. ch_enable and dhw_enable default to true, the others to false.
+  bool ch_enable = true, dhw_enable = true, cooling_enable = false, otc_active = false, ch2_active = false;
```
jesserockz, line 91: If these values have defaults, they should be in the schema itself so it is visible to users in validation dump.

esphome/components/opentherm/hub.h
```diff
@@ -0,0 +1,110 @@
...
+namespace esphome {
+namespace opentherm {
+
+// OpenTherm component for ESPHome
+class OpenthermHub : public Component {
+ protected:
+  // Communication pins for the OpenTherm interface
+  InternalGPIOPin *in_pin_, *out_pin_;
+  // The OpenTherm interface
+  OpenTherm *opentherm_;
```
jesserockz: Since this is only referenced in one place and never shared anywhere else we can safely use a unique ptr here
```suggestion
  std::unique_ptr<OpenTherm> opentherm_;
```

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+void OpenthermHub::process_response(OpenthermData &data) {
+  ESP_LOGD(OT_TAG, "Received OpenTherm response with id %d (%s)", data.id,
+           opentherm_->message_id_to_str((MessageId) data.id));
+  ESP_LOGD(OT_TAG, "%s", opentherm_->debug_data(data).c_str());
+}
+
+void OpenthermHub::setup() {
+  ESP_LOGD(OT_TAG, "Setting up OpenTherm component");
+  this->opentherm_ = new OpenTherm(this->in_pin_, this->out_pin_);  // NOLINT because hub is never deleted
```
jesserockz:
```suggestion
  this->opentherm_ = make_unique<OpenTherm>(this->in_pin_, this->out_pin_);
```

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,331 @@
...
+  if (!this->opentherm_->initialize()) {
+    ESP_LOGE(OT_TAG, "Failed to initialize OpenTherm protocol. See previous log messages for details.");
+    return;
```
jesserockz, line 62:
```suggestion
		this->mark_failed();
    return;
```

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  // Ensure that there is at least one request, as we are required to
+  // communicate at least once every second. Sending the status request is
+  // good practice anyway.
+  this->add_repeating_message(MessageId::STATUS);
+
+  this->current_message_iterator_ = this->initial_messages_.begin();
+  initialized_ = true;
```
jesserockz: With `this->mark_failed()` setting a boolean is not required.
```suggestion
```

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,331 @@
...
+}
+
+void OpenthermHub::on_shutdown() { this->opentherm_->stop(); }
+
+void OpenthermHub::loop() {
+  if (!initialized_)
+    return;
+
```
jesserockz:
```suggestion
```

esphome/components/opentherm/hub.h
```diff
@@ -0,0 +1,110 @@
...
+  // Indicates if we are still working on the initial requests or not
+  bool sending_initial_ = true;
+  // Index for the current request in one of the _requests sets.
+  std::unordered_set<MessageId>::const_iterator current_message_iterator_;
+
+  uint32_t last_conversation_start_ = 0;
+  uint32_t last_conversation_end_ = 0;
+  OperationMode last_mode_ = IDLE;
+  OpenthermData last_request_;
+  bool initialized_ = false;
```
jesserockz:
```suggestion
```

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,331 @@
...
+  if (sync_mode_) {
+    sync_loop_();
```
jesserockz: Please make sure **all** member variables and function calls have `this->` prefixed.
```suggestion
  if (this->sync_mode_) {
    this->sync_loop_();
```

esphome/components/opentherm/opentherm.h
```diff
@@ -0,0 +1,348 @@
...
+#include <sstream>
+#include <iomanip>
+#include "esphome/core/hal.h"
+#include "esphome/core/log.h"
+
+#if defined(ESP32) || defined(USE_ESP_IDF)
+#include "driver/timer.h"
+#endif
+
+#define OT_TAG "opentherm"
```
jesserockz: For consistency across ESPHome, please use `static const char *const TAG` in the c++ files

esphome/components/opentherm/opentherm.h
```diff
@@ -0,0 +1,348 @@
...
+
+#define readBit(value, bit) (((value) >> (bit)) & 0x01)
+#define setBit(value, bit) ((value) |= (1UL << (bit)))
+#define clearBit(value, bit) ((value) &= ~(1UL << (bit)))
+#define writeBit(value, bit, bitvalue) ((bitvalue) ? setBit(value, bit) : clearBit(value, bit))
```
jesserockz: Please make these regular `constexpr` functions in the c++ file.

esphome/components/opentherm/opentherm.h
```diff
@@ -0,0 +1,348 @@
...
+#ifdef ESP8266
+  // ESP8266 timer can accept callback with no parameters, so we have this hack to save a static instance of OpenTherm
+  static OpenTherm *instance_;
+#endif
+};
+
+template<typename T> void int_to_hex(std::stringstream &stream, T i) {
+  std::ostream out(stream.rdbuf());
+  out << std::showbase << std::setfill('0') << std::setw(sizeof(T) * 2) << std::hex << i;
+}
```
jesserockz: `helpers.h` has an existing `format_hex` function to do this too.
Less duplication is better where possible =)

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,331 @@
...
+  data.valueLB = 0;
+
+  // First, handle the status request. This requires special logic, because we
+  // wouldn't want to inadvertently disable domestic hot water, for example.
+  // It is also included in the macro-generated code below, but that will
+  // never be executed, because we short-circuit it here.
+  if (request_id == MessageId::STATUS) {
+    // NOLINTBEGIN
+    bool const ch_enabled = this->ch_enable &&
+#ifdef OPENTHERM_READ_ch_enable
```
jesserockz: Where do these come from exactly? I don't see them being optionally set even...

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,331 @@
+#include "hub.h"
+
+#include <string>
+
+// Disable incomplete switch statement warnings, because the cases in each
+// switch are generated based on the configured sensors and inputs.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wswitch"
```
jesserockz: Which switch does this apply to? I would prefer not to disable rules, but if it does need to be done, it should only be done around the smallest amount of code possible please.

jesserockz (changes requested):
Here is *a* review. Not fully done but submitting as I need to go out and didnt want to leave it pending.

esphome/components/opentherm/__init__.py
olegtarasov: Extracted strings into constants. As for renaming config items for pins, I would rather leave “IN” and “OUT” as they are, since most adapter boards call them this way. If we call them TX and RX, it might create additional confusion for users.

esphome/components/opentherm/hub.h
olegtarasov: They already do have the same defaults in the schema:

```python
cv.Optional(CONF_CH_ENABLE, True): cv.boolean,
cv.Optional(CONF_DHW_ENABLE, True): cv.boolean,
cv.Optional(CONF_COOLING_ENABLE, False): cv.boolean,
cv.Optional(CONF_OTC_ACCTIVE, False): cv.boolean,
cv.Optional(CONF_CH2_ACTIVE, False): cv.boolean,
cv.Optional(CONF_SYNC_MODE, False): cv.boolean
```

esphome/components/opentherm/hub.h
olegtarasov: Done.

esphome/components/opentherm/hub.cpp
olegtarasov: Done.

esphome/components/opentherm/opentherm.h
olegtarasov: Done.

esphome/components/opentherm/opentherm.h
olegtarasov: Done, but placed them into `opentherm.h`, since they will be reused in `hub.cpp` when we implement sensors.

esphome/components/opentherm/opentherm.h
olegtarasov: Done.

esphome/components/opentherm/hub.cpp
olegtarasov: These defines are generated from Python, and I've excluded that code from this batch — that's why you didn't find it. I removed these conditions for now, and will bring them back when we move to implement sensors.

esphome/components/opentherm/hub.cpp
olegtarasov: This also applies to an auto-generated switch statement which handles sensor requests. I've moved the pragmas more closely to the actual place where that statement will live.

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,278 @@
...
+    this->stop_opentherm_();
+    return;
+  }
+
+  // Listen for the response
+  this->opentherm_->listen();
+  if (!this->spin_wait_(1150, [&] { return this->opentherm_->is_active(); })) {
+    ESP_LOGE(TAG, "Hub timeout triggered during receive");
+    this->opentherm_->stop();
+    this->last_conversation_end_ = millis();
```
jesserockz:
```suggestion
    this->stop_opentherm_();
```

esphome/components/opentherm/hub.cpp
```diff
@@ -0,0 +1,278 @@
...
+}
+
+#define ID(x) x
+#define SHOW2(x) #x
+#define SHOW(x) SHOW2(x)
+
+void OpenthermHub::dump_config() {
+  ESP_LOGCONFIG(TAG, "OpenTherm:");
+  ESP_LOGCONFIG(TAG, "  In: GPIO%d", this->in_pin_->get_pin());
+  ESP_LOGCONFIG(TAG, "  Out: GPIO%d", this->out_pin_->get_pin());
```
jesserockz:
```suggestion
  LOG_PIN("  In: ", this->in_pin_);
  LOG_PIN("  Out: ", this->out_pin_);
```

esphome/components/opentherm/hub.h
jesserockz: Somehow I missed that..haha

jesserockz (changes requested):
I think only 1 thing below stood out to me, otherwise good to go. 
I will probably merge this after the beta is cut in 6-7 days and then we have the full cycle to get further PRs in so this one isnt released by itself.

olegtarasov:
@jesserockz That's great to hear, thank you for the review! Meanwhile, I will start preparing the sensors part.

jesserockz (approved):

PR #6647, opened 2024-04-27
Make fast update intervals in qmc5883l work

tronikos:

qmc5883l supports the following date rates: 10Hz, 50Hz, 100Hz, 200Hz. In order to use 200Hz we need to set update_interval to 5ms see https://github.com/esphome/esphome/blob/9bfb36f58bcdaff1d5b62a2ea303d45a10af291e/esphome/components/qmc5883l/sensor.py#L115
The default loop_interval is 16ms so we can only read values from the magnetometer at best every 16ms. This prevents us sampling the sine wave produced by the magnet rotation inside a 30GPM water meter. Their fastest period is 17.3ms. See https://github.com/tronikos/esphome-magnetometer-water-gas-meter/issues/2

So if the update interval of qmc5883l is smaller than the loop_interval start HighFrequencyLoopRequester.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/core/application.h
```diff
@@ -222,6 +222,8 @@ class Application {
    */
   void set_loop_interval(uint32_t loop_interval) { this->loop_interval_ = loop_interval; }
 
+  uint32_t get_loop_interval() { return this->loop_interval_; }
```
jesserockz:
```suggestion
  uint32_t get_loop_interval() const { return this->loop_interval_; }
```

jesserockz (changes requested):

tronikos:
Actually I'm not sure if this is really needed. Note this is a `PollingComponent` which means the loop time isn't relevant?

I added the following code in `QMC5883LComponent::update`:

```cpp
  static uint32_t prev_millis = millis();
  static uint32_t i = 0;
  if (i % 100 == 1) {
    ESP_LOGI("QMC5883LComponent::update", "actual update interval in millis %d", millis() - prev_millis);
  }
  i++;
  prev_millis = millis();
```

I configured the QMC5883L with `update_interval: 5ms`.
Without the change in this PR, the actual update interval of qmc5883l is: 6ms on ESP32, 10ms on ESP8266.
With the change in this PR, it's 5ms on ESP32, 8ms on ESP8266.

So even without the change in this PR, the update interval is less than the loop time (the `debug` `loop_time` reports 12ms on ESP32) which suggests the loop interval isn't relevant for polling components. But then why is there a small improvement with this PR?

jesserockz:
PollingComponents update functions are still run inside the main loop which is every 16ms by default. 

The high frequency loop requester means the main loop will execute as fast as possible, allowing the scheduled 5ms updates to happen as close as possible to the desired time.

tronikos:
In that case this seems like a desired change.

mkmer:
Wondering : Should we do the same to the HMC5883l?  Seems like the same problem exists there too.

jesserockz (changes requested):
Can we actually break the core change into its own PR please?

tronikos:
Sure. Created https://github.com/esphome/esphome/pull/6666

jesserockz (approved):

PR #6653, opened 2024-04-28
WIP: Add support for multiple wakeword/vad models

kahrendt:

This is still a work in progress! I'd appreciate people testing this out and letting me know about any issues. It will be a breaking change due to changes in the yaml syntax for handling multiple wake word models.

This PR adds several features/performance improvements:

- Multiple wake word models can run simultaneously
  - At most two of the current models can run simultaneously without issue
- Adds support for running a Voice Activity Detection model to potentially reduce certain false accepts
  -  The current VAD model can only run with 1 model at the same time. If you try to run VAD and two models all at once, accuracy will suffer greatly
- Several memory improvements
  - Models are loaded and unloaded as mWW starts and stops to save memory when not actively running
  - All buffers (excluding the ring buffer) are freed when not actively running
  - Ring buffer size is reduced (if it filled up before, there was no chance of ever recovering and so 0.5 s of audio was dropped each time)
  - Makes the tensor arena's default allocated memory smaller. The exact space allocated can be set in the codegen stage.

Todo:

- Update the manifest format
  - Add a field for the necessary tensor arena size (currently the values are hardcoded in ``__init__.py``)
  - Add support for a VAD helper model along with its specific parameters
- Allow users to only enable VAD instead of having to point to a specific manifest file (also requires uploading the VAD model to the appropriate default repository) 
- ?Possibly better handle the VAD code? I have added a new preprocessor directive to only compile the relevant code if it is enabled, but I'm not sure if this is the best way to handle it. **Warning: Enabling/disabling a VAD model will require a full recompile when rebuilding, so if you have a slow computer, this may take awhile!**
- Update the documentation (see the example yaml at the end of this PR in the mean time)
- Fix any bugs people encounter in testing

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** unfinished

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml for multiple models

micro_wake_word:
  on_wake_word_detected:
    - voice_assistant.start: 
  models:
    - model: okay_nabu
      sliding_window_average_size: 5
    - model: hey_jarvis
      probability_cutoff: 0.75

```

```yaml
# Example config.yaml for VAD

micro_wake_word:
  on_wake_word_detected:
    - voice_assistant.start: 
  vad_model: 
    model: https://github.com/kahrendt/microWakeWord/releases/download/model/vad_model.json
    sliding_window_average_size: 2
    threshold:
      upper: 0.95
      lower: 0.5
  models:
    - model: alexa

```

## Checklist:

### DIFF
--- a/esphome/components/micro_wake_word/__init__.py
+++ b/esphome/components/micro_wake_word/__init__.py
@@ -30,6 +30,7 @@ from esphome.const import (
     CONF_USERNAME,
     CONF_PASSWORD,
     CONF_RAW_DATA_ID,
+    CONF_THRESHOLD,
     TYPE_GIT,
     TYPE_LOCAL,
 )
@@ -41,9 +42,13 @@ CODEOWNERS = ["@kahrendt", "@jesserockz"]
 DEPENDENCIES = ["microphone"]
 DOMAIN = "micro_wake_word"
 
+CONF_MODELS = "models"
 CONF_PROBABILITY_CUTOFF = "probability_cutoff"
 CONF_SLIDING_WINDOW_AVERAGE_SIZE = "sliding_window_average_size"
 CONF_ON_WAKE_WORD_DETECTED = "on_wake_word_detected"
+CONF_VAD_MODEL = "vad_model"
+CONF_UPPER = "upper"
+CONF_LOWER = "lower"
 
 TYPE_HTTP = "http"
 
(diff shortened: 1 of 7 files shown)


kahrendt:
Closing this pull request as esphome#7032 implements all of these features and more. Note the new pull request has different syntax than what was used here.

PR #6655, opened 2024-04-29
[template/text] Fix lambda config

asergunov:

The template text platform was not working if `lambda` specified. The problem is `default=False` in schema producing the config option even if user is not specifying it, but on validation step we checking if this option doesn't exists in config.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
text:
  - platform: template
    mode: text
    lambda: |
      return std::string{"Hello"};
```

## Checklist:


esphome/components/template/text/__init__.py
```diff
@@ -24,11 +24,11 @@
 
 def validate(config):
     if CONF_LAMBDA in config:
-        if config[CONF_OPTIMISTIC]:
+        if config[CONF_OPTIMISTIC] and config[CONF_OPTIMISTIC]:
```
clydebarrow: No point in repeating the same test:
```suggestion
        if config[CONF_OPTIMISTIC]:
```

esphome/components/template/text/__init__.py
```diff
@@ -24,11 +24,11 @@
...
             raise cv.Invalid("optimistic cannot be used with lambda")
         if CONF_INITIAL_VALUE in config:
             raise cv.Invalid("initial_value cannot be used with lambda")
-        if CONF_RESTORE_VALUE in config:
+        if CONF_RESTORE_VALUE in config and config[CONF_RESTORE_VALUE]:
```
clydebarrow: Since `CONF_RESTORE_VALUE` has a default, it will *always* be present.
```suggestion
        if config[CONF_RESTORE_VALUE]:
```

clydebarrow (changes requested):

esphome/components/template/text/__init__.py
asergunov: Right. Thank you.

asergunov: Is it difficult to add a test? Could you please point me a file to do that?

clydebarrow: There are already tests for `template`, you can just add your sample config in the file `tests/components/template/common.yaml`. Append to the existing `text:` section.

clydebarrow: When adding a test like this, I recommend adding the test *without* your fix, run the test (`script/test_build_components -c template`) and verify failure, then add the fix, run again and verify success.

clydebarrow (approved):

PR #6659, opened 2024-04-29
Add thermocouple type, autoconversion, averaging, and has_fault to max31856

fhriley:

It adds thermocouple type, autoconversion, averaging, and has_fault to the max31856 sensor while maintaining backwards compatibility with polling mode if the `data_ready_pin` is not specified.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3803

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
sensor:
  - platform: max31856
    id: temperature_1
    name: "Temperature 1"
    has_fault:
      name: Temperature 1 Fault
    cs_pin: GPIO12
    data_ready_pin: GPIO13
    type: K
    samples_per_value: 16
  - platform: max31856
    id: temperature_2
    name: "Temperature 2"
    cs_pin: GPIO33
    type: T
    samples_per_value: 2
```

## Checklist:

### DIFF
--- a/esphome/components/max31856/max31856.cpp
+++ b/esphome/components/max31856/max31856.cpp
@@ -8,146 +8,240 @@ namespace max31856 {
 
 static const char *const TAG = "max31856";
 
+enum MAX31856RegisterMasks : uint8_t { SPI_WRITE_M = 0x80 };
+
+enum MAX31856Registers : uint8_t {
+  MAX31856_MASK_REG = 0x02,    ///< Fault Mask register
+  MAX31856_CJHF_REG = 0x03,    ///< Cold junction High temp fault register
+  MAX31856_CJLF_REG = 0x04,    ///< Cold junction Low temp fault register
+  MAX31856_LTHFTH_REG = 0x05,  ///< Linearized Temperature High Fault Threshold Register, MSB
+  MAX31856_LTHFTL_REG = 0x06,  ///< Linearized Temperature High Fault Threshold Register, LSB
+  MAX31856_LTLFTH_REG = 0x07,  ///< Linearized Temperature Low Fault Threshold Register, MSB
+  MAX31856_LTLFTL_REG = 0x08,  ///< Linearized Temperature Low Fault Threshold Register, LSB
+  MAX31856_CJTO_REG = 0x09,    ///< Cold-Junction Temperature Offset Register
+  MAX31856_CJTH_REG = 0x0A,    ///< Cold-Junction Temperature Register, MSB
+  MAX31856_CJTL_REG = 0x0B,    ///< Cold-Junction Temperature Register, LSB
+  MAX31856_LTCBH_REG = 0x0C,   ///< Linearized TC Temperature, Byte 2
+  MAX31856_LTCBM_REG = 0x0D,   ///< Linearized TC Temperature, Byte 1
+  MAX31856_LTCBL_REG = 0x0E,   ///< Linearized TC Temperature, Byte 0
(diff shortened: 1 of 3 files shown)


esphome/components/max31856/max31856.cpp
```diff
@@ -8,148 +8,242 @@ namespace max31856 {
...
+      return 8;
+    case AVE_SAMPLES_16:
+      return 16;
+  }
+  return 16;
+}
+
 void MAX31856Sensor::setup() {
-  ESP_LOGCONFIG(TAG, "Setting up MAX31856Sensor '%s'...", this->name_.c_str());
+  ESP_LOGVV(TAG, "Setting up MAX31856Sensor '%s'...", this->name_.c_str());
```
jesserockz, line 94: Why did you change this? These should be LOGCONFIG

esphome/components/max31856/sensor.py
```diff
@@ -19,22 +28,68 @@
...
+            cv.Exclusive(
+                CONF_DATA_READY_PIN,
+                "mode",
+                exclusive_err_msg,
+                "data_ready_pin will enable autoconversion mode",
+            ): pins.gpio_input_pullup_pin_schema,
+            exclusive_update_interval_with_default: cv.update_interval,
         }
     )
-    .extend(cv.polling_component_schema("60s"))
```
jesserockz, line 37: This should extend cv.COMPONENT_SCHEMA instead then

esphome/components/max31856/sensor.py
```diff
@@ -19,22 +28,68 @@
...
+        accuracy_decimals=3,
         device_class=DEVICE_CLASS_TEMPERATURE,
         state_class=STATE_CLASS_MEASUREMENT,
     )
     .extend(
         {
+            cv.Optional(CONF_HAS_FAULT): binary_sensor.binary_sensor_schema(
+                entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
+                device_class=DEVICE_CLASS_PROBLEM,
+            ),
```
jesserockz, line 76: Sensors should not have binary sensors as children configuration. They should be in their own `binary_sensor.py` file which can link back to the parent.

esphome/components/max31856/sensor.py
```diff
@@ -19,22 +28,68 @@
...
+exclusive_err_msg = "Use only one of data_ready_pin or update_interval"
+
+# Unfortunately, Exclusive doesn't pass a default through to the base Optional so we have to set it manually
+exclusive_update_interval_with_default = cv.Exclusive(
+    CONF_UPDATE_INTERVAL,
+    "mode",
+    exclusive_err_msg,
+    "update_interval will enable polling mode",
+)
+exclusive_update_interval_with_default.default = lambda: "60s"
```
jesserockz, line 61: This is still setting the default update interval to 60s. You can use regular Exclusive and a custom overall validation to add default values for this. 

Then, this component should no longer inherit from PollingComponent, but set its own update `interval` in setup if the time is set but the pin is not.

There are other components that set their own update interval you can reference (I cannot remember them though)

jesserockz (changes requested):
Please add tests for the new configurations

fhriley:
I've already spent too much time on this, and I don't have time to go searching the codebase to figure out how to make a bunch of changes here, unfortunately. You can close this, and I'll just use it from my own repo.

PR #6661, opened 2024-04-30
Fix recent definitions into `defines.h`

edwardtfn:

This fixes missing definitions on `defines.h`:

- `USE_ESP32_BLE`: Should have be added by #6585 
- `USE_NEXTION_TFT_UPLOAD`: Should have be moved by #5667 
- `USE_PSRAM`: Should have be added by #6526 

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** N/A, but looks like this was being reported in some clang.

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:


jesserockz (changes requested):
No, this is not correct. this defines file should basically have every define that is added programatically from .py files.

This is because the CI system runs clang-tidy and clang-format with the direct source code, no code generation, and needs all of these to be defined in this file so it can access that code.

What I was asking for in the other PR is that this line be present in here for both ARDUINO and ESP_IDF because right now the esp-idf code is not being run through clang-tidy

edwardtfn:
Well, if that file shoukd contain all the defines added by py files then there are other missing.
I will take a look at this tomorrow. I have to rest a bit to be able to understand this engine. 🙂

jesserockz:
> Well, if that file shoukd contain all the defines added by py files then there are other missing.

You are not wrong there, but that is not the concern of this specific PR =)

edwardtfn:
Replaced by #6664

Issue #68, opened 2018-06-26
LICENSE


sh0rez:
Esphomeyaml is missing a LICENSE file or any other kind of declaration.
Is this intended? Should all rights be reserved? And does this comply with the Apache-License of homeassistant?

frog32:
Not having a license means I am legally not able to extend this and not having a CLA would prevent me from being able to file a pull request. Some of us have employers that require their employees to be a little picky here so if this is intended to be open please add those files.

OttoWinter:
Yes I understand. esphomeyaml is published under the MIT license as it says in the setup.py file and on pypi.org (https://pypi.org/project/esphomeyaml/#description). I will add a top-level LICENSE file asap.

------

esphomelib, the C++ base for esphomeyaml however is published under GPLv3 as seen here: https://github.com/OttoWinter/esphomelib/blob/master/LICENSE - the reason for this is that I don't really care about anybody copying the python part, as it's honestly quite ugly code rn. The C++ base however was a lot of work, especially all the sensor integrations.

frog32:
Unfortunately that would cause everyone who uses esphomelib to be in violation with the GPLv3 since some underlying libraries are not available in source code and espressif is not able to provide them. Please correct me if I am wrong but as I understand it using esphomelib is not possible if it is under GPLv3.

See https://github.com/espressif/ESP8266_RTOS_SDK/issues/1

frog32:
Maybe that is not a concern but at least esphomeyaml should inform it's users about the fact that their code (configuration) will be linked or combined with GPLv3 code and its implications.
At the same time you might want to provide an alternative option by relicensing it under something more permissive for people that are willing buy such a licence from you.

OttoWinter:
@frog32 Oh noes :( that's not good. I probably should have put more thought into picking the correct license (and not just looking at [this page](https://choosealicense.com/licenses/gpl-3.0/) 😀 )

That being said, I have no problem moving esphome*lib* under MIT too. I mean the code isn't really worth copying anyway 😂. But before I do that, I guess re-licensing would require me to ask any contributors about the change, right? From what I've seen you know way more about this than I do :)

PR #6675, opened 2024-05-03
Add new component mcp3426 7 8

mdop:

Adds support for the 4 channel 16 bit ADC chips of the MCP3426/7/8 family. Structure is based on the ADS1115, but the measurement is done asynchronously.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes -

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3810

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
mcp3428:
  address: 0b1101000

sensor:
  - platform: mcp3428
    name: "Channel 1"
    multiplexer: 1
    resolution: 12
    gain: 8
    id: mcp3428_sensor1

```

## Checklist:


esphome/components/mcp3428/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+CONFIG_SCHEMA = (
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(MCP3428Component),
+            cv.Optional(CONF_CONTINUOUS_MODE, default=False): cv.boolean,
+        }
+    )
+    .extend(cv.COMPONENT_SCHEMA)
+    .extend(i2c.i2c_device_schema(None))
```
jesserockz: Devices with address pins still have a default address if the pins are left floating.
```suggestion
    .extend(i2c.i2c_device_schema(0x58))
```

esphome/components/mcp3428/sensor/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+            cv.Optional(CONF_RESOLUTION, default=16): cv.enum(RESOLUTION, int=True),
+        }
+    )
+    .extend(cv.polling_component_schema("60s"))
+)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await sensor.register_sensor(var, config)
```
jesserockz:
```suggestion
    var = await sensor.new_sensor(config)
```

esphome/components/mcp3428/mcp3428.cpp
```diff
@@ -0,0 +1,168 @@
...
+namespace esphome {
+namespace mcp3428 {
+
+static const char *const TAG = "mcp3426/7/8";
+
+void MCP3428Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up MCP3426/7/8...");
+  uint8_t anwser[3];
+  if (this->read(anwser, 3) != i2c::ErrorCode::NO_ERROR) {
+    this->mark_failed();
```
jesserockz, line 14: Please put useful error messages here so users can look back at logs and see **why** it failed.

esphome/components/mcp3428/mcp3428.cpp
```diff
@@ -0,0 +1,168 @@
...
+    // initial state should be no new measurement in continuous mode
+    config = config | 0b10010000;
+  } else {
+    // don't initiate measurement in single shot mode
+    config = config & 0b01101111;
+  }
+  // leave channel at 1, gain at 1x, and resolution at 12 bit
+
+  if (this->write(&config, 1) != i2c::ErrorCode::NO_ERROR) {
+    this->mark_failed();
```
jesserockz, line 44: Same here

esphome/components/mcp3428/mcp3428.cpp
```diff
@@ -0,0 +1,168 @@
...
+    return;
+  }
+  this->prev_config_ = config;
+  single_measurement_active_ = false;
```
jesserockz: Please make sure to always prefix member variables and functions with `this->`
```suggestion
  this->single_measurement_active_ = false;
```

esphome/components/mcp3428/mcp3428.cpp
```diff
@@ -0,0 +1,168 @@
...
+    voltage = this->convert_anwser_to_voltage_(anwser);
+    single_measurement_active_ = false;
+    this->status_clear_warning();
+    return true;
+  } else {
+    return false;
+  }
+}
+
+float MCP3428Component::convert_anwser_to_voltage_(uint8_t const *anwser) {
```
jesserockz:
```suggestion
float MCP3428Component::convert_answer_to_voltage_(uint8_t const *answer) {
```

esphome/components/mcp3428/mcp3428.cpp
```diff
@@ -0,0 +1,168 @@
...
+  // If continuous mode and config (besides ready bit) are the same there is no need to upload new config, reading the
+  // result should be enough
+  if ((this->prev_config_ & 0b00010000) != 0 and (this->prev_config_ & 0b01111111) == (config & 0b01111111)) {
+    if (millis() - this->last_config_write_ms_ > timeout_wait) {
+      timeout_wait = 0;  // measurement probably immediately available
+    }
+  } else {
+    if (this->write(&config, 1) != i2c::ErrorCode::NO_ERROR) {
+      this->status_set_warning("Error writing configuration to chip.");
+      timeout_wait = 1000;
```
jesserockz: I dont see the point in setting this to an arbitrary high number since it should not be used when false is returned.
```suggestion
```

esphome/components/mcp3428/sensor/mcp3428_sensor.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    }
+                    return RetryResult::RETRY;
+                  });
+}
+
+void MCP3428Sensor::dump_config() {
+  LOG_SENSOR("  ", "MCP3426/7/8 Sensor", this);
+  ESP_LOGCONFIG(TAG, "    Multiplexer: %u", this->multiplexer_);
+  ESP_LOGCONFIG(TAG, "    Gain: %u", this->gain_);
+  ESP_LOGCONFIG(TAG, "    Resolution: %u", this->resolution_);
```
jesserockz: These should be converted back to the human friendly values that were entered in yaml

tests/components/MCP3428/test.esp32-c3-idf.yaml
```diff
@@ -0,0 +1,12 @@
+i2c:
+  - id: i2c_mcp3428
+    scl: 5
+    sda: 4
+
+mcp3428:
+  address: 0b1101000
```
jesserockz: Please use hex here to keep consistent with all other i2c examples.
```suggestion
  address: 0x68
```

jesserockz (changes requested):

esphome/components/mcp3428/mcp3428.cpp
mdop: Added context that communication failed

esphome/components/mcp3428/mcp3428.cpp
mdop: Indeed, not sure what I was thinking.

esphome/components/mcp3428/__init__.py
mdop: Ok, but left floating the address is 0x68, like with grounded pins. Addressed it in a separate commit.

esphome/components/mcp3428/sensor/mcp3428_sensor.cpp
mdop: Seems like I forgot to change this when I changed the numerical values of the enums. Fixed in a commit.

mdop:
Not sure why `CI / Component test MCP3428 (pull_request)` failed, the file `./tests/test_build_components/build_components_base.esp8266.yaml` does exist

lcrockett:
Hello, I was just wondering if there's any love planned for this PR. We have a few use cases for ESP32 + NCD 4-20mA current receiver boards (https://store.ncd.io/product/8-channel-i2c-4-20ma-current-receiver-with-i2c-interface/). It uses a MCP3428 ADC.

mdop:
> Hello, I was just wondering if there's any love planned for this PR. We have a few use cases for ESP32 + NCD 4-20mA current receiver boards (https://store.ncd.io/product/8-channel-i2c-4-20ma-current-receiver-with-i2c-interface/). It uses a MCP3428 ADC.

I'll have a look at the test issues on the weekend. You can already use the component as an external component, though. Just add this to your yaml file:
```
external_components:
  source:
    type: git
    url: https://github.com/mdop/esphome/
    ref: add-MCP3426-7-8
  components: [mcp3428]
  refresh: 300s
```

lcrockett:
@mdop I've hooked up a test setup and am trying out a few configurations. See 'https://pastecode.io/s/nd9joi03' for relevant configuration details.

Based on the receiver board (hooked up with a esp32-wroom-32d) description (https://store.ncd.io/product/mcp3428-8-channel-4-20ma-16-bit-current-receiver-with-iot-interface/) I should be seeing raw ADC values of between 5813-29390.

It appears that this component reports voltage initially ? What i'm unsure about at this point is if I need configuration changes to achieve the readings I want (ADC / mA), or if this is not possible with the hardware i'm using.

See current ESPHome logging at 'https://pastecode.io/s/uovf9tkg'.
- `MCP3428 #68-1` is a 5m submersible depth sensor currently hanging at about 3.5m of depth (very rough estimate)
- `MCP3428 #6e-1` is a pressure sensor 0-10 bar just sitting open
Both sensors are 24v powered and output 4-20mA signals

mdop:
ADCs measure voltage, so that is what the sensor reports. Internal codes are abstracted away to make work easier for the user, but it is a calculated value after all. The reference voltage is 2.048V and is divided into 2¹⁵ ticks in 16bit mode, ie. 62.5µV per tick (internally). With the recommended gain of 2 we get 31.25µV per tick (external voltage). A code of 29390 for 20mA is therefore equivalent to 0.9184V. That would be a 45.9 Ohm sensing resistor, which is a weird value because it doesn't appear in the E96 series. Probably best to check/calibrate the sensing circuit yourself.

mdop:
@lcrockett The component is an ADC which measures a voltage, so that is what it is reporting. It abstracts away all the internal conversions necessary to get the voltage. The product page specifies a code of 29390 at 16 bit resolution and a gain of 2 at 20mA. The code is equivalent to 0.9184375V, or 45.922 V/mA=Ohm. It's a strange sensing resistance value as it doesn't appear in the E96 series, so you may want to check/calibrate the sensing circuit.

lcrockett:
@mdop Thank you for the clarification. I'm going to have to read up more on ADCs working and determine if i'm able to check the sensing circuit.

For now, i've modified your [code](https://github.com/lcrockett/esphome-components-mcp3428) to report only raw ADC values instead of voltage. I can use that to calibrate and report the units I want (bar / CM). The raw ADC readings seem to be very stable relative to the voltage readings (without filters delta).

Your code in this PR does however work as intended with my DAQ circuit, so it's functional and passes usability tests. Thanks again.

PR #6681, opened 2024-05-06
esphome::optional fixes

twasilczyk:

esphome::optional has two problems that prevent its use in code generation
- esphome::nullopt isn't available from cpp_types list (so code generator can't refer to it like to nullptr)
- you can't initialize optional<T> like you would T (so code generator can't just pass e.g. `"some string"` to `esphome::optional<std::string>`) - that's the (8) constructor from [this list](https://en.cppreference.com/w/cpp/utility/optional/optional)

By the way, I have a question: why can't we just use `std::optional`?

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


kbx81:
> By the way, I have a question: why can't we just use std::optional?

There's been some discussion about this on Discord a few times. I don't recall the details, but I don't think there was any specific opposition to using `std::optional`. If you want to look into migrating to it (perhaps that would be a better use of this PR), I don't think we'd be opposed. That said, some of the toolchains may be a bit dated, so if you chase it, be sure to test very, very thoroughly.

twasilczyk:
Ouch. I didn't mean to make a huge change out of this (I did migrations like that at work, they can be nasty). And I already aim at other large change to work on (https://github.com/esphome/feature-requests/issues/1568)...

Let me try to fix this one for now (I see some tests are not passing) and maybe I'll pick up migration in the future.

swoboda1337:
Switch has been made to c++20, closing this

PR #6694, opened 2024-05-07
sen5x: add TPS & PM number concentration

CodeInPolish:
This change adds a new command, listed as being available in [FW >0.7](https://sensirion.com/media/documents/DC018826/637B94B8/Sensirion_SEN5x_Read_Mass_and_Number_Concentrations.pdf)

New entities can be configured:
- pm_n_x_x for PM number concentration in #/cm3 (0.5, 1.0, 2.5, 4 and 10µm are available)
- tps for typical particle size

The implementation is based on sensirion's own Embedded driver, available [here](https://github.com/Sensirion/embedded-i2c-sen5x/blob/448a80ec490858f9fdec559fc95c3df52f0400c0/sen5x_i2c.h#L228).

This command returns the Mass Concentration, the Number Concentration and Typical Particle Size. 
Since we already get the Mass Concentration from the `SEN5X_CMD_READ_MEASUREMENT` command, the new command will not update the Mass Concentration, to prevent doubling the export frequency of those entities.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
  - platform: sen5x
    pm_n_0_5:
      name: "PM <0.5µm Number concentration"
      id: pm_n_0_5
      accuracy_decimals: 1
    pm_n_1_0:
      name: "PM <1.0µm Number concentration"
      id: pm_n_1_0
      accuracy_decimals: 1
    pm_n_2_5:
      name: "PM <2.5µm Number concentration"
      id: pm_n_2_5
      accuracy_decimals: 1
    pm_n_4_0:
      name: "PM <4µm Number concentration"
      id: pm_n_4_0
      accuracy_decimals: 1
    pm_n_10_0:
      name: "PM <10µm Number concentration"
      accuracy_decimals: 1
    pm_tps:
      name: "Typical Particle size"
      id: pm_tps
      accuracy_decimals: 1
```

## Checklist:

image
image


tests/test11.5.yaml
jesserockz, line 1: Please either leave the old test files as they are, or remove this sensor from them

tests/components/sen5x/test.esp32-c3-idf.yaml
jesserockz, line 1: Could you please consolidate the bulk of this yaml into `common.yaml` then include that into the specific platform test files with substitutions for the i2c pins.

esphome/components/sen5x/sen5x.cpp
```diff
@@ -297,38 +325,104 @@ void SEN5XComponent::update() {
...
+
+    if (!this->read_data(measurements, 10)) {
+      this->status_set_warning();
+      ESP_LOGD(TAG, "read data error (%d)", this->last_error_);
+      return;
+    }
+
+    float pm_n_0_5 = measurements[4] / 10.0;
+    if (measurements[4] == 0xFFFF)
+      pm_n_0_5 = NAN;
```
jesserockz: I see the old code already does this, but since changes are being made, it would be better to check for invalid value before doing a division =)

jesserockz: Also without the final `f`, it will use `double` instead of `float`
```suggestion
    float pm_n_0_5 = measurements[4] / 10.0f;
    if (measurements[4] == 0xFFFF)
      pm_n_0_5 = NAN;
```

esphome/components/sen5x/sen5x.cpp
```diff
@@ -297,38 +325,104 @@ void SEN5XComponent::update() {
...
+    this->write_voc_baseline_();
+  }
+
+  this->update_measured_values_();
+
+  if (this->get_pm_number_concentration_and_tps_) {
+    // delay the extra reading to allow update_measured_values to complete.
+    this->set_timeout(30, [this]() { this->update_measured_pm_(); });
+  }
+}
```
jesserockz: There is no guarantee that this will be enough time. Ideally the component would be changes to have a small state machine and run through the stage one at a time in `loop` once triggered by the `update` function.

jesserockz (changes requested):

esphome/components/sen5x/sen5x.cpp
CodeInPolish: That's very true. Wouldn't a loop in the update function block the main event loop though? 

Another approach I was testing was to perform only one reading (either pm with rh/t/nox/vox or full pm) per update. 
If `get_pm_number_concentration_and_tps_` was true, we would need to halve the user specified `update_interval` in the setup with `set_update_interval` and while it was reporting the correct update interval when dumping the config, the sensor was still exporting entities with the original period (i.e. halving it didn't affect the actual period between `update` calls).

I brainstormed a little and came up with another solution:
- use an FSM with 3 states: idle, update_measured_values_progress, update_measured_values_done
- set the idle, update_measured_values_progress and update_measured_values_done accordingly
- create a small helper function to determine the current state of the FSM. This would be responsible for caling itself with a `this->set_timeout(10)` if the state isn't `update_measured_values_done`
- Finally, trigger `update_measured_pm_()` function (from the helper function) when the FSM is in `update_measured_values_done` state and reset the FSM to `idle` in `update_measured_pm_()` 

This would ensure `update_measured_values_()` has had enough time to complete without blocking the main event loop with a loop. What are your thoughts on this?

esphome/components/sen5x/sen5x.cpp
CodeInPolish: Good catch! I'll move all of that validation into its own function to reduce all the repetitions

tests/test11.5.yaml
CodeInPolish: Restored the files as they were.

tests/components/sen5x/test.esp32-c3-idf.yaml
CodeInPolish: Done!

esphome/components/sen5x/sen5x.cpp
jesserockz: No, in the loop state machine you can check the current time against the start time of the last state change and when enough time has passed, you read the data and publish. No delays and timeouts are needed.

CodeInPolish: @jesserockz thank you. I misunderstood what you said earlier (and also didn't know `loop` was still triggered for `PollingComponent`)

I went ahead and implemented the brainstormed solution, because it's less code and less complicated than a full blown FSM for all the tasks to perform. If that's not an acceptable solution, let me know.

PR #6695, opened 2024-05-07
[sensor] First Order High-pass and Low-pass filters

### LINKED ISSUE #2705: Add high pass filter to sensor component
There isn’t a high pass filter available for the esphome sensor component. Propose that one be added. Could be as simple as the example below.

Parameters:
- time_constant
- update_period
- `alpha = time_constant/(time_constant+update_period)`

Difference Equation
`output = alpha * ( previous_output + input - previous_input)`

Attempting to use an accelerometer to detect vibration and would like to be able to filter out DC accelerations (gravity, sensor bias) without having to calibrate or calculate the orientation of the sensor.

**Additional context**
N/A

latonita:

This implements simple but very useful first-order High-pass and Low-pass filters.

High-pass - removing static offset and reduce low-freq noise. `y[i] := α × (y[i−1] + x[i] − x[i−1])`
Low-pass - reducing high-freq noise on signals. `y[i] := y[i-1] + α * (x[i] - y[i-1])`

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2705

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#5584

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

sensor:
  - platform: template
    id: sensor1
    name: Sensor1
    update_interval: 2s
    unit_of_measurement: s
    lambda: |- 
      const float abc[] =  {70, 65.9375, 66.1875, 67, 66, 66.25, 66.25, 66.3125, 66.1875, 66.1875, 66.6875, 66.375, 1, 66.3125, 66.375, 67.125, 78.8125, 66.875, 73.75, 68.625, 70, 73.5, 66.9375, 66.1875, 65.9375, 66, 66.125, 67.4375, 66, 66.375, 66.3125, 66.125, 66.1875, 67, 66.125, 66.375, 89, 66.5625, 66.25, 67.625, 66.125, 66.25, 1, 66.1875, 66.375, 66.375, 1, 66.1875, 67.125, 78.8125, 66.875, 73.75, 68.625, 70, 73.5, 66.9375, 66.1875, 65.9375, 66.375, 66.125, 67.4375, 66.25, 66.375, 66.3125, 66.125, 66.1875, 123, 123, 66.375, 89, 66.5625, 66.25, 67.625, 66.125, 66.25, 66.3125, 66.1875, 20, 66.375, 66.1875, 66.1875};
      static int num = 0;
      float ret = abc[num++];
      if (num >= sizeof(abc)/sizeof(float)) num = 0;
      return ret;
  - platform: template
    id: sensor1_high_pass
    name: Sensor High Pass
    update_interval: 2s
    lambda: return id(sensor1).state;
    unit_of_measurement: s
    filters:
      - high_pass: 0.75
  - platform: template
    id: sensor1_low_pass
    name: Sensor Low Pass
    update_interval: 2s
    lambda: return id(sensor1).state;
    unit_of_measurement: s
    filters:
      - low_pass: 0.25  
    
```

## Checklist:


clydebarrow:
Had you considered generalising this to allow higher-order filters?

latonita:
> Had you considered generalising this to allow higher-order filters?

@clydebarrow When I made these simple 1st order HPF/LPF I actually thought it might be generalized and made configurable.
For example, there is a good library which I used many times (original is https://github.com/dimtass/DSP-Cpp-filters, a little bit more interesting fork is here https://github.com/crylent/DSP-Cpp-filters).
So potentially we can have a generic DSP Filter component, which can be configured - whether its a high pass or low pass , 1st/2nd order, and provide configuration params like cut-off frequencies, width and so on.
However, it might be hard for regular users to understand how to work with it and select proper parameters, especialy having in mind the fact that in esphome we work with quite low-frequency signals and might be tough to operate with 'frequency' concepts.

I might actually to come up with the updated generic dsp filter component so we can review. Its not the rocket science, question is usability. maybe several separate specific components might be easier to use... maybe not.

like 1st_hpf, 1st_lpf, 2nd_hpf, 1st_butterworth, etc.

Ideas, thoughts?

nagyrobi:
> 1st_hpf, 1st_lpf, 2nd_hpf, 1st_butterworth, etc.

Maybe that would be indeed more easy to support (meaning providing support to end users), rather than a universal component with many tweakable settings (knowing the target user base, and seeing the kind of questions arise on Discord).

swoboda1337 (changes requested):
Needs tests and docs.

latonita:
@swoboda1337 
tests added.
docs in new format added (https://github.com/esphome/esphome-docs/pull/5584/). 

it was easier to create new branch, than update from the old 2024 format. can you please delete old docs PR? https://github.com/esphome/esphome-docs/pull/3875 
thanks

esphome/components/sensor/filter.cpp
```diff
@@ -447,6 +447,40 @@ optional<float> RoundFilter::new_value(float value) {
...
+}
+
+LowPassFilter::LowPassFilter(float alpha) : alpha_(alpha) {}
+void LowPassFilter::set_alpha(float alpha) { this->alpha_ = alpha; }
+optional<float> LowPassFilter::new_value(float value) {
+  if (std::isnan(value)) {
+    return {};
+  }
+
+  if (std::isnan(this->last_input_)) {
```
swoboda1337: It seems `last_input_` is never updated so would always be NAN? Was this tested?

esphome/components/sensor/filter.h
```diff
@@ -497,6 +497,30 @@ class RoundFilter : public Filter {
...
+  float alpha_;
+  float last_input_{NAN};
+  float last_output_{NAN};
+};
+
+class LowPassFilter : public Filter {
+ public:
+  explicit LowPassFilter(float alpha);
+  void set_alpha(float alpha);
+  optional<float> new_value(float value) override;
```
swoboda1337, line 644: Change order of functions to match HighPassFilter

swoboda1337 (changes requested):

esphome/components/sensor/filter.cpp
```diff
@@ -447,6 +447,40 @@ optional<float> RoundFilter::new_value(float value) {
...
+HighPassFilter::HighPassFilter(float alpha) : alpha_(alpha) {}
+void HighPassFilter::set_alpha(float alpha) { this->alpha_ = alpha; }
+optional<float> HighPassFilter::new_value(float value) {
+  if (std::isnan(value)) {
+    return {};
+  }
+
+  if (std::isnan(this->last_input_)) {
+    this->last_output_ = value;
+    this->last_input_ = value;
```
swoboda1337: Maybe should return {} here? If the signal is just DC it would avoid outputting a misleading value that doesn't represent the filtered signal.

latonita: agree. also i change the first value to 0.


  - If we set last_output_ = value (input), then on the next sample:
    y = alpha * (value + x2 - value) = alpha * x2
    which leaks absolute signal level into HPF output (wrong bias).
  - If we set last_output_ = 0, next sample is:
    y = alpha * (x2 - x1)
    which matches “no prior high-pass energy yet” and avoids DC injection.

  So 0.0f is a neutral initialization for y[n-1]; value is not.

esphome/components/sensor/filter.cpp
latonita: its a typo, thanks!

esphome/components/sensor/filter.h
latonita: done

PR #6704, opened 2024-05-08
Add PHY register writes to enable external clock on Ethernet with RTL8201

heythisisnate:

When using Ethernet on a board with the `RTL8201` PHY and using an external clock (`clk_mode: GPIO_IN`), there's a good chance that the RMII Mode Setting Register is incorrect. This change fixes the problem by ensuring that the RMII clock mode is set correctly in the PHY register when initializing ethernet.

## Background

Two common IoT boards that use the RTL8201 PHY are the Konnected Alarm Panel Pro and wESP32. The Konnected panel based its design on the wESP32 as a reference, so the ethernet schematics between these two boards are the same. In recent production runs of these boards, suddenly Ethernet performance was severely degraded to the point of non-functional. After months of investigation, we tracked down the problem to a floating pin (pin 12) on the RTL8201FI that _should_ have been pulled up to correctly set the clock mode.

Apparently, there was a recent change in the default behavior of the RTL8201FI chip in date codes starting with _N_. We observed, in collaboration with @xorbit of Silicognition (maker of wESP32), that RTL8201FI chips with date codes _M_ and prior worked despite the floating pin, while those with _N_ and after did not.

Fortunately, we can compensate for the hardware change in firmware and set the PHY register for RMII Mode Settings in the ethernet initialization code, according to the datasheet for the RTL8201FI:
Screenshot 2024-05-07 at 6 13 44 PM

The incorrect setting here that we need to change is the _Rg_rmii_clkdir_ bit to 1 to indicate INPUT direction. 

Changing this is confirmed to fix Konnected Pro boards with RTL8201 chips with _N_ and later date codes.

Interestingly, I observed the hardware set defaults between newer and older boards were the same in both _M_ and _N_ variant chips, indicating that this mode setting is incorrect even on the functional Konnected and wESP32 boards out in the wild. So it's plausible that correctly setting it would improve the ethernet link quality for these boards. It may be that chips with the M and older date code are maybe just more tolerant in some way than the newer ones??

```
[16:39:42][V][ethernet:615]: Hardware default RTL8201 RMII Mode Register is: 0x0FFA
[16:39:42][V][ethernet:622]: Setting RTL8201 RMII Mode Register to: 0x1FFA
```

thank you to @xorbit and @h2zero for the immense help getting to the bottom of this

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
ethernet:
  type: RTL8201
  mdc_pin: GPIO16
  mdio_pin: GPIO17
  clk_mode: GPIO0_IN
  phy_addr: 0

```

## Checklist:

### DIFF
--- a/esphome/components/ethernet/ethernet_component.cpp
+++ b/esphome/components/ethernet/ethernet_component.cpp
@@ -184,6 +184,10 @@ void EthernetComponent::setup() {
     // KSZ8081RNA default is incorrect. It expects a 25MHz clock instead of the 50MHz we provide.
     this->ksz8081_set_clock_reference_(mac);
   }
+  if (this->type_ == ETHERNET_TYPE_RTL8201 && this->clk_mode_ == EMAC_CLK_EXT_IN) {
+    // Change in default behavior of RTL8201FI may require register setting to enable external clock
+    this->rtl8201_set_rmii_mode_(mac);
+  }
 #endif
 
   // use ESP internal eth mac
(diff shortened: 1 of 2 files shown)


esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -584,6 +588,50 @@ void EthernetComponent::ksz8081_set_clock_reference_(esp_eth_mac_t *mac) {
 
 #undef KSZ80XX_PC2R_REG_ADDR
 }
+
+void EthernetComponent::rtl8201_set_rmii_mode_(esp_eth_mac_t *mac) {
+#define RTL8201_RMSR_REG_ADDR (0x10)
```
jesserockz: I just saw the same above and I will change that in its own pr to be a `constexpr uint8_t`
```suggestion
  constexpr uint8_t RTL8201_RMSR_REG_ADDR = 0x10;
```

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -584,6 +588,50 @@ void EthernetComponent::ksz8081_set_clock_reference_(esp_eth_mac_t *mac) {
...
+  ESPHL_ERROR_CHECK(err, "Setting Register 16 RMII Mode Setting failed");
+
+  err = mac->read_phy_reg(mac, this->phy_addr_, RTL8201_RMSR_REG_ADDR, &(phy_rmii_mode));
+  ESPHL_ERROR_CHECK(err, "Read PHY RMSR Register failed");
+  ESP_LOGV(TAG, "Setting RTL8201 RMII Mode Register to: 0x%04X", phy_rmii_mode);
+
+  err = mac->write_phy_reg(mac, this->phy_addr_, 0x1f, 0x0);
+  ESPHL_ERROR_CHECK(err, "Setting Page 0 failed");
+
+#undef RTL8201_RMSR_REG_ADDR
```
jesserockz:
```suggestion
```

jesserockz (changes requested):

esphome/components/ethernet/ethernet_component.cpp
jesserockz: https://github.com/esphome/esphome/pull/6705

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -584,6 +588,48 @@ void EthernetComponent::ksz8081_set_clock_reference_(esp_eth_mac_t *mac) {
 
 #undef KSZ80XX_PC2R_REG_ADDR
 }
+
+void EthernetComponent::rtl8201_set_rmii_mode_(esp_eth_mac_t *mac) {
+  constexpr uint8_t RTL8201_RMSR_REG_ADDR = 0x10;
```
jesserockz:
```suggestion
constexpr uint8_t RTL8201_RMSR_REG_ADDR = 0x10;
void EthernetComponent::rtl8201_set_rmii_mode_(esp_eth_mac_t *mac) {
```

jesserockz (approved):

PR #6717, opened 2024-05-09
telnet + wrench

apbodrov:

Implementation of telnet cli 

## Types of changes

- [+] New feature (non-breaking change which adds functionality)

**Related issue or feature (if applicable):** part of https://github.com/esphome/feature-requests/issues/2707

## Test Environment

- [+] ESP32

```yaml
# Example config.yaml
telnet:
```

## Checklist:

### DIFF
--- a/esphome/components/telnet/__init__.py
+++ b/esphome/components/telnet/__init__.py
@@ -1,0 +1,19 @@
+import esphome.codegen as cg
+import esphome.config_validation as cv
+from esphome.const import CONF_ID
+
+CODEOWNERS = ["@apbodrov"]
+
+telnet_ns = cg.esphome_ns.namespace("telnet")
+TelnetComponent = telnet_ns.class_("TelnetComponent", cg.Component)
+
+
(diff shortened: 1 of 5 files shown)


jesserockz:
This is very niche and I think it should just live as an [external_component](https://esphome.io/components/external_components)

apbodrov:
> This is very niche

This is the same SSH (It’s hard to call it niche), only for the poor. I want to implement through this interface such things as configuration in runtime, and such commands as, for example, `dump_config` to pull out the configuration from the flashed ESP (for me this is really a problem and I saw PR on this topic too)
This will require changes to the core part and can be useful to people (network equipment basically uses CLI insted of something else)

jesserockz:
I have no doubts that it could be useful to **you** and I know what telnet and SSH are. I am saying this in the context of having a telnet server available on a ESPHome device.

We are happy to accommodate smaller core changes that would he helpful to an external component as long as they do not interfere with normal usage.

PR #6724, opened 2024-05-11
fix: Growatt solar, prevent invalid readouts during boot cycle

### LINKED ISSUE #3965: Spikes with integration growatt_solar
### The problem

To avoid doubles see here:
https://community.home-assistant.io/t/energy-dashboard-not-populating/433280

 2022.12.3

Home Assistant Add-on

 2022.12.3 and a lot of them before

ESP8266

### Board

D1 Mini

growatt_solar

### Example YAML snippet

_No response_

_No response_

### Additional information

I have had long discussions on the HA forum. i was pointed here by "tom_i".

Don't know where else to go...

Please help. If not the right place tell me how and what to do else please.

rfpronk:

Early morning when first light hits the solar panels the inverter boots up and powers the USB port on which the ESPhome device is powered.
The first readouts can be invalid, giving either invalid values or values from the previous day, which is annoying when counting energy production per day. (this ruins home assistant energy dashboard graphs).
This PR gives the inverter a bit time (30s) to complete it's boot cycle before ESPHome will pay attention to the data values it will be sending.

I wrote this fix in the evening, but I can't this in the evening. Compilation works and tomorrow when the sun shines again I will flash this branch and let it run for a couple of days to see if the spikes are gone.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes  esphome/issues/issues/3965

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs/pull/3832

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Example will do the trick, https://esphome.io/components/sensor/growatt_solar.html

## Checklist:


jesserockz (changes requested):
This should be a user configurable config of duration that defaults to `0s` (to no break existing configs).

This user provided time should then just be checked inside `update` and just do a `return` if its less than the time, preventing it from ever sending any data requests.

rfpronk:
Thanks for your review.

> This should be a user configurable config of duration that defaults to 0s (to no break existing configs).

~This only happens after startup (when the inverter comes online) and when flashing a new esphome (and thus restart). It will only take a bit longer for the first reading to be available. imho it doesn't break existing config and a user configurable value might work confusing. Would you still prefer a user configurable setting?  I'm just curious to your thoughts here.~ Never mind, I implemented it 👍 

> This user provided time should then just be checked inside update and just do a return if its less than the time, preventing it from ever sending any data requests.

Good point, I will change this soon

This morning it worked fine on my unit, the invalid spike in readout was gone so the idea does the trick on the initial issue 👍

esphome/components/growatt_solar/sensor.py
```diff
@@ -108,6 +109,7 @@
             cv.Optional(CONF_PROTOCOL_VERSION, default="RTU"): cv.enum(
                 PROTOCOL_VERSIONS, upper=True
             ),
+            cv.Optional(CONF_WARM_UP_TIME, default=0): cv.uint8_t,
```
jesserockz, line 112:
```suggestion
            cv.Optional(CONF_WARM_UP_TIME, default="0s"): cv.All(cv.positive_time_period_seconds, cv.uint8_t),
```

esphome/components/growatt_solar/sensor.py
```diff
@@ -163,6 +165,9 @@ async def to_code(config):
 
     cg.add(var.set_protocol_version(config[CONF_PROTOCOL_VERSION]))
 
+    if CONF_WARM_UP_TIME in config:
+        cg.add(var.set_warm_up_time(int(config[CONF_WARM_UP_TIME])))
```
jesserockz: There is a default of 0 so it is always there
```suggestion
    cg.add(var.set_warm_up_time(int(config[CONF_WARM_UP_TIME])))
```

jesserockz (changes requested):
Looks better =)
Just a few minor things below

esphome/components/growatt_solar/sensor.py
rfpronk: ah okay I misunderstood it then, I thought `config` would only be the config text blob without the defaults.

esphome/components/growatt_solar/sensor.py
rfpronk: Sorry for asking, but setting a string as default for an int would cause an issue right? Or does the `Optional` function convert/cast that? The `int()` cast  on line 169 will error on `"0s"` but can handle `"0"`

ssieb:
Are you aware of this filter?  https://esphome.io/components/sensor/#skip-initial

rfpronk:
I was not, thanks for that!
Much simpler solution indeed. 
Will flash this change tomorrow to my unit when the sun shines again and powers the unit

```diff
     energy_production_day:
       name: "Growatt Today's Generation"
       icon: "mdi:solar-power-variant"
+      filters:
+      - skip_initial: 3
     total_energy_production:
       name: "Growatt Total Energy Production"
       icon: "mdi:solar-power-variant"
+      filters:
+      - skip_initial: 3
```

PR #6752, opened 2024-05-15
Allow using allowlist for BLE scanner

feens:
Enables using the esp-idf BLE scanner's whitelist functionality.  This is useful in cases where there's a lot of BLE devices nearby and they overwhelm the scanner, especially in `active` mode.

My own example is using this along with the Bluetooth Proxy component to enable the Home Assistant Yale Bluetooth integration.  I have 3 locks, 2 close to one proxy, one close to another, but the extra BLE traffic can cause significant delays.  I have no other BLE devices that I'm integrating at this point, so filtering down to those devices is very beneficial.

## Types of changes

- [ ] Other

I think this would basically complete https://github.com/esphome/feature-requests/issues/2686 as well (the ask for was for the proxy, but it relies on this component for scanning anyway)

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3848

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
substitutions:
  name: olimex-bluetooth-proxy-00886c
  friendly_name: Bluetooth Proxy (POE)
packages:
  esphome.bluetooth-proxy: github://esphome/firmware/bluetooth-proxy/olimex-esp32-poe-iso.yaml@main
esphome:
  name: ${name}
  name_add_mac_suffix: false
api:

esp32:
  board: esp32-poe-iso
  framework:
    type: esp-idf
    version: 4.4.7

ethernet:
  type: LAN8720
  mdc_pin: GPIO23
  mdio_pin: GPIO18
  clk_mode: GPIO17_OUT
  phy_addr: 0
  power_pin: GPIO12

logger:

ota:

dashboard_import:
  package_import_url: github://esphome/firmware/bluetooth-proxy/olimex-esp32-poe-iso.yaml@main

external_components:
  - source: github://feens/esphome_official
    refresh: 1minutes
    components: [ esp32_ble_tracker ]

esp32_ble_tracker:
  scan_parameters:
    interval: 1100ms
    window: 1100ms
    active: true
    allowlist_address: 

      - 78:9C:85:09:72:10
      - 78:9C:85:10:89:42

bluetooth_proxy:
  active: true
```

## Checklist:


esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
```diff
@@ -213,6 +214,15 @@ class ESP32BLETracker : public Component,
   void gap_event_handler(esp_gap_ble_cb_event_t event, esp_ble_gap_cb_param_t *param) override;
   void ble_before_disabled_event_handler() override;
 
+  static void uint64_to_bd_addr(uint64_t address, esp_bd_addr_t bd_addr) {
```
bdraco, line 244: We should avoid duplicating the code since this function already exists in `bluetooth_proxy.h`

feens: ya...I was trying tot think of the best approach for that.  Can I easily pull in the proxy one, or should the proxy one switch to referencing this one?  The latter seems better as the proxy requires this component anyway.

esphome/components/esp32_ble_tracker/__init__.py
```diff
@@ -30,6 +30,9 @@
 CONF_WINDOW = "window"
 CONF_CONTINUOUS = "continuous"
 CONF_ON_SCAN_END = "on_scan_end"
+CONF_ALLOWLIST_ADDRESS = "allowlist_address"
+MAX_ALLOWLIST = 10
```
feens: The max depends on which board is being used, so I picked what should be a more conservative amount.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.cpp
```diff
@@ -274,15 +274,35 @@ void ESP32BLETracker::start_scan_(bool first) {
...
+    this->scan_params_.scan_filter_policy = BLE_SCAN_FILTER_ALLOW_ALL;
+  }
+
   esp_err_t err = esp_ble_gap_set_scan_params(&this->scan_params_);
   if (err != ESP_OK) {
     ESP_LOGE(TAG, "esp_ble_gap_set_scan_params failed: %d", err);
     return;
   }
+
+  if (!this->allowlist_address_vec_.empty() && !this->allowlist_populated_) {
```
feens, line 333: I wanted to run this in `setup()` as it really only needs to run the first time...but then it wasn't working at all, so I'm just using a flag to save it from running on successive passes.

feens:
@bdraco docs changes are up, should be ready for a proper review now.

esphome/components/esp32_ble_tracker/esp32_ble_tracker.h
bdraco: since proxy depends on esp32_ble_tracker it probably makes sense to move it here.

I'd do that in another PR and than rebase this one on top once its merged

esphome/components/esp32_ble_tracker/__init__.py
nagyrobi: The max allowlist is hardcoded? How does the user know about this? Any way to change this value?

Eg. Using 12 Xiaomi temp/hum sensors and a BLE lock...

feens: @nagyrobi thus far this is the only spot where I've read anything conclusive: https://github.com/espressif/esp-idf/issues/9205

nagyrobi: Should use the max possible for every chip type, or at least make it configurable.

feens: I don't disagree, but I don't have the time to figure out what's possible for every chip (or even know where to look).  Could be a good follow-up PR. I feel like this at least adds a beneficial feature, and the risk of making it configurable is that if an incorrect value is used by the end-user it won't work properly, so this feels like safer approach to me.

nagyrobi: The problem is that otherwise it will generate support questions every time someone runs into that. And, I don't think that 10+ devices is a rare number, especially when used with sensors...

feens: I don't know if I get the logic there.  I would've guessed having the ability to whitelist to any extent, even if the max is a few devices less than the chip allows, is better than not having the option at all.  I poked around bit in the ESP-IDF code, but can't figure out where they define their values. I agree that 10 isn't that many...but given that it seems that the C3 supports 12, and the S3 supports at most 15, I don't know if adding a couple more is going to fix that. 

 I don't really have much more time to offer on this, so I think I can give two options:

1. I can make the max configurable.  I'd still expect this to cause more support issues, as if you don't know what the max is and try and add too many items, it just won't work.
2. I can just close my PR and run this as my own external component. I figured I'd put in a bit extra time to try and give others the option, but if it's not deemed sufficient for the main repo, then perhaps that's the better option.

nagyrobi: Why does this feature have to depend on IDF?

feens: @nagyrobi the methods used are for the ESP-IDF API.  Arduino _may_ have similar functionality but I can't find much info on it.

bdraco: I expect they are also available on Arduino. Switching the framework might be all that is needed to verify

jesserockz: The esp-idf apis are all available even when using Arduino (version dependent for some things)

feens: I'll give Arduino a try when I get a moment and adjust as necessary

feens: sorry for the delay...was pre-occupied with springtime outdoor work for a bit there.  I was finally able to test it w/arduino as a platform, didn't see any issues pop up and I've removed that restriction.

brozikcz:
thanks a lot for PR, it works as expected with `esp-idf` framework

joe-sydney:
Having an ability to whitelist based on MAC would be great, in fact, I had opened a feature request in 2022 for this. However, since then I also need to track IRKs (an iPhone). Would it be possible to extend MAC whitelisting to whitelisting on IRKs as well? OR, it would also be fine to whitelist on MAC for bt proxy, but allow IRKs to still work for ble_presence within esphome.

jcam:
Is this waiting on something to be able to bring it in? 

I have multiple BLE trackers and multiple BLE locks that are far enough from each other that if they bind onto the wrong tracker then controls fail until they are restarted.
I was able to just pull this component in a previous version but it now does not work without conflicting code changes :/

luka-loncaric-kyndryl:
Bump

esphome/components/esp32_ble_tracker/__init__.py
bdraco: https://github.com/espressif/esp-idf/issues/9205#issuecomment-1963441801

12 is probably fine as the min value, the C6 (`VARIANT_ESP32C6`) supports 15

jcam:
Updated here: https://github.com/esphome/esphome/compare/dev...jcam:esphome:2025.5.1-btfilter

bdraco:
Hey @feens, I cleaned up this PR. Is this still something you need? We've made a lot of improvements to BLE performance since then. It's night and day better now, so this might not be necessary anymore.

jcam:
The issues that I had with BLE were not performance, but that it didn't seem to be very smart about choosing the nearest tracker. I have 2 esphome BLE things, one near the front door and one near the back door, and for whatever reason, the door locks would often connect to the *wrong* endpoint. This would cause it to then be nonresponsive to commands due to the weak connection.
I could try running without the allowlist for a while and see if it still happens.

bdraco:
If you're referring to Bluetooth Proxies, the selection algorithm used to be pretty naïve, but it's been improved a lot and is now much more fault-tolerant. I'd definitely give it another try. The downside of using an allowlist is that you lose failover if a device is reachable by multiple proxies, which can reduce reliability when a proxy goes offline or experiences transient interference.

g-kiss:
I have been using this PR for a long time. It is very good at relieving the wifi network of a lot of BLE advertisements, because it only sends traffic from authorized BLE devices to Home Assistant. The 60,000 packets / hour dropped to 600 packets / hour. It would be useful if this feature was included.

RamonBalthazar:
Can someone please provide an update about this?
Would love to use this feature 🙏

robdraaijer:
> Having an ability to whitelist based on MAC would be great, in fact, I had opened a feature request in 2022 for this. However, since then I also need to track IRKs (an iPhone). Would it be possible to extend MAC whitelisting to whitelisting on IRKs as well? OR, it would also be fine to whitelist on MAC for bt proxy, but allow IRKs to still work for ble_presence within esphome.

I would also be highly interested in extending this PR to allow for IRKs as most phones/smartwatches will use them.

david-collett:
I'd also like to see this. I have an annoying device (a yale lock) which causes grief if its constantly active scanned, this would allow me to support devices that require active scan while avoiding that problem device. Seems like a pretty straightforward patch too. :)

bdraco:
Thank you for this contribution and the effort you put into implementing the BLE allowlist functionality. This feature has generated interest from the community and could provide value for users dealing with high BLE traffic environments.

Unfortunately, we haven't received responses to the recent review comments and requests for updates. We understand that contributors' availability can change, and we appreciate the work that has been done on this PR.

To keep our PR queue manageable and focused on active development, we're going to close this PR for now. While this feature could be useful in certain scenarios, it hasn't been a priority for the core team to complete ourselves.

That said, if someone in the community wants to pick this up and complete the implementation:
- Feel free to create a new PR building on the work done here
- Using a feature-specific branch name (rather than `dev`) will make it easier for maintainers to help keep the PR up to date
- We'd be open to reviewing and merging a completed implementation

Thank you again for your contribution to the ESPHome project. We hope either you'll have time to return to it or another contributor who needs this functionality will pick up where this left off.

PR #6759, opened 2024-05-16
Uncouple safe_mode from OTA

kbx81:

Uncouple safe_mode from OTA

## Types of changes

- [x] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3841

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
safe_mode:
```

## Checklist:


esphome/components/safe_mode/safe_mode.cpp
```diff
@@ -0,0 +1,124 @@
...
+    } else {
+      ESP_LOGW(TAG, "SAFE MODE IS ACTIVE");
+    }
+  }
+}
+
+float SafeModeComponent::get_setup_priority() const { return setup_priority::AFTER_WIFI; }
+
+void SafeModeComponent::loop() {
+  if (this->read_rtc_() && (millis() - this->safe_mode_start_time_) > this->safe_mode_enable_time_) {
```
jesserockz: We probably dont want to be loading from preferences every single loop...

kbx81: Cool, I'll just create a flag instead. 🙂

kbx81: Fixed in [`446506c`](https://github.com/esphome/esphome/pull/6759/commits/446506cf34b5b06af32471b6d8546434c50bc444)

esphome/components/ota/__init__.py
jesserockz, line 1: I think for now at least, `safe_mode` should be AUTO_LOAD` by `ota`
If we want to disable it, it can follow the same path as mdns with `disabled: true` in config

kbx81: Ah, good call. I'll do that 😄

kbx81: Done in [`01bbfa4`](https://github.com/esphome/esphome/pull/6759/commits/01bbfa48353eefd5a72dc7a5568dc2f98b8825f1)

jesserockz (approved):

PR #6794, opened 2024-05-22
Add mitsubishi_itp Component

samphonic:

This PR adds the `mitsubishi_itp` component that allows communication with Mitsubishi heat pumps via the CN105 connector.  This integration works similarly to the seminal [SwiCago/HeatPump](https://github.com/SwiCago/HeatPump) library, except it has been written from the ground-up to be an ESPHome component.  It also supports a second UART connection to an MHK2 thermostat for ease of control and integration with existing family workflows 😉.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3860

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example minimal configuration entry

climate:
  - platform: mitsubishi_itp
    uart_heatpump: hp_uart

uart:
  - id: hp_uart
    baud_rate: 2400 # Some devices may require 9600
    parity: EVEN
    rx_pin:
        number: GPIO7
    tx_pin:
        number: GPIO6

```

Additional configuration examples (including adding an MHK2 thermostat and using a Home Assistant sensor as a temperature source) are available here: https://muart-group.github.io/user/configuration

## Checklist:


samphonic:
Some formatting issues missed and I will add myself as codeowner (apologies, it's my first PR here, I'll update those shortly).

samphonic:
Phew, apologies again for a rough first PR there (clearly was not running enough test scripts locally), but the checks are passing now.  Let me know if there's any additional information or changes needed.

paravoid:
Just a drive-by comment but note for posterity that there are multiple out-of-tree efforts to implement this functionality: geoffdavis/esphome-mitsubishiheatpump#131 is a good resource documenting many of the forks. I think it'd be awesome to coalesce efforts around one implementation, and having something in esphome core would help tremendously in that regard. Thanks @Sammy1Am :)

nagyrobi:
Does this support the Mitsubishi SLZ-M35FA / SUZ-M35VA model?

I'm approaching the other way around: I'd like to buy an AC which is (or can be) supported by ESPHome as deep as possible (not just IR blaster hacky) and this component seems to meet this!

Edit: [likely yes](https://github.com/SwiCago/HeatPump/issues/13#issuecomment-658130582)!

samphonic:
> Does this support the Mitsubishi SLZ-M35FA / SUZ-M35VA model?

Just so we're not making any assumptions, the [supported units list](https://muart-group.github.io/user/supported) has started fresh only with units we've confirmed to work; **however** generally speaking anything in the "CN105" rows of [this document](https://q4es.com/wp-content/uploads/2022/03/M-Series-Compatibility-Chart.pdf), or listed as "OK" in [this document](https://www.mitsubishicomfort.com/themes/custom/MitsubishiMegaSite/src/pdf/M_Submittal_kumo_cloud.pdf) should be supported as this library is using the same protocol as the MHK2 / Kumo controllers.

Different equipment might not support all features, but that's more down to the equipment and would be the case even with official controllers.

nagyrobi:
Only SLZ KA and KF are present on the lists you mentioned, not the FA model. I ordered a Mitsubishi SLZ-M35FA / SUZ-M35VA with a wired remote, let's see how it works, as it has a CN105.

Just a small observation, to be conform with the ESPHome naming habits:

```yaml
mitsubishi_cn105: #or maybe more generic mitsubishi_hvac, or perhaps mitsubishi_kumo
  uart_heatpump: hp_uart

uart:
  - id: hp_uart
    baud_rate: 2400 # Some devices may require 9600
    parity: EVEN
    rx_pin:
        number: GPIO7
    tx_pin:
        number: GPIO6
```
Using `_uart` too often as variable names creates confusion among users.
Since this component is specifically designed to support Mitsubishi HVACs through the CN105 connector's protocol, it would be very helpful to clarify that through its name. (I am aware that some older models use CN100 for the same purpose, but it's rare). Even in one of the docs you linked above the official accessories are generically called _CN105 Remote Controllers, Interfaces, and Accessories_ by the factory (they even have a ModBUS adapter!). Let's just call it `mitsubishi_cn105` in ESPHome too.

Instead of `heatpump_uart` and `thermostat_uart` I'd suggest using `uart_heatpump` and `uart_thermostat` to denote that the component is not implementing its own uart, rather just references an ID to an existing ESPHome uart. (gramatically it's "The _uart_ connected to the _heatpump_" in a logic sentence)

samphonic:
Definitely let us know how the *FA works so we can expand the list.

Good suggestions on naming, thanks for those!

- I was hesitant to use `_cn105` at first since that's the "connector" rather than the protocol (which I believe is the kinda-poorly-named "IT Protocol"), but now that you've pointed out the documents are referring to them as "CN105 Controllers" I think I'm sold.
- I haven't found another ESPHome component that has field named other than `uart_id`, but even just being consistent by starting with `uart_*` seems reason enough to change that.

I'll have an update here in the next day or two.

nagyrobi:
Currently testing with [mitsubishi_heatpump_emulator](https://github.com/akamali/mitsubishi_heatpump_emulator) and your component looks great!

I think that the Active Mode switch should be configurable from YAML with in order to be able to configure its `internal` or `disabled_by_default` options because it only has relevance if the user has the separately purchased MHK1/MHK2 thermostat. If users don't have it, this switch shouldn't appear in the UI by default. Would be nice if `restore_mode` would be accessible too, eg. when the node boots somebody would like to have it as `ALWAYS_ON` while somebody else would prefer it restored from a previous state if they do have the accessory etc.

I'd probably add an extra binary sensor to show when the connection between ESPHome and the heatpump is OK. Eg when `[W][muart_bridge:050]: Timeout waiting for response to 5a packet.` happens just set the binary sensor to false, otherwise set it to true. Not extra important though. Just to have it for HA automations to check before sending commands to the node or inform the user that the heatpump is not controllable this way, go and see what's the issue. Currently if I cancel the emulator on the PC ESPHome reboots.

One thing I miss is that the SLZ-Mxx is a cassette model, which fits into ceiling. It has 4 vanes to blow air in different directions, and from the factory remote (according to the manual) the vane vertical positions can be set separately for each (this is a very important feature, I chose this model specifically for this). Also cassette models don't seem to have horrizontal vane positions.
Of course no clue if the emulator supports that (guess not). I think we should have a `cassette_vane_select` with values `All`,`1`,`2`,`3`,`4`, and the user would simply select which vane to adjust from `vane_position_select`.
Have to check if the library supports this at all.

KazWolfe:
I got you on the CN105 Matrix chat, but I'll send a response here for clarity as well.

> I think that the Active Mode switch [...]

I think this makes sense. We can force active mode to always be on if there's no secondary UART defined, and make the toggle default to internal since it's really only intended for development/passive packet captures. Depending on a few other things, we may need to honestly remove the active mode switch entirely (or make it a core flag), but that'll be a later issue.

> I'd probably add an extra binary sensor to show when the connection between ESPHome and the heatpump is OK. [...]

Also seems reasonable, though our climate entity should report the heat pump as `Unavailable` in this state (I think?).

> One thing I miss is that the SLZ-Mxx is a cassette model, which fits into ceiling. It has 4 vanes to blow air in different directions [...]

At present, we are not aware of a way to do this at all via the CN105 port, nor any IT protocol specification for sending these commands. That's not to say they *don't* exist, just that nobody's seen it yet. We'd need to get some kind of packet capture from a CN105 device capable of setting these values, get lucky in finding another packet that reports it, or similar.

I'm interested in this as well though, as my wall-mount unit (MSZ-FS##NA) has similar split vertical vanes. I have not yet been able to find a way to control it though, so more research may be necessary.

> Also cassette models don't seem to have horrizontal vane positions.

I have nagged Sammy about adding autoconf support to the project for a while, and the general consensus seems to be "eventually." For now, we're trying to get this out there to gather more information _overall_ so we can figure out what we can (or can't) collect or control.

We know [a fair bit](https://muart-group.github.io/developer/packet-reference/0x7B-extended-connect-response) about the capability packet, but there are still a few things we need to deal with - including how to (properly) block reporting to HA until we get our capability packet, though the Midea integration does this as is. This initiative can be tracked in https://github.com/muart-group/esphome/issues/13 for those interested.

paravoid:
OK this may sound silly, but how do I report bugs? I can respond to issues on `muart-group/esphome` (and I've already done that on one) but it seems I cannot create new ones, as I get the esphome GitHub issue template, which directs me to `esphome/issues` etc. I see this mention of a CN105 Matrix server above but no idea where that is?

samphonic:
Oh, that's definitely an issue (no pun intended)... I'm going to try *really, really* hard to actually get time to work on this this evening and I'll make sure issues work on muart-group/esphome.

KazWolfe:
@paravoid This is a side effect of us using a fork and both Sammy and I being new to the entire "esphome contribution" thing. We'll fix that eventually, I promise.

You can:

* Open an untemplated issue using [this magic link](https://github.com/muart-group/esphome/issues/new) which bypasses the selector.
* Join us in the [mUART Discord](https://discordapp.com/invite/VRpgcF5JG5), which covers development specifically for this library and our protocol reverse engineering efforts
* Join us in [SwiCago's Gitter](https://app.gitter.im/#/room/#Mitsubishi-Heat-Pump_Lobby:gitter.im), which is the general place where all of CN105 development comes together (but does include other projects).

Apologies for the inconvenience while we clean up the various GitHub-related messes!

samphonic:
Issues can be [created in our fork](https://github.com/muart-group/esphome/issues) now.

samphonic:
Okay, re: @nagyrobi 's suggested naming changes:

- `heatpump_uart` renamed to `uart_heatpump`, same with `thermostat_uart`.  This is a ❗**breaking change**, so apologies, but hopefully a quick config fix at least.
- After giving it some more thought, I think it's best to stick with `mitsubishi_uart` as the component name because:
  - This aligns better with similarly named existing `*_i2c`, `*_spi`, etc. components (where the communication type is appended and matches the base component name (e.g. `i2c:` and `spi:`)
  - `cn105` is the "name" of the physical connector, sort of, but isn't particularly descriptive about its purpose/type (i.e. the connector is *actually* an S05B-PASK-2 (or similar), and the protocol is actually "IT Protocol")

As for other things mentioned in this thread:

----

> I think that the Active Mode switch should be configurable from YAML with in order to be able to configure its internal or disabled_by_default options.

It is accessible already here:
``` yaml
mitsubishi_uart:
# ...
  active_mode_switch:
    name: Active Mode
    disabled_by_default: false
    id: mitsubishi_uart_activemodeswitch_id
    entity_category: config
    icon: mdi:upload-network
```

However, it would be nice to have this disabled entirely if `uart_thermostat` isn't defined since, as you say, it serves no purpose unless a secondary controller is connected.  I've created [this issue](https://github.com/muart-group/esphome/issues/35) to track this enhancement.

----

> I'd probably add an extra binary sensor to show when the connection between ESPHome and the heatpump is OK

Reasonable, though I think tracking through "unavailable" might be the most consistent.  [Issue created](https://github.com/muart-group/esphome/issues/36)

----

Re: Cassette vanes

As @KazWolfe said, this is just a matter of learning *if/how* they can be controlled.  I'll let you guys add an issue since I don't know as much about that one.

----

Re: Autoconf

It sounds really great, but as we keep coming across older/newer units with "exciting" differences, I'm really hesitant to automatically disable anything in case this interferes with some users' ability to control their units.  I'd rather have e.g. a disabled-by-default horizontal vane direction select that I don't see or use, than have a unit with horizontal vanes that incorrectly gets "autoconfigured" without support.  Work on this [is being tracked here](https://github.com/muart-group/esphome/issues/13) though.  So it's not a "no", just something I think should be added very cautiously after we're sure it works.

----

It seems like the renaming thing was the only PR-blocking change suggested so far as the others don't break anything.  However I think the active switch and unavailable enhancements would really improve usability so we'll try to add those soon.

clydebarrow:
I'd strongly suggest adding clarification both in code comments and the docs that this applies to compatible Mitsubishi _Electric_ heat pumps, and does not work with Mitsubishi Heavy Industries products.

nagyrobi:
> Also seems reasonable, though our climate entity should report the heat pump as `Unavailable` in this state (I think?).

The `unavailable` state cannot be set from ESPHome side yet (the entities can go `unavailable` only when the entire node disconnects from HA, see https://github.com/esphome/feature-requests/issues/1568). Afaik `unknown` is also a limited possibility, applicable only to numeric sensors (set it to `nan`). That's why a separate binary sensor would be useful to clearly show this.

nagyrobi:
> After giving it some more thought, I think it's best to stick with `mitsubishi_uart` as the component name because:

As Clyde also pointed a good example case above, I think it's unfair to take the universal `mitsubishi_uart` naming, because Mitsubishi manufactures not only HVACs, but many, many other products that could qualify once to be controlled from ESPHome via UART. And yes, there's not only Mitsubishi Electric but also Heavy Industries. Perhaps use `mitsubishi_muart`, or `muart_hvac` then if you don't like `mitsubishi_c105`.

nagyrobi:
> However, it would be nice to have this disabled entirely if `uart_thermostat` isn't defined since, as you say, it serves no purpose unless a secondary controller is connected. I've created [this issue](https://github.com/muart-group/esphome/issues/35) to track this enhancement.

I was about to suggest the same thing.

nagyrobi:
> It seems like the renaming thing was the only PR-blocking change suggested so far as the others don't break anything. However I think the active switch and unavailable enhancements would really improve usability so we'll try to add those soon.

Reviewing the PRs usually takes longer time. There's no priorization or any kind of thumb rules but generally more complex things require more effort to be reviewed. 

I'll take a stab at the doc PR a bit later to help following better the ESPHome docs structure. Perhaps I'll wait until you finish the changes related to the active switch.

nagyrobi:
How is `selects` option supposed to be used? Getting error:

```
  expected a dictionary.
  selects: 
    - vane_position_select
```

ignisf:
Currently testing this PR with a MSZ-AP50VGK with an esp32 with the IDF framework. Seems to work fine.

```
[01:24:50][C][mitsubishi_itp:115]: Discovered Capabilities: Identify Base Capabilities 
 HeatDisabled:No SupportsVane:Yes SupportsVaneSwing:Yes DryDisabled:No FanDisabled:No ExtTempRange:Yes AutoFanDisabled:No InstallerSettings:No TestMode:No DryTemp:Yes StatusDisplay:Yes
 CoolDrySetpoint:16.000000/31.000000 HeatSetpoint:10.000000/31.000000 AutoSetpoint:16.000000/31.000000 FanSpeeds:5
```

paravoid:
There are at least 6 different ESPHome components implementing this functionality ([geoffdavis/esphome-mitsubishiheatpump](https://github.com/geoffdavis/esphome-mitsubishiheatpump), [wrouesnel/esphome-mitsubishiheatpump](https://github.com/wrouesnel/esphome-mitsubishiheatpump), [echavet/MitsubishiCN105ESPHome](https://github.com/echavet/MitsubishiCN105ESPHome), [dark-nico/esphome-mitsubishiheatpump](https://github.com/dark-nico/esphome-mitsubishiheatpump), PR #5265, MUART). Many of them forks of each other, and some leverage the Swicago Arduino library, which is great  but integrates poorly with ESPHome. As far as I know, these are all friendly forks or sibling projects.

MUART is a clean-room implementation, with code that is upstreamable: implements the protocol directly rather than using Swicago, uses ESPHome's UART abstractions, and is under `climate`. I've reviewed the code and is well-written, well-abstracted and well-documented (and in English rather than French like another project above :grin:). It also has a few features that other projects do not, like interfacing with Mitsubishi's own hardware thermostats.

The project's own [documentation](https://muart-group.github.io/) is also excellent, as they have gone through the trouble of reverse engineering several features of the protocol and documenting the protocol itself in the website, for any implementation to use. Finally, the MUART folks (@Sammy1Am & @KazWolfe in particular) have been great in responding to bugs, feature requests and PRs.

I've used several of the projects I listed above, but have settled into using MUART. I've been using this for many weeks now, and it's been rock stable.

It'd be good to merge this already, because a) there is clearly demand by users for it b) it's in a good working state, c) not having this upstream has resulted into a proliferation of projects that is both confusing to new users, and fragmenting developer activity.

nagyrobi:
How is `selects` option supposed to be used? Getting error:

```
  expected a dictionary.
  selects: 
    - vane_position_select
```

ignisf:
@nagyrobi, I use it as follows:

```yaml
climate:
  - platform: mitsubishi_itp
    name: Living Room Air Conditioner
    uart_heatpump: hp_uart
    temperature_sources:
      - roomtemp
    selects:
      temperature_source_select:
        name: Living Room Air Conditioner Temperature Source

sensor:
  - platform: xiaomi_lywsd03mmc
    mac_address: "XXX"
    bindkey: "XXX"
    temperature:
      name: "Living Room Air Conditioner Remote Temperature"
      internal: true
      id: 'roomtemp'
      force_update: true
```

nagyrobi:
I made a comment in the doc PR that this needs clarification for the users.

esphome/components/mitsubishi_itp/climate.py
```diff
@@ -0,0 +1,462 @@
...
+        # Overwrite name from ENTITY_BASE_SCHEMA with "Climate" as default
+        cv.Optional(CONF_NAME, default="Climate"): cv.Any(
+            cv.All(
+                cv.none,
+                cv.requires_friendly_name(
+                    "Name cannot be None when esphome->friendly_name is not set!"
+                ),
+            ),
+            cv.string,
+        ),
```
jesserockz: We should not be putting default names. This should be user specified.
```suggestion
```

esphome/components/mitsubishi_itp/climate.py
```diff
@@ -0,0 +1,462 @@
...
+
+AUTO_LOAD = [
+    "climate",
+    "select",
+    "sensor",
+    "binary_sensor",
+    "button",
+    "text_sensor",
+    "time",
+]
```
jesserockz: Newer components are not allowed to AUTO_LOAD all of these base platform components.
The code should be written using interfaces/listeners and subclasses or preprocessors (`#ifdef`) to safely compile without the headers being available.

esphome/components/mitsubishi_itp/climate.py
```diff
@@ -0,0 +1,462 @@
...
+DEPENDENCIES = [
+    "uart",
+    "climate",
```
jesserockz: Not required as this is a climate platform, so climate is always going to be loaded.
```suggestion
```

esphome/components/mitsubishi_itp/climate.py
```diff
@@ -0,0 +1,462 @@
...
+]  # These will always be available
+
+validate_custom_fan_modes = cv.enum(CUSTOM_FAN_MODES, upper=True)
+
+BASE_SCHEMA = climate.CLIMATE_SCHEMA.extend(
+    {
+        cv.GenerateID(CONF_ID): cv.declare_id(MitsubishiUART),
+        cv.Required(CONF_UART_HEATPUMP): cv.use_id(uart.UARTComponent),
+        cv.Optional(CONF_UART_THERMOSTAT): cv.use_id(uart.UARTComponent),
+        cv.Optional(CONF_TIME_SOURCE): cv.use_id(time.RealTimeClock),
```
jesserockz: Please use the existing `CONF_TIME_ID` const.
```suggestion
        cv.Optional(CONF_TIME_ID): cv.use_id(time.RealTimeClock),
```

This could also be automatically set if the user has a `time` config:
```suggestion
        cv.OnlyWith(CONF_TIME_ID, "time"): cv.use_id(time.RealTimeClock),
```

esphome/components/mitsubishi_itp/climate.py
```diff
@@ -0,0 +1,462 @@
...
+        for button_designator, (button_name, button_schema) in BUTTONS.items()
+    }
+)
+
+
+CONFIG_SCHEMA = BASE_SCHEMA.extend(
+    {
+        cv.Optional(CONF_SENSORS, default={}): SENSORS_SCHEMA,
+        cv.Optional(CONF_SELECTS, default={}): SELECTS_SCHEMA,
+        cv.Optional(CONF_BUTTONS, default={}): BUTTONS_SCHEMA,
```
jesserockz: These should all be in their own `sensor.py` file etc.

We do not cross mix component types in a single one.

Either this `climate` can be the hub, or make a new top level hub to join everything together.

```yaml
mitsubishi_itp:
  uart_thermostat: ...
  
climate:
  platform: mitsubishi_itp
  uart_id: ...
  
sensor:
  - platform: mitsubishi_itp
    thermostat_temperature:
      name: Thermostat Temperature
    outdoor_temperature:
      name: Outdoor Temperature
    ...

button:
  - platform: mitsubishi_itp
    filter_reset_button:
      name: Filter Reset

...
```

jesserockz (changes requested):
Given the below comments, I have not reviewed most of the code as I believe it will need significant changes to conform to the ESPHome codebase styles.

kpfleming:
I am happy to be a tester/reviewer for this as well, I've been anxiously awaiting a proper ESPHome implementation of this and have lots of hardware ready and waiting :-) I'll definitely be using the MHK2-passthrough mode (on two systems).

esphome/components/mitsubishi_itp/climate.py
kpfleming: This will also eliminate the need for a configuration attribute to 'hide' sensors; if they are separate items which must appear in the YAML config to be visible, then they won't be visible unless the user has included them in their YAML config. This makes the code a little more complex as it has to account for the fact that the user may have decided they don't want the 'filter_reset' button (in which case it will be NULL).

ignisf: As a user, it did indeed stand out as inconsistent with the rest of the component configurations. Functionality-wise the integration works great though.

KazWolfe: > This will also eliminate the need for a configuration attribute to 'hide' sensors; if they are separate items which must appear in the YAML config to be visible, then they won't be visible unless the user has included them in their YAML config. This makes the code a little more complex as it has to account for the fact that the user may have decided they don't want the 'filter_reset' button (in which case it will be NULL).

Something to note here is that we eventually plan to set up autoconf, which may also dynamically hide/show sensors based on hardware support flags. How would this affect our plans to do that, and/or should we just not care until that's all ready?

henrykuijpers:
I hope this PR will not get stalled due to the many change requests. The functionality that this PR is providing is highly desired by the many Mitsubishi air conditioning users. 

I believe this PR is working with Mitsubishi Heavy Industries units? Is it also working with Mitsubishi Electric units? Or am I wrong? 

It would be very valuable to have one centralised Mitsubishi component. 

Also, another question: What does "ITP" stand for?

henrykuijpers:
Oh, and, there is a newer / more recent PCB design submitted by @oliv3r: https://github.com/absalom-muc/MHI-AC-Ctrl/pull/181 it contains some enhancements and makes the MHI-AC-Ctrl work with an ESP32. It would be very nice to make this PR compatible with this version of the PCB as well.

nagyrobi:
`IT Protocol` as it's named the purpose of the CN105 connector in the Mitsu docs.

Not Heavy Industries, just Electric afaik.
This component doesn't need any custom PCB, the ESP can be directly connected to CN105.

esphome/components/mitsubishi_itp/climate.py
ignisf: @KazWolfe, I can think of one example of an optional feature – the receiver functionality of some `climate_ir` platforms (https://esphome.io/components/climate/climate_ir.html#using-a-receiver).

Couldn't you ignore or issue a warning if a given sensor is not supported by the current device?

KazWolfe: @ignisf We probably could - ultimately we'd still need to figure out how to actually run communication init before setting things up. Once that part's working, we can probably do a bunch of things to configure defaults/allow overrides - or at least look at how best to do that while keeping things feeling like ESPHome.

I think @Sammy1Am's priority right now though is to get everything *else* good to PR.

jesserockz: Please focus on the core functionality first. This PR is already big at 3400 lines and ideally would be broken up into smaller PRs :pray: to make the review process go smoother and easier for everyone.

esphome/components/mitsubishi_itp/climate.py
samphonic: I will do some digging for one myself, but if you happen to have an example of a newer component that does this correctly that would be helpful.  (I mostly just used other components as examples when writing the code, but I guess picked older ones)

esphome/components/mitsubishi_itp/climate.py
samphonic: Because a fairly normal heat pump setup may come with ~15 sensors / configuration options out of the box, the goal here was to avoid requiring the user to add all of those explicitly to the `yaml`.  Sure, they could copy and paste a giant block of text from the documentation, but that seems cumbersome and error-prone.

Is there no acceptable way to include sensors or configuration options with a component by default?

> This PR is already big at 3400 lines and ideally would be broken up into smaller PRs 🙏

That *would* be ideal, I agree, but I'm not sure offhand how to break up a fully-functioning component into smaller PRs (other than have PRs that... like add a source file at a time, but aren't yet usable because of the missing pieces?)  Suggestions welcome though, I also don't love having to manage this giant PR.

samphonic:
@kpfleming : You're welcome to start testing with the documentation [available here](https://muart-group.github.io/user) to pull from the mUART repo directly.  The component is pretty stable and functionality-wise there aren't really any upcoming changes. 
 **However**, the requested changes in this PR are likely to break your configuration pretty soon.  You could use `@9ebfd9a` instead of `@dev` in the `external_component` configuration to keep things stable until you're ready to update to the new configuration (when available).

@henrykuijpers : As @nagyrobi said, there's no Heavy Industries support at this time.  I also think it would be nice to have a centralized Mitsubishi component, but if the protocols and features are sufficiently different, it might make more sense to keep them separate (e.g. there is a separate `mitsubishi` component that only has IR support, but it's sufficiently different that we didn't try to just extend that one for UART support).  We switched from `mitsuibishi_uart` to `_itp` as the `IT Protocol` is more specifically what this component is targeting.

Sorry everyone for being absent for a while, going to try to push over the next few days to address the remaining PR comments (left a couple questions up there, so hopefully can get those resolved and move forward ASAP).  The component itself continues to be stable as far as testers have reported, so once the style issues are addressed I'm optimistic it should be a smooth merge).

esphome/components/mitsubishi_itp/climate.py
kpfleming: > Is there no acceptable way to include sensors or configuration options with a component by default?

There isn't. It's a policy in ESPHome that users always configure every entity in their YAML configuration, even if it's just copy-and-paste from an example configuration. This gives them complete control over the presence and configuration of the entities.

samphonic: Alright, a follow-up question because I still can't find any documentation or good examples:
Previously we had `mitsubishi_itp` as a top-level component, but based on feedback moved it under `climate:`.  If it (I think this is what's being called a "hub") is going to be under `climate:`, how do I get a reference to it in `sensors.py`?

Previously I just did `muart_component = cg.new_Pvariable(config[CONF_ID], hp_uart_component)` when it was all together, but now it seems like I could use `get_variable` except the ID will be user-defined so I don't know what it is.

Additionally, am I correct in assuming we'll need to enforce that only one mitsubishi_itp climate component exists now?  Previously there were no technical issues with defining more than one for the same device, but now because the sensors need to be able to reference the hub, there can only be one hub.  Am I understanding correctly?

(Apologies for all the questions-- if there *is* some documentation on this somewhere that I couldn't find I'd happily take a pointer to it.  Otherwise I'm just still trying to wrap my head around how all this works)

clydebarrow: Where hubs are used, multiple hubs of the same type can exist. The link to a hub from a client component is done with `cv.use_id(<idtype>)` which automagically yields an ID if a single one of the correct type has been defined by a hub, fails if none are defined, and requires an explicit user-nominated id if more than one exists. Look at components using `uart` for examples.

kpfleming: Yep, using a hub plus climate/sensor/select components is the way to go.

If there's a single ITP connection, then the hub would look like this:
```
mitsubishi_itp:
  uart_heatpump: <id of uart>
  uart_thermostat: <id of uart>
  temperature_sources:
    - ...
    - ...
```

In this configuration the hub would get an automatically generated ID, or the user can specify one by including an `id: <id>` attribute.

If there are multiple ITP connections (assuming the board has enough UARTs!), then the hub would look like this:
```
mitsubishi_itp:
  - id: itp_one
    uart_heatpump: <id of uart>
  - id: itp_two
    uart_heatpump: <id of uart>
    uart_thermostat: <id of uart>
```

In the configuration schema you'd use `ensure_list` to support these two styles (an 'implicit' list of one item, or an explicit list of one or more items).

For the climate/sensor/select components, with a single ITP connection then the configuration would look like this:
```
climate:
  - platform: mitsubishi_itp
    supported_modes:
      - ...
        ...
```
In this situation no explicit `id` attribute is required because there is only one hub and the ID will be autodetermined (using `cv.use_id()` in the configuration schema).

With multiple ITP connections then configuration would look like this:
```
climate:
  - name: climate_1
    platform: mitsubishi_itp
    itp_id: itp_one
    supported_modes:
      - ...
        ...
  - name: climate_2
    platform: mitsubishi_itp
    itp_id: itp_two
    supported_modes:
      - ...
        ...
```

nagyrobi: Dallas sensor style...

samphonic: Awesome, thank you for the detailed explanation, that makes a lot of sense.  A little annoyed that I had this as a top-level-component initially 😞and only moved it under `climate` based on feedback; but I've learned a lot more about the configuration structure now and think what you're proposing is probably the most organized.

esphome/components/mitsubishi_itp/climate.py
samphonic: Okay, I've done some digging and found [as3935](https://github.com/esphome/esphome/blob/dev/esphome/components/as3935/as3935.cpp) which seems to be a good example of using the preprocessors.  But when we've [got a lot of sensors](https://github.com/muart-group/esphome/blob/pr-changes/esphome/components/mitsubishi_itp/mitsubishi_uart-packetprocessing.cpp) (search for "\_sensor\_" on that page), and sensor logic in various places, it seems like peppering `#ifdef` all over might make things less readable.

I'm not sure what you mean my "interfaces/listeners and subclasses" though; could you just give me a high-level hint at what you mean (or point to a component that does that) so I can see if that direction will help keep the code readable?

esphome/components/mitsubishi_itp/climate.py
samphonic: After a little digging and thought, the "hub" bits and the climate bits are very tightly coupled.  It also looks like all the other climate integrations use a climate-hub, rather than separate bits, so we're going to stick with that for now.

I really like the consistency that breaking out the climate part would provide for where things are configured, but almost every function of the hub communicates with the climate somehow, so as long as climate-hub is a feasible option I think it helps the code.

samphonic:
Apologies for the delay,

### Just merged some changes to address:

- No default names
- Use time_id
- Don't depend on climate
- Break out sensor, binary_sensor, text_sensor, select, and button into their own directories
- Convert sensors and selects into listeners to avoid needing AUTO_LOAD
- List of temperature sensors moved to temperature_source select component (this simplifies the code since the climate-hub doesn't actually need a whole list of temperature sources, just the name of the current one)
- Preferences to store last temperature source moved to temperature_source select component

### These changes break the configuration in the following ways:

The climate component will now require a `name` or `id`:
```yaml
climate:
  - platform: mitsubishi_itp
    name: "Climate"
    uart_heatpump: hp_uart
```

Sensors have been moved out to their own configuration areas instead of being under the climate component:
```yaml
text_sensor:
  - platform: mitsubishi_itp
    actual_fan:
      name: "Actual Fan"
```

By default **NO** sensors will be present.  You have to explicitly define all of them now (this is an ESPHome requirement)

The additional temperature sources are now under the select component:
```yaml
select:
  - platform: mitsubishi_itp
    temperature_source:
      name: "Temperature Source"
      sources:
        - fake_temp
        - home_assistant_temperature
    vane_position:
      name: "Vane Position"
```

----

I think that's all the currently outstanding comments addressed.  It certainly has changed things enough that older comments may no longer apply.  Once the tests pass I will mark this as Ready to Review.

I will also begin updating the documentation to match the new implementation.

samphonic:
ESPHome documentation ([see preview here](https://deploy-preview-3860--esphome.netlify.app/components/climate/mitsubishi_itp)) has been updated in case anyone wants to try it out or help with testing.  I'll update our project documentation soon.

esphome/components/mitsubishi_itp/climate.py
```diff
@@ -0,0 +1,132 @@
...
+    CONF_SUPPORTED_FAN_MODES,
+    CONF_SUPPORTED_MODES,
+    CONF_TIME_ID,
+)
+from esphome.core import coroutine
+from . import mitsubishi_itp_ns, MitsubishiUART
+
+AUTO_LOAD = [
+    "time",
+]
```
jesserockz, line 20: Time cannot be autoloaded as it requires explicit user configuration
```suggestion
```

esphome/components/mitsubishi_itp/muart_packet.h
```diff
@@ -0,0 +1,572 @@
...
+#define CONSOLE_COLOR_CYAN "\033[0;36m"
+#define CONSOLE_COLOR_CYAN_BOLD "\033[1;36m"
+#define CONSOLE_COLOR_WHITE "\033[0;37m"
+
+const std::string FAN_MODE_VERYHIGH = "Very High";
+
+// These are named to match with set fan speeds where possible.  "Very Low" is a special speed
+// for e.g. preheating or thermal off
+const std::array<std::string, 7> ACTUAL_FAN_SPEED_NAMES = {"Off",  "Very Low",        "Low",  "Medium",
+                                                           "High", FAN_MODE_VERYHIGH, "Quiet"};
```
jesserockz, line 27: Why is only one of these defined outside as its own string?

jesserockz (changes requested):
Can we put all platforms (apart from climate) into a separate PR please?

https://github.com/esphome/esphome/pull/6794#discussion_r1710103529

If written correctly (not yet reviewed due to size of PR) then you simply don't need those folders (sensor/binary_sensor/button) etc included in this PR at all and it should still work as expected.
They can be added in a part 2/3/4 PR that can be reviewed and merged quicker after this one is all done and dusted. While it may make things "unusable" in `dev`, we can time the merging of the PRs so that a release does not go out with half the code.

Also, if you want to put all of your Packet classes and processors etc into a standard c++ library, then that also helps to cut down on the amount of code in ESPHome that needs reviewing. (We still look at the library code briefly to make sure there is nothing malicious, but code styles etc are not enforced on that). As long as the library is only doing data processing, and the communications and entities are left in ESPHome, then there is no problem. The `haier` climate does this already.

I think also after a quick look through, the thermostat uart is optional so that and its bridge code could also be put into a followup PR.

Also please make sure that all of the published [codebase standards](https://esphome.io/guides/contributing#codebase-standards) are followed.

Jesse

samphonic:
> Can we put all platforms (apart from climate) into a separate PR please?

I'm happy to attempt this; can you give me some pointers on how best to do so?  Development has cooled down a little in our repo, so it's not *as* difficult to create, like, a separate fork?  Is there a git way to do a PR on only certain files or directories?  (That would be easiest, I think)

> Also, if you want to put all of your Packet classes and processors etc into a standard c++ library, then that also helps to cut down on the amount of code in ESPHome that needs reviewing.

We've talked about this, but haven't really had time to figure out the best way to do so; we sort of made the PR a priority before looking at additional internal improvements.  It might be quicker/easier to make sure the code conforms to codestyles/etc. than to pull the whole thing out right now; if you think this is going to be a blocker for the PR, then we'll have to figure it out.

> I think also after a quick look through, the thermostat uart is optional so that and its bridge code could also be put into a followup PR.

I guess it depends a little on how splitting up the PR is going to work; some of the logic is fairly intertwined so it might be more complicated than just splitting on files.

esphome/components/mitsubishi_itp/climate.py
samphonic: If not explicitly configured we're just setting the thermostat clock to midnight (because it seemed unnecessary to *require* users to configure a time source even if they weren't using a thermostat).  There might be a way to refactor this similar to turning all the sensors, etc. into listeners though, so we'll look into that.

esphome/components/mitsubishi_itp/muart_packet.h
samphonic: "Very High" is defined as a `CUSTOM_FAN_MODE` [here](https://github.com/esphome/esphome/pull/6794/files/dcac9bfedb9aa2fc2f2b9c2c957e97371cee7bf1#diff-5c021af2c0fbf5ba81e7e06fe726c55cfc7fb0995218389efbc794d98cc5cb67R36-R37) because it isn't one of the built-in fan modes available from ESPHome.

"Very Low" also isn't a built-in fan mode, but it's not actually possible for the user to select it so it didn't need to be defined as a custom mode.

I suspect the reason they're not _all_ defined outside as strings is because whatever climate component I was using as an example didn't, but we could definitely define FAN_MODE_VERYLOW ourselves, and try to reference the others from the ESPHome code.

esphome/components/mitsubishi_itp/climate.py
jesserockz: The c++ code can check if the time is configured (`this->time_id_ != nullptr`) and also if `time` is even in the config (`#ifdef USE_TIME`)

jesserockz:
> a separate fork? Is there a git way to do a PR on only certain files or directories? (That would be easiest, I think)

That would be a git branch =)
Instead of committing stuff directly to your `dev` branch, the work should be done in a different branch and a PR made from that branch. This way you can also keep your dev branch with all of your code which people can use and test, and allows the code to be "drip-fed" into the ESPHome codebase in parts.

> We've talked about this, but haven't really had time to figure out the best way to do so; we sort of made the PR a priority before looking at additional internal improvements. It might be quicker/easier to make sure the code conforms to codestyles/etc. than to pull the whole thing out right now; if you think this is going to be a blocker for the PR, then we'll have to figure it out.

It was just a suggestion as it can make the review go easier for everyone as all of your protocol code wont be as scrutinised =)
No problems if you want to leave that as is ans it can always be extracted out later.

> I guess it depends a little on how splitting up the PR is going to work; some of the logic is fairly intertwined so it might be more complicated than just splitting on files.

Yes this part will be more complex because the code is mixed in with other files. It's very difficult to review such large PRs and we really want them to be the minimum amount of code required, which means optional parts should be added later.

esphome/components/mitsubishi_itp/climate.py
samphonic: Yeah, using `#ifdef` in this case should work fine.  This will be addressed in the new PR.

samphonic:
Closing in favor of the smaller PR #7289

PR #6798, opened 2024-05-23
Adding support for PICsimLab

geiseri:

I have been using [PICsimLab](https://github.com/lcgamboa/picsimlab) lately for testing.  It uses qemu on the backend so it requires that  `--fill-flash-size` be set when building the final bin file.  The current change is hamfisted and not desirable by default since it would make for huge bin files for an actual ESP32.  Would there be suggestions as how to fix this?  I was thinking of adding a board type for it since it emulates the DevKitC enough that `esp32dev` works with some adjustments to keep qemu happy.  I was not clear from that how to modify those flash options though.

I am not sure if there is enough interest for this, if there is I would be happy to clean it up.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


geiseri:
@jesserockz is there a way that I can pass something into the script based on the board type?  For picsimlab I also need to set the platform io option `board_build.flash_mode: dio`  to be compatible with qemu emulator's limitations.  Is this something that can be addressed in a board configuration? aka setting `esp32.board` to something like  `picsimlab`.

swoboda1337:
Closing as it is stale and seems there was a bad merge diff doesn't make sense.

PR #6800, opened 2024-05-24
[voice_assistant] Don't allocate buffers until starting the microphone for the first time

jesserockz:

Instead of allocating all of the buffers at setup time which is not required, allocate them later when we first might need them.

This frees up memory for the bootup sequence and other things like BLE when a new device is being provisioned and the voice assistant is not needed yet.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


Taomyn:
@jesserockz this broke my Atom, caused it to loop initialising the buffer when starting the microphone, had to revert ESPHome in Home Assistant to 2024.5.2 and re-compile/flash to get it working again - fyi, I never had the original issue in the first place, the wake word was fine and audio played as expected.

PR #6803, opened 2024-05-24
Add rotation config parameter for stepper component.

cvwillegen:

In certain configurations, you might want a stepper motor to only run in one direction (e.g. when there's an object blocking rotation in one direction). Or when a stepper motor is used to indicate the hands on a clock.

This PR adds the possibility to add a ``rotation`` parameter to the stepper motor config, forcing a specific direction when running the motor. Note! The user must take care of not running onto an infinite loop when using ``set_target`` on the stepper motor, i.e. by using the ``report_position`` when the position of the stepper motor is known.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3864

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: test-stepper
  friendly_name: test_stepper

esp8266:
  board: d1_mini

# Enable logging
logger:

# Enable Home Assistant API
api:
  encryption:
    key: "cxxxxxxxxxxxxxxxxcccxxaaa"

ota:
  password: "xxxxxxxxxxxxxxx"

wifi:
  ssid: !secret wifi_ssid

  ap:
    ssid: "Test-Stepper Fallback Hotspot"
    password: "xxxxxxxxxxx"

captive_portal:

stepper:
  - platform: uln2003
    pin_a: D1
    pin_b: D2
    pin_c: D3
    pin_d: D4
    id: my_stepper
    max_speed: 70 steps/s

globals:
  - id: g_position
    type: int
    restore_value: no
    initial_value: '0'

number:
  - platform: template
    id: position
    min_value: 0
    max_value: 2048
    step: 1
    name: Position
    internal: False
    mode: BOX
    on_value: 
      then:
        - lambda: |-
            id(g_position) = x;
    set_action: 
      then:
        - stepper.set_target: 
            id: my_stepper
            target: !lambda return x;

button:
  - platform: template
    name: CCW
    id: ccw
    on_press:
      - stepper.set_rotation:
          id: my_stepper
          rotation: COUNTERCLOCKWISE

  - platform: template
    name: CW
    id: cw
    on_press:
      - stepper.set_rotation:
          id: my_stepper
          rotation: CLOCKWISE

  - platform: template
    name: Both
    id: both
    on_press:
      - stepper.set_rotation:
          id: my_stepper
          rotation: BOTH
```

## Checklist:


esphome/components/stepper/__init__.py
```diff
@@ -66,11 +68,29 @@ def validate_speed(value):
...
+        f"Expected rotation as 'both', 'cw', 'ccw', 'clockwise', 'counterclockwise', 'counter-clockwise', got {value}"
+    )
+
+
 STEPPER_SCHEMA = cv.Schema(
     {
         cv.Required(CONF_MAX_SPEED): validate_speed,
         cv.Optional(CONF_ACCELERATION, default="inf"): validate_acceleration,
         cv.Optional(CONF_DECELERATION, default="inf"): validate_acceleration,
+        cv.Optional(CONF_ROTATION, default="both"): validate_rotation,
```
jesserockz: Please use an enum in c++ and `cv.enum` in python for this.

jesserockz (changes requested):

esphome/components/stepper/__init__.py
cvwillegen: Thanks for the hint!

esphome/components/stepper/__init__.py
```diff
@@ -66,11 +68,22 @@ def validate_speed(value):
...
 
+Rotation = stepper_ns.enum("Rotation")
+ROTATIONS = {
+    "BOTH": Rotation.ROTATION_BOTH,
+    "CW": Rotation.ROTATION_CW,
+    "CLOCKWISE": Rotation.ROTATION_CW,
+    "CCW": Rotation.ROTATION_CCW,
+    "COUNTERCLOCKWISE": Rotation.ROTATION_CCW,
+    "COUNTER-CLOCKWISE": Rotation.ROTATION_CCW,
+}
```
jesserockz, line 74: Lets not add multiple options per each type. It can add confusion and be less clear in the documentation.

```suggestion
ROTATIONS = {
    "BOTH": Rotation.ROTATION_BOTH,
    "CLOCKWISE": Rotation.ROTATION_CW,
    "COUNTERCLOCKWISE": Rotation.ROTATION_CCW,
}
```

esphome/components/stepper/stepper.cpp
```diff
@@ -38,7 +38,8 @@ int32_t Stepper::should_step_() {
   // assumes this method is called in a constant interval
   uint32_t dt = now - this->last_step_;
   if (dt >= (1 / this->current_speed_) * 1e6f) {
-    int32_t mag = this->target_position > this->current_position ? 1 : -1;
+    int32_t mag = (rotation_ == ROTATION_BOTH ? (this->target_position > this->current_position ? 1 : -1)
+                                              : (rotation_ == ROTATION_CW ? 1 : -1));
```
jesserockz: Please split this out so there are not embedded ternaries

jesserockz (changes requested):

cvwillegen:
The failure looks unrelated to my changes. I'll update the branch later on to see if it makes any difference...

esphome/components/stepper/__init__.py
```diff
@@ -82,6 +92,8 @@ async def setup_stepper_core_(stepper_var, config):
         cg.add(stepper_var.set_deceleration(config[CONF_DECELERATION]))
     if CONF_MAX_SPEED in config:
         cg.add(stepper_var.set_max_speed(config[CONF_MAX_SPEED]))
+    if CONF_ROTATION in config:
+        cg.add(stepper_var.set_rotation(config[CONF_ROTATION]))
```
jesserockz: As there is a default, we dont need to check
```suggestion
    cg.add(stepper_var.set_rotation(config[CONF_ROTATION]))
```

jesserockz (changes requested):
Changes look good. Please add to the test yaml files

PR #6834, opened 2024-05-30
Update const.py added missing millimeter

NonaSuomy:
Added missing millimeter

millimeter was missing from the const.py

Only added this missing value of mm


jesserockz (dismissed):
I can see several places that this unit is already used. 
If you want to add to const.py, please consolidate them all

kbx81 (approved):
Thanks!

PR #6837, opened 2024-05-31
[config] Early termination of validation steps on error

clydebarrow:

If the host: platform is used, and an invalid key is specified in the config for it, the platform never gets set and there is a subsequent exception thrown.
```
  File "/Users/clyde/dev/opensourceprojects/esp/esphome/esphome/core/__init__.py", line 634, in target_platform
    return self.data[KEY_CORE][KEY_TARGET_PLATFORM]
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
KeyError: 'target_platform'
```

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml

esphome:
  name: test

host:
  mac_addr: "62:23:45:AF:B3:DD" # should be mac_address
```

## Checklist:


jesserockz (commented):
Would this not apply to all platforms? They all have this function call at the bottom

clydebarrow:
> Would this not apply to all platforms? They all have this function call at the bottom

You would think so, but with esp32 a bad key is reported as expected, but not on host. It would make sense to move all of them to the top, I just didn't want to make unnecessary changes.

clydebarrow:
> Would this not apply to all platforms? They all have this function call at the bottom

So the ESP32 and ESP8266 and probably the others can't have the validation function first, since those functions use data provided in the config, like the framework and board.

I'm still not 100% sure why there is a difference, but I think I got closer to the root cause - when executing a sequence of validation steps, an error in any step should terminate the process, since subsequent validations may depend on successful earlier ones.

Code updated accordingly. See what you think.

jesserockz (approved):

PR #6841, opened 2024-05-31
[dsmr] Add internal 'telegram' text_sensor to support bridging

quokka2020:
# [DSMR] Adding internal 'telegram' text_sensor to support bridging 

By creating an internal _telegram_ text_sensor it's possible to forward the telegram to another uart (P1 bridge) or simply log it.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3890

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
text_sensor:
- platform: dsmr
   telegram:
      name: "telegram"
      disabled_by_default: true
      on_value:
        then:
          - lambda: |-
              ESP_LOGI("dsrm", "telegram is %s", x.c_str());
              p1_bridge_uart->write_str(x.c_str());
```

## Checklist:


nagyrobi:
What a great idea!

esphome/components/dsmr/dsmr.cpp
```diff
@@ -17,6 +17,10 @@ void Dsmr::setup() {
   if (this->request_pin_ != nullptr) {
     this->request_pin_->setup();
   }
+  // Prevent telegram to be pubished elsewhere
+  if (this->s_telegram_ != nullptr) {
+    this->s_telegram_->set_internal(true);
+  }
```
jesserockz: This should be set in the config schema to default `internal: true` instead. 
The `homeassistant` platform of sensors does this.

jesserockz (changes requested):

esphome/components/dsmr/dsmr.cpp
quokka2020: Hi @jesserockz, first of all thanks for your comment. I'm new to esphome(development), but I can't figger out how to make this sensor _internal_ by default. I assume this can be done in text_sensor.py? Or do you mean it should be done in the yaml for the device?

nagyrobi: See https://github.com/marcovaneck/esphome/blob/c4fe6428fc24558922c73aa6aee7571a092da483/esphome/components/homeassistant/__init__.py#L12 that's all

quokka2020: Hi @jesserockz, the working of default is unclear to me. _default=True_ didn't do the trick. But I agree it's not nice to force internal to be turned off. So I removed the code block and added _internal: true_ to the documentation.

esphome/components/dsmr/text_sensor.py
```diff
@@ -22,6 +22,7 @@
         cv.Optional("water_equipment_id"): text_sensor.text_sensor_schema(),
         cv.Optional("sub_equipment_id"): text_sensor.text_sensor_schema(),
         cv.Optional("gas_delivered_text"): text_sensor.text_sensor_schema(),
+        cv.Optional("telegram"): text_sensor.text_sensor_schema(),
```
jesserockz: This wont pass CI, but is how you do it.
```suggestion
        cv.Optional("telegram"): text_sensor.text_sensor_schema().extend({cv.Optional(CONF_INTERNAL, default=True): cv.boolean),
```

quokka2020: Thanks @jesserockz !
 It takes a while but now I understand how it works. I've tested you're suggested change and it of course it works. The documentation will be updated as well. (strange enough the CI pytest on linux 3.11 fails, but on the other mac/windows they succeed trying to reproduce and fix)

mhendriks:
Thanks Marco for this addition. This works really fine on various hardware (equipped with ESP32S3 and ESP32C3). 

See:
https://github.com/mhendriks/esphome-p1/blob/main/p1-dongle-pro-h2o-v2.yaml
and
https://github.com/mhendriks/esphome-p1/blob/main/ultra.yaml

Would be nice when this PR in merged into the main :-)

jesserockz (approved):

niobos:
I also needed this feature, so I was looking into adding it myself. Good thing that I checked the open PRs! Thank you!

And also a sneaky "ping" to hopefully get this merged into main...

PR #6845, opened 2024-06-01
WebSocket overrides check_origin for reverse proxy configuration

gabest11:
WebSocket server checks the ESPHOME_TRUSTED_DOMAINS environment variable for a comma separated list of allowed domains, behind a reverse proxy.

I could never get websockets work behind IIS's reverse proxy. This is what other docker containers use, I just adopted it for esphome. If the ESPHOME_TRUSTED_DOMAINS is set, the overriden check_origin compares it with the http origin header, otherwise it calls the parent and works just like before.

## Types of changes

- [ ] Other

## Checklist:


gabest11:
For the docs... I'm sure other docker environment variables are documented somewhere, but I cannot find it right now. I know about USERNAME and PASSWORD.

jesserockz (approved):

PR #6850, opened 2024-06-01
Add messages when WiFi and Ethernet components set 'warning' flag.

kpfleming:

Adds messages to the `status_set_warning` calls made by the Ethernet and WiFi components. In these situations there is no actual warning, instead the component is indicating 'warning' because it hasn't established a connection yet. Without a message to accompany the flag, the user reading a log might think they've misconfigured the component or their network.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/ethernet/ethernet_component.cpp
+++ b/esphome/components/ethernet/ethernet_component.cpp
@@ -410,7 +410,7 @@ void EthernetComponent::start_connect_() {
   global_eth_component->ipv6_count_ = 0;
 #endif /* USE_NETWORK_IPV6 */
   this->connect_begin_ = millis();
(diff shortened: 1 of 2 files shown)


clydebarrow:
Is it even necessary to set the warning flag? Why not just change those to LOG_INFO calls?

jesserockz (approved):

jesserockz:
It's because the warning state is used for the status led. So it will flash while disconnected.

PR #6865, opened 2024-06-06
[udp] Implement UDP sensor broadcast

### LINKED ISSUE #52: Automation between nodes
**Is your feature request related to a problem/use-case? Please describe.**

I have two EspHome nodes. One node is a wifi light bulb configured to control the light.
The other node is behind my wall switch and configured as input boolean/switch to control the light using an automation in homeassistant.  
The problem is: when for some reason homeassistant is not running/responding, I can't switch the light anymore.  
When the light and the switch were configured on one EspHome node I could have used an automation in EspHome to still be able to control the light.

**Describe the solution you'd like:**

I would love to have the possibility to create an automation in EspHome between different nodes.

**Additional context:**

Example:

Node1 (light):  
```yaml
esphomeyaml:
  name: example_light

light:
  - platform: cwww
    id: main_light
    name: "Main light"
    cold_white: output_cold
    warm_white: output_warm
    cold_white_color_temperature: 153 mireds
    warm_white_color_temperature: 500 mireds
    
output:
  - platform: esp8266_pwm
    pin: GPIO4
    frequency: 1000 Hz
    id: output_cold
    
  - platform: esp8266_pwm
    pin: GPIO5
    frequency: 1000 Hz
    id: output_warm
```

Node2 (switch):  
```yaml
esphomeyaml:
  name: example_switch

binary_sensor:
  - platform: gpio
    pin:
      number: GPIO5
    name: "Light button"
    on_state:
      - light.toggle: example_light.main_light
```

clydebarrow:

See https://deploy-preview-3918--esphome.netlify.app/components/udp.html for full description.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/52

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3918

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [x] BK72xx
- [x] RTL87xx
- [x] host

```yaml
# Example config.yaml
udp:
  update_interval: 5s
  encryption: "REPLACEME"
  rolling_code_enable: true
  binary_sensors:
    - binary_sensor_id1
  sensors:
    - sensor_id1
  providers:
    - name: some-device-name
      encryption: "REPLACEME with some key"

sensor:
  - platform: udp
    provider: some-device-name
    remote_id: some_sensor_id

binary_sensor:
  - platform: udp
    provider: unencrypted-device
    id: other_binary_sensor_id  # also used as remote_id
```

## Checklist:


nagyrobi:
Whooaah!

nagyrobi:
Just a suggestion: an optional `timeout` could be useful under providers\name entry (or under the sensor config), which could change the sensor to `nan` if no update packet has been received within the specified amount of time.
Or maybe a timeout binary sensor per provider.

Use case is for the consumers to know somehow that the provider is not updating. Eg. a lvgl switch widget could be greyed out when a remote switch ceases to transmit binary sensor state data.
Or for a thermostat climate we could take some measures to prevent damage when temperature sensor data is not coming in anymore.

clydebarrow:
> Just a suggestion: an optional `timeout`

Already available as a filter, and if it were to be built in that should be done in the base sensor component. Probably best left until progress has been made on a proper way to mark a sensor and a binary sensor as unavailable - `NaN` is a hack and HA keeps planning to disable it.

nagyrobi:
You mean [timeout filter](https://esphome.io/components/sensor/#timeout)? That can do it, yes.

nagyrobi:
> best left until progress has been made on a proper way to mark a sensor and a binary sensor as unavailable - `NaN` is a hack and HA keeps planning to disable it.

For reference https://github.com/esphome/feature-requests/issues/1568

esphome/components/udp/binary_sensor.py
```diff
@@ -0,0 +1,26 @@
...
+from esphome.components import binary_sensor
+from . import (
+    SENSOR_SCHEMA,
+    CONF_UDP_ID,
+    CONF_REMOTE_ID,
+    CONF_PROVIDER,
+    require_internal_with_name,
+)
+from ...config_validation import All, has_at_least_one_key
+from ...const import CONF_ID
```
jesserockz:
```suggestion
from esphome.config_validation import All, has_at_least_one_key
from esphome.const import CONF_ID
```

esphome/components/udp/sensor.py
```diff
@@ -0,0 +1,26 @@
...
+from esphome.components import sensor
+from . import (
+    SENSOR_SCHEMA,
+    CONF_UDP_ID,
+    CONF_REMOTE_ID,
+    CONF_PROVIDER,
+    require_internal_with_name,
+)
+from ...config_validation import has_at_least_one_key, All
+from ...const import CONF_ID
```
jesserockz:
```suggestion
from esphome.config_validation import has_at_least_one_key, All
from esphome.const import CONF_ID
```

esphome/components/udp/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+    PLATFORM_BK72XX,
+    PLATFORM_RTL87XX,
+    PLATFORM_HOST,
+    CONF_INTERNAL,
+)
+from esphome.cpp_generator import MockObjClass
+
+CODEOWNERS = ["@clydebarrow"]
+DEPENDENCIES = ["network"]
+AUTO_LOAD = ["socket", "sensor", "binary_sensor"]
```
jesserockz: Please use the defines to not autoload platform_components
```suggestion
AUTO_LOAD = ["socket"]
```

nagyrobi:
Amazing. Finally!

Testing with 
```yaml
external_components:
  - source: github://pr#6865
    refresh: 60min
    components: [ udp ]
```

Deployed on
- an ESP32 (Sonoff Mini R4) gathering temperatures from Xiaomi BLE sensors and pushing them through UDP directly to the thermostats (with proper sensor filtering by ESPHome).
- several ESP8266 (Sonoff Mini R1s, in wall), pushing binary states from lights to other Sonoff Minis and Lanbons, to turn subsequent lights off when exiting the "last" room.
- ESP8266 bedside light switch long-press to handle ESP32-based ceiling lights

Plan to deploy in toilets, to inform the fan relays that light switch was turned on and after a configurable delay turn the fan on.

Finally these very simple and trivial automations can be offloaded from HA! The system is much more snappy and responsive this way. No more random small delays between interactions and reactions.

esphome/components/udp/udp_component.cpp
```diff
@@ -0,0 +1,595 @@
...
+  data += *buf++ << 16;
+  data += *buf++ << 24;
+  return data;
+}
+
+static inline uint16_t get_uint16(uint8_t *&buf) {
+  uint16_t data = *buf++;
+  data += *buf++ << 8;
+  return data;
+}
```
jesserockz, line 138: We already have these in `helpers.h` -> `encode_value(...)`

esphome/components/udp/udp_component.cpp
```diff
@@ -0,0 +1,595 @@
...
+  }
+}
+
+#define CK_LEN(x) \
+  { \
+    if (end - buf < (x)) { \
+      ESP_LOGV(TAG, "Required len %u not available", x); \
+      return; \
+    } \
+  }
```
jesserockz: Not really a fan of "random" macros hiding in the code like this...

esphome/components/udp/udp_component.cpp
```diff
@@ -0,0 +1,595 @@
...
+#endif
+}
+
+void UDPComponent::send_ping_pong_request_() {
+  if (!this->ping_pong_enable_ || !network::is_connected())
+    return;
+  this->ping_key_ = random_uint32();
+  this->ping_header_.clear();
+  add(this->ping_header_, (uint8_t) MAGIC_PING);
+  add(this->ping_header_, (uint8_t) (MAGIC_PING >> 8));
```
jesserockz: Surprised you didn't add another overload for uint16_t... :laughing:

jesserockz (dismissed):
Here is a partial review

esphome/components/udp/udp_component.cpp
clydebarrow: Nope, not the same thing. These increment the pointer. I don't see an equivalent in helpers.h.

esphome/components/udp/udp_component.cpp
clydebarrow: Agreed, it just gets very wordy if it's all written out inline. The decode function is already quite long, I tried to condense the source as much as possible. The macro is defined within the cpp file so no pollution of global namespace is happening.  Do you have a better idea? An inline function isn't going to help since to pass three arguments and check the result will end up as many lines as the code itself.

esphome/components/udp/udp_component.cpp
clydebarrow: good point.

esphome/components/udp/udp_component.cpp
clydebarrow: Also, `encode_value()` in helpers.h treats the data as big-endian, and curiously doesn't seem to be used anywhere in the codebase.

esphome/components/udp/udp_component.cpp
clydebarrow: I've just inlined the checks, which does allow more informative messaging.

nagyrobi:
Applied changes to all my nodes participating in the test, everything works OK.

esphome/components/udp/udp_component.cpp
jesserockz: Ah didnt think think about the pointer itself being incremented still.
Yeah it may not be used...haha
I mostly use (encode_uint32/16), and try to get PRs to use them.

clydebarrow: The naming of the helper functions is confusing (to me at least) - it says "encode" but i would regard that as _decoding_ from a serialised form back to an internal form. I've always preferred `put` and `get`. And the (unused) template form `encode_value()` is clever, but I I'd prefer not to rely on an inferred type in that situation. And then there's the endian-ness issue. 

Perhaps we need something like the [Java `ByteBuffer` class](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html) that could wrap an array and provide a standardised and tested but flexible interface for these operations.

nagyrobi:
Extended test to a total of 9 pcs 8266 and 6pcs ESP32 with various intercommunication scenarios, some encrypted, some not, all work fine.

Having two ESP32's getting 25 `pvvx` BLE sensors forwarding them via UDP to another ESP32 with update intervals of a few seconds each - pretty stressing - all work perfectly with this component.

jesserockz (changes requested):
In general this PR looks good, except for the fact that it is using/sending the `id` outside of the device which we always keep strictly internal. I think either this should use the `name` or another new field inside the udp config itself.

paravoid:
I would propose to add a protocol version to the header. The format may change for various unforeseen reasons (see e.g. BTHome v1->v2), a need for a different encryption schemes may arise, or a more advanced protocol (e.g. using protobuf) may be designed in the future. I see you already have a magic number in the header, so in theory you could always just issue a new magic value, but IMHO it'd be better to encode this in a separate field. Just my 2¢, and with gratitude to all of your work!

paravoid:
If I may offer another (unsolicited) suggestion: this is a novel (simple) layer 7 protocol that runs _on top_ of UDP. Perhaps it'd be more appropriate and less confusing to just give this protocol, and thus this component, a name? You wouldn't name a web server component as a "TCP" after all :)

(No I don't have any great ideas with regards to a name, and no, I personally won't argue for the color of this particular bike shed :smile:)

clydebarrow:
> this is a novel (simple) layer 7 protocol that runs _on top_ of UDP.

The platform hub `udp` component could well support other protocols, unlike the web server which is not implemented as a platform. I did have similar thoughts but a) decided it was fine, and b) couldn't think of a better name.

nielsnl68:
@clydebarrow The idea is very perfect for my needs although I could prefer using TCP or a streaming protocol Or like socket.io or something that can be used both ways.

Everything above could work the same way you implemented it. But just a different way to communicate.

nagyrobi:
@nielsnl68 could be as a separate component. Udp is very good as being stateless, easy on resources.

esphome/components/udp/udp_component.cpp
nielsnl68: I like the idea of having a ByteBuffer. i used it a lot in Javascript/nodejs. Very easy to manipulate strings and other binary data.

esphome/components/udp/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+        cv.Required(CONF_PROVIDER): cv.valid_name,
+        cv.GenerateID(CONF_UDP_ID): cv.use_id(UDPComponent),
+    }
+)
+
+
+def require_internal_with_name(config):
+    if CONF_NAME in config and CONF_INTERNAL not in config:
+        raise cv.Invalid("Must provide internal: config when using name:")
+    return config
```
nielsnl68, line 118: There must be a breaking reason why you have adding this? 
when not, is this not something the user should decide of it wants to send the state over API etc. or not?

nagyrobi: It's to avoid duplication of entities in HA, by default. Same behavior as the `homeassistant` platform sensors.

esphome/components/udp/udp_component.cpp
clydebarrow: > I like the idea of having a ByteBuffer. i used it a lot in Javascript/nodejs. Very easy to manipulate strings and other binary data.

https://github.com/esphome/esphome/pull/6878

esphome/components/udp/__init__.py
clydebarrow: > There must be a breaking reason why you have adding this? when not, is this not something the user should decide of it wants to send the state over API etc. or not?

Yes, the user always gets to decide - the only difference is that the user must be explicit about it, rather than the default behaviour where just providing a name implies the entity is exposed to HA. The alternative was to default to `internal: true` but that would silently change the default behaviour, which is not a Good Thing. This avoids applying the default behaviour but does so noisily.

LumenSoftNL: Ha i see, good point. You can close this then as resolved. :D

nielsnl68: Ha i see, good point. You can close this then as resolved. :D

nagyrobi:
I'm still using this PR widely and it's perfect. No issues of any kind.

clydebarrow:
> I think either this should use the `name` or another new field inside the udp config itself.

I have added a `broadcast_id` field to specify the id used in the broadcast. This allows the internal ids to be confined to the device if that's important to the user, but defaults to the same as the internal id to simplify configuration for a typical use-case.

esphome/components/udp/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    if sensors := config.get(CONF_SENSORS):
+        for sens_conf in sensors:
+            if isinstance(sens_conf, dict):
+                sens_id = sens_conf[CONF_ID]
+                sensor = await cg.get_variable(sens_id)
+                bcst_id = sens_conf.get(CONF_BROADCAST_ID) or sens_id.id
+            else:
+                sensor = await cg.get_variable(sens_conf)
+                bcst_id = sens_conf.id
+            cg.add(var.add_sensor(bcst_id, sensor))
```
jesserockz: I am pretty sure `maybe_simple_value` still puts the entire dict in the config object, its simply a validation to allow the user to only specify a value.
Can also simplify the if/for down to one line by providing a default value (`[]`) to `get`.

```suggestion
    for sens_conf in config.get(CONF_SENSORS, []):
        sens_id = sens_conf[CONF_ID]
        sensor = await cg.get_variable(sens_id)
        bcst_id = sens_conf.get(CONF_BROADCAST_ID, sens_id.id)
        cg.add(var.add_sensor(bcst_id, sensor))
```

clydebarrow: I thought that too, but it failed the first time I tried it. But now it works. I must have screwed something else up.

brotherdust:
> @clydebarrow The idea is very perfect for my needs although I could prefer using TCP or a streaming protocol Or like socket.io or something that can be used both ways.
> 
> 
> 
> Everything above could work the same way you implemented it. But just a different way to communicate.
> 
> 

Network engineer here.
I respectfully disagree. The purpose of this protocol is to **broadcast** the state of something. TCP, by definition, cannot do that. It can only work in point to point use-cases.

That being said, @clydebarrow , have you considered ramifications of UDP broadcasting in ESPHome's use-case?

If not, I humbly ask you to consider the following concerns:

1. Most ESPHome devices are on WiFi. By definition, all broadcasts are sent to all devices in the same broadcast domain. All of these devices **must** process those broadcasts, which means additional processing load for the devices, leading to higher power consumption.

2. Even if the devices are not in the same broadcast domain, broadcasts will cause additional airtime on the AP.

3. Broadcasts in WiFi aren't actually broadcasts. Individual devices may be connected at different data rates. The AP has to deliver the broadcast to every device at the rate they are connected at, meaning it's never going to happen all at once. Yet more airtime.

4. There doesn't appear to be any rate limiting in your implementation, so there's no upper bound on the amount of broadcast packets may be generated by a device. In worst-case scenario, a single device could effectively cause a denial of service. Some enterprise systems have knobs to prevent broadcast storms, but that cannot be relied upon.

So, long story short, it will cause more power consumption, chew up airtime on the WiFi, and possibly soaking up so much of it as to cause a denial of service.

I urge you to consider **not** allowing broadcast and encourage the use of multicast. At the very least, do not set broadcast as default.

Multicast is just as simple to configure. For example, one could simply specify the multicast address of a group to join. They would then be subscribed to all the information being sent to that group. Let the network sort out the rest. In most cases, modern WiFi converts multicast to unicast anyway. This limits the scope of the devices that would be receiving the data, so less processing and less airtime.

Anyway, I appreciate your work here. Just wanted to give you some ideas to think about.

nagyrobi:
I have intensively tested Tasmota's [Device Groups](https://tasmota.github.io/docs/Device-Groups/) feature which relies on multicast. Unfortunately, it turned to be very unreliable at some locations, likely because of poor implementation (either in Tasmota or in the WiFi APs). After rebooting, communication was OK for a while (sometimes hours, sometimes days), and then the target nodes simply didn't respond to commands.

Couldn't tell if it was due to poor network (which probably forgot client subscriptions) or the receiving nodes (which probably didn't tkink of renewing them).
I don't say multicast is bad, I just fear that it's gonna generate a lot of support demand on ESPHome side due to likely poor multicast implementation in SOHO network hardware. I don't believe that _modern WiFi converts multicast to unicast anyway_, imho many brands impement some sort of multicast handling because there's a need on consumer side, but they only test it in a few limited use cases (typically 3play IPTV scenarios) and real settings are not even available for the users to adjust in the equipment.

I use this component by unicasting the data to specific IPs, specifically for the reasons mentioned. Maybe it would be enough to just focus the documentation on that.

clydebarrow:
> Network engineer here.

Thanks for the input. I'll digest it more carefully in due course, but just a couple of quick comments:

Rate limiting - not specifically configured, but it is inherent in the ESPHome `loop()` and `update()` model, so the amount of traffic one node will generate is limited. It may be useful to have an explicit mechanism for rate limiting as well.

Currently the default destination address is the local broadcast address - perhaps the address should be required to be specifically configured, with the docs recommending unicast over broadcast.

Thanks again for the input and I'll give it more thought.

brotherdust:
@nagyrobi ,
Regarding my assertion that most modern WiFi systems just convert unicast to multicast, I should have qualified that more; so I understand your skepticism. Allow me to clarify:

Most _controller-based_ WiFi systems have a feature that converts multicast to unicast. On Ubiquiti UniFi, I believe they call it a multicast helper. It basically acts as a multicast proxy and can be located on the controller or the AP.

There are several reasons for this:

1. True wireless multicast requires that all the subscribing devices and participating AP's lock to the same data rate while the data is transmitted downstream. In order to ensure that as many devices as possible can receive without error, the multicast rate tends to be much lower than the average unicast rate. As a result, the overall throughput of the network will be reduced proportional to the amount multicast data that must be transmitted.

2. There's no guarantee that a given network will correctly support multicast; but unicast is pretty much guaranteed.


I agree with you that there footguns to be aware of. My suggestion to reduce the risk of them would be as follows:

1. Require the address(es) to be explicitly configured and have no default ( @clydebarrow mentioned this already )

2. Support unicast, multicast, and broadcast scenarios; but explicitly call out that unicast is preferred and the other options are left as an exercise for an experienced user and that they should have no expectation it will work reliably.

tomaszduda23:
Great work. I was looking to communicate two esphome over uart. Would you consider splitting it into two components so udp could be replaced by other protocols in the future?

clydebarrow:
> Great work. I was looking to communicate two esphome over uart. Would you consider splitting it into two components so udp could be replaced by other protocols in the future?

Yes, already been giving that some thought,

esphome/components/udp/udp_component.cpp
```diff
@@ -0,0 +1,616 @@
...
+      this->updated_ = true;
+      sensor.updated = true;
+    });
+  }
+#endif
+  if (strlen(this->name_) > 255) {
+    this->mark_failed();
+    this->status_set_error("Device name exceeds 255 chars");
+    return;
+  }
```
jesserockz: Probably makes more sense for this to go as the 2nd line of `setup` as there is no point setting up preferences and generating ping key and setting up callbacks etc.

esphome/components/udp/udp_component.cpp
```diff
@@ -0,0 +1,616 @@
...
+#endif
+  this->flush_();
+  this->updated_ = false;
+  this->resend_data_ = false;
+}
+
+void UDPComponent::update() {
+  this->updated_ = true;
+  this->resend_data_ = this->should_send_;
```
jesserockz, line 359: Just so I am understanding this correctly, every update interval, **all** data will be sent?

esphome/components/udp/sensor.py
```diff
@@ -0,0 +1,27 @@
...
+
+from . import (
+    CONF_PROVIDER,
+    CONF_REMOTE_ID,
+    CONF_UDP_ID,
+    SENSOR_SCHEMA,
+    require_internal_with_name,
+)
+
+AUTO_LOAD = ["udp"]
```
jesserockz: A sensor requires a provider to be set in the main component right? In which case this should be a dependency, not auto loaded
```suggestion
DEPENDENCIES = ["udp"]
```

esphome/components/udp/binary_sensor.py
```diff
@@ -0,0 +1,27 @@
...
+
+from . import (
+    CONF_PROVIDER,
+    CONF_REMOTE_ID,
+    CONF_UDP_ID,
+    SENSOR_SCHEMA,
+    require_internal_with_name,
+)
+
+AUTO_LOAD = ["udp"]
```
jesserockz: Same here?
```suggestion
DEPENDENCIES = ["udp"]
```

esphome/components/udp/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+        sens_id = sens_conf[CONF_ID]
+        sensor = await cg.get_variable(sens_id)
+        bcst_id = sens_conf.get(CONF_BROADCAST_ID, sens_id.id)
+        cg.add(var.add_sensor(bcst_id, sensor))
+    for sens_conf in config.get(CONF_BINARY_SENSORS, ()):
+        sens_id = sens_conf[CONF_ID]
+        sensor = await cg.get_variable(sens_id)
+        bcst_id = sens_conf.get(CONF_BROADCAST_ID, sens_id.id)
+        cg.add(var.add_binary_sensor(bcst_id, sensor))
+    for address in config.get(CONF_ADDRESSES, ("255.255.255.255",)):
```
jesserockz: Defaults should go in the schema itself
```suggestion
    for address in config.get(CONF_ADDRESSES):
```

esphome/components/udp/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+    return config
+
+
+CONFIG_SCHEMA = cv.All(
+    cv.polling_component_schema("15s")
+    .extend(
+        {
+            cv.GenerateID(): cv.declare_id(UDPComponent),
+            cv.Optional(CONF_PORT, default=18511): cv.port,
+            cv.Optional(CONF_ADDRESSES): cv.ensure_list(cv.ipv4),
```
jesserockz:
```suggestion
            cv.Optional(CONF_ADDRESSES, default=["255.255.255.255"]): cv.ensure_list(cv.ipv4),
```

jesserockz (changes requested):

esphome/components/udp/udp_component.cpp
clydebarrow: Correct. UDP is unreliable, so an update pushed when the sensor changes may be lost. The interval is configurable.

jesserockz (approved):

PR #6866, opened 2024-06-06
[e131] Channel offset

jesserockz:

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/e131/e131_addressable_light_effect.cpp
+++ b/esphome/components/e131/e131_addressable_light_effect.cpp
@@ -55,7 +55,7 @@ bool E131AddressableLightEffect::process_(int universe, const E131Packet &packet
   // limit amount of lights per universe and received
   int output_end =
       std::min(it->size(), std::min(output_offset + get_lights_per_universe(), output_offset + packet.count - 1));
-  auto *input_data = packet.values + 1;
+  auto *input_data = packet.values + 1 + (this->channel_offset_ * this->channels_);
 
   ESP_LOGV(TAG, "Applying data for '%s' on %d universe, for %" PRId32 "-%d.", get_name().c_str(), universe,
            output_offset, output_end);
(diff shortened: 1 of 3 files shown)


esphome/components/e131/e131_addressable_light_effect.cpp
```diff
@@ -55,7 +55,7 @@ bool E131AddressableLightEffect::process_(int universe, const E131Packet &packet
   // limit amount of lights per universe and received
   int output_end =
```
ajvpot, line 56: There  may be a buffer overread here because the channel offset is not taken into account for calculating how many channels to read.

esphome/components/e131/e131_addressable_light_effect.cpp
```diff
@@ -55,7 +55,7 @@ bool E131AddressableLightEffect::process_(int universe, const E131Packet &packet
...
       std::min(it->size(), std::min(output_offset + get_lights_per_universe(), output_offset + packet.count - 1));
-  auto *input_data = packet.values + 1;
+  auto *input_data = packet.values + 1 + (this->channel_offset_ * this->channels_);
```
ajvpot, line 58: It may not make sense to multiply the offset by the channel count because in an e.131 system with multiple types of devices, it is likely there will be devices with different channel counts. The starting address is typically defined in an absolute number of channels and may not be a multiple of the channel count of the device.

For my use case, I'm using a MONO device so it's a moot point.

ajvpot:
Thank you so much for your help here! I will test and possibly push some additional changes.

esphome/components/e131/e131_addressable_light_effect.cpp
jesserockz: I did wonder about this. 
I have no pull to either direction here.

PR #6869, opened 2024-06-07
[he60r] Don't publish state unless it has changed. [BUGFIX]

clydebarrow:

State was being published even when it had not changed, which e.g. resulted in spurious triggering of `on_closed`.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/he60r/he60r.h
+++ b/esphome/components/he60r/he60r.h
@@ -25,15 +25,14 @@ class HE60rCover : public cover::Cover, public Component, public uart::UARTDevic
   void control(const cover::CoverCall &call) override;
   bool is_at_target_() const;
   void start_direction_(cover::CoverOperation dir);
-  void update_operation_(cover::CoverOperation dir);
   void endstop_reached_(cover::CoverOperation operation);
   void recompute_position_();
   void set_current_operation_(cover::CoverOperation operation);
   void process_rx_(uint8_t data);
 
-  uint32_t open_duration_{0};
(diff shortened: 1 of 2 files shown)


esphome/components/he60r/he60r.h
```diff
@@ -25,15 +25,14 @@ class HE60rCover : public cover::Cover, public Component, public uart::UARTDevic
...
   void recompute_position_();
   void set_current_operation_(cover::CoverOperation operation);
   void process_rx_(uint8_t data);
 
-  uint32_t open_duration_{0};
-  uint32_t close_duration_{0};
-  uint32_t toggles_needed_{0};
+  unsigned open_duration_{0};
+  unsigned close_duration_{0};
+  unsigned toggles_needed_{0};
```
jesserockz, line 35: Why change these? `uint32_t` is `unsigned int` already

clydebarrow: Because it allows getting rid of the ugly and unreadable PRiu32 printf format. Not material to the fix, but since I was in there... These values will only ever be small so the purely theoretical possibility that `unsigned` is only 16 bits isn't an issue.

jesserockz (approved):

PR #6871, opened 2024-06-07
Sntp for ESP IDF

### LINKED ISSUE #3123: NTP only uses the first server
### The problem

Adding more than one NTP server is futile. Esphome only uses the first.
```
root@GRAPHRT:~# tcpdump -i br-lan port 123
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on br-lan, link-type EN10MB (Ethernet), capture size 262144 bytes
08:52:06.645636 IP GRAPHT2.lan.57862 > GRAPHRT.123: NTPv4, Client, length 48
08:52:19.647165 IP GRAPHT2.lan.65233 > GRAPHRT.123: NTPv4, Client, length 48
08:52:19.665991 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:52:25.669558 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:52:34.672243 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:52:49.673318 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:53:16.676807 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:53:43.680915 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:54:10.685244 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:54:37.687337 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:55:04.691656 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:55:31.696907 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:55:58.699294 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:56:25.703301 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:56:52.706770 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:57:19.711066 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:57:46.714242 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:58:13.719742 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:58:40.724123 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:59:07.725250 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
08:59:34.728671 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
09:00:01.735174 IP GRAPHT2.lan.65234 > GRAPHRT.123: NTPv4, Client, length 48
```
See yaml example:
"Synchronized system clock" never appears.
"Time? / Time is not set." never appears. on_time program never runs. on_boot, if uncommented, also never runs.

2022.3.0-dev

pip

_No response_

ESP32

### Board

M5StickC

NTP

### Example YAML snippet

```yaml
esphome:
  name: bug-test
  platform: ESP32
  board: m5stick-c
#  on_boot:
#    priority: -100
#    then:
#    - logger.log: "Boot script waiting for time..."
#    - wait_until:
#        time.has_time:
#    - logger.log: "Boot script got time!"

wifi:
  ssid: <my_ssid>
  password: <my_pass>
  manual_ip:
    static_ip: 172.27.143.48
    gateway: 172.27.143.1
    subnet: 255.255.255.0
    dns1: 172.27.143.1

time:
  - platform: sntp
    timezone: "Europe/Bucharest"
    servers:
      - 172.27.143.1
      - pool.ntp.org
    on_time_sync:
      then:
      - logger.log: "Synchronized system clock"
    on_time:
      - seconds: /10
        then:
          - logger.log: "Time?"
          - if:
              condition:
                - time.has_time:
              then:
                - logger.log: "Time is valid."
              else:
                - logger.log: "Time not set."
```

Nope.

### Additional information

_No response_

asergunov:

SNTP component for esp-idf builds.

It redirects `set_update_interval`/`get_update_interval` calls directly to ESP IDF API to have only one source of truth. So it doesn't matter how the user change it. This means ESPHome shouldn't call `update` functions because periodical updates are handled by ESP IDF itself. In other the hand the user can trigger NTP update manually by calling `update` function by himself.

The time sync tracked by calling `sntp_get_sync_status()` which returns  `SNTP_SYNC_STATUS_COMPLETED` only once after sync actually happened. So we can clearly see the event in `loop()` and don't need any callbacks.

User can change servers any time. Code takes care about use-after-free issue

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/3123

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esp32:
  board: esp32-c3-devkitm-1
  framework:
    type: esp-idf
    sdkconfig_options:
      CONFIG_LWIP_SNTP_MAX_SERVERS: '3' # Let it use all three servers

# wifi here

time:
  - platform: sntp
    id: sntp_time
    timezone: ${timezone}
    update_interval: 5min
    servers: 
      - 0.pool.ntp.org
      - 1.pool.ntp.org
      - 2.pool.ntp.org
    on_time_sync:
      then:
      - logger.log: "Synchronized"

button:
  - platform: template
    name: Sync now
    entity_category: config
    on_press:
      then:
        - component.update: sntp_time
  - platform: template
    name: Set new servers
    entity_category: config
    on_press:
      then:
        - lambda: id(sntp_time).set_servers({"3.pool.ntp.org", "4.pool.ntp.org", "5.pool.ntp.org"});

number:
  - platform: template
    name: SNTP Sync Interval (s)
    min_value: 1 # ESP-EDF supports minimum 15s interval so result will be 15s if less set
    max_value: 18000
    step: 1    lambda: |
      return id(sntp_time).get_update_interval()/1000;
    set_action: 
      then:
        - lambda: |
            id(sntp_time).set_update_interval(x*1000);
```

## Checklist:


asergunov:
Could someone please help me with check failures. Code looks good for me. Am I missing something?

esphome/components/sntp/sntp_component.h
```diff
@@ -17,21 +17,28 @@ class SNTPComponent : public time::RealTimeClock {
...
-  std::string server_2_;
-  std::string server_3_;
+  void setup_servers_();
+
+ private:
+  // Private because buffer address should stay unchanged
+  std::string servers_[3];
+
+ protected:
+#if defined(USE_ESP_IDF)
```
asergunov:
```suggestion
#if !defined(USE_ESP_IDF)
```

esphome/components/sntp/sntp_component.cpp
```diff
@@ -31,48 +28,103 @@ void SNTPComponent::setup() {
...
 void SNTPComponent::update() {
-#if !defined(USE_ESP_IDF)
   // force resync
-  if (sntp_enabled()) {
+  if (esp_sntp_enabled()) {
+#if defined(USE_ESP_IDF)
+    ESP_LOGD(TAG, "Force resync");
+    sntp_restart();
+#else
     sntp_stop();
```
asergunov, line 83:
```suggestion
#if defined(USE_ESP_IDF)
  if (esp_sntp_enabled()) {
    ESP_LOGD(TAG, "Force resync");
    sntp_restart();
#else
  if (sntp_enabled()) {
    sntp_stop();
```

CarlosGS (approved):
I confirm this PR works well and fixes NTP for the esp-idf platform.
Thank you very much @asergunov!

CarlosGS (changes requested):
I've found a small problem where SNTP was being restarted every update interval.
Here are the suggested changes, open for comment as there are many ways to do this.
https://github.com/asergunov/esphome_esphome/pull/1

esphome/components/sntp/sntp_component.cpp
```diff
@@ -31,48 +28,104 @@ void SNTPComponent::setup() {
...
+  this->servers_was_setup_ = true;
+
+  for (uint8_t i = 0; i < 3; ++i) {
+    const auto &buff = server_name_buffer(servers_[i]);
+    if (buff != nullptr && buff != sntp_getservername(i)) {
+      ESP_LOGCONFIG(TAG, "Can't set server %d", i + 1);
+    }
   }
 #ifdef USE_ESP_IDF
+  this->stop_poller();
```
asergunov: @CarlosGS here I stop any automatic update calls

asergunov:
> I've found a small problem where SNTP was being restarted every update interval.
> 
> Here are the suggested changes, open for comment as there are many ways to do this.
> 
> https://github.com/asergunov/esphome_esphome/pull/1

I disabling calling update function periodically using 'stop_puller()'. But user can call it to manually trigger update. 

@CarlosGS please let me know if I got your request right. 

EDIT: Got it. You mean `stop_puller` doesn't have any effect. Any Idea how to fix it without changing the rest of the logic? Added some comments in code BTW.

esphome/components/sntp/sntp_component.cpp
asergunov: Do you have an idea why this doesn't work?

esphome/components/sntp/sntp_component.cpp
```diff
@@ -31,48 +28,104 @@ void SNTPComponent::setup() {
...
 
+  this->has_time_ = true;
+#endif  // defined(USE_ESP_IDF)
   ESP_LOGD(TAG, "Synchronized time: %04d-%02d-%02d %02d:%02d:%02d", time.year, time.month, time.day_of_month, time.hour,
            time.minute, time.second);
   this->time_sync_callback_.call();
-  this->has_time_ = true;
+}
+#if defined(USE_ESP_IDF)
+void SNTPComponent::set_update_interval(uint32_t update_interval) {
```
asergunov, line 119:
```suggestion
/**
Sets the update interval calling `sntp_set_sync_interval()`.

Calls `sntp_restart()` if update interval is less then it was before as ESP IDF according documentation. Without this cal interval will be updated only after previously scheduled update.
*/
void SNTPComponent::set_update_interval(uint32_t update_interval) {
```

esphome/components/sntp/sntp_component.cpp
```diff
@@ -31,48 +28,104 @@ void SNTPComponent::setup() {
...
+
+  sntp_set_sync_interval(update_interval);
+  const auto new_sync_interval = sntp_get_sync_interval();
+  time::RealTimeClock::set_update_interval(new_sync_interval);
+
+  if (previous_sync_interval > new_sync_interval) {
+    sntp_restart();
+  }
+}
+uint32_t SNTPComponent::get_update_interval() const { return sntp_get_sync_interval(); }
```
asergunov, line 133:
```suggestion
/**
Redirects the call to `sntp_get_sync_interval()` so it's only one source of truth.
*/
uint32_t SNTPComponent::get_update_interval() const { return sntp_get_sync_interval(); }
```

esphome/components/sntp/sntp_component.cpp
asergunov:
```suggestion
  // Stop the puller to prevent periodically update calls. IDF Handles periodic updates by itself.
  // The user still able to call `update` to trigger manual update.
  this->stop_poller();
```

asergunov (commented):
Adding comments explaining update interval logic

asergunov:
@CarlosGS I've deferred the `stop_poller`call. Could you please retest if it helps?

esphome/components/sntp/sntp_component.cpp
asergunov: I've deferred this call.

CarlosGS:
> The purpose to override get_update_interval is to report correct value in case user changed it directly with esp idf API. So there is an only one source of truth in the system.

Ah yes it makes sense. Better this way :+1:

> I've deferred the stop_pollercall. Could you please retest if it helps?

Nice, it just makes it run on the next loop, much cleaner than my approach.
And yep, now it works correctly in the background without restarting the ntp library :)

I hope this gets merged soon, thanks a lot!

CarlosGS (approved):

asergunov:
Keep open

swoboda1337:
> Keep open

There are conflicts that need to be resolved in order for this to be reviewed.

swoboda1337 (changes requested):
Need to fix conflicts

asergunov:
@CarlosGS Hi. I've resolved conflicts. Could you give it a test?

CarlosGS (approved):
All looks good and it seems to be working well.
Thanks for revisiting this! I hope it can be merged soon.
Cheers!

asergunov:
> All looks good and it seems to be working well.
> 
> Thanks for revisiting this! I hope it can be merged soon.
> 
> Cheers!

What board did you test? Esp32 or esp8266?

CarlosGS:
ESP32 C3 with IDF
Screenshot_20251026_112735

asergunov:
We need to check both esp32 and esp8266 in both Arduino and idf builds:
- changing sync period 
- manual sync
- callbacks in both sync cases

I have everything for that. Will try to check by tomorrow.

cypherbits:
And I was wondering why sntp wasn't working on esp-idf, now the recommended and default framework, lost almost 3 hours trying things, but no error messages. And just found this, that sntp is not working on esp-idf since years...

asergunov:
I have no time to test it after merging with latest dev branch and conflicts resolution. That's the only reason it's in draft. Could someone give it a test on different platforms?

The points to test:

- [ ] on_time_sync triggering (just put log message inside)
- [ ] Manual sync triggering with update function
- [ ] Changing the update interval
- [ ] Changing the server list runtime (not critical)

asergunov:
I've tested on esp32idf - all works

esphome/components/sntp/sntp_component.cpp
```diff
@@ -61,46 +60,92 @@ void SNTPComponent::dump_config() {
...
   if (sntp_enabled()) {
     sntp_stop();
     this->has_time_ = false;
     sntp_init();
-  }
 #endif
+  }
 }
 void SNTPComponent::loop() {
-// The loop is used to infer whether we have valid time on platforms where we
```
bdraco, line 78: why not sntp_set_time_sync_notification_cb on esp idf instead of loop ?

esphome/components/sntp/sntp_component.h
```diff
@@ -22,24 +22,28 @@ class SNTPComponent : public time::RealTimeClock {
 
   void setup() override;
   void dump_config() override;
+
+  void set_servers(const std::array<const char *, SNTP_SERVER_COUNT> &servers);
```
bdraco, line 26: I'm not sure we want to support setting servers at runtime as it increases the complexity here.

This also gets very confusing since `SNTP_SERVER_COUNT` is defined from python codegen and can be 1, 2, or 3 based on what is in yaml

esphome/components/sntp/sntp_component.cpp
```diff
@@ -14,44 +14,43 @@ namespace sntp {
...
+  // We can't call `stop_poller` here because it will be started after `setup` call. So defer the execution.
+  this->defer([this] {
+    // Stop the puller to prevent periodically update calls. IDF Handles periodic updates by itself.
+    // The user still able to call `update` to trigger manual update.
+    this->stop_poller();
   });
-#endif
-
+  esp_sntp_set_sync_interval(time::RealTimeClock::get_update_interval());
+  esp_sntp_init();
```
bdraco, line 52: esp_sntp_init is also called on line 24

esphome/components/sntp/sntp_component.cpp
```diff
@@ -61,46 +60,92 @@ void SNTPComponent::dump_config() {
...
-// cannot tell whether SNTP has succeeded.
-// One limitation of this approach is that we cannot tell if it was the SNTP
-// component that set the time.
-// ESP-IDF and ESP8266 use callbacks from the SNTP task to trigger the
-// `on_time_sync` trigger on successful sync events.
-#if defined(USE_ESP32) || defined(USE_ESP8266)
-  this->disable_loop();
-#endif
+#ifdef USE_ESP32
+  // Tah sntp_get_sync_status call returns SNTP_SYNC_STATUS_COMPLETED only once when time just synched. So we don't need
```
bdraco, line 91:
```suggestion
  // The sntp_get_sync_status call returns SNTP_SYNC_STATUS_COMPLETED only once when time just synched. So we don't need
```

esphome/components/sntp/sntp_component.cpp
```diff
@@ -61,46 +60,92 @@ void SNTPComponent::dump_config() {
...
+  // any callbacks
```
bdraco, line 92: The code is mixing `esp_sntp_*` vs `sntp_*` APIs.

 https://github.com/espressif/esp-idf/blob/master/components/lwip/include/apps/esp_sntp.h
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/lwip.html
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/system_time.html#system-time-sntp-sync

https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/system_time.html#sntp-time-synchronization

> Some lwIP APIs, including SNTP functions, are not thread safe, so it is recommended to use [esp_netif component](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_netif.html) when interacting with SNTP module.

Seems like the `esp_` ones are preferred

esphome/components/sntp/sntp_component.cpp
```diff
@@ -14,44 +14,43 @@ namespace sntp {
...
   }
-
-#if defined(USE_ESP8266)
-  settimeofday_cb([this](bool from_sntp) {
-    if (from_sntp)
-      this->time_synced();
+#ifdef USE_ESP32
+  // We can't call `stop_poller` here because it will be started after `setup` call. So defer the execution.
+  this->defer([this] {
+    // Stop the puller to prevent periodically update calls. IDF Handles periodic updates by itself.
```
bdraco, line 47:
```suggestion
    // Stop the poller to prevent periodically update calls. IDF Handles periodic updates by itself.
```

esphome/components/sntp/sntp_component.cpp
```diff
@@ -14,44 +14,43 @@ namespace sntp {
...
+    // The user still able to call `update` to trigger manual update.
+    this->stop_poller();
   });
-#endif
-
+  esp_sntp_set_sync_interval(time::RealTimeClock::get_update_interval());
```
bdraco, line 51: line 23 uses `this->get_update_interval()` ... this is a bit confusing as to why one over the other. Each spot needs comments explaing the choice

esphome/components/sntp/sntp_component.cpp
```diff
@@ -61,46 +60,92 @@ void SNTPComponent::dump_config() {
...
+Redirects the call to `sntp_get_sync_interval()` so it's only one source of truth.
+*/
+uint32_t SNTPComponent::get_update_interval() const { return sntp_get_sync_interval(); }
+#endif  // USE_ESP32
+void SNTPComponent::setup_servers_() {
+  for (size_t i = 0; i < this->servers_.size(); ++i) {
+    const auto &buff = this->servers_[i];
+#ifdef USE_ESP32
+    esp_sntp_setservername(i, buff);
+    if (buff != nullptr && buff != esp_sntp_getservername(i)) {
```
bdraco, line 140: pointer comparison may not be correct here. This might need to be `strcmp` like https://github.com/esphome/esphome/pull/6871/files#diff-aef5cc5434185d4f591b9df6f5be5ecb32ad1e089d3d52be9b2434885fe3888fR34

esphome/components/sntp/sntp_component.cpp
```diff
@@ -61,46 +60,92 @@ void SNTPComponent::dump_config() {
...
+#else
+    sntp_setservername(i, buff);
+    if (buff != nullptr && buff != sntp_getservername(i)) {
```
bdraco, line 143: pointer comparison may not be correct here. This might need to be `strcmp` like https://github.com/esphome/esphome/pull/6871/files#diff-aef5cc5434185d4f591b9df6f5be5ecb32ad1e089d3d52be9b2434885fe3888fR34

esphome/components/sntp/sntp_component.cpp
```diff
@@ -14,44 +14,43 @@ namespace sntp {
...
-#if defined(USE_ESP8266)
-  settimeofday_cb([this](bool from_sntp) {
-    if (from_sntp)
-      this->time_synced();
+#ifdef USE_ESP32
+  // We can't call `stop_poller` here because it will be started after `setup` call. So defer the execution.
+  this->defer([this] {
+    // Stop the puller to prevent periodically update calls. IDF Handles periodic updates by itself.
+    // The user still able to call `update` to trigger manual update.
+    this->stop_poller();
```
bdraco, line 49: It would be cleaner to prevent it from ever starting than using a defer to stop it.


Maybe override `call_setup` and prevent it from starting on esp32

bdraco (changes requested):
Please see review comments above. Thanks!

PR #6885, opened 2024-06-10
[core] Update Entities

jesserockz:

Update Entities!

Following on from the great work in #5586, this utilises that system to provide a native `update` entity for a device. This is done by pointing to a `manifest.json` (sample below) which will be downloaded periodically during runtime (defaults to every 6 hours) and when triggered, the device will download the binary and flash itself.

Related PRs:
- https://github.com/esphome/aioesphomeapi/pull/882
- https://github.com/home-assistant/core/pull/119339
- https://github.com/home-assistant/core/pull/119348

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3935

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
update:
  - platform: http_request
    id: ota_update
    name: "OTA Update"
    source: https://example.com/manifest.json
```

```json
{
  "name": "Awesome Made for ESPHome Device",
  "version": "10",
  "home_assistant_domain": "esphome",
  "builds": [
    {
      "chipFamily": "ESP32",
      "parts": [
        {
          "path": "/firmware-32-idf.factory.bin",
          "offset": 0
        }
      ],
      "ota": {
        "summary": "Here is the changelog:\n\n- Update the thing",
        "release_url": "http://example.com/releases/10",
        "path": "/firmware-32-idf.ota.bin",
        "md5": "1234567890abcdef1234567890abcdef"
      }
    }
  ]
}

```

## Checklist:


oddstr13:
I think md5 is sufficient for protecting against accidental corruption when loading firmware over your LAN, but is it sufficient for internet use, where the manifest can live on a different server than that of the binary files?
(assumption being the manifest is safe, but the binaries them self could be modified by a third party)

I think it would be worth considering whether it is feasible to use cryptographic signatures to verify the origin of OTA updates, and that would likely require stronger checksums than md5 to be useful.

Scenario:
Device is flashed with a firmware including an ed25519 public key for update verification.
The manifest additionally contains a signed version of the checksum.
The device could then verify the integrity and origin of the manifest, and thus confirm the authenticity of the firmware blob once the previously verified checksum matches the data.

This would additionally trivially allow for local caching of updates by using http for the firmware downloads them selves (ala apt-cacher-ng / steamcache). Not that the firmware files are that large, but the potential for many devices doing the update at the same time could cause a decent amount of bandwith usage.

jesserockz:
@oddstr13 Very good point there. We wanted to start with something simple, and we can always extend it later to allow other types of verification. 

One major issue with signed firmware is that you run the risk of soft bricking a device forcing it to be serial flashed.

kbx81 (approved):

oddstr13:
> One major issue with signed firmware is that you run the risk of soft bricking a device forcing it to be serial flashed.

Yes, that is a risk, but the main use-case for it would be huge deployments, where you would probably do staged roll-out, and verify & test updates before deploying them. (Think the scale of IKEA Trådfri or Phillips Hue).
Centrally managed appliance devices.

It would be possible implementing multiple OTA methods, such as allowing the current push-based OTA for local flashing, allowing users to roll their own firmware instead of the signed upstream releases using OTA via polling a http repo.
Additionally, it would be possible to disable verification via a property.

It all depends on how it is set up.

Personally, I'm looking forward to being able to update my deep-sleep nodes, and reducing power usage even more by not having a delay before going back to sleep (to have a *slight* chance of managing to push OTA to them). :rocket:

tomatensaus:
If you have a periodic check eg. 6 hours and you happen to accidentally add a bad firmware. You could end up with all your devices bricked needing to reflash them via USB. The risk of this would mean that I would never use this feature.
Add a page with devices requesting update, and then manually allow/enable the update per device. Then at least you can test a device before rolling out more updates.

Another idea: When you get a bad update, and you end up in safe mode, allow the device to install a factory default via OTA. This could replace the factory default app that takes up valuable space on the flash(I do not use this as I need every KB available). Even if this factory default is just a page that allows the firmware to be manually loaded via file upload, nobody wants to re-flash them via USB because they are mostly installed in hard to reach places

jesserockz:
@tomatensaus There is no automatic updating involved here. It purely creates an entity to tell you there is an update and you have to click a button to start the update. If the user implements automatic triggering of the update, that is on the user.

oddstr13:
Ah, this is for the Home Assistant update notification under settings? I missed that detail.

Is it Home Assistant or the ESPHome device that polls for update availability?
Seems like something that would be best offloaded to Home Assistant, at least for battery operated devices.


My comments above where more for the overall question about md5 which was mentioned in the locked PR #5586, and not specific to this particular set of changes :sweat_smile:

kbx81:
> Is it Home Assistant or the ESPHome device that polls for update availability?

@oddstr13 the ESPHome device makes an HTTP request for a manifest file and uses the content of that file to determine if an update is available (that is, the version in the file is different than that already installed). So, for battery-powered devices, I think this should be advantageous because they can wake up, immediately make and process the request and go right back to sleep if there's no update. 😄

EDIT: To add to that, if it determines that there's an update, this state is immediately published to HA so you'll see it there, too.

nielsnl68:
Hi @jesserockz, @kbx81, @oddstr13 and @tomatensaus,
This is a very interesting conversation here and on Discord.
Would it be possible to move this conversation to discord: 
https://discord.com/channels/429907082951524364/1249857290484318340
I added a proposal there to make this maybe a bit safer?

nielsnl68:
> @oddstr13 Very good point there. We wanted to start with something simple, and we can always extend it later to allow other types of verification.
> 
> One major issue with signed firmware is that you run the risk of soft bricking a device forcing it to be serial flashed.

In many cases this is a good idea. But IMHO with this solution we should have trough all worse case scenarios before this would make it to a PR like this. 

The idea is good. As said above there are to many ways to make it harmful that we should talk it over for some time. And Check other solutions before we allow this PR to see the light.

PR #6899, opened 2024-06-12
add cert-pem configurable

glmnet:

add option to set the cert_pem for http request, in that case the full certs are not required, hence I added a weird condition on the python side.

I guess this component should be MULTI_CONF to support different servers properly.

Let me know if you want me to add more validation python side, or any other requirements, etc. then I'll do docs.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/http_request/__init__.py
+++ b/esphome/components/http_request/__init__.py
@@ -37,6 +37,7 @@ CONF_HTTP_REQUEST_ID = "http_request_id"
 
 CONF_USERAGENT = "useragent"
 CONF_VERIFY_SSL = "verify_ssl"
+CONF_CERTIFICATE_PEM = "certificate_pem"
 CONF_FOLLOW_REDIRECTS = "follow_redirects"
 CONF_REDIRECT_LIMIT = "redirect_limit"
 CONF_WATCHDOG_TIMEOUT = "watchdog_timeout"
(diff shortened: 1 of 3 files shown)


jesserockz (commented):
I like it, I don't like that it is idf only. I will look into whether we can attach the cert for 8266, 32-arduino and rp2040.

esphome/components/http_request/__init__.py
```diff
@@ -105,6 +106,7 @@ def _declare_request_class(value):
                 cv.only_on_esp8266, cv.boolean
             ),
             cv.Optional(CONF_VERIFY_SSL, default=True): cv.boolean,
+            cv.Optional(CONF_CERTIFICATE_PEM): cv.string,
```
guillempages, line 109: I don't know if there is an easy python way for that, but it might be a good idea to validate that the supplied string is a valid certificate here, with an extra validator. Otherwise we risk crashes at runtime if someone puts e.g. a file name here or the copy&paste of the certificate is not complete.
Or am I mistaken here?

jesserockz: Yes this is a good point, although we dont validate the mqtt certificates. (probably should do it)

PR #6905, opened 2024-06-13
[nextion] Optionally skip connection handshake

### LINKED ISSUE #2872: Nextion Editor Simulator Connection
### The problem

I try to connect to the Nextion Editor Simulator ("Debug") feature. This is to avoid having to deal with the physical screen (to avoid any damage) and because is just faster to do it this way.

ESPHome Logs shows the following:

```
[12:10:29][W][nextion:072]: Nextion is not connected! 
[12:10:29][E][uart:015]: Reading from UART timed out at byte 0!
```

And Nextion Editor 

image

Connecting to a real screen with the exact same tft file works OK.

ESPHome logs show the following

```
[12:14:14][C][nextion:117]: Nextion:
[12:14:14][C][nextion:118]:   Device Model:     NX4832K035_011R
[12:14:14][C][nextion:119]:   Firmware Version: 163
[12:14:14][C][nextion:120]:   Serial Number:    ---------------------
[12:14:14][C][nextion:121]:   Flash Size:       33554432
[12:14:14][C][nextion:122]:   Wake On Touch:    True

```

I guess ESPHome tries to get this info at startup and the simulator does not provide it...

Is there a way to override this?

2021.10.3

Docker

_No response_

ESP8266

### Board

Wemos D1 Mini

Nextion

### Example YAML snippet

```yaml
esphome:
  name: esphomethermo-v0
  platform: ESP8266
  board: d1_mini_lite

# Enable logging
logger:
  baud_rate: 0     # Disable UART logging (pins GPIO1/3 are used for Nextion communication)
  
uart:
  id: uart_1
  rx_pin: GPIO3
  tx_pin: GPIO1
  baud_rate: 9600
  
 
# Enable Home Assistant API
api:

ota:
  password: !secret OTA_password

wifi:
  ssid: !secret Wifi_ssid
  password: !secret Wifi_password
  
  manual_ip:
    # Set this to the IP of the ESP
    static_ip: 192.168.1.31
    # Set this to the IP address of the router. Often ends with .1
    gateway: 192.168.1.1
    # The subnet of the network. 255.255.255.0 works for most home networks.
    subnet: 255.255.255.0

  ap:
    ssid: "Esphomethermo-V0"
    password: !secret OTA_password

captive_portal:


    
display:
  - platform: nextion
    id: nextion1
    update_interval: 5s
    uart_id: uart_1
    lambda: |-
      it.set_component_value("dActualTemp", 50);
      it.set_component_text("t0", "Hello World!");
```

```txt
[12:10:29][W][nextion:072]: Nextion is not connected! 
[12:10:29][E][uart:015]: Reading from UART timed out at byte 0!
```

### Additional information

_No response_

edwardtfn:

This adds the possibility to set whether the Nextion display should skip the connection handshake process.
When skip_connection_handshake is true, the connection will be established without performing the handshake.
This can be useful when connecting to Nextion Simulator, as it won't performs the handshake process.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/issues#2872

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3951

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
display:
  - id: disp1
    platform: nextion
    uart_id: tf_uart
    skip_connection_handshake: true
```

## Checklist:


esphome/components/nextion/nextion.h
```diff
@@ -1160,6 +1175,9 @@ class Nextion : public NextionBase, public PollingComponent, public uart::UARTDe
   void set_exit_reparse_on_start_internal(bool exit_reparse_on_start) {
     this->exit_reparse_on_start_ = exit_reparse_on_start;
   }
+  void set_skip_connection_handshake_internal(bool skip_connection_handshake) {
+    this->skip_connection_handshake_ = skip_connection_handshake;
+  }
```
jesserockz: This function seems redundant as this is a simple boolean flag.
```suggestion
```

esphome/components/nextion/nextion_commands.cpp
```diff
@@ -55,6 +55,7 @@ bool Nextion::set_protocol_reparse_mode(bool active_mode) {
   return all_commands_sent;
 }
 void Nextion::set_exit_reparse_on_start(bool exit_reparse) { this->exit_reparse_on_start_ = exit_reparse; }
+void Nextion::set_skip_connection_handshake(bool skip_handshake) { this->skip_connection_handshake_ = skip_handshake; }
```
jesserockz:
```suggestion
```

esphome/components/nextion/nextion.h
```diff
@@ -926,6 +926,21 @@ class Nextion : public NextionBase, public PollingComponent, public uart::UARTDe
...
+   * This can be useful when using Nextion Simulator.
+   *
+   * Example:
+   * ```cpp
+   * it.set_skip_connection_handshake(true);
+   * ```
+   *
+   * When set to true, the display will be marked as connected without performing a handshake.
+   */
+  void set_skip_connection_handshake(bool skip_handshake);
```
jesserockz:
```suggestion
  void set_skip_connection_handshake(bool skip_handshake) { this->skip_connection_handshake_ = skip_handshake; }
```

esphome/components/nextion/nextion.cpp
```diff
@@ -262,8 +276,10 @@ void Nextion::loop() {
       this->goto_page(this->start_up_page_);
     }
 
+    // This could probably be removed from the loop area, as those are redundant.
     this->set_auto_wake_on_touch(this->auto_wake_on_touch_);
     this->set_exit_reparse_on_start(this->exit_reparse_on_start_);
+    this->set_skip_connection_handshake(this->skip_connection_handshake_);
```
jesserockz:
```suggestion
```

esphome/components/nextion/display.py
```diff
@@ -118,6 +120,12 @@ async def to_code(config):
 
     cg.add(var.set_exit_reparse_on_start_internal(config[CONF_EXIT_REPARSE_ON_START]))
 
+    cg.add(
+        var.set_skip_connection_handshake_internal(
```
jesserockz:
```suggestion
        var.set_skip_connection_handshake(
```

jesserockz (dismissed):

esphome/components/nextion/nextion.h
edwardtfn: Indeed, but it is in line with other similar function on this component. I'm fine about simplifying this, but then shouldn't we simplify all the other functions on this component? I'm fine about doing this in another PR later.

jesserockz: Yes, if it is only setting a variable, then it should not have a duplicate `_internal` function that does the same thing. It would be better to make those changes in another PR though and keep this one small and about the connection handshake.

kbx81 (approved):

PR #6908, opened 2024-06-13
Inherit `esp32_ble_beacon` from `esp32_ble`

### LINKED ISSUE #2689: Make `esp32_ble_tracker` and `esp32_ble_beacon` interoperable

I would like to be able to configure my esphome `bluetooth_proxy`'s to also act as `esp32_ble_beacon`s. Currently, adding a `esp32_ble_beacon:` section to the yaml generates an error stating that `Component esp32_ble_beacon cannot be used together with component esp32_ble_tracker.`

I created the [Bermuda](/agittins/bermuda) custom integration which uses gathered advertisements to create location information in HA for BLE beacons, including room-based localisation.

A major goal for the project is full trilateration of the bluetooth environment, which in part requires solving the distances between esphome proxies. I have ways of doing this, but accuracy would benefit from being able to configure the proxies to periodically send out  a beacon packet so that the other proxies can estimate their distance from it. A side-effect of this is the requirement that proxies do this at different times, so retaining the functionality of the `min_interval` and `max_interval` options of `esp32_ble_beacon` would be necessary, and a working config would probably require increasing them significantly so as to not adversely affect `bluetooth_proxy` performance.

Ideally I would probably want to configure them to only send out adverts every hour or so, since the integration will maintain a record of estimated distances.

**Additional context**

Key features desired:
- Allow `esp32_ble_tracker`, `bluetooth_proxy` and `esp32_ble_beacon` to co-exist
- Ideally, allow the beacon to be turned on and off via lambda or have long interval times specified for better proxy performance

The Shelly Plus (or Gen2?) devices appear to do this according to user reports, I have no idea if there's any shared lineage there.

This pull request https://github.com/esphome/esphome/pull/3474 discusses the feature, and looked to have nearly got there but for a change in direction re implementation, but I have not been able to find subsequent attempts at the feature. My understanding from that is that implementing a core `esp32_ble` component would be the primary goal, then making the other ble-related components dependent on that.

Assumptions/observations tested on esphome 2024.3.2, using `esp-idf` framework, `wemos_d1_mini32` board. Happy to provide attempted yaml if relevant.

jpbede:

Inherit the `esp32_ble_beacon` component from `esp32_ble` component, so the `esp32_ble_beacon` can be used along with `bluetooth_proxy` or something like that.

I hope the code fits, my C++ is a little bit older

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2689

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/esp32_ble/__init__.py
+++ b/esphome/components/esp32_ble/__init__.py
@@ -10,6 +10,7 @@ CODEOWNERS = ["@jesserockz", "@Rapsssito"]
 
 CONF_BLE_ID = "ble_id"
 CONF_IO_CAPABILITY = "io_capability"
+CONF_ADVERTISING_CYCLE_TIME = "advertising_cycle_time"
 
 NO_BLUETOOTH_VARIANTS = [const.VARIANT_ESP32S2]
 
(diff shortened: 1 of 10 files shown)


jesserockz (changes requested):
To better make this non-conflicting, the code that builds and calls `esp_ble_gap_config_adv_data_raw` should be moved into the `esp32_ble` component files like we have for other advertising services so it can be managed in one place there.

Rapsssito (approved):

jesserockz:
Hi @jpbede,

I thought about what I proposed and today I came back to this and decided to make it even more modular so that other components can set the raw advertising data as they wish and they will all get turns based on the new `advertising_cycle_time`. 

Let me know what you think. 
I have tested this on a device with improv and beacon and saw it switching nicely via nRF connect on my Android phone.

jpbede:
This makes way more sense. Thanks for improving it further 👍

jesserockz (approved):

PR #6910, opened 2024-06-15
Fix garbled graphics on LILYGO T4-S3 display

manuelkasper:

On my LILYGO T4-S3, shapes drawn on the display often become garbled, with each consecutive line shifted to the left (see photo, generated by the config.yaml below).

The issue seems to be that the `qspi_amoled` driver does not honor the restriction of the RM690B0 and RM67162 that the x/y start coordinates and width/height set using CASET/RASET must be divisible by 2 according to the datasheet.

My patch grows the update "window" as necessary to get even values.

Note: I don't have a LILYGO T-Display S3 AMOLED (RM67162) to test with.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
display:
  - platform: qspi_amoled
    id: amoled_display
    model: RM690B0
    data_rate: 80MHz
    dimensions:
      width: 450
      height: 600
      offset_width: 16
    color_order: rgb
    invert_colors: false
    brightness: 255
    cs_pin: 11
    reset_pin: 13
    enable_pin: 9
    update_interval: 5s
    auto_clear_enabled: false
    lambda: |-
      it.clear();
      auto red = Color(255, 0, 0);
      auto green = Color(0, 255, 0);
      auto blue = Color(0, 0, 255);
      it.filled_circle(100, 110, 50, red);
      it.filled_circle(100, 220, 50, green);
      it.filled_circle(100, 330, 50, blue);

```

## Checklist:

### DIFF
--- a/esphome/components/qspi_amoled/display.py
+++ b/esphome/components/qspi_amoled/display.py
@@ -45,7 +45,7 @@ DATA_PIN_SCHEMA = pins.internal_gpio_output_pin_schema
 
 def validate_dimension(value):
     if value % 2 != 0:
-        raise cv.Invalid("Width/height must be divisible by 2")
+        raise cv.Invalid("Width/height/offset must be divisible by 2")
     return value
 
 
@@ -65,8 +65,12 @@ CONFIG_SCHEMA = cv.All(
                             cv.Required(CONF_HEIGHT): cv.All(
                                 cv.int_, validate_dimension
                             ),
-                            cv.Optional(CONF_OFFSET_HEIGHT, default=0): cv.int_,
-                            cv.Optional(CONF_OFFSET_WIDTH, default=0): cv.int_,
+                            cv.Optional(CONF_OFFSET_HEIGHT, default=0): cv.All(
+                                cv.int_, validate_dimension
+                            ),
+                            cv.Optional(CONF_OFFSET_WIDTH, default=0): cv.All(
+                                cv.int_, validate_dimension
+                            ),
                         }
                     ),
                 ),
--- a/esphome/components/qspi_amoled/display.py
+++ b/esphome/components/qspi_amoled/display.py
@@ -42,6 +42,13 @@ COLOR_ORDERS = {
 }
 DATA_PIN_SCHEMA = pins.internal_gpio_output_pin_schema
 
+
+def validate_dimension(value):
+    if value % 2 != 0:
+        raise cv.Invalid("Width/height must be divisible by 2")
+    return value
+
+
 CONFIG_SCHEMA = cv.All(
     display.FULL_DISPLAY_SCHEMA.extend(
         cv.Schema(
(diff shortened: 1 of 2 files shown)


esphome/components/qspi_amoled/qspi_amoled.cpp
```diff
@@ -26,6 +26,19 @@ void QspiAmoLed::setup() {
 
 void QspiAmoLed::update() {
   this->do_update_();
+  // Start addresses and widths/heights must be divisible by 2 (CASET/RASET restriction in datasheet)
+  if (this->x_low_ % 2 == 1) {
+    this->x_low_--;
+  }
+  if (this->x_high_ % 2 == 0 && this->x_high_ < this->get_width_internal() - 1) {
```
clydebarrow: No need to check the upper bound here as it is clamped in `draw_absolute_pixel_internal()` anyway. Same for the y co-ordinate (and if it _was_ necessary here, it would also be necessary for the lower bound.) Otherwise looks good.

manuelkasper: > No need to check the upper bound here as it is clamped in `draw_absolute_pixel_internal()` anyway.

We're potentially incrementing `x_high_`/`y_high_` after `draw_absolute_pixel_internal()` has done its checks. However, in the usual case that the display width/height are even, the highest "high" value encountered will be odd, so no increment will take place. I still thought it would be prudent to avoid an out-of-bounds situation, but have now removed this check, and instead added a check that the display width/height are even in the config validation, for good measure.

> and if it _was_ necessary here, it would also be necessary for the lower bound.

Since we're only decrementing x_low_/y_low_ if it's uneven, there's no risk of decrementing below 0.

clydebarrow: The point about the lower bound was that it's not required _only_ because it's already limited in `draw_absolute_pixel()`, just like the upper bound. If those checks weren't done there, then you have to check both before setting the draw window, but because they are, you don't need to check either. Checks like this should be symmetrical, if they're not, you should be asking why.

>  in the usual case that the display width/height are even,

This isn't just usual, it's actually required. If a user specified an odd value then the check you had would have papered over that crack rather than fixing the underlying problem, which you have now done with the dimensions check.

You also need to check that `offset_width` and `offset_height` are even.

esphome/components/qspi_amoled/display.py
```diff
@@ -42,6 +42,13 @@
 }
 DATA_PIN_SCHEMA = pins.internal_gpio_output_pin_schema
 
+
+def validate_dimension(value):
+    if value % 2 != 0:
```
clydebarrow, line 48: Also should check that these values are positive - and can then simplify the validation in the schema to just `validate_dimension` rather than the `cv.All( ...`
```suggestion
    value = cv.positive_int(value)
    if value % 2 != 0:
```

clydebarrow (changes requested):

clydebarrow (approved):

clydebarrow:
If this is good to go, mark it ready for review.

manuelkasper:
Sorry, I got sidetracked by another issue with the initial screen never showing if `update()` is called before the `setTimeout()` calls in `setup()` have fired. But that is a different story and somewhat dependant on my particular setup. I'll investigate some more and file a separate PR if necessary. So yes, this change is good to go.

PR #6912, opened 2024-06-16
UART component support added for host platform

paveldn:

Support for UART component for host platform (Linux only). These changes will allow debugging new code for many components that use UART for communication in Linux easily.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4004

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
uart:
  id: ${uart_id}
  baud_rate: 9600
  port: "/dev/ttyUSB0"
  debug:
    direction: BOTH
    dummy_receiver: true
    after:
      delimiter: "\n"
    sequence:
      - lambda: |-
          id(${uart_id}).write_array(&bytes[0], bytes.size());
          UARTDebug::log_hex(uart::UART_DIRECTION_RX, bytes, ' ');

```

## Checklist:


esphome/components/uart/__init__.py
```diff
@@ -181,6 +200,7 @@ def maybe_empty_debug(value):
             cv.Required(CONF_BAUD_RATE): cv.int_range(min=1),
             cv.Optional(CONF_TX_PIN): pins.internal_gpio_output_pin_schema,
             cv.Optional(CONF_RX_PIN): validate_rx_pin,
+            cv.Optional(CONF_NAME): cv.string_strict,
```
jesserockz: `name` could be confusing here, maybe `port` instead?
And you can also use `cv.file_` to ensure it exists at validation time
```suggestion
            cv.Optional(CONF_PORT): cv.All(cv.file_, cv.only_on(PLATFORM_HOST)),
```

esphome/components/uart/__init__.py
```diff
@@ -95,6 +98,20 @@ def validate_invert_esp32(config):
...
+        if CONF_TX_PIN in config or CONF_RX_PIN in config:
+            raise cv.Invalid(
+                "TX and RX pins are not supported for UART on host platform."
+            )
+        if CONF_NAME not in config:
+            raise cv.Invalid("Name is required for UART on host platform.")
+    else:
+        if CONF_NAME in config:
+            raise cv.Invalid("Name is not supported for UART on non-host platforms.")
+    return config
```
jesserockz: With the changes to the field itself, this can be thinned out. 

The `else` is no longer required because of `only_on()`, and the `if CONF_NAME not in config:` will never be reached because of `cv.has_at_least_one_key`

Technically we could put `cv.only_on` on the TX/RX fields to, but that could get messy.

```suggestion
def validate_port_name(config):
    if CORE.is_host:
        if CONF_TX_PIN in config or CONF_RX_PIN in config:
            raise cv.Invalid(
                "TX and RX pins are not supported for UART on host platform."
            )
    return config
```

esphome/components/uart/__init__.py
```diff
@@ -45,13 +46,15 @@
...
 )
+HostUartComponent = uart_ns.class_("HostUartComponent", UARTComponent, cg.Component)
 
 NATIVE_UART_CLASSES = (
     str(IDFUARTComponent),
     str(ESP32ArduinoUARTComponent),
     str(ESP8266UartComponent),
     str(RP2040UartComponent),
     str(LibreTinyUARTComponent),
+    str(HostUartComponent),
```
jesserockz: This list is only used to validate tx/rx pins, and is not needed to add here.
```suggestion
```

esphome/components/uart/uart_component_host.cpp
```diff
@@ -0,0 +1,258 @@
...
+  speed_t baud = get_baud(this->baud_rate_);
+  if (baud == B0) {
+    ESP_LOGE(TAG, "Unsupported baud rate: %d", this->baud_rate_);
+    return;
+  }
+  this->file_descriptor_ = ::open(this->port_name_.c_str(), O_RDWR | O_NOCTTY | O_NDELAY);
+  if (this->file_descriptor_ == -1) {
+    this->update_error_(strerror(errno));
+    return;
+  }
```
jesserockz: These should also mark the component as failed. `this->mark_failed()`

esphome/components/uart/uart_component_host.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+bool HostUartComponent::peek_byte(uint8_t *data) {
+  if (this->file_descriptor_ == -1) {
+    return false;
+  }
+  if (this->available() == 0) {
+    return false;
+  }
+  if (read(this->file_descriptor_, &data, 1) != 1) {
```
jesserockz: For consistency should this (and any other native calls) be `::read`?

jesserockz (changes requested):

esphome/components/uart/__init__.py
```diff
@@ -149,6 +163,11 @@ def maybe_empty_debug(value):
     return DEBUG_SCHEMA(value)
 
 
+def validate_port(value):
+    if not re.match(r"^\/dev\/(?:(?:tty(?:S|USB)|pts\/)\d+)$", value):
```
jesserockz: This does not cover ttyACMx devices

paveldn: Thanks. Completely forgot about them. I will fix it.

clydebarrow:
Any reason this wouldn't work on MacOS?

jesserockz:
I would say test it out and if it does, no reason to exclude it.
I personally don't have access to a Mac.

paveldn:
@clydebarrow  Same situation for me, I don't have a way to test it on Mac. Theoretically, the only thing that is blocking it for MAC is `#ifndef __linux__` in uart_component_host.cpp. If you will test it let me know how it goes.

clydebarrow:
I use a Mac but am travelling right now, so won't be able to test for a few weeks.

paravoid (commented):
Thanks for working on this! I had just started to implement this and have a draft branch in my checkout that I will now happily delete :laughing:

Shouldn't `final_validate_device_schema` also be adjusted to validate the port?

Also, it should probably skip validating the pins, as well as require_tx/require_rx, as ttys are implicitly read/write (assuming both of their underlying serial pins are wired). Basically it'd be great to be able to use the few integrations that use `final_validate_device_schema` without modifying them.

paveldn:
@paravoid 
Thanks. If I knew somebody working on this I wouldn't start it myself 😆. This functionality is important for me because it will allow me to do some basic tests for my component without hardware at all.  As for your remarks, you are completely right I will adjust my code.  I think require_tx/require_rx part is already covered because the HostUartComponent not in NATIVE_UART_CLASSES array. All other things I will implement. Thanks for help!

paveldn:
@paravoid, @jesserockz  I don't think I need to change something in final_validate_device_schema. Because HostUartComponent is not in NATIVE_UART_CLASSES require_tx/require_rx ignored, you can use components that require them without any problem. I tried to build a config with a couple of them (jsn_sr04t, cse7761) and it worked. I added baud rate validation but everything else looks OK. Let me know if you think something else should be adjusted. Thanks.

jesserockz:
Yes correct, the pins are only validated on the native MCU uart types. This was added previously for the weikei uart expanders.

esphome/components/uart/__init__.py
```diff
@@ -95,6 +98,54 @@ def validate_invert_esp32(config):
     return config
 
 
+def validate_host_config(config):
+    if CORE.is_host:
+        if CONF_TX_PIN in config or CONF_RX_PIN in config:
+            raise cv.Invalid(
+                "TX and RX pins are not supported for UART on host platform."
+            )
+        if CORE.is_host:
```
jesserockz: This is already inside a `is_host` block

esphome/components/uart/__init__.py
```diff
@@ -95,6 +98,54 @@ def validate_invert_esp32(config):
...
+                921600,
+                1000000,
+                1152000,
+                1500000,
+                2000000,
+                2500000,
+                3000000,
+                3500000,
+                4000000,
+            ]
```
jesserockz: Please put this outside the validator as HOST_BAUD_RATES

esphome/components/uart/__init__.py
```diff
@@ -95,6 +98,54 @@ def validate_invert_esp32(config):
...
+            if (CONF_BAUD_RATE in config) and (
+                not config[CONF_BAUD_RATE] in baud_rate_supported
+            ):
+                raise cv.Invalid(
+                    f"Host platform doesn't support baud rate {config[CONF_BAUD_RATE]}"
+                )
```
jesserockz: As baud is required, it will always be there
```suggestion
            if config[CONF_BAUD_RATE] not in HOST_BAUD_RATES):
                raise cv.Invalid(
                    f"Host platform doesn't support baud rate {config[CONF_BAUD_RATE]}", 
                    path=CONF_BAUD_RATE,
                )
```

jesserockz (changes requested):

paravoid:
> @paravoid, @jesserockz I don't think I need to change something in final_validate_device_schema. Because HostUartComponent is not in NATIVE_UART_CLASSES require_tx/require_rx ignored, you can use components that require them without any problem. I tried to build a config with a couple of them (jsn_sr04t, cse7761) and it worked. I added baud rate validation but everything else looks OK. Let me know if you think something else should be adjusted. Thanks.

Ah, I hadn't realized that. Thanks so much for looking into it and my apologies for pointing you at the wrong direction!

nagyrobi:
@paveldn any docs update for this?

paveldn:
> @paveldn any docs update for this?

Potentially there can be some documentation update but should we do it? I will need to add a port option description for the UART component that will be valid only for the host platform and I suppose UART for the host shouldn't be used for some "production" units just by developers for testing purposes. So I am afraid updating documentation can confuse other customers trying to configure their devices. 
What is your opinion? Should I update the documentation?

nagyrobi:
A PR for the doc is needed, we can all collaborate on the arrangement and wording, and only merge it when it's appropriate enough. Since `host` platform is already merged and available to anyone (certainly it's an advanced thing, but that's obvious), it's trivial to prepare documentation for this feature too.

paveldn:
@nagyrobi  OK Will prepare PR for doc changes

paveldn:
@nagyrobi Done. Can you please check it? https://github.com/esphome/esphome-docs/pull/4004

nagyrobi:
Here's my use case for this component: https://github.com/hn/linkplay-a31

esphome/components/uart/__init__.py
```diff
@@ -149,6 +201,12 @@ def maybe_empty_debug(value):
     return DEBUG_SCHEMA(value)
 
 
+def validate_port(value):
+    if not re.match(r"^\/dev\/(?:(?:tty(?:S|USB|ACM)|pts\/)\d+)$", value):
```
clydebarrow: I'd suggest just checking that the path starts with `/dev/tty` or `/dev/cu` - the remainder of the device name is going to be driver dependent.

clydebarrow (changes requested):

esphome/components/uart/__init__.py
paveldn: Good point. Implemented.

jesserockz: Even that can be too restrictive sometimes. I personally map certain device ids to named ports eg `/dev/buspirate` :skull_and_crossbones: 
It might make it more flexible in the long term if someone was to have an ESPHome binary running and wanted a consistent port name that could survive reboots, replugs etc.

paravoid: +1 to this, but to add to it, I find even /dev/ to be restrictive. One can create PTYs on their /tmp, or $HOME etc.A use case could be to e.g. create a program to speak the MCU side of a UART protocol (i.e. an emulator), and point ESPHome to this PTY for local development. Or in simpler terms, use socat to bridge a PTY with a TCP/IP socket.

paveldn: OK. I am not against it. Different people have different use cases. Personally, I want to use UART host implementation to have a way to easily debug when developing my Haier AC component using socat. 
So @paravoid what would be your suggestion? Do not validate the port name at all?

clydebarrow: Yeah, sure. It will fail at run time if invalid which since it's on a host will be pretty obvious.

paveldn: OK Maybe we can agree on something like this 
^\/(?:[\w\s.-]+\/)*[\w\s.-]+$
Which will just check if the path is a proper Linux path without other restrictions?

clydebarrow: The ?: is unnecessary since you don't care if the group is capturing or not. That also restricts the allowable characters more than the Linux FS does - is there a good rationale for that? `^/.*[^/]$` should be sufficient - ensures it's absolute and doesn't end with /.

paveldn: OK. Will do it with [^/] but I will keep mine ?: I prefer to do it the right way even if it doesn't matter.

clydebarrow (approved):

nagyrobi:
Next thing to add would be something like ha's [shell_command](https://www.home-assistant.io/integrations/shell_command/)...

To be able to run linux commands from ESPHome triggers...

paveldn:
> Next thing to add would be something like ha's [shell_command](https://www.home-assistant.io/integrations/shell_command/)...
> 
> To be able to run linux commands from ESPHome triggers...

Personally for me running bash scripts as a part of ESPHome is already too much. There are enough ways to have automation on Linux and it is too different from ESP platforms. For me, the best feature of the host platform is to run the same code that running on ESP for debugging and testing. But of course, everybody can have their own use cases and ways to use ESPHome.

paveldn:
@jesserockz  Is there something else I need to do here? I would really like to see it in the next release.

jesserockz (approved):

PR #6925, opened 2024-06-17
making mqtt ON/OFF messages configurable

dariuskt:

Until now `binary_sensor.gpio/fan.binary/output.gpio` sensors could only return strings `ON` or `OFF` via MQTT as intended for HA use only.
This PR lets user to disable this behavior and get a more common bool representation of `1` or `0` via MQTT.
Added configuration option `ha_flavored` to MQTT module which defaults to `true` to keep backward compatibility

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
mqtt:
  ha_flavored: false


binary_sensor:
  - platform: gpio
    pin: D8
    id: bs
    state_topic: ${root_topic}/gpio/bs/get

switch:
  - platform: gpio
    pin: D1
    id: sw
    state_topic: ${root_topic}/gpio/sw/get

fan:
  - platform: binary
    output: fan_out
    id: fn
    state_topic: ${root_topic}/gpio/fn/get

output:
  - platform: gpio
    pin: D2
    id: fan_out

interval:
  - interval: 1s
    then:
      - lambda: |-
          id(bs).publish_state(true);
      - switch.toggle: sw
      - fan.toggle: fn


```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs). esphome/esphome-docs#3959


nagyrobi:
I would have added more flexibility...
Like optional just a couple of configuration items to customize the default strings to anything else. Like not only `1`/`0`, but also `true`/`false`, `ding`/`dong`, `what`/`ever`... the power of MQTT...

dariuskt:
I was considering that approach but decided against it as it raises a lot more architectural, consistency, and implementation questions for me. I would require a lot more time from the community when it is appropriate for such a small feature IMHO. That would become less of a contribution and more coding by proxy. 

I understand this is not the best possible solution, but the messages were hardcoded for many years until now. So I am considering this PR a slight improvement. I also agree that exposing non-optimal configuration to users will make it harder to adopt a better implementation in the future, in case someone will consider doing it...

Whether this is worth it or not is not for me to decide tho ;)

jesserockz:
Any further changes in the future will be breaking if we decide to go a different route. I agree with @nagyrobi and that these could have `payload_on: str` / `payload_off: str` with defaults set to what HA expects, and allows the user to alter them in whatever way they want.

dariuskt:
I slept on the issue and now I fully agree that this implementation is not good for the project long term.
So I'm closing this PR as it's not relevant to a proper solution anyway and I'll see if I can do better.

PR #6958, opened 2024-06-21
[qspi_amoled] Fix display remaining blank after update() before setup completion

manuelkasper:

When using a fairly basic display setup like in the example below on my LILYGO T4 S3, nothing appears on the display. The reason is that `update()` is called before the asynchronous `setTimeout()` tasks started by `setup()` have completed. The desired updates to the display buffer are still made (via `do_update_()` => `draw_absolute_pixel_internal()`), but `draw_pixels_at()` will rightfully refuse to write the updated buffer to the display because `setup_complete_ == false`. However, `update()` will still invalidate the low/high watermarks.

When `update()` is called again later with the same graphics/lambda, no changes to the display buffer will be detected, the watermarks will not be set again, and the display will remain blank, no matter how many further updates are made, unless the content is different.

Admittedly this will only affect users with an `update_interval` who don't use `auto_clear_enabled` and who don't manually call `clear()` in the lambda, e.g. for efficiency reasons if only a small part of the display needs to change. It did bite me though and took some time to figure out.

My simple suggested change is for `update()` not to proceed and invalidate the watermarks if setup has not completed yet. In this way, the first update after setup completion will cause the desired display output, as expected.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
display:
  - platform: qspi_amoled
    id: amoled_display
    model: RM690B0
    data_rate: 80MHz
    dimensions:
      width: 450
      height: 600
      offset_width: 16
    cs_pin: 11
    reset_pin: 13
    enable_pin: 9
    update_interval: 5s
    auto_clear_enabled: false
    lambda: |-
      it.filled_circle(100, 110, 50);
```

## Checklist:

### DIFF
--- a/esphome/components/qspi_amoled/qspi_amoled.cpp
+++ b/esphome/components/qspi_amoled/qspi_amoled.cpp
@@ -25,6 +25,9 @@ void QspiAmoLed::setup() {
 }
 
 void QspiAmoLed::update() {
+  if (!this->setup_complete_) {
+    return;
+  }
   this->do_update_();
   // Start addresses and widths/heights must be divisible by 2 (CASET/RASET restriction in datasheet)
   if (this->x_low_ % 2 == 1) {


clydebarrow (approved):
LGTM

PR #6965, opened 2024-06-21
Add ccache to ESPHome builds

### LINKED ISSUE #975: ccache for building

The builds are slow and even for copy-and-pasted configuration take a long time. Using ccache for the builds would require only single compilation for most of the code (since the node name is built-in, there still would be one compilation and linking, but that's already way shorter than full build for every new device).

It's just to spend less time waiting for the firmware builds.

**Additional context**

None.

jaharkes:

This PR adds support for ccache, which speeds up compilation to detect if the same code is being compiled.

My local testing showed a ~65% cache hit rate on the first build of a random mix of ESP8266 and ESP32 devices, actual wall clock time improvement depends a lot on available CPU and how much time is spend linking/uploading/flashing the final images.

Right now, the compiler cache is stored in its default location which is `$USER/.ccache` so it is cleared whenever the Docker container is recreated or replaced.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes esphome/feature-requests#975

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

No changes necessary, this is completely transparent to the end user.

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -28,6 +28,7 @@ RUN \
     apt-get update \
     # Use pinned versions so that we get updates with build caching
     && apt-get install -y --no-install-recommends \
+        ccache \
         python3-pip=23.0.1+dfsg-1 \
         python3-setuptools=66.1.1-1 \
         python3-venv=3.11.2-1+b1 \
--- a/esphome/use_ccache.py.script
+++ b/esphome/use_ccache.py.script
@@ -5,8 +5,10 @@
 # pylint: disable=E0602
 Import("env")  # noqa
 
-env.Replace(
-  AS="ccache %s" % env["AS"],
-  CC="ccache %s" % env["CC"],
-  CXX="ccache %s" % env["CXX"],
-)
+ccache = env.WhereIs("ccache")
+if ccache is not None:
+    env.Replace(
+      AS="%s %s" % (ccache, env["AS"]),
+      CC="%s %s" % (ccache, env["CC"]),
+      CXX="%s %s" % (ccache, env["CXX"]),
+    )
--- a/esphome/use_ccache.py.script
+++ b/esphome/use_ccache.py.script
@@ -1,0 +1,12 @@
+# Derived from
+#   - https://github.com/platformio/platformio-core/issues/4592
+#   - https://github.com/platformio/platform-espressif32/issues/1073
+
+# pylint: disable=E0602
+Import("env")  # noqa
+
+env.Replace(
+  AS="ccache %s" % env["AS"],
+  CC="ccache %s" % env["CC"],
+  CXX="ccache %s" % env["CXX"],
+)
--- a/esphome/writer.py
+++ b/esphome/writer.py
@@ -142,6 +142,9 @@ def get_ini_content():
     # Sort to avoid changing build flags order
     CORE.add_platformio_option("build_flags", sorted(CORE.build_flags))
 
+    CORE.add_platformio_option("board_build.cmake_extra_args", "-DCCACHE_ENABLE=ON")
+    CORE.add_platformio_option("extra_scripts", ["post:use_ccache.py"])
+
     content = "[platformio]\n"
     content += f"description = ESPHome {__version__}\n"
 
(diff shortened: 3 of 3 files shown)


jaharkes:
It is very easy to make this cache persistent between restarts and upgrades of the Docker container by setting the `CCACHE_DIR=/config/.ccache` environment variable.

jesserockz:
I get no improvement here with two identical configs except for the `name` and `friendly_name`.

```
esphome compile config/m5stack-dial-1.yaml  194.23s user 37.35s system 731% cpu 31.674 total
esphome compile config/m5stack-dial-2.yaml  194.59s user 36.81s system 721% cpu 32.085 total
```

`ccache` is installed locally:
```sh
$ which ccache
/usr/bin/ccache
```

<details>
<summary>tree ~/.cache/ccache</summary>

```sh
/home/jesse/.cache/ccache
├── 0
│   └── 4
│       └── stats
├── 2
│   └── 9
│       └── stats
├── 6
│   └── 6
│       └── stats
├── 7
│   └── 6
│       └── stats
├── 8
│   └── 0
│       └── stats
├── a
│   └── 6
│       └── stats
├── c
│   └── 2
│       └── stats
└── e
    ├── 1
    │   └── stats
    ├── 2
    │   └── stats
    ├── 3
    │   └── stats
    ├── 5
    │   └── stats
    └── 7
        └── stats

21 directories, 12 files
```

</details>

jaharkes:
Thanks for trying it. Not sure what is going on there but it looks like it never even tried to run ccache. I'm rebuilding a container from scratch here to reproduce. Maybe the second commit I created that tries to test if ccache is installed isn't working quite right.

jaharkes:
Not sure what is going on because I'm definitely seeing ccache working. I rebuilt everything in a docker container ran one compilation to make sure platformio components got cached and made sure to remove .esphome/build and ~/.cache/ccache after the first compile. I then compiled the firmware for two of my S31 switches with only name and friendly name differences. The first compilation took a little over 30 seconds. The second compilation took only 22 seconds.
```
# time esphome compile bedroom-ac.yaml
...
real    0m31.501s                                                                                        
user    2m41.987s                                                                                        
sys     0m16.344s
                                                                                        
# time esphome compile office-ac.yaml
...
real    0m22.336s                                                                                        
user    2m1.532s                                                                                         
sys     0m10.334s

# ccache -s
Cacheable calls:    166 /  170 (97.65%)
  Hits:              83 /  166 (50.00%)
    Direct:          83 /   83 (100.0%)
    Preprocessed:     0 /   83 ( 0.00%)
  Misses:            83 /  166 (50.00%)
Uncacheable calls:    4 /  170 ( 2.35%)
Local storage:
  Cache size (GB): 0.00 / 5.00 ( 0.02%)

# tree ~/.cache/ccache  
/root/.cache/ccache                         
├── 0                                       
│   ├── 0                                   
│   │   └── 6211pbm26pn9071o9vmisjmjh2ospd8M
│   ├── 1                                   
│   │   └── stats                           
│   ├── 3                                   
│   │   └── stats                           
│   ├── 5                                   
│   │   ├── e1l83f4ngds0da94n0thlepe9947iaiR
│   │   └── stats                           
│   ├── 7                                   
│   │   ├── 5enfn8e7u2eoqg0gd2edrsb7502lbe6R
│   │   └── stats                           
│   ├── 9                                   
│   │   └── stats                           
│   ├── CACHEDIR.TAG                        
│   ├── a                                   
│   │   ├── 6frmraq104omf9jp0mtk2bv9rv3shrcR
...
│   ├── b                                   
│   │   ├── e986rmg0kjjsi4sbfjf8aa1h2ekca48R
│   │   └── stats                           
│   ├── d                                   
│   │   └── stats                           
│   ├── e                                   
│   │   └── f6svcgcmg2fn4n1mdkoarl0b4u9706kR
│   ├── f                                   
│   │   └── stats                           
│   └── stats                               
└── tmp                                     
                                                    
181 directories, 285 files                  
```
Definitely a head scratcher because in your case it seems ccache was not even used.

jaharkes:
I've started running my builds from the command line inside the docker container instead of using the dashboard. One advantage is that I was able to run `esphome -v compile config.yaml`, which gives tons of extra info. Also added print statements in places.

Long story longer, this looks like a platformio issue, even when the CC and CXX variables are set in the scons environment, it actually only seems to compile a subset of files with what was specified. So I would estimate maybe only half the files ended up being compiled with ccache, which explains why ccache reports a high cache hit ratio, but the speed improvements are modest.

Then there is some per-project build caching by platformio itself, which might explain why in your case ccache was not even used at all.

Either way, I'm closing this until I can figure out what is happening inside of platformio that makes it ignore the settings from the construction environment.

PR #665, opened 2019-07-10
add time based cover, has built in endstop

glmnet:
## Description:
As my blinds has built in end stops I had to add a way to prevent this component to stop the blinds before they fully open or they fully close.
The way my blinds works, they have a remote control (rc switch) with three buttons, up, down and stop. The blinds themselves have the end stop logic to stop on the limits but you can also stop in any position you'd like.
Without this the blinds gets out of sync with the estimated position of the component (it starts at 50% after cold boot) and would likely open only half way and then stay there and only solution is to use original remote controller.

I added the Optional `has_built_in_endstop:` which defaults to false so is backwards compatible, may be this could be an opportunity to individualize built in end stops at open or closed position only.

Off: why a 100% position means fully open cover? is this imposed by HA?
Cover 0% -> cover closed
Cover 100% -> cover open
this just seems awkward to me

**Related issue (if applicable):** fixes not really but somewhat related:  https://github.com/esphome/issues/issues/355

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs# https://github.com/esphome/esphome-docs/pull/302

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #6969, opened 2024-06-22
Remove code that got introduced by #665

saibotma:

#665 but actually has nothing to do with "has_built_in_endstop" and got added a second time as a duplicate with the "manual_control" configuration in https://github.com/esphome/esphome/pull/4249.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3981

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


saibotma:
Keep open

bdraco:
Hey! Thanks for working on this. The code change itself looks right, but I'm having trouble understanding what this PR actually does from the description, and that's probably why it hasn't been merged yet.

Right now the description is written in terms of PR numbers and duplicate code, which makes sense if you've been deep in this code, but doesn't really help me understand what's changing for users or what I should put in the changelog.

When I read "has nothing to do with has_built_in_endstop" but then see you're removing code from the has_built_in_endstop feature, it's confusing. Like, what behavior is changing? Who's affected? What should they do?

Here's what I think is happening after digging through the code and those old PRs:

The `has_built_in_endstop` option currently does two different things:
1. It skips sending stop commands when the cover reaches fully open/closed (the actual endstop thing)
2. It re-triggers commands when you tell the cover to go somewhere it already is

Your PR is removing `#2` because `manual_control` already does that same job. But `#1` stays the same.

The problem is this could break configs if someone is using `has_built_in_endstop: true` specifically because they need that re-triggering behavior to keep their cover in sync. They'd need to know to add `manual_control: true` instead.

Could you rewrite the PR description to explain it more like that? In plain terms that help users understand what's changing and what they might need to update in their config? That would make it way easier to merge and write a good changelog entry.

Something like:
- What behavior is being removed (the re-triggering thing)
- Why it's safe to remove (because manual_control does it)
- What users should do if they relied on it (add manual_control: true)

swoboda1337 (changes requested):
See bdraco's comment, needs a better explanation.

PR #6973, opened 2024-06-24
[http_request] memory leak fix

### LINKED ISSUE #5955: 2024.6.1 http_request GET memory leak
### The problem

After updating to 2024.6.1 and changing http_request and json configurations accordingly in a bigger project I am working on, I have seen that the ESP8266 I am using is reporting out of memory errors after a short while and finally crashes due to that. I have played around to find the root cause and have prepared a minimal configuration (without json) that shows that when using http_request and setting the capture_response variable to true the module leaks memory. For that I have prepared a simple timer that tries to GET a 512 byte txt file and having the debug sensor module enabled showing free heap space. 

When having capture_response set to false the free heap size stays constant but when having capture_response set to true the free heap size steadily declines until the microcontroller runs out of memory and crashes. This also happens when there is "nothing to be done to the response" e.g. when commenting out the "on_response" block or commenting out the debug output in the on_response lambda. 

I have attached the minimal config I was using to reliably reproduce the memory leak. As mentioned, you can try and comment out the logging in the lambda in line 41 or just completely comment out the on_response block in line 37-41 -> when setting capture_response: true in line 36 it will leak memory.

I know there are already other issues (#5949, #5951) open regarding http_request but in this issue I try to provide you with a minimal example and my findings on that matter and my assumption that it is caused by a memory leak in http_request when using capture_response / handling of the response body. I have not found a way to somehow free dangling memory allocated internally by the http_request component from within the on_response lambda.

2024.6.1

Docker

_No response_

ESP8266

### Board

D1 Mini

http_request

### Example YAML snippet

```yaml
esphome:
  name: httprequestmemleakexample
  friendly_name: HttpRequestMemLeakExample

esp8266:
  board: d1_mini

# Enable Home Assistant API
api:
  encryption:
    key: "OFAtXlMrl9spHbpJnP4YKnU9GClHbaUJPPdT9K/j8gE="

ota:
  - platform: esphome
    password: "debugonly"

wifi:
  ssid: !secret wifi_ssid

captive_portal:
    
http_request:
  useragent: esphome/device
  id: http_liquid
  timeout: 10s
  verify_ssl: false

time:
  - platform: sntp
    on_time: 
      - seconds: /1
        then:
          - http_request.get:
              url: "http://krautomation.tech/debug/debug_512.txt"
              capture_response: true # setting this to false and the memory does not leak
              on_response:
                then:
                  - lambda: |-
                      // try and comment the following line out; it still leaks memory when capture_response is set to true
                      ESP_LOGD("http_request_response", "body len: %d", body.length());

debug:
  update_interval: 5s

sensor:
  - platform: debug
    free:
      name: "Heap Free"

# Logger must be at least debug (default)
logger:
  level: debug
```

_No response_

### Additional information

_No response_

gabest11:

With `capture_response: true` the temporary buffer is never freed after copying it into a string.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [<link to issue>](https://github.com/esphome/issues/issues/5955)

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/http_request/http_request.h
+++ b/esphome/components/http_request/http_request.h
@@ -149,6 +149,7 @@ template<typename... Ts> class HttpRequestSendAction : public Action<Ts...> {
         }
         response_body.reserve(read_index);
         response_body.assign((char *) buf, read_index);
+        allocator.deallocate(buf, max_length);
       }
     }
 


jesserockz (approved):

gabest11:
@jesserockz I wonder if the original intent was to allocate external RAM and use std::string assign to **move** the pointer (but it's a copy operation), explaning the missing deallocate. Now it allocates the buffer twice in this function, doubling the need, and it can still fail when std::string runs out of memory. So, I don't really know if this is the right solution, it is correct, but I think it was not designed to do this. You could just use reserve for std::string and directly read into that buffer.

PR #6975, opened 2024-06-24
[modbus_text_sensor] new default ANSI encoding type

dudanov:

New default `ANSI` encoding.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3994

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/modbus_controller/text_sensor/__init__.py
+++ b/esphome/components/modbus_controller/text_sensor/__init__.py
@@ -37,6 +37,7 @@ RAW_ENCODING = {
     "NONE": RawEncoding.NONE,
     "HEXBYTES": RawEncoding.HEXBYTES,
     "COMMA": RawEncoding.COMMA,
+    "ASCII": RawEncoding.ASCII,
 }
 
 CONFIG_SCHEMA = cv.All(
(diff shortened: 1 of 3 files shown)


jesserockz:
Wouldn't this be a potential breaking change for anyone who is using/parsing data that had the nulls in it?
(A breaking change is not a bad thing, just needs to be clear)

dudanov:
> Wouldn't this be a potential breaking change for anyone who is using/parsing data that had the nulls in it?
> (A breaking change is not a bad thing, just needs to be clear)

I don't think this will be a breaking change.  It is unlikely that there are devices that transmit null characters between characters.  Only the case when the register encodes a character.  In this case, it is precisely necessary to filter nulls.  The concept of a C-STRING implies a sequence of characters that ends with a null character.  Even if the device needs to transmit enumeration of strings, separators such as commas, semicolons, etc. will be used.

dudanov:
This might be an option:

We already have a NONE filter, which will pass all characters without filtering. And by default, make sure that null character filtering is always used. I will reflect these changes in the documentation.

What do you think of this option, Jesse?

jesserockz:
Do you mean add another Encoding type that will do the null filtering?

dudanov:
> Do you mean add another Encoding type that will do the null filtering?

Yes, I create a new `Encoding type` and use it instead of `NONE` by default.  And those few users who really need zero characters in their config use the `NONE` `Encoding type`.

dudanov:
Hmmm... Well, somehow the issue of filtering null characters turned into an `ANSI` encoding type. :)

nagyrobi:
Is change this documented?

dudanov:
> Is change this documented?

Yes, I added a link.

jesserockz (approved):

PR #6998, opened 2024-06-26
'uart' and 'improv_serial' need to understand non-UART logger configurations

kpfleming:

Ensures that all hardware UARTs can be used when the logger is using USB_CDC or USB_SERIAL_JTAG. Also adds component tests for `logger` configurations using those interfaces, and adds tests for `uart` configurations with the maximum number of hardware UARTs along with the logger using non-UART interfaces (those tests cannot currently fail as the 'maximum number' check is performed at run-time, not configuration-time or compile-time).

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


jesserockz (approved):

esphome/components/improv_serial/improv_serial_component.cpp
```diff
@@ -99,19 +99,19 @@ void ImprovSerialComponent::write_data_(std::vector<uint8_t> &data) {
 #endif  // !USE_ESP32_VARIANT_ESP32C3 && !USE_ESP32_VARIANT_ESP32S2 && !USE_ESP32_VARIANT_ESP32S3
       uart_write_bytes(this->uart_num_, data.data(), data.size());
       break;
-#if defined(CONFIG_ESP_CONSOLE_USB_CDC) && (defined(USE_ESP32_VARIANT_ESP32S2) || defined(USE_ESP32_VARIANT_ESP32S3))
+#ifdef USE_LOGGER_USB_CDC
```
kbx81, line 102: So I'm just realizing that, when this is set but `CONFIG_ESP_CONSOLE_USB_CDC` is not set, some code is excluded by IDF from the build, resulting in a linker error...

kbx81: Specifically, I believe this is the point of interest: https://github.com/espressif/esp-idf/blob/38eeba213aa695aabfd6d89aa9f5078dbe5a94c3/components/esp_system/port/soc/esp32s3/CMakeLists.txt#L19-L21
...so I think that's not being built, but we are still including code based on this define, which results in the linker error.

kbx81: I think the right move might be to do this:
```suggestion
#if defined(USE_LOGGER_USB_CDC) && defined(CONFIG_ESP_CONSOLE_USB_CDC)
```
@kpfleming any thoughts?

kbx81: We also need to add more Improv serial tests for each ESP that natively supports USB -- it probably would've caught the issue. 🤦🏻

kbx81: The linker error can be seen [here](https://github.com/esphome/esphome/actions/runs/9925278247/job/27417528461?pr=7081).
PR to fix is #7082 -- we can continue discussion there, if necessary.

PR #7000, opened 2024-06-26
Add remote transmitter listener to support 433MHz auto tx on/off 

swoboda1337:

Need a way to turn the radio transmitter on and off for 433MHz radios (transmitters should not be left on all the time). Add a way for radio components to listen for transmit events. This will allow 433MHz radio components to be properly integrated with remote transmitter. This PR is the simplest way to accomplish this integration.

Radio components would have to register for transmit events and then turn on and off the transmitter appropriately. Here is a working example using my radio component:

https://github.com/swoboda1337/sx127x-esphome/compare/main...remote_transmitter_listener

Why this is needed:
- Many transceivers support both rx and tx. Without a way to switch modes the transceiver would have to be locked into rx or tx all the time. With this change the transceiver can switch modes automatically. 
- The transceiver could be locked to tx mode permanently but its generally not a good idea. Even in OOK there will be some leakage. If ASK or FSK are used its a really bad idea as it will be transmitting real power all the time. 

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (changes requested):
We have a `power_supply` component that can do this instead of having `listeners`.

swoboda1337:
> We have a `power_supply` component that can do this instead of having `listeners`.

The power supply component seems to write to a gpio. Normally these radios will need to do a spi write not toggle a gpio. Having a callback like this is very flexible and each radio can do what it needs to (gpio, spi, i2c, uart, etc etc).

jesserockz:
Ah ok. I wasnt sure how you planned to attach to these listeners since there are no yaml configs for it added.

swoboda1337:
> Ah ok. I wasnt sure how you planned to attach to these listeners since there are no yaml configs for it added.

The radio component would have to get the remote transmitter id and register itself. I have a poc working for my radio component here:

https://github.com/swoboda1337/sx127x-esphome/compare/main...remote_transmitter_listener

jesserockz:
Yes, it would have made sense if you put that in the PR description :sweat_smile:

jesserockz (changes requested):
I think the calls to `call_listeners_transmit_`/`complete_` needs to happen outside of the loops. Otherwise they could cause timing issues. It should only need to turn on once at the beginning of an entire transmission, and once off at the end.

swoboda1337:
> I think the calls to `call_listeners_transmit_`/`complete_` needs to happen outside of the loops. Otherwise they could cause timing issues. It should only need to turn on once at the beginning of an entire transmission, and once off at the end.

Yeah it can be moved out. The user could put a long delay between each send. I wasn't sure if its better to turn off the tx or not. This would mess with the timing. I guess its probably best to keep the transmitter on. The user can send multiple times if they want it off.

swoboda1337:
Any chance at getting this PR merged? I want add support for the rf chip but need this first. Thanks!

swoboda1337:
Moved to draft temporarily, possibly a better way to implement this

swoboda1337:
I implemented this in a better way in: https://github.com/esphome/esphome/pull/7483

PR #7008, opened 2024-06-27
[qspi_amoled] Improve drawing performance

manuelkasper:

When drawing large filled shapes (or using the very common `clear()` operation), the separate function calls for each pixel drawn cause a lot of overhead.

This PR reduces the processing time for drawing filled shapes in qspi_amoled by about an order of magnitude (see benchmarks below).

Due to the clipping and rotation logic, I have opted to optimize `filled_rectangle()`, and make `horizontal_line()`  and `vertical_line()` call it with a 1 pixel rect. `horizontal_line()` is used for drawing filled circles and polygons and significantly boosts performance there as well.

I had to make the three optimized drawing functions `virtual` in display.h to render this possible. In the event that this is not acceptable, I can make a new PR that at least overrides the (already virtual) `fill()` for better full-screen fill/clear performance.

### Benchmarks

Time spent drawing graphics to the buffer in `do_update_()`, not including the time taken to send the modified buffer to the display (QSPI). Test device: LILYGO T4 S3, rotation 0.

Test | Time before | Time after PR
---- | ------------ | ------------
test_card | 1407 ms | 147 ms
full screen fill, all pixels changed | 625 ms | 77 ms
full screen fill, nothing changed | 590 ms | 43 ms
draw filled circle with 200 px radius | 332 ms | 42 ms

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/display/display.h
+++ b/esphome/components/display/display.h
@@ -266,17 +266,17 @@ class Display : public PollingComponent {
   void line_at_angle(int x, int y, int angle, int start_radius, int stop_radius, Color color = COLOR_ON);
 
   /// Draw a horizontal line from the point [x,y] to [x+width,y] with the given color.
-  void horizontal_line(int x, int y, int width, Color color = COLOR_ON);
+  virtual void horizontal_line(int x, int y, int width, Color color = COLOR_ON);
 
   /// Draw a vertical line from the point [x,y] to [x,y+width] with the given color.
-  void vertical_line(int x, int y, int height, Color color = COLOR_ON);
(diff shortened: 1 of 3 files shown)


clydebarrow:
Rather than add this to the qspi_amoled component, it would be better implemented in a layer that could be the parent of other buffer oriented display components. Or just use LVGL.

PR #6878, opened 2024-06-10
Implement ByteBuffer

clydebarrow:
# What does this implement/fix?

Based on [Java's ByteBuffer](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html) class, this implements a class that allows wrapping an array of bytes and performing
put and get operations of various data types, with configurable endian-ness.

Follows on from discussion here: https://github.com/esphome/esphome/pull/6865#discussion_r1632394779

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
esphome:
  name: bytebuffer
  on_boot:
    - lambda: |-
        auto buf = ByteBuffer::create(16);
        assert(buf.get_endianness() == Little);
        assert(buf.get_remaining() == 16);
        buf.set_limit(10);
        assert(buf.get_capacity() == 16);
        buf.put_uint8(1);
        assert(buf.get_remaining() == 9);
        buf.put_uint16(0x1234);
        assert(buf.get_remaining() == 7);
        buf.put_uint32(0x12345678UL);
        assert(buf.get_remaining() == 3);
        assert(buf.array()[1] == 0x34);
        assert(buf.array()[2] == 0x12);
        assert(buf.array()[3] == 0x78);
        assert(buf.array()[4] == 0x56);
        assert(buf.array()[5] == 0x34);
        assert(buf.array()[6] == 0x12);
        buf.flip();
        assert(buf.get_capacity() == 16);
        assert(buf.get_remaining() == 7);
        assert(buf.get_uint8() == 1);
        assert(buf.get_uint16() == 0x1234);
        assert(buf.get_uint32() == 0x12345678ul);
        assert(buf.get_remaining() == 0);
        buf.rewind();
        buf.big_endian();
        assert(buf.get_remaining() == 7);
        assert(buf.get_uint8() == 1);
[...]

PR #7009, opened 2024-06-27
[esp32_ble_server] Create custom services, characteristics and descriptors

Rapsssito:

Adds the ability to create a fully custom BLE server with its own services, characteristics and descriptors.

## Types of changes

- [ ] Other

### Aditional changes
- [x] Added `components/event_emitter.h`, which is an EventEmitter templatable class that can emit events with a specific name (it is highly recommended to use an enum class for this) and a list of arguments. Supports multiple listeners for each event (it is used by `BLECharacteristic` for emitting `ON_READ` and `ON_WRITE` events, by `BLEServer` for emitting `ON_CONNECT` and `ON_DISCONNECT` events and by `BLEDescriptor` for emitting `ON_WRITE` events). Could be used in the future to complete the INDICATE implementation in `BLECharacteristic` and many other async interactions without the need of creating an interface class for each use case.

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4002

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esp32_ble:

esp32_ble_server:
  id: ble_server
  manufacturer_data: [0x72, 0x4, 0x00, 0x23]
  manufacturer: ESPHome
  model: Test
  on_connect:
    - lambda: |-
        ESP_LOGD("BLE", "Connection from %d", id);
  on_disconnect:
    - lambda: |-
        ESP_LOGD("BLE", "Disconnection from %d", id);
  services:
    - uuid: 2a24b789-7aab-4535-af3e-ee76a35cc12d
      advertise: false
      characteristics:
        - id: test_notify_characteristic
          description: "Notify characteristic"
          uuid: cad48e28-7fbe-41cf-bae9-d77a6c233423
          read: true
          notify: true
          value: [1, 2, 3, 4]
          descriptors:
            - uuid: cad48e28-7fbe-41cf-bae9-d77a6c111111
              on_write:
                logger.log:
                  format: "Descriptor write id %u, data %s"
                  args: [id, 'format_hex_pretty(x.data(), x.size()).c_str()']
              value:
                data: "123.1"
                type: float
                endianness: BIG
    - uuid: 2a24b789-7a1b-4535-af3e-ee76a35cc42d
      advertise: false
      characteristics:
        - id: test_change_characteristic
          uuid: 2a24b789-7a1b-4535-af3e-ee76a35cc11c
          read: true
          value:
            data: "Initial"
            string_encoding: utf-8
          description: Change characteristic
          descriptors:
            - uuid: 0x4414
              id: test_change_descriptor
              value: "Initial descriptor value"
            - uuid: 0x2312
              value:
                data: 0x12
                type: uint16_t
              on_write:
                - lambda: |-
                    ESP_LOGD("BLE", "Descriptor received: %s from %d", std::string(x.begin(), x.end()).c_str(), id);
        - uuid: 2a24b789-7a1b-4535-af3e-ee76a35cc99a
          write: true
          on_write:
            then:
              - lambda: |-
                  ESP_LOGD("BLE", "Characteristic received: %s from %d", std::string(x.begin(), x.end()).c_str(), id);
              - ble_server.characteristic.set_value:
                  id: test_change_characteristic
                  value: !lambda 'return bytebuffer::ByteBuffer::wrap({0x00, 0x01, 0x02}).get_data();'
              - ble_server.characteristic.notify:
                  id: test_notify_characteristic
              - ble_server.descriptor.set_value:
                  id: test_change_descriptor
                  value: !lambda return bytebuffer::ByteBuffer::wrap({0x03, 0x04, 0x05}).get_data();
```

## Checklist:


clydebarrow (changes requested):
Why not make the value of a characteristic templatable? A read request from a client would trigger evaluation of the lambda. Could also add a `notify` action to trigger evaluation and notification, and separate it from `set_value` - which I'd prefer to be named `ble_server.characteristic.set_value`, i.e. use a dot instead of the underscore.
Instead of a `properties` key, why not make `read`, `write` etc top-level Boolean keys, defaulting to false? I would also prefer `write_nr` to be the full `write_no_response`.

Rapsssito:
> Why not make the value of a characteristic templatable? A read request from a client would trigger evaluation of the lambda. Could also add a `notify` action to trigger evaluation and notification, and separate it from `set_value` - which I'd prefer to be named `ble_server.characteristic.set_value`, i.e. use a dot instead of the underscore. Instead of a `properties` key, why not make `read`, `write` etc top-level Boolean keys, defaulting to false? I would also prefer `write_nr` to be the full `write_no_response`.

The issue I found with setting the value of a characteristic templatable is that it must always be a string (or any other type). I was not able to find a way to overload it.

I will make the other changes ASAP!

Rapsssito:
@clydebarrow, I have your requested changes staged. I don't want to commit them yet as I also want @jesserockz feedback on this.

@jesserockz, is there any way to overload a templatable entry? If it is not the case, do you think the benefits of making the "value" entry templatable will outweigh the benefits of the compatibility of having many types (byte array, int, float, boolean)?

@clydebarrow, regarding `ble_server.characteristic.set_value`, at the moment this only sets the value of the characteristic with the output of the lambda **when the automation is called** because it matches the current paradigm (i.e. the value is not computed when requested, but before). If @jesserockz agrees, we can change the paradigm so the value is always set when it is requested. Either way, I will add a `ble_server.characteristic.notify` automation, as it is already implemented in the C code.

clydebarrow:
The way I would do it is to only compute the value when a read is requested, or the notify action is executed. The value would be templatable (requiring a byte array as value), but also allow fixed values of various types, which get translated to lambdas (templates) yielding a byte array at build time, so the C++ code only has to deal with a template and one type. Or you could store a fixed value as a byte array and not set a lambda for invariant values. Some utility functions callable in lambdas would make providing a dynamic value for e.g. ints or floats simpler.  That is in keeping with other BLE implementations I've worked with. It means that if the value is not fixed, it is freshly computed when needed, rather than representing a computation that took place at some time in the past, so is guaranteed to be as up-to-date as possible.It would avoid the need to run a `set_value` action on an interval or other event, which is pointless if nobody is reading it.

Rapsssito:
@clydebarrow, I have implemented the changes you requested (also marked as outdated my other comments). Let me know what you think!

Rapsssito:
@clydebarrow @jesserockz, pinging you with a friendly reminder about this!

esphome/core/helpers.h
```diff
@@ -441,6 +441,17 @@ std::string base64_encode(const std::vector<uint8_t> &buf);
 std::vector<uint8_t> base64_decode(const std::string &encoded_string);
 size_t base64_decode(std::string const &encoded_string, uint8_t *buf, size_t buf_len);
 
+/// Create a byte vector from multiple types of values.
```
clydebarrow: There should be at least a comment here indicating what the endian-ness of the conversion is.

esphome/components/esp32_ble_server/ble_characteristic.cpp
```diff
@@ -37,57 +37,14 @@ void BLECharacteristic::set_value(std::vector<uint8_t> value) {
...
-}
 void BLECharacteristic::set_value(const uint8_t *data, size_t length) {
   this->set_value(std::vector<uint8_t>(data, data + length));
 }
-void BLECharacteristic::set_value(uint8_t &data) {
-  uint8_t temp[1];
-  temp[0] = data;
-  this->set_value(temp, 1);
-}
-void BLECharacteristic::set_value(uint16_t &data) {
```
clydebarrow, line 51: These functions were storing data in little-endian order, whereas the vector functions added in helpers.cpp appear to store in big-endian format. The endian-ness for characteristic data is going to be application-specific. Have a look at https://github.com/esphome/esphome/pull/6878 as an alternative.

clydebarrow (commented):
This will be a useful addition, but there's a lot to review, have only looked at a little so far.

esphome/components/esp32_ble_server/ble_characteristic.cpp
Rapsssito: I have changed all calls from `to_vector` to ByteBuffer (https://github.com/esphome/esphome/pull/6878) to avoid code duplication. I have also added a few new methods for populating a ByteBuffer from types like double, bool, string, etc.

I have updated the docs (https://github.com/esphome/esphome-docs/pull/4002) accordingly to warn about little-endian behavior by default.

esphome/core/bytebuffer.cpp
```diff
@@ -8,13 +9,74 @@ ByteBuffer ByteBuffer::create(size_t capacity) {
...
 
+ByteBuffer ByteBuffer::wrap(uint8_t value) {
+  ByteBuffer buffer = ByteBuffer::create(1);
+  buffer.put_uint8(value);
+  return buffer;
+}
+
+ByteBuffer ByteBuffer::wrap(uint16_t value, Endian endianness) {
+  ByteBuffer buffer = ByteBuffer::create(2);
+  if (endianness == BIG) {
```
clydebarrow: Could just assign to the property here - that's perfectly fine inside the class, is likely faster than the test, and makes no assumptions about the default.

clydebarrow: Alternatively the create method could take an endianness parameter.

esphome/core/bytebuffer.h
```diff
@@ -29,18 +30,34 @@ enum Endian { LITTLE, BIG };
  */
 class ByteBuffer {
  public:
+  // Default constructor (compatibility with TEMPLATABLE_VALUE)
+  ByteBuffer() = default;
```
clydebarrow: Why is this needed? It leaves the vector uninitialised so is dangerous.

esphome/core/bytebuffer.cpp
Rapsssito: Indeed! I have implemented your suggestion in the last commit (6a7c4b2e595e8c591054b740b095c31db3d0563f)

esphome/core/bytebuffer.h
Rapsssito: It is needed because `TEMPLATABLE_VALUE` initializes the object with the class default constructor when instantiated:
https://github.com/esphome/esphome/blob/80a0f137224c82fdc3cf17dfdd5dd3683ad9e796/esphome/core/automation.h#L66
I have changed the default constructor (https://github.com/esphome/esphome/commit/6a7c4b2e595e8c591054b740b095c31db3d0563f) so the vector is initialized empty.

esphome/components/esp32_ble_server/ble_characteristic.cpp
clydebarrow: If you break out the ByteBuffer changes to a separate PR that can be reviewed and merged quickly. Also the change to const.py and text_sensor etc. It's much easier for testing as if there are no core changes in the PR, since these don't get pulled in by an external component.

esphome/components/esp32_ble_server/__init__.py
```diff
@@ -11,44 +19,337 @@
...
+        result = result | BLECharacteristic_ns.PROPERTY_WRITE
+    if char_conf[CONF_NOTIFY]:
+        result = result | BLECharacteristic_ns.PROPERTY_NOTIFY
+    if char_conf[CONF_BROADCAST]:
+        result = result | BLECharacteristic_ns.PROPERTY_BROADCAST
+    if char_conf[CONF_INDICATE]:
+        result = result | BLECharacteristic_ns.PROPERTY_INDICATE
+    if char_conf[CONF_WRITE_NO_RESPONSE]:
+        result = result | BLECharacteristic_ns.PROPERTY_WRITE_NR
+    return result
```
clydebarrow:
```suggestion
PROPERTY_MAP = {
    CONF_READ: BLECharacteristic_ns.PROPERTY_READ,
    CONF_WRITE: BLECharacteristic_ns.PROPERTY_WRITE,
    CONF_NOTIFY: BLECharacteristic_ns.PROPERTY_NOTIFY,
    CONF_BROADCAST: BLECharacteristic_ns.PROPERTY_BROADCAST,
    CONF_INDICATE: BLECharacteristic_ns.PROPERTY_INDICATE,
    CONF_WRITE_NO_RESPONSE: BLECharacteristic_ns.PROPERTY_WRITE_NR,
}

def parse_properties(char_conf):
    return sum([PROPERTY_MAP[k] for k in char_conf if k in PROPERTY_MAP], start=0)
```

Can then also use the map keys to generate the corresponding entries in the config schema e.g.
```
).extend({
    cv.Optional(k, default=False): cv.boolean for k in PROPERTY_MAP.keys()
})
```

esphome/components/esp32_ble_server/__init__.py
```diff
@@ -11,44 +19,337 @@
...
+        {
+            cv.Required(CONF_ID): cv.use_id(BLECharacteristic),
+        }
+    ),
+)
+async def ble_server_characteristic_notify(config, action_id, template_arg, args):
+    paren = await cg.get_variable(config[CONF_ID])
+    # Check if the NOTIFY property is set from the global configuration
+    assert _ble_server_config is not None
+    for service_config in _ble_server_config[CONF_SERVICES]:
```
clydebarrow: Several issues here; it's not acceptable to raise `cv.Invalid()` in a `to_code` function - it must be done during validation; using a global like that is problematic.

Best solution I can come up with;

* populate a map of characteristics IDs to configs during validation, along with a list of IDs used in notify actions.
* Define a final validate function that checks these:
```
characteristic_map = {}
notifications = set()

def final_validator(config):
    for id in notifications:
        if not characteristic_map[id][CONF_NOTIFY]:
            raise cv.Invalid(f"Characteristic {id} does not support notifications")
    return config

FINAL_VALIDATE_SCHEMA = final_validator

def store_config(config):
    characteristic_map[config[CONF_ID]] = config
    return config


```

The `store_config` function is added to the SERVICE_CHARACTERISTIC_SCHEMA with `cv.All()` Similarly for ids referenced in the notify action.

esphome/components/esp32_ble_server/ble_descriptor.cpp
```diff
@@ -38,8 +38,10 @@ void BLEDescriptor::do_create(BLECharacteristic *characteristic) {
   this->state_ = CREATING;
 }
 
-void BLEDescriptor::set_value(const std::string &value) { this->set_value((uint8_t *) value.data(), value.length()); }
-void BLEDescriptor::set_value(const uint8_t *data, size_t length) {
+void BLEDescriptor::set_value(ByteBuffer buffer) {
+  size_t length = buffer.get_capacity();
+  uint8_t *data = buffer.array();
```
clydebarrow: Only referenced once, just inline the `buffer.array()` call.

esphome/components/esp32_ble_server/ble_server.cpp
```diff
@@ -38,9 +39,27 @@ void BLEServer::loop() {
...
+        for (unsigned i = 0; i < this->services_to_start_.size(); i++) {
+          BLEService *service = this->services_to_start_[i];
+          if (service->is_created()) {
+            service->start();
+          } else {
+            index_to_remove = i + 1;
+          }
+        }
+        // Remove the services that have been started
+        if (index_to_remove > 0) {
```
clydebarrow, line 50: What scenario would mean that the vector can't simply be cleared after starting the services?

clydebarrow (changes requested):
Partial review. My eyes were glazing over so had to stop.

esphome/components/esp32_ble_server/__init__.py
jesserockz: `CORE.data[KEY_ESP32_BLE_SERVER]` can be used to store and retrieve things later

clydebarrow: So where in the core data would you find the notify actions which aren't part of the component config?

clydebarrow: Ah, you're saying that's a good place to store things rather than a random global.

esphome/components/esp32_ble_server/ble_characteristic.cpp
```diff
@@ -32,62 +32,18 @@ BLECharacteristic::BLECharacteristic(const ESPBTUUID uuid, uint32_t properties)
...
-}
 
-void BLECharacteristic::notify(bool notification) {
-  if (!notification) {
-    ESP_LOGW(TAG, "notification=false is not yet supported");
-    // TODO: Handle when notification=false
+void BLECharacteristic::notify(bool require_ack) {
+  if (require_ack) {
+    ESP_LOGW(TAG, "require_ack=true is not yet supported (i.e. INDICATE is not yet supported)");
+    // TODO: Handle when require_ack=true
```
clydebarrow: Need to test here if `service_` is valid - this can be called by an action before the component is setup. If you initialise it in the class definition in the header file it can be tested against `nullptr`, currently it's uninitialised and will be a garbage pointer before being set. Same in `ble_service.h` for `  BLEServer *server_;`

esphome/components/esp32_ble_server/ble_characteristic.cpp
Rapsssito: I have created #7316 with all core changes (and `ble_client`).

esphome/components/esp32_ble_server/__init__.py
Rapsssito: You are right! Thanks! I have included this change in 96026e8d07500f9f2b5db336b546488dd8c20f52

esphome/components/esp32_ble_server/__init__.py
Rapsssito: I have taken your suggestions and used the `extra_schemas` to implement the validation. Let me know what you think of the solution:

```python

def validate_notify_characteristic(char_config):
    if KEY_ESP32_BLE_SERVER_NOTIFY_PROVIDED not in CORE.data:
        CORE.data[KEY_ESP32_BLE_SERVER_NOTIFY_PROVIDED] = dict()
    CORE.data[KEY_ESP32_BLE_SERVER_NOTIFY_PROVIDED][char_config[CONF_ID]] = char_config[CONF_NOTIFY]
    # Check if the NOTIFY property is set if the characteristic has a notify action
    char_ids = CORE.data.get(KEY_ESP32_BLE_SERVER_NOTIFY_REQUIRED, set())
    if not char_config[CONF_NOTIFY] and char_config[CONF_ID] in char_ids:
        raise cv.Invalid("Characteristic has a notify action, but the NOTIFY property is not set")
    return char_config


def validate_notify_action(action_char_id):
    if KEY_ESP32_BLE_SERVER_NOTIFY_REQUIRED not in CORE.data:
        CORE.data[KEY_ESP32_BLE_SERVER_NOTIFY_REQUIRED] = set()
    CORE.data[KEY_ESP32_BLE_SERVER_NOTIFY_REQUIRED].add(action_char_id)
    # Check if the NOTIFY property is set for the characteristic
    char_notify_value = CORE.data.get(KEY_ESP32_BLE_SERVER_NOTIFY_PROVIDED, dict()).get(action_char_id, None)
    if char_notify_value is not None and not char_notify_value:
        raise cv.Invalid("Missing NOTIFY property for characteristic with notify action")
    return action_char_id


SERVICE_CHARACTERISTIC_SCHEMA = cv.Schema(
    {
        cv.GenerateID(): cv.declare_id(BLECharacteristic),
        cv.Required(CONF_UUID): UUID_SCHEMA,
        cv.Optional(CONF_WRITE_NO_RESPONSE, default=False): cv.boolean,
        cv.Optional(CONF_VALUE): CHARACTERISTIC_VALUE_SCHEMA,
        cv.GenerateID(CONF_VALUE_ACTION_ID_): cv.declare_id(
            BLECharacteristicSetValueAction
        ),
        cv.Optional(CONF_DESCRIPTORS, default=[]): cv.ensure_list(DESCRIPTOR_SCHEMA),
        cv.Optional(CONF_ON_WRITE): automation.validate_automation(
            {cv.GenerateID(): cv.declare_id(BLECharacteristic)}, single=True
        ),
    },
    extra_schemas=[validate_on_write, validate_notify_characteristic],
).extend({cv.Optional(k, default=False): cv.boolean for k in PROPERTY_MAP.keys()})


@automation.register_action(
    "ble_server.characteristic.notify",
    BLECharacteristicNotifyAction,
    cv.Schema(
        {
            cv.Required(CONF_ID): cv.All(cv.use_id(BLECharacteristic), validate_notify_action),
        }
    ),
)
```

Regarding `validate_notify_characteristic` and `validate_notify_action`, I had to duplicate the validation logic, as I am not sure which one will be called first. Is the order of validation deterministic?

I am still not sure how to handle the `cv.Invalid` at the end of `parse_characteristic_value` and `parse_descriptor_value`. It handles an invalid state that should never occur since the validation scheme already prevents it. Should I just remove it?

esphome/components/esp32_ble_server/ble_server.cpp
Rapsssito: The thing is that `service->start()` does not start the service right away. It needs to be called once for every characteristic:

https://github.com/esphome/esphome/blob/93f30a0798def5a1b59a0ca891acdc6b975b717c/esphome/components/esp32_ble_server/ble_service.cpp#L85-L101

Rapsssito:
@clydebarrow, just pinging again as a friendly reminder.

clydebarrow:
I started some testing of this. I notice you used my Bluetooth company ID in the example above :-)

 My first attempt resulted in the program crashing, due to this:
```
      characteristics:
        - id: test_notify_characteristic
          uuid: cad48e28-7fbe-41cf-bae9-d77a6c233423
          read: true
          notify: true
          value: !lambda return random_uint32();
```

There were no errors or warnings during compilation, even though the generated code returned a uint32 where a ByteBuffer was expected. I would have expected the compiler to complain.

I changed that wrap the uint32 in a bytebuffer, and that worked.

Next step, this characteristic says it supports notifications, but when connecting to the device, it's not possible to enable notifications, due to the lack of a CCCD.

So I added a descriptor for the CCCD:
```
          descriptors:
            - uuid: 0x2902
              value: 0
```
This then meant that in the client application I could apparently subscribe to notifications - but when doing so, there was no event in the ESPHome code to tell me that the descriptor had been written to. Also, when reading the CCCD it gave an error because 4 bytes were returned instead of 2.

This raises the question then, it is worth supporting descriptors in general? If so it needs to be done in full, with events and lambdas for values. If not (or as well) then the commonly used descriptors (CCCD and User Description) should be implicitly added as required - CCCD added if notify or indicate is enabled, and CUD via a suitable config option e.g. `description:`.

This would also imply a trigger when the CCCD is written to, or just conditions to test if notifications or indications are enabled.

For non-lambda characteristic values it should be possible to specify the size, e.g.:
```
    value: 0
    bytes: 2
```

or something similar.

Basic read and write operations appear to work ok.

esphome/components/esp32_ble_server/__init__.py
```diff
@@ -11,16 +19,161 @@
...
+    CORE.data[_KEY_NOTIFY_REQUIRED].add(action_char_id)
+    # Check if the NOTIFY property is set for the characteristic
+    char_notify_value = CORE.data.get(_KEY_NOTIFY_PROVIDED, {}).get(
+        action_char_id, None
+    )
+    if char_notify_value is not None and not char_notify_value:
+        raise cv.Invalid(
+            "Missing NOTIFY property for characteristic with notify action"
+        )
+    return action_char_id
```
clydebarrow: these two functions appear to be checking the same thing. There should be a single function that is invoked with a FINAL_VALIDATE rather than trying to do it on the fly. Currently this crashes at build time if a notification action precedes the service definition.

esphome/components/esp32_ble_server/ble_2901.cpp
```diff
@@ -8,7 +9,7 @@ namespace esp32_ble_server {
 
 BLE2901::BLE2901(const std::string &value) : BLE2901((uint8_t *) value.data(), value.length()) {}
 BLE2901::BLE2901(const uint8_t *data, size_t length) : BLEDescriptor(esp32_ble::ESPBTUUID::from_uint16(0x2901)) {
-  this->set_value(data, length);
+  this->set_value(ByteBuffer::wrap(data, length));
   this->permissions_ = ESP_GATT_PERM_READ;
 }
```
clydebarrow: This doesn't appear to be used anywhere.

esphome/components/esp32_ble_server/__init__.py
```diff
@@ -29,10 +182,94 @@
...
+            args,
+            ByteBuffer,
+            ByteBuffer_ns.wrap,
+        )
+    for val_method in [
+        cv.boolean,
+        cv.float_,
+        cv.hex_uint8_t,
+        cv.hex_uint16_t,
+        cv.hex_uint32_t,
```
clydebarrow: These validators only check that the value is within range, so e.g. 0x0000 will succeed with `float_` which would be unexpected. So ` value: 0x23` actually delivers 35.0.

esphome/components/esp32_ble_server/__init__.py
```diff
@@ -29,10 +182,94 @@
...
+        start=0,
+    )
+
+
+def parse_uuid(uuid):
+    # If the UUID is a string, use from_raw
+    if isinstance(uuid, str):
+        return ESPBTUUID_ns.from_raw(uuid)
+    # Otherwise, use from_uint32
+    return ESPBTUUID_ns.from_uint32(uuid)
```
clydebarrow: This should validate the format for a string - right now it just passes it through without checks (other than length) which generates a uuid at run time with `1` for any non-hex character.

esphome/components/esp32_ble_server/__init__.py
```diff
@@ -11,16 +19,161 @@
...
+    },
+    extra_schemas=[validate_on_write, validate_notify_characteristic],
+).extend({cv.Optional(k, default=False): cv.boolean for k in PROPERTY_MAP})
 
+SERVICE_SCHEMA = cv.Schema(
+    {
+        cv.GenerateID(): cv.declare_id(BLEService),
+        cv.Required(CONF_UUID): UUID_SCHEMA,
+        cv.Optional(CONF_ADVERTISE, default=False): cv.boolean,
+        cv.Optional(CONF_NUM_HANDLES, default=0): cv.int_,
```
clydebarrow: Why would there be a need to override the calculation of the number of handles required?

clydebarrow (changes requested):

esphome/components/esp32_ble_server/ble_2901.cpp
Rapsssito: You are right! It looks like it was created by @jesserockz in a70a205ace089c30c0ffcf38b9793d7181595b04. Should I remove it?

esphome/components/esp32_ble_server/__init__.py
Rapsssito: Indeed. I have removed it.

esphome/components/esp32_ble_server/__init__.py
Rapsssito: Ups, my bad. I was not familiar with the `cv.hex_uint` validators. I have changed them to  `cv.uint` to avoid these issues.

esphome/components/esp32_ble_server/__init__.py
Rapsssito: Right! I have moved `bt_uuid` validator from `esp32_ble_tracker` to `esp32_ble` and it is now also used here.

esphome/components/esp32_ble_server/__init__.py
Rapsssito: Thanks for the comment! I did not know about the `FINAL_VALIDATE_SCHEMA`. I have followed your suggestion and used it instead of `validate_notify_characteristic`.

esphome/components/esp32_ble_server/__init__.py
clydebarrow: > Ups, my bad

No need to feel bad or apologise BTW. I know that review feedback can sometimes seem a little terse but there's no criticism implied - the whole point of the review process is to improve the code by having another pair (or preferably several pairs) of eyes on it, and it's a learning process for everyone involved.

This is a reasonably large and complex PR so it's inevitable it's going to take several rounds of review to get it right. You can be assured that this is is useful work you're doing and I'm certainly keen to see it get done.

Rapsssito:
> I started some testing of this. I notice you used my Bluetooth company ID in the example above :-)

Oops. I did not know that. I don't even remember where I took it from...

> My first attempt resulted in the program crashing, due to this:
> 
> ```
>       characteristics:
>         - id: test_notify_characteristic
>           uuid: cad48e28-7fbe-41cf-bae9-d77a6c233423
>           read: true
>           notify: true
>           value: !lambda return random_uint32();
> ```
> 
> There were no errors or warnings during compilation, even though the generated code returned a uint32 where a ByteBuffer was expected. I would have expected the compiler to complain.
> 
> I changed that wrap the uint32 in a bytebuffer, and that worked.
> 

I checked with other value types like `return "string"` and discovered the underlying issue. It looks like the lambda is calling the default constructor `ByteBuffer(size_t capacity, Endian endianness = LITTLE)` in the `TEMPLATE_VALUE` with the `uint32` value as the `size_t capacity` parameter. I am not sure if this is fixable.

> Next step, this characteristic says it supports notifications, but when connecting to the device, it's not possible to enable notifications, due to the lack of a CCCD.
> This then meant that in the client application I could apparently subscribe to notifications - but when doing so, there was no event in the ESPHome code to tell me that the descriptor had been written to. Also, when reading the CCCD it gave an error because 4 bytes were returned instead of 2.
> This raises the question then, it is worth supporting descriptors in general? If so it needs to be done in full, with events and lambdas for values. If not (or as well) then the commonly used descriptors (CCCD and User Description) should be implicitly added as required - CCCD added if notify or indicate is enabled, and CUD via a suitable config option e.g. `description:`.

Thank you very much for the feedback. I would like to point out that I am not very familiar with BLE (in fact, I am getting more familiar with it by developing this PR). I didn't think enough about "notify" or "indicate" because I don't need them in my use case. I think the best option is to support descriptors 100%, so I have implemented everything you mentioned. However, I couldn't find if (or how) a descriptor value is supposed to be updated right before it is read, so right now the descriptor `value` configuration entry is just the initial value. For this reason, the templatable value for the descriptor is still not implemented, I will implement it differently depending on when the lambda should be computed. If the lambda is only computed on startup (i.e. just the initial value), I highly recommend including and using the `InstantLambdaExpression` I proposed in #7327. 

Let me know what you think!

esphome/components/esp32_ble_server/ble_2901.cpp
clydebarrow: I think so, it does nothing that can't be done otherwise.

esphome/components/esp32_ble_server/ble_characteristic.cpp
```diff
@@ -32,78 +32,59 @@ BLECharacteristic::BLECharacteristic(const ESPBTUUID uuid, uint32_t properties)
...
-void BLECharacteristic::add_descriptor(BLEDescriptor *descriptor) { this->descriptors_.push_back(descriptor); }
+void BLECharacteristic::add_descriptor(BLEDescriptor *descriptor) {
+  // If the descriptor is the CCCD descriptor, listen to its write event to know if the client wants to be notified
+  if (descriptor->get_uuid() == ESPBTUUID::from_uint16(ESP_GATT_UUID_CHAR_CLIENT_CONFIG)) {
+    descriptor->on(
+      BLEDescriptorEvt::VectorEvt::ON_WRITE,
+      [this](const std::vector<uint8_t> &value, uint16_t conn_id) {
+        if (value.size() != 2)
+          return;
+        uint16_t cccd = (value[1] << 8) | value[0];
```
clydebarrow: There are helper functions to do this in `esphome/core/helpers.h`. The benefit of using them is that it makes it clear what you are doing, it keeps style consistent, and the code is already known to work correctly.
```suggestion
        uint16_t cccd = encode_uint16(value[1], value[0]);
```

esphome/components/esp32_ble_server/ble_characteristic.h
```diff
@@ -71,7 +75,7 @@ class BLECharacteristic {
 
  protected:
   bool write_event_{false};
-  BLEService *service_;
+  BLEService *service_ = nullptr;
```
clydebarrow:
```suggestion
  BLEService *service_{};
```
Same result, more modern idiom, keeps code style consistent.

clydebarrow (commented):
Looking good, will do some more testing, might be a few days though.

clydebarrow:
@Rapsssito I sent you a message via discord. I have some changes I'd like to push, and also need to discuss some other issues with you.

lufton:
Hey @Rapsssito, is there a way to use this implementation right away? I thought I can do:
```yaml
external_components:
  - source: github://pr#7009
    components: [ esp32_ble, esp32_ble_server ]
```

But there are also changes in a `core/event_emitter`. So I'm getting:

```
src/esphome/components/esp32_ble_server/ble_descriptor.h:4:10: fatal error: esphome/core/event_emitter.h: No such file or directory
```

upd: I solved that by copying `event_emitter.h` and `event_emitter.cpp` into esphome/core manually, but this is not convinient and I don't think is even possible via ESPHome Addon for Home Assistant.

clydebarrow:
@Rapsssito I left some comments in the Discord DM channel. Look this over and let me know, but from my perspective it's ready to go.

clydebarrow (approved):

lufton:
Oh, gosh! I was waiting for this to be merged forever! Thanks @clydebarrow @Rapsssito, you are awesome!

bdraco:
Hi everyone,

I wanted to provide some feedback on this PR after working with it in production environments. While the EventEmitter pattern and BLE customization features are powerful additions, we've discovered some significant challenges that have emerged from this implementation:

This PR has added approximately 15-30KB of flash usage (depending on usage), which is substantial for embedded systems. For context, the entire bluetooth_proxy implementation is only ~6.6KB. This has caused memory pressure on devices, particularly affecting vendors shipping products with Improv where the BLE stack now consumes more resources than most other components combined.

The original implementations were severely impacting performance - service lookups were 241x slower before optimizations in #10894 and #10714. The `std::unordered_map` approach with string key generation and hashing, while elegant for desktop applications, isn't suitable for embedded systems where we typically have just 1-3 services.

We've encountered crashes in the ESP32 Improv component that were introduced with these changes (addressed in #10902). The null pointer dereferences and uninitialized pointers have caused devices to crash when BLE client and server are used together.

The advanced features are primarily used by custom automations, which represents a small percentage of use cases, yet the memory and performance costs are paid by all users of ESP32 BLE server.

Moving forward, we're working on:
- Conditionally compiling automations to reduce overhead (#10910)
- Continuing vector-based optimizations (#10900, #10897) achieving 2.3x+ better performance
- Designing a more lightweight event system better suited for embedded constraints
- Ensuring future contributions consider memory/flash/performance impact as primary design constraints

This is a learning opportunity for all of us about balancing feature richness with embedded system constraints. Thank you for your contribution - it sparked important discussions about architecture patterns in ESPHome.

Rapsssito:
Hi @bdraco,

Thanks for your work and apologies for my poorly optimized implementation. I have no previous experience with embedded systems, so I default to design choices more commonly used in high level implementations.

This PR was actually pretty difficult for me to pull off, and I appreciate your work optimizing it. I am actually using it in production, so these new fixes are very welcomed.

Regarding the event emitter, I believe there is redundant code regarding events in the whole codebase. I realized this a few months ago (the PR was already merged), when I was working on the MQTT errors. It would be great if the event emitter functionality was optimized and the optimized version was shared across components.

Thanks again!

PR #7016, opened 2024-06-29
Add remote cover control via homeassistant and handle unavailable and unknown state.

Links2004:

This change allows to control covers managed my homeassistant the same as local covers.
With this its gets very easy to implement push button control for covers.

There are 2 new conditions for the states unavailable and unknown
- `entity.is_unavailable` 
- `entity.is_unknown`

and 2 new triggers:
- `on_unavailable`
- `on_unknown`

The conditions can be implemented by any component by using the class `EntityBase_State`.
`EntityBase_State` implements 2 ways to set the state
- `void set_entity_state(EntityStateType entity_state);`
- `bool handle_state(const char *input);` (parse the state from a string)

`EntityBase_State` has been implemented for all "homeassistant" components.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/4005

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
api:

cover:
  - platform: homeassistant
    id: my_cool_cover
    entity_id: cover.my_cool_cover
    
  - id: dummy_button
    on_multi_click:
      - timing:
          - ON for at most 0.5s
          - OFF for at least 100ms
        then:
          - if:
              condition:
                - lambda: return id(my_cool_cover).current_operation == CoverOperation::COVER_OPERATION_IDLE;
              then:
                - cover.open:  my_cool_cover
              else:
                - cover.stop:  my_cool_cover
      - timing:
          - ON for at least 1s
        then:
          - cover.open:  my_cool_cover
      - timing:
          - ON for at least 1s
          - OFF for at least 100ms
        then:
          - cover.stop:  my_cool_cover
```

## Checklist:

https://github.com/esphome/esphome-docs/pull/4005


Links2004:
docu --> https://github.com/esphome/esphome-docs/pull/4005
for switch --> #7018

nagyrobi:
Nice!
Other very common entilites like cilmate and rgb/dimmable lights would be useful too.

Links2004:
cilmate and rgb lights are much more complicated since there is no easy way I know of to discover the features.
and there are many combinations possible.

simple one color dimmable or on/off lights are possible relatively easy with no effects for example.

nagyrobi:
Do you handle the `unavailable` state somehow?

Links2004:
not directly, the service calls will not work an the HA side if the device is unavailable.

and based on the docs of HA there is no unavailable state in the API used for getting the status.
https://developers.home-assistant.io/docs/core/entity/cover/#states

hance the state in esphome if the HA entity will only be updated when its not unavailable.

nagyrobi:
Ok. We still need to configure it as a text sensor and compare the string to `unavailable` to make local automations aware of that.

Links2004:
can you post the code/yaml of getting the `unavailable` state?
you may using a non documented HA API feature?

nagyrobi:
Nope, all standard. Here I check textually for `opening` and `closing`:
https://deploy-preview-3678--esphome.netlify.app/cookbook/lvgl#cover-status-and-control

`unknown` and `unavailable` works the same, like
```
        state:
            disabled: !lambda return ((0 == x.compare(std::string{"unavailable"})) or (0 == x.compare(std::string{"unknown"})));
```

Links2004:
ok, then the HA API docs are incomplete :O

looks like esphome currently has no component `unavailable` hadling internaly.
https://github.com/esphome/feature-requests/issues/1568

so there is not generic state I can set for the component to indicate that the cover is `unavailable`.

will take a look if a custom condition is possible, like:
```
cover.is_unavailable: my_cool_cover
```

nagyrobi:
Don't forget to trigger a state change in these cases too (even if the state itself will be the same, just this custom condition changes).

Also `cover.is_unknown: my_cool_cover` would be useful, separately from `unavailable`.

Links2004:
feel free to test.
and yes `unavailable` and `unknown` are seperate.
but the state `unavailable` forces `unknown` to be true too.

`publish_state` will be called on change to  `unavailable` and `unknown`
 but the state can only be checked via `cover.is_unavailable` and `cover.is_unknown` since esphome has no concept of component `unavailable` or `unknown` at the moment.

nagyrobi:
The other entities could benefit from this too (sensor, binary sensor etc).

Links2004:
yay, had the same idea.

the question big question is its a good idea to add the unavailable and unknown bools to all entities of esphome as a optional feature, not all components need it but i can see a lot of cases outside of the homeassistant component.
e.g. components that comunicate external
 - Bluetooth
 - Modbus
 - CAN
....

basically every where there is a posibility that the ESP / esphome itself is running but the sensor or device its comunicating with is not ready or present. 

I think some feedback frome the esphome main devs is needed here.

nagyrobi:
Sure... But that's a more complex thing.

Links2004:
no really adding the bools to the EntityBase class is easy and deduplicates a lot of code.
I for sure will not implement the `unavailable` and `unknown` in one go for all component.

Its more a laying the ground work so that every component dev can reuse / implement it if it makes sense for them.
the default if not implemented in this case will be `unavailable` and `unknown` = false.
and only if the component specific code changes the state its gets usefull for the component.

landonr:
I have a lot of these home assistant components built for homeThing. 
I have the [Number component up for review](https://github.com/esphome/esphome/pull/6455) which includes this [HomeAssistantComponent](https://github.com/esphome/esphome/pull/6455/files#diff-a76257bfb481f16e8c550478c9d2202404d32ae0ba77c6eab87993272ee63251) class that's shared between all the components.

Here's my home assistant cover component https://github.com/landonr/esphome-components/blob/main/components/homeassistant_component/cover/HomeAssistantCover.h

Links2004:
interesting had a short look and did not see any handling of the `unavailable` and  or `unknown` state in home assistant.

nagyrobi:
If you implement a native esphome cover (say a time based cover), and unplug the node from network, the entity will become unavailable.

landonr:
@Links2004 you're right I haven't done any work with `unavailable`. I agree that it's good to figure out now. I think the same solution should work for all component types (switch, light, cover, etc)

@nagyrobi if the ESPHome device loses wifi, what do you suggest doing with the state? Should these home assistant components subscribe to the network status?

esphome/core/helpers.cpp
```diff
@@ -409,6 +409,21 @@ ParseOnOffState parse_on_off(const char *str, const char *on, const char *off) {
...
+  if (strcasecmp(str, "closing") == 0)
+    return PARSE_COVER_CLOSING;
+  if (strcasecmp(str, "open") == 0)
+    return PARSE_COVER_OPEN;
+  if (strcasecmp(str, "closed") == 0)
+    return PARSE_COVER_CLOSED;
+  if (strcasecmp(str, "unavailable") == 0)
+    return PARSE_COVER_UNAVAILABLE;
+
+  return PARSE_COVER_UNKNOWN;
```
landonr, line 422: @nagyrobi what do you think about this for unknown / unavailable 
@Links2004

Links2004: this code parses the info comming from `subscribe_home_assistant_state` how the info is handled it importend to.
we most likely need the same for all other entity types, may a generic func that only detects the strings `unavailable` and `unknown` that home_assistant may sends, so that we can easy chain it in all cases.

landonr: How would we keep track of the state? The PARSE_COVER_UNAVAILABLE case is specific to covers.
Maybe there should be a shared value for unknown/unavailable that exists outside of state? I'm also fine with this implementation, and having an unavailable state for every type

Links2004: have rewritten the state handling, now a component only needs to implement the `EntityBase_State` class to get `unavailable` and or `unknown`.

setting the state is as easy as calling `this->set_entity_state` or for string handling `this->handle_state`.

condition names have changed to:
- `entity.is_unavailable`
- `entity.is_unknown`

Links2004:
added trigger support e.g.:

```yaml
binary_sensor:
  - id: !extend btn_boot
    on_click:
      then:
        - if:
            condition:
              - entity.is_unavailable: my_cool_cover
            then:
              - logger.log: my_cool_cover unavailable
        - if:
            condition:
              - entity.is_unknown: my_cool_cover
            then:
              - logger.log: my_cool_cover unknown

cover:
  - platform: homeassistant
    id: my_cool_cover
    entity_id: cover.my_cool_cover
    on_unavailable:
      then:
         - logger.log: trigger my_cool_cover unavailable
    on_unknown:
      then:
         - logger.log: trigger my_cool_cover unknown
```

landonr:
I haven't tested because I don't have a Home Assistant cover (it's in the mail)
Do you have to worry about ignoring the updated from the API? For example if you change the height to different values quickly, Home Assistant might send back old values which you don't want
1. change the height of the cover on ESPHome
2. height sends from ESPHome to Home Assistant
3. Home Assistant entity updates height state
4. ESPHome receives entity height update
I have some code here for handling this https://github.com/esphome/esphome/pull/6455/files#diff-a76257bfb481f16e8c550478c9d2202404d32ae0ba77c6eab87993272ee63251R19-R24

Links2004:
I had no problems in my testing, the HA API sends a update for every 1% in may tests.

look like your code does "optimistic updates" e.g. your `control` function updates the internal esphome state of the entity.

my code does not do this, and the state is only changed when the HA API sends an update.
Which simply avoids the problem.

I dont see why we need "optimistic updates" and if its there I want a conf parameter to disable it ;)
Changing the internal state to what we think may happen can go wrong on so many levels.

e.g.
- if the service call does fail 
- HA entity does control a device with lockout
- the device has special states that happen on service calls 

the HA API will tell us the correct state as soon that the API call is successful no need for any guessing game.

landonr:
> I had no problems in my testing, the HA API sends a update for every 1% in may tests.
> 
> look like your code does "optimistic updates" e.g. your `control` function updates the internal esphome state of the entity.
> 
> my code does not do this, and the state is only changed when the HA API sends an update. Which simply avoids the problem.
> 
> I dont see why we need "optimistic updates" and if its there I want a conf parameter to disable it ;) Changing the internal state to what we think may happen can go wrong on so many levels.
> 
> e.g.
> 
> * if the service call does fail
> * HA entity does control a device with lockout
> * the device has special states that happen on service calls
> 
> the HA API will tell us the correct state as soon that the API call is successful no need for any guessing game.

We want optimistic updates because the entity states can be shown and edited on a menu on the ESP and we don't want the number to jump around. 

For example volume control, light brightness / temp, numbers can be changed immediately, but it might take Home Assistant a full second to return the value. In that time the user can change the state to another value, so we want to update Home Assistant and ignore it

In your example config you should be able to see this happen if you shorten the button delays and change the state quickly

I don't want to block this because my use case, so it can be added later! But what do you think about making a shared class for entity_id, etc like this

```
class HomeAssistantComponent : public Component, public api::CustomAPIDevice {
 public:
  void setup() override{};
  void set_entity_id(const std::string &entity_id) { entity_id_ = entity_id; }
  std::string get_entity_id() { return entity_id_; }
  float get_setup_priority() const override { return setup_priority::LATE; }

 protected:
  std::string entity_id_;
};
```
https://github.com/esphome/esphome/pull/6455/files#diff-6556fc4fcd0ceba3865f509d43f440358ace8c42fd1e25394302ebd74a714ec7R1

Links2004:
I dont see a problem with a shared class for all homeassistant sub components.
will makes thinks easier in the future.

at least in my cases there is no delay to see, the log shows the info all in the same second.
and pressing fast did not result in any problems, at on point I got faster then what my actors allow but thats it.
the state where always good.

what hardware are you running homeassistant on, may the device is to slow?

the button code is in the PR description as example.

```
========= Button pressed
[18:08:47][D][binary_sensor:036]: 'Boot Button': Sending state ON
========= Button release
[18:08:47][D][binary_sensor:036]: 'Boot Button': Sending state OFF
========= service call
[18:08:47][D][cover:076]: 'my_cool_cover' - Setting
[18:08:47][D][cover:084]:   Position: 0%
========= state update via API
[18:08:47][D][homeassistant.cover:033]: 'cover.eg_schlafzimmer_rollo': Got state closing
[18:08:47][D][cover:170]: 'my_cool_cover' - Publishing:
[18:08:47][D][cover:173]:   Position: 96%
[18:08:47][D][cover:186]:   Current Operation: CLOSING
========= more state update via API
....
========= state update via API
18:10:06][D][homeassistant.cover:033]: 'cover.eg_schlafzimmer_rollo': Got state closing
[18:10:06][D][cover:170]: 'my_cool_cover' - Publishing:
[18:10:06][D][cover:173]:   Position: 94%
[18:10:06][D][cover:186]:   Current Operation: CLOSING
========= Button pressed
[18:10:07][D][binary_sensor:036]: 'Boot Button': Sending state ON
====== update during press and release happend 
[18:10:07][D][homeassistant.cover:048]: 'cover.eg_schlafzimmer_rollo': Got position 93.00%
[18:10:07][D][cover:170]: 'my_cool_cover' - Publishing:
[18:10:07][D][cover:173]:   Position: 93%
[18:10:07][D][cover:186]:   Current Operation: CLOSING
========= Button release
[18:10:08][D][binary_sensor:036]: 'Boot Button': Sending state OFF
========= service call
[18:10:08][D][cover:076]: 'my_cool_cover' - Setting
[18:10:08][D][cover:080]:   Command: STOP
========= state update via API
[18:10:08][D][homeassistant.cover:033]: 'cover.eg_schlafzimmer_rollo': Got state open
[18:10:08][D][cover:170]: 'my_cool_cover' - Publishing:
[18:10:08][D][cover:173]:   Position: 93%
[18:10:08][D][cover:186]:   Current Operation: IDLE
```

In my opinion its fine to having optimistic later on as a option, 
but I think the default should always be none optimistic, since that is how other components do it.

sames goes for a "x second" ignore since this is not what a user expects by default ;)
And the ignore can have a lot of edge cases where you easy have the wrong state internally, 
since the HA API sends each change only once.
for this reason its basically a necessity that the users opt in to this since they need to be aware of it,
this will result in a lot of open issues other wise ;)

But this is only my opinion, happy to hear what other devs think.

landonr:
@Links2004 that timing looks good, you're right it shouldn't be an issue :D I'll update #6455 with the unknown/unavailable code tonight, can you take a look after?

Links2004:
I hate to ping people but any feedback from the esphome Team for thes PR is highly appreciated.

landonr:
Sorry for the delay - been afk, i'll try to check this out this week!

jesserockz (changes requested):
A quick review shows way too many different files changed for this. 

This PR should be split up into the different parts.

I also think the unavailable state handling etc should go in `homeassistant` top level folder only as it is only useful for the platforms of it.

Also see the recent addition that uses `get_home_assistant_state` instead of `get_home_assistant_state` to get things like the supported features attribute for parsing and setting the traits correctly.

Links2004:
@jesserockz

I can move the `EntityBase_State::handle_state(const char *input)` func to `homeassistant` but I am not sure If that is what is requested.

The EntityBase_State base class makes sense for me in a lot of cases other then the homeassistant use case.
e.g.  any component with external comunication (e.g. BLE, Modbus, ...),
there is also a feature-requests for it [here](https://github.com/esphome/feature-requests/issues/1568) where
this class can be used a starting point (the comunication back to HA is missing currently).
since the first step is to have a generic way to track the unavailable and unknown state at all.
currently there is only the `NaN` case with numbers but this does not work for any other data type.

In the long run it may is a good idea to have this all in the `EntityBase`.
But a deticated class makes it very easy for the  components to indecate if they support unknown/unavailable states. 
Directly integrating in to EntityBase may create issues like its not working for X.
The compiler will complian with a dedicated class.

For splitting up I think of the following PRs:

1. EntityBase_State implementation with triggers and conditions
2. updating exsisting homeassistant components
3. adding cover

I can do the split up if this sounds good.

PR #7019, opened 2024-06-30
Enable devcontainer linters

orland0m:

* Devcontainer improvements

## Types of changes

- [] Bugfix (non-breaking change which fixes an issue)
- [x] Other - devcontainer linters

## Test VS devcontainer
Tested locally with dev container:
* Ran negative test with flake8/pylint by editing config and confirming errors/warnings show up (screenshots below).
* Tested black auto format on save (video below).

https://github.com/esphome/esphome/assets/3383373/a25af5e4-43e4-45ea-a559-761e2f975130

flake8_negative_test
pylint_negative_test

## Checklist:
  - [N/A] Tests have been added to verify that the new code works (under `tests/` folder).

  - [N/A] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,9 @@
 {
   "name": "ESPHome Dev",
   "image": "ghcr.io/esphome/esphome-lint:dev",
-  "postCreateCommand": ["script/devcontainer-post-create"],
+  "postCreateCommand": [
+    "script/devcontainer-post-create"
+  ],
   "containerEnv": {
     "DEVCONTAINER": "1",
     "PIP_BREAK_SYSTEM_PACKAGES": "1",
@@ -27,6 +29,9 @@
       "extensions": [
         // python
         "ms-python.python",
+        "ms-python.pylint",
+        "ms-python.flake8",
+        "ms-python.black-formatter",
         "visualstudioexptteam.vscodeintellicode",
         // yaml
         "redhat.vscode-yaml",
@@ -38,9 +43,21 @@
       "settings": {
         "python.languageServer": "Pylance",
         "python.pythonPath": "/usr/bin/python3",
-        "python.linting.pylintEnabled": true,
-        "python.linting.enabled": true,
-        "python.formatting.provider": "black",
+        "pylint.args": [
+          "--rcfile=${workspaceFolder}/pyproject.toml"
+        ],
+        "flake8.args": [
+          "--config=${workspaceFolder}/.flake8"
+        ],
+        "black-formatter.args": [
+          "--config",
+          "${workspaceFolder}/pyproject.toml"
+        ],
+        "[python]": {
+          // VS will say "Value is not accepted" before building the devcontainer, but the warning
+          // should go away after build is completed.
+          "editor.defaultFormatter": "ms-python.black-formatter"
+        },
         "editor.formatOnPaste": false,
         "editor.formatOnSave": true,
         "editor.formatOnType": true,


docker/Dockerfile
```diff
@@ -34,31 +34,28 @@ RUN \
...
           libopenjp2-7=2.5.0-2 \
           libtiff6=4.5.0-6+deb12u1 \
           cargo=0.66.0+ds1-1 \
           pkg-config=1.8.1-1 \
           gcc-arm-linux-gnueabihf=4:12.2.0-3; \
-    fi; \
-    rm -rf \
-        /tmp/* \
-        /var/{cache,log}/* \
-        /var/lib/apt/lists/*
```
orland0m: Removed this cleanup section from `base` because the distribution lists are required later by child images. Instead, I made sure child images had this cleanup step at the end.

Reference: https://community.fly.io/t/e-unable-to-locate-package-python3-pip/18797/2

docker/Dockerfile
```diff
@@ -34,31 +34,28 @@ RUN \
...
         iputils-ping=3:20221126-1 \
         git=1:2.39.2-1.1 \
-        curl=7.88.1-10+deb12u5 \
+        curl=7.88.1-10+deb12u6 \
         openssh-client=1:9.2p1-2+deb12u2 \
         python3-cffi=1.15.1-5 \
         libcairo2=1.16.0-7 \
         libmagic1=1:5.44-3 \
-        patch=2.7.6-7; \
+        patch=2.7.6-7
```
orland0m: Prior `RUN` script would silently fail due to usage of semicolon, decided to split command but let me know if `&&` is preferred instead.

Reference: https://unix.stackexchange.com/questions/187145

.devcontainer/devcontainer.json
```diff
@@ -38,9 +43,21 @@
       "settings": {
         "python.languageServer": "Pylance",
         "python.pythonPath": "/usr/bin/python3",
-        "python.linting.pylintEnabled": true,
```
orland0m, line 41: These settings are deprecated in favor of plugins

Ref: https://stackoverflow.com/questions/77498238

docker/Dockerfile
```diff
@@ -137,8 +134,11 @@ ENTRYPOINT ["/entrypoint.sh"]
 # When no arguments given, start the dashboard in the workdir
 CMD ["dashboard", "/config"]
 
-
-
+RUN \
+    rm -rf \
+    /tmp/* \
+    /var/{cache,log}/* \
+    /var/lib/apt/lists/*
```
jesserockz: The reason this is in the same `RUN` step is that it has no effect on the image size if you run it outside the step the files are added.
This is the way docker works (and it is annoying but we have to work with that limitation)

jesserockz (changes requested):
Thanks for the PR.

Could you please make this PR only have a single purpose - the devcontainer linters?

I am going to bump the other dependencies too in my other PR, then look at fixing the actual dockerfile itself to not hide the errors. It has hit a few times in the past too and I just have not gotten to fixing it yet.

jesserockz:
I have also reimplemented the RUN command in #7021 (hopefully it works, I cant run armv7 images locally for some reason today.)

Thanks for the push to fix it =)

docker/Dockerfile
orland0m: > RE: Could you please make this PR only have a single purpose - the devcontainer linters?

Done

> RE: then look at fixing the actual dockerfile itself to not hide the errors. It has hit a few times in the past too and I just have not gotten to fixing it yet.

Got it, I saw a different PR with successful checks and I assumed I had broken the builds :D

orland0m:
Yeah, I realize the change

jesserockz (approved):

PR #7027, opened 2024-07-01
Make crc8 const-correct

nattgris:

The crc8 helper doesn't and shouldn't modify its data argument, so it should be declared const, like in the other crc helpers.

Without it, external components using the Dallas CRC8 fail to compile if calling crc8() on a const buffer. 

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/core/helpers.cpp
+++ b/esphome/core/helpers.cpp
@@ -93,7 +93,7 @@ std::string to_string(long double value) { return str_snprintf("%Lf", 32, value)
 // Mathematics
 
 float lerp(float completion, float start, float end) { return start + (end - start) * completion; }
-uint8_t crc8(uint8_t *data, uint8_t len) {
+uint8_t crc8(const uint8_t *data, uint8_t len) {
   uint8_t crc = 0;
 
   while ((len--) != 0u) {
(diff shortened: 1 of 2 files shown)


nielsnl68:
@nattgris, a quick look around in the esphome codebase i see that it looks like the helper CRC8 function is being used in multiple components and not only with the Dallas components.

Did you have a look where this function is used as well and if this function still compiles with your suggested changes?

nattgris:
@nielsnl68, as far as I can see it's just used in `dallas_temp.cpp` and `one_wire_bus.cpp` in tree. There are several more uses of other CRC8 algorithms, but not this 1-Wire variant in `helpers.cpp`.

Anyway, a _pointer-to-T_ is convertible to _pointer-to-const-T_ so just adding `const` like this to a parameter cannot break compilation. At least not in plain old C, I'm not particularly good at C++.

I thought the CI would build all components, but I don't see it in the checks. The affected components build locally at least.

jesserockz (approved):

PR #7028, opened 2024-07-01
Tuya integration no longer waits for wifi

szupi-ipuzs:

In the original implementation Tuya integration setup priority was LATE, which means it wasn't started until after wifi was connected. This was probably because the integration author wanted real wifi state to be sent to Tuya mcu.
There are 2 problems with this:
1) It might take a long time for the real wifi to connect - and during this time tuyamcu expects the module to respond to notifications. If there is no response - tuyamcu assumes errornous state and might require a reset.
2) There is really no point to forward real wifi state to tuya mcu. These devices were designed to be controlled via cloud and some of them even block their functionality until the cloud is connected. The esphome users might want to use these devices even without connecting them to wifi. Why not let them?

TLDR:
This PR moves the initialization of Tuya integration from LATE to DATA (before wifi) and fakes the network status when asked by Tuya MCU.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/tuya/tuya.cpp
+++ b/esphome/components/tuya/tuya.cpp
@@ -1,17 +1,8 @@
 #include "tuya.h"
-#include "esphome/components/network/util.h"
 #include "esphome/core/gpio.h"
 #include "esphome/core/helpers.h"
 #include "esphome/core/log.h"
 #include "esphome/core/util.h"
-
-#ifdef USE_WIFI
-#include "esphome/components/wifi/wifi_component.h"
-#endif
-
-#ifdef USE_CAPTIVE_PORTAL
-#include "esphome/components/captive_portal/captive_portal.h"
-#endif
 
 namespace esphome {
 namespace tuya {
@@ -20,6 +11,10 @@ static const char *const TAG = "tuya";
 static const int COMMAND_DELAY = 10;
 static const int RECEIVE_TIMEOUT = 300;
 static const int MAX_RETRIES = 5;
+
+static const uint8_t NET_STATUS_WIFI_CONNECTED = 0x03;
+static const uint8_t NET_STATUS_CLOUD_CONNECTED = 0x04;
+static const uint8_t FAKE_WIFI_RSSI = 100;
 
 void Tuya::setup() {
   this->set_interval("heartbeat", 15000, [this] { this->send_empty_command_(TuyaCommandType::HEARTBEAT); });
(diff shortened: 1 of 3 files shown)


jesserockz (changes requested):
The whole point of this was to relay the real wifi/api status to the device. I personally have a tuya mcu device that has a wifi indicator and I would prefer that it show me the real status, and not just fake it.

szupi-ipuzs:
Well, then let's make this configurable somehow. Although I don't know if setup priority can be based on yaml config?
Anyway, I think keeping the implementation as it is means sacrificing support for many potential devices for the sake of seeing the real indicator status. Don't you agree?

nagyrobi:
You could just fake it for the time it takes to move from DATA to LATE. When reached final state, show real info....

szupi-ipuzs:
> You could just fake it for the time it takes to move from DATA to LATE. When reached final state, show real info....

But if you tell tuyamcu it is not connected to the cloud anymore, it might stop working.
No, I think it's better to make this configurable (either fake the network status or not).

nagyrobi:
```
if (connnected) {
  send real data;
} else {
  fake it as connected;
}
```

szupi-ipuzs:
So, in other words - always send "connected" state, right? How is it different to what I did?
And this also means jessierockz's need to see the real state (eg. disconnected) is not fulfilled.

jesserockz:
We can just change the setup priority to be earlier on, and it will send offline until it is actually online which is then 100% real, no faking required and mimicks how it is on a new device out of the box. The original firmware wouldnt have been connected to wifi/cloud faster than the secondary mcu boots up so it should be able to handle not being connected to wifi, otherwise why implement this command/result.

szupi-ipuzs:
> We can just change the setup priority to be earlier on, and it will send offline until it is actually online which is then 100% real, no faking required and mimicks how it is on a new device out of the box. The original firmware wouldnt have been connected to wifi/cloud faster than the secondary mcu boots up so it should be able to handle not being connected to wifi, otherwise why implement this command/result.

@jesserockz with all due respect - trying to mimmick the original tuya firmware behavior is usually not why I have flashed my devices to esphome.
A tuyamcu device might not be operable until the "connected" status is reported. And we might want to do some automation even before wifi, based on the data reported via datapoints (like eg. close the valve, turn on/off light, etc.), 
All I'm saying is - let's give users a choice. My proposal: allow faking the wifi status __optionally__, eg. via an additional yaml key which would be false by default:

```yaml
tuya:
  fake_wifi_status: True
```
(or please propose a better name).

If you agree, I will update this PR so that the tuya component:
1) Reads the new yaml key and based on the setting:
2a) Waits for wifi/mqtt before reporting "wifi connected" to tuyamcu.
or
2b) Reports "wifi ready" immediately after tuyamcu initialization.

In both cases I think it would be beneficial to move the setup priority as early as possible (to DATA) to speed up the initialization.

ssieb:
Have you seen any device that didn't function if the wifi wasn't connected yet?

szupi-ipuzs:
Not personally to be honest.
But if you read through the elektroda iot forum where they routinely install openbeken on tuya devices, they always recommend to fake the wifi state. You can [read the explanation here](https://www.elektroda.com/rtvforum/viewtopic.php?p=20975937#20975937), let me quote the appropriate part:

`
Do not forget the tuyaMcu_defWiFiState 4 line. By default, OBK sends "paired" state (0x04) only when MQTT is on, but some devices always require the WiFi state to be "paired" (0x04) before sending data. This will control the WiFi LED of the MCU and even may control buzzer. Some devices will send buzzer audio signals when not paired. 
`

PR #7032, opened 2024-07-02
[micro_wake_word] Version 2

kahrendt:

Adds support for microWakeWord version 2 models.

- Adds ops to support more efficient and accurate 'MixedNet' models which can run on an ESP32
- Adds support for running multiple models at once
- Adds support for running a Voice Activity Model to reduce false activations from non-speech noises
- Moved spectrogram feature generation to code in C for a large speed improvement using ESPMicroSpeechFeatures: 
  - Source Code: https://github.com/kahrendt/ESPMicroSpeechFeatures/releases/tag/v1.0.0
  - PlatformIO registry: https://registry.platformio.org/libraries/kahrendt/ESPMicroSpeechFeatures
- Models load and unload as mWW is started and stopped to reduce memory usage
- Update the model manifest to version 2 to support setting the spectrogram feature step size and setting the amount of memory to allocate for the tensor arena.

The old microWakeWord models will still work with this version of the component, but they cannot run simultaneously with the new models. It is a breaking change as the YAML configuration has changed to support listing multiple models.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4015

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Ensure you are running the ESPHome dev branch!

```yaml
# Example config.yaml
external_components:
  - source:
      type: git
      url: https://github.com/kahrendt/esphome
      ref:  mww-v2-external-library
    refresh: 0s
    components: [ micro_wake_word ]  

micro_wake_word:
  on_wake_word_detected:
    - voice_assistant.start: 
  vad:
    model: https://github.com/kahrendt/microWakeWord/releases/download/v2.1_models/vad.json
  models:
    - model: https://github.com/kahrendt/microWakeWord/releases/download/v2.1_models/okay_nabu.json
    - model: https://github.com/kahrendt/microWakeWord/releases/download/v2.1_models/alexa.json
    - model: https://github.com/kahrendt/microWakeWord/releases/download/v2.1_models/hey_jarvis.json
    - model: https://github.com/kahrendt/microWakeWord/releases/download/v2.1_models/hey_mycroft.json


```

## Checklist:


esphome/components/micro_wake_word/__init__.py
```diff
@@ -167,10 +214,21 @@ def _process_http_source(config):
     if not isinstance(manifest_data, dict):
         raise cv.Invalid("Manifest file must contain a JSON object")
 
-    try:
-        MANIFEST_SCHEMA_V1(manifest_data)
-    except cv.Invalid as e:
-        raise cv.Invalid(f"Invalid manifest file: {e}") from e
+    if manifest_version := manifest_data["version"]:
```
jesserockz: We should use the const here, and also change to `.get` in case it is not in the file as we dont want to raise an unhandled key error
```suggestion
    if manifest_version := manifest_data.get(KEY_VERSION):
```

esphome/components/micro_wake_word/__init__.py
```diff
@@ -282,16 +368,30 @@ def _load_model_data(manifest_path: Path):
...
+                raise cv.Invalid(f"Invalid manifest file: {e}") from e
+        elif manifest_version == 2:
+            try:
+                MANIFEST_SCHEMA_V2(manifest)
+            except cv.Invalid as e:
+                raise cv.Invalid(f"Invalid manifest file: {e}") from e
+        else:
+            raise cv.Invalid("Invalid manifest version")
+    else:
+        raise cv.Invalid("Invalid manifest file, missing 'version' key.")
```
jesserockz: Lets combine this with the above so we dont have a large chunk of duplicate code

esphome/components/micro_wake_word/__init__.py
```diff
@@ -260,18 +318,46 @@ def _validate_source_shorthand(value):
...
         {
             cv.GenerateID(): cv.declare_id(MicroWakeWord),
             cv.GenerateID(CONF_MICROPHONE): cv.use_id(microphone.Microphone),
-            cv.Optional(CONF_PROBABILITY_CUTOFF): cv.percentage,
-            cv.Optional(CONF_SLIDING_WINDOW_AVERAGE_SIZE): cv.positive_int,
+            cv.Required(CONF_MODELS): cv.ensure_list(MODEL_SCHEMA),
             cv.Optional(CONF_ON_WAKE_WORD_DETECTED): automation.validate_automation(
                 single=True
             ),
-            cv.Required(CONF_MODEL): MODEL_SOURCE_SCHEMA,
```
jesserockz, line 273: Please leave removed keys in the schema with a validation of `cv.invalid("This has moved... etc")`

esphome/components/micro_wake_word/__init__.py
```diff
@@ -319,42 +412,83 @@ async def to_code(config):
...
+                    manifest[KEY_MICRO][CONF_TENSOR_ARENA_SIZE],
+                )
+            )
+
+        model_step_size = manifest[KEY_MICRO][CONF_FEATURE_STEP_SIZE]
+
+        if features_step_size is None:
+            features_step_size = model_step_size
+        elif features_step_size != model_step_size:
+            raise cv.Invalid("Cannot load models with differet features step sizes.")
```
jesserockz: We cannot raise cv.Invalid errors inside `to_code`. This should be checked in the validation stage.
Also a typo =)

```suggestion
            raise cv.Invalid("Cannot load models with different features step sizes.")
```

esphome/components/micro_wake_word/micro_wake_word.cpp
```diff
@@ -55,58 +56,55 @@ static const LogString *micro_wake_word_state_to_string(State state) {
...
-  ESP_LOGCONFIG(TAG, "  Wake Word: %s", this->get_wake_word().c_str());
-  ESP_LOGCONFIG(TAG, "  Probability cutoff: %.3f", this->probability_cutoff_);
-  ESP_LOGCONFIG(TAG, "  Sliding window size: %d", this->sliding_window_average_size_);
+  ESP_LOGCONFIG(TAG, "microWakeWord models:");
+  for (auto &model : this->wake_word_models_) {
+    model.log_model_config();
+  }
+#ifdef USE_MWW_VAD
+  this->vad_model_->log_model_config();
+#endif
```
jesserockz:
```suggestion
  ESP_LOGCONFIG(TAG, "microWakeWord:");
  ESP_LOGCONFIG(TAG, "  models:");
  for (auto &model : this->wake_word_models_) {
    model.log_model_config();
  }
#ifdef USE_MWW_VAD
  this->vad_model_->log_model_config();
#endif
```

esphome/components/micro_wake_word/__init__.py
```diff
@@ -319,42 +412,83 @@ async def to_code(config):
...
+        )
 
-    elif model_config[CONF_TYPE] == TYPE_LOCAL:
-        file = Path(model_config[CONF_PATH])
+    features_step_size = None
 
-    elif model_config[CONF_TYPE] == TYPE_HTTP:
-        file = _compute_local_file_path(model_config) / "manifest.json"
+    if vad_model := config.get(CONF_VAD):
+        cg.add_define("USE_MWW_VAD")
```
jesserockz: I know its long, but we need to keep consistency and use the actual component name =)
Also please add it to `esphome/core/defines.h` in the esp32 section so it is set during development and CI.
```suggestion
        cg.add_define("USE_MICRO_WAKE_WORD_VAD")
```

esphome/components/micro_wake_word/streaming_model.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+static const char *const TAG = "micro_wake_word";
+
+namespace esphome {
+namespace micro_wake_word {
+
+void WakeWordModel::log_model_config() {
+  ESP_LOGCONFIG(TAG, "  - Wake Word: %s", this->wake_word_.c_str());
+  ESP_LOGCONFIG(TAG, "    Probability cutoff: %.3f", this->probability_cutoff_);
+  ESP_LOGCONFIG(TAG, "    Sliding window size: %d", this->sliding_window_size_);
```
jesserockz: I split the logconfig text in the main component and prefixed "models" so these should also move over.
```suggestion
  ESP_LOGCONFIG(TAG, "    - Wake Word: %s", this->wake_word_.c_str());
  ESP_LOGCONFIG(TAG, "      Probability cutoff: %.3f", this->probability_cutoff_);
  ESP_LOGCONFIG(TAG, "      Sliding window size: %d", this->sliding_window_size_);
```

esphome/components/micro_wake_word/streaming_model.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+void VADModel::log_model_config() {
+  ESP_LOGCONFIG(TAG, "  - VAD Model");
+  ESP_LOGCONFIG(TAG, "    Probability cutoff: %.3f", this->probability_cutoff_);
+  ESP_LOGCONFIG(TAG, "    Sliding window size: %d", this->sliding_window_size_);
```
jesserockz:
```suggestion
  ESP_LOGCONFIG(TAG, "    - VAD Model");
  ESP_LOGCONFIG(TAG, "      Probability cutoff: %.3f", this->probability_cutoff_);
  ESP_LOGCONFIG(TAG, "      Sliding window size: %d", this->sliding_window_size_);
```

esphome/components/micro_wake_word/streaming_model.h
```diff
@@ -0,0 +1,93 @@
...
+                const std::string &wake_word, size_t tensor_arena_size);
+
+  void log_model_config() override;
+
+  /// @brief Checks for the wake word by comparing the mean probability in the sliding window with the probability
+  /// cutoff
+  /// @return True if wake word is detected, false otherwise
+  bool determine_detected() override;
+
+  std::string get_wake_word() { return this->wake_word_; }
```
jesserockz:
```suggestion
  const std::string &get_wake_word() const { return this->wake_word_; }
```

jesserockz (changes requested):
Here is my initial review. I have not looked at everything yet.

We have the tflite component loaded now in CI, so `#ifndef CLANG_TIDY` can be removed from the source files.

esphome/components/micro_wake_word/__init__.py
kahrendt: I've created a new function to validate this and just set ``FINAL_VALIDATE_SCHEMA = _feature_step_size_validate`` I think this is the right approach from looking at other components, but I'm not 100% sure.

esphome/components/micro_wake_word/__init__.py
kahrendt: I get lazy sometimes :)

kahrendt:
> Here is my initial review. I have not looked at everything yet.
> 
> We have the tflite component loaded now in CI, so `#ifndef CLANG_TIDY` can be removed from the source files.

Thanks for the initial review! I made those changes and fixed the things clang tidy was complaining about:
- add final specifier to the ``WakeWordModel`` and ``VADModel`` classes (I'm not 100% this is the best way to handle the virtual destructor issue)
- use unique_ptr for the VAD model and the TFLite interpreter
- use constexpr in the ``generate_features_for_window_`` function
- use ``emplace_back`` instead of ``push_back`` when adding WakeWordModels

jesserockz (approved):

jesserockz (changes requested):
Forgot we should change the shorthand url builder to have the v2 directory
https://github.com/esphome/esphome/pull/7032/files#diff-7dd4ab6d5bf5e2bc507a1a67e15a8aa9b206c8d285bbe47be228a71f0de749c2R271

jesserockz (approved):

PR #7037, opened 2024-07-03
Don't test for IPv6 addresses when min_ipv6_addr_count is 0

colmbuckley:
Alter the `#if` slightly here to avoid this test when `min_ipv6_addr_count` is zero.

Avoids the trivial comparison when the number of required IPv6 addresses is set to zero (eliminating a source of compiler warnings).

As things stand, if IPv6 is enabled but `min_ipv6_addr_count == 0` (the default), we are here testing whether a `uint8_t` value is greater or equal to zero here. This is of course always true, so the compiler emits a warning (due to `-Wtype-limits`). Adding a test to the `#ifdef` to avoid making this comparison when the RHS is zero.

Trivially tested.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/wifi/wifi_component_esp_idf.cpp
+++ b/esphome/components/wifi/wifi_component_esp_idf.cpp
@@ -757,7 +757,7 @@ void WiFiComponent::wifi_process_event_(IDFWiFiEvent *data) {
 
 WiFiSTAConnectStatus WiFiComponent::wifi_sta_connect_status_() {
   if (s_sta_connected && this->got_ipv4_address_) {
-#if USE_NETWORK_IPV6
+#if USE_NETWORK_IPV6 && (USE_NETWORK_MIN_IPV6_ADDR_COUNT > 0)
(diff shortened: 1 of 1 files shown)


colmbuckley:
It's probable that similar changes should be made in `components/ethernet/ethernet_component.cpp`, 
`components/wifi/wifi_component_esp32_arduino.cpp` and 
`components/wifi/wifi_component_esp8266.cpp` but these don't correspond to my environment so I can't easily test them. A scan of the code suggests that the corresponding changes should be fairly trivial and low-impact.

HeMan (commented):
This should work fine. I'm looking at how and if this is applicable to the other environments as well.

jesserockz (approved):

PR #7044, opened 2024-07-04
jsn_sr04t component: AJ_SR04M compatibility mode in checksum calculation

### LINKED ISSUE #2784: jsn_sr04t component: AJ_SR04M compatibility mode
The AJ_SR04M ultarsonic module is very similar to the JSN SR04T and has similar operation modes.

There is only one differences in the protocol, which is the checksum calculation that does not include the first byte.

Instead of adding a new component, would it be possible to add a configuration variable (e.g. `ajsq04m_mode`, which defaults to false), which, if set to true, changes the checksum calculation from

`uint8_t checksum = this->buffer_[0] + this->buffer_[1] + this->buffer_[2];`
to
`uint8_t checksum = this->buffer_[1] + this->buffer_[2];`
?  

see above

**Additional context**

see above

soeffi:

The AJ_SR04M ultarsonic module is very similar to the JSN SR04T and has similar operation modes.

There is only one differences in the protocol, which is the checksum calculation that does not include the first byte.

This PR adds a configuration option to use the alternative checksum calculation mode.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2784

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4022

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: jsn_sr04t
    ajsr04m_mode: true
```

## Checklist:


soeffi:
I did my best to compile a PR. Be kind, this is my first PR ever.

Let me know what you think and I'll work on the open checklist items

Mafus1:
@soeffi I just looked at the PR, and it looks good to me! I also tested it on my JSN-RT04T and it behaves as expected. It would be great if you could also update the documentation, so other people know how to use this option, too.

soeffi:
> It would be great if you could also update the documentation

Done, added PR for docs as well

esphome/components/jsn_sr04t/sensor.py
```diff
@@ -25,6 +26,11 @@
     )
     .extend(cv.polling_component_schema("60s"))
     .extend(uart.UART_DEVICE_SCHEMA)
+    .extend(
+        {
+            cv.Optional(CONF_AJSR04M_MODE, default=False): cv.boolean,
```
jesserockz: Lets change this to an enum with `CONF_MODEL` with the default being "jsn_sr04t"

jesserockz (changes requested):

esphome/components/jsn_sr04t/sensor.py
soeffi: > Lets change this to an enum with `CONF_MODEL` with the default being "jsn_sr04t"

Done, PR for code and docs updated. Could you please review and let me know if that works?

esphome/components/jsn_sr04t/jsn_sr04t.h
```diff
@@ -20,7 +25,7 @@ class Jsnsr04tComponent : public sensor::Sensor, public PollingComponent, public
 
  protected:
   void check_buffer_();
-  bool ajsr04m_;
+  Model model_ = JSN_SR04T;
```
jesserockz:
```suggestion
  Model model_;
```

esphome/components/jsn_sr04t/jsn_sr04t.cpp
```diff
@@ -52,8 +58,16 @@ void Jsnsr04tComponent::check_buffer_() {
...
-                this->ajsr04m_ ? "aj_sr04m (exclude start byte)" : "jsn_sr04t (include start byte)");
+  switch (this->model_) {
+    case JSN_SR04T:
+      ESP_LOGCONFIG(TAG, "  sensor model: jsn_sr04t");
+      break;
+    case AJ_SR04M:
+      ESP_LOGCONFIG(TAG, "  sensor model: aj_sr04m");
+      break;
+    default:
+      ESP_LOGE(TAG, "  sensor model undefined, set 'model' option in YAML");
```
jesserockz: It will always be set, no need to handle default.
```suggestion
```

jesserockz (changes requested):
Changes look good. Just a couple of things below

Mafus1:
@soeffi, the changes look good to me! I tested it and it behaves as expected. I like the option with the enum to extend it to further models in the future.
@jesserockz made two comments about small improvements, maybe you can apply them, and then I think this should be ready to merge.

esphome/components/jsn_sr04t/jsn_sr04t.cpp
soeffi: Typically, I'd try to check as much as possible, just in case. However, for embedded devices, it's probably better to reduce code size...

Mafus1 (approved):

jesserockz (approved):

esphome/components/jsn_sr04t/jsn_sr04t.cpp
jesserockz: Yes its good to cover everything. Our CI will fail if all switch cases are not handled for this specifically though.

PR #7056, opened 2024-07-08
Add support for the Gree YAC1FB9 in climate_ir

### LINKED ISSUE #6024: Gree climate_ir component doesn't work with YAC1FB9
### The problem

The Gree YAC1FB9 heat air pump doesn't directly work with the "gree" climate_ir component (configured as "yac" or "yaa"), as it seems to require slightly different timings (4500 vs 4000 for header space, and 19980 vs 19000 for message space).

2024.6.6

Home Assistant Add-on

2024.7.1

ESP32

### Board

esp32cam

gree

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

topeju:
This model appears to be more similar to the YAA than YAC (it doesn't have horizontal swing), but also requires slightly different space timings. (It does, however, have I-Feel, though I've not implemented this support and the existing gree component also doesn't seem to support it.)

Adds a new model "yac1fb9" for the "gree" climate_ir component

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/6024

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4029

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
climate:
  - platform: gree
    name: "Gree/Alpicair heat air pump"
    id: gree_climate
    model: yac1fb9
    transmitter_id: gree_ir_xmit
    sensor: room_temp
    supports_cool: true
    supports_heat: true

remote_transmitter:
  pin: 13
  id: gree_ir_xmit
  carrier_duty_percent: 50%
```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs). Pull request: esphome/esphome-docs#4029

### DIFF
--- a/esphome/components/gree/climate.py
+++ b/esphome/components/gree/climate.py
@@ -16,6 +16,7 @@ MODELS = {
     "yan": Model.GREE_YAN,
     "yaa": Model.GREE_YAA,
     "yac": Model.GREE_YAC,
+    "yac1fb9": Model.GREE_YAC1FB9,
 }
 
 CONFIG_SCHEMA = climate_ir.CLIMATE_IR_WITH_RECEIVER_SCHEMA.extend(
--- a/esphome/components/gree/gree.cpp
+++ b/esphome/components/gree/gree.cpp
@@ -24,7 +24,7 @@ void GreeClimate::transmit_state() {
     remote_state[4] |= (this->horizontal_swing_() << 4);
   }
 
-  if (this->model_ == GREE_YAA || this->model_ == GREE_YAC) {
+  if (this->model_ == GREE_YAA || this->model_ == GREE_YAC || this->model_ == GREE_YAC1FB9) {
     remote_state[2] = 0x20;  // bits 0..3 always 0000, bits 4..7 TURBO,LIGHT,HEALTH,X-FAN
     remote_state[3] = 0x50;  // bits 4..7 always 0101
     remote_state[6] = 0x20;  // YAA1FB, FAA1FB1, YB1F2 bits 4..7 always 0010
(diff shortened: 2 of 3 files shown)


topeju:
This is a very straightforward way to implement the new model, but it might not be a bad idea to refactor the component a bit. For example, it might be beneficial to allow changing the timing values though the yaml configuration?

nagyrobi:
Does it support toggling the light/xfan/turbo/health bits the same as `yaa`?

topeju:
At least the light bit is supported (if it's the one toggling the temperature display on and off), I'm not quite sure about xfan/turbo -- will check all of these tonight. (Edit 2024-07-10: Forgot to do so yesterday, will try to remember to do that tonight.)

nagyrobi:
Those bits apparently work for all supported models, as i've just tried:
https://github.com/nagyrobi/esphome/blob/gree-with-extras/esphome/components/gree/gree.cpp

My goal would be to have later a switch platform under `gree` to expose them as switches like
```yaml
switch:
  - platform: gree
    turbo:
      name: Room AC Turbo
    light:
      name: Room AC Light
    health:
      name: Room AC Ionizer
    xfan:
      name: Room AC Econo
```

topeju:
Ah, right, of course: those bits (light/turbo/health/x-fan) aren't controllable in the current implementation, and this PR is not really about adding support for them either. As far as I know, the bits themselves do work. `remote_state[2]` is just hard-coded to the value 0x20, which sets the light bit on, i.e. turns the goal temperature display on.

jesserockz (approved):

PR #7049, opened 2024-07-06
[nrf52, core] nrf52 core based on zephyr

tomaszduda23:
# What does this implement/fix?

It requires:
- https://github.com/esphome/esphome/pull/8351
- https://github.com/esphome/esphome/pull/8352
- https://github.com/esphome/esphome/pull/7715
- https://github.com/esphome/esphome/pull/8139
- https://github.com/esphome/esphome/pull/8321
- https://github.com/esphome/esphome/pull/8412

The next steps:
- https://github.com/esphome/esphome/pull/8410
- https://github.com/esphome/esphome/pull/8319
- https://github.com/esphome/esphome/pull/8315
- https://github.com/esphome/esphome/pull/8150
- https://github.com/esphome/esphome/pull/7340
- https://github.com/esphome/esphome/pull/6075
- https://github.com/esphome/esphome/pull/8992
- https://github.com/esphome/esphome/pull/9319

Core part of https://github.com/esphome/esphome/pull/6075. The rest could be added as external components (with exception to ble tools for logging and flashing). Things which works:

- digital GPIO
- logger cdc/uart
- upload of mcuboot based firmware with SWD `python -m esphome upload tests/test12.2.yaml --device PYOCD` 
- upload of adafruit based firmware with CDC (after shorting reset pin to ground twice) `python -m esphome upload tests/test12.3.yaml`
- [x] test clock

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** 

- esphome/esphome-docs#4111

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] nrf52840

## Example entry for `config.yaml`:

```yaml
---
# based on mcuboot bootloader
external_components:
  - source: github://pr#7049
    components: [ logger, nrf52, time, zephyr ]
    refresh: always

nrf52:
  board: adafruit_feather_nrf52840

esphome:
  name: nrf52-test-nrf

logger:
  level: DEBUG

time:

switch:
  - platform: gpio
    pin:
      number: 15
      inverted: true
      mode:
        output: true
[...]

PR #7057, opened 2024-07-08
add way to exclude tests for given platform

tomaszduda23:

There is `test.all.yaml` which is great idea. This add way to exclude given platform from build. Example of usage: `tests/components/datetime/exclude.test.nrf52-mcumgr`. It is part of https://github.com/esphome/esphome/pull/7049

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


jesserockz:
I'm not sure we should have a need for this. Anywhere that has a `test.all.yaml` file should be supported on any and all platforms that ESPHome supports. 

I am guessing datetime is failing on not having a `time` platform available for the nrf?

At the very least this should not be a YAML file as this will never contain YAML, but simply be an empty marker file that simply exists.

tomaszduda23:
1. I didn't want to add more code to this PR. I assume it would be possible to fix compilation of time component.
2. I've noticed that after some changes which I made in the past some configuration stopped compiling even though CI was all green. I like the idea of `test.all.yaml`. Most components tests could be tested that way. Some kind of GPIO unification or aliases would be needed but it does not sounds very difficult. Than if needed `exclude` could be used to disable test for some platform. Current design requires adding explicit new target to all components which might be difficult to maintain over time.

I'm not sure if this is good place to discus the future strategy for testing though :smile:

tomaszduda23:
It looks that timezones are not supported by zephyr https://docs.zephyrproject.org/3.5.0/services/portability/posix.html. It would required using external library which I don't want to add  e.g. https://github.com/josmithua/utz. I will implement `time` component without timezone support for now. The easiest solution would be to add time offset and dst flag. It is not perfect either.

tomaszduda23:
@jesserockz would you mind to take a look again? Now the use cases look more clear.

PR #7066, opened 2024-07-10
components/host: remove -lsodium

ferbar:

@clydebarrow I think libsodium is not required
thx!

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] Host

## Checklist:


nagyrobi:
Isn't it required for the encrypted API?

clydebarrow:
> libsodium is not required

Yes, it is required when using encryption with `api:`. If you have a real problem with that, you'll need to come up with a way of only adding the option when required. Just removing it is not going to work.

ferbar:
I can confirm that it doesn't compile when encrypted api is enabled, Sorry for asking stupid questions

PR #7077, opened 2024-07-12
Update esp-idf i2c timeout unit conversion

dwhitlow:

When using ESP-IDF framework with the i2c component, setting any timeout value higher than zero on variants esp32c6 and esp32h2 results in the error `i2c timing value error`.

The I2C component integration with ESP-IDF multiplies the `timeout` value (in us) by 80 before passing it to `i2c_set_timeout()`. According to the [ESP-IDF documentation](https://docs.espressif.com/projects/esp-idf/en/v4.4.7/esp32/api-reference/peripherals/i2c.html#_CPPv415i2c_set_timeout10i2c_port_ti), the unit of the timeout value should be APB 80 MHz clock cycles. However, different ESP32 variants have different Advanced Peripheral Bus clock frequencies (ex: the [C6](https://github.com/espressif/esp-idf/blob/5ca9f2a49aaabbfaf726da1cc3597c0edb3c4d37/components/soc/esp32c6/include/soc/soc.h#L142) and [H2](https://github.com/espressif/esp-idf/blob/5ca9f2a49aaabbfaf726da1cc3597c0edb3c4d37/components/soc/esp32h2/include/soc/soc.h#L140) have 40 MHz and 32 MHz APB cloud frequencies respectively).

Testing with the C6, I noted that the timeout value returned by `i2c_get_timeout()` defaults to 10 (400 us @ 40 MHz). Further testing showed that the valid range of values that could be passed to `i2c_set_timeout()` were 1 to 31 (inclusive) (40 us - 1240 us).

This change updates the `i2c_set_timeout()` call to divide by the selected variant's `APB_CLK_FREQ` (in MHz) instead of multiplying by 80 MHz, allowing the timeout value specified in the ESPHome config to accurately reflect the value set on the ESP32.

This PR could alter the behavior of ESPHome projects that meet the following conditions:
* Using ESP-IDF framework
* Using ESP32
* Using `i2c` component with a non-zero value for `timeout`

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: "esp-idf-i2c-timeout-test"
  platformio_options:
    platform: "https://github.com/platformio/platform-espressif32/archive/refs/tags/v6.7.0.zip"

esp32:
  board: "esp32-c6-devkitm-1"
  variant: "esp32c6"
  flash_size: "4MB"
  framework:
    type: "esp-idf"
    version: "5.2.2"
    sdkconfig_options:
      CONFIG_ESPTOOLPY_FLASHSIZE_4MB: "y"

i2c:
- sda: "GPIO10"
  scl: "GPIO11"
  scan: true
  frequency: "280kHz"
  timeout: "1240us"
```

## Checklist:

### DIFF
--- a/esphome/components/i2c/i2c_bus_esp_idf.cpp
+++ b/esphome/components/i2c/i2c_bus_esp_idf.cpp
@@ -50,13 +50,14 @@ void IDFI2CBus::setup() {
       ESP_LOGW(TAG, "i2c timeout of %" PRIu32 "us greater than max of 13ms on esp-idf, setting to max", timeout_);
       timeout_ = 13000;
     }
-    err = i2c_set_timeout(port_, timeout_ / (APB_CLK_FREQ / 1000000));
+    uint32_t timeout_ticks = timeout_ / (APB_CLK_FREQ / 1000000);
+    err = i2c_set_timeout(port_, timeout_ticks);
     if (err != ESP_OK) {
       ESP_LOGW(TAG, "i2c_set_timeout failed: %s", esp_err_to_name(err));
       this->mark_failed();
       return;
     } else {
-      ESP_LOGV(TAG, "i2c_timeout set to %" PRIu32 " ticks (%" PRIu32 " us)", timeout_ / (APB_CLK_FREQ / 1000000), timeout_);
+      ESP_LOGV(TAG, "i2c_timeout set to %" PRIu32 " ticks (%" PRIu32 " us)", timeout_ticks, timeout_);
     }
   }
   err = i2c_driver_install(port_, I2C_MODE_MASTER, 0, 0, ESP_INTR_FLAG_IRAM);
--- a/esphome/components/i2c/i2c_bus_esp_idf.cpp
+++ b/esphome/components/i2c/i2c_bus_esp_idf.cpp
@@ -50,13 +50,13 @@ void IDFI2CBus::setup() {
       ESP_LOGW(TAG, "i2c timeout of %" PRIu32 "us greater than max of 13ms on esp-idf, setting to max", timeout_);
       timeout_ = 13000;
     }
-    err = i2c_set_timeout(port_, timeout_ * 80);  // unit: APB 80MHz clock cycle
+    err = i2c_set_timeout(port_, timeout_ / (APB_CLK_FREQ / 1000000));
     if (err != ESP_OK) {
       ESP_LOGW(TAG, "i2c_set_timeout failed: %s", esp_err_to_name(err));
       this->mark_failed();
       return;
     } else {
-      ESP_LOGV(TAG, "i2c_timeout set to %" PRIu32 " ticks (%" PRIu32 " us)", timeout_ * 80, timeout_);
+      ESP_LOGV(TAG, "i2c_timeout set to %" PRIu32 " ticks (%" PRIu32 " us)", timeout_ / (APB_CLK_FREQ / 1000000), timeout_);
     }
   }
   err = i2c_driver_install(port_, I2C_MODE_MASTER, 0, 0, ESP_INTR_FLAG_IRAM);
--- a/tests/components/i2c/test.esp32-c3-idf.yaml
+++ b/tests/components/i2c/test.esp32-c3-idf.yaml
@@ -2,3 +2,6 @@ i2c:
   - id: i2c_i2c
     scl: 5
     sda: 4
+    scan: true
+    frequency: "280kHz"
+    timeout: "1000us"
--- a/tests/components/i2c/test.esp32-idf.yaml
+++ b/tests/components/i2c/test.esp32-idf.yaml
@@ -2,3 +2,6 @@ i2c:
   - id: i2c_i2c
     scl: 16
     sda: 17
+    scan: true
+    frequency: "280kHz"
+    timeout: "1000us"


dwhitlow:
Okay, I dug into the source a bit more while working on a corresponding documentation PR, and found this ESP-IDF thread while investigating default values: https://www.esp32.com/viewtopic.php?t=32281 .

The valid range for `i2c_set_timeout()` is enforced by a [check in the function definition](https://github.com/espressif/esp-idf/blob/release/v4.4/components/driver/i2c.c#L839) (`I2C_LL_MAX_TIMEOUT`). `I2C_LL_MAX_TIMEOUT` is defined in ESP-IDF in `components/hal/<variant>/include/hal/i2c_ll.h` (ex: [esp32s2](https://github.com/espressif/esp-idf/blob/release/v4.4/components/hal/esp32s2/include/hal/i2c_ll.h#L86)). This value is typically set to `I2C_TIME_OUT_REG_V`, which is defined in `components/soc/<variant>/include/soc/i2c_reg.h` (ex: [esp32s2](https://github.com/espressif/esp-idf/blob/release/v4.4/components/soc/esp32s2/include/soc/i2c_reg.h)) and is typically `0xFFFFF`. For some variants, this is set to `I2C_TIME_OUT_VALUE` (ex: [esp32c6](https://github.com/espressif/esp-idf/blob/release/v5.2/components/hal/esp32c6/include/hal/i2c_ll.h)) and is typically `0x1F` (ex: [esp32c6](https://github.com/espressif/esp-idf/blob/release/v5.2/components/soc/esp32c6/include/soc/i2c_reg.h#L237)).

Based on my understanding so far, the units for `i2c_set_timeout` and `i2c_get_timeout` depend on the value of `I2C_LL_MAX_TIMEOUT`. Where `timeout_` is the value set via ESPHome in microseconds, and `esp_idf_timeout` is the argument passed to `i2c_set_timeout`:
* `I2C_LL_MAX_TIMEOUT == 0x1F` (c2, c3, c5, c6, c61, h2, p4, s3): `timeout_ = 2^esp_idf_timeout * (1000000 / APB_CLK_FREQ)`, or `esp_idf_timeout = log(timeout_ * (APB_CLK_FREQ / 1000000)) / log(2)`
* `I2C_LL_MAX_TIMEOUT == 0xFFFFF` (all other variants): current behavior, or `esp_idf_timeout = timeout_ * (APB_CLK_FREQ / 1000000)`

If this is true, this PR can be updated so the only currently-supported variants that this will alter the behavior of are the esp32c3 and esp32h2.

dwhitlow:
Alternatively, I could update the PR to add support in the `i2c` component for an optional `raw_timeout` argument. When set, this value would be cast to the appropriate type and passed to the relevant set_timeout function. This approach makes the interface a bit more complicated, but has the advantage of not introducing any breaking changes.

A third option would be to add support for the new ESP-IDF i2c_master driver. According to the documentation, this new driver seems to have consistent behavior for setting timeouts across variants. Unfortunately, the new driver is only available on v5.2 and higher, and implementing this change would likely require changes to every i2c-based ESPHome component.

rvalitov:
Looks interesting. Do you think it can solve this issue https://github.com/esphome/issues/issues/6365 ?
I see that you use `scan: true` - it does not lead to crash at your end?

tests/components/i2c/test.esp32-c3-idf.yaml
```diff
@@ -2,3 +2,6 @@ i2c:
   - id: i2c_i2c
     scl: 5
     sda: 4
+    scan: true
+    frequency: "280kHz"
```
rvalitov, line 6:
```suggestion
```
Why add `frequency` field if you don't use it in code and use `APB_CLK_FREQ` instead?

tests/components/i2c/test.esp32-idf.yaml
```diff
@@ -2,3 +2,6 @@ i2c:
   - id: i2c_i2c
     scl: 16
     sda: 17
+    scan: true
+    frequency: "280kHz"
```
rvalitov, line 6:
```suggestion
```
Why add `frequency` field if you don't use it in code and use `APB_CLK_FREQ` instead?

tests/components/i2c/test.esp32-c3-idf.yaml
```diff
@@ -2,3 +2,6 @@ i2c:
   - id: i2c_i2c
     scl: 5
     sda: 4
+    scan: true
```
rvalitov, line 5: Why we need this line?

esphome/components/i2c/i2c_bus_esp_idf.cpp
rvalitov, line 1: Looks good!

rvalitov (changes requested):

rvalitov:
> A third option would be to add support for the new ESP-IDF i2c_master driver. According to the documentation, this new driver seems to have consistent behavior for setting timeouts across variants. Unfortunately, the new driver is only available on v5.2 and higher, and implementing this change would likely require changes to every i2c-based ESPHome component.

I think this is the only way to go. At least the I2C master driver should be started to be used. ESPHome does not support ESP-IDF version 5.0 or higher, so it's OK to start making such changes without the need to rewrite the whole code base at once. I made an issue for this https://github.com/esphome/issues/issues/6366

Usually people use newer ESP-IDF version if they have a newer boards, like my ESP32-C6. For example, it crashes when using parameter `scan: true` in I2C https://github.com/esphome/issues/issues/6365

swoboda1337:
The new i2c driver has scl_wait_us and there is no need to convert

PR #7078, opened 2024-07-12
[modbus_controller] Add on_command_sent trigger

leejoow:

Add an on_command trigger to the modbus_controller which fires every time a command is sent on the modbus. In the lambda the `function_code` and `address` are available.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4039

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
modbus_controller:
  - id: modbus_con
    address: 0x01
    update_interval: 5min
    offline_skip_updates : 0
    on_command:
      - number.increment: modbus_writes
```

## Checklist:


esphome/components/modbus_controller/const.py
```diff
@@ -6,6 +6,7 @@
 CONF_FORCE_NEW_RANGE = "force_new_range"
 CONF_MODBUS_CONTROLLER_ID = "modbus_controller_id"
 CONF_MODBUS_FUNCTIONCODE = "modbus_functioncode"
+CONF_ON_COMMAND = "on_command"
```
jesserockz: Probably better to be more specific here and say `on_command_sent`
I feel like `on_command` implies an incoming command

```suggestion
CONF_ON_COMMAND_SENT = "on_command_sent"
```

jesserockz (changes requested):

esphome/components/modbus_controller/const.py
leejoow: I agree, I will make the changes

jesserockz (approved):

PR #7085, opened 2024-07-14
jsn_sr04t component: Add RCWL-1655 module compatibility

### LINKED ISSUE #2804: jsn_sr04t component: RCWL-1655 compatibility
I bought a JSN_SR04T sensor module from ebay, thinking it would work with the existing implementation in esphome. Turned out this is a slightly different flavor of controller which didn't work with the existing sensor component.

As it turns out the module is a "RCWL-1655", which requires a different request format and delivers the measurement in 24 instead of 16 bits and without a checksum
https://makerhero.com/img/files/download/RCWL-1655-Datasheet.pdf

**Additional context**
I've implemented support for this flavor of the distance sensor here: 
https://github.com/esphome/esphome/compare/dev...fraxinas:esphome:jsn_sr04t_RCWL-1655

fraxinas:

## Types of changes

**Related issue or feature (if applicable):** fixes [<link to issue>](https://github.com/esphome/feature-requests/issues/2804)

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
  - platform: "jsn_sr04t"
    name: "Distance"
    model: "rcwl_1655"
    update_interval: 60s
```

## Checklist:

I saw @soeffi working on this. In case I missed anything, please let me know

### DIFF
--- a/esphome/components/jsn_sr04t/jsn_sr04t.cpp
+++ b/esphome/components/jsn_sr04t/jsn_sr04t.cpp
@@ -10,23 +10,47 @@ namespace jsn_sr04t {
 static const char *const TAG = "jsn_sr04t.sensor";
 
 void Jsnsr04tComponent::update() {
-  this->write_byte(0x55);
+  switch (this->model_) {
+    case JSN_SR04T:
+    case AJ_SR04M:
+      this->write_byte(0x55);
+      break;
+    case RCWL_1655:
+      this->buffer_.clear();
+      this->write_byte(0xA0);
+      break;
+  }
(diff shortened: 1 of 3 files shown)


jesserockz (changes requested):
Considering this has both a different sending and receive flow for this, I think we dont want to bunch this under the same unrelated `jsn_sr04t` component name.

Instead, I propose we rename this component entirely to `ultrasonic_uart` to indicate that it can be extended more in the future with other models.

If you want to do this work, please do. :pray:

PR #7088, opened 2024-07-15
Drop max BLE client connections limitation

syssi:

In the past an ESP32 wasn't able to handle more than 3 simultaneous BLE client connections because of limited resources. More powerful ESP32 models are able to handle more than 3 connections. Also the ESP32-WROOM model can talk to a bunch of devices if you establish the connections sequentially. 

In other words: The limit is pretty hard and should be relaxed.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** No official issue

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** None

The explicit limitation isn't mentioned at the docs anymore. The general warning is still valid:

> The BLE software stack on the ESP32 consumes a significant amount of RAM on the device.

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/ble_client/__init__.py
+++ b/esphome/components/ble_client/__init__.py
@@ -65,9 +65,7 @@ CONF_ON_PASSKEY_NOTIFICATION = "on_passkey_notification"
 CONF_ON_NUMERIC_COMPARISON_REQUEST = "on_numeric_comparison_request"
 CONF_AUTO_CONNECT = "auto_connect"
 
-# Espressif platformio framework is built with MAX_BLE_CONN to 3, so
-# enforce this in yaml checks.
-MULTI_CONF = 3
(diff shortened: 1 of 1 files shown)


paolss:
yes it is wrong and not working :)

ingrimsch:
I would love to see this merged as this would reduce the amount of ble monitoring hardware/devices required in my special usecase 👍

clydebarrow (approved):
I could not find a definitive answer as to how many connections are actually allowed, but I agree it's best left to the underlying implementation to enforce the limit.

PR #7098, opened 2024-07-16
[docker] Reduce need for privileged mode a bit more

oliv3r:
Currently the general advise is to just use privileged mode with docker/compose. From a security point of view this is obviously undesirable.

One reason why privileged mode is needed, is because some packages [0] incorrectly set permissions on their (`-rwxr-x--- 501:staff package.json`) file. One would rightfully argue, that this doesn't matter, because docker runs as root and thus all permissions are granted anyway. However when working with volumes, this is no longer true, the docker user is mapped to some random UID and within our container we now get a permission denied error, because our remapped root user no longer has access.

While we obviously should aim to fix this error upstream wherever possible, this could take years. Meanwhile lets put the root user into the `staff` group, which seems to be the permissions set by these misbehaving packages. Obviously this will break again once upstream or debian changes the mapping/permission at a later time.

[0]: https://community.platformio.org/t/incorrect-permissions-for-several-packages-on-package-json/41733

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


oliv3r:
With this change, I am now able to use the following compose file to run an unpriviledged container. Note that obviously I am not forwarding any device to my container (as this is a server without anything connected in terms of ESP devices). `--device /dev/X:/dev/Y` would still be needed of course, with probably a good udev rule to ensure that `/dev/X` is always a fixed esphome based entry.

```yaml
etworks:
  esphome: {}

volumes:
  esphome_build:
  esphome_cache:

services:
  esphome:
    image: ghcr.io/esphome/esphome:stable
    cap_add:
      - fowner
      - chown
    cap_drop:
      - all
    depends_on:
      - avahi
    env_file:
      - common.env
      - ./esphome/secret.env
    volumes:
      - esphome_build:/build:rw
      - esphome_cache:/cache:rw
      - ./esphome/config:/config:rw
    networks:
      - esphome
    expose:
      - "6052/tcp"
    restart: unless-stopped
```

jesserockz (changes requested):
This should go in its own RUN step as it has nothing to do with the package installation.

PR #7101, opened 2024-07-17
[http_request] Fix ESP-IDF follow redirect

dentra:

Follow redirects ESP-IDF http client implementation works only when using esp_http_client_perform, so current implementation completely does not work. We can't use esp_http_client_perform because of stream reading (esp_http_client_read), so I implement redirect functionality internally.

~~I also added parameters to change the client's read/write buffer, this will allow, for example, to process redirects from Github correctly.~~ 

At this moment request and response are limited to 512 bytes. PR #7125 resolve this limitation.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
http_request:
  buffer_size_rx: 1024
  buffer_size_tx: 1024
ota:
  - platform: http_request
    id: ota_http_request
update:
  - platform: http_request
    name: Firmware
    device_class: firmware
    source: https://github.com/$OWNER/$REPO/releases/latest/download/manifest.json
```

## Checklist:

### DIFF
--- a/esphome/components/http_request/__init__.py
+++ b/esphome/components/http_request/__init__.py
@@ -1,7 +1,7 @@
-import esphome.codegen as cg
-import esphome.config_validation as cv
 from esphome import automation
+import esphome.codegen as cg
 from esphome.components import esp32
+import esphome.config_validation as cv
 from esphome.const import (
     CONF_ESP8266_DISABLE_SSL_SUPPORT,
     CONF_ID,
@@ -40,8 +40,6 @@ CONF_VERIFY_SSL = "verify_ssl"
 CONF_FOLLOW_REDIRECTS = "follow_redirects"
 CONF_REDIRECT_LIMIT = "redirect_limit"
 CONF_WATCHDOG_TIMEOUT = "watchdog_timeout"
-CONF_BUFFER_SIZE_RX = "buffer_size_rx"
-CONF_BUFFER_SIZE_TX = "buffer_size_tx"
 
 CONF_MAX_RESPONSE_BUFFER_SIZE = "max_response_buffer_size"
 CONF_ON_RESPONSE = "on_response"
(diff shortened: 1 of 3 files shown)


esphome/components/http_request/__init__.py
```diff
@@ -110,6 +112,8 @@ def _declare_request_class(value):
                 cv.positive_not_null_time_period,
                 cv.positive_time_period_milliseconds,
             ),
+            cv.Optional(CONF_BUFFER_SIZE_RX): cv.All(cv.uint16_t, cv.only_with_esp_idf),
+            cv.Optional(CONF_BUFFER_SIZE_TX): cv.All(cv.uint16_t, cv.only_with_esp_idf),
```
jesserockz: The default is 512 for both of these in esp-idf, so we may as well set that here using `cv.SplitDefault`

jesserockz (changes requested):

esphome/components/http_request/__init__.py
jesserockz: Also, as the buffer size is a new feature, it wont be merged into a patch release, but the redirect can be if you split the buffer size setting into its own PR.

dentra: > The default is 512 for both of these in esp-idf, so we may as well set that here using `cv.SplitDefault`

Done

> Also, as the buffer size is a new feature, it wont be merged into a patch release,

If the workaround using external_components is not suitable, of course I will do it.

dentra: Mmm, it would be necessary to update the documentation...

jesserockz: > If the workaround using external_components is not suitable, of course I will do it.

Well I would like to be able to release the redirect fixes in a patch release, but cant do that if it is included in the same PR as a new feature.
If they are in their own PRs then I can =)

dentra: Now split buffer size configuration to PR #7125

dentra: But two PRs broke of external_components usage 😢

jesserockz: If you want to combine them for use externally, then you just push a new branch on your repo and dont open a PR and use that branch in yaml =)

jesserockz (approved):

PR #7118, opened 2024-07-22
Allow unsetting of speaker and media player

HarvsG:

```yaml
  on_start: 
    - if:
        condition:
          lambda: return id(reply_location).state != "Another Media Player";
        then: 
          - lambda: id(va).set_speaker(echo_speaker);
        else
          - lambda: id(va).set_speaker(nullptr); # does not work
```
Currently a user can set a speaker dynamically, however they cannot unset the speaker because `local_output_` will always be true once `set_speaker()` is called.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/voice_assistant/voice_assistant.h
+++ b/esphome/components/voice_assistant/voice_assistant.h
@@ -90,7 +90,7 @@ class VoiceAssistant : public Component {
 #ifdef USE_SPEAKER
   void set_speaker(speaker::Speaker *speaker) {
     this->speaker_ = speaker;
-    this->local_output_ = (this->speaker_ != nullptr)l
+    this->local_output_ = (this->speaker_ != nullptr);
   }
 #endif
 #ifdef USE_MEDIA_PLAYER
--- a/esphome/components/voice_assistant/voice_assistant.h
+++ b/esphome/components/voice_assistant/voice_assistant.h
@@ -96,12 +96,7 @@ class VoiceAssistant : public Component {
 #ifdef USE_MEDIA_PLAYER
   void set_media_player(media_player::MediaPlayer *media_player) {
     this->media_player_ = media_player;
-    if (this->media_player_ != nullptr) {
-      this->local_output_ = true;
-    } else {
-      this->local_output_ = false;
-    }
-  }
+    this->local_output_ = (this->media_player_ != nullptr);
 #endif
 
   uint32_t get_legacy_version() const {
--- a/esphome/components/voice_assistant/voice_assistant.h
+++ b/esphome/components/voice_assistant/voice_assistant.h
@@ -90,13 +90,21 @@ class VoiceAssistant : public Component {
 #ifdef USE_SPEAKER
   void set_speaker(speaker::Speaker *speaker) {
     this->speaker_ = speaker;
-    this->local_output_ = true;
+    if (this->speaker_ != nullptr) {
+      this->local_output_ = true;
+    } else {
+      this->local_output_ = false;
+    }
   }
 #endif
 #ifdef USE_MEDIA_PLAYER
   void set_media_player(media_player::MediaPlayer *media_player) {
     this->media_player_ = media_player;
-    this->local_output_ = true;
+    if (this->media_player_ != nullptr) {
+      this->local_output_ = true;
+    } else {
+      this->local_output_ = false;
+    }
   }
 #endif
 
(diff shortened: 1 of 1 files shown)


HarvsG:
I have not tested this locally as I do not have & and do not know how to set up a dev environment

esphome/components/voice_assistant/voice_assistant.h
```diff
@@ -90,13 +90,21 @@ class VoiceAssistant : public Component {
 #ifdef USE_SPEAKER
   void set_speaker(speaker::Speaker *speaker) {
     this->speaker_ = speaker;
-    this->local_output_ = true;
+    if (this->speaker_ != nullptr) {
+      this->local_output_ = true;
+    } else {
+      this->local_output_ = false;
+    }
```
nielsnl68:
```suggestion
    this->local_output_ = (this->speaker_ != nullptr)l
```
You can simplify your code into the above.

esphome/components/voice_assistant/voice_assistant.h
```diff
@@ -90,13 +90,21 @@ class VoiceAssistant : public Component {
...
 #ifdef USE_MEDIA_PLAYER
   void set_media_player(media_player::MediaPlayer *media_player) {
     this->media_player_ = media_player;
-    this->local_output_ = true;
+    if (this->media_player_ != nullptr) {
+      this->local_output_ = true;
+    } else {
+      this->local_output_ = false;
+    }
   }
```
nielsnl68:
```suggestion
    this->local_output_ = (this->media_player_ != nullptr);
```

nielsnl68 (changes requested):

nielsnl68:
> I have not tested this locally as I do not have & and do not know how to set up a dev environment

Good suggestion. But why do you want to disconnect a speaker component?

HarvsG:
> Good suggestion. But why do you want to disconnect a speaker component?

At the moment, once a speaker is connected there is no way to disable it. Voice assistant will continue to play replies over it.
Two main reasons I can see.
1) To act as a pseudo mute button as I2S audio does not support volume control. (VA supports `media_player` but esp-idf doesn't and many devices are space limited)
2) To disable the speaker when replies are sent to an external media player (this is my use case)
   - although I have a working work-around for this use case https://gist.github.com/HarvsG/40468d1d0e3f444d9a1907f26a503bc6#file-m5stack-atom-echo-mww2beta-reply-selector-full-config-yaml

jesserockz:
> I have not tested this locally as I do not have & and do not know how to set up a dev environment

You need to test this before it can be accepted. Not knowing how to set up a dev environment doesnt stop it from being tested. https://esphome.io/components/external_components

nielsnl68:
> > Good suggestion. But why do you want to disconnect a speaker component?
> 
> At the moment, once a speaker is connected there is no way to disable it. Voice assistant will continue to play replies over it. Two main reasons I can see.
> 
> 1. To act as a pseudo mute button as I2S audio does not support volume control. (VA supports `media_player` but esp-idf doesn't and many devices are space limited)
> 2. To disable the speaker when replies are sent to an external media player (this is my use case)
>    
>    * although I have a working work-around for this use case https://gist.github.com/HarvsG/40468d1d0e3f444d9a1907f26a503bc6#file-m5stack-atom-echo-mww2beta-reply-selector-full-config-yaml

Sorry, i still not see the reason why you want to detach the speaker. Is using the speaker.stop action not enough?

HarvsG:
> Sorry, i still not see the reason why you want to detach the speaker. Is using the speaker.stop action not enough?

No 1) it doesn't seem to do anything. And 2) I want to prevent the speaker from playing. Not stop it half way through playing.

nielsnl68:
Okay, got it. 
I belief it is better to add the mute option to the speaker component. 
The easy thing would be to disallow the play method to add new data to be send to the speaker.

HarvsG:
> Okay, got it. I belief it is better to add the mute option to the speaker component. The easy thing would be to disallow the play method to add new data to be send to the speaker.

That may well be true but is this code change not appropriate anyway? There is a function which can take a nullptr as an argument but behaves as if it can't.

PR #7137, opened 2024-07-26
enhancement's speaker component 

nielsnl68:

This PR updates the speaker and his i2s_audio platform component to allow a smoother control over the data that is send to the speaker. And making better us of the freeRTOS Task support. 

- replacing play xQueue- into a xStreambuffer, this allows smother playing and can return the available free buffer space.
- Added a new action named **speaker.finish** this will stop the speaker after playing all data in the buffer.
- Added a new condition **speaker.is_stopped** this will return true when the speaker is really stopped.
- fixed waring introduced in pr#6872
- allow setting the 16/32 bit word to send to the speaker.
- make sure that the last word is sent is 0; to prevent speaker burning.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4083

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
  
```

## Checklist:


esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -79,195 +80,219 @@ void I2SAudioSpeaker::player_task(void *params) {
...
-    size_t bytes_written;
-
-    memmove(buffer, data_event.data, data_event.len);
-    size_t remaining = data_event.len / 2;
-    size_t current = 0;
-
-    while (remaining > 0) {
-      uint32_t sample = (buffer[current] << 16) | (buffer[current] & 0xFFFF);
+    if (this_speaker->buffer_queue_ != nullptr) {
+      int ret = xStreamBufferReceive(this_speaker->buffer_queue_, &sample, wordsize, portMAX_DELAY);
```
pyos, line 136: reading 4 bytes at a time isn't slow?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -79,195 +80,219 @@ void I2SAudioSpeaker::player_task(void *params) {
...
+void I2SAudioSpeaker::stop_() {
+  if (this->has_buffered_data()) {
+    return;
   }
+
+  vTaskSuspend(this->player_task_handle_);
+  // make sure the speaker has no voltage on the pins before closing the I2S poort
+  size_t bytes_written;
+  uint32_t sample = 0;
+  i2s_write(this->parent_->get_port(), &sample, 4, &bytes_written, (10 / portTICK_PERIOD_MS));
```
pyos, line 208: that seems like a side effect of suspending a task while it's potentially in i2s_write, are you sure this is safe?..

esand:
I decided to give this a test after having issues with the current i2s speaker data queue filling up before the end of the audio clip ([i2s_audio_speaker not playing all audio data](esphome/issues#6222)).

This seems promising, as it plays through more of my data, but it still doesn't play through all of it (and I've only got a 1.6s clip). This code also seems to require the audio data to be in stereo to play at proper speed compared to current main branch code (though, kind of makes sense due to the i2s channel config setting).

FYI, changing the channel config to I2S_CHANNEL_FMT_ONLY_LEFT and providing mono (single channel) raw audio allows more audio to play but it still cuts out eventually. Also, with setting use_16bit_mode off (default), audio seems to play better, but the raw data length must be evenly divisible by 4 to avoid a warning.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
nielsnl68: The idea behind this is to make sure that the output has any positive or negative value still outputting to the speaker DAC. I burned some speakers that way.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
nielsnl68: We could always update the value later, i think.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
pyos: yes, and the reason it's outputting a voltage to the speaker is likely that you stopped the task doing the output at a random point instead of signaling it to exit once it's done with the current iteration. suspending threads is a complete disaster in pthreads in my experience, and i doubt it's any better in smt freertos.

PR #7138, opened 2024-07-26
[esp32_ble] allow setting of min/max key_size and auth_req_mode

whitty:

Allows configuration of additional security params along the same lines as `IoCapability`

The intent is to not change any behaviour unless explicitly configured.

A downstream project forked `esphome` previously to achieve the same end, and I'd like to upstream the change in a generic way.

https://github.com/esphome/esphome/compare/dev...WeekendWarrior1:esphome:emerald_ble#diff-c581745ebf09e9c23ad69f4bc9c01c49858759277e3c347afec5b3596509f521

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#5769  (nee esphome/esphome-docs#4084)

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

However you'd need a device matching those requirements

```yaml
esp32_ble:
    io_capability: keyboard_display
    max_key_size: 16
#    min_key_size: 7   # use default
    auth_req_mode: sc_mitm_bond
```

## Checklist:

I don't understand the structure under `tests/components/esp32_ble` - is it sane to just add params to `tests/components/esp32_ble_client/common.yaml` - or should there be conditional cases? ie different sets of parameters?


whitty:
My general concern with this change is that I don't know if a global object `esp32_ble` is the appropriate scope.  The original change and the similar `io_capability` were all global, but the parameter seem to me like they should be tied to a connection (though the espressif API doesn't seem to allow this).

bdraco:
@whitty Would you please resolve conflicts and add a compile test.

I think these options are rare so they could be behind a define to enable conditional compilation.

whitty:
> Would you please resolve conflicts

Resolved 

>  and add a compile test.

I think I've done that - not sure if that's exactly what you meant.

> I think these options are rare so they could be behind a define to enable conditional compilation.

Will do, not done yet

bdraco:
[Run esphome/pre-commit-action@43cd1109c09c544d97196f7730ee5b2e0cc6d81e: tests/components/esp32_ble_client/test-extended-auth-req-params.esp32-idf.yaml#L2](https://github.com/esphome/esphome/pull/7138/files#annotation_42629030680)
2:5 [indentation] wrong indentation: expected 2 but found 4

whitty:
Sorry - you've caught me at a bad time - I will resolve the issues here hopefully within a week (seems to now be a merge loop/issue introduced from your push too)

bdraco:
No rush. We just pushed 2025.12.0 beta so generally we only merge fixes for the next week anyways.

bdraco (approved):

PR #7145, opened 2024-07-28
build host with debug to show usefull callstack in gdb

tomaszduda23:

gdb show proper lines number that way. It is useful for debugging.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/host/__init__.py
+++ b/esphome/components/host/__init__.py
@@ -1,15 +1,15 @@
+import esphome.codegen as cg
+import esphome.config_validation as cv
 from esphome.const import (
+    CONF_MAC_ADDRESS,
     KEY_CORE,
     KEY_FRAMEWORK_VERSION,
     KEY_TARGET_FRAMEWORK,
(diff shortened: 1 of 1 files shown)


clydebarrow:
I think this hardly needs to be baked into the build, when you can simply add it in any yaml where required:

```
esphome:
  name: lvgl
  platformio_options:
    build_flags: [-g3]
```

tomaszduda23:
debug symbols are pretty standard thing e.g. ubuntu provides it by default. It doesn't make side effects besides of size of binary.

clydebarrow:
It might be a common thing to do, but not always what you want, e.g. when adding optimisation options like -O3 which does not play nicely with -g.

The default options should be only those that are required for the build to succeed, any others desired for a specific use case can be added in the yaml.

PR #7146, opened 2024-07-28
Add tca9555 GPIO driver

mobrembski:

This PR adds driver for Texas Instruments TCA9555 I2C GPIO Port Expander (simillar to PCF8574 and MCP2003).

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4089

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Please have a look at esphome-docs PR.

## Checklist:


esphome/components/tca9555/tca9555.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+void TCA9555Component::dump_config() {
+  ESP_LOGCONFIG(TAG, "TCA9555:");
+  LOG_I2C_DEVICE(this)
+  if (this->is_failed()) {
+    ESP_LOGE(TAG, "Communication with TCA9555 failed!");
+  }
+}
+bool TCA9555Component::digital_read(uint8_t pin) {
+  this->read_gpio_();
```
jesserockz: Something we would like to implement on all GPIO expanders is a "pin reading cache".
Basically it would work like this:
1. _Something_ reads a gpio pin from this expander `digital_read`
2. `digital_read` checks if this pin has been read already in this `loop` 
  a. If it has not, returns the cached value (from `input_mask_` etc)
  b. If is has been read before, then force a new read from the hardware before returning the newly cached value.

This would significantly speed up the GPIO expanders, because right now, every pin read triggers a hardware read, and `gpio` `binary_sensor`s read every configured pin, every loop.

If this is something you think you are able to add to this PR, we would be very grateful as we can then apply it to all of the expanders.

esphome/components/tca9555/tca9555.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+void TCA9555Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up TCA9555...");
+  if (!this->read_gpio_()) {
+    ESP_LOGE(TAG, "TCA9555 not available under 0x%02X", this->address_);
+    this->mark_failed();
+    return;
+  }
+
+  this->write_gpio_();
```
jesserockz: I know other expander components do this already, but we would like to remove it from them as it causes pins to flip on reboot.

One solution to avoid this is to read the output register from the hardware into the `output_mask` so that after setup, any pin changes will only affect the pin being changed and not the others.

esphome/components/tca9555/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    if not (value[CONF_INPUT] or value[CONF_OUTPUT]):
+        raise cv.Invalid("Mode must be either input or output")
+    if value[CONF_INPUT] and value[CONF_OUTPUT]:
+        raise cv.Invalid("Mode must be either input or output")
+    return value
+
+
+TCA9555_PIN_SCHEMA = pins.gpio_base_schema(
+    TCA9555GPIOPin,
+    cv.int_range(min=0, max=17),
```
jesserockz: Values here are inclusive
```suggestion
    cv.int_range(min=0, max=15),
```

esphome/components/tca9555/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+
+@pins.PIN_SCHEMA_REGISTRY.register(CONF_TCA9555, TCA9555_PIN_SCHEMA)
+async def tca9555_pin_to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    parent = await cg.get_variable(config[CONF_TCA9555])
+
+    cg.add(var.set_parent(parent))
+
+    num = config[CONF_NUMBER]
+    cg.add(var.set_pin(num))
```
jesserockz:
```suggestion
    cg.add(var.set_pin(config[CONF_NUMBER]))
```

esphome/components/tca9555/tca9555.h
```diff
@@ -0,0 +1,62 @@
...
+  /// Mask for the pin mode - 1 means output, 0 means input
+  uint16_t mode_mask_{0x00};
+  /// The mask to write as output state - 1 means HIGH, 0 means LOW
+  uint16_t output_mask_{0x00};
+  /// The state read in read_gpio_ - 1 means HIGH, 0 means LOW
+  uint16_t input_mask_{0x00};
+};
+
+/// Helper class to expose a TCA9555 pin as an internal input GPIO pin.
+class TCA9555GPIOPin : public GPIOPin {
```
jesserockz:
```suggestion
class TCA9555GPIOPin : public GPIOPin, public Parented<TCA9555Component> {
```

esphome/components/tca9555/tca9555.h
```diff
@@ -0,0 +1,62 @@
...
+ public:
+  void setup() override;
+  void pin_mode(gpio::Flags flags) override;
+  bool digital_read() override;
+  void digital_write(bool value) override;
+  std::string dump_summary() const override;
+
+  void set_parent(TCA9555Component *parent) { parent_ = parent; }
```
jesserockz:
```suggestion
```

esphome/components/tca9555/tca9555.h
```diff
@@ -0,0 +1,62 @@
...
+  void set_pin(uint8_t pin) { pin_ = pin; }
+  void set_inverted(bool inverted) { inverted_ = inverted; }
+  void set_flags(gpio::Flags flags) { flags_ = flags; }
+
+ protected:
+  TCA9555Component *parent_;
```
jesserockz:
```suggestion
```

esphome/components/tca9555/tca9555.h
```diff
@@ -0,0 +1,62 @@
...
+  void setup() override;
+  void pin_mode(gpio::Flags flags) override;
+  bool digital_read() override;
+  void digital_write(bool value) override;
+  std::string dump_summary() const override;
+
+  void set_parent(TCA9555Component *parent) { parent_ = parent; }
+  void set_pin(uint8_t pin) { pin_ = pin; }
+  void set_inverted(bool inverted) { inverted_ = inverted; }
+  void set_flags(gpio::Flags flags) { flags_ = flags; }
```
jesserockz:
```suggestion
  void set_pin(uint8_t pin) { this->pin_ = pin; }
  void set_inverted(bool inverted) { this->inverted_ = inverted; }
  void set_flags(gpio::Flags flags) { this->flags_ = flags; }
```

esphome/components/tca9555/tca9555.cpp
```diff
@@ -0,0 +1,113 @@
...
+    ESP_LOGE(TAG, "Output register set failed!");
+    return false;
+  }
+
+  this->status_clear_warning();
+  return true;
+}
+float TCA9555Component::get_setup_priority() const { return setup_priority::IO; }
+
+void TCA9555GPIOPin::setup() { pin_mode(flags_); }
```
jesserockz:
```suggestion
void TCA9555GPIOPin::setup() { this->pin_mode(this->flags_); }
```

esphome/components/tca9555/tca9555.cpp
```diff
@@ -0,0 +1,113 @@
...
+void TCA9555GPIOPin::pin_mode(gpio::Flags flags) { this->parent_->pin_mode(this->pin_, flags); }
+bool TCA9555GPIOPin::digital_read() { return this->parent_->digital_read(this->pin_) != this->inverted_; }
+void TCA9555GPIOPin::digital_write(bool value) { this->parent_->digital_write(this->pin_, value != this->inverted_); }
+std::string TCA9555GPIOPin::dump_summary() const {
+  char buffer[32];
+  snprintf(buffer, sizeof(buffer), "%u via TCA9555", pin_);
+  return buffer;
```
jesserockz:
```suggestion
  return str_sprintf("%u via TCA9555", this->pin_);
```

esphome/components/tca9555/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    }
+)
+
+
+@pins.PIN_SCHEMA_REGISTRY.register(CONF_TCA9555, TCA9555_PIN_SCHEMA)
+async def tca9555_pin_to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    parent = await cg.get_variable(config[CONF_TCA9555])
+
+    cg.add(var.set_parent(parent))
```
jesserockz:
```suggestion
    await cg.register_parented(var, config[CONF_TCA9555])
```

jesserockz (changes requested):

esphome/components/tca9555/tca9555.cpp
mobrembski: I can implement it, give me some time this week, i'll try :)

esphome/components/tca9555/tca9555.cpp
mobrembski: +1, good idea. I didnt think about software reboots. Indeed, registers should preserve their contents since there is no reset on expander chip. Will do.

mobrembski:
@jesserockz Feel free to comment.

My idea is to add a "proxy" class  which will implement digitial_read and write, and based on flag (which resets every loop) it calls main component (here tca9555) to update its "cache" content.
I did it because as you said:

> ...we would be very grateful as we can then apply it to all of the expanders.

So i tried to make it more generic.

So far, cached_gpio class is added at tca9555 level in tree. If you like the idea, i suggest to place it somewhere else since it's not tca9555 specific.

jesserockz:
I have made a few more changes to hopefully make it a bit clearer what is happening and moved the class to its own component folder for importing by other expanders.

Let me know if it causes issues with your tca9555. Ill try to test it with an expander I actually have at some point.

mobrembski:
> I have made a few more changes to hopefully make it a bit clearer what is happening and moved the class to its own component folder for importing by other expanders.
> 
> Let me know if it causes issues with your tca9555. Ill try to test it with an expander I actually have at some point.

There was a small bug - we've been missing setting output flags so tca9555 gets always same value. But already fixed ;)

jesserockz:
Oops, not sure how I dropped that out..haha
Thanks for fixing and confirming.

jesserockz (approved):

PR #6721, opened 2024-05-11
[modem] pppos tcp/ip for UART modem (esp32)

oarcher:
# What does this implement/fix?

This implement pppos tcp/ip for esp32, on lte/4G networks.   It conflict with `ethernet` component, but can be used with Wifi AP with #7147

See https://github.com/oarcher/piotech for supported boards, and yaml example.

See:
  - [Modem Component Doc](https://deploy-preview-4056--esphome.netlify.app/components/modem)  
  - [Modem Sensor Doc](https://deploy-preview-4056--esphome.netlify.app/components/sensor/modem.html)
  - [Modem Text Sensor Doc](https://deploy-preview-4056--esphome.netlify.app/components/text_sensor/modem.html) 
  - [Wifi Component](https://deploy-preview-4091--esphome.netlify.app/components/wifi.html) (to setup a Wifi AP that will use the modem connection).

You can join me on this [discord thread](https://discord.com/channels/429907082951524364/1238611620545167401)

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** 
fixes https://github.com/esphome/feature-requests/issues/2303
fixes https://github.com/esphome/feature-requests/issues/477
fixes https://github.com/esphome/feature-requests/issues/1379
fixes https://github.com/esphome/feature-requests/issues/1810

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4056

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml

# While not merged, this PR is also available as an external component
external_components:
    - source: github://pr#6721
      components: [ network, modem ]  # also add 'watchdog' if not found
      refresh: 10min
   # To create a Wifi AP that will use the modem connection
   # - source: github://pr#7147
   #    components: [ wifi ]


modem:
  id: atmodem
  rx_pin: 26
  tx_pin: 27
  model: SIM7600  # or SIM7670 BG96 SIM7000 SIM800 GENERIC
  apn: orange
  pin_code: "0000"


```

## Checklist:
  - [x] The code change is tested and works locally.
[...]

PR #7147, opened 2024-07-28
[wifi] [ethernet] Allow Wifi AP to reuse existing network interface (ethernet or modem) for NAT

oarcher:

This allow the WiFi component to use another network interface (like ethernet, but not limited to) for bridging to another network.
 
This can be done only if the with is only in AP mode (no STA).

So wifi clients connected to the AP can reach internet if the ethernet component is able to do so.

This will be also used by the `modem` component  (#6721) 

## Types of changes

- [ ] Other

#### For WiFi component

During setup for idf, the WiFi component looks for an existing netif by `s_gw_netif = esp_netif_next(nullptr);`
If it exists, `setup()` assume that `esp_netif_init()` and `esp_event_loop_create_default()` was already called by a previous component, and doesn't recall them.

When configuring the wifi AP in `WiFiComponent::wifi_ap_ip_config_`, if `s_gw_netif ` is valid, it fetch the DNS server to offer to the dhcp clients, and setup NAT with `ip_napt_enable`

On the `__init__.py` side, `wifi_has_sta` has been added, so other components can check if the WiFi configuration is AP only.

#### For Ethernet component

The setup priority has been increased, to be sure the created netif will exist during WiFi setup.
The final validation check that there is no STA configuration for WiFi with
```python
def _final_validate(config):
    if wifi_config := fv.full_config.get().get(CONF_WIFI, None):
        if wifi_has_sta(wifi_config):
            raise cv.Invalid("Wifi must be AP only when using ethernet")
```
            

Note that python imports where reoganized by https://github.com/esphome/esphome/pull/7103

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4091

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
wifi:
  # only AP must be used
  ap: 
    ssid: espap
    password: mypassword

ethernet:
  type: ...

```

## Checklist:


jesserockz:
I was so confused by the purpose of this PR by reading the title and description.

Would have been clear if it simply said: `bridge wifi ap to ethernet` or similar =)

oarcher:
Sorry for the title, just changed it!

nagyrobi:
It's still not clear... A bridge shouldn't do NAT, interfaces in a bridge are always in the same network. A bit simplified, a bridge is like a switch/hub.

If there's NAT between the interfaces, that's not a bridge.
Does it also offer DHCP addresses to clients connecting?

oarcher:
@nagyrobi , Yes, the wifi AP still offer DHCP for clients. It also send the previously found DNS server to.

It then configure NAT.

The doc title is "Allow Wifi AP to reuse existing network interface for NAT", and I might be better.

nagyrobi:
Any plans to make configurable some DHCP-specific options? Like custom DNS, NTP, no gateway, etc...

oarcher:
> Any plans to make configurable some DHCP-specific options? Like custom DNS, NTP, no gateway, etc...

Not for the moment, I want to keep this PR as small as possible. It's a basic feature, but I think it's very useful for the modem component (#6721)

rosbeef:
Hello thank you for starting this PR :) i need this functionality, but if i undertand your code, i'm not really conviced by the level3 nat. why not l2 bridging ? or i misunderstood the way it's done.
Is something like this should not be lighter for cpu https://github.com/dokmic/eth2wlan ?

oarcher:
@rosbeef Yes, L2 bridging is probably more efficient and faster than the current implementation.

This version relies on `ip_napt_enable`, which does most of the work. It's convenient for small use cases, like acting as an AP for other ESP devices using MQTT.  
But it's quite slow for general-purpose AP usage, especially when clients are laptops or smartphones.

L2 bridging is harder to implement, since it requires handling raw packets and forwarding them between interfaces using event callbacks. I will try to have a look for this implementation one day...

And sorry for late answer ;)

oarcher:
I've tried to setup L2 bridging, but it seems impossible to use it with a ppp netif. I was able to copy packets from the wifi netif to the ppp netif, but but from from the ppp netif to wifi netif, because there are no driver‑level packet handling for ppp netif.

As I want this PR works also with the modem component, I keep the current `ip_napt_enable`  implementation.
It's slower, but the primary bottleneck for the modem is the UART baud rate (max 3686400).

rosbeef:
Thanks @oarcher .If i understand well, 2l bridging should work with ethernet but as you want it to work with modem you prefer to not include it in your PR isn't it ? Sorry my English compression fail sometimes.

oarcher:
Yes @rosbeef , it's exactly what I mean

esphome/components/wifi/__init__.py
```diff
@@ -427,6 +431,10 @@ def add_sta(ap, network):
         )
         cg.add(var.set_ap_timeout(conf[CONF_AP_TIMEOUT]))
         cg.add_define("USE_WIFI_AP")
+        if CORE.is_esp32 and CORE.using_esp_idf:
+            add_idf_sdkconfig_option("CONFIG_LWIP_IP_FORWARD", True)
+            add_idf_sdkconfig_option("CONFIG_LWIP_IPV4_NAPT", True)
```
bdraco: It seems like this should be opt-in as everyone has to pay for these even if they aren't using them with the current design

oarcher: Those options are only added if a wifi AP is defined. To add other condition, I can:
  1. check if another network component is loaded (modem, ethernet...)
  2. use an 'napt: ' option to the AP configuration
  3. check if the wifi configuration is AP only, with no STA (as this PR introduced this feature)
  
  1/ has the inconvenient to have the list of others network components and 2/ add a new option, so I prefer 3/....

oarcher: Now `if CORE.is_esp32 and CORE.using_esp_idf and not wifi_has_sta(config):` (3)

esphome/components/ethernet/__init__.py
```diff
@@ -35,13 +36,14 @@
...
     CONF_VALUE,
+    CONF_WIFI,
     KEY_CORE,
     KEY_FRAMEWORK_VERSION,
 )
 from esphome.core import CORE, TimePeriodMilliseconds, coroutine_with_priority
 import esphome.final_validate as fv
 
-CONFLICTS_WITH = ["wifi"]
+CONFLICTS_WITH = ["captive_portal"]
```
bdraco, line 59: We definitely have some code and assumptions that will break if we take out the assumption that wifi and ethernet are mutually exclusive. We need to do research to determine what needs to change. We may need to move all of that into a step that happens after we decide which components are loaded and only disable unused code/paths after we know for sure which network methods are enabled

bdraco: Some things to consider and potential challenges. I did not do a full analysis on each of these to see what is a problem or not.

- The WiFi component has a runtime check that prevents AP mode when another network interface exists.
-  Many components (API, MQTT, mDNS, web servers) depend on the network component but assume single interface operation
- The network utilities assume a single primary interface `get_ip_addresses()` `get_use_address()` 
- Wifi `__init__.py` disables some sdk options automatically based on ap mode or not
- Ethernet `__init__.py` disables some WiFi sdk options automatically assuming no wifi

oarcher: the goal to remove the confict with wifi is:
  - allow ethernet to set up a netif
  - allow the wifi component to set up an AP *only* if no STA is defined (`not wifi_has_sta`) (an error is raised if STA is used)
  - the AP will scan for existing netif (from ethernet or modem) to setup NAT with `ip_napt_enable`
  - the network utilities will return the primary (ie the first) interface with `get_ip_addresses()`. So Wifi has to be checked after primary interfaces. It's ok for ethernet, and for the modem, I've added `USE_MODEM` before `USE_WIFI`. see the modem component at https://github.com/esphome/esphome/pull/6721/files#diff-8da1ea97754e500dc543ea031621f454809482da54930c7e7009996a27900576

oarcher: I haven’t extensively tested this PR with Ethernet, but I have thoroughly tested it with the modem component (#6721), including MQTT, HTTP requests OTA updates, and WireGuard.

oarcher:
Keep open until https://github.com/esphome/esphome/pull/6721#issuecomment-3366238958

tbrasser:
What if you want the other way around (for example to add wifi to an ethernet-only device)? Is that supported?

oarcher:
@tbrasser , if ethernet and wifi are available, this PR assume that ethernet is the *primary* interface (ie WAN, internet). Wifi will be used to set an AP, so wifi client can access the WAN (dhcp + nat).

The other way around is not supported: if the wifi is connected to a WAN, a client device can't be connected to ethernet (no dhcp server on ethernet).

oarcher:
My latest commit introduces `WiFiComponent::wifi_ap_nat(esp_netif_t *gateway_netif)`, which is intended to be called explicitly by another network component (Ethernet, modem, etc.), i.e. the component that owns the primary / gateway netif.

The idea is the following:

During setup:
- The primary network interface (the gateway) is initialized first by the modem or Ethernet component, which may already call `esp_event_loop_create_default()` (see #13896).
- The WiFi AP component silently ignores `ESP_ERR_INVALID_STATE` from `esp_event_loop_create_default()`, since it is setting up a secondary interface.
- Later, once the primary (gateway) interface is fully up, the component owning that interface explicitly asks the WiFi component to reconfigure the AP so that traffic is routed through the gateway netif. This includes restarting DHCP server, updating DNS, and enabling NAT.

For example, from the modem component:

```cpp
#ifdef USE_WIFI_AP
  esphome::wifi::global_wifi_component->wifi_ap_nat(this->modem_handler->ppp_netif);
#endif
```

PR #7162, opened 2024-07-30
[pid] Add get_min_integral() and get_max_integral()

FreeBear-nc:
Kp, Ki, and Kd all have get & set methods. Having get calls for min & max integrals is handy to have when initializing templates that go on to clamp these parameters.


jesserockz (approved):

jesserockz:
Please make sure to use the PR template correctly and do not just delete it all.

PR #7164, opened 2024-07-30
[micro_wake_word] Fix VAD detection and modify detection computation

kahrendt:

This PR fixes a bug where voice activity was nearly always detected. I originally compared the quantized unsigned integer output of the VAD model to a floating point number between 0 and 1. This PR computes the floating point equivalent of the quantized output before comparing.

Additionally, this PR now uses a sliding window average rather than the max over the sliding window for the VAD probability. This keeps the VAD detection result more stable.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4097

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

micro_wake_word:
  vad:
  models:
    - model: okay_nabu
    - model: hey_mycroft
  on_wake_word_detected:
    then:
      - voice_assistant.start:

```

## Checklist:

### DIFF
--- a/esphome/components/micro_wake_word/streaming_model.cpp
+++ b/esphome/components/micro_wake_word/streaming_model.cpp
@@ -148,7 +148,7 @@ WakeWordModel::WakeWordModel(const uint8_t *model_start, float probability_cutof
 };
 
 bool WakeWordModel::determine_detected() {
-  int32_t sum = 0;
+  uint32_t sum = 0;
   for (auto &prob : this->recent_streaming_probabilities_) {
     sum += prob;
   }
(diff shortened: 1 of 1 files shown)


esphome/components/micro_wake_word/streaming_model.cpp
```diff
@@ -175,12 +175,14 @@ VADModel::VADModel(const uint8_t *model_start, float probability_cutoff, size_t
 };
 
 bool VADModel::determine_detected() {
-  uint8_t max = 0;
+  int32_t sum = 0;
```
jesserockz: `prob` is a `uint8_t`, so it doesn't make sense to have this as signed.
```suggestion
  uint32_t sum = 0;
```

jesserockz (approved):

PR #5535, opened 2023-10-15
Read/Write bootloader, partition table and any partition via OTA

angelnu:
# What does this implement/fix?

This adds support to update the bootloader and partition tables with OTA. It allows migrating from
firmwares with different layouts (new Tasmota versions) or repair botloaders (such as the locked new Shellies).

Updating the bootloader and partition tables only works with the esp-idf framework. For arduino [arduino-esp32](https://github.com/espressif/arduino-esp32/releases/tag/3.0.0-alpha3) is needed. This is currently in alpha so no platformio package is available. It is possible to move from esp-idf to arduino OTA so you can migrate Tasmota -> esphome(esp-idf) -> esphome(arduino).

## New CLI commands

```bash
# update OTA the bootloader without rebooting
python -m esphome -v upload test.yaml --bootloader --no-reboot
# update OTA the partition table without rebooting (requires esp32-idf version > 5.1)
python -m esphome -v upload test.yaml --partition-table --no-reboot
# update OTA partition with type and subtype
python -m esphome -v upload --partition-type=0x0 --partition-subtype=0x10 test.yaml
# update OTA partition with label
python -m esphome -v upload --partition-label=app0 test.yaml
# download OTA the bootloader
python -m esphome -v download test.yaml --bootloader --file=bootloader.bin
# download OTA the partition table without rebooting (requires esp32-idf version > 5.1)
python -m esphome -v download test.yaml --partition-table --file=partition_table.bin
# download OTA partition with type and subtype
python -m esphome -v download --partition-type=0x0 --partition-subtype=0x10 test.yaml  --file=firmware.bin
# download OTA partition with label
python -m esphome -v download --partition-label=app0 test.yaml  --file=firmware.bin
# update OTA all esphome parts (equivalent to a factory reset)
python -m esphome upload-factory-ota test.yaml
```

The `upload-factory-ota` command makes a backup of the partition table before flashing the new one. If any operation fails then the backup is automatically restored. This usually happens when the new firmware would override the running partition.

## Tasmota migration

If you are migrating from Tasmota 1.12 or newer using the Safeboot partition you need the following steps. There are 2 options depending on the used esp-idf framework.

### Single command (requires esp-idf>5.1)

This is how to do it without custom partition_tables.

1. Build an ESPHome (`test.yaml`) with a minimal configuration (with wifi and OTA) with esp-idf>5.1
```yaml
# Example for esp32-c3 (Shelly Plus 1PM Mini)
esphome:
  name: test
  platformio_options:
[...]

PR #7172, opened 2024-08-01
ade7953_spi wrong size specified in read_array call

SimoPk:

read_array is called to read a 16 bit value into a 2 bytes array, but the read_array second parameter was erroneously 4. This resulted in frequent crashes/reboots when testing the component on a Shelly Pro 4PM, which has 2 aded7953 on the same spi bus.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/ade7953_spi/ade7953_spi.cpp
+++ b/esphome/components/ade7953_spi/ade7953_spi.cpp
@@ -60,7 +60,7 @@ bool AdE7953Spi::ade_read_16(uint16_t reg, uint16_t *value) {
   this->write_byte16(reg);
   this->transfer_byte(0x80);
   uint8_t recv[2];
-  this->read_array(recv, 4);
+  this->read_array(recv, 2);
   *value = encode_uint16(recv[0], recv[1]);
   this->disable();
   return false;


angelnu (approved):
Good catch!

I also have 4x Pro 4PM and they were periodically freezing - I was thinking that it was due to the complicated code in the display (it takes 120 ms so a lot of warnings) but this bug is more likely to be the reason.

We should get this fix with the next patch release if possible @jesserockz .

SimoPk:
I started from your example configuration (https://github.com/angelnu/esphome/blob/master/devices/shelly/pro4pm/base_v1.yaml) for mine, and I noticed that after at most 10 minutes (often much less) I had a reboot without anything on the log.

I decided to restart from a minimal configuration, removing the display code for sure and also every device/sensor on the config. Since this experiment worked and I had no more reboots, I started adding the components one by one (I also thought it was the display at first). This way I narrowed the problem down to the second ade7953. Out of curiosity I read the code for the component to check if there were any race conditions or something else, and I stumbled upon the typo I signaled in this PR... fortunately it worked and now I have no reboots!

Btw, thanks again for your great work, I was able to start from your configuration and get the device working with esphome in no time!
Another minor thing: I haven't understand exactly if you flashed your devices by first uploading tasmota or directly through the serial port. On the internet I haven't found any info about the 4pm serial port so I tried by myself and succeeded, and so I can confirm that the pins next to the LAN port are the serial interface with the same pinout as other shelly devices, GND being the bottom pin. Hope this helps!

jesserockz (approved):

ptr727:
Is `_countof()` an esphome used pattern?

angelnu:
> I started from your example configuration (https://github.com/angelnu/esphome/blob/master/devices/shelly/pro4pm/base_v1.yaml) for mine, and I noticed that after at most 10 minutes (often much less) I had a reboot without anything on the log.
> 
> I decided to restart from a minimal configuration, removing the display code for sure and also every device/sensor on the config. Since this experiment worked and I had no more reboots, I started adding the components one by one (I also thought it was the display at first). This way I narrowed the problem down to the second ade7953. Out of curiosity I read the code for the component to check if there were any race conditions or something else, and I stumbled upon the typo I signaled in this PR... fortunately it worked and now I have no reboots!
> 
> Btw, thanks again for your great work, I was able to start from your configuration and get the device working with esphome in no time! Another minor thing: I haven't understand exactly if you flashed your devices by first uploading tasmota or directly through the serial port. On the internet I haven't found any info about the 4pm serial port so I tried by myself and succeeded, and so I can confirm that the pins next to the LAN port are the serial interface with the same pinout as other shelly devices, GND being the bottom pin. Hope this helps!

I used Tasmota as intermediate step using https://github.com/esphome/esphome/pull/5535

Mine were taking days to hang but I did not check for reboots. Great you were mite melodic debuging it :-)

PR #7173, opened 2024-08-01
Update MiCS Values

### LINKED ISSUE #6106: mics_4514 sensor reading way off
### The problem

I noticed that the sensor readings for the mics_4514 are way off. The current code two issues: Low concentrations are ignored, higher values are calculated wrong. 

Here is the graph for the datasheet for the RED sensor (the OX based sensors have the same issue):

image

I marked the thresholds used in code for the various sensors with colored dots. Two issues for each sensor:
- only concentrations on the right side of these dot are reported! Especially dangerous with the CO sensor that ignores values below 11ppm (red dot). Please note 6ppm is the max CO concentration for 24h, totally ignored by this implementation
- The value calculations are wrong! The datasheet graph is logarithmic, but the value interpolation is  done linear (only correct for the max concentrations).

Firmware: 2024.6.6 (Jul 8 2024, 07:53:53)

Home Assistant Add-on

core 2024.7

mics_4514

### Board

ESP32-C3 (Apollo Air-1 device)

mics_4514

### Example YAML snippet

n/a the issue is in the component code itself. Here a code snippet of the component. 
- Arrow 1 marks the threshold (red CO dot in the graph)
- Arrow 2 the linear interpolation formula (only correct for the lowest red_f value 0.01):

image
(https://github.com/esphome/esphome/blob/3920029aff9dd85fd5f0775945b8426c03da471c/esphome/components/mics_4514/mics_4514.cpp#L71-L80)

```txt
N/A
```

### Additional information

_No response_

TrevorSchirmer:

Corrects values reported in the MiCS component to match datasheet.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/6106

## Test Environment

- [ x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/mics_4514/mics_4514.cpp
+++ b/esphome/components/mics_4514/mics_4514.cpp
@@ -92,10 +92,8 @@ void MICS4514Component::update() {
 
   if (this->methane_sensor_ != nullptr) {
     float methane = 0.0f;
-    if (red_f > 0.9f) {
+    if (red_f > 0.9f || red_f < 0.5) {  // outside the range->unlikely
       methane = 0.0;
-    } else if (red_f < 0.5) {
-      methane = 0.0;  // outside the range->unlikely
     } else {
       methane = 630 / pow(red_f, 4.4);
     }
--- a/esphome/components/mics_4514/mics_4514.cpp
+++ b/esphome/components/mics_4514/mics_4514.cpp
@@ -122,10 +122,8 @@ void MICS4514Component::update() {
 
   if (this->ammonia_sensor_ != nullptr) {
     float ammonia = 0.0f;
-    if (red_f > 0.98f) {
+    if (red_f > 0.98f || red_f < 0.2532) {  // outside the ammonia range->unlikely
       ammonia = 0.0;
-    } else if (red_f < 0.2532) {
-      ammonia = 0.0;  // outside the ammonia range->unlikely
     } else {
       ammonia = 0.9 / pow(red_f, 4.6);
     }
--- a/esphome/components/mics_4514/mics_4514.cpp
+++ b/esphome/components/mics_4514/mics_4514.cpp
@@ -112,10 +112,8 @@ void MICS4514Component::update() {
 
   if (this->hydrogen_sensor_ != nullptr) {
     float hydrogen = 0.0f;
-    if (red_f > 0.9f) {
+    if (red_f > 0.9f || red_f < 0.02) {  // outside the range->unlikely
       hydrogen = 0.0;
-    } else if (red_f < 0.02) {
-      hydrogen = 0.0;  // outside the range->unlikely
     } else {
       hydrogen = 0.85 / pow(red_f, 1.75);
     }
--- a/esphome/components/mics_4514/mics_4514.cpp
+++ b/esphome/components/mics_4514/mics_4514.cpp
@@ -102,10 +102,8 @@ void MICS4514Component::update() {
 
   if (this->ethanol_sensor_ != nullptr) {
     float ethanol = 0.0f;
-    if (red_f > 1.0f) {
+    if (red_f > 1.0f || red_f < 0.02) {  // outside the range->unlikely
       ethanol = 0.0;
-    } else if (red_f < 0.02) {
-      ethanol = 0.0;  // outside the range->unlikely
     } else {
       ethanol = 1.52 / pow(red_f, 1.55);
     }
--- a/esphome/components/mics_4514/mics_4514.cpp
+++ b/esphome/components/mics_4514/mics_4514.cpp
@@ -85,7 +85,7 @@ void MICS4514Component::update() {
     if (ox_f < 0.3f) {
       nitrogendioxide = 0.0;
     } else {
-      nitrogendioxide = 0.164 * pow(0x_f, 0.975);
+      nitrogendioxide = 0.164 * pow(ox_f, 0.975);
     }
     this->nitrogen_dioxide_sensor_->publish_state(nitrogendioxide);
   }
--- a/esphome/components/mics_4514/mics_4514.cpp
+++ b/esphome/components/mics_4514/mics_4514.cpp
@@ -70,72 +70,70 @@ void MICS4514Component::update() {
 
   if (this->carbon_monoxide_sensor_ != nullptr) {
     float co = 0.0f;
-    if (red_f <= 0.425f) {
-      co = (0.425f - red_f) / 0.000405f;
-      if (co < 1.0f)
-        co = 0.0f;
-      if (co > 1000.0f)
-        co = 1000.0f;
+    if (red_f > 3.4f) {
+      co = 0.0;
+    } else if (red_f < 0.01) {
+      co = 1000.0;
+    } else {
+      co = 4.2 / pow(red_f, 1.2);
     }
     this->carbon_monoxide_sensor_->publish_state(co);
   }
 
   if (this->nitrogen_dioxide_sensor_ != nullptr) {
     float nitrogendioxide = 0.0f;
-    if (ox_f >= 1.1f) {
-      nitrogendioxide = (ox_f - 0.045f) / 6.13f;
-      if (nitrogendioxide < 0.1f)
-        nitrogendioxide = 0.0f;
-      if (nitrogendioxide > 10.0f)
-        nitrogendioxide = 10.0f;
+    if (ox_f < 0.3f) {
+      nitrogendioxide = 0.0;
+    } else {
+      nitrogendioxide = 0.164 * pow(0x_f, 0.975);
     }
     this->nitrogen_dioxide_sensor_->publish_state(nitrogendioxide);
   }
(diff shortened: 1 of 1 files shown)


erkr (approved):
Looks good, to enable easy verification of the calculations, it could help to add two 'raw' sensors:
- One for the red_f value 
- One for the ox_f value

esphome/components/mics_4514/mics_4514.cpp
```diff
@@ -70,72 +70,70 @@ void MICS4514Component::update() {
...
-    if (ox_f >= 1.1f) {
-      nitrogendioxide = (ox_f - 0.045f) / 6.13f;
-      if (nitrogendioxide < 0.1f)
-        nitrogendioxide = 0.0f;
-      if (nitrogendioxide > 10.0f)
-        nitrogendioxide = 10.0f;
+    if (ox_f < 0.3f) {
+      nitrogendioxide = 0.0;
+    } else {
+      nitrogendioxide = 0.164 * pow(0x_f, 0.975);
```
manafount: Looks like a small typo here: `0x_f` should be `ox_f`

esphome/components/mics_4514/mics_4514.cpp
```diff
@@ -70,72 +70,70 @@ void MICS4514Component::update() {
 
   if (this->carbon_monoxide_sensor_ != nullptr) {
     float co = 0.0f;
-    if (red_f <= 0.425f) {
-      co = (0.425f - red_f) / 0.000405f;
-      if (co < 1.0f)
-        co = 0.0f;
-      if (co > 1000.0f)
-        co = 1000.0f;
+    if (red_f > 3.4f) {
```
erkr, line 73: As CO has the largest range on the RED sensor, it triggers by all gasses supported.
CO false positives are very alarming, so I propose a small addition to this method to avoid false positives.

- for both low and high levels of CO the RED sensor detection is CO only, only the mid region of RED is not conclusive
- Assumption is that CO gradually builds up, so before measuring other gasses, (low) concentrations of CO (outside the range of other gasses) will be measured first.
- so the approach I propose is to keep track if CO was measured a few cycles before entering the mid region. 
- Only report the CO measurement if RED was a few cycles in a CO only range, before entering the mid range

The pseudo code (not tested/ compiled):
``` 
// extra code to keep track if CO was detected initially
static long exclusive_co=0;// counter
if (red_f > 3.4f) {
  exclusive_co=0;
}
else if (red_f > 1.0f|| red_f < 0.02f) {
  exclusive_co++; // increment only in the CO only range
}
// only the first line below is changed in order to wait 2 cycles before reporting
if (exclusive_co < 3) { // 3 is my guestimate. to be tested
 co = 0.0f;
}
else if (red_f < 0.01f) {
  co = 1000.0f;
}
else {
  co = 4.2/ pow(red_f,1.2f);
}
```

TrevorSchirmer:
Updated my AIR-1 and instantly had much better readings in the low ppm

@manafount @erkr If you'd like to test here is the change in ESPHome addon to the yaml

```
packages:
  ApolloAutomation.AIR-1:
    url: https://github.com/ApolloAutomation/AIR-1
    ref: Mics_Test 
    files: [Integrations/ESPHome/AIR-1.yaml]
    refresh: 1min 
```

jpiszcz:
Hello, I have had an issue as reported here, Ammonia levels on one if my AIR-1's which should always be 0.00ppm has been spiking after sometime, I have written the following block below to the AIR-1 and I do see there is now a lot of logging in the ESPHome console, I will see if this issue I have been having presents itself again.


```
packages:
  ApolloAutomation.AIR-1:
    url: https://github.com/ApolloAutomation/AIR-1
    ref: Mics_Test 
    files: [Integrations/ESPHome/AIR-1.yaml]
    refresh: 1min
```

AIR-1 Dining Room: 0.00ppm
AIR-1 Garage: 11.76ppm (as shown in screenshot)

I moved the AIR-1 Dining Room into the Garage (above the other unit) and it remained 0.00ppm.
The uptime on the AIR-1 Garage was ~13d, I am guessing that this bug takes awhile to occur.
After updating the AIR-1 Garage unit I placed it back in the garage on 8/29, I will see if the levels of Ammonia spike again (there are no sources of Ammonia in the garage, (used for storage only and no natural gas here), and the other AIR-1 showed 0ppm for Ammonia).

sensor-comparison
a2

Will see if the Ammonia levels spike again with this new version.

jpiszcz:
Update: I have updated two AIR-1s, one in the garage, another in the house, changes I have noticed:

1. Ammonia in garage before (10-20 ppm) -> now 1.06ppm (maybe fixed?)
2. Carbon Monoxide in the house (4.20ppm) -> is this right (it was always 0.00ppm before the update)
3. Ethanol in the house -> 1.52ppm is this right (it was always 0.00ppm before the update)
4. Nitrogen Dioxide in the house (0.16ppm) -> is this right?

air1-compare

jpiszcz:
Around ~1-2hrs later after the HVAC ran for awhile it seems most of everything appears to be accurate **indoors** with the exception of Carbon Monoxide, none of my other devices (3) are showing any carbon monoxide in the house.

bb2

esphome/components/mics_4514/mics_4514.cpp
```diff
@@ -70,72 +70,70 @@ void MICS4514Component::update() {
...
-    if (red_f <= 0.786f) {
-      methane = (0.786f - red_f) / 0.000023f;
-      if (methane < 1000.0f)
-        methane = 0.0f;
-      if (methane > 25000.0f)
-        methane = 25000.0f;
+    if (red_f > 0.9f) {
+      methane = 0.0;
+    } else if (red_f < 0.5) {
+      methane = 0.0;  // outside the range->unlikely
```
jesserockz:
```suggestion
    if (red_f > 0.9f || red_f < 0.5) {  // outside the range->unlikely
      methane = 0.0;
```

esphome/components/mics_4514/mics_4514.cpp
```diff
@@ -70,72 +70,70 @@ void MICS4514Component::update() {
...
-    if (red_f <= 0.306f) {
-      ethanol = (0.306f - red_f) / 0.00057f;
-      if (ethanol < 10.0f)
-        ethanol = 0.0f;
-      if (ethanol > 500.0f)
-        ethanol = 500.0f;
+    if (red_f > 1.0f) {
+      ethanol = 0.0;
+    } else if (red_f < 0.02) {
+      ethanol = 0.0;  // outside the range->unlikely
```
jesserockz:
```suggestion
    if (red_f > 1.0f || red_f < 0.02) {  // outside the range->unlikely
      ethanol = 0.0;
```

esphome/components/mics_4514/mics_4514.cpp
```diff
@@ -70,72 +70,70 @@ void MICS4514Component::update() {
...
-    if (red_f <= 0.279f) {
-      hydrogen = (0.279f - red_f) / 0.00026f;
-      if (hydrogen < 1.0f)
-        hydrogen = 0.0f;
-      if (hydrogen > 1000.0f)
-        hydrogen = 1000.0f;
+    if (red_f > 0.9f) {
+      hydrogen = 0.0;
+    } else if (red_f < 0.02) {
+      hydrogen = 0.0;  // outside the range->unlikely
```
jesserockz:
```suggestion
    if (red_f > 0.9f || red_f < 0.02) {  // outside the range->unlikely
      hydrogen = 0.0;
```

esphome/components/mics_4514/mics_4514.cpp
```diff
@@ -70,72 +70,70 @@ void MICS4514Component::update() {
...
-    if (red_f <= 0.8f) {
-      ammonia = (0.8f - red_f) / 0.0015f;
-      if (ammonia < 1.0f)
-        ammonia = 0.0f;
-      if (ammonia > 500.0f)
-        ammonia = 500.0f;
+    if (red_f > 0.98f) {
+      ammonia = 0.0;
+    } else if (red_f < 0.2532) {
+      ammonia = 0.0;  // outside the ammonia range->unlikely
```
jesserockz:
```suggestion
    if (red_f > 0.98f || red_f < 0.2532) {  // outside the ammonia range->unlikely
      ammonia = 0.0;
```

jesserockz (changes requested):

jesserockz (approved):

ha-elguano:
I wanted to chime in and say that implementing the test branch of the MICS fix above, I'm also now getting accurate readings at low concentrations. Sensors that have ALWAYS read zero now display reasonable values and ranges. 

mics2

jpiszcz:
Agree with ha-elguano’s comment above.  Once this version is merged into the ESPHome, will the debugging window remain enabled?

PR #7178, opened 2024-08-02
adjust to new python pre-commit hooks

tomaszduda23:

It applies formatting by new pre-commit hooks and lint which can be merge without the rest. 

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/climate/climate_traits.h
```diff
@@ -168,21 +168,21 @@ class ClimateTraits {
 
  protected:
   void set_mode_support_(climate::ClimateMode mode, bool supported) {
-    if (supported) {
+    if (supported) {  // NOLINT(bugprone-branch-clone)
```
jesserockz: Im confused as to why these are causing this rule to fire. they are clearly different bodies

jesserockz (changes requested):
Could you split this PR up even further so there is 1 for components, 1 for scripts (partially done in https://github.com/esphome/esphome/pull/7198), and 1 for core changes?

esphome/components/climate/climate_traits.h
tomaszduda23: for some reason it does not pick that error now

jesserockz (approved):

PR #7167, opened 2024-07-31
Update i2s_audio_speaker.cppi2s_audio/speaker: Fix fallthrough compiler warning

Mimoja:
# What does this implement/fix?

The current startup code is relying on a fallthrough on startup. This is causing a compiler warning, which we fix by adding the [[fallthrough]] attribute to the respective switch statement.

**Requires C++17, no helper was added as local build of the firmware did work out of the box**

Warning addressed:
```
src/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp: In member function 'virtual void esphome::i2s_audio::I2SAudioSpeaker::loop()':
src/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp:235:19: warning: this statement may fall through [-Wimplicit-fallthrough=]
       this->start_();
       ~~~~~~~~~~~~^~
src/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp:236:5: note: here
     case speaker::STATE_RUNNING:
     ^~~~
```

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

## Test Environment

- [X] ESP32 (Build test)
- [X] ESP32 IDF (Build test)
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:
The ready made Voice assistand configuration for echo atom

## Checklist:
  - [X] The code change is tested and works locally.
  - [ ] ~~Tests have been added to verify that the new code works (under `tests/` folder).~~

If user exposed functionality or configuration variables are added/changed:
  - ~~[ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs)~~

PR #7183, opened 2024-08-03
Move I2S config settings the the base i2sAudio files. Phase 1

nielsnl68:

This PR allows us to have the same audio options for both speaker, microphone and is part of multiple upcoming PR's. Also this PR will remove some unused and duplicated yaml variable. 

To make the Speaker component compatible with the microphone component there will be some breaking change by changing "mode" variable will be changed into "channel". 
In addition you can now set the bits per sample, the sample rates and clock provider Internal (= primary) or External (= secondary).

The *channel* can be set to *left*, *right* or *stereo* (we could also add *all_left* and *all_right*)

This PR includes also the change proposed by @Mimoja in #7167 so the fixed warning is gone with this PR as well.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

esp_adf:

i2s_audio:
  i2s_lrclk_pin: GPIO33
  i2s_bclk_pin: GPIO19

microphone:
  - platform: i2s_audio
    id: echo_microphone
    i2s_din_pin: GPIO23
    adc_type: external
    pdm: true

speaker:
  - platform: i2s_audio
    id: echo_speaker
    dac_type: external
    i2s_dout_pin: GPIO22
    bits_per_sample: 16bit
    channel: right
```

## Checklist:


esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -233,6 +233,7 @@ void I2SAudioSpeaker::loop() {
   switch (this->state_) {
     case speaker::STATE_STARTING:
       this->start_();
+      [[fallthrough]];  // thanks to @Mimoja pr#7167
```
oliv3r: Are comments of gratitude, while nice, really needed?

oliv3r:
Forgive me for not providing a reference, but I'm sure I read somewhere (probably the snapclient esp32 port) that esp_adf was considered bad and should be avoided.

I can imagine esp_adf being used for the various codecs/decoders, but it is truely needed for just the I2S bits?

Just out of curiosity, does this replace the legacy and deprecated i2s audio stuff from esp and replaces it with the new API?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -64,17 +64,17 @@ void I2SAudioSpeaker::player_task(void *params) {
...
 
   i2s_driver_config_t config = {
-      .mode = (i2s_mode_t) (I2S_MODE_MASTER | I2S_MODE_TX),
-      .sample_rate = 16000,
-      .bits_per_sample = I2S_BITS_PER_SAMPLE_16BIT,
-      .channel_format = I2S_CHANNEL_FMT_RIGHT_LEFT,
+      .mode = (i2s_mode_t) (this_speaker->i2s_mode_ | I2S_MODE_TX),
+      .sample_rate = this_speaker->sample_rate_,
+      .bits_per_sample = this_speaker->bits_per_sample_,
+      .channel_format = this_speaker->channel_,
```
pyos, line 70: if you set this to I2S_CHANNEL_FMT_ONLY_LEFT or I2S_CHANNEL_FMT_ONLY_RIGHT, the data provided to i2s_write should be half as long, so the code that does `uint32_t sample = (buffer[current] << 16) | (buffer[current] & 0xFFFF);` below becomes incorrect

Hedda:
> This PR allows us to have the same audio options for both speaker, microphone and is part of multiple upcoming PR's. Also this PR will remove some unused and duplicated yaml variable.

Do you plan to do same/similar consolidation/move/refactoring with I2S config settings from I²S Audio Media Player component?

* https://esphome.io/components/media_player/i2s_audio

oliv3r:
> > This PR allows us to have the same audio options for both speaker, microphone and is part of multiple upcoming PR's. Also this PR will remove some unused and duplicated yaml variable.
> 
> Do you plan to do same/similar consolidation/move/refactoring with I2S config settings from I²S Audio Media Player component?
> 
>     * https://esphome.io/components/media_player/i2s_audio

I'd almost argue, could we not have a generic audio layer, that both speaker and audio player hook into, having to do this work only once ...

nielsnl68:
> I'd almost argue, could we not have a generic audio layer, that both speaker and audio player hook into, having to do this work only once ...

Go for it :D

nielsnl68:
Thanks for the fixes @jesserockz

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
jesserockz: I am planning to merge your PR after this one (and some rebasing) where you handle this

jesserockz (approved):

PR #7188, opened 2024-08-04
Adds MQTT component to Alarm Control panel component

heythisisnate:

Makes the Alarm Panel Component implement MQTT using the MQTT component.

After this change, when MQTT is enabled, the Alarm Control Panel component will publish its state to a MQTT state topic and receive commands by subscribing to a MQTT command topic, following the conventions in the MQTT component.

States are published as strings matching these [HA states](https://www.home-assistant.io/integrations/alarm_control_panel.mqtt/).
Valid command strings are (case insensitive): `ARM_AWAY`, `ARM_HOME`, `ARM_NIGHT`, `ARM_VACATION`, `ARM_CUSTOM_BYPASS`, `DISARM`, `PENDING`, and `TRIGGERED`

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4107

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
mqtt:
  broker: $mqtt_addr

alarm_control_panel:
  - platform: template
    name: Alarm Panel
    binary_sensors:
      - input: zone1
      - input: zone2

```

## Checklist:

### DIFF
--- a/esphome/components/mqtt/mqtt_alarm_control_panel.cpp
+++ b/esphome/components/mqtt/mqtt_alarm_control_panel.cpp
@@ -45,8 +45,16 @@ void MQTTAlarmControlPanelComponent::setup() {
 void MQTTAlarmControlPanelComponent::dump_config() {
   ESP_LOGCONFIG(TAG, "MQTT alarm_control_panel '%s':", this->alarm_control_panel_->get_name().c_str());
   LOG_MQTT_COMPONENT(true, true)
+  ESP_LOGCONFIG(TAG, "  Supported Features: %" PRIu32, this->alarm_control_panel_->get_supported_features());
+  ESP_LOGCONFIG(TAG, "  Requires Code to Disarm: %s", YESNO(this->alarm_control_panel_->get_requires_code()));
+  ESP_LOGCONFIG(TAG, "  Requires Code To Arm: %s", YESNO(this->alarm_control_panel_->get_requires_code_to_arm()));
 }
-void MQTTAlarmControlPanelComponent::send_discovery(JsonObject root, mqtt::SendDiscoveryConfig &config) {}
+
+void MQTTAlarmControlPanelComponent::send_discovery(JsonObject root, mqtt::SendDiscoveryConfig &config) {
+  root[MQTT_SUPPORTED_FEATURES] = this->alarm_control_panel_->get_supported_features();
+  root[MQTT_CODE_DISARM_REQUIRED] = this->alarm_control_panel_->get_requires_code();
+  root[MQTT_CODE_ARM_REQUIRED] = this->alarm_control_panel_->get_requires_code_to_arm();
+}
 
 std::string MQTTAlarmControlPanelComponent::component_type() const { return "alarm_control_panel"; }
 const EntityBase *MQTTAlarmControlPanelComponent::get_entity() const { return this->alarm_control_panel_; }
--- a/esphome/components/mqtt/mqtt_alarm_control_panel.cpp
+++ b/esphome/components/mqtt/mqtt_alarm_control_panel.cpp
@@ -81,9 +81,41 @@ const EntityBase *MQTTAlarmControlPanelComponent::get_entity() const { return th
 bool MQTTAlarmControlPanelComponent::send_initial_state() { return this->publish_state(); }
 bool MQTTAlarmControlPanelComponent::publish_state() {
   bool success = true;
-  const auto state = this->alarm_control_panel_->get_state();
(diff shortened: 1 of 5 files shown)


esphome/components/mqtt/mqtt_alarm_control_panel.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+std::string MQTTAlarmControlPanelComponent::component_type() const { return "alarm_control_panel"; }
+const EntityBase *MQTTAlarmControlPanelComponent::get_entity() const { return this->alarm_control_panel_; }
+
+bool MQTTAlarmControlPanelComponent::send_initial_state() { return this->publish_state(); }
+bool MQTTAlarmControlPanelComponent::publish_state() {
+  bool success = true;
+  const auto state = this->alarm_control_panel_->get_state();
+  const char *state_s = LOG_STR_ARG(alarm_control_panel_state_to_string(state));
+  if (!this->publish(this->get_state_topic_(), state_s))
```
jesserockz: Have you tested this with Home Assistant?
If I am reading it correctly, HA expects the state payload to be lower case.

esphome/components/mqtt/mqtt_alarm_control_panel.cpp
```diff
@@ -0,0 +1,68 @@
...
+    }
+    call.perform();
+  });
+}
+
+void MQTTAlarmControlPanelComponent::dump_config() {
+  ESP_LOGCONFIG(TAG, "MQTT alarm_control_panel '%s':", this->alarm_control_panel_->get_name().c_str());
+  LOG_MQTT_COMPONENT(true, true)
+}
+void MQTTAlarmControlPanelComponent::send_discovery(JsonObject root, mqtt::SendDiscoveryConfig &config) {}
```
jesserockz: This should be populated with supported features and code settings etc.

jesserockz (changes requested):
ESPHome's default MQTT implementation should always work with Home Assistant if the platform is implemented there.

https://github.com/home-assistant/core/blob/dev/homeassistant/components/mqtt/alarm_control_panel.py

heythisisnate:
@jesserockz thanks for the quick review.
Good point. I wasn't thinking of HA as a client and wasn't using it for testing. I will adjust to make sure it works.

heythisisnate:
@jesserockz Working perfectly with HA now with lowercase state payloads.

The only thing I'm unsure about is if the `supported_features` data format is correct. It looks like this is stored as an integer which has a mapping to some values but HA is expecting a list of strings. Can you give me a hint or example how to decode this into the strings that HA wants to see?

heythisisnate:
Nevermind, I figured it out. This should be good now. Working great with HA and MQTT discovery.

esphome/components/mqtt/mqtt_alarm_control_panel.cpp
```diff
@@ -0,0 +1,96 @@
...
+std::string MQTTAlarmControlPanelComponent::component_type() const { return "alarm_control_panel"; }
+const EntityBase *MQTTAlarmControlPanelComponent::get_entity() const { return this->alarm_control_panel_; }
+
+bool MQTTAlarmControlPanelComponent::send_initial_state() { return this->publish_state(); }
+bool MQTTAlarmControlPanelComponent::publish_state() {
+  bool success = true;
+  const auto state = this->alarm_control_panel_->get_state();
+  const std::string state_lower = str_lower_case(LOG_STR_ARG(alarm_control_panel_state_to_string(state)));
+  const char *state_s = state_lower.c_str();
+  if (!this->publish(this->get_state_topic_(), state_s))
```
jesserockz: I dont think we should be using the log argument for sending the string state. Its there for logging, hence the name :sweat_smile: 

https://github.com/esphome/esphome/blob/f737ca6e286f36e243b5bae1bcb7515b9a4bf857/esphome/components/mqtt/mqtt_climate.cpp#L255-L284

While this would basically be repeating the same code as the log helper, I think for consistency it is better to implement similarly to the climate.

If we do a bigger refactor on them, they will be done together.

jesserockz (changes requested):
Thanks for making those changes. 
I don't have my HA set up with mqtt so I cannot test it easily. 

Please also add acp to the `tests/components/mqtt` files so it gets compiled in CI.

jesserockz (approved):

PR #7192, opened 2024-08-05
hydreon_rgxx: Fix parsing of data line

domob1812:

This fixes parsing of received data lines from a RG-15 sensor with the `hydreon_rgxx` component.  In the situation that the `Acc` sensor is not enabled in the yaml configuration, the data line (which starts with `Acc` as the first reported value) would not have been detected as an actual data line, and instead treated as unparsable, with the values not filled into the sensors that are configured.

The error logged in that case is:

    [hydreon_rgxx.sensor:288]: Got unknown line: Acc  0.00 mm, EventAcc  0.00 mm, TotalAcc  0.23 mm, RInt  0.00 mmph

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: hydreon_rgxx
    model: RG_15
    id: hydreon
    uart_id: uart_rg15
    resolution: high
    # acc: is intentionally not configured here
    total_acc:
      name: "RG-15 Rain Total"
    r_int:
      name: "RG-15 Rain Intensity"
    update_interval: 60s
```

## Checklist:

### DIFF
--- a/esphome/components/hydreon_rgxx/hydreon_rgxx.cpp
+++ b/esphome/components/hydreon_rgxx/hydreon_rgxx.cpp
@@ -236,7 +236,7 @@ void HydreonRGxxComponent::process_line_() {
   }
   bool is_data_line = false;
   for (int i = 0; i < NUM_SENSORS; i++) {
-    if (this->sensors_[i] != nullptr && this->buffer_starts_with_(PROTOCOL_NAMES[i])) {
+    if (this->sensors_[i] != nullptr && this->buffer_.find(PROTOCOL_NAMES[i]) != std::string::npos) {
       is_data_line = true;
       break;
     }


functionpointer (approved):
Looks good. I have tested with RG-9 it still works, so that's good.
I assume you have tested with RG-15?

It is a little crazy that nobody has run into this before.

domob1812:
Thanks for taking a look!  Yes, I'm running this with RG-15, and with the config shown in the OP (i.e. without `Acc`).  I was a bit baffled, too, but it was easy enough to spot while looking at the code.  I guess everyone else just has all the sensors configured in ESPHome and just disables them in HASS as needed.

jesserockz (approved):

PR #7193, opened 2024-08-05
[http_request] Add custom UpdateInfo checks

dentra:

This PR adds the ability to do additional custom UpdateInfo checks to decide whether an update is available.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
http_request:
ota:
  - platform: http_request
update:
  - platform: http_request
    name: Firmware
    source: $FW_MANIFEST_JSON
    check_update: |-
      return !str_endswith(x.latest_version, "-beta");
```

## Checklist:


esphome/components/http_request/update/http_request_update.h
```diff
@@ -25,10 +25,15 @@ class HttpRequestUpdate : public update::UpdateEntity, public PollingComponent {
...
 
+  void set_check_update(std::function<bool(const update::UpdateInfo &)> &&check_update) {
+    this->check_update_ = std::move(check_update);
+  }
+
  protected:
   HttpRequestComponent *request_parent_;
   OtaHttpRequestComponent *ota_parent_;
   std::string source_url_;
+  std::function<bool(const update::UpdateInfo &)> check_update_{[](const update::UpdateInfo &) { return true; }};
```
jesserockz:
```suggestion
  std::function<bool(const update::UpdateInfo &)> check_update_{};
```

esphome/components/http_request/update/http_request_update.cpp
```diff
@@ -125,8 +125,17 @@ void HttpRequestUpdate::update() {
 
   this->update_info_.current_version = current_version;
 
-  if (this->update_info_.latest_version.empty() || this->update_info_.latest_version == update_info_.current_version) {
+  if (this->update_info_.latest_version.empty() ||
+      this->update_info_.latest_version == this->update_info_.current_version ||
+      !this->check_update_(this->update_info_)) {
```
jesserockz:
```suggestion
      (this->check_update_ != nullptr && !this->check_update_(this->update_info_))) {
```

esphome/components/http_request/update/http_request_update.cpp
```diff
@@ -125,8 +125,17 @@ void HttpRequestUpdate::update() {
...
     this->state_ = update::UPDATE_STATE_NO_UPDATE;
+
+    this->update_info_.latest_version = this->update_info_.current_version;
+    this->update_info_.title = "";
+    this->update_info_.summary = "";
+    this->update_info_.firmware_url = "";
+    this->update_info_.release_url = "";
+    this->update_info_.md5 = "";
```
jesserockz, line 138: Resetting all of this will break the `force` update functionality.

jesserockz (changes requested):

esphome/components/http_request/update/http_request_update.cpp
dentra: if I change the return value to enum (AVAILABLE, LIMITED, DENIED) and process it accordingly, will that be enough?

jesserockz: What is limited and denied?

dentra: Available - check was success and allow any update
Limited - check was failed but allow to do force update
Denied - check was failed and completely disallow any update

your names are welcome :)

dentra: For example we have a new firmware that use new partition mapping or something other critical and do not want to flash it via ota, it update will just broke device for user. So even a forced update in this case should be prohibited.

Another case is the classic beta version, in this case, a forced update is what is needed.

And the usual case is when ota is installed automatically, without the need for the user to force a special operation.

dentra: Some additional enum suggestion: accept_any, only_forced, reject_all

PR #7209, opened 2024-08-06
remove dependencies to components from core part 2

tomaszduda23:

This is first part of my old commit https://github.com/esphome/esphome/pull/4470. The goal is to remove hardcoded component lists in core. It allows to add new entities by external components.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] ESP32-S3

```yaml
# Example config.yaml

```

## Checklist:


bdraco:
Can you give a bit more color about what the goal is?

tomaszduda23:
1. It was made to allow things like https://github.com/esphome/esphome/pull/3917 to be used as external components
2. Now all component types are hardcoded in core. It is impossible to extend it without modifying the core
3. As for now the main reason to push it forward would be to remove code duplication in application.h. Probably you could do similar thing with macros (which is great is some case). I just somehow prefer templates for this kind of job.

bdraco:
I appreciate the effort to reduce code duplication in `application.h` - that's a worthwhile improvement.

I'm supportive of this change as long as the primary goal remains code de-duplication. While I understand the extensibility benefits for external components, I want to ensure core platforms remain hard-coded in ESPHome for stability and performance because it gives us the ability to make much better decisions at compile time.

The reduction in boilerplate is great, and if it makes external components easier to add without touching core, that's a nice side benefit. Just want to confirm we're aligned that this isn't moving toward making core platforms dynamically loadable.

tomaszduda23:
This is based on std::tuple which had to be defined at compilation time. Under the hood, it should work exactly like the original code. No extra dynamic allocation.
The only difference is that list of components for std::tuple is generated in python.

PR #7214, opened 2024-08-06
hx711: Check for DOUT going high after a reading

dwmw2:
The protocol for the HX711 is very simple, but it *is* possible to detect
that the device is absent and not just report zeroes as it does at the
moment:

 ﻿[D][hx711:033]: 'HX711 value': Got value 0

After a reading, when the 25th and subsequent pulses of the SCK line
are indicating the gain for the next reading, DOUT is supposed to be
high. Check for that, and discard the reading if it doesn't happen:

 [W][hx711:066]: HX711 DOUT pin not high after reading (data 0x0)!

The board which actually *does* have an HX711 attached also still works.

 [D][hx711:031]: 'HX711 value': Got value -1073110
 ﻿[D][sensor:094]: 'HX711 value': Sending state 66.13105 kg with 0 decimals of accuracy

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/hx711/hx711.cpp
+++ b/esphome/components/hx711/hx711.cpp
@@ -39,8 +39,8 @@ bool HX711Sensor::read_sensor_(uint32_t *result) {
     return false;
   }
 
-  this->status_clear_warning();
   uint32_t data = 0;
+  bool final_dout;
 
   {
     InterruptLock lock;
@@ -59,7 +59,16 @@ bool HX711Sensor::read_sensor_(uint32_t *result) {
       this->sck_pin_->digital_write(false);
       delayMicroseconds(1);
     }
+    final_dout = this->dout_pin_->digital_read();
   }
+
+  if (!final_dout) {
+    ESP_LOGW(TAG, "HX711 DOUT pin not high after reading (data 0x%" PRIx32 ")!", data);
+    this->status_set_warning();
+    return false;
+  }
+
+  this->status_clear_warning();
 
   if (data & 0x800000ULL) {
     data |= 0xFF000000ULL;


esphome/components/hx711/hx711.cpp
```diff
@@ -10,7 +10,9 @@ static const char *const TAG = "hx711";
 void HX711Sensor::setup() {
   ESP_LOGCONFIG(TAG, "Setting up HX711 '%s'...", this->name_.c_str());
   this->sck_pin_->setup();
+  this->sck_pin_->pin_mode(gpio::FLAG_OUTPUT);
   this->dout_pin_->setup();
+  this->dout_pin_->pin_mode(gpio::FLAG_INPUT | gpio::FLAG_PULLUP);
```
jesserockz: The pin modes are already set to these by default in the config/code generation (take a look at the generated main.cpp) 
Pull-up is not, but that depends on user config.

jesserockz (changes requested):

esphome/components/hx711/hx711.cpp
dwmw2: I double-checked the SparkFun HX711 board schematic, and it *doesn't* have an external pullup. That implies that it's internal to the HX711 chip. So configuring it for pullup probably isn't necessary if there's a chip present.

Now that I've found the trick of checking for DOUT to go high after a reading, that should suffice to catch the cases where there *isn't* a chip present. I'll just drop that part.

Thanks.

dwmw2: Thanks. I double-checked the [schematic](https://cdn.sparkfun.com/assets/f/5/5/b/c/SparkFun_HX711_Load_Cell.pdf) of the SparkFun HX711 board, and it doesn't have an external pullup on the DOUT line. So it must be internal to the chip, and the pullup on the GPIO shouldn't really be necessary when there's a chip attached.

Now that I've found the trick of waiting for DOUT to go high after the reading, we can live without it because we can detect absent chips that way.

I'll drop the GPIO config part and retest.

jesserockz (approved):

PR #7216, opened 2024-08-07
Add mac address filtering to bluetooth_proxy

jeblair:

This adds support for filtering mac addresses in the bluetooth proxy. This can be useful for environments where a user only needs to proxy a small number of BLE devices.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/4118

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
bluetooth_proxy:
  active: true
  mac_address:
    - "12:34:56:AB:CD:EF"

```

## Checklist:

### DIFF
--- a/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
+++ b/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
@@ -35,6 +35,15 @@ bool BluetoothProxy::parse_device(const esp32_ble_tracker::ESPBTDevice &device) 
   return true;
 }
 
+bool BluetoothProxy::check_bda_matches(esp_ble_gap_cb_param_t::ble_scan_result_evt_param &adv) {
+  if (this->address_vec_.empty()) {
+    return true;
+  }
+
+  return std::find(this->address_vec_.begin(), this->address_vec_.end(), esp32_ble::ble_addr_to_uint64(adv.bda)) !=
+         this->address_vec_.end();
+}
+
 bool BluetoothProxy::parse_devices(esp_ble_gap_cb_param_t::ble_scan_result_evt_param *advertisements, size_t count) {
   if (!api::global_api_server->is_connected() || this->api_connection_ == nullptr || !this->raw_advertisements_)
     return false;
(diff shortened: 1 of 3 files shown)


esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -35,30 +35,44 @@ bool BluetoothProxy::parse_device(const esp32_ble_tracker::ESPBTDevice &device)
...
+      uint8_t length = result.adv_data_len + result.scan_rsp_len;
+      adv.data.reserve(length);
+      for (uint16_t i = 0; i < length; i++) {
+        adv.data.push_back(result.ble_adv[i]);
+      }
+
+      resp.advertisements.push_back(std::move(adv));
+      ESP_LOGV(TAG, "Proxying 1 packet");
+      this->api_connection_->send_bluetooth_le_raw_advertisements_response(resp);
+    }
```
jesserockz: This is going to significantly slow down the bluetooth proxy, even for those not using this filtering.

jesserockz: There is also another PR #6752 which filters the mac addresses at the lower levels so the packets from other mac addresses dont even make it to ESPHome.

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
```diff
@@ -35,30 +35,44 @@ bool BluetoothProxy::parse_device(const esp32_ble_tracker::ESPBTDevice &device)
...
 
+bool BluetoothProxy::check_bda_matches(esp_ble_gap_cb_param_t::ble_scan_result_evt_param &adv) {
+  if (!this->address_vec_.empty()) {
+    uint64_t u64_addr = esp32_ble::ble_addr_to_uint64(adv.bda);
+    if (std::find(this->address_vec_.begin(), this->address_vec_.end(), u64_addr) == this->address_vec_.end()) {
+      return false;
+    }
+  }
+  return true;
+}
```
samueltardieu: Nitpick: YMMV, but I find the following organization clearer as it returns early with a positive answer if no address list has been set.

```suggestion
  if (this->address_vec_.empty()) {
    return true;
  }
  const uint64_t u64_addr = esp32_ble::ble_addr_to_uint64(adv.bda);
  return std::find(this->address_vec_.begin(), this->address_vec_.end(), u64_addr) != this->address_vec_.end();
}
```

(I would even have used a `||` and inlined `u64_addr` but that is not the point here)

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
jeblair: Thanks for taking a look.  I definitely did not prematurely optimize this.  :)

Regarding 6752, that sounds like a great idea, and I would prefer it except that I share the concern about the statically defined limit of 10 devices.  My own use case is starting with proxying 10 ATC thermometers; I'm only likely to increase that over time.  Given that, I don't think it makes sense to abandon this in favor of the other PR (as much as I would love to).

If you're interested in this PR, I'd be happy to optimize it.  What's the level of your concern about performance?
We could:
A) Implement the short-circuit suggested by @samueltardieu so that this only affects people that use the filter
B) Look into optimizing the matching further to get the best (or at least, consistent) performance for folks that do use the filter (maybe a bloom filter, or trie, or... I'm open to suggestions).

Let me know how you'd like to proceed.

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
jeblair: That sounds good to me, thanks.  I might need to learn a little bit more about the code generator, but I think there's a good chance we could use a macro or #ifdef or similar to optimize out that case at compile time (assuming the compiler doesn't do that already).

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
jesserockz: I think that if somebody has not set a filter list, the code that executes now should stay the same.

jeblair: Okay, there should be no change in execution now if the mac addr filter is unused.

esphome/components/bluetooth_proxy/bluetooth_proxy.cpp
jeblair: Done.

joe-sydney:
Is there a possibility to add IRKs to the whitelist as well? Back in 2022, I had put in a feature request for MAC whitelist, but since then I also have a device (iPhone) that uses IRK and I presume a MAC whitelist wouldn't work for that.

PR #7226, opened 2024-08-07
Support Setting Dashboard Port from Environment Variable

### LINKED ISSUE #2718: Specify dashboard port for docker container

I'd like the possibility to specify the dashboard port number in the docker container. We can already put our username and password in an environment variable, so why not the port number? 

I always move away from standard port numbers on any docker container I use, but with the esphome docker container, this is not possible. 
One workaround is to add this command to the compose file: 
``` 
command: "dashboard /config --port ${ESPHOME_PORT} --username ${ESPHOME_USER} --password ${ESPHOME_PASS}"
```
The dashboard is started this way with my specified port number, but now the healtcheck fails as this is still done on port 6052. 

**Additional context**

zhudhjen:

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/2718

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


zhudhjen:
(previous [PR](https://github.com/esphome/esphome/pull/7224) closed due to messed up commit history)

The implementation has finished. Tests have been performed locally regarding dashboard port configuration: 
Screenshot 2024-08-07 104609

Changes are only made to the dashboard, so no on-board testing was done. The dashboard was able to discover and adopt my devices properly.

Doc has not been updated yet. I did not find any current environment variables being documented in the doc. I can add it to [here](https://esphome.io/guides/cli.html#cmdoption-esphome-dashboard-port), but is there a convention to omit such configurations? As `USERNAME` and `PASSWORD` seem not documented either.

esphome/__main__.py
```diff
@@ -889,7 +889,7 @@ def parse_args(argv):
         "--port",
         help="The HTTP port to open connections on. Defaults to 6052.",
         type=int,
-        default=6052,
+        default=0,
```
jesserockz: Untested, but i'm sure that this is really all that is needed for this to work
```suggestion
        default=os.environ.get("ESPHOME_DASHBOARD_PORT", 6052),
```

jesserockz (changes requested):

esphome/__main__.py
zhudhjen: Updated according to your suggestion. Tested and it works: 
image

tests/dashboard/test_web_server.py
```diff
@@ -35,14 +35,14 @@ async def fetch(self, path: str, **kwargs) -> HTTPResponse:
...
+@pytest_asyncio.fixture(name="dashboard")
+async def fixture_dashboard() -> DashboardTestHelper:
     sock, port = bind_unused_port()
     args = Mock(
         ha_addon=True,
         configuration=get_fixture_path("conf"),
-        port=port,
     )
+    os.environ["ESPHOME_DASHBOARD_PORT"] = str(port)
     DASHBOARD.settings.parse_args(args)
```
jesserockz, line 46: I actually think the port here is not doing anything as the `sock` is being used directly to "host" the webserver and the `port` number is directly being used by the `DashboardTestHelper` lower down.

zhudhjen: Tested by removing the `os.environ` line and it still passed, so this is true. Then how should I test this port? Or do I have to test it? As it looks like other environment vars are not being tested either, e.g. `USERNAME`/`PASSWORD`.

jesserockz: You cant really test it in these unit tests due to the way the port assigning works.

Basically you would need to write a new test that tests `__main__.parse_args` to ensure that the env var is in the `port` argument

zhudhjen: I cannot write a test to cover all the parse arg behaviors now -- there are too many arguments to cover and I'm not familiar with most of them. I can add `test_parse_args.py` in `tests/unit_tests` to cover the test of the `port` argument. Potentially other args can be added to this test later. Do you think it is acceptable?

jesserockz: Yes. You are only changing port, so testing only port is fine

PR #7247, opened 2024-08-11
[api] Error log when NONE Update command is sent

oarcher:

Since #7190, the [web_server test](https://github.com/esphome/esphome/actions/runs/10341300837/job/28622985737?pr=6721) fail for #672, #7239 and #7049 with error:
```
src/esphome/components/api/api_connection.cpp: In member function 'virtual void esphome::api::APIConnection::update_command(const esphome::api::UpdateCommandRequest&)':
src/esphome/components/api/api_connection.cpp:1339:21: error: format '%d' expects argument of type 'int', but argument 5 has type 'long unsigned int' [-Werror=format=]
 1339 |       ESP_LOGW(TAG, "Unknown update command: %d", msg.command);
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

As I'm not sure that enums are always `'long unsigned int` for every platform, I've handled all `UpdateCommandRequest` enums in `update_command`. The `default` case is no more handled, but this should not be a problem as `clang-diagnostic-switch` is enabled.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (dismissed):

jesserockz:
> The default case is no more handled, but this should not be a problem as clang-diagnostic-switch is enabled.

This was actually more for when an aioesphomeapi client that may have new options sends to a device that is not updated with the new commands.

jesserockz (approved):

PR #7248, opened 2024-08-11
Allow project name and version as improv_serial identity

### LINKED ISSUE #2591: [PR offer] Configurable improv_serial firmware name and version identifier

When building a firmware image using ESPHome and the improv_serial component, ESPHome will configure the improv version response to be "ESPHome", the ESPHome version used to build the firmware and the device name. The code for this -- if I'm not mistaken -- can be found here:
https://github.com/esphome/esphome/blob/dev/esphome/components/improv_serial/improv_serial_component.cpp#L166-L170
I would like both the version string and firmware identifier to be configureable from YAML. 

When using improv and ESP Web Tools to flash the resulting firmware image to a MCU, these strings are presented to the user to identify the firmware and firmware version. This makes sense for a lot of use-cases, but in others one might want to use the use-case specific firmware's name and version. For example, I am using ESPHome to create a firmware image for a device I am distributing to end-users. I would like the firmware to identify itself via improv as the use-case specific firmware and use my own versioning system.

**Additional context**

Currently however, this is not possible. The improv_serial component does not allow for these values to be configured or overridden from the ESPHome yaml.

I would see two possible options to implement an override mechanism:

1) An explicit override configuration through the improv_serial component. E.g. through optional `firmware_name` and `firmware_version` configuration fields. If not provided, the ESPHome defaults are used, otherwise the values provided through the configuration fields are used.

2) Alternatively `project: name` and `project: version` from the `esphome` section could be used. An additional configuration field (e.g. `identify_with_project`) in improv_serial might be introduced to enable this override, which should be off by default. So for existing configurations they would continue using the ESPHome defaults, even if they have project_name and project_version currently set.

I was wondering if this is an oversight or just so far has not been a priority or if this is a specific conscious decision? Would a pull request to add the option to override the firmware identifier and version from YAML be something that the project is willing to accept? If so, some feedback on the proposed approach would be greatly appreciated.

AzonInc:

With this change the improv_serial component will use the project name and version as identify whenever project data is available.

When using ESP Web Tools it's possible to show the project name and version instead of ESPHome. This allows the use of the same firmware check paired with the manifest.json.

Without this change it's not possible to determine if the same ESPHome project firmware is running on the ESP, thus not possible to see the update button. Also it's not possible to install a firmware without erasing first because it's always a new firmware for ESP Web Tools.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** resolves https://github.com/esphome/feature-requests/issues/2591
**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4125

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

improv_serial:
```

## Checklist:

  - [] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

### DIFF
--- a/esphome/components/improv_serial/improv_serial_component.cpp
+++ b/esphome/components/improv_serial/improv_serial_component.cpp
@@ -171,6 +171,9 @@ std::vector<uint8_t> ImprovSerialComponent::build_rpc_settings_response_(improv:
 
 std::vector<uint8_t> ImprovSerialComponent::build_version_info_() {
   std::vector<std::string> infos = {"ESPHome", ESPHOME_VERSION, ESPHOME_VARIANT, App.get_name()};
+#ifdef ESPHOME_PROJECT_NAME
+  infos = {ESPHOME_PROJECT_NAME, ESPHOME_PROJECT_VERSION, ESPHOME_VARIANT, App.get_name()};
+#endif
   std::vector<uint8_t> data = improv::build_rpc_response(improv::GET_DEVICE_INFO, infos, false);
   return data;
 };
--- a/esphome/components/improv_serial/improv_serial_component.cpp
+++ b/esphome/components/improv_serial/improv_serial_component.cpp
@@ -170,9 +170,10 @@ std::vector<uint8_t> ImprovSerialComponent::build_rpc_settings_response_(improv:
 }
 
 std::vector<uint8_t> ImprovSerialComponent::build_version_info_() {
-  std::vector<std::string> infos = {"ESPHome", ESPHOME_VERSION, ESPHOME_VARIANT, App.get_name()};
 #ifdef ESPHOME_PROJECT_NAME
-  infos = {ESPHOME_PROJECT_NAME, ESPHOME_PROJECT_VERSION, ESPHOME_VARIANT, App.get_name()};
+  std::vector<std::string> infos = {ESPHOME_PROJECT_NAME, ESPHOME_PROJECT_VERSION, ESPHOME_VARIANT, App.get_name()};
+#else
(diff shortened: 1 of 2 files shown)


esphome/components/improv_serial/improv_serial_component.cpp
```diff
@@ -171,6 +171,9 @@ std::vector<uint8_t> ImprovSerialComponent::build_rpc_settings_response_(improv:
 
 std::vector<uint8_t> ImprovSerialComponent::build_version_info_() {
   std::vector<std::string> infos = {"ESPHome", ESPHOME_VERSION, ESPHOME_VARIANT, App.get_name()};
+#ifdef ESPHOME_PROJECT_NAME
+  infos = {ESPHOME_PROJECT_NAME, ESPHOME_PROJECT_VERSION, ESPHOME_VARIANT, App.get_name()};
+#endif
```
jesserockz:
```suggestion
#ifdef ESPHOME_PROJECT_NAME
  std::vector<std::string> infos = {ESPHOME_PROJECT_NAME, ESPHOME_PROJECT_VERSION, ESPHOME_VARIANT, App.get_name()};
#else
  std::vector<std::string> infos = {"ESPHome", ESPHOME_VERSION, ESPHOME_VARIANT, App.get_name()};
#endif
```

jesserockz (changes requested):

jesserockz (approved):

PR #7250, opened 2024-08-12
bl0942: Fix init sequence, add address and line_frequency options

### LINKED ISSUE #6143: BL0942 register initialisation looks wrong
### The problem

Our bl0942 register definitions look like [this](https://github.com/esphome/esphome/blob/dev/esphome/components/bl0942/bl0942.cpp#L14-L19):
```
static const uint8_t BL0942_WRITE_COMMAND = 0xA8;
static const uint8_t BL0942_REG_I_FAST_RMS_CTRL = 0x10;
static const uint8_t BL0942_REG_MODE = 0x18;
static const uint8_t BL0942_REG_SOFT_RESET = 0x19;
static const uint8_t BL0942_REG_USR_WRPROT = 0x1A;
static const uint8_t BL0942_REG_TPS_CTRL = 0x1B;
```
They don't seem to match the [OpenBeken](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/src/driver/drv_bl0942.c#L38-L41) definitions:
```
#define BL0942_REG_USR_WRPROT 0x1D
#define BL0942_USR_WRPROT_DISABLE 0x55

// User operation register (read and write)
#define BL0942_REG_MODE 0x19
#define BL0942_MODE_DEFAULT 0x87
#define BL0942_MODE_RMS_UPDATE_SEL_800_MS (1 << 3)
```

The [datasheet](https://www.belling.com.cn/media/file_object/bel_product/BL0942/datasheet/BL0942_V1.06_en.pdf) appears to agree with OpenBeken:
Screenshot from 2024-08-12 10-26-26

dev

pip

N/A

BK72XX

### Board

Tongou DIN rail power meter

bl0942

### Example YAML snippet

_No response_

_No response_

### Additional information

I have a new one arriving today and will sniff what the Tuya firmware sends before reflashing it.

dwmw2:
Closes: https://github.com/esphome/issues/issues/6143

Fix the register definitions and init sequence for the BL0942 sensor.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/6143

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4126

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
uart:
  id: uart_bus
  tx_pin: TX1
  rx_pin: RX1
  baud_rate: 4800
  stop_bits: 1

sensor:
  - platform: bl0942
    uart_id: uart_bus
    line_frequency: 60Hz
    address: 0
    update_interval: 20s
    current:
      name: BL0942 Current
    voltage:
      name: BL0942 Voltage
    power:
      name: BL0942 Power
      filters:
        multiply: -1
    energy:
      name: BL0942 Energy
    frequency:
      name: BL0942 Frequency
      accuracy_decimals: 2

```

## Checklist:

### DIFF
--- a/esphome/components/bl0942/bl0942.cpp
+++ b/esphome/components/bl0942/bl0942.cpp
@@ -2,6 +2,8 @@
 #include "esphome/core/log.h"
 #include <cinttypes>
 
+// Datasheet: https://www.belling.com.cn/media/file_object/bel_product/BL0942/datasheet/BL0942_V1.06_en.pdf
+
 namespace esphome {
 namespace bl0942 {
 
@@ -12,33 +14,41 @@ static const uint8_t BL0942_FULL_PACKET = 0xAA;
 static const uint8_t BL0942_PACKET_HEADER = 0x55;
 
 static const uint8_t BL0942_WRITE_COMMAND = 0xA8;
-static const uint8_t BL0942_REG_I_FAST_RMS_CTRL = 0x10;
-static const uint8_t BL0942_REG_MODE = 0x18;
-static const uint8_t BL0942_REG_SOFT_RESET = 0x19;
-static const uint8_t BL0942_REG_USR_WRPROT = 0x1A;
+
+static const uint8_t BL0942_REG_I_RMSOS = 0x12;
+static const uint8_t BL0942_REG_WA_CREEP = 0x14;
+static const uint8_t BL0942_REG_I_FAST_RMS_TH = 0x15;
+static const uint8_t BL0942_REG_I_FAST_RMS_CYC = 0x16;
+static const uint8_t BL0942_REG_FREQ_CYC = 0x17;
+static const uint8_t BL0942_REG_OT_FUNX = 0x18;
+static const uint8_t BL0942_REG_MODE = 0x19;
+static const uint8_t BL0942_REG_SOFT_RESET = 0x1C;
+static const uint8_t BL0942_REG_USR_WRPROT = 0x1D;
 static const uint8_t BL0942_REG_TPS_CTRL = 0x1B;
 
-// TODO: Confirm insialisation works as intended
-const uint8_t BL0942_INIT[5][6] = {
-    // Reset to default
-    {BL0942_WRITE_COMMAND, BL0942_REG_SOFT_RESET, 0x5A, 0x5A, 0x5A, 0x38},
-    // Enable User Operation Write
-    {BL0942_WRITE_COMMAND, BL0942_REG_USR_WRPROT, 0x55, 0x00, 0x00, 0xF0},
-    // 0x0100 = CF_UNABLE energy pulse, AC_FREQ_SEL 50Hz, RMS_UPDATE_SEL 800mS
-    {BL0942_WRITE_COMMAND, BL0942_REG_MODE, 0x00, 0x10, 0x00, 0x37},
-    // 0x47FF = Over-current and leakage alarm on, Automatic temperature measurement, Interval 100mS
-    {BL0942_WRITE_COMMAND, BL0942_REG_TPS_CTRL, 0xFF, 0x47, 0x00, 0xFE},
-    // 0x181C = Half cycle, Fast RMS threshold 6172
-    {BL0942_WRITE_COMMAND, BL0942_REG_I_FAST_RMS_CTRL, 0x1C, 0x18, 0x00, 0x1B}};
+static const uint32_t BL0942_REG_MODE_RESV = 0x03;
+static const uint32_t BL0942_REG_MODE_CF_EN = 0x04;
+static const uint32_t BL0942_REG_MODE_RMS_UPDATE_SEL = 0x08;
+static const uint32_t BL0942_REG_MODE_FAST_RMS_SEL = 0x10;
+static const uint32_t BL0942_REG_MODE_AC_FREQ_SEL = 0x20;
+static const uint32_t BL0942_REG_MODE_CF_CNT_CLR_SEL = 0x40;
+static const uint32_t BL0942_REG_MODE_CF_CNT_ADD_SEL = 0x80;
+static const uint32_t BL0942_REG_MODE_UART_RATE_19200 = 0x200;
+static const uint32_t BL0942_REG_MODE_UART_RATE_38400 = 0x300;
+static const uint32_t BL0942_REG_MODE_DEFAULT =
+    BL0942_REG_MODE_RESV | BL0942_REG_MODE_CF_EN | BL0942_REG_MODE_CF_CNT_ADD_SEL;
+
+static const uint32_t BL0942_REG_SOFT_RESET_MAGIC = 0x5a5a5a;
+static const uint32_t BL0942_REG_USR_WRPROT_MAGIC = 0x55;
 
 void BL0942::loop() {
   DataPacket buffer;
   if (!this->available()) {
     return;
   }
-  if (read_array((uint8_t *) &buffer, sizeof(buffer))) {
-    if (validate_checksum(&buffer)) {
-      received_package_(&buffer);
+  if (this->read_array((uint8_t *) &buffer, sizeof(buffer))) {
+    if (this->validate_checksum_(&buffer)) {
+      this->received_package_(&buffer);
     }
   } else {
     ESP_LOGW(TAG, "Junk on wire. Throwing away partial message");
@@ -47,8 +57,8 @@ void BL0942::loop() {
   }
 }
 
-bool BL0942::validate_checksum(DataPacket *data) {
-  uint8_t checksum = BL0942_READ_COMMAND;
+bool BL0942::validate_checksum_(DataPacket *data) {
+  uint8_t checksum = BL0942_READ_COMMAND | this->address_;
   // Whole package but checksum
   uint8_t *raw = (uint8_t *) data;
   for (uint32_t i = 0; i < sizeof(*data) - 1; i++) {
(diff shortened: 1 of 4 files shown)


dwmw2:
No material changes to anything but bl0942; just making `CONF_LINE_FREQUENCY` common as CI demanded it.

esphome/components/bl0942/sensor.py
```diff
@@ -61,6 +68,10 @@
                 device_class=DEVICE_CLASS_FREQUENCY,
                 state_class=STATE_CLASS_MEASUREMENT,
             ),
+            cv.Optional(CONF_LINE_FREQUENCY, default="50HZ"): cv.enum(
+                LINE_FREQS, upper=True
+            ),
```
jesserockz:
```suggestion
            cv.Optional(CONF_LINE_FREQUENCY, default="50HZ"): cv.All(
                cv.frequency,
                cv.enum(
                    LINE_FREQS, int=True
                ),
            )
```

esphome/components/bl0942/sensor.py
```diff
@@ -1,27 +1,34 @@
...
     UNIT_VOLT,
     UNIT_WATT,
-    UNIT_HERTZ,
-    STATE_CLASS_TOTAL_INCREASING,
 )
 
+LINE_FREQS = {
+    "50HZ": 50,
+    "60HZ": 60,
+}
```
jesserockz: This needs to be moved down in this file
```suggestion
LineFrequency = bl0942_ns.enum("LineFrequency")
LINE_FREQS = {
    50: LineFrequency.LINE_FREQUENCY_50HZ,
    60: LineFrequency.LINE_FREQUENCY_60HZ,
}
```

esphome/components/bl0942/bl0942.h
```diff
@@ -35,9 +35,10 @@ class BL0942 : public PollingComponent, public uart::UARTDevice {
   void set_power_sensor(sensor::Sensor *power_sensor) { power_sensor_ = power_sensor; }
   void set_energy_sensor(sensor::Sensor *energy_sensor) { energy_sensor_ = energy_sensor; }
   void set_frequency_sensor(sensor::Sensor *frequency_sensor) { frequency_sensor_ = frequency_sensor; }
+  void set_line_freq(int freq) { line_freq_ = freq; }
```
jesserockz:
```suggestion
  void set_line_freq(LineFrequency freq) { this->line_freq_ = freq; }
```

esphome/components/bl0942/bl0942.h
```diff
@@ -35,9 +35,10 @@ class BL0942 : public PollingComponent, public uart::UARTDevice {
...
+  void set_address(int address) { address_ = address; }
```
jesserockz:
```suggestion
  void set_address(int address) { this->address_ = address; }
```

esphome/components/bl0942/bl0942.cpp
```diff
@@ -111,6 +158,11 @@ void BL0942::received_package_(DataPacket *data) {
 
 void BL0942::dump_config() {  // NOLINT(readability-function-cognitive-complexity)
   ESP_LOGCONFIG(TAG, "BL0942:");
+  ESP_LOGCONFIG(TAG, "  Address: %d", this->address_);
+  ESP_LOGCONFIG(TAG, "  Nominal line frequency: %d Hz", this->line_freq_);
+  int modereg = read_reg_(BL0942_REG_MODE);
+  if (modereg >= 0)
+    ESP_LOGCONFIG(TAG, "  Mode register: 0x%x", modereg);
```
jesserockz: Data transactions should not be done in `dump_config`. This is only to print out the configuration variables, or data that was previously gathered (in `setup` for example)

jesserockz (changes requested):

esphome/components/bl0942/bl0942.cpp
dwmw2: Thanks. I'll move it to `setup()` as a sanity check that the device is really there and we aren't just sending serial bytes into the void.

esphome/components/bl0942/bl0942.h
dwmw2:
```
src/esphome/components/bl0942/bl0942.h:38:22: error: 'LineFrequency' has not been declared
   38 |   void set_line_freq(LineFrequency freq) { this->line_freq_ = freq; }
      |                      ^~~~~~~~~~~~~
```
Do I have to define the corresponding enum in C++ code too, or should it be autogenerated somewhere?

dwmw2: Answered my own question by looking for other examples. I note the other examples of 'LineFrequency' don't do it this way; I had just copied from those.

dbuezas:
I unfortunately don't have my bl0942 bulb anymore, was this confirmed to work?
When I first made a custom component for it (which was used as a base for this integration and the openbeken one), the datasheet wasn't available so I reversed engineered the protocol.
It may well be that that initialization code was imperfect and this PR makes it fully compliant

dwmw2:
Yes, I'm reading back the mode register and checking it works, and have checked the order of the soft reset vs. write protect by checking when the mode is reset. Full details in the linked issue.

esphome/components/bl0942/bl0942.h
jesserockz: > I note the other examples of 'LineFrequency' don't do it this way; I had just copied from those.

I will add to my list to update the others.

dwmw2:
@dbuezas Where did you get your calibration data from?

```
static const float BL0942_PREF = 596;              // taken from tasmota
static const float BL0942_UREF = 15873.35944299;   // should be 73989/1.218
static const float BL0942_IREF = 251213.46469622;  // 305978/1.218
static const float BL0942_EREF = 3304.61127328;    // Measured
```
I'm tempted to make those calibration options, but that's not very helpful unless I can document for the user how to set them. _(And it would be great if we now know enough about decrypting the Tuya storage partitions that we could even find the factory settings!)._

As I understand it, the inputs of the "calibration-free IC" are a linear factor of the voltage and the current. The IC then measures the phase difference between the two, remembers everything I've forgotten about AC power theory, and uses it to calculate the actual power, as well as accumulating that to count total energy flow.

So calculating `UREF` and `IREF` ought to be as simple as applying an external meter and comparing with the raw values from the IC. And all that **I** need to remember about AC power theory is that the I and the V are on the top of the equation, so both power and energy factors ought to be proportional to (`UREF` * `IREF`) ?

dwmw2:
Let's deal with the calibration separately, in https://github.com/esphome/issues/issues/6155. This PR should be ready to go, and just addresses the setup issues.

dbuezas:
@dwmw2 

> Where did you get your calibration data from?

I'm sorry, I confused this with the BP5758, which is an LED driver. Looking at my PR for this one I see that PRef was from tasmota and the others just measured.

dwmw2:
@jesserockz I believe I've made the changes you asked for and clicked the 'Ready for review' button as directed, but it still seems to say '1 change requested'. Have I done something wrong?

Apologies for hassling, if it's just in your queue and you'll get round to it when you can.

dwmw2:
Hm, why does the pull request list still say this has 'Changes requested'?

Screenshot from 2024-08-16 18-14-48

esphome/components/bl0942/bl0942.cpp
```diff
@@ -12,32 +14,40 @@ static const uint8_t BL0942_FULL_PACKET = 0xAA;
...
 
 void BL0942::loop() {
   DataPacket buffer;
   if (!this->available()) {
     return;
   }
   if (read_array((uint8_t *) &buffer, sizeof(buffer))) {
-    if (validate_checksum(&buffer)) {
+    if (validate_checksum_(&buffer)) {
       received_package_(&buffer);
```
jesserockz:
```suggestion
    if (this->validate_checksum_(&buffer)) {
      this->received_package_(&buffer);
```

jesserockz (changes requested):
Please also add the line_frequency and address to 1 or more yaml tests.

esphome/components/bl0942/bl0942.h
```diff
@@ -28,16 +28,22 @@ struct DataPacket {
...
+
 class BL0942 : public PollingComponent, public uart::UARTDevice {
  public:
   void set_voltage_sensor(sensor::Sensor *voltage_sensor) { voltage_sensor_ = voltage_sensor; }
   void set_current_sensor(sensor::Sensor *current_sensor) { current_sensor_ = current_sensor; }
   void set_power_sensor(sensor::Sensor *power_sensor) { power_sensor_ = power_sensor; }
   void set_energy_sensor(sensor::Sensor *energy_sensor) { energy_sensor_ = energy_sensor; }
   void set_frequency_sensor(sensor::Sensor *frequency_sensor) { frequency_sensor_ = frequency_sensor; }
+  void set_line_freq(LineFrequency freq) { this->line_freq_ = freq; }
+  void set_address(int address) { this->address_ = address; }
```
jesserockz:
```suggestion
  void set_address(uint8_t address) { this->address_ = address; }
```

esphome/components/bl0942/bl0942.cpp
jesserockz:
```suggestion
      this->received_package_(&buffer);
```

esphome/components/bl0942/bl0942.cpp
```diff
@@ -12,32 +14,40 @@ static const uint8_t BL0942_FULL_PACKET = 0xAA;
...
+
+static const uint32_t BL0942_REG_SOFT_RESET_MAGIC = 0x5a5a5a;
+static const uint32_t BL0942_REG_USR_WRPROT_MAGIC = 0x55;
 
 void BL0942::loop() {
   DataPacket buffer;
   if (!this->available()) {
     return;
   }
   if (read_array((uint8_t *) &buffer, sizeof(buffer))) {
```
jesserockz:
```suggestion
  if (this->read_array((uint8_t *) &buffer, sizeof(buffer))) {
```

jesserockz (changes requested):

jesserockz:
@dwmw2 Please dont force push/rebase PRs once a review has started. It makes seeing followup changes more difficult and means the entire PR has to be reviewed again

dwmw2:
> @dwmw2 Please dont force push/rebase PRs once a review has started. It makes seeing followup changes more difficult and means the entire PR has to be reviewed again

Ack. Would you like me to undo that, or would that make it worse now?

jesserockz (approved):

jesserockz:
> Ack. Would you like me to undo that, or would that make it worse now?

It can't be fixed once done.

dwmw2:
OK. I have rebased #7299 though, as that was based on this one, and I don't believe the review had really started. Hope that's OK.

PR #7261, opened 2024-08-13
Add total_daily_energy mode:lifetime option

nkinnan:

Added an optional mode to the **total_daily_energy** component which allows it to never reset back to zero.  

In order to facilitate this, the variable **total_energy_** was renamed to **today_energy_** and then **lifetime_energy_** was added.  This allows the existing logic to remain almost completely unchanged, the only difference is, when mode is set to lifetime:
* If restore:true then the value is restored to **lifetime_energy_** instead of **today_energy_**
* On publish and save, **lifetime_energy_** is added back into **today_energy_** before processing the value
* When the day ticks over at midnight, **today_energy_** is still zeroed, but is first added into **lifetime_energy_**

This implementation also avoids the "adding a really small value to a really large value in floating point sometimes does nothing" issue.  It does so preserving the existing logic around **today_energy_** (was: **total_energy_**) by still zeroing it every day.  But then adds a "second level" of aggregation, once a day, into **lifetime_energy_**.  

I'm not sure if that issue with floating point is the reason why there's a **total_daily_energy** component but not a **total_energy** version?  Either way, this basically sidesteps the issue.  Sampling once per second, it would take 236 years for the difference in magnitude between **today_energy_** and **lifetime_energy_** to equal the magnitude difference between **[new reading to add]** and **today_energy_** just before midnight each day.  For lower sample rates that would be an even larger number of centuries.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4131

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

  - source:
      type: git
      ref: "add-tde-lifetime"
    components: [ total_daily_energy ]
    refresh: 60s


  - platform: total_daily_energy
    name: "Active Energy Resets to Zero At Midnight And On Reboot"
    power_id: active_power_sensor
    restore: false

  - platform: total_daily_energy
    name: "Active Energy Resets to Zero At Midnight"
    power_id: active_power_sensor
    restore: true

  - platform: total_daily_energy
    name: "Active Energy Lifetime"
    power_id: active_power_sensor
    restore: true
    mode: lifetime

```

## Checklist:

### DIFF
--- a/esphome/components/total_daily_energy/sensor.py
+++ b/esphome/components/total_daily_energy/sensor.py
@@ -8,6 +8,7 @@ from esphome.const import (
     CONF_TIME_ID,
     DEVICE_CLASS_ENERGY,
     CONF_METHOD,
+    CONF_MODE,
     STATE_CLASS_TOTAL_INCREASING,
     CONF_UNIT_OF_MEASUREMENT,
     CONF_ACCURACY_DECIMALS,
@@ -19,10 +20,15 @@ DEPENDENCIES = ["time"]
 CONF_POWER_ID = "power_id"
 total_daily_energy_ns = cg.esphome_ns.namespace("total_daily_energy")
 TotalDailyEnergyMethod = total_daily_energy_ns.enum("TotalDailyEnergyMethod")
+TotalDailyEnergyMode = total_daily_energy_ns.enum("TotalDailyEnergyMode")
 TOTAL_DAILY_ENERGY_METHODS = {
     "trapezoid": TotalDailyEnergyMethod.TOTAL_DAILY_ENERGY_METHOD_TRAPEZOID,
     "left": TotalDailyEnergyMethod.TOTAL_DAILY_ENERGY_METHOD_LEFT,
     "right": TotalDailyEnergyMethod.TOTAL_DAILY_ENERGY_METHOD_RIGHT,
+}
+TOTAL_DAILY_ENERGY_MODES = {
+    "daily": TotalDailyEnergyMode.TOTAL_DAILY_ENERGY_MODE_DAILY,
+    "lifetime": TotalDailyEnergyMode.TOTAL_DAILY_ENERGY_MODE_LIFETIME,
 }
 TotalDailyEnergy = total_daily_energy_ns.class_(
     "TotalDailyEnergy", sensor.Sensor, cg.Component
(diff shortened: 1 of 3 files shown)


jesserockz:
That would go against the name total\_**daily**\_energy...

We have an [integration](https://esphome.io/components/sensor/integration) sensor that doesnt reset daily. Is this not the same thing?

nkinnan:
I wasn't aware of that sensor, thank you for pointing me to it.  After taking a look, I have a couple thoughts that I'd like your feedback on:

1) **integration** only does one-level aggregation, so you'll run into the problem where the new delta being added to the saved value is so small that the floating point operation does nothing.  The saved value is so large that the delta is below it's floating point resolution (and you'll lose resolution even before hitting that point).  One of the reasons I created this PR is that I was doing the same thing as **integration** does with a global variable, and after a couple of weeks, it simply stopped changing value any more.

2) I could add similar logic to **integration** but it doesn't have a dependency on **time** like **total_daily_energy** does, and I'm not sure how I'd trigger the two-level aggregation.  

3) I agree that adding a "lifetime" mode to the *daily* sensor is a bit confusing.  I can think of a couple alternate options, what do you think?  Or do you have a different suggestion?
  
  A) Add a dependency of [time] to [integration] similar to **total_daily_energy**, to allow for the two level daily aggregation.
  
  B) Rename *total_daily_energy* to *total_energy**, then add back **total_daily_energy** as a wrapper of **total_energy** which sets mode to **daily** in order to preserve compatibility with the old name.

  D) Other?
  
Thoughts?

nkinnan:
Thanks for pointing me to that.  I think it does make my PR redundant.  Let me try it tomorrow and I will probably close.

nkinnan:
Yes, that appears to do the same thing, and due to using double instead of float it does not have the same resolution issues as **total_daily_energy** (when not resetting at midnight).  I will abort this PR and instead create one to update the **total_daily_energy** documentation to point to **integration** as non-resetting option.  Thanks again for the pointer.

nkinnan:
Doc update link to point to integration for this functionality: https://github.com/esphome/esphome-docs/pull/4135

PR #7263, opened 2024-08-13
remove extra number from pronto

### LINKED ISSUE #6147: Pronto IR dump is not conforming to standard
### The problem

When dumping Pronto data from the ``remote_receiver`` component the output is not valid according to the specs. Simplifyingly speaking the Pronto data consists of a header and two optional burst blocks. Each of these must contain an even number of data words, namely an on cycles word and an off cycles word.
While this is maintained in the component, an additional ``off word`` is added unconditionally [here](https://github.com/esphome/esphome/blob/2a70ef05d17c0645fd4e9023d6b1bc5c2ea078ca/esphome/components/remote_base/pronto_protocol.cpp#L222).
This makes the word count odd, causing f.i. IRScrutinizer to bail out with an error.

The addition of that word seems to be intending to add a safety gap between two IR bursts, but will fail doing that.

- as an off cycle always will follow an on cycle which is missing here, it will not be applied
- most IR remotes will add such off interval after their sent sequence anyway
- detecting an IR transmission will require the receiver to wait for some quiet time to detect the end of the signal. This is obviously done in ``remote_receiver``, so the last off cycle dumped usually is much longer already.

newest

Home Assistant Add-on

_No response_

ESP8266

### Board

_No response_

remote_receiver

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

ssieb:

The proto decoder dumps an extra number that makes the sequence invalid.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/6147

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/remote_base/pronto_protocol.cpp
+++ b/esphome/components/remote_base/pronto_protocol.cpp
@@ -201,9 +201,6 @@ std::string ProntoProtocol::compensate_and_dump_sequence_(const RawTimings &data
     out += dump_duration_(t_duration, timebase);
   }
 
-  // append minimum gap
-  out += dump_duration_(PRONTO_DEFAULT_GAP, timebase, true);
-
   return out;
 }
 


Miq1:
May I suggest a little more elaborate change?

When at the final element of the ``data`` array (that will be the final off cycle), check if ``t_duration`` is less than ``PRONTO_DEFAULT_GAP``. If it is, add the default to ``out`` instead.

This way you will preserve whatever gap the respective remote was using, except when it is dangerously short.

ssieb:
This change matches the other decoders.  They don't include a delay after.  That's up to the user to add when transmitting.

kbx81 (approved):

ssieb:
Some further research indicates that it's an arbitrary number.  For esphome, the last number ends up being the idle timeout value because that's where the receiver stops recording.  The `06C3` was then added on, but that doesn't make sense.  I found out that a certain arduino IR library does end with the `06C3`, but I expect that's probably just their timeout value as well.

PR #7291, opened 2024-08-15
[network] Always allow ``enable_ipv6: false``

### LINKED ISSUE #6144: IPv6 can't be enabled for libretiny on BK72XX platform
### The problem

```
Failed config

network: [source /home/dwmw/esphome/base.yaml:74]
  
  This feature is only available on ['esp32', 'esp8266', 'rp2040'].
  enable_ipv6: True
```
I have IPv6 enabled in my base config; I can turn it off for specific devices by overriding it to `False` again if I have to, but even *that* doesn't work here as it refuses even to parse the default configuration.

But aside from the premature failure, *why* doesn't it work with libretiny? Looking at libretiny documentation, I see hints that there is IPv6 support: https://docs.libretiny.eu/ltapi/class_wi_fi_class/#function-enableipv6

Is this specific to BK72xx rather than libretiny itself? https://github.com/esphome/esphome/pull/6387 seemed to block this configuration without any link to an upstream bug or feature request.

dev

pip

N/A

BK72XX

### Board

Tongou DIN rail power meter

network

### Example YAML snippet

```yaml
network:
  enable_ipv6: true
```

_No response_

### Additional information

_No response_

dwmw2:
I have a default configuration which uses IPv6, but some platforms only support Legacy IP. It doesn't suffice just to override enable_ipv6=false in their own config; we are never even allowed to *mention* IPv6 or the build fails.
Fix that, by adding a boolean_false validator. And put it first, so that the error the user actually sees if they *do* set it to true, is the one about it not being supported on that platform.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):**  fixes https://github.com/esphome/issues/issues/6144

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
network:
  enable_ipv6: false
```

## Checklist:

### DIFF
--- a/esphome/components/network/__init__.py
+++ b/esphome/components/network/__init__.py
@@ -18,13 +18,18 @@ IPAddress = network_ns.class_("IPAddress")
 
 CONFIG_SCHEMA = cv.Schema(
     {
+        cv.Optional(CONF_ENABLE_IPV6, default=False): cv.boolean,
         cv.SplitDefault(
             CONF_ENABLE_IPV6,
             esp8266=False,
             esp32=False,
             rp2040=False,
         ): cv.All(
-            cv.boolean, cv.only_on([PLATFORM_ESP32, PLATFORM_ESP8266, PLATFORM_RP2040])
+            cv.boolean,
+            cv.Any(
+                cv.only_on([PLATFORM_ESP32, PLATFORM_ESP8266, PLATFORM_RP2040]),
+                cv.boolean_false,
+            ),
         ),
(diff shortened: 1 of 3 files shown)


jesserockz (approved):

jesserockz:
Your particular use case can actually be solved without this PR =)

```yaml
network:
  enable_ipv6: !remove
```

jesserockz (dismissed):

PR #7299, opened 2024-08-16
Support BL0942 calibration

### LINKED ISSUE #6155: BL0942 calibration is wrong/inconsistent.
### The problem

All of Tasmota, OpenBeken, and ESPHome, appear to have values for the calibration which are pulled out of a hat. Ours are:
```
static const float BL0942_PREF = 596;              // taken from tasmota
static const float BL0942_UREF = 15873.35944299;   // should be 73989/1.218
static const float BL0942_IREF = 251213.46469622;  // 305978/1.218
static const float BL0942_EREF = 3304.61127328;    // Measured
```

I've spent a little time with the data sheet and trying to remember my high school physics, and come up with the following:
```
// The BL0942 IC is "calibration-free", which means that it doesn't care
// at all about calibration, and that's left to software. It measures a
// voltage differential on its IP/IN pins which linearly proportional to
// the current flow, and another on its VP pin which is proportional to
// the line voltage. It never knows the actual calibration; the values
// it reports are solely in terms of those inputs.
//
// The datasheet refers to the input voltages as I(A) and V(V), both
// in millivolts. It measures them against a reference voltage Vref,
// which is typically 1.218V (but that absolute value is meaningless
// without the actual calibration anyway).
//
// The reported I_RMS value is 305978 I(A)/Vref, and the reported V_RMS
// value is 73989 V(V)/Vref. So we can calibrate those by applying a
// simple meter with a resistive load.
//
// The chip also measures the phase difference between voltage and
// current, and uses it to calculate the power factor (cos φ). It
// reports the WATT value of 3537 * I_RMS * V_RMS * cos φ).
//
// It also integrates total energy based on the WATT value. The time for
// one CF_CNT pulse is 1638.4*256 / WATT.
//
// So... how do we calibrate that?
//
// Using a simple resistive load and an external meter, we can measure
// the true voltage and current for a given V_RMS and I_RMS reading,
// to calculate BL0942_UREF and BL0942_IREF. Those are in units of
// "305978 counts per amp" or "73989 counts per volt" respectively.
//
// We can derive BL0942_PREF from those. Let's eliminate the weird
// factors and express the calibration in plain counts per volt/amp:
// UREF1 = UREF/73989, IREF1 = IREF/305978.
//
// Next... the true power in Watts is V * I * cos φ, so that's equal
// to WATT/3537 * IREF1 * UREF1. Which means
// BL0942_PREF = BL0942_UREF * BL0942_IREF * 3537 / 305978 / 73989.
//
// Finally the accumulated energy. The period of a CF_CNT count is
// 1638.4*256 / WATT seconds, or 419230.4 / WATT seconds. Which means
// the energy represented by a CN_CNT pulse is 419230.4 WATT-seconds.
// Factoring in the calibration, that's 419230.4 / BL0942_PREF actual
// Watt-seconds (or Joules, as the physicists like to call them).
//
// But we're not being physicists today; we we're being engineers, so
// we want to convert to kWh instead. Which we do by dividing by 1000
// and then by 3600, so the energy in kWh is
// CF_CNT * 419230.4 / BL0942_PREF / 3600000
//
// Which makes BL0942_EREF = BL0942_PREF * 3600000 / 419430.4
```

Looking at the values used in Tasmota and OpenBeken, they seem within a reasonable amount of my assertion that PREF = UREF * IREF * 3537 / 305978 / 73989. Ours is a bit off (should be 623.0 but we have 596).

I propose that we make UREF and IREF configurable in our bl0942 driver, and make it derive EREF and PREF values from them as described above (once I confirm empirically). Should I bother with a backward compatible mode to retain the previous values for legacy users who might not care about the truth, but only *relative* power consumption in their own environment?

Perhaps it could default to the existing values if neither UREF nor IREF are given in the config, but derive EREF/PREF if the UREF/IREF calibration values *are* specified?

dev

pip

_No response_

BK72XX

### Board

Tongou DIN rail power meter

bl0942

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

dwmw2:

Calibration of BL0942 power sensor

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/6155
**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** https://github.com/esphome/esphome-docs/pull/4163

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
  - platform: bl0942
    uart_id: uart_bus
    line_frequency: 50Hz
    address: 0
    update_interval: 20s
    current:
      name: ${tongou_name} Current
    voltage:
      name: ${tongou_name} Voltage
    power:
      name: ${tongou_name} Power
      filters:
        multiply: -1
    energy:
      name: ${tongou_name} Energy
    frequency:
      name: ${tongou_name} Frequency
      accuracy_decimals: 2
    voltage_reference: ${voltage_ref}
    current_reference: ${current_ref}
    power_reference: ${power_ref}
    energy_reference: ${energy_ref}

```

## Checklist:


dwmw2:
This is on top of the simpler bug fix in https://github.com/esphome/esphome/pull/7250

esphome/components/bl0942/bl0942.cpp
```diff
@@ -104,13 +155,15 @@ void BL0942::received_package_(DataPacket *data) {
...
            watt, cf_cnt, total_energy_consumption, frequency, data->status);
 }
 
 void BL0942::dump_config() {  // NOLINT(readability-function-cognitive-complexity)
   ESP_LOGCONFIG(TAG, "BL0942:");
+  ESP_LOGCONFIG(TAG, "  Address: %d", this->address_);
+  ESP_LOGCONFIG(TAG, "  Nominal line frequency: %d Hz", this->line_freq_);
   LOG_SENSOR("", "Voltage", this->voltage_sensor_);
   LOG_SENSOR("", "Current", this->current_sensor_);
   LOG_SENSOR("", "Power", this->power_sensor_);
```
descipher, line 207: LOG_SENSOR("", "Energy", this->energy_sensor_);
  LOG_SENSOR("", "frequency", this->frequency_sensor_);

frequency can be capitalized

dwmw2:
@dbuezas Are you OK for me to mark myself as a co-owner for this component?

esphome/components/atm90e32/sensor.py
```diff
@@ -7,6 +7,7 @@
     CONF_FORWARD_ACTIVE_ENERGY,
     CONF_FREQUENCY,
     CONF_ID,
+    CONF_LINE_FREQUENCY,
```
CircuitSetup: Im unsure why this needs to be changed. Care to explain?

dwmw2: The CI complained at me that as I was adding the third instance of `CONF_LINE_FREQUENCY` I now needed to consolidate them into `const.py`

descipher: It's just CI forced cleanup, the two constants are identical so it's prefered to use the const.py defined one vs creating a local const. with the same or possibly overriding one.

CircuitSetup: got it, thanks!

esphome/components/bl0942/bl0942.cpp
dwmw2: Done

esphome/components/atm90e32/sensor.py
jesserockz: I have made https://github.com/esphome/esphome/pull/7351 to remove those changes from this PR

jesserockz:
Coverted to draft until https://github.com/esphome/esphome/pull/7250 is merged

dwmw2:
Added calibration fields to YAML tests.

esphome/components/bl0942/sensor.py
```diff
@@ -106,3 +115,11 @@ async def to_code(config):
         cg.add(var.set_frequency_sensor(sens))
     cg.add(var.set_line_freq(config[CONF_LINE_FREQUENCY]))
     cg.add(var.set_address(config[CONF_ADDRESS]))
+    if config.get(CONF_CURRENT_REFERENCE):
+        cg.add(var.set_current_reference(config[CONF_CURRENT_REFERENCE]))
```
jesserockz:
```suggestion
    if (current_reference := config.get(CONF_CURRENT_REFERENCE, None)) is not None:
        cg.add(var.set_current_reference(current_reference))
```

esphome/components/bl0942/bl0942.cpp
```diff
@@ -109,6 +109,18 @@ void BL0942::update() {
...
+  // If either current or voltage references are set explicitly by the user,
+  // calculate the power reference from it unless that is also explicitly set.
+  if ((this->current_reference_set_ || this->voltage_reference_set_) && !this->power_reference_set_) {
+    this->power_reference_ = (this->voltage_reference_ * this->current_reference_ * 3537.0 / 305978.0) / 73989.0;
+    this->power_reference_set_ = true;
+  }
+
+  // Similarly for energy reference, if the power reference was set by the user
+  // either implicitly or explicitly.
+  if (!this->power_reference_set_)
```
jesserockz:
```suggestion
  if (this->power_reference_set_ && !this->energy_reference_set_)
```

esphome/components/bl0942/bl0942.cpp
```diff
@@ -109,6 +109,18 @@ void BL0942::update() {
...
+    this->energy_reference_ = this->power_reference_ * 3600000 / 419430.4;
```
jesserockz, line 135:
```suggestion
    this->energy_reference_ = this->power_reference_ * 3600000 / 419430.4;
    this->energy_reference_set_ = true
```

jesserockz (changes requested):

esphome/components/bl0942/bl0942.cpp
dwmw2: Nobody cares if it's set any more, but I suppose it makes sense. Will need `{}` around it though.

esphome/components/bl0942/sensor.py
```diff
@@ -106,3 +115,11 @@ async def to_code(config):
         cg.add(var.set_frequency_sensor(sens))
     cg.add(var.set_line_freq(config[CONF_LINE_FREQUENCY]))
     cg.add(var.set_address(config[CONF_ADDRESS]))
+    if (current_reference := config.get(CONF_CURRENT_REFERENCE, None)) is not None:
```
jesserockz, line 118: The same changes need to be applied to the 3 below too =)

jesserockz (changes requested):

esphome/components/bl0942/sensor.py
dwmw2: ~~Does that make a difference? When I omit those parameters from the yaml configuration, nothing shows up in main.cpp for them. They already get elided somewhere, it seems?~~

Ah, it's just a style thing about *how* we check for `None`, and only doing the lookup once. More coffee...

jesserockz (approved):

esphome/components/bl0942/bl0942.cpp
```diff
@@ -122,6 +122,20 @@ void BL0942::update() {
...
+    this->power_reference_ = (this->voltage_reference_ * this->current_reference_ * 3537.0 / 305978.0) / 73989.0;
+    this->power_reference_set_ = true;
+  }
+
+  // Similarly for energy reference, if the power reference was set by the user
+  // either implicitly or explicitly.
+  if (this->power_reference_set_ && !this->energy_reference_set_) {
+    this->energy_reference_ = this->power_reference_ * 3600000 / 419430.4;
+    this->energy_reference_set_ = true;
+  }
```
jesserockz, line 137: Maybe for another PR in the future if you are up for it, it would be better to set the defaults in python, and do this calculation in python so at runtime it will always have the values set.

PR #7301, opened 2024-08-17
Add CSE7766 reactive power

nkinnan:

Adds a missing Reactive Power sensor under the CSE7766 component.

It's possible to generate the ReactivePower value with a template sensor, but it's update rate will not be synced to the CSE7766 readings and it would be less accurate than integrating this sensor output into the CSE7766 component directly.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4129

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

external_components:

  - source:
      type: git
      ref: "add-cse-7766-reactive-power"
    components: [ cse7766 ]
    refresh: 60s

sensor: 

  - platform: cse7766

    reactive_power:
      name: "Reactive Power"
      id: reactive_power_sensor
      accuracy_decimals: 1
      unit_of_measurement: VAR
      device_class: reactive_power
```

## Checklist:


esphome/components/cse7766/cse7766.cpp
```diff
@@ -147,6 +147,7 @@ void CSE7766Component::parse_data_() {
   float power = 0.0f;
   if (power_cycle_exceeds_range) {
     // Datasheet: power cycle exceeding range means active power is 0
+    have_power = true;
```
nkinnan, line 150: We do have a power reading here: 0 per the code comment immediately above this line

esphome/components/cse7766/cse7766.cpp
```diff
@@ -178,6 +179,16 @@ void CSE7766Component::parse_data_() {
     if (this->apparent_power_sensor_ != nullptr) {
       this->apparent_power_sensor_->publish_state(apparent_power);
     }
+    if (have_power && this->reactive_power_sensor_ != nullptr) {
+      const float reactive_power = apparent_power - power;
+      if (reactive_power < 0.0f) {
```
nkinnan, line 184: Sometimes the power reading is zero, but V*A is still positive, which causes negative reactive power to be calculated.  So this check is needed for the same reason as all the checks around invalid power factor calculations.

nkinnan: I wish I knew the best way to "catch" this situation and zero something upstream out, but the chip is apparently reporting readings that are (together) impossible.  If I find out more I'll do another PR if it's fixable since this issue is applicable to all readings.  For example another thing I see is after switching a high power load off, active power shows as 0 but apparent power is still high, and it sits in that state for a few seconds - presumably since no new readings are coming in since the load is physically disconnected and there is nothing to report.

nkinnan:
Clang seems to have dumped core while processing this pull request and I'm not sure how to kick it.

kbx81 (approved):

PR #7304, opened 2024-08-17
[speaker] Fix header includes

nielsnl68:

It seems that using the speaker component together with the adf_pipeline external component from @gnumpi the speaker component did not compile correctly. It was just missing 2 includes.

Strangely it works fine using the i2s_audio component directly.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/speaker/speaker.h
+++ b/esphome/components/speaker/speaker.h
@@ -1,5 +1,8 @@
 #pragma once
 
+#include "esphome/core/entity_base.h"
+#include "esphome/core/helpers.h"
+
 namespace esphome {
 namespace speaker {
 
--- a/esphome/components/speaker/speaker.h
+++ b/esphome/components/speaker/speaker.h
@@ -1,7 +1,8 @@
 #pragma once
 
-#include "esphome/core/entity_base.h"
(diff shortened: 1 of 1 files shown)


esphome/components/speaker/speaker.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+#include "esphome/core/entity_base.h"
```
jesserockz: Speaker is not an entity, this include makes no sense at all.

esphome/components/speaker/speaker.h
```diff
@@ -1,5 +1,8 @@
...
+#include "esphome/core/helpers.h"
```
jesserockz: This file is not using anything from the helper file so this include doesn't make sense

jesserockz (changes requested):

jesserockz (approved):

PR #7306, opened 2024-08-18
[i2s_audio] Add more options to speakers and microphones

pyos:

Now I2S speakers also have the following options that already exist for microphones:

 * `i2s_mode` (primary/secondary; default: primary)
 * `use_apll` (true/false, default: false)
 * `channel` (left/right/mono/stereo, default: mono)
 * `sample_rate` (default: 16000)
 * `bits_per_sample` (default: 16bit)

Both speakers and microphones also have:

 * `bits_per_channel` (default: equal to `bits_per_sample`)

`channel` has new values:

 * stereo: input/output data is interpreted as a sequence of (right, left) pairs (this option works for both speakers and microphones);
 * mono: for speakers, this is the old behavior - each sample from the input is sent to both left and right channels.

Also, speakers have an extra option:

 * `timeout` (default: 100ms) - how long the background tasks should wait before releasing the bus if the buffer is empty.

The `mode` option of speakers was removed, because it didn't actually do anything.

The `pdm` option for microphones is now optional and defaults to `false`.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4166

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
i2s_audio:
- i2s_lrclk_pin: 26
  i2s_bclk_pin: 27
  id: i2so
- i2s_lrclk_pin: 33 # WS
  i2s_bclk_pin: 32 # SCK
  id: i2si

microphone:
- platform: i2s_audio
  id: sound_in
  adc_type: external
  i2s_din_pin: 25 # SD
  i2s_audio_id: i2si
  sample_rate: 24000
  bits_per_sample: 32bit
  channel: right
  on_data:
    lambda: |-
      id(sound_out)->play((const uint8_t*)x.data(), x.size() * sizeof(x[0]));

speaker:
- platform: i2s_audio
  id: sound_out
  dac_type: external
  i2s_dout_pin: 14
  i2s_audio_id: i2so
  sample_rate: 24000
  bits_per_sample: 32bit
  channel: mono
  timeout: 500ms
```

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/__init__.py
+++ b/esphome/components/i2s_audio/__init__.py
@@ -30,6 +30,10 @@ CONF_I2S_MODE = "i2s_mode"
 CONF_PRIMARY = "primary"
 CONF_SECONDARY = "secondary"
 
+CONF_USE_APLL = "use_apll"
+CONF_BITS_PER_SAMPLE = "bits_per_sample"
+CONF_BITS_PER_CHANNEL = "bits_per_channel"
+CONF_MONO = "mono"
 CONF_LEFT = "left"
 CONF_RIGHT = "right"
 CONF_STEREO = "stereo"
@@ -46,6 +50,30 @@ i2s_mode_t = cg.global_ns.enum("i2s_mode_t")
 I2S_MODE_OPTIONS = {
     CONF_PRIMARY: i2s_mode_t.I2S_MODE_MASTER,  # NOLINT
     CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,  # NOLINT
+}
+
+i2s_channel_fmt_t = cg.global_ns.enum("i2s_channel_fmt_t")
+I2S_CHANNELS = {
+    CONF_MONO: i2s_channel_fmt_t.I2S_CHANNEL_FMT_ALL_LEFT,
+    CONF_LEFT: i2s_channel_fmt_t.I2S_CHANNEL_FMT_ONLY_LEFT,
+    CONF_RIGHT: i2s_channel_fmt_t.I2S_CHANNEL_FMT_ONLY_RIGHT,
+    CONF_STEREO: i2s_channel_fmt_t.I2S_CHANNEL_FMT_RIGHT_LEFT,
+}
+
+i2s_bits_per_sample_t = cg.global_ns.enum("i2s_bits_per_sample_t")
+I2S_BITS_PER_SAMPLE = {
+    8: i2s_bits_per_sample_t.I2S_BITS_PER_SAMPLE_8BIT,
+    16: i2s_bits_per_sample_t.I2S_BITS_PER_SAMPLE_16BIT,
+    24: i2s_bits_per_sample_t.I2S_BITS_PER_SAMPLE_24BIT,
+    32: i2s_bits_per_sample_t.I2S_BITS_PER_SAMPLE_32BIT,
+}
+
+i2s_bits_per_chan_t = cg.global_ns.enum("i2s_bits_per_chan_t")
+I2S_BITS_PER_CHANNEL = {
+    8: i2s_bits_per_chan_t.I2S_BITS_PER_CHAN_8BIT,
+    16: i2s_bits_per_chan_t.I2S_BITS_PER_CHAN_16BIT,
+    24: i2s_bits_per_chan_t.I2S_BITS_PER_CHAN_24BIT,
+    32: i2s_bits_per_chan_t.I2S_BITS_PER_CHAN_32BIT,
 }
 
 # https://github.com/espressif/esp-idf/blob/master/components/soc/{variant}/include/soc/soc_caps.h
@@ -70,14 +98,12 @@ I2S_BITS_PER_SAMPLE = {
     32: i2s_bits_per_sample_t.I2S_BITS_PER_SAMPLE_32BIT,
 }
 
-INTERNAL_ADC_VARIANTS = [VARIANT_ESP32]
-PDM_VARIANTS = [VARIANT_ESP32, VARIANT_ESP32S3]
-
 _validate_bits = cv.float_with_unit("bits", "bit")
 
 
 def i2s_audio_component_schema(
     class_: MockObjClass,
+    *,
     default_sample_rate: int,
     default_channel: str,
     default_bits_per_sample: str,
@@ -96,6 +122,10 @@ def i2s_audio_component_schema(
             cv.Optional(CONF_I2S_MODE, default=CONF_PRIMARY): cv.enum(
                 I2S_MODE_OPTIONS, lower=True
             ),
+            cv.Optional(CONF_USE_APLL, default=False): cv.boolean,
+            cv.Optional(CONF_BITS_PER_CHANNEL): cv.All(
+                cv.float_with_unit("bits", "bit"), cv.enum(I2S_BITS_PER_CHANNEL)
+            ),
         }
     )
 
@@ -107,6 +137,11 @@ async def register_i2s_audio_component(var, config):
     cg.add(var.set_channel(config[CONF_CHANNEL]))
     cg.add(var.set_sample_rate(config[CONF_SAMPLE_RATE]))
     cg.add(var.set_bits_per_sample(config[CONF_BITS_PER_SAMPLE]))
+    bits_per_channel = config.get(
+        CONF_BITS_PER_CHANNEL, i2s_bits_per_chan_t.I2S_BITS_PER_CHAN_DEFAULT
+    )
+    cg.add(var.set_bits_per_channel(bits_per_channel))
+    cg.add(var.set_use_apll(config[CONF_USE_APLL]))
 
 
 CONFIG_SCHEMA = cv.Schema(
--- a/esphome/components/i2s_audio/__init__.py
+++ b/esphome/components/i2s_audio/__init__.py
@@ -78,6 +78,7 @@ I2S_BITS_PER_SAMPLE = {
 
 i2s_bits_per_chan_t = cg.global_ns.enum("i2s_bits_per_chan_t")
 I2S_BITS_PER_CHANNEL = {
+    "default": i2s_bits_per_chan_t.I2S_BITS_PER_CHAN_DEFAULT,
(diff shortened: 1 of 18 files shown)


esphome/components/i2s_audio/__init__.py
```diff
@@ -42,6 +50,30 @@
     CONF_SECONDARY: i2s_mode_t.I2S_MODE_SLAVE,  # NOLINT
 }
 
+i2s_channel_fmt_t = cg.global_ns.enum("i2s_channel_fmt_t")
+I2S_CHANNELS = {
+    CONF_MONO: i2s_channel_fmt_t.I2S_CHANNEL_FMT_ALL_LEFT,
```
nielsnl68, line 65: In what case would you suggest this is useful?  I have played with the same ideas as you and i also already added some PR's for some suggested changes as well. 
In the end i saw no use case to have the option for  `i2s_channel_fmt_t.I2S_CHANNEL_FMT_ALL_LEFT,` available. only the other 3 are interesting.

esphome/components/i2s_audio/__init__.py
```diff
@@ -60,6 +92,27 @@
...
+            ),
+            cv.Optional(CONF_USE_APLL, default=False): cv.boolean,
+            cv.Optional(CONF_CHANNEL, default=default_channel): cv.enum(I2S_CHANNELS),
+            cv.Optional(CONF_SAMPLE_RATE, default=16000): cv.int_range(min=1),
+            cv.Optional(CONF_BITS_PER_SAMPLE, default=default_bits_per_sample): cv.All(
+                cv.float_with_unit("bits", "bit"), cv.enum(I2S_BITS_PER_SAMPLE)
+            ),
+            cv.Optional(CONF_BITS_PER_CHANNEL): cv.All(
+                cv.float_with_unit("bits", "bit"), cv.enum(I2S_BITS_PER_CHANNEL)
+            ),
```
nielsnl68: Just not sure we should offer this. imho it makes everything a bit difficult. But when you see a really use case. Then i don't mind.

esphome/components/i2s_audio/microphone/__init__.py
```diff
@@ -88,7 +62,7 @@ def validate_esp32_variant(config):
             "external": BASE_SCHEMA.extend(
                 {
                     cv.Required(CONF_I2S_DIN_PIN): pins.internal_gpio_input_pin_number,
-                    cv.Required(CONF_PDM): cv.boolean,
+                    cv.Optional(CONF_PDM, default=False): cv.boolean,
```
nielsnl68: In my honest opinion this should be an option for the ADC_TYPE enum like:
internal, standard, PDM

esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
```diff
@@ -58,7 +58,7 @@ void I2SAudioMicrophone::start_() {
       .tx_desc_auto_clear = false,
       .fixed_mclk = 0,
       .mclk_multiple = I2S_MCLK_MULTIPLE_256,
-      .bits_per_chan = I2S_BITS_PER_CHAN_DEFAULT,
+      .bits_per_chan = this->bits_per_channel_,
```
nielsnl68, line 61: look at the remark above about this option.

esphome/components/i2s_audio/speaker/__init__.py
```diff
@@ -44,28 +44,34 @@ def validate_esp32_variant(config):
     return config
 
 
+BASE_SCHEMA = i2s_component_schema(
+    speaker.SPEAKER_SCHEMA,
+    I2SAudioSpeaker,
+    default_channel=CONF_MONO,
```
nielsnl68: My suggestion is to use "CONF_RIGHT" here.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -114,50 +129,48 @@ void I2SAudioSpeaker::player_task(void *params) {
   event.type = TaskEventType::STARTED;
   xQueueSend(this_speaker->event_queue_, &event, portMAX_DELAY);
 
-  int16_t buffer[BUFFER_SIZE / 2];
+  int32_t buffer[BUFFER_SIZE];
 
   while (true) {
-    if (xQueueReceive(this_speaker->buffer_queue_, &data_event, 100 / portTICK_PERIOD_MS) != pdTRUE) {
+    if (xQueueReceive(this_speaker->buffer_queue_, &data_event, this_speaker->timeout_) != pdTRUE) {
```
nielsnl68:
```suggestion
    if (xQueueReceive(this_speaker->buffer_queue_, &data_event, this_speaker->timeout_ / portTICK_PERIOD_MS) != pdTRUE) {
```

esphome/components/i2s_audio/speaker/i2s_audio_speaker.h
```diff
@@ -2,7 +2,7 @@
 
 #ifdef USE_ESP32
 
-#include "../i2s_audio.h"
+#include "esphome/components/i2s_audio/i2s_audio.h"
```
nielsnl68: ?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.h
```diff
@@ -38,18 +37,18 @@ struct DataEvent {
...
 
   void setup() override;
   void loop() override;
 
+  void set_timeout(uint32_t ms) { this->timeout_ = ms / portTICK_PERIOD_MS; }
   void set_dout_pin(uint8_t pin) { this->dout_pin_ = pin; }
 #if SOC_I2S_SUPPORTS_DAC
   void set_internal_dac_mode(i2s_dac_mode_t mode) { this->internal_dac_mode_ = mode; }
 #endif
-  void set_external_dac_channels(uint8_t channels) { this->external_dac_channels_ = channels; }
```
nielsnl68: where did you moved this to?

nielsnl68 (changes requested):
It looks like we had almost the same ideas, already. 
see: 
https://github.com/esphome/esphome/pull/7183
https://github.com/esphome/esphome/pull/7137

esphome/components/i2s_audio/speaker/i2s_audio_speaker.h
pyos: nowhere, it was not used

esphome/components/i2s_audio/speaker/i2s_audio_speaker.h
pyos: copied from a custom_components directory, forgot to revert

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
pyos: already divided in the setter

esphome/components/i2s_audio/speaker/__init__.py
pyos: CONF_MONO is equivalent to the current behavior

esphome/components/i2s_audio/__init__.py
pyos: technically INMP441 for example has 24-bit data in 32-bit channels, but since 32-bit data mode discards the low 8 bits anyway it doesn't really matter

esphome/components/i2s_audio/__init__.py
pyos: this mode plays the same sound on both channels, so it's useful for switching the same 2-amplifier/speaker arrangement between mono and stereo mode in software (you can call speaker.set_channel in a lambda while it's not playing)

pyos: in your PRs my `channel: mono` is your `use_16bit_mode: true` with `channel: stereo`, I just put it as a separate `channel` option because combining `use_16bit_mode: true` with `channel: left` and `channel: right` is wrong

esphome/components/i2s_audio/speaker/i2s_audio_speaker.h
nielsnl68: Ha yes, this is one of the phantom variable @jesserockz added to the speaker component :P

esphome/components/i2s_audio/speaker/__init__.py
nielsnl68: I know, but still i do think it is not necessary

esphome/components/i2s_audio/__init__.py
nielsnl68: > in your PRs my channel: mono is your use_16bit_mode: true with channel: stereo, I just put it as a separate channel option because combining use_16bit_mode: true with channel: left and channel: right is wrong

True that was a mistake in later work i removed this option in favor of I2S_CHANNELS like you do.

jesserockz:
I have rebased this PR after merging #7183. The tests compile but I have not actually looked at the code or reviewed or tested yet.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -56,6 +56,21 @@ void I2SAudioSpeaker::start_() {
...
+    return reinterpret_cast<const uint8_t *>(from);
+  }
+  const b *result = to;
+  for (size_t i = 0; i < bytes; i += sizeof(a)) {
+    b value = static_cast<b>(*from++) << (sizeof(b) - sizeof(a)) * 8;
+    *to++ = value;
+    if (repeat)
+      *to++ = value;
+  }
+  bytes *= (sizeof(b) / sizeof(a)) * (repeat ? 2 : 1);
```
jesserockz: AFAICT this should be safe
https://clang.llvm.org/extra/clang-tidy/checks/bugprone/sizeof-expression.html#suspicious-usage-of-sizeof-sizeof
```suggestion
  bytes *= (sizeof(b) / sizeof(a)) * (repeat ? 2 : 1);  // NOLINT
```

jesserockz (approved):

PR #7312, opened 2024-08-19
[Modbus Controller] Added preference to change command retries

0x3333:

Currently, if a command fails (no response received), it will retry 5 times. This needs to be configurable by the user; otherwise, in fast environments, it could cause some trouble.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4202

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

modbus_controller:
- id: modbus_device
  address: 0x1   ## address of the Modbus slave device on the bus
  modbus_id: modbus1
  setup_priority: -10
  max_cmd_retries: 1
```

## Checklist:


esphome/components/modbus_controller/modbus_controller.h
```diff
@@ -490,6 +493,8 @@ class ModbusController : public PollingComponent, public modbus::ModbusDevice {
   bool module_offline_;
   /// how many updates to skip if module is offline
   uint16_t offline_skip_updates_;
+  /// how many times we'll retry a command that got no response
+  uint8_t max_cmd_retries_;
```
clydebarrow:
```suggestion
  uint8_t max_cmd_retries_{5};
```
If it's not set in the constructor, it should be initialised to something, even though in the usual case this will get overwritten later.

esphome/components/modbus_controller/__init__.py
```diff
@@ -254,6 +256,8 @@ async def add_modbus_base_properties(
 async def to_code(config):
     var = cg.new_Pvariable(config[CONF_ID])
     cg.add(var.set_command_throttle(config[CONF_COMMAND_THROTTLE]))
+    if CONF_MAX_CMD_RETRIES in config:
+        cg.add(var.set_max_cmd_retries(config[CONF_MAX_CMD_RETRIES]))
```
clydebarrow:
```suggestion
    cg.add(var.set_max_cmd_retries(config[CONF_MAX_CMD_RETRIES]))
```
It will always be set since a default is provided.

esphome/components/modbus_controller/modbus_controller.cpp
```diff
@@ -22,7 +22,7 @@ bool ModbusController::send_next_command_() {
     auto &command = command_queue_.front();
 
     // remove from queue if command was sent too often
-    if (command->send_countdown < 1) {
+    if (command->send_counter >= this->max_cmd_retries_) {
```
clydebarrow:
```suggestion
    if (command->send_counter == this->max_cmd_retries_) {
```
Can't ever be greater. But this logic could be moved into the `ModbusCommandItem` class so `send_counter` becomes protected, and gets modified only within the class. Either that or move the increment of the counter into this function - the logic should be confined to one class to make it easier to understand and validate.

esphome/components/modbus_controller/__init__.py
```diff
@@ -129,6 +130,7 @@
             cv.Optional(
                 CONF_COMMAND_THROTTLE, default="0ms"
             ): cv.positive_time_period_milliseconds,
+            cv.Optional(CONF_MAX_CMD_RETRIES, default=5): cv.positive_int,
```
clydebarrow:
```suggestion
            cv.Optional(CONF_MAX_CMD_RETRIES, default=5): cv.int_range(min=1),
```
A value of zero would mean the command is never sent. I'm assume that's not meant to be an option.

esphome/components/modbus_controller/modbus_controller.cpp
```diff
@@ -559,7 +561,7 @@ bool ModbusCommandItem::send() {
     modbusdevice->send_raw(this->payload);
   }
   ESP_LOGV(TAG, "Command sent %d 0x%X %d", uint8_t(this->function_code), this->register_address, this->register_count);
-  send_countdown--;
+  send_counter++;
```
clydebarrow: Either move the test into this class, or move this increment into the controller class.

clydebarrow (changes requested):
Also note that the convention is that class properties like `command_queue_` should always be referenced as `this->command_queue_`.

esphome/components/modbus_controller/modbus_controller.h
0x3333: I moved the initialization to the generated code, but I got your point, will fix. Is there any instrumentation that can "sync" both values? The one in C++ and the one in Python?

esphome/components/modbus_controller/__init__.py
0x3333: Got it wrong, will fix.

esphome/components/modbus_controller/__init__.py
0x3333: This is a little confusing I admit, but if this is a retry, I could have this as zero. Because if the command is sent and receives a response, there is no retry per se. What are your thoughts about it?

0x3333: I'll keep the way it is, and make it clear in the documentation that this number includes the initial sent. If you agree, of course.

clydebarrow: Sure, though I can't imagine why anyone would want to specify this and never have it sent. An alternative would be to allow it to be zero, but add 1 when setting to allow for the initial send. That would make sense in that the "retries" value is just that.

0x3333: The issue is just the nomenclature; if we say this is the retry count, the initial transmission is not a retry, so it should not be counted. I'll add 1 and make sure the documentation makes it clear.

0x3333:
> Also note that the convention is that class properties like `command_queue_` should always be referenced as `this->command_queue_`.

Despite the fact that this PR is not related to this, but I added `this->` to all `command_queue_` calls.

esphome/components/modbus_controller/modbus_controller.h
```diff
@@ -458,6 +462,13 @@ class ModbusController : public PollingComponent, public modbus::ModbusDevice {
   bool get_module_offline() { return module_offline_; }
   /// Set callback for commands
   void add_on_command_sent_callback(std::function<void(int, int)> &&callback);
+  /// called by esphome generated code to set the max_cmd_retries.
+  void set_max_cmd_retries(uint8_t max_cmd_retries) {
+    // We add 1 so the initial transmission doesn't count when checking against command->send_counter.
+    this->max_cmd_retries_ = max_cmd_retries + 1;
```
clydebarrow: Only problem here is that `get_max_cmd_retries()` will return a different number to what was set.
```suggestion
    this->max_cmd_retries_ = max_cmd_retries;
```

esphome/components/modbus_controller/modbus_controller.h
```diff
@@ -322,9 +321,9 @@ class ModbusCommandItem {
       on_data_func;
   std::vector<uint8_t> payload = {};
   bool send();
-  // wrong commands (esp. custom commands) can block the send queue
-  // limit the number of repeats
-  uint8_t send_countdown{MAX_SEND_REPEATS};
+  /// Check if the command should be retried based on the max_send_count parameter
+  bool should_retry(uint8_t max_send_count) { return this->send_count_ < max_send_count; };
```
clydebarrow: Change the test so the first one always gets sent.
```suggestion
  bool should_retry(uint8_t max_retries) { return this->send_count_ <= max_retries; };
```

clydebarrow (changes requested):
Nearly there. Otherwise looks fine.

esphome/components/modbus_controller/modbus_controller.h
0x3333: In this case we need `max_cmd_retries_` to be 4, but it would represent 5 transmissions. Am I right? So I can update the documentation.

clydebarrow: Correct. One initial send, 4 retries. On the basis that there will always be the initial send.

esphome/components/modbus_controller/modbus_controller.h
```diff
@@ -490,6 +498,8 @@ class ModbusController : public PollingComponent, public modbus::ModbusDevice {
   bool module_offline_;
   /// how many updates to skip if module is offline
   uint16_t offline_skip_updates_;
+  /// How many times we will send a command if we get no response(Includes the first one).
```
clydebarrow:
```suggestion
  /// How many times we will retry a command if we get no response
```

clydebarrow (dismissed):

clydebarrow:
Just need an update to the docs now.

clydebarrow:
Oh, and add the config to one or more the test files under `tests/components/modbus`

0x3333:
I’ll do the docs and test. No worries.

0x3333:
> Just need an update to the docs now.

The docs base is next, right?

clydebarrow:
> The docs base is next, right?

Correct. When creating the PR make sure you change the base branch in the drop-down, can't be done afterwards. DAMHIK.

0x3333:
> Oh, and add the config to one or more the test files under `tests/components/modbus`

Done

> Correct. When creating the PR make sure you change the base branch in the drop-down, can't be done afterwards. 

Done, PR esphome/esphome-docs#4202

jesserockz (approved):

PR #7316, opened 2024-08-19
feat: Expand ByteBuffer

Rapsssito:

Extracted core changes from #7009 (and a few others) as commented in https://github.com/esphome/esphome/pull/7009#discussion_r1717801326.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esphome:
  name: bytebuffer
  on_boot:
    - lambda: |-
            auto buf = ByteBuffer::create(16);
            assert(buf.get_endianness() == Endian::LITTLE);
            assert(buf.get_remaining() == 16);
            buf.set_limit(10);
            assert(buf.get_capacity() == 16);
            buf.put_uint8(1);
            assert(buf.get_remaining() == 9);
            buf.put_uint16(0x1234);
            assert(buf.get_remaining() == 7);
            buf.put_uint32(0x12345678UL);
            assert(buf.get_remaining() == 3);
            assert(buf.array()[1] == 0x34);
            assert(buf.array()[2] == 0x12);
            assert(buf.array()[3] == 0x78);
            assert(buf.array()[4] == 0x56);
            assert(buf.array()[5] == 0x34);
            assert(buf.array()[6] == 0x12);
            buf.flip();
            assert(buf.get_capacity() == 16);
            assert(buf.get_remaining() == 7);
            assert(buf.get_uint8() == 1);
            assert(buf.get_uint16() == 0x1234);
            assert(buf.get_uint32() == 0x12345678ul);
            assert(buf.get_remaining() == 0);
            buf.rewind();
            buf.big_endian();
            assert(buf.get_remaining() == 7);
            assert(buf.get_uint8() == 1);
            assert(buf.get_uint16() == 0x3412);
            buf.mark();
            assert(buf.get_uint32() == 0x78563412ul);
            assert(buf.get_remaining() == 0);
            buf.reset();
            assert(buf.get_remaining() == 4);
            assert(buf.get_uint32() == 0x78563412ul);
            auto buf1 = ByteBuffer::wrap(buf.array(), buf.get_limit());
            buf.clear();
            assert(buf.get_position() == 0);
            assert(buf.get_capacity() == 16);
            assert(buf.get_limit() == 16);
            assert(buf1.get_remaining() == 7);
            assert(buf1.get_capacity() == 7);
            buf1.set_position(3);
            assert(buf1.get_uint32() == 0x12345678ul);
            buf1.clear();
            assert(buf1.get_limit() == 7);
            assert(buf1.get_capacity() == 7);
            assert(buf1.get_position() == 0);
            float f = 1.2345;
            buf1.put_float(f);
            buf1.flip();
            assert(buf1.get_remaining() == 4);
            assert(buf1.get_float() == f);
            buf1.clear();
            buf1.put_uint16(-32760);
            buf1.put_uint24(-302760);
            buf1.flip();
            assert(buf1.get_int16() == -32760);
            assert(buf1.get_int24() == -302760);
            auto buf2 = ByteBuffer::wrap({0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08}, Endian::BIG);
            assert(buf2.get_remaining() == 8);
            assert(buf2.get_capacity() == 8);
            assert(buf2.get_uint64() == 0x0102030405060708);
            auto buf3 = ByteBuffer::wrap(true);
            assert(buf3.get_remaining() == 1);
            assert(buf3.get_bool() == true);
```

## Checklist:


esphome/core/bytebuffer.cpp
```diff
@@ -1,20 +1,69 @@
...
+ByteBuffer ByteBuffer::create(size_t capacity, Endian endianness) {
   std::vector<uint8_t> data(capacity);
-  return {data};
+  ByteBuffer buffer = {data};
+  buffer.endianness_ = endianness;
+  return buffer;
 }
 
-ByteBuffer ByteBuffer::wrap(uint8_t *ptr, size_t len) {
+ByteBuffer ByteBuffer::wrap(const uint8_t *ptr, size_t len) {
```
clydebarrow: This should have an endianness parameter like the other `wrap` methods. I appreciate that it's not strictly essential since it has no impact on the function itself, but it's inconsistent with the others and might still need the endianness set after the wrap call.

esphome/core/bytebuffer.cpp
```diff
@@ -1,20 +1,69 @@
...
+  return buffer;
+}
+
+ByteBuffer ByteBuffer::wrap(double value, Endian endianness) {
+  ByteBuffer buffer = ByteBuffer::create(sizeof(double), endianness);
+  buffer.put_double(value);
+  return buffer;
+}
+
+ByteBuffer ByteBuffer::wrap(const std::string &data) {
```
clydebarrow: This does not encode a string in any commonly recognised format - if a user wants to do this it's easy enough but including it here would be misleading.

esphome/core/bytebuffer.cpp
```diff
@@ -126,9 +243,60 @@ void ByteBuffer::put_uint32(uint32_t value) {
...
+    for (uint8_t byte_part : byte_array) {
+      this->data_[this->position_++] = byte_part;
+    }
+  } else {
+    for (size_t i = sizeof(double); i > 0; i--) {
+      this->data_[this->position_++] = byte_array[i - 1];
+    }
+  }
+}
+void ByteBuffer::put_string(const std::string &value) {
```
clydebarrow: I'd suggest leaving this out - it's not a complete solution to storing strings, which need the length encoded as well. Best left up to the individual requirement.

esphome/core/bytebuffer.cpp
```diff
@@ -56,35 +135,73 @@ uint32_t ByteBuffer::get_uint32() {
...
+    }
+  }
+  double value;
+  std::memcpy(&value, byte_array, sizeof(double));
+  return value;
 }
-float ByteBuffer::get_float() {
-  auto value = this->get_uint32();
-  return *(float *) &value;
+std::string ByteBuffer::get_string(size_t length) {
```
clydebarrow: As per the other string related functions best left out.

esphome/core/bytebuffer.cpp
```diff
@@ -1,20 +1,69 @@
...
+  buffer.put_double(value);
+  return buffer;
+}
+
+ByteBuffer ByteBuffer::wrap(const std::string &data) {
+  std::vector<uint8_t> buffer(data.begin(), data.end());
+  return {buffer};
+}
+
+ByteBuffer ByteBuffer::wrap(std::initializer_list<uint8_t> values) {
```
clydebarrow: Needs the endianness parameter.

esphome/core/event_emitter.cpp
```diff
@@ -0,0 +1,12 @@
+#include "event_emitter.h"
```
clydebarrow: Break this out to a separate PR - it's unrelated to the ByteBuffer stuff and needs Jesse's review I think.

clydebarrow (changes requested):

esphome/core/bytebuffer.cpp
Rapsssito: Included in db28347a1f267d07e263900665b217fe5eb77cb8

esphome/core/bytebuffer.cpp
```diff
@@ -1,19 +1,69 @@
...
+  return buffer;
+}
+
+ByteBuffer ByteBuffer::wrap(double value, Endian endianness) {
+  ByteBuffer buffer = ByteBuffer::create(sizeof(double), endianness);
+  buffer.put_double(value);
+  return buffer;
+}
+
+ByteBuffer ByteBuffer::wrap(std::initializer_list<uint8_t> values, Endian endianness) {
```
clydebarrow: The endianness parameter isn't used here. Did you put together a test script similar to what I had in the original PR? I'm guessing not :-) This is what I did - I'd recommend expanding it to include all the new methods and running it (the host platform is Really Useful for that.) Given that many of these methods won't be used right away, it is important to make sure they actually work as designed now. Sadly the ESPHome test files only test compilation, not execution.

I'd particularly make sure you test the big-endian cases, since those are non-default, and would have picked up the omission above.

```
esphome:
  name: bytebuffer
  on_boot:
    - lambda: |-
        auto buf = ByteBuffer::create(16);
        assert(buf.get_endianness() == Little);
        assert(buf.get_remaining() == 16);
        buf.set_limit(10);
        assert(buf.get_capacity() == 16);
        buf.put_uint8(1);
        assert(buf.get_remaining() == 9);
        buf.put_uint16(0x1234);
        assert(buf.get_remaining() == 7);
        buf.put_uint32(0x12345678UL);
        assert(buf.get_remaining() == 3);
        assert(buf.array()[1] == 0x34);
        assert(buf.array()[2] == 0x12);
        assert(buf.array()[3] == 0x78);
        assert(buf.array()[4] == 0x56);
        assert(buf.array()[5] == 0x34);
        assert(buf.array()[6] == 0x12);
        buf.flip();
        assert(buf.get_capacity() == 16);
        assert(buf.get_remaining() == 7);
        assert(buf.get_uint8() == 1);
        assert(buf.get_uint16() == 0x1234);
        assert(buf.get_uint32() == 0x12345678ul);
        assert(buf.get_remaining() == 0);
        buf.rewind();
        buf.big_endian();
        assert(buf.get_remaining() == 7);
        assert(buf.get_uint8() == 1);
        assert(buf.get_uint16() == 0x3412);
        buf.mark();
        assert(buf.get_uint32() == 0x78563412ul);
        assert(buf.get_remaining() == 0);
        buf.reset();
        assert(buf.get_remaining() == 4);
        assert(buf.get_uint32() == 0x78563412ul);
        auto buf1 = ByteBuffer::wrap(buf.array(), buf.get_limit());
        buf.clear();
        assert(buf.get_position() == 0);
        assert(buf.get_capacity() == 16);
        assert(buf.get_limit() == 16);
        assert(buf1.get_remaining() == 7);
        assert(buf1.get_capacity() == 7);
        buf1.set_position(3);
        assert(buf1.get_uint32() == 0x12345678ul);
        buf1.clear();
        assert(buf1.get_limit() == 7);
        assert(buf1.get_capacity() == 7);
        assert(buf1.get_position() == 0);
        float f = 1.2345;
        buf1.put_float(f);
        buf1.flip();
        assert(buf1.get_remaining() == 4);
        assert(buf1.get_float() == f);
        buf1.clear();
        buf1.put_uint16(-32760);
        buf1.put_uint24(-302760);
        buf1.flip();
        assert(buf1.get_int16() == -32760);
        assert(buf1.get_int24() == -302760);

```

clydebarrow (changes requested):

esphome/core/bytebuffer.cpp
Rapsssito: You are so right... I am sorry. I updated the PR description with the same code, but with a little addition to check for `get_uint64`, `get_bool` and a `wrap` with BE. I do not know how many more checks I should add. Let me know!

esphome/core/bytebuffer.cpp
```diff
@@ -54,11 +60,6 @@ ByteBuffer ByteBuffer::wrap(double value, Endian endianness) {
   return buffer;
 }
 
-ByteBuffer ByteBuffer::wrap(const std::string &data) {
-  std::vector<uint8_t> buffer(data.begin(), data.end());
-  return {buffer};
-}
```
nielsnl68: Why did you remove this?

esphome/core/bytebuffer.cpp
```diff
@@ -191,15 +192,6 @@ double ByteBuffer::get_double() {
...
 }
-std::string ByteBuffer::get_string(size_t length) {
-  assert(this->get_remaining() >= length);
-  std::string value;
-  value.reserve(length);
-  for (size_t i = 0; i < length; i++) {
-    value.push_back(this->data_[this->position_++]);
-  }
-  return value;
-}
```
nielsnl68: ~~Why did no remove this?~~ I see now that @clydebarrow asked to do so. But i would love to have this.

esphome/core/bytebuffer.cpp
```diff
@@ -293,10 +285,4 @@ void ByteBuffer::put_double(double value) {
     }
   }
 }
-void ByteBuffer::put_string(const std::string &value) {
-  assert(this->get_remaining() >= value.size());
-  for (char c : value) {
-    this->data_[this->position_++] = c;
-  }
-}
```
nielsnl68: This are important methods to handle with the buffer class. Although i think this would be better using c_str()

nielsnl68 (changes requested):
~~I see no reason why this PR should be merged.~~ This is no longer the case.

esphome/core/bytebuffer.cpp
nielsnl68: I totally disagree on that, This is just for me a must have method to easily store strings to a buffer without to call .c_str() everytime you want to store a string inside the buffer.

esphome/core/bytebuffer.cpp
nielsnl68: and same as above: I totally disagree on that, This is just for me a must have method to easily store strings to a buffer without to call .c_str() everytime you want to store a string inside the buffer.

esphome/core/bytebuffer.cpp
Rapsssito: I can add it back. In NodeJS is common to use a Buffer (somewhat similar to this class) to encode whole strings, which would remove the issue of knowing the string length in advance. We could remove `get_string(length)` (removing also the `put_string()`) and add a `to_string()` method where the whole Buffer is returned. @clydebarrow, what do you think?

clydebarrow: Remember that every convenience function here gets compiled in and takes up space in the object file whether used or not. I'm already a little concerned with the number of these for that reason so it's best to keep them to a minimal set that provides the building blocks for more complicated use cases.

But, I appreciate there is a need to be able to get a range of bytes so a `get_vector(size)` function would be useful, which can then easily be used to construct a string if that's what's needed. And it would be more efficient. Something like this:

```
 std::vector<uint8_t> get_vector(size_t size) {
   auto start = this->data_.begin() + position_;
   this->position_ += size;
   return {start, start + size};
  }
```

If you want a string you then would just write in user code:

```
auto vec = buffer.get_vector(size);
auto str = std::string(vec.begin(), vec.end());
```

Similarly a `put_vector()` function would be useful.

clydebarrow: And WRT the idea a buffer can just wrap a string, you don't even need this class for that. The whole point of the class is to be able to pack _multiple_ objects into a buffer, you can already easily convert a string to a vector. So I'm wondering whether the various `wrap` methods that just wrap a single scalar value are of sufficient value to be worth taking up valuable space. Do you have an example of how they would be used?

Rapsssito: I have added the `get_vector()` and `put_vector()` methods.

WRT the convenience `wrap` methods, they are heavily used in the new BLE custom server setup in #7009. They remove the need for extra functions (https://github.com/esphome/esphome/pull/7009#discussion_r1716723073) that the user would need for passing templatable values to the C++. In addition, they remove the need for extra convenience functions that were already present `BLECharacteristic::set_value` and `BLEDescriptor::set_value`, so they now can be removed. Converting simple types to byte arrays in a centralized manner (i.e. this class) is a really useful feature. In my opinion, `wrap` methods remove more duplicate code than they add (even more in the long run).

nielsnl68: I see your point @clydebarrow. Although in my use case i need to use this conversion many time and need to create an extra function or extend the bufferByte to add this as methods myself.

In my opinion we have two solutions:
 1) we add a `#define` check around the string methods and when the programmer needs to string function it needs to set the `#define` to enable the extra methods.
 2) we create an extended class in a separate file with the string handling options. so the programmer can just choose to use that class instead.

To me option 2 would be my favorite. I havent seen option 1 being used so far.

Just on question: how could we change the above so we can store dynamic sized string (vector)? I C(++) the strings ands up with a zero and in pascal the size was stored in the string itself at the front of the string.

Rapsssito: @nielsnl68, you can actually create a `ByteBuffer` from a `std::string` with a one-liner:
```cpp
ByteBuffer::wrap(std::vector<uint8_t>(str.begin(), str.end()));
```

This would be the same with the new `put_vector()`.

clydebarrow: std::string does not have a terminating null, the length is stored in the object. The terminating null is C convention, and has plenty of drawbacks. This is my point about string conversions - there is no single standard for how they should be encoded into a byte stream.

When I've designed encoding formats and need to encode a string I typically store the length (as a fixed size integer) followed by the data, so if that's the case you would use a put_uint8 followed by a put_vector - or maybe put_uint16 then vector if the length was allowed to exceed 255.

If your convention was to use a null terminator you'd do put_vector followed by put_uint8(0). But those conventions shouldn't be hardwired into a general purpose feature like this.

That's why put_vector is useful to include, because it's general-purpose, put_string isn't because you would have to choose how to encode it which might not suit the application.

If you had an application where you frequently wanted to encode strings in a particular format, you'd just create your own convenience function for that purpose.

esphome/core/bytebuffer.cpp
```diff
@@ -1,19 +1,70 @@
...
   std::vector<uint8_t> data(ptr, ptr + len);
-  return {data};
+  ByteBuffer buffer = {data};
+  buffer.endianness_ = endianness;
+  return buffer;
 }
 
-ByteBuffer ByteBuffer::wrap(std::vector<uint8_t> data) { return {std::move(data)}; }
+ByteBuffer ByteBuffer::wrap(std::vector<uint8_t> data, Endian endianness) {
+  ByteBuffer buffer = {std::move(data)};
```
clydebarrow:
```suggestion
ByteBuffer ByteBuffer::wrap(std::vector<uint8_t> const &data, Endian endianness) {
  ByteBuffer buffer = {data};
```

Constructing a new vector always does a copy. But I've realised there are unnecessary copies in the existing constructor and create method that should be streamlined (I was not thinking about move/copy semantics at the time.) It will be simpler if I just push some changes to your branch - are you ok with that?

clydebarrow (changes requested):

esphome/core/bytebuffer.cpp
clydebarrow: > Converting simple types to byte arrays in a centralized manner (i.e. this class) is a really useful feature. In my opinion, `wrap` methods remove more duplicate code than they add (even more in the long run).

Fair enough. I don't see the get_vector and put_vector methods though.

nielsnl68: I fully agree on and aware of this. What every we decide it is none standard. Both have there pro's and con's. 
Anyway there are multiple roads that lead to Rome. I was hoping we could defined a global setup so we do not end up having multiple instances doing almost the same thing.

esphome/core/bytebuffer.cpp
Rapsssito: @clydebarrow, of course! I think I gave write permissions when opening the PR (if not, please, let me know).

clydebarrow: Yes, I have the power :-) But always like to ask first. I'll get onto that tomorrow, it's late here. I think we're nearly there with this though.

nielsnl68:
I have still two open questions:

1) for @clydebarrow, just wondering in your original PR you added `uint8_t *array() {return this->data_.data();}`. Is there a reason behind not using ``uint8_t *data()` here?

2) As if i correctly understand the new **`warp()`** methods are basically doing the same as the "from()" for NodeJs is it not? When so would we not want to set `Endian endianness = LITTLE` at the "`static ByteBuffer wrap(uint8_t value);`" as well, for future use?

Anyway i like this class as it is already it is very flexible in the uses.

clydebarrow:
> 2\. When so would we not want to set `Endian endianness = LITTLE` at the "`static ByteBuffer wrap(uint8_t value);`"

It would be only for cosmetic reasons if done - i.e. just to maintain consistency with the other methods. A 1 byte buffer can't make use of endianness since the only thing you can ever get or put is a single byte. There is no provision for expanding the capacity of a buffer, that's a design choice.

clydebarrow:
> * Is there a reason behind not using ``uint8_t *data()` here?
> *

Just that `array()` is what is used by the Java class on which the design was based.

So I changed it to `get_data()` returning the vector - more useful I think.

esphome/core/bytebuffer.cpp
clydebarrow: If the multiple instances are doing _almost_ the same thing that's why they need to be multiple instances.

clydebarrow:
I think this is done - @nielsnl68  and @Rapsssito please look over and see if there's anything more. If not I'll merge it.

nielsnl68:
> > 2. When so would we not want to set `Endian endianness = LITTLE` at the "`static ByteBuffer wrap(uint8_t value);`"
> 
> It would be only for cosmetic reasons if done - i.e. just to maintain consistency with the other methods. A 1 byte buffer can't make use of endianness since the only thing you can ever get or put is a single byte. There is no provision for expanding the capacity of a buffer, that's a design choice.

I see, i was mislead by the **`set_limit()`** method name, and i expected (without checking the code behind it) that you could resize the vector buffer. But it is instead possible to virtual reduce the limit size. When it is just to reduce the limit size maybe we should rename it to **`reduce_limit()`**.

That said i do not wrap small uint's into this ByteBuffer? Could you guys give me some use cases for them? which cant be done using normal byte array's?

clydebarrow:
> When it is just to reduce the limit size maybe we should rename it to **`reduce_limit()`**.

You're confusing the limit (which is the end of the data that has so far been appended to the buffer) with the capacity - which is the maximum possible size. The limit will change depending on what has been written to or read from the buffer, it just can't ever exceed the capacity.

As far as use cases for the wrap functions, refer to @Rapsssito 's comments earlier - that's different to the things I want to do with it but if it can serve multiple purposes that's fine.

nielsnl68:
> I think this is done - @nielsnl68 and @Rapsssito please look over and see if there's anything more. If not I'll merge it.

Currently there is a lot of duplicated code to get and put uint's into the buffer. i suggest to centralize them to on method like:

```cpp
uint64_t ByteBuffer::get_uint(uint8_t bytes, Endian endianness /* = LITTLE */) {
  assert(this->get_remaining() >= bytes);
  uint64_t value = 0;
  if (endianness == LITTLE) {
    value |= this->data_[this->position_++];
    if (bytes>1) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 8;
    if (bytes>2) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 16;
    if (bytes>3) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 24;
    if (bytes>4) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 32;
    if (bytes>5) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 40;
    if (bytes>6) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 48;
    if (bytes>7) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 56;
  } else {
    if (bytes>7) value != static_cast<uint64_t>(this->data_[this->position_++]) << 56;
    if (bytes>6) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 48;
    if (bytes>5) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 40;
    if (bytes>4) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 32;
    if (bytes>3) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 24;
    if (bytes>2) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 16;
    if (bytes>1) value |= static_cast<uint64_t>(this->data_[this->position_++]) << 8;
    value |= this->data_[this->position_++];
  }
  return value;
}

uint64_t ByteBuffer::get_uint64() {
  return this->get_uint(8, endianness_);
}

uint32_t ByteBuffer::get_uint32() {
  return static_cast<uint32_t>(this->get_uint(4, endianness_));
}

uint16_t ByteBuffer::get_uint16() {
  return static_cast<uint16_t>(this->get_uint(2, endianness_));
}
```
etc.

clydebarrow:
> Currently there is a lot of duplicated code to get and put uint's into the buffer. i suggest to centralize them to on method like:
> etc.

How much difference to code size and speed will that make? The size will be smaller, but slower, but by how much?

nielsnl68:
> How much difference to code size and speed will that make? The size will be smaller, but slower, but by how much?

I have always learned to reduce duplicated code as much as possible and moving it into a sperate function/method and frankly speed issues where never an issue.
So I do not know about the speed difference, to me it is also about readable and maintenance.

nielsnl68:
Anyway i would love to see the `put_uint()` and `get_uint()` being added, i would love to use that to store MAC addresses inside the buffer that now are stored inside a uint64 and i only need the first 6 bytes in BIG endianness being stored and retrieved.

esphome/core/bytebuffer.h
```diff
@@ -15,32 +15,59 @@ enum Endian { LITTLE, BIG };
...
+  }
+  static ByteBuffer wrap(int64_t value, Endian endianness = LITTLE) {
+    return wrap(static_cast<uint64_t>(value), endianness);
+  }
+  static ByteBuffer wrap(float value, Endian endianness = LITTLE);
+  static ByteBuffer wrap(double value, Endian endianness = LITTLE);
+  static ByteBuffer wrap(bool value) { return wrap(static_cast<uint8_t>(value)); }
+  static ByteBuffer wrap(std::initializer_list<uint8_t> values, Endian endianness = LITTLE) {
+    return wrap(std::vector<uint8_t>(values), endianness);
+  }
```
Rapsssito: I actually added this `wrap(std::initializer_list)` because I thought it was necessary if provided with a braced-init-list `{1, 2, 3, 4, 5}` (which are the ones generated by the Python code). However, I checked and saw that the compiler went automatically to the `wrap(std::vector)` in those cases, so `wrap(std::initializer_list)` is not necessary in my use case (I removed it in a81d45945d79f6b2f762ef2f3f06864170695462). Feel free to remove it if you think it won't be needed.

esphome/core/bytebuffer.cpp
```diff
@@ -93,42 +181,74 @@ void ByteBuffer::put_uint8(uint8_t value) {
...
+    this->data_[this->position_++] = static_cast<uint8_t>(value >> 16);
+    this->data_[this->position_++] = static_cast<uint8_t>(value >> 8);
+    this->data_[this->position_++] = static_cast<uint8_t>(value);
+  }
+}
+void ByteBuffer::put_float(float value) {
+  static_assert(sizeof(float) == sizeof(uint32_t), "Float sizes other than 32 bit not supported");
+  assert(this->get_remaining() >= sizeof(float));
+  this->put_uint32(*reinterpret_cast<uint32_t *>(&value));
+}
```
Rapsssito: I get a warning when compiling `get_float`, `put_float`, `get_double`, `put_double`:

> warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]

I know it is just a warning, but I previously changed the code so it also allowed for compatibility with different sizes of floats and doubles (and removed the warning). It was more "verbose", but might be worth the compatibility bonus.

```cpp
float ByteBuffer::get_float() {
  assert(this->get_remaining() >= sizeof(float));
  uint8_t byte_array[sizeof(float)];
  if (this->endianness_ == LITTLE) {
    for (uint8_t &byte_part : byte_array) {
      byte_part = this->data_[this->position_++];
    }
  } else {
    for (size_t i = sizeof(float); i > 0; i--) {
      byte_array[i - 1] = this->data_[this->position_++];
    }
  }
  float value;
  std::memcpy(&value, byte_array, sizeof(float));
  return value;
}
```

If `get_uint()` is finally added. It would be even smaller:
```cpp
float ByteBuffer::get_float() {
  assert(this->get_remaining() >= sizeof(float));
  std::memcpy(&value, &this->get_uint(sizeof(float)), sizeof(float));
  return value;
}
```

Rapsssito:
> So I do not know about the speed difference, to me it is also about readable and maintenance.

A "for" loop would probably be faster here, as CPU branch predictors are usually highly optimized for them. Either way, the performance difference is probably a few CPU cycles in all cases.

nielsnl68:
> > So I do not know about the speed difference, to me it is also about readable and maintenance.
> 
> A "for" loop would probably be faster here, as CPU branch predictors are usually highly optimized for them. Either way, the performance difference is probably a few CPU cycles in all cases.

You are totally right.

esphome/core/bytebuffer.h
clydebarrow: Ah, good catch.

esphome/core/bytebuffer.cpp
clydebarrow: Yes, I can see the appeal of `get_uint()` - will go that route. I'm not seeing the warning - which compiler is that with?

clydebarrow: Ah, esp-idf 5.x

nielsnl68:
@clydebarrow the std::vecter.resize() is not available in the current compiler that we use, is it?

jesserockz:
Yes, we use resize in various places I believe.

Can the float warning fix be extracted to it's own pr so I can release in a patch?

clydebarrow:
> Can the float warning fix be extracted to it's own pr so I can release in a patch?

Not really, no. As it is it depends on other changes. It would be possible to produce a separate patch fix. Or this can go out as is, I'm happy with it.

clydebarrow (approved):

clydebarrow:
> std::vecter.resize()

A resize method could be added - just needs some thought on what effect that should have on position etc. since there is no existing model to follow.

nielsnl68 (approved):

esphome/core/bytebuffer.cpp
```diff
@@ -27,108 +72,102 @@ void ByteBuffer::clear() {
...
-    this->data_[this->position_++] = (uint8_t) (value >> 8);
-    this->data_[this->position_++] = (uint8_t) value;
-  }
-}
-void ByteBuffer::put_float(float value) { this->put_uint32(*(uint32_t *) &value); }
-void ByteBuffer::flip() {
-  this->limit_ = this->position_;
-  this->position_ = 0;
+void ByteBuffer::put_float(float value) {
+  static_assert(sizeof(float) == sizeof(uint32_t), "Float sizes other than 32 bit not supported");
```
Rapsssito, line 155: It should work with `<= sizeof(uint32_t)` too, right? Same with double.

clydebarrow: It could, but you'd need to test for that case and use an appropriate size integer to copy into and out of. And there's no current need since no platform we're targeting uses other sizes. If that changes in future, the static_assert will ensure that it's noticed and extended.

clydebarrow:
Drawing a line and merging now - any further tweaks can be done in a future PR.

Rapsssito:
LGTM!

PR #7319, opened 2024-08-19
[core] Initialize submodules in external component from git

szupi-ipuzs:

If external component is added from a repo that containes submodules, the submodules are not downloaded. If they contain the wanted components, an error is thrown.

This PR fixes the above issue. The fix works, but it's probably not the the best one, as it forces downloading of all the submodules even the ones you might not need.
Maybe the submodules should be specified directly in yaml, like the components are?
Or maybe we should assume the submodules are named exactly the same as the components (so a separate list is not needed)?

I await your ideas :)

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
external_components:
  - source:
      type: git
      url: https://github.com/szupi-ipuzs/esphome-external-components
    components: [ deep_sleep_libretiny ]
```

## Checklist:


jesserockz (changes requested):
I don't think we should blindly init all submodules. 
If we should add submodules to external_components, then there should be a config variable for it. Possible values could be "all", "components", or a list of strings which would add the correct argument to `clone_or_update`

PR #7321, opened 2024-08-20
Warn about data loss on API communication

syssi:

Every now and then (if a lot of sensors gets published very fast) the `tx_buf_` isn't empty and `can_write_without_blocking()` returns `false` two times in a row. In this case a message gets silently dropped. If it's a state update of a `binary_sensor` the sensor entity at Home Assistant doesn't get updated and will be out of sync. 

It's hard to get an idea what's going on because the issue disappear/is unlikely if you increase the log level.

This PR doesn't resolve the issue. It just helps to make the issue more visible if you experience loss of state updates too.

## Types of changes

- [ ] Other

## How to reproduce the issue

[test.yaml.txt](https://github.com/user-attachments/files/16670706/test.yaml.txt)

Some states are out of sync because of dropped messages:
states

## Checklist:

### DIFF
--- a/esphome/components/api/api_connection.cpp
+++ b/esphome/components/api/api_connection.cpp
@@ -1886,7 +1886,7 @@ bool APIConnection::send_buffer(ProtoWriteBuffer buffer, uint32_t message_type) 
     if (!this->helper_->can_write_without_blocking()) {
       // SubscribeLogsResponse
       if (message_type != 29) {
-        ESP_LOGV(TAG, "Cannot send message because of TCP buffer space");
+        ESP_LOGW(TAG, "Unable to send message: TX buffer not empty. Message discarded");
       }
       delay(0);
       return false;


guillempages (approved):

tricky10:
I would like to confirm the issue. The more sensors I put into the configuration some of them are not being refreshed for hours. This is a random thing. In case of decreasing the amount of sensors the stability comes back but this is very annoying. 
Are there any workarounds ? I've attached my configuration which contains all sensors i support  
[jkbms-easun.txt](https://github.com/user-attachments/files/17281942/jkbms-easun.txt)

DjordjeMandic:
How likely is this to cause excessive log spam?

syssi:
It's hard to guess because it depends on the complexity / number of entities / number of blocking operations. Users with lightweight setups aren't affected by the "data loss" and won't see this warning. If you are affected you will see the message once per dropped message.

DjordjeMandic (approved):

shvmm:
@nkinnan Your say on this 'excessive log spam' concern ?

clydebarrow:
https://github.com/esphome/esphome/pull/7547

syssi:
Awesome job @nkinnan! Feel free to close this MR if it's obsolete.

nkinnan:
> @nkinnan Your say on this 'excessive log spam' concern ?

I agree with the original intent to promote this to warning level since it did result in a form of data loss, however https://github.com/esphome/esphome/pull/7547 has made the log message obsolete.  It should no longer be possible to hit this condition.  

That said there is no harm in retaining this PR, and potentially still some benefit - seeing it would indicate a regression in behavior or a newly introduced bug.  Or even an oversight in the implementation of a new sensor type that doesn't take advantage of the new queue functionality.  So I'd suggest keeping this open personally.

clydebarrow:
Here's another thought - instead of logging when this happens, which may make the problem worse, keep a count of the number of times it happens (only when the API connection is in place) and publish it as a sensor. Do the same for the number of API disconnects. Publish the sensors on an interval. That way there is no overhead unless those sensors are configured, and looking at the numbers vs uptime will give a sense of how bad the problem, if any, is. Beats looking tthrough the log IMO.

nkinnan:
> Here's another thought - instead of logging when this happens, which may make the problem worse, keep a count of the number of times it happens (only when the API connection is in place) and publish it as a sensor. Do the same for the number of API disconnects. Publish the sensors on an interval. That way there is no overhead unless those sensors are configured, and looking at the numbers vs uptime will give a sense of how bad the problem, if any, is. Beats looking tthrough the log IMO.

Perhaps as a new sensor under the debug component? https://esphome.io/components/debug.html Perhaps even as part of a general effort to get more debug info like this available in general... Having stats on the queue implemented in https://github.com/esphome/esphome/pull/7547 itself would be neat. But not trying to push scope creep on syssi, I just like the idea. ;)

clydebarrow:
> Perhaps as a new sensor under the debug component

Just under the `api` component I would have thought. Otherwise you introduce a dependency from `debug` to `api`. It would look something like this in config:

```
sensor:
  - platform: api
    type: dropped_messages
    id: dropped_msgs
    name: "Dropped API message count"
  - platform: api
    type: dropped_connections
    id: dropped_connections
    name: "Dropped API connection count"
```

nkinnan:
You're right, I hadn't considered the dependency.  I'd only suggest adding delayed message count to track how many times the api queue was utilized to prevent drop.

shvmm:
umm, something is definitely wrong.
Even with latest ESPHome release 25.3.0
I keep hitting the condition that shows this log message

> > @nkinnan Your say on this 'excessive log spam' concern ?
> 
> I agree with the original intent to promote this to warning level since it did result in a form of data loss, however #7547 has made the log message obsolete. It should no longer be possible to hit this condition.
> 
> That said there is no harm in retaining this PR, and potentially still some benefit - seeing it would indicate a regression in behavior or a newly introduced bug. Or even an oversight in the implementation of a new sensor type that doesn't take advantage of the new queue functionality. So I'd suggest keeping this open personally.

```[09:17:57.937][I][main:752]: Received CAN ID: 0x305 ACK
[09:17:57.938][I][jk_bms:071]: Status frame received
[09:17:57.938][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.938][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.938][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.938][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.938][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.939][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.939][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.939][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.939][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.939][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.939][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.939][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.940][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.940][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:57.940][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:58.452][W][api.connection:1889]: Unable to send message: TX buffer not empty. Message discarded
[09:17:58.453][W][component:237]: Component jk_modbus took a long time for an operation (150 ms).
[09:17:58.453][W][component:238]: Components should block for at most 30 ms.
[09:18:00.475][I][main:1074]: send CAN ID: 0x359 HEX: 0 0 0 0 1 64 0 0
[09:18:00.476][I][main:1109]: send CAN ID: 0x351 HEX: 38 2 f4 1 20 3 e0 1
[09:18:00.476][I][main:1110]: Charge Status : Bulk
[09:18:00.476][I][main:1151]: send CAN ID: 0x355 HEX: 5 0 5d 0 36 d f6 c
[09:18:00.476][I][main:1175]: send CAN ID: 0x356 HEX: e7 14 b5 0 5 1 4 1
[09:18:00.476][I][main:1215]: send CAN ID: 0x35C HEX: c0 0 f3 0
[09:18:00.476][I][main:1221]: send CAN ID: 0x35E ASCII : SHVM
[09:18:00.477][I][main:752]: Received CAN ID: 0x305 ACK
```

nkinnan:
@shvmm Are you seeing a behavior change? Less dropped readings? It's been so many months since I created this PR...

shvmm:
> @shvmm Are you seeing a behavior change? Less dropped readings? It's been so many months since I created this PR...

Not much. I've removed this for now so as to avoid potential bogging down of connection.
 I'm trying to diagnose a separate issue CANBus status dropping out frequently since update

PR #7326, opened 2024-08-20
[uponor_smatrix] Modifies sending algorithm

skasi7:

Modifies sending algorithm to allow sending packages even if the controller is continuously polling the thermostats.

The current code expects the controller to have idle periods of 10s each time a round of polls is made, but my controller doesn't have those periods, making impossible to modify the target temperatures and/or synchronizing time.

This PR aims to fixing this behavior using the small gaps of ~200ms between thermostat polls, given that the `loop` code is called each 5-15ms and sending data doesn't take more than 50ms, it's safe to change this behavior.

Also, removes a variable tracking the last polling start, not needed anymore.

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


kroimon:
Hi @skasi7, thanks for your PR and for using my component!

Can you provide more info on your setup? Which controller do you have and how many thermostats are connected?
I wrote down my protocol analysis [here ](https://github.com/kroimon/uponor-smatrix-analysis)when I wrote this component. Anything you can provide to improve this analysis is highly appreciated! Maybe your system uses different values for the fields I could not fully decipher or maybe you have additional packet dumps?

Only 200 ms of bus silence is a tight fit, especially if there is more than one packet in the send queue. We might need to think about limiting ourselves to sending a single packet per gap in this case.
Edit: Oh I see, you already limited the sending window to 150 ms after the last received byte.

skasi7:
Hi @kroimon!

First of all, thanks for your incredible work deciphering the protocol and coding the component. This is really helpful to avoid the very expensive R-208.

Now, answering your questions. My setup is exactly the same that you describe in your analysis (X-245 controller + 5x T-148 thermostats). I've been taking [logs](https://github.com/user-attachments/files/16679304/logs.txt)
 modifying the component and the only difference between your dumps and mine is that the controller does not stop after the last thermostat, but continue with the first right after the usual 200ms interval. I don't know why my controller behaves like this, but if I log the `last_poll_start_` only increases and never resets.

My solution aims at using the gaps between thermostat polls, trying to leave enough time before and after the packages and only sending one package in each gap to avoid collisions. After reading some RS-485 documentation, the 50ms window implemented should be more than enough to avoid collisions.

Thanks for the review. Best regards,

kroimon (approved):
I finally had time to test this PR on my setup. Everything seems to be working. 👍 

The only downside is that with the 10 second polling interval on my system, most of the bus silence is now wasted because only a single packet is now sent immediately after each received packet. We could probably build a more sophisticated way of better estimating expected bus silence periods, but... I can live with that - floor heating system's aren't known for their fast reaction times anyway 😄 

Approved.

jesserockz (approved):

PR #7327, opened 2024-08-20
feat(MQTT): Add templatable values for discovery, connection and topics

Rapsssito:

Added templated options for MQTT `broker`, `port`, `username`, `password`, `client_id`, `discovery`, `discover_ip`, `discovery_retain`, `discovery_prefix` and `topic_prefix` (and the topics for `birth_message`, `will_message`, `shutdown_message` and `log_topic`). These templates are only computed once during startup.

It provides flexibility and allows for more complex setups. For example, when the device turns on and off repeatedly in a cyclical manner (so discovery is not always necessary) or when the connection information might be negotiated dynamically.

The `topic_prefix` templatability fixes [this forum post](https://community.home-assistant.io/t/esphome-i-will-be-able-to-use-name-add-mac-suffix-on-mqtt-topic-prefix/429599/16) (I do not know if it has an associated issue).

**NOTE**: HA tries to read the logs from the wrong MQTT topic if `log_topic` is a template. I think the best approach is to avoid trying to automatically connect to the logs in this case. I was not able to disable (nor show a warning) because I am not familiar with that part of the codebase.

### `codegen.py` changes

I have implented an `instant_templatable` method that generates a `LambdaExpression` that is called intermediately on declaration.

```python
async def instant_templatable(
    value: Any,
    output_type: Optional[SafeExpType],
    to_exp: Union[Callable, dict] = None,
):
    """Generate code for a templatable config option that is executed
    immediately on setup.

    If `value` is a templated value, the lambda expression is returned
    with inline exectution. Otherwise the value is returned as-is
    (optionally process with to_exp).

    :param value: The value to process.
    :param output_type: The output type of the lambda expression.
    :param to_exp: An optional callable to use for converting non-templated values.
    :return: The potentially templated value.
    """
    lambda_ = await templatable(value, [], output_type, to_exp)
    # If the lambda is a LambdaExpression, we execute it immediately
    if isinstance(lambda_, LambdaExpression):
        return CallExpression(lambda_, None)
    return lambda_
```

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4176

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
mqtt:
  broker: "192.168.178.84"
  port: !lambda 'return 1883;'
  username: debug
  password: debug
  client_id: someclient
  use_abbreviations: false
  discovery: !lambda 'return true;'
  topic_prefix: !lambda 'return "testing-topic-prefix";'
```

## Checklist:


Rapsssito:
@jesserockz, let me know what you think!

Rapsssito:
@jesserockz, @clydebarrow, what do you think? Sending you this friendly reminder. It is quite a small PR and might be very useful for dynamic MQTT connections (different users, passwords, IPs, etc.).

Rapsssito:
@jesserockz, @clydebarrow, what do you think? Pinging you as a second friendly reminder. I have updated and increased the scope of the PR. I will update the docs once I have your blessing!

jesserockz (changes requested):
A better and more user friendly option would be to make most of the required config options optional, which would tell the mqtt client not to try to connect on setup, and add a new action `mqtt.configure` or similar to allow setting them in an automation

PR #7333, opened 2024-08-21
[deep_sleep] touch_wakeup capability on ESP32

### LINKED ISSUE #3440: ESP32 Touch + Deep Sleep not working, fully implemented?
### The problem

Hi all,
New to the project but navigating with some basic C++ understandings.

I am trying to build a project using touch pins to detect moisture, while sending the device into deep sleep as much as possible so that it can be driven by batteries. What I have noticed is: any touch pin reading change, even though well above touch binary sensor thresholds, will wake up the device with a sensor OFF message and wake up reason 2: ESP_SLEEP_WAKEUP_EXT0, Wakeup caused by external signal using RTC_IO.

I checked the ESP32 implementation and found the "touchAttachInterrupt" needs to be called to specify the threshold, for example here:
https://randomnerdtutorials.com/esp32-touch-wake-up-deep-sleep/

I then searched around the entire ESPHome project but didn't find any call to this function. This makes me wonder if touch wake up from deep sleep has been fully implemented. Kindly suggest if I missed something. Thanks you!

2022.6.1

Docker

2022.3.5

ESP32

### Board

all ESP32 boards

_No response_

### Example YAML snippet

```yaml
esphome:
  name: abcde-04aabb
  on_boot:
    priority: 900
    then:
      lambda: |-
        id(wake_up_reason) = esp_sleep_get_wakeup_cause();

esp32:
  board: esp32dev
  framework:
    type: arduino

# Enable logging
logger:
  level: NONE

wifi:
  ssid: !secret wifi_ssid
  fast_connect: true

deep_sleep:
  run_duration: 10s
  sleep_duration: 3590s
  touch_wakeup: true
  wakeup_pin: T4
  wakeup_pin_mode: KEEP_AWAKE

esp32_touch:
  iir_filter: 10ms
  setup_mode: false

binary_sensor:
  - platform: esp32_touch
    name: esphome-leak
    device_class: moisture
    pin: T4
    threshold: 100

globals:
  - id: wake_up_reason
    type: int
    restore_value: no
    initial_value: '0'
    
sensor:
  - platform: template
    name: "Wake-up Reason"
    accuracy_decimals: 0
    lambda: |-
      return (id(wake_up_reason));

mqtt:
  broker: *****
  username: ******
  password: ******
  birth_message:
   topic: /status
   payload: "Online"
  will_message:
   topic: /status
   payload: "Offline"
```

_No response_

### Additional information

_No response_

jnuk:

When using the `touch_wakeup` option in the `deep_sleep` component on an ESP32, even with the corresponding `binary_sensor` configured with a `wakeup_threshold`, the ESP does not wake up on touch events. (see linked [issue 3440](https://github.com/esphome/issues/issues/3440)).
As @lhac5vet pointed out in the issue's thread, changing the `esp_sleep_pd_config` from `ESP_PD_OPTION_ON` to `ESP_PD_OPTION_AUTO` fixed this issue, according to a note in the ESP32 API-Sleep Modes ([source](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/sleep_modes.html#_CPPv432esp_sleep_enable_touchpad_wakeupv)): 

> On ESP32, touch wakeup source can not be used when RTC_PERIPH power domain is forced to be powered on (ESP_PD_OPTION_ON) or when ext0 wakeup source is used.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes [issue 3440](https://github.com/esphome/issues/issues/3440)

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esp32_touch:

binary_sensor:
  - platform: esp32_touch
    id: touch_test
    pin: GPIO4
    threshold: 400
    wakeup_threshold: 400

deep_sleep:
  touch_wakeup: True

```

## Checklist:

### DIFF
--- a/esphome/components/deep_sleep/deep_sleep_esp32.cpp
+++ b/esphome/components/deep_sleep/deep_sleep_esp32.cpp
@@ -81,7 +81,7 @@ void DeepSleepComponent::deep_sleep_() {
 
   if (this->touch_wakeup_.has_value() && *(this->touch_wakeup_)) {
     esp_sleep_enable_touchpad_wakeup();
-    esp_sleep_pd_config(ESP_PD_DOMAIN_RTC_PERIPH, ESP_PD_OPTION_ON);
+    esp_sleep_pd_config(ESP_PD_DOMAIN_RTC_PERIPH, ESP_PD_OPTION_AUTO);
   }
 #endif
 #if defined(USE_ESP32_VARIANT_ESP32C3) || defined(USE_ESP32_VARIANT_ESP32C6)


geiseri:
What is the impact on power use? My impression was that this keeps power on.

jnuk:
> What is the impact on power use? My impression was that this keeps power on.

From a quick and dirty measurement I'm getting about 1.4 mA in deep sleep with the current way (`esp_sleep_pd_config` set to `ESP_PD_OPTION_ON`), while setting it to `ESP_PD_OPTION_AUTO` as this PR proposes, I'm getting about 1.6-1.8 mA. However, I'm just using a mulitmeter, so not sure how reliable these results are.

lhac5vet:
Hi, my own experience in measuring very low current in my designs pushed me to buy a dedicated tool. It is called "CurrentRanger". This tool brought me a much better accurate value for this level of currents.
In my current design (based on a DFRobot "Firebeetle 2" board) I can measure a current of 300uA in deep-sleep which is still the double of 150uA worst case of the ESP32 deep-sleep mode as specified in the spec. But there are still chips around the ESp32 like voltage regulator, etc... who are not powered off during deep-sleep.  
For me, there is no measurable difference between the touch wake-up and other wake-up like timer because all make use of the RTC controller to survey and to generate wake-up events. And all use the RTC clock to synchronize the logic block related to each.

benwmcg:
Is there any movement on this being merged? I'd love to be able to wake from deep sleep on touch

nielsnl68 (approved):

nielsnl68:
Not sure if this PR is really needed, i tried deepsleep/esp32_touch without this PR and it works fine for me.

lhac5vet:
I didn't test since I mentioned the problem because I used the component in local. I have checked the new deepsleep component source code and it has been completely rewritten. May be the reason why it works now. Will give it a try.

PR #7335, opened 2024-08-21
feat(WiFi): Add wifi.configure action

Rapsssito:

Added `wifi.configure` action that connects to ab SSID and password (optionally saving it in persistent memory after checking for a successful connection). This allows for more complex setups where the SSID and password might be negotiated dynamically.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4186

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
script:
  - id: change_wifi_sta
    then:
      - wifi.configure:
          ssid: MySSID
          password: password1
          save: true
          on_connect: 
            - logger.log: "Connected to WiFi!"
          on_error:
            - logger.log: "Failed to connect to WiFi!"
```

## Checklist:


Rapsssito:
@jesserockz, what do you think? It is quite a small PR and might be very useful for dynamic WiFi connections.

jesserockz:
Not against it, but what would be better would be that the action actually sets the details on the wifi component, makes sure it connects successfully, only then would it save to prefs, otherwise logs an error.

improv does this too:

```c++
  wifi::global_wifi_component->set_sta(sta);
  wifi::global_wifi_component->start_connecting(sta, false);

  ... 
  // Sets state and keeps checking in `loop` if connected before saving.
  ...

  if (wifi::global_wifi_component->is_connected()) {
    wifi::global_wifi_component->save_wifi_sta(this->connecting_sta_.get_ssid(),
                                             this->connecting_sta_.get_password());
  }
```

Rapsssito:
@jesserockz, you are right! I have updated the method so it now checks for a connection before saving it to the device. I had to use `wifi.disable()` and `wifi.enable()` to make it work (the WiFi component does not like to be disconnected from the current network). Let me know what you think!

I will update the documentation once it is approved, to avoid having to change it again.

Rapsssito:
@jesserockz, just pinging you again as a friendly reminder!

pzich:
@jesserockz is this good to merge? Or are there changes that need to be made?

jesserockz:
Will review again after release this week

esphome/components/wifi/__init__.py
```diff
@@ -483,3 +486,36 @@ async def wifi_enable_to_code(config, action_id, template_arg, args):
 @automation.register_action("wifi.disable", WiFiDisableAction, cv.Schema({}))
 async def wifi_disable_to_code(config, action_id, template_arg, args):
     return cg.new_Pvariable(action_id, template_arg)
+
+
+@automation.register_action(
+    "wifi.set_sta",
```
jesserockz: Not sure I like the action name being `set_sta`. This refers to an internal function name and is not exactly user friendly for the yaml side. 
Maybe simply `wifi.connect`?
```suggestion
    "wifi.connect",
```

jesserockz (changes requested):
Further to my requested name change below, I think adding a `save: bool` config to the schema to allow users to not save to prefs (can default to true) would be nice too.

esphome/components/wifi/__init__.py
pzich: I feel like `connect` could be confused with just enabling/disabling it. Maybe `update`, `change`, or even just `set`?

jesserockz: This is true, maybe `wifi.configure`?

pzich:
Probably outside of the scope of this PR, but it might be useful if this could optionally take `networks` instead (similar to the wifi config itself) to update/replace the whole set of ssid/pws in the list.

Rapsssito:
@jesserockz, @pzich, I have implemented your suggestions. Let me know what you think!

Regarding multiple networks, sadly I am not familiar with that kind of setup. @pzich, feel free to build on top of my PR!

jesserockz (approved):

PR #7342, opened 2024-08-23
[ledc] Fix maximum brightness on ESP-IDF 5.1

clydebarrow:

The LEDC component output drops to zero when at maximum brightness using ESP-IDF 5.1. This problem is not observed with ESP-IDF 4.x or with Arduino. I could find no reports of bugs in ESP-IDF or anything in the docs or source code to explain this.

This PR stops the PWM output in the ON state at maximum duty cycle, and in the OFF state at 0.

Testing with IDF 4.x and 5.1 demonstrates correct behaviour.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

output:
  - platform: ledc
    pin: GPIO15
    id: backlight_output
    frequency: 2441Hz

light:
  - platform: monochromatic
    output: backlight_output
    name: LCD Backlight
    id: led
    restore_mode: ALWAYS_ON
```

## Checklist:

### DIFF
--- a/esphome/components/ledc/ledc_output.cpp
+++ b/esphome/components/ledc/ledc_output.cpp
@@ -115,8 +115,8 @@ void LEDCOutput::write_state(float state) {
   const uint32_t max_duty = (uint32_t(1) << this->bit_depth_) - 1;
   const float duty_rounded = roundf(state * max_duty);
   auto duty = static_cast<uint32_t>(duty_rounded);
+  ESP_LOGV(TAG, "Setting duty: %" PRIu32 " on channel %u", duty, this->channel_);
 #ifdef USE_ARDUINO
-  ESP_LOGV(TAG, "Setting duty: %u on channel %u", duty, this->channel_);
   ledcWrite(this->channel_, duty);
 #endif
 #ifdef USE_ESP_IDF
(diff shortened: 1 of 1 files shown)


kbx81 (dismissed):
Gah! I just saw this...not long after I merged #7336. 😅 I like this better. Thanks @clydebarrow!

esphome/components/ledc/ledc_output.cpp
```diff
@@ -1,3 +1,4 @@
+#include <cinttypes>
```
kbx81:
```suggestion
```

kbx81 (approved):

kbx81:
I just realized -- I don't think this will work as expected if the output is inverted...

clydebarrow:
> don't think this will work if the output is inverted

Yes it does, I tested it.

kbx81:
Perfect! Thanks again 🍻

clydebarrow:
> I just realized -- I don't think this will work as expected if the output is inverted...

I initially thought it would need the stop value inverted, but realized the inversion is done earlier by :
```
  if (this->pin_->is_inverted())
    state = 1.0f - state;
```

PR #7024, opened 2024-07-01
Tuya Number: allow to set hidden datapoints

szupi-ipuzs:
# What does this implement/fix?

Some Tuya devices do not publish all of their datapoints at initialization, but they still allow writing to the so-called "hidden" datapoints.
Current implementation of Tuya Number requires to know the datapoint type upfront. In case of the "hidden" datapoints, their type remains unknown and thus it is not possible to set such a Number.

This PR adds a yaml config block which allows specifying missing datapoint type and also (optionally) the initial value to be written to datapoint upon the initialization.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4010

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
    - platform: "tuya"
      name: "Alarm at maximum"
      number_datapoint: 116
      min_value: 0
      max_value: 100
      datapoint_hidden:
        datapoint_type: int
        initial_value: 85
```

## Checklist:
  - [ ] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #7346, opened 2024-08-23
Tuya Number: allow restoring value of hidden datapoints

szupi-ipuzs:

This is a continuation of #7024.
The hidden datapoints are actually stored (and restored) by TuyaMCU itself, but there seem to be no way of reading them (because TuyaMCU does not publish them either at all or does it only only _after_ they are written).

My previous change (#7024) allowed the user to set the number to a specific and always the same initial_value, thus overwriting whatever might have been restored by TuyaMCU.

This PR adds the _restore_value_ yaml key to allow storing the hidden Tuya Number value on esphome side and restoring it at startup.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4191

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
    - platform: "tuya"
      name: "Alarm at maximum"
      number_datapoint: 116
      min_value: 0
      max_value: 100
      datapoint_hidden:
        datapoint_type: int
        initial_value: 85
        restore_value: True
```

## Checklist:


esphome/components/tuya/number/tuya_number.cpp
```diff
@@ -23,8 +37,22 @@ void TuyaNumber::setup() {
   });
 
   this->parent_->add_on_initialized_callback([this] {
-    if ((this->initial_value_) && (this->type_)) {
-      this->control(*this->initial_value_);
+    if (this->type_) {
+      float value;
+      if (!this->restore_value_) {
+        value = *this->initial_value_;
```
jesserockz: This is not checking that `inital_value` actually has a value

jesserockz (changes requested):

jesserockz (approved):

PR #7347, opened 2024-08-24
[core] fix running script\build_codeowners.py on windows.

nielsnl68:

windows setup needs to iinstall python-magic-bin. I was not able to run the script\build_codeowners.py before this was installed on my system.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:
As far as I can see the only use for libmagic is to identify an SVG file, but to do that it's bringing in a rather heavyweight package that depends on an external native library. And that library apparently just identifies an SVG file by searching for the string "<svg" so maybe that's all that needs doing, and then the requirement for libmagic can be removed.

clydebarrow:
Something like this:
```
diff --git a/esphome/components/image/__init__.py b/esphome/components/image/__init__.py
index e5a205f1e..2e38f2b96 100644
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -6,7 +6,8 @@ import hashlib
 import io
 from pathlib import Path
 import re
-from magic import Magic
+
+from PIL import UnidentifiedImageError

 from esphome import core
 from esphome.components import font
@@ -280,17 +281,18 @@ async def to_code(config):
     except Exception as e:
         raise core.EsphomeError(f"Could not load image file {path}: {e}")

-    mime = Magic(mime=True)
-    file_type = mime.from_buffer(file_contents)
-
     resize = config.get(CONF_RESIZE)
-    if "svg" in file_type:
-    else:
+    try:
         if resize:
             image.thumbnail(resize)
-
+    except UnidentifiedImageError:
+        if "<svg" in str(file_contents):
+            image = load_svg_image(file_contents, resize)
+        else:
+            raise core.EsphomeError(
+                f"Could not load image file {path}"
+            )
     width, height = image.size

     if CONF_RESIZE not in config and (width > 500 or height > 500):
```

jesserockz (dismissed):

nielsnl68:
> As far as I can see the only use for libmagic is to identify an SVG file, but to do that it's bringing in a rather heavyweight package that depends on an external native library. And that library apparently just identifies an SVG file by searching for the string "<svg" so maybe that's all that needs doing, and then the requirement for libmagic can be removed.

I see, Not sure why the **`build_codeowners.py`** needs to check if a file has a "<svg"?

nielsnl68:
@clydebarrow, About the given solution, i disagree on that. I Believe it would work for sure.
But i belief it is as nesty fix as what i suggest to do in this PR.

As i asked above why should the **`build_codeowners.py`** needs to check if a file has a "<svg"?
In my belief **`build_codeowners.py`** should validate of the **`CODEOWNERS`** is in the `__init__.py` etc. and the other stuff inside those files should be untouched.

clydebarrow:
the problem is that the script loads all the components in the ESPHome tree to be able to check if there is a CODEOWNERS property in each one, and it's the image component that references libmagic, so when it gets loaded it complains if the library is absent. I'll raise a separate PR to remove that dependency, which is a pain point irrespective of its effect on the codeowners script.

clydebarrow:
https://github.com/esphome/esphome/pull/7365

PR #7362, opened 2024-08-27
Allow Update Timezone

RebbePod:

Allows setting and applying a different timezone

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/time/real_time_clock.h
+++ b/esphome/components/time/real_time_clock.h
@@ -23,6 +23,9 @@ class RealTimeClock : public PollingComponent {
   /// Set the time zone.
   void set_timezone(const std::string &tz) { this->timezone_ = tz; }
 
+  // Apply the time zone.
+  void apply_timezone() { this->apply_timezone_(); }
+
   /// Get the time zone currently in use.
   std::string get_timezone() { return this->timezone_; }
 
@@ -36,8 +39,6 @@ class RealTimeClock : public PollingComponent {
   time_t timestamp_now() { return ::time(nullptr); }
 
   void call_setup() override;
-  
-  void apply_timezone() { this->apply_timezone_(); }
 
   void add_on_time_sync_callback(std::function<void()> callback) {
     this->time_sync_callback_.add(std::move(callback));
--- a/esphome/components/time/real_time_clock.h
+++ b/esphome/components/time/real_time_clock.h
@@ -36,6 +36,7 @@ class RealTimeClock : public PollingComponent {
   time_t timestamp_now() { return ::time(nullptr); }
 
   void call_setup() override;
+  void apply_timezone_();
 
   void add_on_time_sync_callback(std::function<void()> callback) {
     this->time_sync_callback_.add(std::move(callback));
@@ -46,7 +47,6 @@ class RealTimeClock : public PollingComponent {
   void synchronize_epoch_(uint32_t epoch);
 
   std::string timezone_{};
-  void apply_timezone_();
 
   CallbackManager<void()> time_sync_callback_;
 };
--- a/esphome/components/time/real_time_clock.h
+++ b/esphome/components/time/real_time_clock.h
@@ -36,7 +36,8 @@ class RealTimeClock : public PollingComponent {
   time_t timestamp_now() { return ::time(nullptr); }
 
   void call_setup() override;
-  void apply_timezone_();
+  
+  void apply_timezone() { return this->apply_timezone_(); }
 
   void add_on_time_sync_callback(std::function<void()> callback) {
     this->time_sync_callback_.add(std::move(callback));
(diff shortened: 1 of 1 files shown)


dwmw2:
Not sure about exposing this internal function on its own. Perhaps we should expose a new method which does the second half of what `synchronize_epoch_()` already does — logs the `Synchronized time:` message, and invokes the `time_sync_callback_`.

dwmw2:
Or maybe just make the existing `set_timezone()` method also apply the change, if invoked after setup?

RebbePod:
The latter makes sense to me

RebbePod:
Refresh

RebbePod:
Refresh

RebbePod:
Bump

bdraco:
There are conflicts on this PR. Please update the PR description to explain the purpose of this PR and resolve the conflicts.

Thanks

RebbePod:
@bdraco It's been years lol
The idea if I'm not mistaken was to be able to set the timezone from the webserver. For example for vendors that would sell a device the user can configure the timezone directly and simply.

See the discussion here
https://discord.com/channels/429907082951524364/1275608813881524338/1278132346012307538

About the conflict I don't remember the ins and outs anymore, do you know what's needed to get it to fit nicely?

bdraco:
We now have an API to send a pre-parsed TimeZone and its now auto set from Home Assistant:

https://github.com/esphome/aioesphomeapi/blob/169206671766f4ef24c5896b7648cb153f4b66af/aioesphomeapi/posix_tz.py#L42
https://github.com/esphome/esphome/blob/62d0c25a2bfd3da6afcfdd3c9f869c18432cbe18/esphome/components/api/api.proto#L876

The current plan is to remove the on device POSIX timezone parser to reduce the binary bloat in 2026.9.0 since nothing is using it currently except for legacy Home Assistant clients: https://github.com/esphome/esphome/pull/14237

RebbePod:
What about people using esphome without home assistant?

bdraco:
> What about people using esphome without home assistant?

What do you mean?  

Its been added to the native API so it can be set via Home Assistant, using `aioesphomeapi`, or any client that implements the native API in addition to codegen / YAML.

RebbePod:
What would be a sample yaml to be able to set the timezone via web server?

bdraco:
Nobody has support for setting it via web_server so thats not possible

edwardtfn:
> We now have an API to send a pre-parsed TimeZone and its now auto set from Home Assistant:
> 
> https://github.com/esphome/aioesphomeapi/blob/169206671766f4ef24c5896b7648cb153f4b66af/aioesphomeapi/posix_tz.py#L42
> 
> https://github.com/esphome/esphome/blob/62d0c25a2bfd3da6afcfdd3c9f869c18432cbe18/esphome/components/api/api.proto#L876
> 
> The current plan is to remove the on device POSIX timezone parser to reduce the binary bloat in 2026.9.0 since nothing is using it currently except for legacy Home Assistant clients: #14237

But POSIX could be solved from customer side (HA or something else) and sent just the useful info like `EST5EDT,M3.2.0/2:00:00,M11.1.0/2:00:00`. Even if ESPHome doesn't know what `EST5EDT` means, the rest of the string would be sufficient for calculating the time properly.

bdraco:
> > We now have an API to send a pre-parsed TimeZone and its now auto set from Home Assistant:
> > [esphome/aioesphomeapi@`1692066`/aioesphomeapi/posix_tz.py#L42](https://github.com/esphome/aioesphomeapi/blob/169206671766f4ef24c5896b7648cb153f4b66af/aioesphomeapi/posix_tz.py#L42)
> > https://github.com/esphome/esphome/blob/62d0c25a2bfd3da6afcfdd3c9f869c18432cbe18/esphome/components/api/api.proto#L876
> > 
> > The current plan is to remove the on device POSIX timezone parser to reduce the binary bloat in 2026.9.0 since nothing is using it currently except for legacy Home Assistant clients: #14237
> 
> But POSIX could be solved from customer side (HA or something else) and sent just the useful info like `EST5EDT,M3.2.0/2:00:00,M11.1.0/2:00:00`. Even if ESPHome doesn't know what `EST5EDT` means, the rest of the string would be sufficient for calculating the time properly.

Parsing could be done in JS to set the fields via a new web_server API. Realistically I think you would just want a button to send the systems local time in the browser, but maybe someone would want something more complex. 

Regardless this isn't the right place for this type of [discussion](https://github.com/orgs/esphome/discussions)

RebbePod:
@bdraco I think it would be really usefully to have a way to set it via web server.

Would you like me to open a new feature request?

RebbePod:
Any update on this?

PR #7365, opened 2024-08-28
[image] Remove dependency on libmagic

clydebarrow:

Use of libmagic is problematic on windows and requires native libraries installed on other host systems, and running the codeowners script requires it to be installed due to it being used by `image`. It's only used to detect an SVG file and in any case does so solely by searching for a string. This PR replicates that behaviour without bringing in a heavyweight external library.

Also change `display` from a dependency to an auto-load since image does not depend on having a display configured, only the header file available.

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/esphome/pull/5896

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [ ] host

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/image/__init__.py
```diff
@@ -239,12 +238,11 @@ def _file_schema(value):
 
 def load_svg_image(file: bytes, resize: tuple[int, int]):
     # Local import only to allow "validate_pillow_installed" to run *before* importing it
```
paravoid: This comment needs to be moved alongside the import.

esphome/components/image/__init__.py
```diff
@@ -1,18 +1,17 @@
...
-
 import hashlib
 import io
+import logging
 from pathlib import Path
 import re
-from magic import Magic
 
-from esphome import core
+from PIL import UnidentifiedImageError
```
paravoid: Doesn't this need to be a local import, like other pillow imports, for `validate_pillow_installed` etc.?

nielsnl68:
Did you intended to push 2 changes in this PR :D

esphome/components/image/__init__.py
```diff
@@ -280,17 +278,16 @@ async def to_code(config):
...
-    if "svg" in file_type:
-        image = load_svg_image(file_contents, resize)
-    else:
+    try:
         image = Image.open(io.BytesIO(file_contents))
         if resize:
             image.thumbnail(resize)
-
+    except UnidentifiedImageError as exc:
+        if "<svg" in str(file_contents):
```
paravoid: For what it's worth, libmagic has these rules:
```
0       string          \<?xml\ version=
>14     regex           ['"\ \t]*[0-9.]+['"\ \t]*
>>19    search/4096     \<svg                   SVG Scalable Vector Graphics image
!:mime  image/svg+xml
!:ext   svg

0       string          \<svg                   SVG Scalable Vector Graphics image
!:mime  image/svg+xml
!:ext   svg
```

In other words: `file_contents.startswith("<svg") or ( file_contents.startswith("<?xml version=") and "<svg" in file_contents[:4096]) )`.

Checking for "<svg" in the whole file seems a little broad, but I can't imagine it being _that_ risky - in the worst case `load_svg_image()` fails.

paravoid (changes requested):

clydebarrow:
> Did you intended to push 2 changes in this PR

Yes, without those it fails CI.

esphome/components/image/__init__.py
paravoid: @clydebarrow I don't think this was actually fixed?

nielsnl68 (approved):

esphome/components/image/__init__.py
clydebarrow: > I don't think this was actually fixed?

It was, but `ruff` un-fixed it.

guillempages:
@clydebarrow can you add issue 5896 to the "fixes" line in the description? As far as I can see, that issue is exactly what you are trying to fix here

guillempages:
IMHO the `invert_colors` changes on the tests belong in a separate PR... They are unrelated to this change, but to the one that made it mandatory

clydebarrow:
> IMHO the `invert_colors` changes on the tests belong in a separate PR

True, but without those (actually now only one) this PR won't pass CI. I'm not going to raise a separate PR to change a single line, sorry Jesse, but I'm drawing a line here :-)

jesserockz:
> True, but without those (actually now only one) this PR won't pass CI. I'm not going to raise a separate PR to change a single line, sorry Jesse, but I'm drawing a line here :-)

Your PR is adding `invert_colors` to the `spi` config :thinking: 
It already exists on the `display` for this test...

And its not failing becuase the 8266 commmon file is not used anywhere :see_no_evil:

jesserockz:
Another alternative is https://pypi.org/project/puremagic/
Especially if other file types are added and need to be identified before some code runs in the future.

kahrendt:
> Another alternative is https://pypi.org/project/puremagic/ Especially if other file types are added and need to be identified before some code runs in the future.

I like this option! I want to determine the type of an audio file (wav, flac, or mp3), and I'm currently using libmagic for it. Puremagic is pretty much a straight drop in replacement.

guillempages:
Since there seems to be no movement here, I've implemented the "puremagic" version in #7536.
So probably only one of the two PRs should be merged, since they address the same issue in two different ways

PR #7377, opened 2024-08-29
Add voice assistant announce

synesthesiam:

Adds a new `VoiceAssistantAnnounce` message that will allow media URLs to be announced if the device supports it.

* https://github.com/esphome/aioesphomeapi/pull/939
* https://github.com/home-assistant/core/pull/125157
* https://github.com/esphome/voice-kit/pull/64

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/api/api.proto
+++ b/esphome/components/api/api.proto
@@ -1553,12 +1553,27 @@ message VoiceAssistantTimerEventResponse {
   bool is_active = 6;
 }
 
-message VoiceAssistantAnnounce {
+message VoiceAssistantAnnounceRequest {
   option (id) = 120;
   option (source) = SOURCE_CLIENT;
   option (ifdef) = "USE_VOICE_ASSISTANT";
 
   string media_id = 1;
+  string text = 2;
+}
+
+message VoiceAssistantAnnounceFinished {
+  option (id) = 121;
+  option (source) = SOURCE_SERVER;
+  option (ifdef) = "USE_VOICE_ASSISTANT";
+
+  string media_id = 1;
+}
+
+message VoiceAssistantTextToSpeechFormat {
+  string format = 1;
+  uint32 sample_rate = 2;
+  uint32 num_channels = 3;
 }
 
 // ==================== ALARM CONTROL PANEL ====================
--- a/esphome/components/api/api.proto
+++ b/esphome/components/api/api.proto
@@ -1566,6 +1566,8 @@ message VoiceAssistantAnnounceFinished {
   option (id) = 120;
   option (source) = SOURCE_SERVER;
   option (ifdef) = "USE_VOICE_ASSISTANT";
+
+  bool success = 1;
 }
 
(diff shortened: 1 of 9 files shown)


synesthesiam:
Leaving as a draft since this is still under discussion.

esphome/components/api/api.proto
```diff
@@ -1554,6 +1554,14 @@ message VoiceAssistantTimerEventResponse {
   bool is_active = 6;
 }
 
+message VoiceAssistantAnnounce {
```
jesserockz:
```suggestion
message VoiceAssistantAnnounceRequest {
```

esphome/components/api/api.proto
```diff
@@ -1554,6 +1554,29 @@ message VoiceAssistantTimerEventResponse {
...
+
+  string media_id = 1;
+}
+
+message VoiceAssistantTextToSpeechFormat {
+  string format = 1;
+  uint32 sample_rate = 2;
+  uint32 num_channels = 3;
+}
+
```
jesserockz: I can see this is not being used yet, It should be removed from this PR, or added to the `aioesphomeapi` one to match please.

esphome/components/api/api.proto
```diff
@@ -1554,6 +1554,29 @@ message VoiceAssistantTimerEventResponse {
...
+  string text = 2;
+}
+
+message VoiceAssistantAnnounceFinished {
+  option (id) = 121;
+  option (source) = SOURCE_SERVER;
+  option (ifdef) = "USE_VOICE_ASSISTANT";
+
+  string media_id = 1;
+}
```
bdraco: We have a conflict with allocation of `id`s.  119 is the next available one, or we are missing a merge/change on the aioesphomeapi side

synesthesiam: 119 was the MediaPlayerSupported format, but I put an id on it not realizing that it wasn't needed (it's a sub-message). Should we use 119 and 120 here instead?

esphome/components/api/api.proto
synesthesiam: Removed

esphome/components/api/api.proto
jesserockz: 119 was accidentally assigned to `MediaPlayerSupportedFormat` but it doesnt need an id as it is not a message that is sent or received by itself, I will fix that to remove the id there. Then yes this PR needs to change to 119 as the first id.

jesserockz: https://github.com/esphome/esphome/pull/7406

jesserockz: https://github.com/esphome/aioesphomeapi/pull/948

jesserockz (approved):

PR #7384, opened 2024-09-01
Use "puremagic" instead of "magic" python module

guillempages:

Use the puremagic module to avoid extra dependencies, that sometimes cause problems in windows due to missing binary libraries. puremagic is a cross-platform pure python module that does not have binary dependencies.

The tradeoff is a lower number of detected files; for the current use-case should be enough though.

## Types of changes

- [x] Other -> Python only fix for Windows

**Related issue or feature (if applicable):** fixes #5896

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -1,18 +1,17 @@
 from __future__ import annotations
 
-import logging
-
 import hashlib
 import io
+import logging
 from pathlib import Path
 import re
-from magic import Magic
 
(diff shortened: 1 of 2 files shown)


guillempages:
Already done in #7365

PR #7390, opened 2024-09-02
Better support for task blocking ring buffer reads and writes

kahrendt:

Adds better support for task blocking when reading and writing. This allows FreeRTOS tasks to be set up more efficiently, as they don't need to run until they have enough data to proceed. 

When reading, if ``ticks_to_wait`` is nonzero, it will block until the requested amount of data is available or the amount of ticks has passed. If it is the default 0 ticks, then it will continue to use the current behavior and return as much data that is available instantly. The new ``write_without_replacement`` function never overwrites the oldest data in the ring buffer. It has an optional ``ticks_to_wait`` parameter so that the calling task blocks until the requested amount of data is free or the amounts of ticks has passed. Finally, I added comments to better describe what each function does.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/core/ring_buffer.cpp
+++ b/esphome/core/ring_buffer.cpp
@@ -26,6 +26,8 @@ std::unique_ptr<RingBuffer> RingBuffer::create(size_t len) {
 }
 
 size_t RingBuffer::read(void *data, size_t len, TickType_t ticks_to_wait) {
+  if (ticks_to_wait > 0)
+    xStreamBufferSetTriggerLevel(this->handle_, len);
   return xStreamBufferReceive(this->handle_, data, len, ticks_to_wait);
 }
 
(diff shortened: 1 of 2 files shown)


esphome/core/ring_buffer.cpp
```diff
@@ -26,6 +26,8 @@ std::unique_ptr<RingBuffer> RingBuffer::create(size_t len) {
 }
 
 size_t RingBuffer::read(void *data, size_t len, TickType_t ticks_to_wait) {
+  if (ticks_to_wait > 0)
+    xStreamBufferSetTriggerLevel(this->handle_, len);
```
jesserockz, line 30: Do we need to reset/set this back to 1 afterwards?

kahrendt: That's a good idea!

jesserockz (approved):

PR #7392, opened 2024-09-02
[image] Add Invert Colors option

boussouira:

This PR add a new  configuration variable to to invert the image colors. This is useful when drawing a dark image on dark background for example.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4217

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml
image:
  - file: fan.png
    id: ac_button_image
    resize: 50x50
    type: RGB565
    use_transparency: true
    reverse_colors: true


```

## Checklist:

### DIFF
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -313,7 +313,7 @@ async def to_code(config):
         )
 
     transparent = config[CONF_USE_TRANSPARENCY]
-    invert_colors = config.get(CONF_REVERSE_COLORS)
+    reverse_colors = config.get(CONF_REVERSE_COLORS)
 
     dither = (
         Image.Dither.NONE
@@ -337,7 +337,7 @@ async def to_code(config):
 
     elif config[CONF_TYPE] == "RGBA":
         image = image.convert("RGBA")
-        if invert_colors:
+        if reverse_colors:
             image = invert_image_colors(image)
 
         pixels = list(image.getdata())
@@ -355,7 +355,7 @@ async def to_code(config):
 
     elif config[CONF_TYPE] == "RGB24":
         image = image.convert("RGBA")
-        if invert_colors:
+        if reverse_colors:
             image = invert_image_colors(image)
 
         pixels = list(image.getdata())
@@ -379,7 +379,7 @@ async def to_code(config):
 
     elif config[CONF_TYPE] in ["RGB565"]:
         image = image.convert("RGBA")
-        if invert_colors:
+        if reverse_colors:
             image = invert_image_colors(image)
 
         pixels = list(image.getdata())
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -260,6 +260,7 @@ def load_svg_image(file: bytes, resize: tuple[int, int]):
 
     return Image.open(io.BytesIO(svg_image))
 
+
 def invert_image_colors(image):
     from PIL import Image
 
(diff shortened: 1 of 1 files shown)


esphome/components/image/__init__.py
```diff
@@ -47,6 +47,7 @@
 }
 
 CONF_USE_TRANSPARENCY = "use_transparency"
+CONF_REVERSE_COLORS = "reverse_colors"
```
guillempages, line 50: You might want to use "CONF_INVERT_COLORS" (from const.py) instead; to keep consistency with the displays.

Also, to me "reverse colors" sounds more like make BGR from RGB; not swap black and white ;-)

esphome/components/image/__init__.py
```diff
@@ -225,6 +226,7 @@ def _file_schema(value):
             # Not setting default here on purpose; the default depends on the image type,
             # and thus will be set in the "validate_cross_dependencies" validator.
             cv.Optional(CONF_USE_TRANSPARENCY): cv.boolean,
+            cv.Optional(CONF_REVERSE_COLORS, default=False): cv.boolean,
```
guillempages, line 229: Does reverse (or invert) colors actually make sense for anything except black&white images?

esphome/components/image/__init__.py
```diff
@@ -259,6 +261,18 @@ def load_svg_image(file: bytes, resize: tuple[int, int]):
...
+def invert_image_colors(image):
+    from PIL import Image
+
+    r, g, b, a = image.split()
+
+    def invert(image):
+        return image.point(lambda p: 255 - p)
+
+    r, g, b = map(invert, (r, g, b))
+    return Image.merge(image.mode, (r, g, b, a))
```
guillempages, line 273: Why not use pillow to do the inversion, instead of reinventing the wheel?
```suggestion
    import PIL.ImageOps
    
    return PIL.ImageOps.invert(image)
```

esphome/components/image/__init__.py
```diff
@@ -301,6 +315,7 @@ async def to_code(config):
         )
 
     transparent = config[CONF_USE_TRANSPARENCY]
+    reverse_colors = config.get(CONF_REVERSE_COLORS)
```
guillempages, line 318:
```suggestion
    invert_colors = config.get(CONF_INVERT_COLORS)
```

esphome/components/image/__init__.py
guillempages: This needs some testing though; it might fail for RGBA images

guillempages:
Why are the binary and greyscale images not inverted? To me those *are* the ones that make the most sense to invert 🤔

nagyrobi:
I would do it at compile time too, more over, i'd put it under a `filters` parent key to make room for further processing possibilities like rotate, sharpen etc:

```yaml
# Example config.yaml
image:
  - file: fan.png
    id: ac_button_image
    resize: 50x50
    type: RGB565
    use_transparency: true
    filters:
      - reverse_colors: true
      - flip_vertical: true #idea for future
```

ed6767:
this would be a very useful feature :)

PR #7415, opened 2024-09-05
[libretiny] Report version 1.7.0 for 'dev' and 'latest'

dwmw2:
- (Depends on https://github.com/esphome/esphome/pull/7408 )

Rather than reporting 0.0.0 for 'dev' and 'latest' branches, report 1.7.0 which is (currently) the truth.

Like other frameworks, this will be perpetually out of date because we don't update it at runtime (although perhaps we could have a script which updates it prior to making an ESPHome release). It's not that much of a problem though, as we mostly use this for requiring a new enough version for certain features, and those tend not to depend on *future* versions.

So reporting 1.7.0 is enough for the upcoming IPv6 support, and it's OK for it to stay on 1.7.0 even if the git respository has something newer, until such time as something in ESPHome actually *cares* about checking for a newer version.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/libretiny/__init__.py
+++ b/esphome/components/libretiny/__init__.py
@@ -1,10 +1,6 @@
 import json
 import logging
-from os.path import (
-    dirname,
-    isfile,
-    join,
-)
+from os.path import dirname, isfile, join
 
 import esphome.codegen as cg
 import esphome.config_validation as cv
@@ -177,8 +173,8 @@ def _notify_old_style(config):
 # NOTE: Keep this in mind when updating the recommended version:
 #  * For all constants below, update platformio.ini (in this repo)
 ARDUINO_VERSIONS = {
-    "dev": (cv.Version(0, 0, 0), "https://github.com/libretiny-eu/libretiny.git"),
-    "latest": (cv.Version(0, 0, 0), None),
+    "dev": (cv.Version(1, 7, 0), "https://github.com/libretiny-eu/libretiny.git"),
+    "latest": (cv.Version(1, 7, 0), "libretiny"),
     "recommended": (cv.Version(1, 5, 1), None),
 }
 
--- a/esphome/components/libretiny/__init__.py
+++ b/esphome/components/libretiny/__init__.py
@@ -172,7 +172,7 @@ def _notify_old_style(config):
 
 # NOTE: Keep this in mind when updating the recommended version:
 #  * For all constants below, update platformio.ini (in this repo)
-# The dev and latest branches will be at *least* this versiom, which is what matters.
+# The dev and latest branches will be at *least* this version, which is what matters.
 ARDUINO_VERSIONS = {
     "dev": (cv.Version(1, 7, 0), "https://github.com/libretiny-eu/libretiny.git"),
     "latest": (cv.Version(1, 7, 0), "libretiny"),
--- a/esphome/components/libretiny/__init__.py
+++ b/esphome/components/libretiny/__init__.py
@@ -172,6 +172,7 @@ def _notify_old_style(config):
 
 # NOTE: Keep this in mind when updating the recommended version:
 #  * For all constants below, update platformio.ini (in this repo)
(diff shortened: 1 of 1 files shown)


kuba2k2:
One problem here is that setting "latest" will now add the version number to PlatformIO. It used to put simply "libretiny" into platformio.ini, which allowed to use whatever version was installed on the system - including development versions, i.e. not pushed to git, with unsaved changes. This feature is crucial for development, so I'd rather have it available this way.

dwmw2:
Ah, I thought I'd tested that and it still worked (and I've certainly still been building with locally modified versions of LibreTiny). Will take another look.

May fall back to allowing 0.0.0 as "new enough" for IPv6 support as a special case.

dwmw2:
This seems to work:
```
ARDUINO_VERSIONS = {
    "dev": (cv.Version(1, 7. 0), "https://github.com/libretiny-eu/libretiny.git"),
    "latest": (cv.Version(1, 7, 0), "libretiny"),
    "recommended": (cv.Version(1, 5, 1), None),
}
```

kuba2k2:
It should indeed work when applied together with #7408. But alone, it will make the following config:
```yaml
framework:
  version: latest
```
result in this framework string: `libretiny @ 1.7.0`, which is a breaking change (e.g. for developers).

I'd suggest joining these two PRs - they are related to the same bugfix/feature and they're small enough to be merged as a single PR.

Another issue - which was also present in ESPHome before - is that using custom `source:` is not really user-friendly. It requires setting either `version: 0.0.0` (which is totally odd) or setting the "real" version string. Ideally, setting `source:` shouldn't require setting `version:` at all - after all, `source:` is most likely used by developers/beta testers only.

So, when specifying `source:` without `version:`, the function `_check_framework_version()` could instead populate `CONF_VERSION` with the version number from `ARDUINO_VERSIONS["dev"]`.

Even better - why use two keys altogether? Something like `version: https://github.com/....` would be perfectly fine IMHO. The same `version:` property could solely be used to select the version used, whether it's dev/recommended/latest, a custom version number, or a custom Git branch.

dwmw2:
> I'd suggest joining these two PRs - they are related to the same bugfix/feature and they're small enough to be merged as a single PR.

I've no aversion to pulling the PRs together. My natural inclination would have been to make them *all* just separate commits in a single overall PR for BK72xx IPv6 support. I'm mostly just trying to fit in with @jesserockz's preferences and normal workflow, which seems to be using separate PRs and squashing each PR into a single commit in the `dev` branch. It does also have the benefit that the simpler, uncontentious changes can go in first and be out of the way while we concentrate on refining the more interesting parts.

So as long as #7408 is merged *first*, I think that's fine.

> Another issue - which was also present in ESPHome before - is that using custom `source:` is not really user-friendly

I suppose so, but a custom `source:` ideally isn't a *user* option anyway. That should be for developers. I do agree that setting `source:` shouldn't require setting `version:` though. Ideally we'd be able to *infer* the version from the `platform.json` in the source tree that we're pointed to — although it doesn't get checked out until much later, so that's hard.

But I don't see why we can't just default to `0.0.0` or `None`. For the purpose of ESPHome (especially #7409 and #7410), that would always be considered "new enough". And then yes, using a single key for either a version# or a direct source link ought to be fine. Slightly orthogonal to this issue though.

dwmw2:
@kuba2k2 Some of the discussion above is hypothetical, about things we *might* like to change. Just for the purposes of making it clear whether *this* CR is OK to merge... please could you confirm that you're happy with it, as long as #7408 goes in first? With explicitly specifying `libretiny` as the source in the `latest` case, I think it's all working correctly and your initial feedback is addressed?

Thanks.

kuba2k2 (approved):

jesserockz (approved):

PR #7425, opened 2024-09-09
Make time dependency optional

nkinnan:

This change allows decoupling the datetime component from any root level time component being defined.  This is useful in the case where, for example, your ESP is communicating with a second device that has an RTC and you would like to use the datetime component to expose reading and setting that value on the secondary device.  Defining a dummy time, or just using ntp or homeassistantapi, or using the secondary device's rtc as the esp's source of time may be a valid option for many, but there are situations in which you may not want or care for the ESP to know what time it is at all, and/or may not desire for the ESP's time to come *from* that secondary device either.  All you care about is exposing the secondary device's datetime without additional constraints.  This allows you to do so by only requiring time_id when an on_time trigger is defined, otherwise simply not compiling in the time dependencies.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4234

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

<yaml without time defined anywhere at the root level>

datetime:
  - platform: my_platform

    my_datetime_component:
      name: "Second System's non-authoritative Date and Time"
```

## Checklist:


RFDarter:
Are the wraps `#ifdef USE_TIME` even needed?

Maybe instead of making it only possible for custom components to define a dateime without the requirement of a time component it would make sense for the datetime component to require a time component only if `on_time` is set in the config?

nkinnan:
> Are the wraps `#ifdef USE_TIME` even needed?
> 
> Maybe instead of making it only possible for custom components to define a dateime without the requirement of a time component it would make sense for the datetime component to require a time component only if `on_time` is set in the config?

@RFDarter I like that idea, but I can't do that with runtime conditionals, it requires excluding a header file and not declaring some variables defined in that header at compile time.  If there is some way of getting a #define based on whether the dependency is in the YAML or not, I think this would be a good change to make.

In that case why not tie it to time_id directly?  But again, has to be compile-time.

RFDarter:
Couldn't we just set a define if `on_time` is set in the config?

nkinnan:
> Couldn't we just set a define if `on_time` is set in the config?

@RFDarter Can we? I'm not familiar with how the python and the codegen and the build system interact. If it's possible, I'd be happy to do so, though I think it might be better to set the define based on whether time_id is present instead.

nkinnan:
@RFDarter Figured it out, REQUIRES_TIME is now defined for the preprocessor and controls whether the header and rtc/trigger stuff is included.  It's controlled by init.py via the same parameter requires_time.  Much cleaner I think, thanks for the idea :)

RFDarter:
Hi, have you actually tested providing a `on_time` event in the config without setting a `time_id`?

To me it seems if you do that `config[CONF_TIME_ID]` would get set to `None` by this
```py
    if CONF_ON_TIME in config and CONF_TIME_ID not in config:
        time_id = cv.use_id(time.RealTimeClock)(None)
        config[CONF_TIME_ID] = time_id
```

and so this part would fail since `if CONF_TIME_ID in config:` is always true
```py
rtc = await cg.get_variable(config[CONF_TIME_ID])
```

EDIT: i see how it works now, `time_id = cv.use_id(time.RealTimeClock)(None)` will let the config fail if no time component is set, so all works as intended

nkinnan:
> Hi, have you actually tested providing a `on_time` event in the config without setting a `time_id`?

@RFDarter Per discussion on discord, `cv.use_id` is throwing the proper exception.  Thanks for the assist with testing while the build is broken, and for raising the concern.

(And thanks Jesse for moving the check into validate.)

RFDarter:
I tested it locally and it all seems to work as intended

jesserockz (approved):

PR #7429, opened 2024-09-09
[font] Add support for "glyphsets"

### LINKED ISSUE #2791: font renderer: support the notion of "glyphsets"

I'm Greek. I'd like to be able to show text, using the display/font component, in the Greek alphabet, in addition to the Latin alphabet.

For this purpose, there is the `glyph` attribute in the component's configuration, where I can add every character. However, that is not trivial to define once you get started. You need to enumerate all characters, as well as all the permutations of characters + accents (as no canonicalization form is applied). I've done the effort of course, but it's a shame for every Greek speaker to have to go through the same trouble.

It'd be great if there was a way to configure the component with some notion of "latin", "greek", and so on and so forth.

A few alternatives I've considered and seeking input for:
1. Use the notion of Unicode blocks. In my case, it would be "Greek and Coptic", U+0370 - U+03FF. stdlib's unicodedata does not include this functionality, but we could use a library like https://github.com/nagisa/unicodeblocks (unfortunately unmaintained). The other downside is that Unicode's blocks are fairly wide, e.g. in this case it includes a bunch of Coptic characters, which won't useful to me (or anyone else, really).
1. Statically define a few alphabets as map to to character ranges, in the ESPHome code, seeded from an original small list and then updated as needed, from users across the globe.
1. Use the Google Fonts Glyphsets Python library (`pip install glyphsets`), which includes glyph ranges (not blocks!) as defined by the Google Font team, such as `GF_Latin_Core`,  and `GF_Greek_Core`. These are way more accurate, but the library a) brings a bunch of dependencies b) has a few rough edges, as it was really designed for font authors etc. 

I actually have code that implements option (3), but I'm still wondering if it's the way to go and was hoping to agree on the way forward before opening up a PR :) For the dependencies, I've actually used the same trick as we do for `pillow`, where it's dynamically imported and if not found, prints an error - not sure if that's desired here?

The diffstat for the record is:
```
 esphome/components/font/__init__.py | 44 ++++++++++++++++++++++++++++++++++++++++++--
 esphome/const.py                    |  1 +
```

**Additional context**

List of Unicode blocks, for option 1: https://en.wikipedia.org/wiki/Unicode_block

List of Google Fonts glyphsets, for option 3:
 * GF_Arabic_Core
 * GF_Arabic_Plus
 * GF_Cyrillic_Core
 * GF_Cyrillic_Historical
 * GF_Cyrillic_Plus
 * GF_Cyrillic_Pro
 * GF_Greek_AncientMusicalSymbols
 * GF_Greek_Archaic
 * GF_Greek_Coptic
 * GF_Greek_Core
 * GF_Greek_Expert
 * GF_Greek_Plus
 * GF_Greek_Pro
 * GF_Latin_African
 * GF_Latin_Beyond
 * GF_Latin_Core
 * GF_Latin_Kernel
 * GF_Latin_Plus
 * GF_Latin_PriAfrican
 * GF_Latin_Vietnamese
 * GF_Phonetics_APA
 * GF_Phonetics_DisorderedSpeech
 * GF_Phonetics_IPAHistorical
 * GF_Phonetics_IPAStandard
 * GF_Phonetics_SinoExt
 * GF_TransLatin_Arabic
 * GF_TransLatin_Pinyin
(per `glyphsets.defined_glyphsets()`)

A few examples for `glyphsets`:
- `print("".join([chr(x) for x in glyphsets.unicodes_per_glyphset("GF_Latin_Kernel")]))`
- `print("".join([chr(x) for x in glyphsets.unicodes_per_glyphset("GF_Latin_Core")]))`
- `print("".join([ chr(x) for x in sorted(set(glyphsets.unicodes_per_glyphset("GF_Latin_Kernel")) | set(glyphsets
    ...: .unicodes_per_glyphset("GF_Greek_Core"))) ]))`

paravoid:

For space and efficiency reasons, the font component currently defaults to a small subset of glyphs, numbers, Latin alphabet characters and a few symbols.

It also allows customizing this set of glyphs, by specifying a different list of glyphs, or an additional set of glyphs. For speakers of non-Latin based languages it is, however, non-trivial and error prone for users to enumerate entire alphabets, especially when taking into account all the permutations of characters and accents.

This adds a new "glyphsets" attribute, that leverages Google Fonts Glyphset definitions, such as "GF_Latin_Core" and GF_Greek_Core", to allow users to concisely include glyphs for their own script/language.

Other changes either necessary or related that have been incorporated:

* Add tests for bitmap fonts
* Fix path handling for bitmap fonts
* Check for undefined glyphs, where specifically requested.
* The default glyphset (GF_Latin_Kernel) will be pared down to the subset supported by the specified font file.

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes esphome/feature-requests#2791

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4245

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] Host

```yaml
font:
  - file: "gfonts://Roboto"
    id: roboto_greek
    size: 24
    glyphsets:
      - GF_Latin_Core
      - GF_Greek_Core
```

## Checklist:


esphome/components/font/__init__.py
```diff
@@ -71,13 +73,34 @@ def validate_glyphs(value):
...
+    if isinstance(value, list):
+        value = cv.Schema([cv.string])(value)
+    value = cv.Schema([cv.string])(list(value))
+
+    for glyphset in value:
+        if glyphset not in glyphsets.defined_glyphsets():
+            raise cv.Invalid(f"{glyphset} is not a valid glyphset")
+    return value
+
+
```
clydebarrow:
```suggestion
```

esphome/components/font/__init__.py
```diff
@@ -313,6 +336,7 @@ def _file_schema(value):
         cv.Required(CONF_ID): cv.declare_id(Font),
         cv.Required(CONF_FILE): FILE_SCHEMA,
         cv.Optional(CONF_GLYPHS, default=DEFAULT_GLYPHS): validate_glyphs,
+        cv.Optional(CONF_GLYPHSETS): validate_glyphsets,
```
clydebarrow:
```suggestion
        cv.Optional(CONF_GLYPHSETS): cv.ensure_list(cv.one_of(*glyphsets.defined_glyphsets())),
```
In addition, if `glyphsets` is provided, the default for `glyphs` should be empty, I would think.

esphome/const.py
```diff
@@ -330,6 +330,7 @@
 CONF_GATEWAY = "gateway"
 CONF_GLASS_ATTENUATION_FACTOR = "glass_attenuation_factor"
 CONF_GLYPHS = "glyphs"
+CONF_GLYPHSETS = "glyphsets"
```
clydebarrow: This does not need to be in `const.py` since it's only used in one place. Define it in `font/__init__.py`.

esphome/components/font/__init__.py
```diff
@@ -71,13 +73,34 @@ def validate_glyphs(value):
...
+    for glyphset in config.get(CONF_GLYPHSETS, []):
+        for up in glyphsets.unicodes_per_glyphset(glyphset):
+            glyph = chr(up)
+            if glyph not in glyphs:
+                glyphs.append(glyph)
+
+    validate_glyphs(glyphs)
+
+    font_list = [(EFont(config[CONF_FILE], config[CONF_SIZE], glyphs.copy()))]
+
```
clydebarrow: This could be done a lot more simply and cleanly, but the rest of the function (i.e. my code) also needs some cleaning up. Basically best to use a set to accumulate the glyphs to avoid duplicates, then convert the set to a list and sort.

~Also, this does not need to be done as part of validation so should be called from `to_code`.~
The side-effect of checking for duplicate glyphs does need to be done as part of validation.

clydebarrow (dismissed):
I think the basic idea is sound, just the code needs cleaning up. You can tackle that yourself and I'll re-review, or I could just push a commit with the changes I think need happening - they're a little bit more complex than can easily be spelt out in a review and it's a good opportunity to clean up my earlier code. Your choice. Either way you should do the doc PR - this _will_ get done.

esphome/const.py
paravoid: I guess then `CONF_GLYPHS` also needs to move? Is that part of the cleanups to the original code you're envisioning here?

esphome/components/font/__init__.py
paravoid: I used sets originally and went back and forth myself. You a) basically lose the insertion order, which may or may not be the same as sorting order (at least for the "glyphs" option; I didn't check what glyphset does).
b) you need to set(glyphs) | set(...) for all these sets, which may get a little dense. Basically the data set is a few dozen items and figured that open coding this may make for an easier read :)

I definitely don't have a strong opinion about this, though.

esphome/components/font/__init__.py
paravoid: Fully agreed on the former, that's awesome :)

On the latter point w.r.t. emptying `glyphs`... it's a good point, and I think I agree with you but maybe worth thinking it through. I think in at least latin-adjacent countries (like Greek) it would uncommon to not have Latin glyphs at all in the set. (There are no fonts that support Greek but not Latin, for example). One could obviously add `GF_Latin_Kernel` or `GF_Latin_Core` to their list, but maybe that's superfluous?

paravoid:
@clydebarrow thanks for the super quick review, much appreciated!

> I think the basic idea is sound, just the code needs cleaning up. You can tackle that yourself and I'll re-review, or I could just push a commit with the changes I think need happening - they're a little bit more complex than can easily be spelt out in a review and it's a good opportunity to clean up my earlier code. Your choice.

I'm not sure if you mean something additional to the comments you've already left? I think I got a pretty good grasp of those, but I also want to be respectful of your time and avoid any potential additional back-and-forths, so feel free to make any changes you deem necessary! I have no attachments to anything in there, nor any particular reason to want to make changes myself :)

>  Either way you should do the doc PR - this _will_ get done.

Will do!

On another note:
- DEFAULT_GLYPHS = `` !"%()+=,-.:/?0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz°`` (78)
- GF_Latin_Kernel = `` !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~ ¢£¥©®°·×÷–—‘’“”•…€™−`` (116)
- Diff: ``#$&'*;<>@[\]^`{|}~ ¢£¥©®·×÷–—‘`` (38)

That's a +50% increase which is significant, but OTOH DEFAULT_GLYPHS is missing some that feel like very basic characters (`#`, `&`, `@`, `[`, `]` etc.). Perhaps `glyphset: GF_Latin_Kernel` should be the new default?

esphome/components/font/__init__.py
clydebarrow: > You a) basically lose the insertion order, which may or may not be the same as sorting order

The insertion order is unimportant - the generated font code requires the glyph table to be sorted so that a binary search can be done on it. The problem with the code isn't so much what you added, but that what I wrote there isn't all that readable anyway - I know that because coming back to it now after a few months I have trouble working out what it's doing :-( Some thought needs to go into the duplicate glyph check because clearly it shouldn't fail if you add multiple overlapping glyph sets, but should probably fail if a manually listed codepoint also exists in a glyphset. And glyphsets should probably be allowed in the `extras` schema as well.

esphome/const.py
clydebarrow: No, that was historic, so should be left untouched - it probably should not have gone in there originally, but now it's there there's not a good reason to move it. Basically when updating components it's best to avoid core changes unless necessary or at least with obvious benefit, for several reasons, not least because core changes make a branch or PR unusable as an external component.

clydebarrow:
> Perhaps `glyphset: GF_Latin_Kernel` should be the new default?

Yes I think that's a good idea. I did find the default a little restrictive.

clydebarrow:
> so feel free to make any changes you deem necessary! I have no attachments to anything in there, nor any particular reason to want to make changes myself :)

Roger that - I'll go ahead and push a commit - will probably take less time than back and forth. You can review after I've done it.

clydebarrow:
@paravoid Ok, have a look and see what you think. The  validation and code generation are more cleanly separated now.

Tests ok for me here.

esphome/const.py
```diff
@@ -331,7 +331,6 @@
 CONF_GAS_RESISTANCE = "gas_resistance"
 CONF_GATEWAY = "gateway"
 CONF_GLASS_ATTENUATION_FACTOR = "glass_attenuation_factor"
-CONF_GLYPHS = "glyphs"
```
paravoid: I think this (and its associated changes) is the opposite of what you intended, no? i.e. I understood that you wanted to move the newly introduced `CONF_GLYPHSETS` out of const, and keep `CONF_GLYPHS` for legacy reasons.

esphome/components/font/__init__.py
```diff
@@ -480,14 +445,26 @@ def __init__(self, data_len, offset_x, offset_y, width, height):
...
+    """
+    Collect all glyph codepoints, construct a map from a codepoint to a font file.
+    Codepoints are either explicit (glyphs key in top level or extras) or part of a glyphset.
+    Codepoints listed in extras use the extra font and override codepoints from glyphsets.
+    Achieve this by processing the base codepoints first, then the extras
+    """
+    # get the base codepoints from glyphs and glyphsets as a flat de-duped list.
+    setpoints = [glyphsets.unicodes_per_glyphset(x) for x in config[CONF_GLYPHSETS]]
+    setpoints = [chr(x) for x in sum(setpoints, [])]
+    codepoints = list(set(sum([list(x) for x in config[CONF_GLYPHS]], setpoints)))
```
paravoid: As I feared, this is a bit dense :)

In particular the use of `sum()` here is idiomatic and not technically correct (Python says "This function is intended specifically for use with numeric values and may reject non-numeric types."). `itertools.chain(*setpoints)` or `itertools.chain.from_iterable(setpoints)` can have the same effect of flattening the list of lists.

The merger with CONF_GLYPHS can be done in a simpler way by making setpoints a set, and then unioning it, i.e. `setpoints = set(setpoints).union(config[CONF_GLYPHS])` . Not sure where you stand on functional vs. list comprehension but you could also use `map`, e.g. `setpoints = map(glyphsets.unicodes_per_glyphset, config[CONF_GLYPHSETS])`.

So all in all:
```python
setpoints = map(glyphsets.unicodes_per_glyphset, config[CONF_GLYPHSETS])
setpoints = itertools.chain.from_iterable(setpoints)  # flatten
codepoints = set(map(chr, setpoints))  # set of unique characters
codepoints = codepoints.union(config[CONF_GLYPHS])  # extend with manually defined glyphs
```

It's a little too functional for my own taste, but I find it more readable than your version. YMMV :smile:

(untested)

esphome/const.py
clydebarrow: Doh!

esphome/components/font/__init__.py
clydebarrow: I personally think dense code with a few explanatory comments is easier to comprehend than verbose code, with or without comments, if only because there's less to look at.

The use of `map()` has been reviewed unfavourably vs. list or dict comprehension in ESPHome when I've used it in previous PRs. There's little in it either way so I don't care.

itertools would indeed be a better choice given the ambiguity in the Python docs about sum().

>
```
codepoints = codepoints.union(config[CONF_GLYPHS])  # extend with manually defined glyphs
```

won't work as is, `config[CONF_GLYPHS]` is a list of possibly multi-character strings that has to be flattened first. That's just how the schema is written, and I don't want to break it. As an aside, that aspect of the schema could be better explained in the docs when you're doing a PR.

I'll revise it with itertools.

clydebarrow:
There are pre-existing issues handling image font files - as in it is broken. But even when fixed it also doesn't work with the default glyphset - will need to sort that tomorrow. Pushing work so far.

esphome/components/font/__init__.py
paravoid: > I personally think dense code with a few explanatory comments is easier to comprehend than verbose code, with or without comments, if only because there's less to look at.

A long time ago in a different life I wrote (and read) a lot of Perl, so perhaps I'm slightly traumatized :smile: I don't feel strongly either way.

> config[CONF_GLYPHS] is a list of possibly multi-character strings that has to be flattened first

Ugh, I missed that part! Good catch.

Side note: this could really benefit from type annotations to both avoid errors creeping in, and enhancing readability. How do you feel about them, and do you know if these are welcome in ESPHome? (I don't see mypy in CI, so there's at least that downside)

clydebarrow: Type annotations are welcome and used a bit, though I've run across places where they were used, but wrong, so without enforcement it's a bit of a mixed bag since a misleading annotation is worse than none.

clydebarrow:
Well, this turned into more of a marathon than I thought. 

The bitmap font loading was not being tested, and failed when tested. Fixed path handling for that.
Checks are now done for missing glyphs that are specifically requested (not for default glyphs)

paravoid:
Wow, OK, didn't expect to nerdsnipe you to refactor it all! It'll take me some time to review it all.

As an initial point, I don't think this "missing glyphs" check you added is a good idea. Many fonts have partial support for some of these glyphsets but are otherwise perfectly usable.

For example, even with Roboto (which has a pretty good coverage generally) a compilation currently fails even with GF_Latin_Core alone:
```
  Roboto@400@False@v1.ttf is missing glyphs: ̂ (b'\\u0302'), ̄ (b'\\u0304'), ̆ (b'\\u0306'), ̇ (b'\\u0307'), ̈ (b'\\u0308'), ̊ (b'\\u030a'), ̋ (b'\\u030b'), ̌ (b'\\u030c'), ̦ (b'\\u0326'), ̧ (b'\\u0327'), ̨ (b'\\u0328'), ẞ (b'\\u1e9e').
```
It also fails with GF_Greek_Core for what it's worth, due to some infrequently used characters (fun fact: two of these glyphs were a matter of intense debate between Adobe and Google on whether they should be part of the core set!).

clydebarrow:
> I don't think this "missing glyphs" check you added is a good idea.

It may need some tweaking. It already is suppressed when using a default glyphset (which would otherwise be a breaking change) so maybe it needs an option to say "I'm ok with missing glyphs in the sets I asked for". Or maybe never complain about missing ones that come only from a set. But then the risk is that someone will be surprised when they asked for Greek and didn't get it, and weren't told in advance. I would prefer the option to permit

The current situation where you can request specific glyphs and get no feedback that they don't exist in the font is unacceptable IMO, 

Another possible requirement would be to list glyphs to omit - was not needed before, but with the glyphsets available I can see that being requested. And the Google glyphsets don't seem to include an ASCII set, so maybe that needs to be added too. Or the ability to specify a range perhaps. It's a bit tedious to have to list the whole alphabet.

I have hijacked this PR but it highlighted existing issues that really need addressing, as well as raising new ones - including CI build issues.The end result should be great though.

paravoid:
> It may need some tweaking. It already is suppressed when using a default glyphset (which would otherwise be a breaking change) so maybe it needs an option to say "I'm ok with missing glyphs in the sets I asked for". Or maybe never complain about missing ones that come only from a set. But then the risk is that someone will be surprised when they asked for Greek and didn't get it, and weren't told in advance.

I see where you are going with this, and selecting a font that doesn't e.g. support Greek at all while specifying `GF_Greek_Core` in glyphsets is a real problem that it'd be ideal to catch at build time. But I fear it's (much) more complicated than that. Glyphsets are not universally standardized and fonts that claim to support e.g. "Greek" are not required to include all glyphs anyway.

As I mentioned above, there is a specific example where Adobe (and its font foundry) [disagree with Google](https://github.com/googlefonts/glyphsets/issues/12) on whether some glyphs are "core" or not, and thus whether they should be included in all of their fonts that claim to support Greek (or not). Other characters are not practically in use, for instance no operating system that I know of emits ";" (U+037E, "Greek Question Mark", the Greek equivalent of "?") in their Greek keyboard maps. They emit instead ";" (U+003B, "Semicolon"). Therefore fonts, such as Roboto, sometimes don't include the glyph for the former for space-saving reasons. Then there's the matter of composed characters such as accented characters, and Unicode normalization forms, which is its own mess.

> I would prefer the option to permit

That was trimmed, not sure what you wanted to say here!

> Another possible requirement would be to list glyphs to omit - was not needed before, but with the glyphsets available I can see that being requested.

That doesn't sound like a bad feature to have but I'd say that's orthogonal. It's not a good solution for the problem above as it would be quite tedious to have to enumerate characters that are not in the font, especially given that they are typically rare characters that you probably are not even able to type using a standard keyboard.

> And the Google glyphsets don't seem to include an ASCII set, so maybe that needs to be added too.

Maybe? Or maybe we should let ASCII finally die! What's the use case you're thinking of?

> Or the ability to specify a range perhaps. It's a bit tedious to have to list the whole alphabet.

This is dangerous. Range of what? "a-z" means different things depending on the language. For example, does that range include ẞ? Å?

You could include ranges of (numeric) codepoints, but then it assumes that users are aware of how scripts are arranged within Unicode blocks, which is non-trivial by itself.

(i18n is hard!)

> I have hijacked this PR but it highlighted existing issues that really need addressing, as well as raising new ones - including CI build issues.The end result should be great though.

Thanks so much for that & I think so too! I do fear of scope creep a little bit (bitmap fonts et al) and how this will get reviewed and merged, but you've had much more complicated changes merged so I don't worry too much :wink:

nagyrobi:
We need at least a warning if the user chooses a font file which misses certain glyphs.

clydebarrow:
Re ASCII - the use case is me, and every other denizen of English-speaking countries that use $ as the currency symbol. But this could be addressed by implementing ranges - as you point out this should probably require numeric codepoints rather than characters, so I could just ask for something like `U20-U7F` for example, under the `glyphsets` key to avoid misinterpretation. Or maybe `U20 .. U7F` - any other ideas?

I don't think you should worry too much about user knowledge of Unicode blocks - anyone using that feature above 0x7F would have to at least know what a codepoint is.

Omitting glyphs could use a key `omit_glyphs` and/or `omit_glyphsets` - for the latter it would mainly be for ranges.

And an option `ignore_missing_glyphs` defaulting to false would work to satisfy the need for build-time alerts without being burdensome I think.

Since now it's possible to tell what glyphs are in a font (Pillow doesn't expose that, I had to use freetype to achieve that) perhaps the default glyphset if none is specified should be the entirety of the font. Though that would be a breaking change - not badly breaking but breaking nonetheless.

Or maybe require at least one of glyphs and glyphsets? Again a breaking change, but not a silent one.

clydebarrow:
Oh, maybe allow `glyphs: all` to specify inclusion of the entire font.

paravoid:
Honestly I'm getting a little lost with all the subthreads! Broadly speaking as I mentioned before I think this has scope-creeped a bit, and we're now discussing multiple additional features, stacked to each other, some of which we don't have consensus on. Don't get me wrong: I'm very appreciative of the additional work that has gone into what I started as a simple PR! But I'd prefer it if we kept this PR focused, make some progress and then discuss all these additional features in their own issues/PRs?

Additionally, my POV is that the component should remain lean, and features should be added once a clear use case has been demonstrated, perhaps through a user feature request.  Unicode is complicated, and the more we expose to users the more confusing it can get. YMMV :)

I'lll try to summarize my thoughts on some of the subthreads here:

### Missing glyphs warnings/errors

First of all, I fully agree that for the glyphs that the user _specifically_ requested (under the `glyphs` key), it makes sense to emit warnings or even errors.

On the other hand, I strongly believe that is not a good idea when it comes to glyphsets. Fonts are not required to implement (Google's) notion of a glyphset before claiming they support a script and/or language. Basically, glyphsets are aspirational, and sub-100% coverage does not make a font unusable.

Even extremely popular fonts do not have 100% coverage of basic sets such as `GF_Latin_Core`. Some examples just from my system: Helvetica, Helvetica Neue, all of the MS Core Fonts (Andale Mono, Arial, Comic Sans MS, Courier New, Georgia, Impact, Times New Roman, Trebuchet, Verdana), newer MS fonts (Carlita, Caladea), Gill Sans, Myriad Pro are all not at 100% and would error out with this revision of the PR.

For `GF_Greek_Core` it's even worse, partially but not exclusively because of the Adobe/Google disagreement I mentioned before: again all of the MS Core fonts, Liberation, Source Sans 4/Serif, FreeSans, Liberation, etc. would all emit errors. I think beyond a few fonts that are explicitly attempting to have 100% coverage (like DejaVu and Noto, I'd have a hard time finding a font with 100% `GF_Greek_Core` coverage.

Basically, I don't think it's sensible to say that Arial and Times New Roman are not usable with the core Latin set, or the core Greek set for that matter.

As such, I believe emitting errors or even warnings is just going to be annoying for the vast majority of `glyphset` users. Not even word editors like LibreOffice do that, though, so I dispute the necessity. I'd expect most users that explicitly configure a font _and_ glyphset to know whether their font of choice supports the glyphset of their choice.

But, if we /really/ want to avoid users misconfiguring their system in this way, then the best I can think of is to calculate a ratio of coverage (`missing_glyphs/codepoints`), and then emit warnings over some threshold, e.g. if more than 10-20% of a glyphset is missing. 

### Adding an ASCII glyphset

> Re ASCII - the use case is me, and every other denizen of English-speaking countries that use $ as the currency symbol

I should have been more clear! What is the use case for ASCII-only, as opposed to `GF_Latin_Kernel` (which is: "Support ASCII + necessary punctuation and symbols for English language")

### Glyph ranges

I am both unconvinced by the use case (who/what is this for?) and also fear that it's going to get extra confusing. The intricacies of Unicode and how characters are arranged into ranges/blocks is not something I think we should expose users to, unless there is good evidence that this is going to be helpful. Maybe wait until someone specifically asks for this, with a clear use case?

### Other misc features

`glyphs: all` sounds good to me, although it has a bit of a shoot-your-own-feet potential, e.g. someone specifying it for Material Design Icons.

I don't think `omit_glyphs` is a bad idea per se, but I also don't see the use case just yet? If someone is space/memory-constrained and/or have tight control of the text presented, then they can fallback in enumerating the glyphs manually rather than using glyphsets, no?

nagyrobi:
> emitting errors or even warnings is just going to be annoying for the vast majority of `glyphset` users.

Nope. It would be annoying if those warnings would appear on the devices/screens themselves, not at compile time. The user has to be able to know _before_ flashing a binary that certain glyphs are missing he might want to use, rather than running into displays showing garbage at runtime or printing errors in the console, which would ultimately result into increasing the load on the people supporting this project, by having to explain each time why is this.
A (properly documented) `omit_glyphs` could not only mitigate this, but also help with omitting the ones not used, to indeed save resources.

paravoid:
I hear where you're coming from! I do agree that (good) warnings can helpful and avoid user confusion (and reduce bug reports). My point is, however, that warnings that are emitted most of the time, or that are nonsensical, do more harm than good, either by creating more confusion, or a feeling of "alert fatigue".

But, warning users when they use basic (primarily Latin) fonts like "Arial" and "Times New Roman" with the Latin set would be nonsensical and confusing, as it would be to ask users to `omit_glyphs` rare glyphs that they didn't even know existed before.

The out-of-box experience should be e.g. "I want to support text in English and Greek in my firmware", set a couple of keys, and unless you want to do something special, to not have to open up the Unicode book and enumerate specific glyphs — or understand intricacies like composed vs. non-composed characters and NFD/NFC normalization forms for that matter (which is in some of the cases the reason of missing glyphs).

As another example, on my Debian system with a set of (largely hand-picked) fonts, the majority of the fonts that I use on my day-to-day as a Greek speaker do not have 100% coverage of `GF_Greek_Core`. The distribution is (number of glyphs vs. `GF_Greek_Core` coverage):
```
    139 100%
    200 99%
     25 98%
     11 97%
     40 96%
     35 95%
[...]
      1 4%
      4 3%
     13 2%
     49 1%
    118 0%
```

Many of these 98% or 99% fonts are fonts from large professional font foundries, with full Greek support, and are perfectly usable.

For comparison, if instead of Google we choose Adobe as our source of truth, and the [Adobe Greek 1 glyphset](https://github.com/adobe-type-tools/adobe-greek-charsets) specifically, we have 450 fonts with 100% coverage (all of the 95-100% in the Google list), and another 22 in the 90-99%. Basically even Google's and Adobe's definition of Greek differs by 5%.

Adobe even states: _"**These character sets are informative, not normative.** They are guidelines. We reserve the right to update, modify, replace or withdraw them at any time without prior notice."_ The same obviously applies to Google's definition. Emitting warnings for even a single missing glyph would be an example of treating them as normative.

clydebarrow:
> Emitting warnings for even a single missing glyph would be an example of treating them as normative.

Quite the opposite IMO. Since the character sets are not set in stone, a warning that a glyph you had been happily using is now absent from an updated glyphset would be exactly what I would want. It's not about treating anything as normative, it's about avoiding unexpected behaviour. The current status where no warnings are issued but the glyphs don't work is more likely to have come about because Pillow provides no means of verifying a glyph's status rather than a deliberate design decision - having to flash and run the code just to check if the glyphs you want are there is not an efficient workflow.

 I agree that you don't want to require users to list every unwanted glyph, which is why I suggested the availability of ranges (which would allow entire blocks to be omitted) and an overriding option to shut the warning up, allowing those who really don't care what goes in their font to just get on with it. But I don't think that should be the default.

The font component has had very little love since it was introduced, but it's going to be used a lot more in the future with LVGL now available, and a plethora of cheap high-resolution display boards on the market. It deserves to be the best it can.

At this point I think the best course of action would be for me to raise a separate PR fixing the historic issues, adding warnings, and generally cleaning up the code, then the glyphsets can be implemented in your PR as a separate step.

It has been unfair to have hijacked what on the surface seems a simple change with all the other stuff - it's only because you were proposing bolting on new structure to an unstable foundation that I went down that rabbit hole.

nagyrobi:
> My point is, however, that warnings that are emitted most of the time, or that are nonsensical, do more harm than good, either by creating more confusion, or a feeling of "alert fatigue".

Again, you're overthinking it. Fatigue would be if those alerts would appear on the displays themselves during usage, and not during compile, in a terminal window. You compile the firmware once, flash it to the device and you see no more warnings.

Other components also handle similar situations using an optional `ignore...` config variable which defaults to false. Look at `ignore_strapping_warning` for example or seach for `ignore` to see how various other components offer handling similar issues. All do it correctly not by muting the situation, but informing the user correctly on what's expected to happen, and still offering a _manual_ choice to silence some of the warnings (and/or trigger the user to change component behavior in a beneficial way).

This has to be handled consistently with the other parts of the system.

Repeated comparison with desktop operating systems and text editors has no point here, this kind of usage, as employed by ESPHome is barely legal in case of the majority of the vectorial font files anyway - if the users would respect the licensing terms, as it would turn out if they would at least read them... (responsibility is of course still the user's). Also the target userbase of ESPHome is nowhere comparable (in sizes and expertise at least) with a text/graphic editor's.

paravoid:
> I agree that you don't want to require users to list every unwanted glyph, which is why I suggested the availability of ranges (which would allow entire blocks to be omitted) and an overriding option to shut the warning up, allowing those who really don't care what goes in their font to just get on with it. But I don't think that should be the default.

It seems that I'm not persuading you, which is fine! I don't have to get convinced, but I think it would really help me and the discussion if this proposal (and similar like them) were a bit more specific, i.e. came with examples? Like, I found zero fonts where the missing glyphs sit neatly as part of a range, so I don't know how this would work. But I wrote a few scripts myself, and checked very specific fonts (the ones I found on my system), so I may have missed something.

I wonder if you've tested on how many fonts the code is erroring out and/or how this warning would look like, and how easy it would be to silence it with `omit_glyphs`? Maybe try it at least with `GF_Latin_Core` and a few fonts like the ones I mentioned, like e.g. Courier New?
 
> At this point I think the best course of action would be for me to raise a separate PR fixing the historic issues, adding warnings, and generally cleaning up the code, then the glyphsets can be implemented in your PR as a separate step.
> 
> It has been unfair to have hijacked what on the surface seems a simple change with all the other stuff - it's only because you were proposing bolting on new structure to an unstable foundation that I went down that rabbit hole.

I am highly appreciative of the fact that you fixed those issues! I also think that we have agreement that the error on missing glyphs should go, and the matter of debate was about warnings, right? So could we at least merge that?

paravoid:
> Other components also handle similar situations using an optional `ignore...` config variable which defaults to false. Look at `ignore_strapping_warning` for example or seach for `ignore` to see how various other components offer handling similar issues. All do it correctly not by muting the situation, but informing the user correctly on what's expected to happen, and still offering a _manual_ choice to silence some of the warnings (and/or trigger the user to change component behavior in a beneficial way).

Actually that is a good example of a spurious compile-time warning that annoyed users enough to the point where an `ignore` setting was made :wink: 

It wouldn't be my most preferred option, but a warning (not error) + `ignore_missing_glyphs` setting would be OK, I guess. Definitely better than `omit_glyphs` (with or without ranges).

> Repeated comparison with desktop operating systems and text editors has no point here, this kind of usage, as employed by ESPHome is barely legal in case of the majority of the vectorial font files anyway - if the users would respect the licensing terms, as it would turn out if they would at least read them... (responsibility is of course still the user's). Also the target userbase of ESPHome is nowhere comparable (in sizes and expertise at least) with a text/graphic editor's.

I think you may have misunderstood. The Debian reference was irrelevant - it was about getting a sample of .ttf files. I could have done it with Google Fonts, it was just easier to check fonts found in my filesystem. I did mention LibreOffice above though (once). And while the comparison is not apples-to-apples, I do think it's helpful to look at what others have done in more complete implementations. Unicode is hard and fonts are hard, and I don't think any of us three are experts in this domain -- I certainly am not. I do have [Yannis Haralambous' "Fonts & Encodings"](https://www.oreilly.com/library/view/fonts-encodings/9780596102425/) book sitting on my bookshelf, but at 1016 pages long I never managed to read it in full. I read enough to get a headache and to know to be careful when dealing with those matters :)

nagyrobi:
> that is a good example of a spurious compile-time warning that annoyed users enough to the point where an `ignore` setting was made 😉

Yes. And it was handled properly, not silently muted. We should be consistent like that.

> a warning (not error) + `ignore_missing_glyphs` setting would be OK, 

The warning has to list all the missing glyphs by their code. `ignore_missing_glyphs` should be `false` by default.

paravoid:
> The warning has to list all the missing glyphs by their code. `ignore_missing_glyphs` should be `false` by default.

Can you give an example of how that YAML config would look like with a few fonts of your choice?

nagyrobi:
What do you mean? The example in the op is good.

paravoid:
> What do you mean? The example in the op is good.

A few representative examples where this proposed `ignore_missing_glyphs` is used, i.e. with fonts that are missing glyphs for certain glyphsets.

nagyrobi:
Simply just
```yaml
font:
  - file: "gfonts://Roboto"
    id: roboto_greek
    size: 24
    glyphsets:
      - GF_Latin_Core
      - GF_Greek_Core
    ignore_missing_glyphs: true
```

clydebarrow:
> I am highly appreciative of the fact that you fixed those issues! I also think that we have agreement that the error on missing glyphs should go, and the matter of debate was about warnings, right? So could we at least merge that?

If a glyph is requested specifically in `glyphs` then it should still be an error, but if implicitly via `glyphsets` then just a warning, and if no `glyphs` or `glyphsets` is provided, so uses the default set, then no warning or error and just outputs the intersection of the glyphset and the font. Happy with that?

paravoid:
> > I am highly appreciative of the fact that you fixed those issues! I also think that we have agreement that the error on missing glyphs should go, and the matter of debate was about warnings, right? So could we at least merge that?
> 
> If a glyph is requested specifically in `glyphs` then it should still be an error, but if implicitly via `glyphsets` then just a warning, and if no `glyphs` or `glyphsets` is provided, so uses the default set, then no warning or error and just outputs the intersection of the glyphset and the font. Happy with that?

:laughing: I think that's a good compromise, yes. Thanks!

paravoid:
Awesome! This looks almost there, thanks so much! Perhaps nitpicking or overengineering a little bit but...

One of the issues (e.g. identified with Roboto 2.001101 from 2014) is:
```
WARNING Font Roboto@400@False@v1.ttf is missing 6 glyphs:
    ̈ (b'\\u0308')
    ʹ (b'\\u0374')
    ͵ (b'\\u0375')
    ; (b'\\u037e')
    Ϗ (b'\\u03cf')
    ϗ (b'\\u03d7')
```

This has the following problems:
* We repr() a byte, so we have this b'' indirection
* We leak Python's representations (\x, \u, \U) etc. We should emit the conventional "U+NNNN..." instead (4+ digits)
* Combining characters (suc as U+308 here, but also U+301, U+302 etc. with Latin Core) are not represented properly. The right way to do this per [Unicode (16.0, 2.11.4)](https://www.unicode.org/versions/Unicode16.0.0/UnicodeStandard-16.0.pdf): "_Nonspacing combining marks used by the Unicode Standard may be exhibited in apparent isolation by applying them to U+00A0 NO-BREAK SPACE. This convention might be employed, for example, when talking about the combining mark itself as a mark, rather than using it in its normal way in text (that is, applied as an accent to a base letter or in other combinations)._". Now, older versions of Unicode allowed (breaking) spaces, and here we _do_ have a space prepended due to the indentation; but we shouldn't rely on that.

While at it, and because we're going to need to import stdlib's `unicodedata` for (c), perhaps we should also print the name for the character? So all in all:
```python
import unicodedata

def char_for_display(char: str):
    display = f"U+{ord(char):04X}"
    name = unicodedata.name(char)
    if unicodedata.combining(char):
        # Prepend no-break space to combining characters for display per
        # Unicode 16.0.0 2.11.4 Exhibiting Nonspacing Marks in Isolation
        char = "\N{NO-BREAK SPACE}" + char
    return char + " (" + display + " " + name + ")"
```

nagyrobi:
Lvgl just prints `U+NNNN...` in the console/log when runs into a missing glyph...

jesserockz (approved):

PR #7434, opened 2024-09-11
[workflows] Use alpine base image for docker builds

clydebarrow:

Use Alpine instead of Debian for docker builds. Has potential to use wheels hosted on wheels.home-assistant.io.

Build tools used to build wheels not available to download are installed, but removed from the final image, so although this slows the build, it does not bloat the image. The images generated are substantially smaller than the current Debian based images.

Local builds for the `docker` image for all 3 architectures have been tested. The `lint` and `hassio` images have not yet been built or tested.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -181,7 +181,7 @@ RUN \
         clang15-extra-tools=15.0.7-r19 \
         nano=8.0-r0 \
         build-base=0.5-r3 \
-        python3-dev=3.12.6-r0 \
+        python3-dev=3.12.6-r0
 
 COPY requirements_test.txt /
 RUN if [ "$TARGETARCH$TARGETVARIANT" = "armv7" ]; then \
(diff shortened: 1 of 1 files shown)


ptr727:
Slightly off-topic; I had previously tried to use `python:3.12-alpine` as a base image, and then installed `esphome` using `pip`, but I get `sh: xtensa-lx106-elf-g++: not found` errors during compilation. Per google this looks like PIO pulling the wrong xtensa platform binaries, or maybe `glibc` vs. `musl`.

Have you experienced anything similar?

Update: I found that I needed to install `gcompat` and `build-base` for the `glibc` based toolchain to work on `musl` Alpine.

PR #3515, opened 2022-05-29
The NEC IR protocol uses 8bit commands

Emrvb:
# What does this implement/fix?

The current implementation for the NEC IR protocol encoding/decoding is incomplete.

The NEC IR protocol uses 8bit ~~addresses and~~ commands. The other bits are inverses for error checking.

~~This patch will gracefully accept ~~16bit addresses and~~ commands for encoding but will emit a warning if you do. The decoder now only sets the proper 8bit ~~address and~~ command.~~

~~This patch will break existing implementations which expect to receive 16bit addresses and commands.~~

This patch will break existing implementations that rely on sending or receiving "made up" commands that fail the error checking.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes N/A

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here> WIP

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
switch:
  - platform: template
    name: Transmit NEC
    turn_on_action:
      - remote_transmitter.transmit_nec:
          address: 0x1
          command: 0xC

remote_receiver:  
  pin:
    number: ?
    inverted: ?
  dump:
    - nec
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #7458, opened 2024-09-16
NEC IR protocol: add padding bytes

domdomegg:

Normally, the NEC protocol is made of 4 bytes: 2 address bytes and 2 command bytes. However, both are often really just 1 byte with a padding byte (which is the inverse of the actual byte).

This means in many product manuals and other documentation only the one byte is provided. This has lead to confusion setting up the integration (as seen via Home Assistant forum posts), and requires people read the long note in the docs.

This PR pads addresses automatically, provided they are given as just one byte. This would make setting up IR control with ESPHome easier.

This is different to a previous PR, https://github.com/esphome/esphome/pull/3515, which enforced the use of 1 byte (8 bit) codes. That PR would prevent users from entering in 2 (or even 4) byte custom NEC codes, breaking compatibility with some devices. Whereas this PR just automatically pads addresses and commands provided as just a single byte. It doesn't touch them otherwise, so longer codes still work just fine.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#WIP - plan to delete 'Additionally, ESPHome does not automatically generate parity bits or pad values to 2 bytes. So, in order to send command 0x0, you need to use 0xFF00 (0x00 being the command and 0xFF being the logical inverse).', and update the address and command parameters to say '8-bit or 16-bit'

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
button:
  - platform: template
    name: NEC short
    on_press:
      remote_transmitter.transmit_nec:
        address: 0x32
        command: 0x02
```

## Checklist:

### DIFF
--- a/esphome/components/remote_base/nec_protocol.cpp
+++ b/esphome/components/remote_base/nec_protocol.cpp
@@ -13,8 +13,17 @@ static const uint32_t BIT_ONE_LOW_US = 1690;
 static const uint32_t BIT_ZERO_LOW_US = 560;
 
 void NECProtocol::encode(RemoteTransmitData *dst, const NECData &data) {
-  ESP_LOGD(TAG, "Sending NEC: address=0x%04X, command=0x%04X command_repeats=%d", data.address, data.command,
-           data.command_repeats);
+  // If the address or command is <= 0xFF the user has only supplied one byte
+  // We need to generate parity bits to pad the value to 2 bytes
+  // The NEC protocol requires us to prepend the inverse byte, e.g. 0x21 -> 0xDE21
+  uint16_t address = data.address <= 0xFF
+    ? data.address | ((uint16_t)(~data.address) << 8)
+    : data.address;
+  uint16_t command = data.command <= 0xFF
+    ? data.command | ((uint16_t)(~data.command) << 8)
+    : data.command;
+
+  ESP_LOGD(TAG, "Sending NEC: address=0x%04X, command=0x%04X command_repeats=%d", address, command, data.command_repeats);
 
   dst->reserve(2 + 32 + 32 * data.command_repeats + 2);
   dst->set_carrier_frequency(38000);
@@ -22,7 +31,7 @@ void NECProtocol::encode(RemoteTransmitData *dst, const NECData &data) {
   dst->item(HEADER_HIGH_US, HEADER_LOW_US);
 
   for (uint16_t mask = 1; mask; mask <<= 1) {
-    if (data.address & mask) {
+    if (address & mask) {
       dst->item(BIT_HIGH_US, BIT_ONE_LOW_US);
     } else {
       dst->item(BIT_HIGH_US, BIT_ZERO_LOW_US);
@@ -31,7 +40,7 @@ void NECProtocol::encode(RemoteTransmitData *dst, const NECData &data) {
 
   for (uint16_t repeats = 0; repeats < data.command_repeats; repeats++) {
     for (uint16_t mask = 1; mask; mask <<= 1) {
-      if (data.command & mask) {
+      if (command & mask) {
         dst->item(BIT_HIGH_US, BIT_ONE_LOW_US);
       } else {
         dst->item(BIT_HIGH_US, BIT_ZERO_LOW_US);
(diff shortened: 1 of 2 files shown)


DjordjeMandic (changes requested):
Current implementation with changes to source would break extended nec protocol. Support for padding bytes should be implemented only in python (code validation/generation). No changes to source are needed. In order to prevent breaking a lot of configs multiple times, if you're interested we could implement it together in #8015 

#8015 Already introduces breaking changes, could be done once then. 

Only thing that needs to be discussed is whether to have padding enabled by default or not. Padding can be done in `async def nec_action(var, config, args)` and `def nec_binary_sensor(var, config)`. I just dont know how to implement it for `
@register_trigger("nec", NECTrigger, NECData)`

Emrvb:
Hello DjordjeMandic,

Responding here since you mentioned my pull request. 

What you call "padding" is actually a kind of checksum, it's to verify the command integrity.

I guess your proposed change still breaks existing implementations depending on this "bug" (users abusing the implied 16 bit address space), where they actually want to send the first byte zeroed.

DjordjeMandic:
> Hello DjordjeMandic,
> 
> Responding here since you mentioned my pull request. 
> 
> What you call "padding" is actually a kind of checksum, it's to verify the command integrity.

Yes, it is checksum (inverted byte). Command by protocol def always has higher byte inverted. Command config option can be 0-255 (1 byte) but that would break functionality for non standard uses. 


> I guess your proposed change still breaks existing implementations depending on this "bug" (users abusing the implied 16 bit address space), where they actually want to send the first byte zeroed.

Well I don't see how would that break it. Users can currently supply manually whole 16 bits (either extended or normal 8bit addressing). New config option (for example calculate_inverted_higher_byte) could be added to control whether higher byte is automatically calculated(if calculate_inverted_higher_byte true) for values under 256 (dec). That leaves both current and new implementation available to users

EDIT:
For example:

```yaml
# 8 bit nec protocol
on_press:
      address: 0x40
      command: 0x0B
      calculate_inverted_higher_byte: true
```

```yaml
# still 8 bit nec protocol, same as above
on_press:
      address: 0xBF40
      command: 0xF40B
      calculate_inverted_higher_byte: false
```

```yaml
# still 8 bit nec protocol, same as above
on_press:
      address: 0xBF40
      command: 0xF40B
      calculate_inverted_higher_byte: true # ignored because supplied address/command is higher than 0xFF
```


```yaml
# 16 bit addr nec protocol
on_press:
      address: 0xAA40
      command: 0xF40B
      calculate_inverted_higher_byte: false
```
```yaml
# 16 bit addr nec protocol, same as above
on_press:
      address: 0xAA40
      command: 0xF40B
      calculate_inverted_higher_byte: true # ignored because supplied address/command is higher than 0xFF
```

And if calculate_inverted_higher_byte is set to true with value supplied being over 0xFF treat is as calculate_inverted_higher_byte was set to false. It also allows non/standard commands to be sent

calculate_inverted_higher_byte defaults to false in config schema => no current configurations are broken, but allows easier input of address / command if needed.

Emrvb:
It boils down to this:

In my opinion the default behavior should be to follow the spec, an 8 bit command with calculated inverse. An option to allow "abuse" of the full address space, like you suggest, would be fine. But that change in default behavior could still break existing implementations.

Using the old behavior as a default and relying on users to properly read the documentation is only a marginal improvement.

DjordjeMandic:
> It boils down to this:
> 
> In my opinion the default behavior should be to follow the spec, an 8 bit command with calculated inverse. An option to allow "abuse" of the full address space, like you suggest, would be fine. But that change in default behavior could still break existing implementations.

You're absolutely correct, spec should be followed in yaml config. Cpp Source can be left as is. 

> Using the old behavior as a default and relying on users to properly read the documentation is only a marginal improvement.

Will talk with devs on discord about this, I'm not fan of breaking changes but when they are necessary then I like to do it in one release. I will probably open one PR on top of #8015 for automatic inverse calculation before code generation of supplied config address. 

In the end I agree with you. Schema should follow the standard by default with optional option to allow the abuse of address and commands (16 bit). 

Ideally schema should detect amount of characters in hex int. If it is more than 4 (0x00 - 0xFF is 4 chars), treat it as 16bit (both addr and command) and ignore the automatic calculation. This would allow old configurations.

Emrvb:
> You're absolutely correct, spec should be followed in yaml config. Cpp Source can be left as is.

We do not agree on this. I don't want to calculate the command inverse in my C code and your proposed change is still breaking if it breaks or changes the behavior of existing yaml configs.

PR #7463, opened 2024-09-17
Add OpenTherm component (part 2: sensors)

olegtarasov:

This is the second (and final) part of OpenTherm component. It contains the code to support all the sensors, inputs and outputs. Sensors configured in YAML are transformed to a set of C++ macros, which in turn generate the final code that handles sensors. I extracted these macros to `opentherm_macros.h` and tried to explain them in detail.

## Types of changes

- [ ] Other

Part 1: #6645 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3800

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml

esphome:

esp8266:
  board: d1_mini

opentherm:
  in_pin: 1
  out_pin: 2

number:
  - platform: opentherm
    t_set:
```

## Checklist:


sergeantd83:
Based on the OpenTherm specs, I would suggest for the max_rel_mod_level number in the `esphome/components/opentherm/schema.py`:
`step=1`
`max_value=100`

OpenTherm protocol documentation isn’t clear for the `ID15` MaxCapacityMinModLevel. The MaxCapacityMinModLevel represents 2 things:
u8_hb: The Max capacity in kW
u8_lb: The minimum "real" minimum modulation level that the boiler is able to modulate in a range 0-100%
Let’s say the u8_lb is 25. It means that the minimum "real" boiler modulation level is 25% and represents the relative modulation at 0%. The other 75% of the "real" boiler modulation represents the 0 - 100% scale of the relative modulation sensor.

olegtarasov:
@sergeantd83 I would like to ask @FreeBear-nc to weigh in on this particular sensor, as it was him who created a PR to add it. My boiler doesn't support these messages, so I can only rely on other people who can actually use and test these changes.

FreeBear-nc:
Would need to do further testing with my boiler (Viessmann 050). With winter approaching, I'll make a start.
If ID15:HB sets the upper limit of capacity, and ID15:LB sets the lower, then it would be reasonable to assume that ID14 is a percentage between these two bounds.
ID17 (rel_mod_level) would then be a percentage bound by ID15:LB and ID14. If you are thinking of computing the real instantaneous output in kW and exporting the data as a new sensor, I would advise against it. The Viessmann I have will often report max_capacity as 255 after running for an extended period. A reset of the ESP32 clears the false data.

cadavre:
Another winter season comes shortly, so I thought I'm gonna browse through GitHub repos to see if someone updated their `esphome-opentherm` repos. And what a nice surprise to see this going to be merged into official ESPHome library! **Thank you!**

I can help you with some testing and figuring out different OT specs implementations with my boiler:
https://www.termet.com.pl/en/produkt/integra-comfort/5

I faced previously a situation where one function added to support single boilers' feature broken my integration completely. I think it may be nice to have another boilers for validation.

sergeantd83:
> ID17 (rel_mod_level) would then be a percentage bound by ID15:LB and ID14. 

No, when you set the ID14 to a certain value ( e.g 60% ) then the boiler’s upper relative modulation limit will be 60%. OpenTherm documentation isn’t clear regarding modulation unfortunately.

FreeBear-nc:
> > ID17 (rel_mod_level) would then be a percentage bound by ID15:LB and ID14.
> 
> No, when you set the ID14 to a certain value ( e.g 60% ) then the boiler’s upper relative modulation limit will be 60%. OpenTherm documentation isn’t clear regarding modulation unfortunately.

You may be right on the first point, and certainly correct on the second.
Got a heat meter to connect to my boiler over the weekend - That should give me an absolute value in kW for heat generated. From there, we can check what the upper bound is for rel_mod_level.

sergeantd83:
This [link](http://otgw.tclcode.com/background.html) may clear things out.

esphome/components/opentherm/generate.py
```diff
@@ -0,0 +1,140 @@
+from collections.abc import Awaitable
+from typing import Any, Callable
+
+import esphome.codegen as cg
+from esphome.const import CONF_ID
+from . import const
+from .schema import TSchema
+
+opentherm_ns = cg.esphome_ns.namespace("esphome::opentherm")
```
jesserockz: Shouldnt need this part
```suggestion
opentherm_ns = cg.esphome_ns.namespace("opentherm")
```

olegtarasov:
After digging OpenTherm documentation some more and checking the code, I decided to apply suggestions from @sergeantd83 since they seem to be backed by OT docs. Moreover, there was an incorrect message code in `auto_min_value`, which I also fixed. Would still appreciate feedback from people who can actually test this with their boilers, but at least this sensor doesn't look broken now.

FreeBear-nc:
> After digging OpenTherm documentation some more and checking the code, I decided to apply suggestions from @sergeantd83 since they seem to be backed by OT docs. Moreover, there was an incorrect message code in `auto_min_value`, which I also fixed. Would still appreciate feedback from people who can actually test this with their boilers, but at least this sensor doesn't look broken now.

ID15:LB is the minimum absolute modulation level of the boiler. For mine, this is reported as 12%.
The message ID is completely wrong in my opinion - It is not a relative value, nor a maximum.
max_rel_mod_level will always be 100%, and if you are reporting the value of ID15:LB, it should be as min_mod_level.

olegtarasov:
After reading https://otgw.tclcode.com/background.html, I think that @FreeBear-nc you are also correct and `max_rel_mod_level` always has the range [0..100], and `ID15:LB` is more of an informational value that we can use to calculate boiler capacity when relative modulation level is set to 0. So the correct approach would be to remove `auto_min_value` from `max_rel_mod_level` altogether.

olegtarasov:
@jesserockz Just letting you know that the second part of OpenTherm integration is ready for review :)

FreeBear-nc:
> After reading https://otgw.tclcode.com/background.html, I think that @FreeBear-nc you are also correct and `max_rel_mod_level` always has the range [0..100], and `ID15:LB` is more of an informational value that we can use to calculate boiler capacity when relative modulation level is set to 0. So the correct approach would be to remove `auto_min_value` from `max_rel_mod_level` altogether.

Started to use my boiler again this year - Will be able to do some testing over the weekend. Also have a new heat meter connected so that I can test some assumptions regarding minimum output & modulation levels.

jesserockz (changes requested):
I started to review this, then got completely lost trying to understand the defines. 

I think that they are being used way too much and it makes it too hard to understand and review this PR properly.

Update: CI also does not run any python parts so code that is "hidden" inside defines cannot be linted or checked against common errors that may occur.

Can we also make this PR (or make a new PR) that is a lot smaller and only includes 1 platform type, eg `sensor` so the underlying structure there can be reviewed and decided on.

Jesse

olegtarasov:
@jesserockz I've created a new PR #7529 which contains only the `sensor` platform. I'm not sure what's the problem with clangtidy there, but you can start the review and I will figure it out.

olegtarasov:
Closing since #7529 got merged. Will prepare the final PR with rest of the sensors later today.

PR #7470, opened 2024-09-18
Omit unused functions when not needed (rtttl.cpp)

colmbuckley:
If `USE_SPEAKER` or `USE_OUTPUT` are not defined, these functions are not called and therefore should not be included.

Save space and suppress a warning.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/rtttl/rtttl.cpp
+++ b/esphome/components/rtttl/rtttl.cpp
@@ -364,6 +364,7 @@ void Rtttl::finish_() {
   ESP_LOGD(TAG, "Playback finished");
 }
 
+#if defined(USE_OUTPUT) || defined(USE_SPEAKER)
 static const LogString *state_to_string(State state) {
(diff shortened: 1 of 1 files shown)


colmbuckley:
Hmm; not sure this is the right approach actually. My issue is that, if the RTTTL component is included in an esphome config, but nothing actually *calls* the RTTTL functions, then esphome optimizes most of the functions out of the source code, but `state_to_string` remains, so we get a warning about its being defined but unused. Might be better to include this mapping logic into the `set_state_()` function, so that they are added/removed as a unit.

colmbuckley:
Let me try a different approach.

PR #7473, opened 2024-09-19
Don't compile `state_to_string()` unless debugging.

colmbuckley:
Under certain circumstances (eg: debug logging not enabled), `ESP_LOGD` will be defined to be null. Consequently, the `state_to_string` function will be defined but not used, causing unused-function warnings.

This changes it to only be compiled when the debug level is set to DEBUG or higher.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


jesserockz (changes requested):
This defeats the entire purpose of the function using LogString, and the `LOG_STR_ARG`.

The simplest solution is to wrap the ESP_LOGD line in an ifdef that checks the logging is at at least debug level.

```c++
#if ESPHOME_LOG_LEVEL >= ESPHOME_LOG_LEVEL_DEBUG
ESP_LOGD(...)
#endif
```

nielsnl68:
Most of the audio components has a similar construction to show the state as string.
Would you mind checking those as well?

colmbuckley:
> This defeats the entire purpose of the function using LogString, and the `LOG_STR_ARG`.

Thanks Jesse - I admit I wasn't sure what the purpose of the `LOG_STR_*` macros were; I read log.h a bit more closely and I thought we could use `LOG_STR_LITERAL()` here but that fails tests. Is there a doc anywhere describing what the various `LOG_STR_*` macros do?

Unfortunately there's not a lot of consistency across the codebase regarding how states are exposed to debugging macros. My quick thoughts on best practice would be to have a class member function which returns either `const char *` or `std::string` with a human-readable state; and to cast or copy that as needed in the logging functions.

> The simplest solution is to wrap the ESP_LOGD line in an ifdef that checks the logging is at at least debug level.

It's the `state_to_string()` function which is orphaned; it's already not called when we're at log levels < debug. We could of course comment *it* out when we're not in debugging, but I think it's possible that there will be other use cases for the `state_to_string_()` function so I'd keep it as a protected class member.

If we're concerned about the hundred bytes or so taken by the strings when we're not in debug mode, I'd put an `#if` as you suggest *inside* that function and just return an empty string when we're not debugging.

colmbuckley:
Pending any longer-term discussion about "the right thing"; I've gone with the simplest solution, which is to `#if` out the `state_to_string()` function when we're not at debug-or-higher. This fixes the warnings.

jesserockz (approved):

PR #7477, opened 2024-09-21
Glitch filter to filter out RF false positives

ScramblerUSA:
Glitch filter for binary sensor

In noisy RF environments it is sometimes possible to get a false trigger of binary_sensor attached to remote_receiver, which ultimately leads to undesired outcome (a light switches on unintentionally, etc.) This filter is designed to remove such glitches and only pass true signals through. It does so by ignoring a few first state transitions and passes the rest through. Since transmitters send the code repeatedly, it works wonders.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4265

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
    binary_sensor:
      - platform: remote-receiver
        id: remote_button
        # ...
        filters:
          - glitch:
              count: 4
              duration: 200ms
          - settle: 500ms
```

## Checklist:


swoboda1337:
I actually thought about adding one for sensors too. Would just provide a delta value. For example temp values will wont change by 50 degrees on one reading.  I was going to call it glitch too.

PR #7478, opened 2024-09-21
[core][esp32_rmt_led_strip] Migrate ExternalRAMAllocator to RAMAllocator

PxPert:

This PR include these Changes:
- **New Feature**: Added a FORCE_INTERNAL Flag to the ExternalRAMAllocator Class. With this flag it's possible to the use internal malloc call for components using ExternalRAMAllocator even if PSRAM is enabled, in runtime. This could be useful to fix weird problems with some components/devices or to keep PSRAM free for other components wich do need as much memory possible.
- **New Feature/Bugfix**: Added optional use_psram parameter to esp32_rmt_led_strip component (Depends on the FORCE_INTERNAL Flag Feature). With this flag the user can fix weird behaviours with its led strip in case PSRam is enabled, by forcing the allocation into the internal memory. The default setting is **true** so this don't break existing configurations.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4266

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
  - platform: esp32_rmt_led_strip
    rgb_order: GRB
    chipset: ws2812
    rmt_channel: 1
    use_psram: false
```

## Checklist:


clydebarrow:
Why not just change `esp32_rmt_led_strip` to use `malloc()` when desired rather than complicate things by adding a flag to `ExternalRAMAllocator` to tell it not do to what it is meant to do?

PxPert:
Yes, I was initally tempted to do it, but:
- This flag could apply to every component relying on ExternalRAMAllocator
- Putting the logic of managing selective use of ExternalRAMAllocator/malloc inside the component itself would be more complex as it's not a simple change in a call (ExternalRAMAllocator is a class which needs to be initialized and has its own methods), and this would need to be repeated everytime  ExternalRAMAllocator is used in the component. if another component should need this in future, the logic has to be repeated again and again (and most probably in a different way from one component to the other). All of this would translate in more code, so less readibility and more probability of bugs.

Implementing the magic inside the ExternalRAMAllocator itself makes all of this transparent to the component developer, which should only add the flag when needed. After all this logic is already present because it would work the same way as if you don't have an ESP32 or PSRAM component enabled

guillempages:
I'm with @clydebarrow here... The allocator is called **External**RAMAllocator 🤔  So if the component uses this, it is trying to use external RAM if available. If the component does not want to use, it can really just use `malloc`/`new` instead.

It *might* be an idea to just create a generic "ESPHomeAllocator" or similar (or rename the `ExternalRAMAllocator`, that uses PSRAM when available, or internal RAM otherwise or when explicitly told it to, to make memory allocation more generic/consistent between components, but this might also be overkill, and most components would probably just ignore that and keep using `malloc` when they really need internal memory.

PxPert:
I agree with the fact that creating a new Class or rewriting its name would be overkill and would need to refactor the code depending on it.
The idea of the additional Flag is to keep the current behaviour of the ExternalAllocator Class with a minimal change consisting on adding another "fallback" mechanism to the already existant ones (The dependency on ESP32 and on the PSRAM component), with the difference that this can be manually set by the user.
I know this doesn't totally stand with the logic (and the name) of the ExternalAllocator, but this is just a borderline case, where the component **could** use external allocator, but the user wants to force malloc usage (For example due to possible incompatibilities with some devices) and so chooses to disable it via yaml. This could potentially happen with every other components with optional dependency on PSRAM, so moving this logic inside the component itself would be way more complex and difficult to maintain

jesserockz:
IMHO I am for renaming it to remove the `External` from the name. We can add an interim `using` so we dont have to update every reference at once and can go through them one PR at a time.
We have a lint rule that fails when `malloc` is detected in the code and this then forces more `NOLINT` comments. I would prefer that all of the different components can rely on a single dynamic implementation of memory allocation.

PxPert:
Maybe that's the best solution, and now that we have this flag it's possible to use it to remove all the nolinted malloc calls.
If needed I can change the PR to rename the class and alias it with the old name.
Just one suggestion:
I know it's a bit hacky, but could we use a macro instead of a using, to just warn the user/developer the old name is deprecated and going to be removed?
Something like this:
```
#define ExternalRAMAllocator _Pragma( "GCC warning \"ExternalRAMAllocator name is deprecated. Please rename it to RAMAllocator\"" ); RAMAllocator
 ```

**Comment by clydebarrow** [comment]
With the rename, I'd suggest flags like `ALLOC_INTERNAL` and `ALLOC_EXTERNAL` rather than anything called "FORCE". The default, if neither is set (or both are set) should be the same as the current implementation.Otherwise with only one flag set it should fail if memory from the requested space is not available.

**Comment by guillempages** [comment]
I'm not sure about the warning. It is not something the user can change/ fix,  it is something for the developers/ maintainer.
The warning might confuse some users and make them create issues or search support for something that is not really an issue with the configuration.

**Comment by jesserockz** [comment]
@PxPert 
https://github.com/esphome/esphome/blob/21fbbc5fb9477704be8b1ec8293d2a729f7a0072/esphome/components/light/addressable_light.h#L20

**Comment by PxPert** [comment]
Oh I didn't know of ESPDEPRECATED macro, my bad. I'll use it then.
Talking about the flags: 
@clydebarrow  You're right. Now that we're going to rename the class it's better to also rename FORCE_INTERNAL to ALLOC_INTERNAL.
Could we also replace REFUSE_INTERNAL with ALLOC_EXTERNAL? I don't see any code using it right now, so maybe we could just rename it without keeping both of them.
Let me know so I'll proceed with the changes

**Comment by clydebarrow** [comment]
> Could we also replace REFUSE_INTERNAL with ALLOC_EXTERNAL?

Yes. If it's currently unused then that should be easy.

File: esphome/core/helpers.h [hunk 1]
```diff
@@ -688,9 +695,13 @@ template<class T> class ExternalRAMAllocator {
   }
 
  private:
-  Flags flags_{Flags::ALLOW_FAILURE};
+  uint8_t flags_{Flags::ALLOW_FAILURE};
 };
 
+template<class T>
+using ExternalRAMAllocator ESPDEPRECATED("RAMAllocator is deprecated, use RAMAllocator instead.",
+                                         "2024.10.0") = RAMAllocator<T>;
```
jesserockz: Lets actually not add the ESPDEPRECATED right away as I can see the issue tracker getting bombarded with this.

PxPert: No problem, I've just removed the ESPDEPRECATED Flag ;)

esphome/core/helpers.h
```diff
@@ -651,32 +651,39 @@ void delay_microseconds_safe(uint32_t us);
...
-/** An STL allocator that uses SPI RAM.
+/** An STL allocator that uses SPI or internal RAM.
  *
- * By setting flags, it can be configured to don't try main memory if SPI RAM is full or unavailable, and to return
- * `nulllptr` instead of aborting when no memory is available.
+ * By setting flags, it can be configured to:
+ * - perform external allocation falling back to main memory if SPI RAM is full or unavailable
+ * - perform external allocation only
+ * - perform internal allocation only
+ * - return `nulllptr` instead of aborting when no memory is available.
```
guillempages:
```suggestion
 * - return `nullptr` instead of aborting when no memory is available.
```

esphome/core/helpers.h
```diff
@@ -651,32 +651,39 @@ void delay_microseconds_safe(uint32_t us);
...
 
   enum Flags {
-    NONE = 0,
-    REFUSE_INTERNAL = 1 << 0,  ///< Refuse falling back to internal memory when external RAM is full or unavailable.
-    ALLOW_FAILURE = 1 << 1,    ///< Don't abort when memory allocation fails.
+    NONE =
+        0,  ///< Perform external allocation and fall back to internal memory when external RAM is full or unavailable.
+    ALLOC_EXTERNAL = 1 << 0,  ///< Perform external allocation only.
+    ALLOC_INTERNAL = 1 << 1,  ///< Perform internal allocation only.
+    ALLOW_FAILURE = 1 << 2,   ///< Don't abort when memory allocation fails.
```
jesserockz: I think we should never abort, the code requesting memory should always handle a failed allocation

jesserockz (dismissed):

esphome/core/helpers.h
PxPert: The abort was kept to mantain backwards compatibility. Anyway we could just proceed in more steps, like these:
- Push this PR keeping the actual functionality
- Remove the abort and just return nullptr keeping the ALLOW_FAILURE flag just for compatibility
- Remove the ALLOW_FAILURE and class alias after all the dependant components are refactored

What do you think about it?

clydebarrow: I can only find one place in ESPHome where an allocator is created *without* the ALLOW_FAILURE flag (font component) and it will abort anyway if the allocation fails.

So I would suggest keeping the flag definition for now, for backwards compatibility, but ignoring it and always return the failure to the calling function.

PxPert: Ok then, I've just removed the abort condition ;)

clydebarrow (changes requested):
The logic for handling the flags is a little backwards I think:

```
    if ((this->flags_ & Flags::ALLOC_INTERNAL) == 0) {
      ptr = static_cast<T *>(heap_caps_malloc(size, MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT));
    }
```

Here the internal flag is being tested to decide whether to allocate externally.

I'd suggest as follows:

* In the constructor, if neither INTERNAL nor EXTERNAL is set, set them both.
* Test the EXTERNAL flag to attempt external allocation
* Test the INTERNAL flag to attempt internal allocation.

That will preserve the current behaviour, but make setting the flags behave as expected.

esphome/core/helpers.h
```diff
@@ -651,35 +651,40 @@ void delay_microseconds_safe(uint32_t us);
...
 
   enum Flags {
-    NONE = 0,
-    REFUSE_INTERNAL = 1 << 0,  ///< Refuse falling back to internal memory when external RAM is full or unavailable.
-    ALLOW_FAILURE = 1 << 1,    ///< Don't abort when memory allocation fails.
+    NONE =
+        0,  ///< Perform external allocation and fall back to internal memory when external RAM is full or unavailable.
+    ALLOC_EXTERNAL = 1 << 0,  ///< Perform external allocation only.
+    ALLOC_INTERNAL = 1 << 1,  ///< Perform internal allocation only.
+    ALLOW_FAILURE = 1 << 2,   ///< Does nothing. Kept for compatibility.
```
clydebarrow: Nitpicking: long comment breaks the line; comment indicators are unusual.

```suggestion
    NONE = 0,  // Perform external allocation and fall back to internal memory 
    ALLOC_EXTERNAL = 1 << 0,  // Perform external allocation only.
    ALLOC_INTERNAL = 1 << 1,  // Perform internal allocation only.
    ALLOW_FAILURE = 1 << 2,   // Does nothing. Kept for compatibility.
```

PxPert: Yup, this seemed strange for me, too, but I've kept the same format as the original comments thinking it was a sort of internal standard :)

clydebarrow: I have seen similar formatting in Espressif code, but it's not AFAIK commonly used in ESPHome and I don't personally seen any benefit in it.

esphome/components/esp32_rmt_led_strip/led_strip.cpp
```diff
@@ -22,7 +22,8 @@ void ESP32RMTLEDStripLightOutput::setup() {
 
   size_t buffer_size = this->get_buffer_size_();
 
-  ExternalRAMAllocator<uint8_t> allocator(ExternalRAMAllocator<uint8_t>::ALLOW_FAILURE);
+  RAMAllocator<uint8_t> allocator(RAMAllocator<uint8_t>::ALLOW_FAILURE |
```
clydebarrow: The ALLOW_FAILURE flag can be removed now that it's non-functional.

clydebarrow (changes requested):

esphome/components/esp32_rmt_led_strip/led_strip.cpp
PxPert: You're right, removing right now at least from the led_strip component

clydebarrow (changes requested):
@PxPert Did you see my suggestion here? https://github.com/esphome/esphome/pull/7478#pullrequestreview-2355639891

As it stands if an allocator is constructed with both ALLOC_INTERNAL and ALLOC_EXTERNAL set it will always fail, which is counter-intuitive.

PxPert:
> @PxPert Did you see my suggestion here? [#7478 (review)](https://github.com/esphome/esphome/pull/7478#pullrequestreview-2355639891)
> 
> As it stands if an allocator is constructed with both ALLOC_INTERNAL and ALLOC_EXTERNAL set it will always fail, which is counter-intuitive.

I'm sorry I completed missed your comment. Anyway you're right, I'll make the condition inclusive and not exclusive, with the default (no flag set) to prefer external allocation in case it's an ESP32. Do you agree with that?

PxPert:
Changes just pushed (and CI Testing). This is the allocator behaviour now:
- No flags (or to be more precise both ALLOC_EXTERNAL and  ALLOC_INTERNAL not set): Prefer External allocation / Fallback to Internal
- ALLOC_EXTERNAL: External allocation / nullptr if allocation fails
- ALLOC_INTERNAL: Internal allocation / nullptr if allocation fails
- ALLOC_INTERNAL | ALLOC_EXTERNAL: Prefer External allocation / Fallback to Internal

If device != ESP32: Completely ignore ALLOC_ Flags and perform internal allocation. nullptr if allocation fails

clydebarrow (approved):

PxPert:
Thanks for merging my PR, I really appreciate the effort you all put into reviewing it ;)

PR #7479, opened 2024-09-21
Allow starting and stopping intervals

guillempages:

Add actions to start/stop a specific interval, so that its actions are not executed while stopped.
This can be used e.g. to pause a running animation, to stop automatic page changes in displays, ...

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4268
## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
interval:
  - interval: 1s
    id: toggleable_interval
    autostart: false
    then:
      - logger.log: Tick

switch:
  - platform: template
    id: pause_interval
    optimistic: true
    turn_on_action:
      - interval.start: toggleable_interval
    turn_off_action:
      - interval.stop: toggleable_interval
```

## Checklist:

### DIFF
--- a/esphome/components/interval/__init__.py
+++ b/esphome/components/interval/__init__.py
@@ -1,25 +1,59 @@
+from esphome import automation
 import esphome.codegen as cg
 import esphome.config_validation as cv
-from esphome import automation
 from esphome.const import CONF_ID, CONF_INTERVAL, CONF_STARTUP_DELAY
 
 CODEOWNERS = ["@esphome/core"]
+CONF_AUTOSTART = "autostart"
+CONF_START_CONDITIONS = "start_conditions"
+
 interval_ns = cg.esphome_ns.namespace("interval")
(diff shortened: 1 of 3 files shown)


jesserockz:
I have not verified, but I think you already can do this with [`component.resume`](https://esphome.io/automations/actions#component-resume-action) and [`component.suspend`](https://esphome.io/automations/actions#component-suspend-action) as it is a `PollingComponent`

guillempages:
Oh. My bad. Sorry about that. I've tested it and it works.

The existing solution is probably even better; I assume tha the `update` method will thus not even be called, instead of just skipped with my solution

PR #7490, opened 2024-09-24
[sx127x] Add sx127x component

### LINKED ISSUE #2634: LORA support in the ESPHome

I have already the home automation setup with help of ESPHome and Home Assistant all using the WiFi. But I want extend the coverage outside the Home area where LORA will be helpful. Please add the Lora into ESPHome

   I have purchased the SX1262 SPI based Lora Module and tried both the Custom SPI Device(with RadioLib library) and External Components (https://github.com/PaulSchulz/esphome-lora-sx126x), but it is failed. and also difficult for me to port

**Additional context**

### LINKED ISSUE #2886: Davis vantage pro weather station info to esphome-homeassistant

Davis vantage pro weather station info to esphome-homeassistant

could be something like that
https://github.com/dcbo/ISS-MQTT-Gateway

**Additional context**

swoboda1337:

The SX127X component allows you to configure/control the SX1276, SX1277, SX1278 and SX1279 transceivers in ESPHome. Supported frequencies range from 137 MHz to 1020 MHz. Supported modulations include OOK, FSK, GFSK, MSK and GMSK. Support for both receive and transmit. These chips contain two modems. The LoRa modem and FSK/OOK modem are both supported.

Datasheet: https://www.semtech.com/products/wireless-rf/lora-connect/sx1278#documentation

There are many devices in the market with an ESP32 already packaged with these radios, examples: https://www.lilygo.cc/en-ca/products/lora3, https://heltec.org/project/wifi-lora-32v2/ and https://shop.m5stack.com/products/m5stack-lora-module-433mhz-v1-1

There are many people using my component. I currently use it for my 433 MHz AcuRite weather sensors (https://github.com/swoboda1337/acurite-esphome). It would be nice to get these radios officially supported.

Docs can be found here: https://deploy-preview-4278--esphome.netlify.app/components/sx127x
SX126X PR: https://github.com/esphome/esphome/pull/8516

## Types of changes

- [ ] Other

- resolves https://github.com/esphome/feature-requests/issues/2886
- resolves https://github.com/esphome/feature-requests/issues/2634

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4278

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Docs can be found here: https://deploy-preview-4278--esphome.netlify.app/components/sx127x

```yaml
external_components:
  - source: github://pr#7490
    components: [ sx127x ]
    refresh: 5min
    
sx127x:
  nss_pin: GPIO18
  rst_pin: GPIO23
  frequency: 433920000
  modulation: OOK
  rx_start: true
  rx_bandwidth: 50_0kHz
  rx_floor: -90

remote_receiver:
  pin: GPIO32
  dump: raw
```

## Checklist:


lboue:
Build is OK for M5stack Cores3 and [Module-LoRa868_V1.1](https://docs.m5stack.com/en/module/Module-LoRa868_V1.1).
```
[19:20:18][C][spi:064]: SPI bus:
[19:20:18][C][spi:065]:   CLK Pin: GPIO36
[19:20:18][C][spi:066]:   SDI Pin: GPIO35
[19:20:18][C][spi:067]:   SDO Pin: GPIO37
[19:20:18][C][spi:072]:   Using HW SPI: SPI
[19:20:18][C][sx127x:128]: SX127x:
[19:20:18][C][sx127x:129]:   NSS Pin: GPIO6
[19:20:18][C][sx127x:130]:   RST Pin: GPIO5
[19:20:18][C][sx127x:131]:   PA Pin: BOOST
[19:20:18][C][sx127x:132]:   PA Power: 17 dBm
[19:20:18][C][sx127x:133]:   Frequency: 869.524963 MHz
[19:20:18][C][sx127x:134]:   Modulation: OOK
[19:20:18][C][sx127x:135]:   Rx Bandwidth: 50.0 kHz
[19:20:18][C][sx127x:136]:   Rx Start: true
[19:20:18][C][sx127x:137]:   Rx Floor: -90.0 dBm
[19:20:18][C][sx127x:138]:   FSK Fdev: 5000 Hz
[19:20:18][C][sx127x:139]:   FSK Ramp: 40 us
[19:20:18][C][sx127x:147]:   FSK Shaping: NONE
```

swoboda1337:
Converted to draft temporarily to do more fsk testing,

swoboda1337:
M5 stack doesn't have dio2 connected. Will add support for packet mode as well.

herostrat:
No review, just a thank you for writing this component.
It works for me (heltec lora 32 v2)!

swoboda1337:
Packet mode is done, works great with fsk there is an on_packet trigger. M5stack can be used now as it only supports packet mode. Need to add a few automations and it's done.

swoboda1337:
Everything is done, will be updating the docs this week.

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,327 @@
...
+  uint32_t fdev = std::min(this->fsk_fdev_ / 61, (uint32_t) 0x3FFF);
+  this->write_register_(REG_FDEV_MSB, (uint8_t) ((fdev >> 8) & 0xFF));
+  this->write_register_(REG_FDEV_LSB, (uint8_t) ((fdev >> 0) & 0xFF));
+
+  // set the channel bw
+  this->write_register_(REG_RX_BW, this->rx_bandwidth_);
+
+  // set bitrate
+  if (this->bitrate_ > 0) {
+    uint64_t bitrate = (32000000u + this->bitrate_ / 2) / this->bitrate_;
```
paravoid: Perhaps define this hardcoded constant (32000000u), as it's also used elsewhere. I believe this is FXOSC, right? Also perhaps explain why this calculation is built this way?

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,327 @@
...
+    this->write_register_(REG_BITRATE_MSB, (uint8_t) ((bitrate >> 8) & 0xFF));
+    this->write_register_(REG_BITRATE_LSB, (uint8_t) ((bitrate >> 0) & 0xFF));
+    bit_sync = BIT_SYNC_ON;
```
paravoid: Should the bit synchronizer be unilaterally on? AIUI the data sheet recommends it, but does not mandate it. If not configurable through YAML, perhaps allow a way for other components to disable it, if necessary?

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,327 @@
...
+    this->mark_failed();
+    return;
+  }
+
+  // set modulation and make sure transceiver is in sleep mode
+  this->write_register_(REG_OP_MODE, this->modulation_ | MODE_SLEEP);
+  delay(1);
+
+  // set freq
+  uint64_t frf = ((uint64_t) this->frequency_ << 19) / 32000000;
```
paravoid: This is dividing by FSTEP = FXOSC / 2**19, correct? Perhaps use define and/or comments to avoid the magic numbers?

esphome/components/sx127x/sx127x.h
```diff
@@ -0,0 +1,286 @@
...
+  OOK_THRESH_DEC_1_2 = 0x20,
+  OOK_THRESH_DEC_1 = 0x00
+};
+
+enum SX127xPacketConfig2 : uint8_t {
+  CONTINUOUS_MODE = 0x00,
+  PACKET_MODE = 0x40,
+};
+
+enum SX127xRxBw : uint8_t {
```
paravoid: Where did you get these values? From the data sheet I could only find the table of RxBwMant/RxBwExp/RxBw. I saw that you're calculating `rx_bw_mant` and `rx_bw_exp` from these values, but I couldn't figure out what the source information is? Are these precalculated?

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+        cv.Optional(CONF_PREAMBLE_ERRORS, default=0): cv.int_range(min=0, max=31),
+        cv.Optional(CONF_RX_FLOOR, default=-94): cv.float_range(min=-128, max=-1),
+        cv.Optional(CONF_RX_START, default=True): cv.boolean,
+        cv.Optional(CONF_RX_BANDWIDTH, default="50_0kHz"): cv.enum(RX_BW),
+        cv.Optional(CONF_RX_DURATION, default="0ms"): cv.All(
+            cv.positive_time_period_microseconds,
+            cv.Range(max=TimePeriod(microseconds=1000000000)),
+        ),
+        cv.Optional(CONF_PA_PIN, default="BOOST"): cv.enum(PA_PIN),
+        cv.Optional(CONF_PA_POWER, default=17): cv.int_range(min=0, max=17),
```
paravoid: Any plans to also support setting PaDac = "the +20dBm option on PA_BOOST pin"?

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+        cv.GenerateID(): cv.declare_id(SX127x),
+        cv.Optional(CONF_DIO0_PIN): pins.internal_gpio_input_pin_schema,
+        cv.Optional(CONF_DIO2_PIN): pins.internal_gpio_input_pin_schema,
+        cv.Required(CONF_RST_PIN): pins.internal_gpio_output_pin_schema,
+        cv.Required(CONF_NSS_PIN): pins.internal_gpio_output_pin_schema,
+        cv.Required(CONF_FREQUENCY): cv.int_range(min=137000000, max=1020000000),
+        cv.Required(CONF_MODULATION): cv.enum(MOD),
+        cv.Optional(CONF_SHAPING, default="NONE"): cv.enum(SHAPING),
+        cv.Optional(CONF_BITRATE, default=0): cv.int_range(min=0, max=300000),
+        cv.Optional(CONF_FSK_FDEV, default=5000): cv.int_range(min=0, max=100000),
```
paravoid: I think this, as well as CONF_FSK_RAMP, should not be configured on OOK mode, correct? If so, perhaps a final_validate is needed to raise an invalid config exception if these are defined but mode is set to OOK?

paravoid (commented):
Disclaimer: I know very little about the SX127s, so consider this a mostly uninformed review.

Broadly speaking, there are a lot of magic numbers in the code (some I've left comments at, but there are plenty more) and I had to go back and forth to the datasheet to unpack the code. So for clueless folks like me :grin:, perahps it would be useful to add names to many (most?) of them, especially when the datasheet already has names defined for them (e.g. FXOSC).

Hope this review helps!

esphome/components/sx127x/sx127x.cpp
swoboda1337: It is configurable in the YAML, if the bitrate is 0 the bit synchronizer if off. Definitely need the ability to turn it off. You may not know what the bitrate of your rf device is before you listen to it. So need the ability to do raw unsynchronized data. I have run it with the bit synchronizer off for months and it is fine. If you had two signals with different bit rates you may also want it off so you can receive both without getting a second rf device.

swoboda1337: I am working on finishing the docs now

esphome/components/sx127x/sx127x.h
swoboda1337: Just taking the values from the table and using register description putting them together. I can add a comment.

esphome/components/sx127x/__init__.py
swoboda1337: Honestly it wasn't clear from the register description how to do that. I can look again but yeah it was confusing.

esphome/components/sx127x/__init__.py
swoboda1337: Yes I could add that. It doesn't break OOK mode, it does set the register it just doesn't have any effect on the signal.  So its not actually invalid. If you did want to switch between fsk and ook it might be convenient to set these ahead of time? Not sure might leave this.

esphome/components/sx127x/__init__.py
swoboda1337: image

In boost mode Pout is 17 - (15 - OutputPower) and OutputPower is only 4 bits... so the max I can set is 17. I know the docs say 20 db but I have no idea how to do that.

esphome/components/sx127x/sx127x.cpp
paravoid: Sorry I think I wasn't clear! I meant that right now as the code is structured, it's not possible to set a bitrate _and_ turn off the bit synchronizer. Shouldn't that be possible?

esphome/components/sx127x/sx127x.h
paravoid: So why is e.g. RX_BW_2_6 = 0x17 and  RX_BW_25_0 = 0x0C? I didn't find that in the table, but I may have missed it!

esphome/components/sx127x/__init__.py
paravoid: I believe this is a different setting:

image

So...
```
// REG_PA_DAC = 0x4d
// PA_DAC_ENABLED  = 0x07
// PA_DAC_DISABLED = 0x04
this->write_register_(REG_PA_DAC, PA_DAC_ENABLED);
```
...but only when output power is 17?

(untested)

swoboda1337: Yeah I didnt spend a lot of time on that. I can add support for it.

esphome/components/sx127x/sx127x.cpp
swoboda1337: I don't think that setting would do anything with it off. Can set it sure but I dont think it would do anything.

swoboda1337: I will double check this but dont think the bit rate setting matters if it is off.

esphome/components/sx127x/sx127x.h
swoboda1337: Table 40 on page 88 has all the possible bw mant and exp settings. For 25 the exp would be 4 and the mant would be 01/20. For 2.6 the exp would be 7 and the mant would be 10/24. Since bw is FXOSC / (mant * 2^(exp +2)) it works out.

esphome/components/sx127x/sx127x.cpp
swoboda1337: I guess I can add this just in case something is affected by the bitrate without the synchronizer on.

esphome/components/sx127x/__init__.py
swoboda1337: Ok I read a bit more about this and I don't think its worth it. Has a lot of restrictions including duty cycle, could damage the board. Have to adjust the over current protection too.

swoboda1337:
Docs are done and I made some updates after review comments. Should be ready to go

esphome/components/sx127x/__init__.py
```diff
@@ -134,6 +135,12 @@ def validate_config(config):
         raise cv.Invalid("PA power must be <= 15 dbm when using the RFO pin")
     if config[CONF_PA_PIN] == "BOOST" and config[CONF_PA_POWER] < 2:
         raise cv.Invalid("PA power must be >= 2 dbm when using the BOOST pin")
+    if CONF_BITRATE in config and CONF_BITSYNC not in config:
+        raise cv.Invalid(
+            "Bitrate is configured but not bitsync; add 'bitsync: true' for original functionality"
+        )
+    if CONF_BITSYNC in config and CONF_BITRATE not in config:
```
paravoid: This fails with:
```
# no bitrate
bitsync: false
```

Doesn't this need to check for `config[CONF_BITSYNC] is True`? Like `config.get(CONF_BITSYNC, False) is True and CONF_BITRATE not in config`?

swoboda1337: Thanks yeah don't know why I missed this

swoboda1337:
Moved to draft until new RMT drivers are integrated.

swoboda1337:
Cleaned up docs. Removed hacks that were required with the older RMT drivers. Removed rx_duration debug mode as it was hacky and too complicated. PR is ready to go.

esnet146:
Tested on board heltec_wifi_lora_32_V2. 
Remote_receiver and remote_transmitter works great!
DIO2 Connected in GPIO34 (INPUT ONLY)
To operate the transmitter, you can connect GPIO34 and GPIO13.
We are impatiently waiting for the PR to be received!
```
spi:
  clk_pin: GPIO5
  mosi_pin: GPIO27
  miso_pin: GPIO19
sx127x:
  dio0_pin: GPIO26
  nss_pin: GPIO18
  rst_pin: GPIO14
  frequency: 433920000
  pa_pin: BOOST
```

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,261 @@
...
+CONF_BITRATE = "bitrate"
+CONF_BITSYNC = "bitsync"
+CONF_PAYLOAD_LENGTH = "payload_length"
+CONF_PREAMBLE_SIZE = "preamble_size"
+CONF_PREAMBLE_POLARITY = "preamble_polarity"
+CONF_PREAMBLE_ERRORS = "preamble_errors"
+CONF_SYNC_VALUE = "sync_value"
+CONF_FSK_FDEV = "fsk_fdev"
+CONF_FSK_RAMP = "fsk_ramp"
+CONF_ON_PACKET = "on_packet"
```
kbx81: Please alphabetize lists of constants/variables like this. 🙂

tests/components/sx127x/test.esp32-ard.yaml
kbx81, line 1: Please consolidate the tests into a `common.yaml` file and use substitutions for the pin numbers (as is done in most other components).

kbx81 (dismissed):
Couple thoughts

swoboda1337:
More clean up, please use the standard CS_PIN instead of NSS_PIN

swoboda1337:
Code clean up complete and review comments addressed. Ready to go.

swoboda1337:
Code clean up complete and review comments addressed. Should be ready to go.

swoboda1337:
Sorry, now its ready, found a few more things to clean up

swoboda1337:
Stress testing revealed a few corner cases in packet mode when switching from rx to tx. 

Resolve by switching directly from rx to tx (the only way to clear the fifo) and increasing fs delays. 

Stress testing is complete and things look good.

swoboda1337:
Ok LoRa support has been added after a few people requested it. There were a few breaking changes but they should be obvious.

esnet146:
Please add packet_transport platform lora for this component.
https://github.com/esphome/esphome/pull/8214

Checked the operation of the module sx1278 ( XL1278-SMT)  on the board KinCony KC868-A6 
```
sx127x:
  dio0_pin:
    number: GPIO02
  cs_pin:
    number: GPIO05
  rst_pin: GPIO21
  pa_pin: BOOST
  pa_power: 17
  bandwidth: 250_0kHz
  crc_enable: true
  frequency: 433920000
  modulation: LORA
  rx_start: true
  sync_value: 0x33 
  spreading_factor: 7 
  coding_rate: CR_4_5 
  on_packet:
    then:
  ```
In packet transmission mode, two-way communication to standard antennas within the city limits is 400 meters, outside the city up to 1 km

**Comment by swoboda1337** [comment]
> Please add packet_transport platform lora for this component. #8214
> 
> In packet transmission mode, two-way communication to standard antennas within the city limits is 400 meters, outside the city up to 1 km

You can increase the range if you change the spreading_factor and code_rate. SF12 and CR_4_8 would be give a much larger range. SF12 takes a very long time on the air though SF10 might be better. 

That PR is not merged yet. I will look at integrating it once that happens.

**Comment by mortenx** [comment]
What could be the distance with the best settings?

**Comment by swoboda1337** [comment]
Depends on the environment. If its line of sight with a good antenna they claim 5km, some people say even 10km but would need a perfect environment like one side elevated on hill or something. 

To get the max range I would also have to add support for 20dbm tx, probably should do that.

**Comment by esnet146** [comment]
> SF12 and CR_4_8

~30% ! in city

**Comment by swoboda1337** [comment]
> > SF12 and CR_4_8
> 
> ~30% ! in city

30% of what?

**Comment by esnet146** [comment]
> 30% of what?

range. 400 -> 600 metеrs

**Comment by dala318** [comment]
> Please add packet_transport platform lora for this component. 

@esnet146 and @swoboda1337 
I've made an attempt in appending this base with packet_transport in https://github.com/dala318/esphome/pull/3 and it seem to be possible without much extras. The later addition of `register_listener` was a good help!

In my config I have set the sx1276 in LoRa mode, which don't have a pre-defined payload length. I suppose that would be a requirement if using packet_transport. (have not tried how the other FSK/OOK modes behave)

Also, as I commented in https://github.com/esphome/feature-requests/issues/2634#issuecomment-2761064325 it would be a nice addition if both sx126x and sx127x could be wrapped by a `lora` component. If so (and if LoRa mode is required) maybe this lora component is the one that should get packet_transport added?

**Comment by swoboda1337** [comment]
> Also, as I commented in https://github.com/esphome/feature-requests/issues/2634#issuecomment-2761064325 it would be a nice addition if both sx126x and sx127x could be wrapped by a lora component. If so (and if LoRa mode is required) maybe this lora component is the one that should get packet_transport added?

Yeah eventually I think a higher layer LoRa component would be a good idea especially if we ever want to support LoRaWAN. Hopefully the packet transport stuff gets merged soon.

**Comment by wildekek** [comment]
This is an amazing component, works great on the LilyGo LoRa32 V1.0. It would be amazing to have this merged to dev. This could be what LoRa needs to get broad adoption in the HA community, especially when people build LoraWan and Meshtastic support on top. This PR deserves more attention from the core devs.

File: esphome/components/sx127x/__init__.py [hunk 9]
```diff
@@ -0,0 +1,332 @@
...
+            cv.Optional(CONF_RX_START, default=True): cv.boolean,
+            cv.Optional(CONF_SHAPING, default="NONE"): cv.enum(SHAPING),
+            cv.Optional(CONF_SPREADING_FACTOR, default=7): cv.int_range(min=6, max=12),
+            cv.Optional(CONF_SYNC_VALUE, default=[]): cv.ensure_list(cv.hex_uint8_t),
+        },
+    )
+    .extend(cv.COMPONENT_SCHEMA)
+    .extend(spi.spi_device_schema(True, 8e6, "mode0")),
+    validate_config,
+)
```
dala318: Is the `cv.All(..)` really needed?
```suggestion
CONFIG_SCHEMA = cv.Schema(
    {
        cv.GenerateID(): cv.declare_id(SX127x),
        cv.Optional(CONF_AUTO_CAL, default=True): cv.boolean,
        cv.Optional(CONF_BANDWIDTH, default="125_0kHz"): cv.enum(BW),
        cv.Optional(CONF_BITRATE): cv.int_range(min=500, max=300000),
        cv.Optional(CONF_BITSYNC): cv.boolean,
        cv.Optional(CONF_CODING_RATE, default="CR_4_5"): cv.enum(CODING_RATE),
        cv.Optional(CONF_CRC_ENABLE, default=False): cv.boolean,
        cv.Optional(CONF_DEVIATION, default=5000): cv.int_range(min=0, max=100000),
        cv.Optional(CONF_DIO0_PIN): pins.internal_gpio_input_pin_schema,
        cv.Required(CONF_FREQUENCY): cv.int_range(min=137000000, max=1020000000),
        cv.Required(CONF_MODULATION): cv.enum(MOD),
        cv.Optional(CONF_ON_PACKET): automation.validate_automation(single=True),
        cv.Optional(CONF_PA_PIN, default="BOOST"): cv.enum(PA_PIN),
        cv.Optional(CONF_PA_POWER, default=17): cv.int_range(min=0, max=17),
        cv.Optional(CONF_PA_RAMP, default="40us"): cv.enum(RAMP),
        cv.Optional(CONF_PAYLOAD_LENGTH, default=0): cv.int_range(min=0, max=256),
        cv.Optional(CONF_PREAMBLE_DETECT, default=0): cv.int_range(min=0, max=3),
        cv.Optional(CONF_PREAMBLE_ERRORS, default=0): cv.int_range(min=0, max=31),
        cv.Optional(CONF_PREAMBLE_POLARITY, default=0xAA): cv.All(
            cv.hex_int, cv.one_of(0xAA, 0x55)
        ),
        cv.Optional(CONF_PREAMBLE_SIZE, default=0): cv.int_range(min=0, max=65535),
        cv.Required(CONF_RST_PIN): pins.internal_gpio_output_pin_schema,
        cv.Optional(CONF_RX_FLOOR, default=-94): cv.float_range(min=-128, max=-1),
        cv.Optional(CONF_RX_START, default=True): cv.boolean,
        cv.Optional(CONF_SHAPING, default="NONE"): cv.enum(SHAPING),
        cv.Optional(CONF_SPREADING_FACTOR, default=7): cv.int_range(min=6, max=12),
        cv.Optional(CONF_SYNC_VALUE, default=[]): cv.ensure_list(cv.hex_uint8_t),
    },
).extend(
    cv.COMPONENT_SCHEMA
).extend(
    spi.spi_device_schema(True, 8e6, "mode0")
).add_extra(
    validate_config
)
```

swoboda1337: Doesnt make much of a difference but sure can use add_extra instead

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,332 @@
...
+    cg.add(var.set_bandwidth(config[CONF_BANDWIDTH]))
+    cg.add(var.set_frequency(config[CONF_FREQUENCY]))
+    cg.add(var.set_deviation(config[CONF_DEVIATION]))
+    cg.add(var.set_modulation(config[CONF_MODULATION]))
+    cg.add(var.set_pa_pin(config[CONF_PA_PIN]))
+    cg.add(var.set_pa_ramp(config[CONF_PA_RAMP]))
+    cg.add(var.set_pa_power(config[CONF_PA_POWER]))
+    cg.add(var.set_shaping(config[CONF_SHAPING]))
+    if CONF_BITRATE in config:
+        cg.add(var.set_bitrate(config[CONF_BITRATE]))
```
dala318: Just a suggestion, makes the code a bit cleaner (in my oppinion).
```suggestion
    if bitrate := config.get(CONF_BITRATE):
        cg.add(var.set_bitrate(bitrate))
```

esphome/components/sx127x/sx127x.h
```diff
@@ -0,0 +1,123 @@
...
+  Trigger<std::vector<uint8_t>, float, float> *packet_trigger_{new Trigger<std::vector<uint8_t>, float, float>()};
+  std::vector<SX127xListener *> listeners_;
+  std::vector<uint8_t> sync_value_;
+  InternalGPIOPin *dio0_pin_{nullptr};
+  InternalGPIOPin *rst_pin_{nullptr};
+  SX127xBw bandwidth_;
+  uint32_t bitrate_;
+  uint32_t deviation_;
+  uint32_t frequency_;
+  uint32_t payload_length_;
```
dala318, line 104: Rename to `max_payload_length_`? Could otherwise be mistaken for actual payload length.

tests/components/sx127x/common.yaml
```diff
@@ -0,0 +1,44 @@
...
+  shaping: NONE
+  preamble_size: 2
+  preamble_detect: 2
+  preamble_errors: 8
+  preamble_polarity: 0x55
+  on_packet:
+    then:
+      - sx127x.send_packet:
+          data: [0xC5, 0x51, 0x78, 0x82, 0xB7, 0xF9, 0x9C, 0x5C]
+
```
dala318, line 33: Add a `packet_transport` instance as well?

dala318 (commented):
Just some proposals. Thinks it look good in general.

esphome/components/sx127x/automation.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include "esphome/core/component.h"
+#include "esphome/core/automation.h"
+#include "esphome/components/sx127x/sx127x.h"
+
+namespace esphome {
+namespace sx127x {
+
+template<typename... Ts> class RunImageCalAction : public Action<Ts...> {
```
dala318: Is there a reason to why SendPacketAction is `public Parented<SX127x>` but the other actions not?

tests/components/sx127x/common.yaml
swoboda1337: packet_transport is its own thing, dont need actions: https://esphome.io/components/packet_transport/index.html#example-configuration

swoboda1337:
> Just some proposals. Thinks it look good in general.

Thanks! I will look at these and probably make a few changes.

esphome/components/sx127x/sx127x.h
swoboda1337: If the value is 0 its variable length. If it has a non zero value that is the actual length.

esphome/components/sx127x/automation.h
swoboda1337: There is not, I will update them all to use `public Parented<SX127x>`

esphome/components/sx127x/__init__.py
swoboda1337: It looks more consistent with the other lines though:
```
    cg.add(var.set_pa_ramp(config[CONF_PA_RAMP]))
    cg.add(var.set_pa_power(config[CONF_PA_POWER]))
    cg.add(var.set_shaping(config[CONF_SHAPING]))
    if CONF_BITRATE in config:
        cg.add(var.set_bitrate(config[CONF_BITRATE]))
    else:
        cg.add(var.set_bitrate(4800))
```

esphome/components/sx127x/packet_transport/sx127x_transport.cpp
```diff
@@ -0,0 +1,27 @@
+#include "esphome/core/log.h"
+#include "esphome/core/application.h"
+#include "esphome/components/network/util.h"
```
mcuelenaere: I don't believe this include is being used. Having it here, requires the board to be configured with wifi/network enabled.

swoboda1337: I will take a look

mcuelenaere: (I can confirm this PR works without that include. I forked your code to remove it, otherwise I couldn't get it to work on a board with wifi disabled)

swoboda1337: Ok I will update the pr

M66B:
@swoboda1337 Since Lora is often used for IoT devices out of range of Wi-Fi networks, it would be nice to be able to OTA update via Lora.

Something like:

```
ota:
  platform: sx127x
```

A spreading factor of 7 has a theoretical data rate of 5465 kbps, let's say 5 KByte/sec. Sending a 2MB image would take 2000/5 seconds = ~ 7 minutes.

The sending esp device could have a web interface to upload to another esp device.

```
web_server:
  ota: true
    platform: sx127x    
```

Is this feasible? No, of course, is also an answer. I'm just thinking out loud.

swoboda1337:
> Is this feasible? No, of course, is also an answer. I'm just thinking out

Is it possible to do, probably could hack something together. 

Is it feasible, not atm. Would need a device to act as a proxy and forward the update. There is no way to do that atm without hacks.

esphome/components/sx127x/__init__.py
jesserockz: You should be setting the default inside the config_schema, then you can assume there is always a value. Same for the below ones too

swoboda1337: I did this because I made a breaking change and needed to know if it was set in the validation code. It's been long enough though. I will clean it up.

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+        if config[CONF_BANDWIDTH] == "500_0kHz":
+            raise cv.Invalid(f"{config[CONF_BANDWIDTH]} is only available with LORA")
+        if CONF_BITSYNC not in config:
+            raise cv.Invalid("Config 'bitsync' required with FSK/OOK")
+        if CONF_PACKET_MODE not in config:
+            raise cv.Invalid("Config 'packet_mode' required with FSK/OOK")
+        if config[CONF_PACKET_MODE] and CONF_DIO0_PIN not in config:
+            raise cv.Invalid("Config 'dio0_pin' required in packet mode")
+        if config[CONF_PAYLOAD_LENGTH] > 64:
+            raise cv.Invalid("Payload length must be >= 64 with FSK/OOK")
```
bdraco: message doesn't match check

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,478 @@
...
+        uint8_t rssi = this->read_register_(REG_PKT_RSSI_VALUE);
+        int8_t snr = (int8_t) this->read_register_(REG_PKT_SNR_VALUE);
+        std::vector<uint8_t> packet(bytes);
+        this->write_register_(REG_FIFO_ADDR_PTR, addr);
+        this->read_fifo_(packet);
+        if (this->frequency_ > 700000000) {
+          this->call_listeners_(packet, (float) rssi - 157, (float) snr / 4);
+        } else {
+          this->call_listeners_(packet, (float) rssi - 164, (float) snr / 4);
+        }
```
bdraco: nit: Magic numbers count use comments or constants here

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,478 @@
...
+  ESP_LOGCONFIG(TAG, "SX127x:");
+  LOG_PIN("  CS Pin: ", this->cs_);
+  LOG_PIN("  RST Pin: ", this->rst_pin_);
+  LOG_PIN("  DIO0 Pin: ", this->dio0_pin_);
+  ESP_LOGCONFIG(TAG, "  Auto Cal: %s", TRUEFALSE(this->auto_cal_));
+  ESP_LOGCONFIG(TAG, "  Frequency: %" PRIu32 " Hz", this->frequency_);
+  ESP_LOGCONFIG(TAG, "  Bandwidth: %" PRIu32 " Hz", BW_HZ[this->bandwidth_]);
+  ESP_LOGCONFIG(TAG, "  PA Pin: %s", this->pa_pin_ == PA_PIN_BOOST ? "BOOST" : "RFO");
+  ESP_LOGCONFIG(TAG, "  PA Power: %" PRIu8 " dBm", this->pa_power_);
+  ESP_LOGCONFIG(TAG, "  PA Ramp: %" PRIu16 " us", RAMP[this->pa_ramp_]);
```
bdraco: could be optimized a bit https://developers.esphome.io/architecture/logging/

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,478 @@
...
+    }
+    if (this->preamble_size_ > 0) {
+      ESP_LOGCONFIG(TAG, "  Preamble Size: %" PRIu16, this->preamble_size_);
+    }
+  } else {
+    ESP_LOGCONFIG(TAG, "  Modulation: %s", this->modulation_ == MOD_FSK ? "FSK" : "OOK");
+    ESP_LOGCONFIG(TAG, "  Bitrate: %" PRIu32 "b/s", this->bitrate_);
+    ESP_LOGCONFIG(TAG, "  Bitsync: %s", TRUEFALSE(this->bitsync_));
+    ESP_LOGCONFIG(TAG, "  Rx Start: %s", TRUEFALSE(this->rx_start_));
+    ESP_LOGCONFIG(TAG, "  Rx Floor: %.1f dBm", this->rx_floor_);
```
bdraco: https://developers.esphome.io/architecture/logging/

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,478 @@
...
+void SX127x::call_listeners_(const std::vector<uint8_t> &packet, float rssi, float snr) {
+  for (auto &listener : this->listeners_) {
+    listener->on_packet(packet, rssi, snr);
+  }
+  this->packet_trigger_->trigger(packet, rssi, snr);
+}
+
+void SX127x::loop() {
+  if (this->modulation_ == MOD_LORA) {
+    if (this->dio0_pin_->digital_read()) {
```
bdraco: Could use early return here to make this a bit easier to follow

esphome/components/sx127x/sx127x.cpp
```diff
@@ -0,0 +1,478 @@
...
+      this->write_register_(REG_FIFO, packet.size());
+    }
+    this->write_fifo_(packet);
+    this->set_mode_tx();
+  }
+  uint32_t start = millis();
+  while (!this->dio0_pin_->digital_read()) {
+    if (millis() - start > 4000) {
+      ESP_LOGE(TAG, "Transmit packet failure");
+      break;
```
bdraco, line 286: Is there a potential concern about blocking the event loop here?

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+SetModeStandbyAction = sx127x_ns.class_(
+    "SetModeStandbyAction", automation.Action, cg.Parented.template(SX127x)
+)
+
+
+def validate_raw_data(value):
+    if isinstance(value, str):
+        return value.encode("utf-8")
+    if isinstance(value, str):
+        return value
```
bdraco: Looks unreachable

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+            "62_5kHz",
+            "125_0kHz",
+            "250_0kHz",
+            "500_0kHz",
+        ]
+        if config[CONF_BANDWIDTH] not in bws:
+            raise cv.Invalid(f"{config[CONF_BANDWIDTH]} is not available with LORA")
+        if CONF_DIO0_PIN not in config:
+            raise cv.Invalid("Cannot use LoRa without dio0_pin")
+        if config[CONF_PREAMBLE_SIZE] > 0 and config[CONF_PREAMBLE_SIZE] < 6:
```
bdraco:
```suggestion
        if 0 < config[CONF_PREAMBLE_SIZE] < 6:
```

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+            "10_4kHz",
+            "15_6kHz",
+            "20_8kHz",
+            "31_3kHz",
+            "41_7kHz",
+            "62_5kHz",
+            "125_0kHz",
+            "250_0kHz",
+            "500_0kHz",
+        ]
```
bdraco, line 162: nit: Could be a constant frozenset to avoid linear search

esphome/components/sx127x/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+        return value
+    if isinstance(value, list):
+        return cv.Schema([cv.hex_uint8_t])(value)
+    raise cv.Invalid(
+        "data must either be a string wrapped in quotes or a list of bytes"
+    )
+
+
+def validate_config(config):
+    if config[CONF_MODULATION] == "LORA":
```
bdraco, line 150: cv.typed_schema with discriminator=CONF_MODULATION might be cleaner in the schema to avoid hand rolled validate_config but it looks like there is enough overlap that it might not be worth doing

bdraco:
Besides some minor nits, https://github.com/esphome/esphome/pull/7490#discussion_r2181229415 looks like the only potential area of concern remaining

esphome/components/sx127x/sx127x.cpp
swoboda1337: Its basically the same as remote transmitter which blocks as well. Typically you will do other things after transmitting like delay plus transmit again. Making this non blocking can get messy and I think will cause misunderstandings.

The 4 second delay isn't representative of the typical delay. Like remote transmitter this should normally be like 50 ms or less. 

Even if a blocking mode is added I think the default should be blocking.

esphome/components/sx127x/__init__.py
swoboda1337: Must have been a copy and paste error

esphome/components/sx127x/__init__.py
swoboda1337: yeah should be <= wups

esphome/components/sx127x/sx127x.cpp
bdraco: Would you please add some comments explaining this behavior so we don't need to revisit it in the future.

esphome/components/sx127x/sx127x.cpp
swoboda1337: Yeah wrote this a year ago... will update

esphome/components/sx127x/__init__.py
swoboda1337: Only used once, the savings are probably offset by creating the frozenset. Would save like 1 us max and add two lines of code.

bdraco (approved):

bdraco:
Thanks @swoboda1337

PR #7496, opened 2024-09-25
Add default color order

ties-s:

Aligns default value with documentation

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:
I think it's better to fix the docs rather than introduce a default that could be wrong 50% of the time. If the user has to supply a colour order then if it's wrong they will at least have a starting place to tweak rather than diagnosing a silent incorrect default.

ties-s:
I'm also okay with that

PR #7499, opened 2024-09-26
Change priority of ili9xxx from HARDWARE to HARDWARE_LATE 

SqyD:

Simple fix to prevent too early calls to the SPI bus I changed the priority of the ili9xxx display component from HARDWARE to HARDWARE_LATE. 

This resolved an issue where the ili9341 display on the [mch2022 badge](https://badge.team/docs/badges/mch2022/) tried to use the SPI bus too early, causing error ```[15:29:06][W][spi-esp-idf:041]: spi_setup called before initialisation ```
As suggested [in the component source here](https://github.com/esphome/esphome/blob/21fbbc5fb9477704be8b1ec8293d2a729f7a0072/esphome/core/component.h#L27) I changed the priority from ``` HARDWARE``` to ```HARDWARE_LATE``` which resolved the issue. I suspect this will prevent similar issues for others in the future and I don't think it likely will cause issues for existing users. 

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

No user facing changes in configuration or documentation for this fix.

```

## Checklist:

### DIFF
--- a/esphome/components/ili9xxx/ili9xxx_display.cpp
+++ b/esphome/components/ili9xxx/ili9xxx_display.cpp
@@ -126,7 +126,7 @@ void ILI9XXXDisplay::dump_config() {
   LOG_UPDATE_INTERVAL(this);
 }
 
-float ILI9XXXDisplay::get_setup_priority() const { return setup_priority::HARDWARE; }
+float ILI9XXXDisplay::get_setup_priority() const { return setup_priority::HARDWARE_LATE; }
 
 void ILI9XXXDisplay::fill(Color color) {
   if (!this->check_buffer_())


clydebarrow:
Where is the information on the issue that triggered this, i.e. steps to reproduce?

It should not be possible for the ili9xxx component to call SPI functions before SPI setup, because the SPI bus should have priority BUS which is very early. If this is happening then there is likely something that needs fixing rather than papering over the cracks by changing priority.

SqyD:
I'm not sure if the issue is reproducible on other hardware than the mch2022 badge I linked to. The badge has an a-typical setup as it shares the SPI bus between the display and an fpga chip. This could likely contribute to the issue of a slight delay over situations where the SPI is dedicated to the display. The error occurs at boot with any variation of default configuration options for an otherwise standard ili9341 display. I will be glad to send you one of my spare ones if you are interested.

I just did a quick check of setup_priority settings of other display components. The vast majority of them, including inkplate, ssdxxxx and st7xxx displays all have it set to ```PROCESSOR``` which is even after ```HARDWARE_LATE``` so it appears the Ili9xxx is the exception here. As the comments in the components.h file I linked also suggest using ```HARDWARE_LATE``` for displays I thought to at least offer my change as a PR.

If you are of the opinion this is expected behavior or If this is indeed just related to the a-typical board I'm working with I'm fine running my own fork of the component that fixes this for this limited volume hardware.

Cheers!

jesserockz:
@SqyD Can you link to the config you are using for the board?
As Clyde mentioned above, its not possible with the current priorities for this to happen unless something in the board specific YAML is doing stuff early.

SqyD:
Sure thing, [here is my work-in-progress](https://github.com/badgeteam/esphome-badge/blob/master/mch2022/mch2022-badge.yaml#L96)

clydebarrow:
If it's specific to that hardware, then all you need do is set `setup_priority:` in the display config for now to override the default priority. It's an option available to all components. It appears to be undocumented, not sure if this is an oversight or a deliberate choice, but it's definitely available.

SqyD:
> If it's specific to that hardware, then all you need do is set `setup_priority:` 

That is indeed a great solution that I haven't seen anywhere before but completely matches my needs. It appears to work fine for my board. Thanks a lot for that suggestion!

The offer to send out one of these badges still stands. Just ping me with your shipping address at paul@krischer.nl and I'll send one over. (without lipo battery, shipping companies don't appear to like those). No strings attached, just for your enjoyment as gift of a big esphome fan.
Thanks again!

PR #2399, opened 2021-09-26
Option to ignore CRC for EFuse MAC address

mmakaay:
# What does this implement/fix? 

Some devices exist out there for which somehow the MAC address as burnt into EFuse, does not match its accompanying CRC value. When the esp-idf function `esp_efuse_mac_get_default()` is called, that function sees the mismatch and raises an error `Base MAC address from BLK0 of EFUSE CRC error` and calls `abort()`.

When using the Arduino framework, there is no fix available, other than actually hacking esp-idf to treat this issue as a warning in the abovementioned function. Reason for this is that the Arduino library code will always call the function that breaks.
I have [a repository](https://github.com/mmakaay/arduino-esp32-unicore-no-mac-crc-builder) that contains scripting to patch esp-idf and build framework packages that fix the issue. Very cumbersome really.

Now esp-idf is becoming available as a framework, a different fix is in reach. When a custom MAC address is set before this framework looks up what MAC address to use for WiFI, the custom MAC address is used instead of reading it from EFuse.

This PR (specific to ESP32 + framework esp-idf) introduces a new option `ignore_efuse_mac_crc` for the `esp32` component. When set to true, the EFuse MAC address will be read and used as-is, without a CRC check.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [x] ESP32
- [ ] ESP8266

## Example entry for `config.yaml`:

This configuration works for my Xiaomi Bedside Lamp2, which not only needs the CRC fix, but is also a single core ESP32-WROOM-32D.

```yaml
esp32:
  board: esp32doit-devkit-v1
  framework:
    type: esp-idf
    sdkconfig_options:
      CONFIG_FREERTOS_UNICORE: y
    advanced:
      ignore_efuse_mac_crc: true
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).
  
If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #7502, opened 2024-09-26
[wifi] Replace ``USE_ESP32_IGNORE_EFUSE_MAC_CRC`` with IDF's ``CONFIG_ESP_MAC_IGNORE_MAC_CRC_ERROR``

kbx81:

The define `USE_ESP32_IGNORE_EFUSE_MAC_CRC` is redundant and we can just use IDF's native `CONFIG_ESP_MAC_IGNORE_MAC_CRC_ERROR` flag, instead. See IDF's code [here](https://github.com/espressif/esp-idf/blob/e499576efdb086551abe309a72899302f82077b7/components/esp_hw_support/mac_addr.c#L131-L144) and [here](https://github.com/espressif/esp-idf/blob/e499576efdb086551abe309a72899302f82077b7/components/esp_hw_support/mac_addr.c#L98-L105).

This is effectively a streamlining of #2399; note that only the ESP32 (not variants) supports CRC checking of the MAC address.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/core/helpers.cpp
+++ b/esphome/core/helpers.cpp
@@ -662,13 +662,9 @@ void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parame
   static const uint8_t esphome_host_mac_address[6] = USE_ESPHOME_HOST_MAC_ADDRESS;
   memcpy(mac, esphome_host_mac_address, sizeof(esphome_host_mac_address));
 #elif defined(USE_ESP32)
-#if defined(CONFIG_SOC_IEEE802154_SUPPORTED) || defined(USE_ESP32_IGNORE_EFUSE_MAC_CRC)
+#if defined(CONFIG_SOC_IEEE802154_SUPPORTED)
   // When CONFIG_SOC_IEEE802154_SUPPORTED is defined, esp_efuse_mac_get_default
-  // returns the 802.15.4 EUI-64 address. Read directly from eFuse instead.
-  // On some devices, the MAC address that is burnt into EFuse does not
-  // match the CRC that goes along with it. For those devices, this
-  // work-around reads and uses the MAC address as-is from EFuse,
-  // without doing the CRC check.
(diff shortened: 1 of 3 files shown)


esphome/core/helpers.cpp
```diff
@@ -664,7 +664,7 @@ void get_mac_address_raw(uint8_t *mac) {  // NOLINT(readability-non-const-parame
   static const uint8_t esphome_host_mac_address[6] = USE_ESPHOME_HOST_MAC_ADDRESS;
   memcpy(mac, esphome_host_mac_address, sizeof(esphome_host_mac_address));
 #elif defined(USE_ESP32)
-#if defined(CONFIG_SOC_IEEE802154_SUPPORTED) || defined(USE_ESP32_IGNORE_EFUSE_MAC_CRC)
+#if defined(CONFIG_SOC_IEEE802154_SUPPORTED)
   // When CONFIG_SOC_IEEE802154_SUPPORTED is defined, esp_efuse_mac_get_default
   // returns the 802.15.4 EUI-64 address. Read directly from eFuse instead.
   // On some devices, the MAC address that is burnt into EFuse does not
```
jesserockz: The second half of this comment can go too

kbx81: 🤦🏻‍♂️😅

jesserockz:
@mmakaay Would you be able to confirm this works for you as the original author that added this?

jesserockz (approved):

mmakaay:
> @mmakaay Would you be able to confirm this works for you as the original author that added this?

I had pretty much the same PR prepared a while ago, but I must have forgotten to actually drop it in here :-/ 
I will check it against my devices!

PR #7506, opened 2024-09-29
Add OpenThread support on ESP-IDF

### LINKED ISSUE #1397: Zigbee/Thread support on the new upcoming ESP32-H2 variant?

A new variant of the ESP32 [was announced last month](https://www.espressif.com/en/news/ESP32_H2), the ESP32-H2. It supports 802.15.4 and therefore Zigbee and Thread, as well as BLE, and now runs on a RISC-V architecture which should be a lot more efficient. It also still supports WiFi.

This looks super interesting because it will finally give home users a way to easily build Zigbee/Thread versions instead of WiFi ones. I think ESPHome supporting this would be a brilliant step forward.

There are no alternatives to ESPHome for Zigbee as far as I know? Or really any simple development environments for home users.

**Additional context**

Of course the H2 isn't out yet. But I think it'd be good to get a discussion on this going before it is released. Zigbee is definitely a lot more friendly to larger home installations in many ways.

Support for battery operated end Zigbee devices would also be a huge leap forward. And I would think that should be possible on the new H2.

mrene:
Opening this as a draft to get early feedback while it's being cleaned up.

Remaining TODOs:
- [x] Split ESP-IDF specific code into openthread_esp.cpp (ideally the main file would only reference openthread methods)
- [x] Handle Off-Mesh-Routable prefix changes and update the SRP entries
  - OpenThread has a `otSrpClientEnableAutoHostAddress` which does this automatically
- [x] Add sensors for the current thread channel/role/networkname/etc
- [ ] ~~Find if we can get rid of this pool_alloc pattern for passing things to openthread~~
- [ ] Write some docs

This add OpenThread as a possible network stack on supported ESP-IDF platforms. It's been tested on an ESP32-C6 and ESP32-H2.

I've adapted the SRP component from the Zephyr esphome implementation and integrated it with the mdns component so that it populates the TXT records accordingly (mDNS is delegated to an SRP server in thread networks to avoid RF traffic).

## Types of changes

- closes esphome/feature-requests#1397
- closes esphome/feature-requests#2644

- esphome/esphome-docs#4306

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: openthread

esp32:
  board: esp32-c6-devkitm-1
  framework:
    type: esp-idf

network:
  enable_ipv6: true

openthread:
  channel: 13
  network_name: OpenThread-8f28
  network_key: 0xdfd34f0f05cad978ec4e32b0413038ff
  panid: 0x8f28
  extpanid: 0xd63e8e3e495ebbc3
  pskc: 0xc23a76e98f1a6483639b1ac1271e2e27

logger:
  level: DEBUG

api:

# https://github.com/esphome/esphome/pull/7506
external_components:
  - source: github://pr#7506
    components: [ network, mdns, openthread ]
    refresh: 0s
  - source: github://pr#8975
    components: esp32
    refresh: 0s
```

## Checklist:


optimusprimespace:
Using your test openthread config from the text file [test-ot.yaml](https://github.com/esphome/esphome/pull/7506/files#diff-0a6afaf2588163104ee0db222a1dcf343bb8361612387ee8c139457d618ec7cd) with a xiao esp32-c6`
external_components:
  - source:
      type: git
      url: https://github.com/mrene/esphome
      ref: openthread
    components: [ openthread ]
    refresh: 0s

esp32:
  board: esp32-c6-devkitc-1
  flash_size: 4MB
  variant: esp32c6
  framework:
    type: esp-idf
    version: "5.3.0"
    platform_version: 6.8.1
    sdkconfig_options:
      CONFIG_OPENTHREAD_ENABLED: y
      CONFIG_ENABLE_WIFI_STATION: y
      CONFIG_USE_MINIMAL_MDNS: y
      CONFIG_ESPTOOLPY_FLASHSIZE_4MB: y` I got the following error: src/esphome/components/openthread/openthread.cpp: In member function 'void esphome::openthread::OpenThreadComponent::srp_setup()':
src/esphome/components/openthread/openthread.cpp:253:41: error: 'class esphome::mdns::MDNSComponent' has no member named 'get_services'
  253 |     this->mdns_services_ = this->mdns_->get_services();
      |                                         ^~~~~~~~~~~~

mrene:
> Using your test openthread config from the text file [test-ot.yaml](https://github.com/esphome/esphome/pull/7506/files#diff-0a6afaf2588163104ee0db222a1dcf343bb8361612387ee8c139457d618ec7cd) with a xiao esp32-c6` 

are you using the code from this PR? there are changes to the `mdns` and `network` component that you need to pull if you want to use this off-tree (you can also reference them via external components iirc)

lboue:
Are you also facing this `OPENTHREAD_BUILD_DATETIME` error during building?
```
Compiling .pioenvs/openthread/openthread/openthread/src/core/api/history_tracker_api.cpp.o
Compiling .pioenvs/openthread/openthread/openthread/src/core/api/icmp6_api.cpp.o
Compiling .pioenvs/openthread/openthread/openthread/src/core/api/instance_api.cpp.o
Compiling .pioenvs/openthread/openthread/openthread/src/core/api/ip6_api.cpp.o
<command-line>: error: invalid digit "9" in octal constant
/data/cache/platformio/packages/framework-espidf/components/openthread/openthread/src/core/api/instance_api.cpp:177:47: note: in expansion of macro 'OPENTHREAD_BUILD_DATETIME'
  177 |                                          "; " OPENTHREAD_BUILD_DATETIME
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~
Compiling .pioenvs/openthread/openthread/openthread/src/core/api/jam_detection_api.cpp.o
Compiling .pioenvs/openthread/openthread/openthread/src/core/api/joiner_api.cpp.o
Compiling .pioenvs/openthread/openthread/openthread/src/core/api/link_api.cpp.o
/data/cache/platformio/packages/framework-espidf/components/openthread/openthread/src/core/api/instance_api.cpp: In function 'const char* otGetVersionString()':
<command-line>: error: expected ',' or ';' before numeric constant
/data/cache/platformio/packages/framework-espidf/components/openthread/openthread/src/core/api/instance_api.cpp:177:47: note: in expansion of macro 'OPENTHREAD_BUILD_DATETIME'
  177 |                                          "; " OPENTHREAD_BUILD_DATETIME
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~
*** [.pioenvs/openthread/openthread/openthread/src/core/api/instance_api.cpp.o] Error 1
========================= [FAILED] Took 595.69 seconds =========================

```

mrene:
> Are you also facing this `OPENTHREAD_BUILD_DATETIME` error during building?

@lboue Can you paste your config?

lboue:
@mrene 
Here it is:

```yaml
esphome:
  name: openthread

esp32:
  board: esp32-c6-devkitm-1
  variant: esp32c6
  framework:
    type: esp-idf
    version: "5.3.0"
    platform_version: "https://github.com/pioarduino/platform-espressif32/releases/download/53.03.10%%2Brc1/platform-espressif32.zip"

network:
  enable_ipv6: true

openthread:
  channel: 13
  network_name: OpenThread-8f28
  network_key: dfd34f0f05cad978ec4e32b0413038ff
  panid: 0x8f28
  extpanid: d63e8e3e495ebbc3
  pskc: c23a76e98f1a6483639b1ac1271e2e27

# The web server will cause the HA integration to fail, see note in the README
# web_server:
#   port: 80

api:
  encryption:
    key: "n4HolGwz01UdfUKKS6DZZ/0OfHlTBduVAmzrezSXIxI="

# https://github.com/esphome/esphome/pull/7506
external_components:
  # equivalent shorthand for GitHub pull request
  - source: github://pr#7506
    components: [ network, mdns, openthread ]
    refresh: 0s

# This is the "Boot" button on my dev board
binary_sensor:
  - platform: gpio
    id: button1
    name: "Button"
    publish_initial_state: true
    pin:
      number: GPIO9
      mode: INPUT_PULLUP
      inverted: true

light:
  - platform: esp32_rmt_led_strip
    chipset: WS2812
    pin: GPIO8
    num_leds: 1
    rmt_channel: 0
    name: "Status LED RGB"
    id: statusledlight
    icon: "mdi:led-outline"
    rgb_order: GRB

# text_sensor:
#   - platform: template
#     name: "OTT Thread RLOC16"
#     lambda: |-
#       if (!esp_openthread_lock_acquire(1)) {
#           return {"Unknown"};
#       }
#       auto instance = esp_openthread_get_instance();
#       auto rloc16 = otThreadGetRloc16(instance);
#       esp_openthread_lock_release();
#       char buf[10];
#       snprintf(buf, sizeof(buf), "%04x", rloc16);
#       return {buf};
#   - platform: template
#     name: "OTT Thread Role"
#     lambda: |-
#       if (!esp_openthread_lock_acquire(1)) {
#           return {"Unknown"};
#       }
#       auto instance = esp_openthread_get_instance();
#       auto role = otThreadDeviceRoleToString(otThreadGetDeviceRole(instance));
#       esp_openthread_lock_release();
#       return {role};
```

mrene:
> openthread.yaml.txt

@lboue odd, I can build that config just fine here (under linux)

optimusprimespace:
When I successfully compile mine i get this error: `OPENTHREAD: Host connection mode none
phy_init: phy_version 310,dde1ba9,Jun  4 2024,16:38:11
phy: libbtbb version: 04952fd, Jun  4 2024, 16:38:26
OPENTHREAD:[I] ChildSupervsn-: Timeout: 0 -> 190
OPENTHREAD: OpenThread attached to netif
OPENTHREAD:[N] Mle-----------: Role disabled -> detached
I (936) OT_STATE: netif up
I(1456) OPENTHREAD:[N] Mle-----------: Attach attempt 1, AnyPartition reattaching with Active Dataset
E (6436) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
E (6436) task_wdt:  - loopTask (CPU 0)
E (6436) task_wdt: Tasks currently running:
E (6436) task_wdt: CPU 0: ot_cli
E (6436) task_wdt: Aborting.` And it constantly reboots the Xiao ESP32-C6. PS: I don't have anything that the ESP can connect to, but i'm certain that it wouldn't restart every time it tries to connect to a OT network.

optimusprimespace:
I get this error too: `<command-line>: error: invalid digit "9" in octal constant
/data/cache/platformio/packages/framework-espidf@3.50300.0/components/openthread/openthread/src/core/api/instance_api.cpp:177:47: note: in expansion of macro 'OPENTHREAD_BUILD_DATETIME'
  177 |                                          "; " OPENTHREAD_BUILD_DATETIME
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~
Compiling .pioenvs/esp32-c6-openthread/openthread/openthread/src/core/api/jam_detection_api.cpp.o
Compiling .pioenvs/esp32-c6-openthread/openthread/openthread/src/core/api/joiner_api.cpp.o
Compiling .pioenvs/esp32-c6-openthread/openthread/openthread/src/core/api/link_api.cpp.o
/data/cache/platformio/packages/framework-espidf@3.50300.0/components/openthread/openthread/src/core/api/instance_api.cpp: In function 'const char* otGetVersionString()':
<command-line>: error: expected ',' or ';' before numeric constant
/data/cache/platformio/packages/framework-espidf@3.50300.0/components/openthread/openthread/src/core/api/instance_api.cpp:177:47: note: in expansion of macro 'OPENTHREAD_BUILD_DATETIME'
  177 |                                          "; " OPENTHREAD_BUILD_DATETIME` when using     ```platform_version: "https://github.com/pioarduino/platform-espressif32/releases/download/53.03.10%%2Brc1/platform-espressif32.zip"```  try using it like: `platform_version: 6.8.1` for example

mrene:
@optimusprimespace this compile error is intermittent? 

I've reproduced (and fixed) the watchdog reset problem, it happens because openthread wants to spawn its command line interface on the same uart as the logger (I had been using the usb_serial_jtag logger instead of the uart, which avoided the problem).  

Even if you have nothing to connect to, thread is a mesh network, it'll simply start a new one and become its leader. 

<details><summary>Example logs (at debug level)</summary>
<p>

```
[18:57:45]I (565) main_task: Started on CPU0
[18:57:45]I (592) main_task: Calling app_main()
[18:57:45][I][logger:156]: Log initialized
[18:57:45][I][app:029]: Running through setup()...
[18:57:45][D][esp-idf:000]: I (748) gpio: GPIO[9]| InputEn: 1| OutputEn: 0| OpenDrain: 0| Pullup: 1| Pulldown: 0| Intr:0
[18:57:45]
[18:57:45][D][binary_sensor:034]: 'Button': Sending initial state OFF
[18:57:45][I][openthread:046]: Setting up OpenThread...
[18:57:45][D][esp-idf:000][ot_main]: I (752) OPENTHREAD: Host connection mode none
[18:57:45]
[18:57:45][D][esp-idf:000][ot_main]: I (775) phy: phy_version: 310,2, 8fa9cb4, Jun  5 2024, 14:19:10
[18:57:45]
[18:57:45][D][esp-idf:000][ot_main]: I (776) phy: libbtbb version: a024da6, Jun  5 2024, 14:19:22
[18:57:45]
[18:57:45][D][esp-idf:000][ot_main]: I(790) OPENTHREAD:
[18:57:45][D][esp-idf:000][ot_main]: [I] ChildSupervsn-: Timeout: 0 -> 190
[18:57:45][D][esp-idf:000][ot_main]:
[18:57:45]
[18:57:45][D][esp-idf:000][ot_main]: I (803) OPENTHREAD: OpenThread attached to netif
[18:57:45]
[18:57:45][I][openthread:131][ot_main]: Activating dataset...
[18:57:45][D][esp-idf:000][ot_main]: I(832) OPENTHREAD:
[18:57:45][D][esp-idf:000][ot_main]: [N] Mle-----------: Role disabled -> detached
[18:57:45][D][esp-idf:000][ot_main]:
[18:57:45]
[18:57:45][D][esp-idf:000][ot_main]: I (851) OT_STATE: netif up
[18:57:45]
[18:57:45][D][esp-idf:000][ot_main]: I(864) OPENTHREAD:
[18:57:45][D][esp-idf:000][ot_main]: [N] Mle-----------: Attach attempt 1, AnyPartition reattaching with Active Dataset
[18:57:45][D][esp-idf:000][ot_main]:
[18:57:45]
[18:57:52][D][esp-idf:000][ot_main]: I(7395) OPENTHREAD:
[18:57:52][D][esp-idf:000][ot_main]: [N] RouterTable---: Allocate router id 52
[18:57:52][D][esp-idf:000][ot_main]:
[18:57:52]
[18:57:52][D][esp-idf:000][ot_main]: I(7398) OPENTHREAD:
[18:57:52][D][esp-idf:000][ot_main]: [N] Mle-----------: RLOC16 fffe -> d000
[18:57:52][D][esp-idf:000][ot_main]:
[18:57:52]
[18:57:52][D][esp-idf:000][ot_main]: I(7412) OPENTHREAD:
[18:57:52][D][esp-idf:000][ot_main]: [N] Mle-----------: Role detached -> leader
[18:57:52][D][esp-idf:000][ot_main]:
[18:57:52]
[18:57:52][D][esp-idf:000][ot_main]: I(7433) OPENTHREAD:
[18:57:52][D][esp-idf:000][ot_main]: [N] Mle-----------: Partition ID 0x530fe6
[18:57:52][D][esp-idf:000][ot_main]:
[18:57:52]
```

</p>
</details> 

If you have another c6/h2, you can flash esp-idf's `ot_rcp` example, and use it to form a border router on one of your machines (it's the same kind of firmware that runs on these thread usb dongles),

lboue:
I tried to build on Raspberry Pi 4 (aarch64).

lboue:
I tried ESPHome 2024.9.0 and 2024.10.0 and it fails for both on x86_64 VM too:
- [logs_thread_compile_2024.10.0-dev.txt](https://github.com/user-attachments/files/17188298/logs_thread_compile_2024.10.0-dev.txt)
- [logs_thread_compile_2024.9.0-dev.txt](https://github.com/user-attachments/files/17188300/logs_thread_compile_2024.9.0-dev.txt)

```
#ifdef OPENTHREAD_BUILD_DATETIME
                                         "; " OPENTHREAD_BUILD_DATETIME
```
https://github.com/openthread/openthread/blob/main/src/core/api/instance_api.cpp#L174-L175


Could you tell me which version works correctly?

Regards

optimusprimespace:
As i've noted above i only get the error when using `platform_version: "https://github.com/pioarduino/platform-espressif32/releases/download/53.03.10%%2Brc1/platform-espressif32.zip" `, if i switch to `platform_version: 6.8.1` it works without any compile error. PS: don't forget to set version to `version: "5.3.1"` too. I think platformio's new version 6.9 has some problem with the new ESP-IDF version, but not entirely sure.

optimusprimespace:
I'm new to the OT standard, but i have two ESP32-C6 boards. If i want to have a working connection to HA, do i need to have a 3rd ESP32 to connect with my C6 (like here: https://github.com/espressif/esp-idf/blob/master/examples/openthread/ot_br/image/thread-border-router-esp32-esp32h2.jpg) or i can use it directly with HA? I read somewhere that it can only work with Thread or Wi-fi only. I want this to work as a weather station with one border router and one end device sending measured sensor data back to Home Assistant.

lboue:
> I'm new to the OT standard, but i have two ESP32-C6 boards. If i want to have a working connection to HA, do i need to have a 3rd ESP32 to connect with my C6 (like here: https://github.com/espressif/esp-idf/blob/master/examples/openthread/ot_br/image/thread-border-router-esp32-esp32h2.jpg) or i can use it directly with HA? I read somewhere that it can only work with Thread or Wi-fi only. I want this to work as a weather station with one border router and one end device sending measured sensor data back to Home Assistant.

You should ask this in [discussions](https://github.com/esphome/esphome/discussions).

lboue:
> platform_version: 6.8.1
Thanks. It works:
```yaml
esp32:
  board: esp32-c6-devkitm-1
  variant: esp32c6
  framework:
    type: esp-idf
    version: "5.3.1"
    platform_version: 6.8.1
```

mrene:
> I'm new to the OT standard, but i have two ESP32-C6 boards. If i want to have a working connection to HA, do i need to have a 3rd ESP32 to connect with my C6 (like here: https://github.com/espressif/esp-idf/blob/master/examples/openthread/ot_br/image/thread-border-router-esp32-esp32h2.jpg) or i can use it directly with HA? I read somewhere that it can only work with Thread or Wi-fi only. I want this to work as a weather station with one border router and one end device sending measured sensor data back to Home Assistant.

@optimusprimespace You're right that ESPs can only connect to Thread or Wifi, but not both. Espressif's example for setting up a border router involves doing so using only ESP devices, but it's also possible to use your home assistant machine to become a border router [with the appropriate extension](https://www.home-assistant.io/integrations/otbr/). You will need a Thread dongle, or a Thread-capable ESP flashed with the `ot_rcp` example (thread radios follow a standard protocol over USB/serial). 

Once you have a border router, you'll be able to access the thread network from your local network via IPv6 (if you don't already have IPv6 connectivity, it'll automatically assign address ranges and devices in your LAN will automatically assign addresses on it). Feel free to open a discussion as suggested to discuss further.

lboue:
Do you know how to fix failed CI tests?

mrene:
> Do you know how to fix failed CI tests?

@lboue They are mostly comestic / formatting failures, except for the "doesn't have documentation" failure :sweat_smile:

lboue:
I started to work on documentation:
https://deploy-preview-4306--esphome.netlify.app/components/openthread/

lboue:
> > Do you know how to fix failed CI tests?
> 
> @lboue They are mostly comestic / formatting failures, except for the "doesn't have documentation" failure 😅

Could you reference this doc PR in this PR: https://github.com/esphome/esphome-docs/pull/4306?

tomaszduda23:
Cool work. I was looking into OpenThread for nrf52840 https://github.com/esphome/esphome/pull/7049. It could be extended later on.

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            "CONFIG_OPENTHREAD_NETWORK_NAME", f"{config[CONF_NETWORK_NAME]}"
+        )
+    if config[CONF_EXTPANID]:
+        add_idf_sdkconfig_option(
+            "CONFIG_OPENTHREAD_NETWORK_EXTPANID", f"{config[CONF_EXTPANID]}"
+        )
+    if config[CONF_PSKC]:
+        add_idf_sdkconfig_option(
+            "CONFIG_OPENTHREAD_NETWORK_PSKC", f"{config[CONF_PSKC]}"
+        )
```
austindrenski: Looks like these are optional here

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            cv.GenerateID(CONF_MDNS_ID): cv.use_id(MDNSComponent),
+            cv.Required(CONF_NETWORK_NAME): cv.string_strict,
+            cv.Required(CONF_CHANNEL): cv.int_,
+            cv.Required(CONF_NETWORK_KEY): cv.string_strict,
+            cv.Required(CONF_PSKC): cv.string_strict,
+            cv.Required(CONF_PANID): cv.int_,
+            cv.Required(CONF_EXTPANID): cv.string_strict,
+        }
+    ),
+)
```
austindrenski: But then being required here. Can you clarify what is actually required?

image

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+        add_idf_sdkconfig_option(
+            "CONFIG_OPENTHREAD_NETWORK_PSKC", f"{config[CONF_PSKC]}"
+        )
+
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_DNS64_CLIENT", True)
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_SRP_CLIENT", True)
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_SRP_CLIENT_MAX_SERVICES", 5)
+
+    # TODO: Add suport for sleepy end devices
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_FTD", True)  # Full Thread Device
```
austindrenski, line 76: Personally very interested in `CONFIG_OPENTHREAD_MTD` support, fwiw

esphome/components/openthread/__init__.py
mrene: It looks like I forgot to update the schema. 

This needs to be tested but according to the spec only the network key is required to join a thread network, with the caveat that it'll have to scan until it finds an advertisement packet it can decrypt.

To be on the safer side, I intended to have channel/network_name/network_key as required fields.

esphome/components/openthread/__init__.py
mrene: Likewise. There's an [ot_sleepy_device](https://github.com/espressif/esp-idf/tree/master/examples/openthread/ot_sleepy_device) example in esp-idf that can be used for inspiration (I haven't looked at how it integrates with esphome's sleep yet)

mrene:
I think this is ready for an initial review pass.

There's one last thing that I'm looking for feedback on:

I'm wondering the best way to handle openthread dataset (network) changes. The problem is that thread comes with a management protocol where any of the values can change remotely. This is very handy when migrating between channels, rotating keys, etc, but it means that the initial network set in  esphome's configuration is not what's going to be used unless it's the first execution, because afterwards it'll just read the dataset from non-volatile storage (which esp-idf maintains for us).  This gets worse because the nvs partition isn't cleared when flashing the device, you have to swap over to idf.py erase-flash to be able to change thread networks.

I'm tempted to compute a hash of the values, save it as a preference, and override the nvs data if that hash has changed, but I wonder if someone has a better idea.

lkm:
> > platform_version: 6.8.1
> > Thanks. It works:
> 
> ```yaml
> esp32:
>   board: esp32-c6-devkitm-1
>   variant: esp32c6
>   framework:
>     type: esp-idf
>     version: "5.3.1"
>     platform_version: 6.8.1
> ```

@lboue: Unfortunately this doesn't work for the h2 boards as the `esp32-h2-devkitm-1` doesn't exist in either 6.8.1 nor 6.9. I've tried both 5.3.0 with pioarduino 53.03.10-rc1 and 5.3.1 with rc2. Same compile error: OPENTHREAD_BUILD_DATETIME

lboue:
> @lboue: Unfortunately this doesn't work for the h2 boards as the `esp32-h2-devkitm-1` doesn't exist in either 6.8.1 nor 6.9. I've tried both 5.3.0 with pioarduino 53.03.10-rc1 and 5.3.1 with rc2. Same compile error: OPENTHREAD_BUILD_DATETIME

Check this: https://github.com/esphome/esphome/discussions/7737

lkm:
Thanks @lboue, however, I'm not sure I get that. It's not the config but the compile failure I'm concerned about. I've solved that problem by manually patching framework-espidf package. There's definitely something that going wrong with `OPENTHREAD_BUILD_DATETIME` the in the macro there.

@mrene I'm now trying to get the esp-h2 to join an existing thread network which doesn't seem possible. I've checked and tested that I can join with those credentials using another node. I suspect the panid isn't passed down to the openthread library correctly and is always set to 0x1234

mrene:
> @mrene I'm now trying to get the esp-h2 to join an existing thread network which doesn't seem possible. I've checked and tested that I can join with those credentials using another node. I suspect the panid isn't passed down to the openthread library correctly and is always set to 0x1234

@lkm I'm assuming you flashed the example beforehand. As mentioned in https://github.com/esphome/esphome/pull/7506#issuecomment-2425246399 the device will remember the previous thread dataset  (because they have been passed to esp-idf/openthread and can be remotely updated) - try erasing the device's flash using `esptool.py erase_flash` and it should use the new thread dataset once you re-flash it.

Edit: I've been told that a factory reset button component will also erase non-volatile storage but I haven't tested it.

martin-a-o:
I'm having the same problem connecting to an existing thread network. By tweaking sdkconfig.project I get the correct pan id (all other openthread settings are from the project yaml) in the text_sensor output, but the device announces itself as leader and isn't connected to the established thread network.

Regarding remembering the previous thread dataset: I have experienced this when trying the examples from the esp matter sdk. Before trying this esphome-on-thread, my h2 device was a functional matter device connected to my established thread network.

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_CLI", False)
+
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_ENABLED", True)
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_NETWORK_PANID", config[CONF_PANID])
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_NETWORK_CHANNEL", config[CONF_CHANNEL])
+    add_idf_sdkconfig_option(
+        "CONFIG_OPENTHREAD_NETWORK_MASTERKEY", f"{config[CONF_NETWORK_KEY]}"
+    )
+
+    if config[CONF_NETWORK_NAME]:
```
sturlig: This will give an error when it isn't in the config. Please use `if CONF_NETWORK_NAME in config:`
Same for the next 2 if's.

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+OpenThreadComponent = openthread_ns.class_("OpenThreadComponent", cg.Component)
+OpenThreadSrpComponent = openthread_ns.class_("OpenThreadSrpComponent", cg.Component)
+
+CONFIG_SCHEMA = cv.All(
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(OpenThreadComponent),
+            cv.GenerateID(CONF_SRP_ID): cv.declare_id(OpenThreadSrpComponent),
+            cv.GenerateID(CONF_MDNS_ID): cv.use_id(MDNSComponent),
+            cv.Required(CONF_PANID): cv.int_,
```
sturlig: PANID should go in hex notation in the sdkconfig file. When it is saved as integer, it won't be used. To fix this please change `cv.int_` to `cv.hex_int`

mrene:
> I'm having the same problem connecting to an existing thread network. By tweaking sdkconfig.project I get the correct pan id (all other openthread settings are from the project yaml) in the text_sensor output, but the device announces itself as leader and isn't connected to the established thread network.

@martin-a-o why did you have to tweak the sdkconfig, was a value there wrong? Maybe you hit this issue? (https://github.com/esphome/esphome/pull/7506#discussion_r1853798352)
I have sometimes seen this behavior if the device doesn't receive any response when asking for a parent, but the device ended up joining the network after it receives an advertisement from another device, usually within ~30s (it notices it's in a newer network partition than the existing one).

@lkm I've added a `force_dataset` option (boolean) that ignores any previously joined thread dataset (network) and uses the one from the config at every boot, this should make testing easier. I've hopped between two thread networks by changing the config with this option set.

You can add it to the config like this:
```yaml
openthread:
  channel: 13
  network_name: OpenThread-8f28
  network_key: dfd34f0f05cad978ec4e32b0413038ff
  panid: 0x1234
  extpanid: d63e8e3e495ebbc3
  pskc: c23a76e98f1a6483639b1ac1271e2e27
  force_dataset: true
```

martin-a-o:
> @martin-a-o why did you have to tweak the sdkconfig, was a value there wrong? Maybe you hit this issue? ([#7506 (comment)](https://github.com/esphome/esphome/pull/7506#discussion_r1853798352)) I have sometimes seen this behavior if the device doesn't receive any response when asking for a parent, but the device ended up joining the network after it receives an advertisement from another device, usually within ~30s (it notices it's in a newer network partition than the existing one).

Yes, the panid was an int in sdkconfig.project.esphomeinternal, but 0x1234 in sdkconfig.project. After changing it in sdkconfig.project the correct value was printed by openthread_info, but it still wasn't able to connect to the network.

I'll try again with the latest commits and force_dataset: true.

Edit: I got some progress now using the correct channel, which I had set wrong in the first attempts.

martin-a-o:
I've got my device into Home assistant now, it has been running for a day, but mDNS is unstable. The device stops mDNS announcements and appears as "offline" in Esphome - but stays connected to Home assistant if the IP is the same.

Now after a code update it got a new IP however, mDNS refuse to cooperate and it's offline in HA.

Edit: I think it's a problem with my network and border routers.

stefanklug:
@lkm @mrene @optimusprimespace I also hit the ```OPENTHREAD_BUILD_DATETIME``` issue. I dug a bit deeper and it seems this might be a cmake issue (not sure yet). 

A workaround is to edit 
```.platformio/packages/framework-espidf/components/openthread/CMakeLists.txt``` and 
replace 
```string(TIMESTAMP OT_BUILD_TIMESTAMP "%Y-%m-%d %H:%M:%S UTC" UTC)```
with
```string(TIMESTAMP OT_BUILD_TIMESTAMP "%Y-%m-%d_%H:%M:%SUTC" UTC)``` (spaces removed or replaced by underscore)

I don't know if that timestamp is needed later in the original format, but it compiles.

The issue seems to be somewhere between CMake and the compiler call. In the CMakeLists.txt there is the following code:
```
target_compile_definitions(
        ${COMPONENT_LIB}
        PUBLIC
        "OPENTHREAD_CONFIG_FILE=\"openthread-core-esp32x-${CONFIG_FILE_TYPE}-config.h\""
            "${device_type}"
        PRIVATE
        "PACKAGE_VERSION=\"${IDF_VERSION_FOR_OPENTHREAD_PACKAGE}-${OPENTHREAD_VERSION}\""
        "OPENTHREAD_BUILD_DATETIME=\"${OT_BUILD_TIMESTAMP}\""
        )
```
when ```OT_BUILD_TIMESTAMP``` contains spaces, OPENTHREAD_BUILD_DATETIME gets passed to the compiler as 
```
"-DOPENTHREAD_BUILD_DATETIME=""2025-01-20 22:23:00 UTC"""
```
while it should be
```
"-DOPENTHREAD_BUILD_DATETIME=\"2025-01-20 22:23:00 UTC\""
```
Making the above modifications, it gets passed as
```
-DOPENTHREAD_BUILD_DATETIME=\"2025-01-20_22:23:00UTC\"
```
I don't know where that quoting error occurs. cmake? ninja?. But I don't think openthread is directly to blame here although passing defines with escaped double quotes does feel errorprone. But at least the workaround is relatively easy to apply... and maybe someone has an idea where to look for the root cause.

File: esphome/components/openthread_info/openthread_info_text_sensor.cpp [hunk 6]
```diff
@@ -0,0 +1,24 @@
...
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace openthread_info {
+
+static const char *const TAG = "openthread_info";
+
+void IPAddressOpenThreadInfo::dump_config() { LOG_TEXT_SENSOR("", "OpenThreadInfo IPAddress", this); }
+void RoleOpenThreadInfo::dump_config() { LOG_TEXT_SENSOR("", "OpenThreadInfo Role", this); }
+void ChannelOpenThreadInfo::dump_config() { LOG_TEXT_SENSOR("", "OpenThreadInfo Role", this); }
```
MrEbbinghaus: Looks like a copy&paste error here. Should be "Channel" instead of "Role".

benwa:
How are we supposed to get the required configuration parameters so the device can join an already established Thread network?

mathbou:
> How are we supposed to get the required configuration parameters so the device can join an already established Thread network?

If you are running an OTBR container/vm and can log into it, you can use the `ot-ctl` command:

```
ot-ctl channel
ot-ctl networkname
ot-ctl networkkey
ot-ctl panid
ot-ctl extpanid
ot-ctl pskc
```

Otherwise, it will depend on what your hardware exposes on the UI side. If obtaining the `network` and `pan` infos should be quite easy, the pskc is rarely exposed.
It is, however, written in the `dataset TLV `, near the end. On mine, it starts at the 177 character, or 44 from the end, for a 32char length.

benwa:
> > How are we supposed to get the required configuration parameters so the device can join an already established Thread network?
> 
> If you are running an OTBR container/vm and can log into it, you can use the `ot-ctl` command:
> 
> ```
> ot-ctl channel
> ot-ctl networkname
> ot-ctl networkkey
> ot-ctl panid
> ot-ctl extpanid
> ot-ctl pskc
> ```
> 
> Otherwise, it will depend on what your hardware exposes on the UI side. If obtaining the `network` and `pan` infos should be quite easy, the pskc is rarely exposed. It is, however, written in the `dataset TLV `, near the end. On mine, it starts at the 177 character, or 44 from the end, for a 32char length.

It's a Google TV Streamer (4K), so I doubt I'd be able to get it. How would an off the shelf product join that network? Shouldn't that be doable here too?

oli-the-kitty:
I've tried this on ESP32C6, mostly following the example config. Got it to compile and run but it does not connect to the network properly. I'm running a Sonoff ZBDongle-E flashed with SiliconLabs Multiprotocol. Here's the log from the SiliconLabs Multiprotocol addon from when I plug in the ESP:
```
otbr-agent[303]: 01:54:06.048 [N] MeshForwarder-: Failed to send IPv6 UDP msg, len:83, chksum:9edd, ecn:no, to:0xffff, sec:yes, error:NoAck, prio:net
otbr-agent[303]: 01:54:06.048 [N] MeshForwarder-:     src:[fe80:0:0:0:481e:903a:e3f4:a4f]:19788
otbr-agent[303]: 01:54:06.048 [N] MeshForwarder-:     dst:[ff02:0:0:0:0:0:0:1]:19788
otbr-agent[303]: 01:54:06.051 [W] Platform------: Error processing result: InvalidState
otbr-agent[303]: 01:54:06.052 [W] Platform------: Error waiting response: InvalidState
otbr-agent[303]: 01:54:06.052 [W] SubMac--------: RadioReceive() failed, error: InvalidState
otbr-agent[303]: 02:05:42.523 [N] Ip6-----------: Error UDP Checksum
otbr-agent[303]: 02:05:42.523 [N] Ip6-----------: Failed to handle payload: Drop
otbr-agent[303]: 02:05:45.361 [N] Ip6-----------: Error UDP Checksum
otbr-agent[303]: 02:05:45.361 [N] Ip6-----------: Failed to handle payload: Drop
otbr-agent[303]: 02:05:51.036 [N] Ip6-----------: Error UDP Checksum
otbr-agent[303]: 02:05:51.037 [N] Ip6-----------: Failed to handle payload: Drop
otbr-agent[303]: 02:06:02.383 [N] Ip6-----------: Error UDP Checksum
otbr-agent[303]: 02:06:02.383 [N] Ip6-----------: Failed to handle payload: Drop
otbr-agent[303]: 02:06:25.081 [N] Ip6-----------: Error UDP Checksum
otbr-agent[303]: 02:06:25.082 [N] Ip6-----------: Failed to handle payload: Drop
```

And my config:
```yaml
esphome:
  name: esp32c6-openthread-test
  friendly_name: ESP32C6 OpenThread test

esp32:
  variant: ESP32C6
  board: esp32-c6-devkitm-1
  flash_size: 8MB
  framework:
    type: esp-idf
    version: "5.3.1"
    platform_version: 6.8.1
    
# Enable logging
logger:
  level: VERBOSE

# Enable Home Assistant API
api:
  encryption:
    key: "removed"

network:
  enable_ipv6: true

openthread:
  channel: 0x0f
  network_name: ha-thread-9347
  network_key: removed
  panid: 0xremoved
  extpanid: removed
  pskc: removed
  force_dataset: true

# https://github.com/esphome/esphome/pull/7506
external_components:
  - source: github://pr#7506
    components: [ network, mdns, openthread ]
    refresh: 0s
```
Compiled using latest ESPHome Device Builder (2025.2.2) under Home Assistant 2025.2.5 on Raspberry Pi 4.
If anyone has any ideas, I'm willing to test and experiment

MrEbbinghaus:
@benwa I got the credentials from the Home Assistant app. (on iOS)
Settings -> Companion app -> Debug -> Thread 

There you can get the Active Operational Dataset, network name, network key, and extended PAN-ID.

You can use the openthread cli to decode the dataset by configuring a new thread network with the dataset and then query the pskc and channel. (see https://openthread.io/reference/cli/concepts/dataset)

(It would be great if the esphome `openthread` integration could take the encoded dataset. In the future maybe directly from Home Assistant.)

Hedda:
@mrene FYI, developer @felipejfc has just submitted a new pull request based on this but adds OpenThread client device support to the expirimental nrf52 platform (which it itself has not yet been merged into ESPHome however is it looking close):

- https://github.com/esphome/esphome/pull/8410

  - https://github.com/felipejfc/esphome/tree/nrf52-ot

Hedda:
> This add OpenThread as a possible network stack on supported ESP-IDF platforms. It's been tested on an ESP32-C6 and ESP32-H2.

May I suggest using this new M5Stack Unit Gateway H2 as reference hardware for tests & proof-of-concept as it has a Grove port:

* https://shop.m5stack.com/products/esp32-h2-thread-zigbee-gateway-unit

* https://docs.m5stack.com/en/unit/Unit%20Gateway%20H2

* https://www.youtube.com/watch?v=wjfyz15jWmw&ab_channel=M5Stack

M5Stack’s Unit Gateway H2 is sold for ~$7 as a development kit with a Grove port and powered by a ESP32-H2-MINI-1 module:

* https://www.cnx-software.com/2025/03/17/m5stack-unit-gateway-h2-tiny-esp32-h2-thread-zigbee-gateway/


Image

Image

Image

PS: FYI, [Home Assistant Voice Preview Edition](https://www.home-assistant.io/blog/2024/12/19/voice-preview-edition-the-era-of-open-voice/) also has a Grove post since allow easy plug-and-play to test sensors with ESPHome.

* https://wiki.seeedstudio.com/Grove_System/

* https://voice-pe.home-assistant.io/guides/grove_port/

lboue:
@Hedda 
[Module Gateway H2 + CoreS3  ESP Thread Boarder Router doc](https://docs.m5stack.com/en/guide/thread/module_gateway_h2/thread_border_router)

alanlaird:
The M5stack product linked above seems to be out of stock.  Would this ESP32-C6 part be an acceptable alternative since it also has a grove port?

https://shop.m5stack.com/products/m5stack-nanoc6-dev-kit

image

Hedda:
> _Would this ESP32-C6 part be an acceptable alternative since it also has a grove port?_
> 
> _https://shop.m5stack.com/products/m5stack-nanoc6-dev-kit_

Yes for development and testing (plus that $6 M5Stack NanoC6 Dev Kit looks like an awesome development kit for prototyping and proof-of-concept), however I believe for real-world use cases in ”production” (i.e. deploying the final products as permanent devices) you will more likely want to implement the ESP32-H2 for battery-powered sleepy devices (such as low-power sensors like example a motion sensor or a temperature sensor that do not wake-up very often to report state changes).

I think for real-world use cases in ”production” you probably only would want to implement the ESP32-C6 for mains-powered non-sleepy devices that need to report state changes more often as well perhaps as also being able to act as a Thread Router (Thread repeater) or Thread Border Router to extend the overall range and coverage of your Thread network.

For reference; ESP32-H2 is an ultra-low power SoC that has just IEEE 802.15.4 (for Thread or Zigbee) + BLE 5 radios and its single-core can only be clocked up to 96 MHz and is as such specifically targeting energy-efficient applications, while the more power-hungry ESP32-C6 is still "low power" it also has a WiFi 6 radio and its core can be clocked up to 160 MHz so is really targeting applications that require WiFi and/or higher performance.

rwrozelle:
I'm testing this using a SuperMini ESP32-H2, I've got it to attach to my PAN as a Router.  I added a switch component, but nothing is showing up in HA.  Am I mistaken about the intended functionality?  I thought with the API component, we would be able to use other esphome components.

bobrozelle:
I got it working on a ESP32H2 Super Mini, when started, it attaches as a ROUTER and I can use other ESPHOME components such as Platform Button and servo.  Not sure why, but I had to set the local mesh prefix to get it to connect as a ROUTER:
esp32:
  board: esp32-h2-devkitm-1
  variant: esp32h2
  framework:
    type: esp-idf
    # This contains the esp32-h2-devkit board
    platform_version: 51.03.04
    sdkconfig_options: 
      CONFIG_MBEDTLS_SSL_PROTO_DTLS: "y"
      CONFIG_MBEDTLS_KEY_EXCHANGE_ECJPAKE: "y"
      CONFIG_MBEDTLS_ECJPAKE_C: "y"
      CONFIG_OPENTHREAD_JOINER: "y"
      CONFIG_OPENTHREAD_MESH_LOCAL_PREFIX: "fd52:135f:ed21:ad80::/64"

Hedda:
FYI, new battery-powered Seeed Studio IoT Button is based on an ESP32-C6 SoC; ”_supports ESPHome or Zigbee firmware_” (though their provided Zigbee firmware is not based on ESPHome, this hardware might be good proof-of-concept as it as can technically support WiFi, Zigbee, or Thread):

- https://www.cnx-software.com/2025/04/13/battery-powered-seeed-studio-esp32-c6-iot-button-esphome-zigbee-firmware/

- https://www.seeedstudio.com/Seeed-IoT-Button-p-6419.html

- https://github.com/Seeed-Projects/Seeed_IoT_Button

- https://wiki.seeedstudio.com/getting_started_with_seeed_iot_button/

- https://files.seeedstudio.com/wiki/iot_button_zigbee/Seeed_Studio_IoT_Button_PCB.zip

- https://www.aliexpress.com/item/1005008801906619.html
- 

image

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+# TODO: Doesn't conflict with wifi if you're using another ESP as an RCP (radio coprocessor), but this isn't implemented yet
+CONFLICTS_WITH = ["wifi"]
+DEPENDENCIES = ["esp32"]
+
+CONF_NETWORK_NAME = "network_name"
+CONF_NETWORK_KEY = "network_key"
+CONF_PSKC = "pskc"
+CONF_PANID = "panid"
+CONF_EXTPANID = "extpanid"
+CONF_MDNS_ID = "mdns_id"
```
bobrozelle: CONF_MESHLOCALPREFIX = "meshlocalprefix"

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+    add_idf_sdkconfig_option("CONFIG_OPENTHREAD_NETWORK_CHANNEL", config[CONF_CHANNEL])
+    add_idf_sdkconfig_option(
+        "CONFIG_OPENTHREAD_NETWORK_MASTERKEY", f"{config[CONF_NETWORK_KEY]}"
+    )
+
+    if CONF_NETWORK_NAME in config:
+        add_idf_sdkconfig_option(
+            "CONFIG_OPENTHREAD_NETWORK_NAME", f"{config[CONF_NETWORK_NAME]}"
+        )
+    if CONF_EXTPANID in config:
```
bobrozelle: if CONF_MESHLOCALPREFIX in config:
        add_idf_sdkconfig_option(
            "CONFIG_OPENTHREAD_MESH_LOCAL_PREFIX", f"{config[CONF_MESHLOCALPREFIX]}"
        )

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+            cv.GenerateID(): cv.declare_id(OpenThreadComponent),
+            cv.GenerateID(CONF_SRP_ID): cv.declare_id(OpenThreadSrpComponent),
+            cv.GenerateID(CONF_MDNS_ID): cv.use_id(MDNSComponent),
+            cv.Required(CONF_PANID): cv.hex_int,
+            cv.Required(CONF_CHANNEL): cv.int_,
+            cv.Required(CONF_NETWORK_KEY): cv.string_strict,
+            cv.Optional(CONF_EXTPANID): cv.string_strict,
+            cv.Optional(CONF_NETWORK_NAME): cv.string_strict,
+            cv.Optional(CONF_PSKC): cv.string_strict,
+            cv.Optional(CONF_FORCE_DATASET): cv.boolean,
```
bobrozelle, line 112: cv.Optional(CONF_MESHLOCALPREFIX): cv.string_strict,

rwrozelle:
pioarduino has a fix in development for compiling issue with OPENTHREAD_BUILD_DATE https://github.com/esphome/esphome/discussions/7737

Below worked on 5/25/2025:
framework:
type: esp-idf
version: 5.4.1
platform_version: https://github.com/pioarduino/platform-espressif32.git#develop

Hedda:
Off-topic but I think that with the introduction of native OpenThread in ESPHome there will a huge interest in making low-power battery-operated devices combining the Thread protocol and ESPHome, so wondering if you also have a roadmap for extending ESPHome built-in power-management funtionality to make sleepy devices using sleeping technics and/or even deep sleep features?

jesserockz:
I have pushed a bunch of (breaking) changes to this PR, mostly around the configuration value keys and types. I also introduced a new config key `tlv` which allows taking the entire tlv value from the Thread information in Home Assistant and the individual values will be automatically extracted from it. If you want to see the individual values, you can use `esphome config ...` or the `validate` button to print out the config with those values.

I have also made a bunch of PRs to support this one better. #8973 #8974 #8975

With all of those changes, this PR works with the default versions of the platform and idf in ESPHome.

Thanks for getting the PR this far.

lboue:
Do you have an example of the new config key `tlv` configuration so we can update doc PR?
Is it like this?
```yaml
network:
  enable_ipv6: true

openthread:
  tlv: 0e080000000000010000000300001035060004001fffe00208e227ac6a7f24052f0708fdb753eb517cb4d3051062b2442a928d9ea3b947a1618fc4085a030f4f70656e5468726561642d393837330102987304105330d857354330133c05e1fd7ae81a910c0402a0f7f8
```

Dataset from doc here: https://openthread.io/reference/cli/concepts/dataset

rwrozelle:
So you were able to resolve compiling issue with OPENTHREAD_BUILD_DATE (https://github.com/esphome/esphome/discussions/7737) without having to use development branch of platformio ?

lboue:
> So you were able to resolve compiling issue with OPENTHREAD_BUILD_DATE (#7737) without having to use development branch of platformio ?

This time it worked fine with ESPHome 2025.5.1 default setting. This downloaded the following versions:
https://github.com/pioarduino/platform-espressif32/releases/download/53.03.13/platform-espressif32.zip
https://github.com/pioarduino/esp-idf/releases/download/v5.3.2/esp-idf-v5.3.2.zip

There is a fix in the pioarduino develop branch:
- https://github.com/esphome/esphome/discussions/7737#discussioncomment-13263148
- https://github.com/pioarduino/platform-espressif32/commit/7d84d3208c1829a8155e655a411969aa2336e9ab

jesserockz:
Yes, I asked to get the patch included in the platform-espressif32 53.03.13 release so we don't have to have different versions just for openthread.

https://github.com/pioarduino/platform-espressif32/pull/185#issuecomment-2922170389

lboue:
Now it builds OK but I only have the bootrom log a boot:
```
[22:51:28]ESP-ROM:esp32c6-20220919
[22:51:28]Build:Sep 19 2022
[22:51:28]rst:0x1 (POWERON),boot:0xc (SPI_FAST_FLASH_BOOT)
[22:51:28]SPIWP:0xee
[22:51:28]mode:DIO, clock div:2
[22:51:28]load:0x40875720,len:0x18ec
[22:51:28]load:0x4086c110,len:0xd9c
[22:51:28]load:0x4086e610,len:0x2f80
[22:51:28]entry 0x4086c110
```
And no ESPHome log. Anyone facing the same issue?

rwrozelle:
tlv is found at: HA -> Settings -> Devices & services -> Thread -> Services -> Configure -> Information (the little i in a circle)

esphome/components/openthread_info/openthread_info_text_sensor.h
```diff
@@ -0,0 +1,228 @@
...
+    otIp6AddressToString(&*address, addressAsString, 40);
+    std::string ip = addressAsString;
+
+    if (this->last_ip_ != ip) {
+      this->last_ip_ = ip;
+      this->publish_state(this->last_ip_);
+    }
+  }
+  float get_setup_priority() const override { return setup_priority::AFTER_WIFI; }
+  std::string unique_id() override { return get_mac_address() + "-openthreadinfo-ip"; }
```
jesserockz, line 47: Remove in a PR after merging this

esphome/components/openthread_info/openthread_info_text_sensor.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "openthread_info_text_sensor.h"
+#ifdef USE_OPENTHREAD
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace openthread_info {
+
+static const char *const TAG = "openthread_info";
+
+void IPAddressOpenThreadInfo::dump_config() { LOG_TEXT_SENSOR("", "OpenThreadInfo IPAddress", this); }
```
jesserockz, line 11: Remove string prefix

esphome/components/openthread/openthread_esp.cpp
```diff
@@ -0,0 +1,166 @@
...
+  }
+  return {};
+}
+
+std::optional<InstanceLock> InstanceLock::acquire() {
+  while (!esp_openthread_lock_acquire(100)) {
+    esp_task_wdt_reset();
+  }
+  return InstanceLock();
+}
```
jesserockz, line 158: I don't like this as it has the potential to block forever, the code calling it should be written to try again in another loop (to be removed in a future PR)

esphome/components/openthread/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+async def to_code(config):
+    cg.add_define("USE_OPENTHREAD")
+
+    ot = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(ot, config)
+
+    srp = cg.new_Pvariable(config[CONF_SRP_ID])
+    cg.add(srp.set_host_name(cg.RawExpression(f'"{CORE.name}"')))
```
jesserockz, line 143: This should be using `App.get_name()` in the c++ code, not doing it here. Otherwise it will not get the mac suffix.

jesserockz (approved):

lboue:
Doc preview: https://deploy-preview-4306--esphome.netlify.app/components/openthread

PR #7494, opened 2024-09-25
tcs34725: optimize fetch time with burst read for RGB and clear values

RubenKelevra:
# What does this implement/fix?

- Switched to a single burst read operation for R, G, B, and clear channels. This replaces four separate I2C commands with one, allowing us to pull all 16 bytes in one go.
- Significantly reduces roundtrip overhead, improving the overall performance.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Other: Small optimization

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

<details>

```yaml
esphome:
  name: $devicename
  platform: ESP8266
  board: d1_mini
  friendly_name: $friendly_name

substitutions:
  devicename: ambient-light-sensor
  friendly_name: "Ambient Light Sensor"

logger:
  level: DEBUG

mdns:
  
time:
  - platform: homeassistant

api:
  encryption: 
    key: !secret api_key-ambient-light-sensor

ota:
  platform: esphome
  password: !secret ota_password_ambient-light-sensor

wifi:
  networks:
    ssid: !secret wifi_ssid
    password: !secret wifi_password
    manual_ip:
      static_ip: !secret ip-ambient-light-sensor
      gateway: !secret wifi_gateway
      subnet: !secret wifi_subnet
      dns1: !secret wifi_dns1

i2c:
  sda: D1
  scl: D2
  scan: true
  id: bus_a

sensor:
  - platform: tcs34725
    i2c_id: bus_a
    red_channel:
      name: "$friendly_name Red Channel relative"
      accuracy_decimals: 1
    green_channel:
      name: "$friendly_name Green Channel relative"
      accuracy_decimals: 1
    blue_channel:
      name: "$friendly_name Blue Channel relative"
      accuracy_decimals: 1
    clear_channel:
      name: "$friendly_name Clear Channel"
      accuracy_decimals: 1
    illuminance:
      name: "$friendly_name Illuminance"
[...]

PR #7508, opened 2024-09-29
TCS34725: add accurate sensor saturation; remove clear channel

RubenKelevra:
<strike>This PR contains two commits which are filed separately as PR #7492 and PR #7494, which need to be merged first.</strike>

# What does this implement?

1. **Removal of Clear Channel Handling:**
   - The `clear_channel` attribute and associated code have been removed from the `TCS34725Component`.
   - The `calculate_temperature_and_lux_` function has been modified to no longer use or handle the clear channel (`c`) as an input parameter.
   - `clear_channel` related sensor configuration and data handling have been replaced with the newly introduced `sensor_saturation`.

2. **Introduction of Sensor Saturation:**
   - The `sensor_saturation` attribute has been added to provide a direct representation of sensor saturation, making it more informative.
   - This value is now logged and used to determine if samples should be discarded or used, based on specified saturation thresholds.

3. **Adjustment of Saturation and Minimum Raw Value Handling:**
   - The code now uses the maximum of the RGB channels (`peak_raw_value`) to calculate saturation instead of relying on the clear channel.
   - As the clear channel is no longer used, we now use the minimum raw value (`min_raw_value`) of R, G, B to determining underexposure conditions. The minimum raw value of 5 is used in accordance to the spec sheet, to avoid sensor noise on very low light conditions. In other circumstances, 1 is used.

4. **Dynamic Range Improvement:**
   - By removing the dependency on the clear channel, the overall dynamic range of the sensor has been increased by approximately 269%.
   - This is achieved by relying solely on the RGB channels, which prevents the clear channel from prematurely limiting the sensor’s range due to overexposure.

5. **Test Updates:**
   - Test configuration files have been updated to reflect the removal of the `clear_channel` and the addition of `sensor_saturation`.
   - All relevant test cases and configurations have been adapted to the new attribute and its handling.

# What does this fix?

- fix low light noisy output, where the color temperature will just flop up and down:

Screenshot_2024-09-29-08-06-32-082_io homeassistant companion android

- fix overexposure issues, as the red channel was never checked for overexposure. So very reddish light could outperform the clear channel in its value and clip. The spec sheet shows, that the red channel response can be 110% of what the clear channel response is in the red region under some circumstances

- Very limited bright light performance - 8,000-12,000 lux depending on the white light characteristics, as described [here](https://github.com/esphome/issues/issues/3575). Now this sensor will read values as high as 83,000 lux before clipping.

Screenshot_2024-09-29-10-31-28-906_io homeassistant companion android

# Rationale

It may be counterintuitive at the first glance, but removing the clear channel improves this sensor's output:

The clear channel is a sum of the spectrum of the R, G, B channels. Meaning, its value is also a sum. The clear channel therefore will overexpose long before the R, G, B channels will. This is artificially reducing the capabilities of this sensor, as the R, G, B channels are all we really need for color, color temperature and lux calculations.

The clear channel is also absolutely not useful for any purpose, as its response curve is quite similar to the R, G, B channel ones, but cannot be corrected, as it's a sum. So a user can just sum up the R, G, B channels, to get to the clear channel's output, if there's a need for this value.

Removing the clear channel gives us around 269% more dynamic range, with no downsides.

# Limits of this PR

As I was asked to keep the PRs small, this PR will break the percentage based color calculations. The color calculation will be added back in the next PR. So this is not meant to be released on its own, but together with the next PR. 

The next PR is #7517 which adds color output back, as absolut instead of relative outputs like before, so it's more useful. 

The PR after that, #7629, adds CIE1931 color output to the sensor, based on the limited calibration data provided by the vendor. 

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

<details>

```yaml
esphome:
  name: $devicename
  platform: ESP8266
  board: d1_mini
  friendly_name: $friendly_name

substitutions:

logger:
  level: DEBUG

mdns:
  
time:
  - platform: homeassistant

api:
  encryption: 

ota:
  platform: esphome

wifi:
  networks:
    ssid: !secret wifi_ssid
    manual_ip:
      gateway: !secret wifi_gateway
      subnet: !secret wifi_subnet
      dns1: !secret wifi_dns1

i2c:
  sda: D1
  scl: D2
  scan: true
  id: bus_a

sensor:
  - platform: tcs34725
    i2c_id: bus_a
    red_channel:
      name: "$friendly_name Red Channel relative"
      accuracy_decimals: 1
    green_channel:
      name: "$friendly_name Green Channel relative"
      accuracy_decimals: 1
    blue_channel:
      name: "$friendly_name Blue Channel relative"
      accuracy_decimals: 1
    sensor_saturation:
      accuracy_decimals: 1
    illuminance:
      accuracy_decimals: 0
    color_temperature:
      accuracy_decimals: 0
    gain: 1x
    integration_time: auto
    glass_attenuation_factor: 1.0
    address: 0x29   

output:
  - platform: gpio
    pin: D3
    id: white_led
    inverted: false
  
light:
  - platform: binary
    output: white_led
```

</details>

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs). ([PR 4288](https://github.com/esphome/esphome-docs/pull/4288))

### DIFF
--- a/esphome/components/tcs34725/sensor.py
+++ b/esphome/components/tcs34725/sensor.py
@@ -10,6 +10,7 @@ from esphome.const import (
     CONF_INTEGRATION_TIME,
     DEVICE_CLASS_ILLUMINANCE,
     ICON_LIGHTBULB,
+    ICON_GAUGE,
     STATE_CLASS_MEASUREMENT,
     UNIT_PERCENT,
     ICON_THERMOMETER,
@@ -23,6 +24,7 @@ CONF_RED_CHANNEL = "red_channel"
 CONF_GREEN_CHANNEL = "green_channel"
 CONF_BLUE_CHANNEL = "blue_channel"
 CONF_CLEAR_CHANNEL = "clear_channel"
+CONF_SENSOR_SATURATION = "sensor_saturation"
 
 tcs34725_ns = cg.esphome_ns.namespace("tcs34725")
 TCS34725Component = tcs34725_ns.class_(
@@ -66,6 +68,12 @@ color_channel_schema = sensor.sensor_schema(
     accuracy_decimals=1,
     state_class=STATE_CLASS_MEASUREMENT,
 )
+sensor_saturation_schema = sensor.sensor_schema(
+    unit_of_measurement=UNIT_PERCENT,
+    icon=ICON_GAUGE,
+    accuracy_decimals=1,
+    state_class=STATE_CLASS_MEASUREMENT,
+)
 color_temperature_schema = sensor.sensor_schema(
     unit_of_measurement=UNIT_KELVIN,
     icon=ICON_THERMOMETER,
@@ -86,7 +94,10 @@ CONFIG_SCHEMA = (
             cv.Optional(CONF_RED_CHANNEL): color_channel_schema,
             cv.Optional(CONF_GREEN_CHANNEL): color_channel_schema,
             cv.Optional(CONF_BLUE_CHANNEL): color_channel_schema,
-            cv.Optional(CONF_CLEAR_CHANNEL): color_channel_schema,
+            cv.Optional(CONF_CLEAR_CHANNEL): cv.invalid(
+                "The 'clear_channel' configuration option has been removed. Use 'sensor_saturation' instead."
+            ),
+            cv.Optional(CONF_SENSOR_SATURATION): sensor_saturation_schema,
             cv.Optional(CONF_ILLUMINANCE): illuminance_schema,
             cv.Optional(CONF_COLOR_TEMPERATURE): color_temperature_schema,
             cv.Optional(CONF_INTEGRATION_TIME, default="auto"): cv.enum(
@@ -121,9 +132,9 @@ async def to_code(config):
     if CONF_BLUE_CHANNEL in config:
         sens = await sensor.new_sensor(config[CONF_BLUE_CHANNEL])
         cg.add(var.set_blue_sensor(sens))
-    if CONF_CLEAR_CHANNEL in config:
-        sens = await sensor.new_sensor(config[CONF_CLEAR_CHANNEL])
-        cg.add(var.set_clear_sensor(sens))
+    if CONF_SENSOR_SATURATION in config:
+        sens = await sensor.new_sensor(config[CONF_SENSOR_SATURATION])
+        cg.add(var.set_sensor_saturation(sens))
     if CONF_ILLUMINANCE in config:
         sens = await sensor.new_sensor(config[CONF_ILLUMINANCE])
         cg.add(var.set_illuminance_sensor(sens))
--- a/esphome/components/tcs34725/tcs34725.cpp
+++ b/esphome/components/tcs34725/tcs34725.cpp
@@ -1,8 +1,9 @@
 #include "tcs34725.h"
 #include "esphome/core/log.h"
 #include "esphome/core/hal.h"
-#include <algorithm>
 #include "esphome/core/helpers.h"
+#include <algorithm>
+#include <cmath>
 
 namespace esphome {
 namespace tcs34725 {
@@ -50,7 +51,7 @@ void TCS34725Component::dump_config() {
   }
   LOG_UPDATE_INTERVAL(this);
 
-  LOG_SENSOR("  ", "Clear Channel", this->clear_sensor_);
+  LOG_SENSOR("  ", "Sensor Saturation", this->sensor_saturation_);
   LOG_SENSOR("  ", "Red Channel", this->red_sensor_);
   LOG_SENSOR("  ", "Green Channel", this->green_sensor_);
   LOG_SENSOR("  ", "Blue Channel", this->blue_sensor_);
@@ -68,12 +69,16 @@ float TCS34725Component::get_setup_priority() const { return setup_priority::DAT
  *          Green value
  *  @param  b
  *          Blue value
- *  @param  c
- *          Clear channel value
+ *  @param  current_saturation
+ *          Sensor saturation in percent
+ *  @param  min_raw_value
+ *          lowest raw value reported by the sensor
  *  @return Color temperature in degrees Kelvin
  */
-void TCS34725Component::calculate_temperature_and_lux_(uint16_t r, uint16_t g, uint16_t b, uint16_t c) {
-  float sat; /* Digital saturation level */
+void TCS34725Component::calculate_temperature_and_lux_(uint16_t r, uint16_t g, uint16_t b, float current_saturation,
+                                                       uint16_t min_raw_value) {
+  float sat_limit;
+  uint16_t min_raw_limit;
 
   this->illuminance_ = NAN;
   this->color_temperature_ = NAN;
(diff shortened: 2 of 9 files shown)


RubenKelevra:
Sorry, git messed up the rebase of the commits made in the meantime in `dev` that's why the bot added the wrong tags/mentioned the wrong people.

RubenKelevra:
@jesserockz quick question: My next PR has a sensor calibration file - derived from the spec sheet, and two scripts to calculate the sensor calibration based off of that.

As they may come handy on any other color sensor with calibration data in the specs, I was wondering if there's a place in the repo to put such stuff, or if I should publish it in a repo of my own?

RubenKelevra:
Clang-tidy complains about std::clamp() not existing, are we talking about pre-C++17 standard here? I mean, the code compiles fine on my ESPHome installation, so it does exist in the std lib shipped with ESPHome.

RubenKelevra:
I revised the "fix low light noisy output, where the color temperature will just flop up and down" fix, to be less stringent. It will now no longer ensure a minimum of 6 (raw value) but a minimum of 1, as the sensor seems to have less noise than anticipated by the spec sheet.

jesserockz:
> @jesserockz quick question: My next PR has a sensor calibration file - derived from the spec sheet, and two scripts to calculate the sensor calibration based off of that.
> 
> As they may come handy on any other color sensor with calibration data in the specs, I was wondering if there's a place in the repo to put such stuff, or if I should publish it in a repo of my own?

Your own repo would be a better place for this

esphome/components/tcs34725/sensor.py
```diff
@@ -86,7 +93,7 @@
             cv.Optional(CONF_RED_CHANNEL): color_channel_schema,
             cv.Optional(CONF_GREEN_CHANNEL): color_channel_schema,
             cv.Optional(CONF_BLUE_CHANNEL): color_channel_schema,
-            cv.Optional(CONF_CLEAR_CHANNEL): color_channel_schema,
```
jesserockz, line 89: Please leave the key there but with a value of `cv.invalid("Some short message explaining the removal")`

esphome/components/tcs34725/tcs34725.cpp
```diff
@@ -198,26 +208,31 @@ void TCS34725Component::update() {
   uint16_t raw_g = encode_uint16(data[5], data[4]);  // Green channel
   uint16_t raw_b = encode_uint16(data[7], data[6]);  // Blue channel
 
-  ESP_LOGV(TAG, "Raw values clear=%d red=%d green=%d blue=%d", raw_c, raw_r, raw_g, raw_b);
+  ESP_LOGD(TAG, "Raw values clear=%d red=%d green=%d blue=%d", raw_c, raw_r, raw_g, raw_b);
```
jesserockz:
```suggestion
  ESP_LOGV(TAG, "Raw values clear=%d red=%d green=%d blue=%d", raw_c, raw_r, raw_g, raw_b);
```

jesserockz (changes requested):

RubenKelevra:
> > @jesserockz quick question: My next PR has a sensor calibration file - derived from the spec sheet, and two scripts to calculate the sensor calibration based off of that.
> > As they may come handy on any other color sensor with calibration data in the specs, I was wondering if there's a place in the repo to put such stuff, or if I should publish it in a repo of my own?
> 
> Your own repo would be a better place for this

Alright, then they stay at https://github.com/RubenKelevra/tcs34725-irradiance-correction

swifty99:
Hey [@RubenKelevra](https://github.com/RubenKelevra),

thanks for improving this. Nice.

One comment on saturation:
the ga glass attenuation will lead to limited output too. I think the limit should be done on sensor information and not on the user provided glass attenuation. I use(d) a grey filter to not cap at super bright situations and the ga set accordingly.


The code here:
```
  // Lux Calculation (DN40 3.2)

  float g1 = R_COEF * (float) r + G_COEF * (float) g + B_COEF * (float) b;
  float cpl = (this->integration_time_ * this->gain_) / (ga * DF);

  this->illuminance_ = std::max(g1 / cpl, 0.0f);

  if (this->illuminance_ > MAX_ILLUMINANCE) {
    ESP_LOGW(TAG, "Calculated illuminance greater than limit (%f), setting to NAN", this->illuminance_);
    this->illuminance_ = NAN;
    return;
  }
```

I think it would be better to apply ga after the limitation.

RubenKelevra:
> thanks for improving this. Nice.

You're welcome!
 
> One comment on saturation: the ga glass attenuation will lead to limited output too. I think the limit should be done on sensor information and not on the user provided glass attenuation. I use(d) a grey filter to not cap at super bright situations and the ga set accordingly.
> 
> The code here:
> 
> ```
>   // Lux Calculation (DN40 3.2)
> 
>   float g1 = R_COEF * (float) r + G_COEF * (float) g + B_COEF * (float) b;
>   float cpl = (this->integration_time_ * this->gain_) / (ga * DF);
> 
>   this->illuminance_ = std::max(g1 / cpl, 0.0f);
> 
>   if (this->illuminance_ > MAX_ILLUMINANCE) {
>     ESP_LOGW(TAG, "Calculated illuminance greater than limit (%f), setting to NAN", this->illuminance_);
>     this->illuminance_ = NAN;
>     return;
>   }
> ```
> 
> I think it would be better to apply ga after the limitation.

Yeah, makes sense. But I think that's better put into a new PR, so I can test the change individually. 

For this PR it maybe makes sense to raise the limit to 200,000 lux in the meantime, to not cap it?

I just want to avoid putting inf lux out, which we did before.

RubenKelevra:
@swifty99 temporary fix has been added here.

Permanent fix is in PR #7574

esphome/components/tcs34725/tcs34725.cpp
```diff
@@ -226,49 +241,54 @@ void TCS34725Component::update() {
...
   if (this->integration_time_auto_) {
     // change integration time an gain to achieve maximum resolution an dynamic range
     // calculate optimal integration time to achieve 70% satuaration
     float integration_time_ideal;
 
-    integration_time_ideal = 60 / ((float) std::max((uint16_t) 1, raw_c) / 655.35f) * this->integration_time_;
+    integration_time_ideal = 60 / ((float) std::max((uint16_t) 1, peak_raw_value) / 655.35f) * this->integration_time_;
 
     uint8_t gain_reg_val_new = this->gain_reg_;
     // increase gain if less than 20% of white channel used and high integration time
```
swifty99: maybe this comment should reflect the new behavior

swifty99 (approved):
Looks good to me. Nice work.

esphome/components/tcs34725/tcs34725.cpp
RubenKelevra: Sorry which comment?

RubenKelevra: yeah, I went over some of the code comments. Thanks!

Here's the diff:

https://github.com/esphome/esphome/compare/ecea95d50c9403df1791eae4fd0312211ac93811..c438e04602d43e803a50ae5818e82ad6fa1ffbb6

swifty99:
I have this running since a few days. Found no trouble.

RubenKelevra:
@swifty99 thanks for your confirmation!

RubenKelevra:
@jesserockz can you please take another look? Thanks!

swifty99:
what a bummer. So much effort put into this.

PR #7517, opened 2024-10-01
TCS34725: spec based irradiance output

RubenKelevra:

This PR is the second half of https://github.com/esphome/esphome/pull/7508.

This PR introduces a new feature for the TCS34725 sensor component: the ability to report irradiance (in µW/cm²) for each color channel (red, green, and blue). The implementation extracts the spectral responsivity curves from the TCS34725 datasheet and leverages them to convert the raw sensor readings into irradiance values for more precise and meaningful measurements.

Summary of what was done:

- Spectral responsivity curves for the red, green, blue, and clear channels were extracted from the TCS34725 datasheet:

Screenshot_20240925_183102

- The datasheet provides irradiance measurements for the clear channel, with three specified light sources:

Screenshot_20240925_200800

- I've simulated those light sources onto the clear channel to get a conversion factor for the clear channel from unitless responsivity specified in the graph to the counts/irradiance specification.
- The median of those three measurements (which are pretty close) can be used to convert other responsivity values to counts/irradiance, so I did this for the red, green, blue channels.
- Now the spectrum width of this different channels need to be factored in, too, and then we can use the final conversion factor to convert the “counts” the sensor gives us to irradiance measurements.

The source code for these calculations can be found [here](https://github.com/RubenKelevra/tcs34725-irradiance-correction), as ESPHome will just use the result for performance reasons. 

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

<details>

```yaml
esphome:
  name: $devicename
  platform: ESP8266
  board: d1_mini
  friendly_name: $friendly_name

substitutions:

logger:
  esp8266_store_log_strings_in_flash: false
  level: DEBUG

mdns:
  
time:
  - platform: homeassistant

api:
  encryption: 

ota:
  platform: esphome

wifi:
  networks:
    ssid: !secret wifi_ssid
    manual_ip:
      gateway: !secret wifi_gateway
      subnet: !secret wifi_subnet
      dns1: !secret wifi_dns1
      #dns2: !secret wifi_dns2
  fast_connect: true
  reboot_timeout: 5min

i2c:
  sda: D1
  scl: D2
  scan: true
  id: bus_a

sensor:
  - platform: adc
    pin: VCC
    name: "VCC Voltage"
    entity_category: "diagnostic"
    update_interval: 1s
    filters:
      - or:
          - throttle: 600s
          - delta: 0.2
  - platform: uptime
    name: Uptime
    entity_category: "diagnostic"
    update_interval: 300s
  - platform: wifi_signal
    name: "WiFi Signal"
    accuracy_decimals: 1
    entity_category: "diagnostic"
    update_interval: 5s
    filters:
      - or:
          - throttle: 600s
          - delta: 2
  - platform: tcs34725
    i2c_id: bus_a
    red_channel_irradiance:
      name: "$friendly_name Red Channel Irradiance"
      accuracy_decimals: 1
      filters:
        - or:
            - throttle: 300s
            - delta: 0.01
    green_channel_irradiance:
      name: "$friendly_name Green Channel Irradiance"
      accuracy_decimals: 1
      filters:
        - or:
            - throttle: 300s
            - delta: 0.01
    blue_channel_irradiance:
      name: "$friendly_name Blue Channel Irradiance"
      accuracy_decimals: 1
      filters:
        - or:
            - throttle: 300s
            - delta: 0.01
    sensor_saturation:
      accuracy_decimals: 1
      filters:
        - or:
            - throttle: 300s
            - delta: 0.2
    illuminance:
      accuracy_decimals: 2
      filters:
        - or:
            - throttle: 300s
            - delta: 0.1
    color_temperature:
      accuracy_decimals: 2
      filters:
        - or:
            - throttle: 300s
            - delta: 2
    integration_time: auto
    glass_attenuation_factor: 1.0
    address: 0x29
    update_interval: 1s    
 

output:
  - platform: gpio
    pin: D3
    id: white_led
    inverted: false
  
light:
  - platform: binary
    output: white_led

```

</details>

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs), see https://github.com/esphome/esphome-docs/pull/4293


RubenKelevra:
I've noticed a couple of bugs in my calculations. I'm currently testing the fix. Will send it as "Ready for review" as soon as testing is complete and fix is pushed.

RubenKelevra:
Applied same solution for the now missing config options, as [requested](https://github.com/esphome/esphome/pull/7508#discussion_r1789285120) for PR #7508.

RubenKelevra:
@swifty99 would you mind taking a look at this second PR, which restores color output, after #7508 makes the previous method of calculating color percentages impossible?

swifty99:
Looks good to me. Once it in dev I will test it with my sensors. Unfortunately the 
```
external_components:
  # equivalent shorthand for GitHub pull request
  - source: github://pr#7517
    components: [ tcs34725 ]

```
did not work for me, as py constants were changed too. The code looks good though.

RubenKelevra:
> Looks good to me. Once it in dev I will test it with my sensors. Unfortunately the
> 
> ```
> external_components:
>   # equivalent shorthand for GitHub pull request
>   - source: github://pr#7517
>     components: [ tcs34725 ]
> ```
> 
> did not work for me, as py constants were changed too. The code looks good though.

Yeah, I've patched it out for testing. You can test all 3 PRs by using this:

```yaml
external_components:
  - source: github://RubenKelevra/esphome_esphome@tcs34725_adjust_max_lux_for_glass_attenuation_v0.2
    components: [ tcs34725 ]  
```

RubenKelevra:
> Looks good to me. Once it in dev I will test it with my sensors.

Btw: Do you got more than one sensor, or just one single one?

I'm curious because I like to implement X, Y, Z CIE colors for them, to calculate more accurate Lux in the process, but I'm not sure how big the spread is across those sensors.

I've bought an AS7261 and would check my work or correct it, based on my two TCS34725 sensors, but if they are by chance very big outliers, it may be completely useless to calibrate with them.

So if you got like 5 or 10 of them and can expose them to similar light, I would be happy to see the irradiation results off of that, to get a better sense of the spread. :)

swifty99:
Hi,

I have 2 sensors. Although I did not had than side by side for a long time I would guess for color signals this should be reliable enough.

However, big disclaimer, take good care what you are "looking" at. For sunlight this works great. For artificial light not at all! I wrote a controller to adjust artificial lights to represent the same color distribution. It did work, the measurements were near identical. My eyes would not agree. Funky colors at least. Artificial and adjustable lights have a narrow spectrum an not necessarily aligned to the sensor. 
I wrote a blueprint in Home assistant to use RGBW lights as color controlled lights. I tried to describe the theory there.

RubenKelevra:
> However, big disclaimer, take good care what you are "looking" at. For sunlight this works great. For artificial light not at all! I wrote a controller to adjust artificial lights to represent the same color distribution. It did work, the measurements were near identical. My eyes would not agree. Funky colors at least. Artificial and adjustable lights have a narrow spectrum an not necessarily aligned to the sensor.
> I wrote a blueprint in Home assistant to use RGBW lights as color controlled lights. I tried to describe the theory there.

Wow, that's interesting!

Yeah, my interest regarding this sensor and the other one is specifically for the using them to measure artificial light sources and compare them.

But without this spec based irradiation output, I've implemented here, the color sensor was bound to output weird stuff, as the percentage values are not absolute and not corrected.

What this change here does is correct the individual channels, to respond equally to the same amount of light, but this is still not useable, as our eyes react differently to light for the different channels.

So the idea was to use a modern CIE human eye response and the responses of each channel, to average both against each other to get conversion factors, based on the human eye for color vision, to correct this.

I started working on a fix for the Lux calculation, because the current approach for Lux calculation falls apart on very blueish light, because of the negative coefficient for the blue channel. I tried a math hack for this, but it turned out to be a couple of percents off, which sounded too much for me.

So I tried a different approach:

The CIE published with 018:2019 a human eye response for each 1 nm channel, which I took and used to compare this to the corrected sensor response, to create a factor for how humans perceive the average of each color channel ([see here](https://github.com/RubenKelevra/tcs34725-irradiance-correction/tree/master/lux)).

The issue with this approach is, that the conversion to Lux is just meant to work with white light and D65 (6500 K), while we currently use here some inofficial D40 (4000 K) conversion. So technically our Lux values only work properly on 4000 K light, and everything else isn't really correct.

So instead of changing to to only working on 6500 K correctly, I thought about converting it to X, Y, Z colors and convert the illuminance signal of that back to Lux. This should then work more reliably over all types of colors, regardless of color temperature.

And sure, some artificial light sources, like CRI >=70 LEDs got a lot of blue light, we barely perceive as humans. But might give a wrong impression to the sensor in terms of energy - because it can not distinguish between different blue light types, so it's limited to assume a Gaussian distribution of the blue light, which isn't really correct.

That's also the reason cameras have a hard time with the white balance on those types of lighting conditions. But modern high quality CRI >=90/95 LEDs should normally work fine, as their blue spike is pretty muted.

swifty99:
Interesting. If you have a link to the project/result whatsoever I would like to follow it.
If you want to compare the same artificial lights, then no problem. If you got different ones, probably impossible.

My takeaway was:
- For reasonable color science a full blown spectrometer is needed. All DIY projects in this regard so far were off my limits or not appropriate for me
- For outside light measurement RGB sensors perform great
- For calibrating artificial lights I use a known reference and my eyes (no one else to be pleased)

I use this sensor to "copy" color temperature from outside to the inside. That works reasonably well.

RubenKelevra:
> Interesting. If you have a link to the project/result whatsoever I would like to follow it. 

I keep you posted. :) 

> If you want to compare the same artificial lights, then no problem. If you got different ones, probably impossible.
> 
> My takeaway was:
> 
> * For reasonable color science a full blown spectrometer is needed. All DIY projects in this regard so far were off my limits or not appropriate for me
> * For outside light measurement RGB sensors perform great
> * For calibrating artificial lights I use a known reference and my eyes (no one else to be pleased)

Yeah I don't want to use this sensor, but the AS7261 for that. It got more channels and is calibrated. It's like 40 bucks. 

It is designed to reflect human vision very closely, according to CIE1972 and will output calibrated CIE 1931 / CIE1972 colors with very high accuracy, and calibrated Lux as well as calibrated color temperature. 

Sure, it won't work with very bad LEDs, but the rest should work fine.

Im working on the support for ESPHome for it. 

> I use this sensor to "copy" color temperature from outside to the inside. That works reasonably well.

Yeah I tried that at well, works nicely. But the high brightness scenarios killed it a bit before ;)

PR #7533, opened 2024-10-03
[bang-bang] Remove ``assert()``

kbx81:

Removes an `assert()` from bang-bang -- this doesn't need to crash the device here. Changed to print a warning and move along.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/bang_bang/bang_bang_climate.cpp
+++ b/esphome/components/bang_bang/bang_bang_climate.cpp
@@ -157,8 +157,11 @@ void BangBangClimate::switch_to_action_(climate::ClimateAction action) {
     default:
       trig = nullptr;
   }
-  assert(trig != nullptr);
-  trig->trigger();
+  if (trig != nullptr) {
+    trig->trigger();
+  } else {
+    ESP_LOGW(TAG, "trig not set - unsupported action");
+  }
   this->action = action;
   this->prev_trigger_ = trig;
   this->publish_state();


esphome/components/bang_bang/bang_bang_climate.cpp
```diff
@@ -157,8 +157,11 @@ void BangBangClimate::switch_to_action_(climate::ClimateAction action) {
...
   }
-  assert(trig != nullptr);
-  trig->trigger();
+  if (trig != nullptr) {
+    trig->trigger();
+  } else {
+    ESP_LOGW(TAG, "trig not set - unsupported action");
+  }
   this->action = action;
   this->prev_trigger_ = trig;
```
guillempages, line 166: Wouldn't it be better to not override the previous trigger in the nullptr case? I.e. move the `this->prev_trigger_ = trig;` line to within the `if (trig != nullptr) {` block?

guillempages: Or maybe even the action as well?

guillempages: Ok, scratch that... The previous action will be stopped regardless of whether the new action is supported or not, and the prev_trigger is also always nullptr when this code is reached so actually setting prev_trigger to nullptr if trig is also null doesn't hurt as it is now.

jesserockz (approved):

PR #7536, opened 2024-10-03
Use "puremagic" instead of "magic" python module

### LINKED ISSUE #5896: Unable to compile S3 Voice Assistant on Windows
### The problem

I followed the instructions [HERE](https://esphome.io/guides/installing_esphome.html) to install Python and ESPHome on my Windows machine.

Then I tried to adopt a previously flashed ESP32 S3 Box 3.

The compiling process crashed with the message 'ImportError: failed to find libmagic.  Check your installation' and 'Component not found: image'

2024.5.5

Home Assistant Add-on

_No response_

ESP32

### Board

_No response_

_No response_

### Example YAML snippet

```yaml
substitutions:
  name: esp32-s3-box-3-redacted
  friendly_name: S3 Box A
packages:
  esphome.voice-assistant: github://esphome/firmware/wake-word-voice-assistant/esp32-s3-box-3.yaml@main
esphome:
  name: ${name}
  name_add_mac_suffix: false
api:
  encryption:
    key: 4ZtOEJ/bUR7/gcLp9vR33JprrzI8mO0YipdLhC72k+E=

wifi:
  ssid: !secret wifi_ssid
```

_No response_

### Additional information

I installed Python 3.12.4 amd-64 for Windows.

Further digging revealed a related post [HERE](https://github.com/esphome/issues/issues/5427) which might help fixing the problem.

guillempages:
Use the puremagic module to avoid extra dependencies, that sometimes cause problems in windows due to missing binary libraries. puremagic is a cross-platform pure python module that does not have binary dependencies.

The tradeoff is a lower number of detected files; for the current use-case should be enough though.

## Types of changes

- [x] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/5896 

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/image/__init__.py
+++ b/esphome/components/image/__init__.py
@@ -6,7 +6,7 @@ import logging
 from pathlib import Path
 import re
 
-from magic import Magic
+import puremagic
 
 from esphome import core, external_files
 import esphome.codegen as cg
@@ -237,8 +237,8 @@ CONFIG_SCHEMA = cv.All(font.validate_pillow_installed, IMAGE_SCHEMA)
 
 
 def load_svg_image(file: bytes, resize: tuple[int, int]):
-    # Local import only to allow "validate_pillow_installed" to run *before* importing it
-    # This import is only needed in case of SVG images; adding it
+    # Local imports only to allow "validate_pillow_installed" to run *before* importing it
+    # cairosvg is only needed in case of SVG images; adding it
     # to the top would force configurations not using SVG to also have it
     # installed for no reason.
     from cairosvg import svg2png
@@ -281,8 +281,7 @@ async def to_code(config):
     except Exception as e:
         raise core.EsphomeError(f"Could not load image file {path}: {e}")
 
-    mime = Magic(mime=True)
-    file_type = mime.from_buffer(file_contents)
+    file_type = puremagic.from_string(file_contents, mime=True)
 
     resize = config.get(CONF_RESIZE)
     if "svg" in file_type:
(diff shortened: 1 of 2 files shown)


clydebarrow (approved):

guillempages:
Now that I think about it; I'm not sure if this should be considered "breaking", since it adds (actually replaces) one python dependency (`puremagic`). I assume the docker containers and the "automagical" installations should use the requirements.txt file and thus work out of the box, but for manual installations/venv installations, this probably needs to be installed manually...

@jesserockz @kbx81 any thoughts about that?

clydebarrow:
What would it break? If the dependency is in requirements.txt it will be automatically added when building containers or installing via pip, and `script/setup` will take care of it when installing from git.

guillempages:
> What would it break? If the dependency is in requirements.txt it will be automatically added when building containers or installing via pip, and `script/setup` will take care of it when installing from git.

I don't know why, but I had to manually `pip install puremagic` on my venv before even being able to build an unrelated configuration... But I'm working off a git checkout of esphome that I already had, so I agree it's not the generic use case.

I just wanted to highlight, that there might be cases where it needs to be manually installed; if you say it's not an issue, it's fine for me.

clydebarrow:
> but I had to manually `pip install puremagic` on my venv

Existing venv perhaps, but creating a new one with `script/setup` it does install puremagic (just tested.) So I think we're good to go.

PR #7538, opened 2024-10-04
web_server: ensure fair network sharing + prevent lost state changes via deferred publish at high event load

nkinnan:

If the AsyncEventSource queue is full, instead save just enough information (8 bytes total per event including the queue entry) into a "deferred queue" for later publish.  This prevents lost state changes, and because the queue is FIFO no single sensor or set of sensors (or even log output) can dominate the network bandwidth.  All sensors will eventually have their latest state published to all event listeners.

This is part 1 for the web_server component.  A part 2 will be coming soon for the api component to give it the same functionality.

requires https://github.com/esphome/ESPAsyncWebServer/pull/43

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
  - source:
      type: git
      ref: "deferred_publish"
    components: [ web_server, web_server_base, web_server_idf ]
    refresh: 1s
```

## Checklist:


.gitignore
jesserockz, line 1: If you want to add this, please do so in its own PR.

esphome/components/event/event.cpp
jesserockz, line 1: Please also move the `event` changes to their own PR

esphome/components/web_server_idf/web_server_idf.cpp
```diff
@@ -8,6 +8,11 @@
 #include "esp_tls_crypto.h"
 
 #include "utils.h"
+
+#define USE_WEBSERVER  // jank linting doesn't replicate the build env
```
jesserockz: The problem is that the defines used for dev and CI need to be added to all of the supported sections instead of just here in `USE_ARDUINO`: https://github.com/esphome/esphome/blob/ee3ee3a63b784767ea907f2f52935db34f0e1267/esphome/core/defines.h#L89-L90

esphome/components/web_server_idf/__init__.py
```diff
@@ -8,6 +8,8 @@
     cv.only_with_esp_idf,
 )
 
+AUTO_LOAD = ["web_server"]
```
jesserockz, line 11: This should not be necessary

jesserockz (dismissed):

esphome/components/web_server_idf/web_server_idf.cpp
jesserockz: https://github.com/esphome/esphome/pull/7703

nkinnan: if you believe 7703 will resolve it, I can test removing this once that is committed

esphome/components/web_server_idf/__init__.py
nkinnan: web_server_idf now requires headers from web_server, and they will not be available unless web_server is loaded

esphome/components/event/event.cpp
nkinnan: the event change is required for the same reason as https://github.com/esphome/esphome/pull/7547 requires it and is identical, it supports this change (both PRs in fact) and is not an unrelated tweak hitching a ride

.gitignore
nkinnan: the change seems to have disappeared, did you push that yourself?

esphome/components/event/event.cpp
jesserockz: It should still be in its own PR as it is a change to something outside of the main focus of this PR.

esphome/components/web_server_idf/__init__.py
jesserockz: `web_server_idf` is loaded from `web_server`, so it is always present.

esphome/components/event/event.cpp
nkinnan: It is not outside the focus of this PR.  It allows this PR to work for Event.  For the same reason it allows https://github.com/esphome/esphome/pull/7547 to work for Event.  Which is why it's in both PRs.  It literally serves no purpose outside these two PRs. Which is why ssieb also pushed back against this comment on the API side https://github.com/esphome/esphome/pull/7547/files/8c7b9ef276543c31d81566d7c3d141313f076913#diff-7b561e90fd36d0b77851bf5be3c1cc9948a9590250d6e6cbbf907ad051b16664R26

esphome/components/web_server_idf/__init__.py
nkinnan: The CI fails if it is not present.

nkinnan:
Can we get some eyes on this?  I think it is an important and valuable change that improves the reliability of the communications protocol and prevents out of date values in HA + device webserver.  The design was suggested by ssieb.

esphome/components/web_server/web_server.cpp
```diff
@@ -1394,9 +1633,9 @@ std::string WebServer::valve_json(valve::Valve *obj, JsonDetail start_config) {
 
 #ifdef USE_ALARM_CONTROL_PANEL
 void WebServer::on_alarm_control_panel_update(alarm_control_panel::AlarmControlPanel *obj) {
-  if (this->events_.count() == 0)
+  if (this->events_.empty())
     return;
-  this->events_.send(this->alarm_control_panel_json(obj, obj->get_state(), DETAIL_STATE).c_str(), "state");
+  this->events_.deferrable_send_state(obj, "state", alarm_control_panel_state_json_generator);
```
nkinnan, line 1638: (for the CI error a few lines down)

1) not my change
2) it's a string type mismatch - value() returns an arduino "class String" whereas set_code takes a std::string, so removing the call to .c_str() call would break things
3) not sure how to make the CI error go away

nkinnan: nolinted

nkinnan:
I give up.  If you want this and https://github.com/esphome/esphome/pull/7547 to go in, ask the devs: https://discord.com/channels/429907082951524364/524177279270780928

clydebarrow (approved):

nkinnan:
Thank you Clyde!

AzonInc:
@nkinnan mind taking a look at https://github.com/esphome/issues/issues/6940
Highly appreciate your improvements <3

nkinnan:
> @nkinnan mind taking a look at [esphome/issues#6940](https://github.com/esphome/issues/issues/6940) Highly appreciate your improvements <3

Thanks for letting me know, I'm following up at https://github.com/esphome/issues/issues/6940 since it does appear to be related.

PR #7025, opened 2024-07-01
Add new component max6921

endym:
# What does this implement/fix?

This PR introduces the component MAX6921 [(datasheet)](https://www.analog.com/media/en/technical-documentation/data-sheets/MAX6921-MAX6931.pdf). This chip is a vacuum-fluorescent display (VFD) tube driver. For example it can be used to drive a IV18 display tube.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4074

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:
```
spi:
  clk_pin: GPIO18
  mosi_pin: GPIO23

display:
  - platform: max6921
    id: iv18
    load_pin: GPIO5
    out_pin_mapping:
        seg_to_out_map:
            seg_a_pin: 0
            seg_b_pin: 2
            seg_c_pin: 5
            seg_d_pin: 6
            seg_e_pin: 4
            seg_f_pin: 1
            seg_g_pin: 3
            seg_p_pin: 7
        pos_to_out_map:
            # IV18 positions starts at the right side!
            pos_0_pin: 15
            pos_1_pin: 14
            pos_2_pin: 13
            pos_3_pin: 16
            pos_4_pin: 12
            pos_5_pin: 17
            pos_6_pin: 11
            pos_7_pin: 18
    lambda: |-
      it.print("01234567");
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs): esphome/esphome-docs#4074

PR #7545, opened 2024-10-05
Add new component max6921 (2nd try)

endym:

This PR introduces the component MAX6921 [(datasheet)](https://www.analog.com/media/en/technical-documentation/data-sheets/MAX6921-MAX6931.pdf). This chip is a vacuum-fluorescent display (VFD) tube driver. For example it can be used to drive a IV18 display tube.

Note: My initial [PR](https://github.com/esphome/esphome/pull/7025) was closed by an error of mine. That's why I created this PR once again. Sorry for that!

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4074

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```
spi:
  clk_pin: GPIO18
  mosi_pin: GPIO23

display:
  - platform: max6921
    id: iv18
    load_pin: GPIO5
    out_pin_mapping:
        seg_to_out_map:
            seg_a_pin: 0
            seg_b_pin: 2
            seg_c_pin: 5
            seg_d_pin: 6
            seg_e_pin: 4
            seg_f_pin: 1
            seg_g_pin: 3
            seg_p_pin: 7
        pos_to_out_map:
            # IV18 positions starts at the right side!
            pos_0_pin: 15
            pos_1_pin: 14
            pos_2_pin: 13
            pos_3_pin: 16
            pos_4_pin: 12
            pos_5_pin: 17
            pos_6_pin: 11
            pos_7_pin: 18
    lambda: |-
      it.print("01234567");
```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs): https://github.com/esphome/esphome-docs/pull/4074


jesserockz:
Not sure why GH would close it.
But also in general, don't force push a branch that has a PR with reviews unless requested by the reviewer.

jesserockz:
> Currently the load pin is handled similarly to a SPI CS pin. However, the load pin has a completely different task with a quite flexible handling. For example, a HIGH pulse could also be made after the data transfer (see figure 4 in the data manual).
I would prefer to leave the load pin configuration as it is.

But, this PR does not do that, it only ever sets the pin to high at the beginning of transfer, and low after transfer. (It also sets the load pin to high during setup until the first data is sent? Is that correct as the external device really should just sit there waiting until an actual transaction happens)

If at some point you want to modify the component to do what you mention, you can still use `this->cs_` directly in the component, it's not taken away.

endym:
> But, this PR does not do that, it only ever sets the pin to high at the beginning of transfer, and low after transfer. (It also sets the load pin to high during setup until the first data is sent? Is that correct as the external device really should just sit there waiting until an actual transaction happens)

No, the load pin is initiated to low at setup which means latching the current state of shift register output (low after power-on). As I mentioned, the load pin task is not comparable with the cs pin task.


However, is that your idea to use the SPI cs pin functionality:

- Rename the YAML configuration variable `load_pin` to `cs_pin`

```
display:
  - platform: max6921
    id: iv18
    cs_pin: GPIO5  # define MAX6921 LOAD pin
    ...
```

- Adapt code

```
void MAX6921Component::setup() {
  ...
  this->spi_setup();
  this->enable();  // set LOAD to low to disable update of output latch
  ...
}
```
```
void HOT MAX6921Component::write_data(uint8_t *ptr, size_t length) {
  ...
  this->enable();  // set LOAD to low
  ...
  this->transfer_array(...);
  this->disable();  // set LOAD to high to update output latch
}
```

jesserockz:
> However, is that your idea to use the SPI cs pin functionality

Yes

Why are you holding the load pin low from setup exactly?

endym:
> Why are you holding the load pin low from setup exactly?

Low means that the MAX6921 output latch will not be updated from shift register.

endym:
I've tried to implement the discussed changes (replace LOAD with CS pin), but it doesn't work. The CS methods `enable()` and `disable()` contains a lot of SPI-specific logic (incl. OS sync methods like mutex) which on the one hand is completely unnecessary (and worsens the performance) and on the other breaks the wifi functionality somehow. I haven't gone deep into debugging (because it's not worth the effort IHMO), but I assume that it has to do with the fact that the data transfer runs asynchronous in the context of a pinned thread.
Therefore I can only point out once again that the LOAD pin function really has nothing to do with the SPI CS function. It just so happens that the LOAD pin is switched in the same way.

endym:
@jesserockz How should we proceed now?

endym:
@jesserockz Any chance that this PR will be merged?

esphome/components/max6921/max6921.cpp
```diff
@@ -0,0 +1,98 @@
...
+ *          DOUT | x  | x  | x  | x  | 19 | 18 | ... | 1 | 0
+ */
+void HOT MAX6921Component::write_data(uint8_t *ptr, size_t length) {
+  uint8_t data[3];
+  static bool first_call_logged = false;
+
+  assert(length == 3);
+  memcpy(data, ptr, sizeof(data));  // make copy of data, because transfer buffer will be overwritten with SPI answer
+  if (!first_call_logged)
+    ESP_LOGVV(TAG, "SPI(%u): 0x%02x%02x%02x", length, data[0], data[1], data[2]);
```
jesserockz, line 54:
```suggestion
  if (!first_call_logged) {
    ESP_LOGVV(TAG, "SPI(%u): 0x%02x%02x%02x", length, data[0], data[1], data[2]);
  }
```

esphome/components/max6921/max6921.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace esphome {
+namespace max6921 {
+
+static const char *const TAG = "max6921";
+
+float MAX6921Component::get_setup_priority() const { return setup_priority::HARDWARE; }
+
+void MAX6921Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up MAX6921...");
+  // global_max6921 = this;
```
jesserockz, line 21: No commented out code

esphome/components/max6921/max6921.cpp
```diff
@@ -0,0 +1,98 @@
...
+  uint8_t data[3];
+  static bool first_call_logged = false;
+
+  assert(length == 3);
+  memcpy(data, ptr, sizeof(data));  // make copy of data, because transfer buffer will be overwritten with SPI answer
+  if (!first_call_logged)
+    ESP_LOGVV(TAG, "SPI(%u): 0x%02x%02x%02x", length, data[0], data[1], data[2]);
+  first_call_logged = true;
+  this->disable_load_();  // set LOAD to low to disable update of output latch during data transfer
+  this->transfer_array(data, sizeof(data));
```
jesserockz, line 57: Please still wrap this in `this->enable()` / `this->disable()` even when cs pin is not being used as the SPI bus still needs to be configured each time for this specific components settings.

esphome/components/max6921/max6921.h
```diff
@@ -0,0 +1,53 @@
...
+class MAX6921Component;
+class Max6921Display;
+
+using max6921_writer_t = std::function<void(MAX6921Component &)>;
+
+class MAX6921Component : public PollingComponent,
+                         public spi::SPIDevice<spi::BIT_ORDER_MSB_FIRST, spi::CLOCK_POLARITY_LOW,
+                                               spi::CLOCK_PHASE_LEADING, spi::DATA_RATE_4MHZ> {
+ public:
+  std::unique_ptr<Max6921Display> display_;
```
jesserockz, line 25: This should be protected

esphome/components/max6921/display.cpp
```diff
@@ -0,0 +1,953 @@
...
+    }
+    ESP_LOGVV(TAG, "%s(): segment data: 0x%06" PRIx32, __func__, out_data);
+#if 0
+    // At the moment an unsupport character is equal to blank (' ').
+    // To distinguish an unsupported character from blank we would need to
+    // increase font data type from uint8_t to uint16_t!
+    if (out_data == SEG_UNSUPPORTED_CHAR) {
+      ESP_LOGW(TAG, "Encountered character '%c (0x%02x)' with no display representation!", *vi_text, *vi_text);
+    }
+#endif
```
jesserockz, line 586: This is equivalent to commented out code, which we do not allow.

esphome/components/max6921/max6921.cpp
jesserockz, line 1: I don't understand the need to split this into 2 separate classes when you are just calling function between them back and forward. It might as well just be a single class.

tests/components/max6921/test.esp32-ard.yaml
jesserockz, line 1: Please deduplicate these test files into a `common.yaml` file that is included into the rest.

esphome/components/max6921/display.cpp
```diff
@@ -0,0 +1,953 @@
...
+  this->init_font_();
+
+  // display output buffer...
+  this->out_buf_size_ = this->num_digits_ * 3;
+  this->out_buf_ = new uint8_t[this->out_buf_size_];  // NOLINT
+  memset(this->out_buf_, 0, this->out_buf_size_);
+
+  // find smallest segment DOUT number...
+  this->seg_out_smallest_ = 19;
+  for (unsigned char i : this->seg_to_out_map_) {
```
jesserockz, line 142: We do not use `unsigned char` 
```suggestion
  for (uint8_t i : this->seg_to_out_map_) {
```

esphome/components/max6921/display.cpp
```diff
@@ -0,0 +1,953 @@
...
+  this->pos_to_out_map_ = pos_to_out_map;
+  this->num_digits_ = pos_to_out_map.size();
+  ESP_LOGCONFIG(TAG, "Display digits: %u", this->num_digits_);
+
+  // setup font...
+  this->init_font_();
+
+  // display output buffer...
+  this->out_buf_size_ = this->num_digits_ * 3;
+  this->out_buf_ = new uint8_t[this->out_buf_size_];  // NOLINT
```
jesserockz, line 137: Please use the `RAMAllocator` or a std::vector for this instead of bypassing the linter with NOLINT

esphome/components/max6921/display.cpp
```diff
@@ -0,0 +1,953 @@
...
+void Max6921Display::dump_config() {
+  char seg_name[3];
+
+  // display segment to DOUTx mapping...
+  for (uint i = 0; i < this->seg_to_out_map_.size(); i++) {
+    if (i < 7) {
+      seg_name[0] = 'a' + i;
+      seg_name[1] = 0;
+    } else
+      strncpy(seg_name, "dp", sizeof(seg_name));
```
jesserockz, line 214: All branches should have braces
```suggestion
    } else {
      strncpy(seg_name, "dp", sizeof(seg_name));
    }
```

esphome/components/max6921/display.cpp
```diff
@@ -0,0 +1,953 @@
...
+ * @param c character to check
+ *
+ * @return true, if character activates the point segment only, otherwise false
+ */
+bool Max6921Display::is_point_seg_only(char c) { return ((c == ',') || (c == '.')); }
+
+void Max6921Display::init_font_() {
+  uint8_t seg_data;
+
+  this->ascii_out_data_ = new uint8_t[ARRAY_ELEM_COUNT(ASCII_TO_SEG)];  // NOLINT
```
jesserockz, line 235: Same as above

esphome/components/max6921/display.h
```diff
@@ -0,0 +1,127 @@
...
+  DISP_MODE_LAST_ENUM
+};
+
+enum TextAlignT { TEXT_ALIGN_LEFT, TEXT_ALIGN_CENTER, TEXT_ALIGN_RIGHT, TEXT_ALIGN_LAST_ENUM };
+
+enum TextEffectT {
+  TEXT_EFFECT_NONE,         // show text at given position, cut if too long
+  TEXT_EFFECT_BLINK,        // blink
+  TEXT_EFFECT_SCROLL_LEFT,  // scroll left, start with 1st char at right position
+  // TEXT_EFFECT_SCROLL_RIGHT, // scroll right, start with last char at left position
```
jesserockz, line 30: Remove or utilise

jesserockz (changes requested):
There are many PRs open to review and not so many reviewers. Pinging (and sending emails) will only get a PR ignored in the future.

If you followed all of the guidelines on https://esphome.io/guides/contributing, then it would be less of a task for someone to review your PR instead of going over minor things that are listed there.

Edit: task, not talk

endym:
> There are many PRs open to review and not so many reviewers. Pinging (and sending emails) will only get a PR ignored in the future.

Imagine you create a PR (first time for this project) and after the review has already started ... 6 weeks no response to any question. What would you think about? A short message like _"At the moment I'm very busy. Come back to you asap."_ would have been sufficient.

> If you followed all of the guidelines on https://esphome.io/guides/contributing, then it would be less of a talk for someone to review your PR instead of going over minor things that are listed there.

You are totally right! If people would do everything 100% right, we would have to talk less. 
But honestly, it's unrealistic to expect all contributors to know the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) by heart. If you really want 100% compliant code with no review effort, then you have to extend the automatic checks accordingly.

endym:
It was a lot of work to bring this PR to its current state - on the one hand because it was my first contribution to esphome and on the other hand because hard time requirements are not easy to implement in the esphome environment.
The new required review changes would again require a lot of effort, which I am no longer willing to put in. This is also partly due to the fact that I did not feel that my contribution was of interest during the entire review process.
However, the current component state works for me without any problems. But in future I would no longer realize such a development as an esphome component, but as a MQTT device. The advantages for me would be that it can be used more flexibly (not only in HA) and the development itself would be more independent and therefore easier.
I nevertheless think esphome is a great thing that makes it easy - and not just for non-programmers - to create your own smart home devices.
endym

PR #7548, opened 2024-10-06
[pulse_counter_ulp][const] Add pulse counter ulp

### LINKED ISSUE #826: ESP32 ULP Coprocessor

Possibility to set the counter of impulses on the coprocessor. And the main core of ESP32 is sleeping.

Pulse counter for metering water consumption. Less battery consumption!

**Additional context**

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/ulp.html
https://github.com/tanakamasayuki/UlpDebug
Real Project Whater Counter:
https://github.com/OloloevReal/Waterius32/tree/master/src

brisk0:

This introduces a new type of pulse counter that uses the ESP32's Ultra-Low Power (ULP) coprocessor to keep counting through deep sleep with the goal of enabling simple battery operated sensors in esphome (this is specifically to drive my own electricity meter reader, which is ironically difficult to power otherwise).

The sensor tracks how long the ULP program takes to run while it is awake, so that it can estimate the time passed on wake-up by maintaining a count of times the ULP program ran. All variables that need to persist are stored by the ULP program (which I believe uses RTC_SLOW_MEM).

The Pulse Counter ULP sensor is (perhaps obviously) based on the Pulse Counter sensor. I was hoping it would be possible to maintain it as a configuration of Pulse Counter, but there were a few difficulties that pushed away from that idea:

- There would have to be breaking changes to the configuration, as Pulse Counter assumes only a single variation (`has_pcnt`) and assumes all configuration is valid for all storage types.
- The relationship between the "storage" part of the sensor and the main part of the sensor has to be much tighter in the ULP version, because anything that persists must be known to the ULP program, including time information. This would have meant substantially rewriting the component at the risk of introducing new bugs, and a lot of conditional execution only relevant to the ULP variant.
- A few hardware abstractions had to be broken through to make this work, meaning that the ULP sensor cannot compile on e.g. an ESP8266, which would lead to a lot of conditional compilation.

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/feature-requests/issues/826

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4307

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

substitutions:
  device_name: deepsleep-test
  friendly_name: Deep Sleep
  project_version: "1.0"
  device_description: "Test deep sleep pulse counter component"

  # Define the GPIO pins
  pulse_pin: GPIO13

esphome:
  name: '${device_name}'
  comment: '${device_description}'
  project:
    name: "brisk.deepsleep-test"
    version: "${project_version}"

esp32:
  board: esp32dev
  framework:
    type: esp-idf
    version: recommended

# WiFi credentials #
wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_key
  fast_connect: true

# Enable logging
logger:
  level: VERBOSE

deep_sleep:
  run_duration: 25s
  sleep_duration: 20s

sensor:
  - platform: pulse_counter_ulp
    name: ${friendly_name} - Number of Pulses
    pin:
      number: ${pulse_pin}
      inverted: true
      mode:
        pullup: true
        input: true
    update_interval: 5s


```

## Checklist:


zane-lazare:
Is this going anywhere? I shelved my own pulse meter due to not being able to run deep sleep in September, then came across this last month! Pretty keen to see it happen

brisk0:
> Is this going anywhere? I shelved my own pulse meter due to not being able to run deep sleep in September, then came across this last month! Pretty keen to see it happen

This is fully functional as far as I'm concerned so you can pull this branch and give it a go!

Per https://esphome.io/guides/contributing.html#can-i-help-review-prs this could help it get merged if you report back.

zane-lazare:
Is this the right code to include the pull request?

````
external_components:
  - source: github://pr#7548
    components: [ pulse_counter_ulp ]
````

 Haven't tried to utilise external components before, and compilation is throwing up a CMake error.

<details><summary>Log</summary>
<p>

```
Processing esp-glow (board: esp32-c3-devkitm-1; framework: espidf; platform: platformio/espressif32@5.4.0)
--------------------------------------------------------------------------------
HARDWARE: ESP32C3 160MHz, 320KB RAM, 4MB Flash
 - framework-espidf @ 3.40408.0 (4.4.8) 
 - tool-cmake @ 3.16.4 
 - tool-ninja @ 1.7.1 
 - toolchain-esp32ulp @ 2.35.0-20220830 
 - toolchain-riscv32-esp @ 8.4.0+2021r2-patch5
Reading CMake configuration...
-- Building ESP-IDF components for target esp32c3
-- Project sdkconfig file /data/build/esp-glow/sdkconfig.esp-glow
-- App "esp-glow" version: 2024.12.0-dev
-- Adding linker script /data/build/esp-glow/.pioenvs/esp-glow/esp-idf/esp_system/ld/memory.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_system/ld/esp32c3/sections.ld.in
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_rom/esp32c3/ld/esp32c3.rom.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_rom/esp32c3/ld/esp32c3.rom.api.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_rom/esp32c3/ld/esp32c3.rom.libgcc.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_rom/esp32c3/ld/esp32c3.rom.newlib.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_rom/esp32c3/ld/esp32c3.rom.version.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_rom/esp32c3/ld/esp32c3.rom.newlib-time.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/esp_rom/esp32c3/ld/esp32c3.rom.eco3.ld
-- Adding linker script /data/cache/platformio/packages/framework-espidf/components/soc/esp32c3/ld/esp32c3.peripherals.ld
-- IPV6 support is disabled so the asio component will not be built
-- IPV6 support is disabled so the coap component will not be built
-- Configuring incomplete, errors occurred!
See also "/data/build/esp-glow/.pioenvs/esp-glow/CMakeFiles/CMakeOutput.log".

fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
fatal: not a git repository: /data/cache/platformio/packages/framework-espidf/components/openthread/openthread/../../../.git/modules/components/openthread/openthread
CMake Error at src/CMakeLists.txt:25 (ulp_embed_binary):
  Unknown CMake command "ulp_embed_binary".
```
</p>
</details> 

Full .yaml file for context:

<details><summary>glow.yaml</summary>
<p>

```
external_components:
  - source: github://pr#7548
    components: [ pulse_counter_ulp ]
    

substitutions:
  device_name: esp-glow
  friendly_name: Glow

  pulse_pin: GPIO08
  scl_pin: GPIO10
  sda_pin: GPIO09
  batt_pin: GPIO04

esphome:
  name: '${device_name}'
  friendly_name: '${friendly_name}'

esp32:
  board: esp32-c3-devkitm-1
  framework:
    type: esp-idf
    version: recommended
        
# Enable logging
logger:
  level: VERBOSE

# Enable Home Assistant API
api:
  encryption:
    key: "dKLev9iIxenl04qCFIU61zvwIZ3PK/HjS4ldtSxlpNg="

ota:
  - platform: esphome
    password: "5a94f788707834a5dc11abbb6e9a415e"

wifi:
  ssid: !secret wifi_ssid
  fast_connect: True

  ap:
    ssid: "Glow Fallback Hotspot"
    password: "aXFsvZNIXgdu"

#captive_portal:

deep_sleep:
  run_duration: 25s
  sleep_duration: 20s

i2c:
  sda: ${sda_pin}
  scl: ${scl_pin}
  scan: true

sensor:
  - platform: pulse_counter_ulp
    name: ${friendly_name} - Number of Pulses
    pin:
      number: ${pulse_pin}
      inverted: true
      mode:
        pullup: true
        input: true
    update_interval: 5s

  - platform: bme280_i2c
    temperature:
      name: "BME280 Temperature"
    pressure:
      name: "BME280 Pressure"
    humidity:
      name: "BME280 Humidity"

  - platform: adc
    pin: ${batt_pin}
    name: "Battery Voltage"
    update_interval: 10s
    accuracy_decimals: 2
    filters:
      - multiply: 2.0  # Adjust this based on your voltage divider ratio

```

</p>
</details>

**Comment by brisk0** [comment]
@zane-lazare 
I'm not sure that this is what's going wrong, and I'll try replicating with the external component system, but I'm sorry to say this component won't work with the ESP32C3 as there is no ULP coprocessor in the C series

**Comment by zane-lazare** [comment]
> @zane-lazare I'm not sure that this is what's going wrong, and I'll try replicating with the external component system, but I'm sorry to say this component won't work with the ESP32C3 as there is no ULP coprocessor in the C series

Seems I had mucked up at my end, needed to run clean build files before trying to use ESP-IDF (only been using arduino). Compiled without issue after that.... The lack of co-processor is definately an oversight on my behalf though! I'll get it running on another device and report back!

**Comment by lboue** [comment]
Build fails for me with ESP32-S3 board:

[ulp-s3.yaml.txt](https://github.com/user-attachments/files/17775080/ulp-s3.yaml.txt)

```yaml
esp32:
  board: esp32-s3-devkitc-1
  framework:
    type: esp-idf
    version: recommended
```

```
[1/7] Generating ulp_fsm.ld
[2/7] Generating pulse_cnt.ulp.S
[3/7] Building ASM object CMakeFiles/ulp_main.dir/pulse_cnt.ulp.S.obj
[4/7] Linking ASM executable ulp_main
FAILED: ulp_main 
: && esp32ulp-elf-ld                                   -A elf32-esp32s2ulp -nostdlib CMakeFiles/ulp_main.dir/pulse_cnt.ulp.S.obj -o ulp_main  -T"/data/build/ulp-s3/.pioenvs/ulp-s3/esp-idf/src/ulp_main/ulp_fsm.ld"  -Map="/data/build/ulp-s3/.pioenvs/ulp-s3/esp-idf/src/ulp_main/ulp_main.map" && :
esp32ulp-elf-ld: invalid length for memory region ram
ninja: build stopped: subcommand failed.
*** [.pioenvs/ulp-s3/esp-idf/src/ulp_main/ulp_main.h] Error 1
========================= [FAILED] Took 53.76 seconds =========================
```


[logs_ulp-s3_compile.txt](https://github.com/user-attachments/files/17775065/logs_ulp-s3_compile.txt)

zane-lazare:
Seems all compiles and runs fine on my ESP32 D1 mini. Any tips for how long this can record pulses before needing to wake?

brisk0:
@lboue looks like I need to add S3 specific sdkconfig options to turn on the ULP Coprocessor (CONFIG_ESP32S3_ULP_COPROC_ENABLE). I don't suppose you have any insight into why S3 is special? I'm a bit worried there might be other device specific options I'm still missing once I've added this one.

@zane-lazare it depends on the max rate of your pulse. Pulses are counted in a uint16 so you can Count up to 65535 pulses

brisk0:
@zane-lazare Actually, the larger restriction will be the run-counter, which is also a uint16_t.
So the max time between reads would be approximately sleep_duration * 65535. At the default 20ms sleep_duration you'll get about 21.85 minutes. I would shave a bit off to accommodate startup time and the time between the last read and the start of the deep_sleep.

I'll make sure the documentation covers the restriction explicitly.

brisk0:
@lboue I've pushed additional sdkconfig parameters and a unit test that appear to test and remedy the failure when building for an ESP32-S3. Please pull and try again at your convenience.

brisk0:
@zane-lazare I have updated the documentation to address maximum deep sleep time. Please let me know if it is sufficiently clear:

https://deploy-preview-4307--esphome.netlify.app/components/sensor/pulse_counter_ulp

Out of curiosity, is the limit problematic for your use case? It could probably be extended in future if there's motivation.

lboue:
> @lboue I've pushed additional sdkconfig parameters and a unit test that appear to test and remedy the failure when building for an ESP32-S3. Please pull and try again at your convenience.

Thanks. Build is OK now.

zane-lazare:
> @zane-lazare I have updated the documentation to address maximum deep sleep time. Please let me know if it is sufficiently clear:
> 
> https://deploy-preview-4307--esphome.netlify.app/components/sensor/pulse_counter_ulp
> 
> Out of curiosity, is the limit problematic for your use case? It could probably be extended in future if there's motivation.

Nah, not particularly. I'm using this device as a solar powered Glow, plus weather station. Getting an update every 5 minutes or so is probably adequate for power usage monitoring, and for specific cases I'll imploy a "high res" update switch or something.

I've got the code compiled fine. It isn't reporting pulses though. This may be because I've put the photoresistor on the ADC0 pin, and didn't realize until I put it back outside. I'll have a fiddle tonight and see if I can get it going

zane-lazare:
Not having much luck, I've included my yaml below. Sensor posts/updates intermittently, sometimes every couple of minutes, sometimes 30, sometimes not at all. For some reason the value is always 75.x

Update: seem to be getting semi-regular updates, always reported as 75.x,  seems the sensor is only updating every 15mins or 30 mins, and reporting an average pulse/min? Screenshot attached


```
external_components:
  - source: github://pr#7548
    components: [ pulse_counter_ulp ]
    
substitutions:
  pulse_pin: GPIO34
  scl_pin: GPIO22
  sda_pin: GPIO21
  batt_pin: GPIO033

esphome:
  name: '${device_name}'
  friendly_name: '${friendly_name}'
  on_boot:
    priority: 300.0
    then: 
      - delay: 15s
      - if:
          condition:
            not:
              mqtt.connected:
          then:
           - logger.log: "MQTT connection failed. Sleeping for 30 seconds"
           - deep_sleep.enter:
               id: deep_sleep_1
               sleep_duration: 30s
          
esp32:
  board: esp32dev
  framework:
    type: esp-idf
    version: recommended

# Enable logging
logger:
  level: VERBOSE

wifi:
  ssid: !secret wifi_ssid
  fast_connect: True

mqtt:
  broker: 192.168.1.167
  username: !secret mqtt_user
  birth_message: #Empty
  will_message: #Empty
  on_message:
    - topic: esphome/ota_mode
      payload: 'ON'
      then:
        - logger.log: "OTA mode is ENABLED"
        - deep_sleep.prevent: deep_sleep_1
    - topic: esphome/ota_mode
      payload: 'OFF'
      then:
        - logger.log: "OTA mode is DISABLED"
        - deep_sleep.allow: deep_sleep_1


deep_sleep:
  run_duration: 10s
  sleep_duration: 50s
  id: deep_sleep_1

sensor:
  - platform: pulse_counter_ulp
    name: ${friendly_name} - Number of Pulses
    pin: 
      number: ${pulse_pin}
      inverted: true
      mode:
        input: true
    update_interval: 5s
    accuracy_decimals: 0
```

SmartSelect_20241118_214713_Home Assistant

brisk0:
@zane-lazare I haven't had a chance to try to replicate but I should be able to in the next couple of days. If you have a chance in the meantime, it might help fault-find if you're able to test whether the behaviour is as expected when not using deep-sleep, or when using the original pulse-counter (without deep-sleep)

zane-lazare:
> @zane-lazare I haven't had a chance to try to replicate but I should be able to in the next couple of days. If you have a chance in the meantime, it might help fault-find if you're able to test whether the behaviour is as expected when not using deep-sleep, or when using the original pulse-counter (without deep-sleep)

How embarrassing. It works! Looks like I was chasing a bloody damaged wire on the data pin on the diode. Original code I posted works a charm. Epic work dude!

michaelt88:
This looks great!
I will try it soon

michaelt88:
I have tried running this

I am using this yaml config
```
deep_sleep:
  run_duration: 25s
  sleep_duration: 20s

sensor:
  - platform: pulse_counter_ulp
    name: Number of Pulses
    pin:
      number: 14
      inverted: true
      mode:
        pullup: true
        input: true
    update_interval: 5s
```

My observations are as follows:
* When the unit is awake it is counting the pulses and reporting every 5s
* The unit goes to sleep, I continue to pulse pin 14 while it is in sleep
* When the unit wakes up and reconnects, I don't see any pulses counted while the unit was in deep sleep

I'm not 100% sure how it is supposed to work, but my expectation was that while the unit is in deep sleep it will continue to count the pulses, and then when it wakes up it will send the total number of pulses that occurred while in deep sleep.

Please let me know if I'm doing something wrong?

brisk0:
@michaelt88 it sends a rate rather than a total count, but otherwise that's correct, and your yaml looks fine.

What output are you seeing?

michaelt88:
> @michaelt88 it sends a rate rather than a total count, but otherwise that's correct, and your yaml looks fine.
> 
> What output are you seeing?

image

I have attached logs - similar test - pulse when unit awake, pulse when unit asleep, no pulses captured when unit wakes up
[logs_esphome-web-805598_logs.txt](https://github.com/user-attachments/files/18030460/logs_esphome-web-805598_logs.txt)

It is configured for 30s awake, 30s sleep, but it seems to be offline for a lot longer than 30s. One theory is that the unit wakes up, but takes a while to connect to wifi, and the pulse it sends in that time is lost?

Having the "total" counter like the regular pulse_counter function would also be handy.

brisk0:
@michaelt88 I have successfully replicated your issue. I was not seeing that behaviour before as I have been using "fast_connect" on the wifi configuration, which seems to set-up fast enough on my network for Home Assistant to receive the first message publication.

brisk0:
@michaelt88 Regarding the "total" counter, I had planned to include this originally. Pulse_counter implements this through an automation, which when I started working on this branch was able to keep count through deep sleep.
Unfortunately, I believe there has been a change to how automations store state since then as more recent experiments show the "total sensor" automation failing to retain state through deep sleep (I have not investigated this particularly thoroughly). So I can't use the same implementation. I'm sure it's doable, but IMO out of scope of this PR

brisk0:
@michaelt88 I have pushed a commit changing the setup priority. It should now wait until wifi is set up before pulse_counter_ulp will set up. Let me know if this does or does not resolve what you're seeing.

michaelt88:
> @michaelt88 I have pushed a commit changing the setup priority. It should now wait until wifi is set up before pulse_counter_ulp will set up. Let me know if this does or does not resolve what you're seeing.

@brisk0 yes that seems to have fixed it

In the below example I pulsed the pin while it was sleeping, then the pulses were captured when the device woke up
image

> @michaelt88 I have successfully replicated your issue. I was not seeing that behaviour before as I have been using "fast_connect" on the wifi configuration, which seems to set-up fast enough on my network for Home Assistant to receive the first message publication.

Yes that would probably explain it. I am new to the ESP Home system so was not familiar with fast_connect. Good to know though as I was puzzled why it was taking so long to connect.

Although if I enable fast_connect, now I seem to have the issue of pulses not being counted during deep sleep..

michaelt88:
> @michaelt88 Regarding the "total" counter, I had planned to include this originally. Pulse_counter implements this through an automation, which when I started working on this branch was able to keep count through deep sleep. Unfortunately, I believe there has been a change to how automations store state since then as more recent experiments show the "total sensor" automation failing to retain state through deep sleep (I have not investigated this particularly thoroughly). So I can't use the same implementation. I'm sure it's doable, but IMO out of scope of this PR

Not sure if it helps, but with Arduino ESP development environment, persistent data can be stored through RTC memory during deep sleep as follows
`
RTC_DATA_ATTR uint32_t pulse_count = 0;   // Persistent counter
`

It seems like keeping a count of total pulses, and then transmitting the total count would be a more robust way to avoid issues if some transmissions are not sent.

Then it could also send the increase and pulses, the elapsed time, and the pulse rate between each successful transmission.

m0uH:
I tried the proposed changes as I'm interested in using it for a water meter.

Thanks to the hint of @michaelt88 I was able to implement an optional total pulse count sensor (like in the original pulse count component). Using only the rate was quite unreliable when using deep sleep.

@brisk0 I'd like to propose adding my changes to this PR, probably best adding a PR to the branch of your fork? I also do not know where to adapt the documentation for the additional 'total' sensor.

As I need to use an active sensor module, I'd also like to know, if others would be interested in optionally support for enabling a pin for checking the pulse state. I am using a LED to detect a rotating metal and would be interested in reducing the power consumption of this LED by only powering it while checking the state.

michaelt88:
> Thanks to the hint of @michaelt88 I was able to implement an optional total pulse count sensor (like in the original pulse count component). Using only the rate was quite unreliable when using deep sleep.

Nice work! That would be great if this change could be added to the PR

> As I need to use an active sensor module, I'd also like to know, if others would be interested in optionally support for enabling a pin for checking the pulse state. I am using a LED to detect a rotating metal and would be interested in reducing the power consumption of this LED by only powering it while checking the state.

Interesting! I am currently using a prox sensor to monitor the rotating metal on the water meter, this draws around 1-2 mA. Pulsing an LED and using an LDR to detect the rotating metal sounds like a great solution that would allow you to dramatically lower power consumption. Are you able to share any details on how you have physically setup the LED & LDR to the water meter?

m0uH:
I've now also added an additional option to wake up the main SoC after a user defined number of detected edges. By this one can use a much longer deep sleep duration, which should drastically improve average power consumption in case pulses are generated only from time to time.

E.g. in the water meter use case (assuming water is not running all the time) one can set the deep sleep duration to 1800s and let the ULP wake up the main SoC after detecting a single pulse.

> Interesting! I am currently using a prox sensor to monitor the rotating metal on the water meter, this draws around 1-2 mA. Pulsing an LED and using an LDR to detect the rotating metal sounds like a great solution that would allow you to dramatically lower power consumption. Are you able to share any details on how you have physically setup the LED & LDR to the water meter?

I did use a standard white LED, increased the resistor such that only 0.3mA are consumed by the LED. For detection I did use a pre-soldered sensor module with a photoresistor (like https://www.amazon.de/dp/B07P5Z2XD4) , which provides an analog and a digital signal output. The digital signal can be calibrated easily with a screw-driver. Only downside is that it has two quite brite LEDs. The two status LEDs consume together >2mA, such that the whole module is at about 3mA.

brisk0:
@m0uH that sounds fantastic, please feel free to make a PR to this branch of my fork

https://github.com/brisk0/esphome/tree/pulse-counter-ulp

Likewise the associated branch for documentation changes

https://github.com/brisk0/esphome-docs/tree/pulse-counter-ulp

(PR https://github.com/esphome/esphome-docs/pull/4307)

If the total count functionality is very similar to pulse-counter, cribbing from the documentation page for that should be fine I'm sure.

m0uH:
@brisk0 I've added my changes in https://github.com/brisk0/esphome/pull/1 and https://github.com/brisk0/esphome-docs/pull/1.

While I also added the automation to update the total sensor via esphome API, I did not add it to the documentation, as it is currently not that helpful, as it would need to be run after each reset.

Unfortunately I did not manage to get how the sensor value is kept between resets for the non-ULP pulse-counter (if at all?). May be someone more experienced in esphome development can assist here.

May be one could write it to flash, but in case the sleep interval is short, this might wear out the flash memory quite fast.

For now one needs to add up the sensor values using e.g. an utility-meter.

michaelt88:
> While I also added the automation to update the total sensor via esphome API, I did not add it to the documentation, as it is currently not that helpful, as it would need to be run after each reset.
>
> Unfortunately I did not manage to get how the sensor value is kept between resets for the non-ULP pulse-counter (if at all?). May be someone more experienced in esphome development can assist here.
> 
> May be one could write it to flash, but in case the sleep interval is short, this might wear out the flash memory quite fast.

@m0uH my understanding is the RTC slow memory is not subject to write limits.

> The ESP8266 RTC has 512 bytes of RAM that persist across restarts but not during power outages. While this RAM isn't persistent across power outages it also doesn't have the write limitations that flash memory has, making it appropriate for frequent writes which don't need to survive power loss.

The RTC slow memory can be used by initialising the variable as follows
`RTC_DATA_ATTR uint32_t pulse_count;   // Persistent counter`

Some more information here
https://stackoverflow.com/questions/78118223/rtc-noinit-attr-vs-rtc-data-attr-slow-vs-fast-memory

I haven't looked at how the code is integrated into ESPHome stack, I have done most development using Arduino IDE with the espressif ESP board package. However I assume it should be similar, if the variable is initialised it should retain the value when waking up from deep sleep.

In an similar project to this using Arduino, not ESPHome, I also added a step to reset the counter if a power-cycle was detected instead of a deep-sleep wake-up, as I believe it's possible for the RTC memory to initialise in a random state. I did this by adding a unique initialisation flag in RTC slow memory.

I've added this project to github so you can take a look if that is helpful
https://github.com/michaelt88/ULP_pulse_counter

The problem I found with not having the total counted on the device, is if you miss a message from the ESP meter, then you can't get an accurate total. Having this functionality available through ESPHome would be awesome.

m0uH:
Thank you for the explanation. That's also how my implementation of the total sensor works. You get the pulses summed up across deep sleeps. But not across power outage or external reset of the board (did not try rebooting it via software, if that's possible).

After looking at your code, I do think it works the same. I also think it's the same for the non-ULP pulse counter, after reading a bit through the corresponding pull request - the absolute count does not survive power outages. One could use the integration component or total_daily_energy component.

> The problem I found with not having the total counted on the device, is if you miss a message from the ESP meter, then you can't get an accurate total. Having this functionality available through ESPHome would be awesome.

That is solved by my changes, as the pulses are counted across deep sleeps.

m0uH:
We might get around the failing pipeline check by encapsulating the global variables in a static function of PulseCounterUlpSensor:

> int& PulseCounterUlpSensor::get_pulse_count_persist_handle()
{
  RTC_DATA_ATTR static int pulse_count_persist = 0;
  return pulse_count_persist;
}
int& PulseCounterUlpSensor::get_mean_exec_time_handle()
{
  RTC_DATA_ATTR static int mean_exec_time = 0;
  return mean_exec_time;
}

brisk0:
@m0uH  That was my next plan after static class variables, but static class variables behaved. Mostly.

I assume the remaining clang-tidy issue is that it's trying to run for an ESP8266 env at all, given that the RTC_* vars are ESP32 only. I've yet to find a way to insist that the component is ESP32 only that will skip the ESP8266 clang-tidy check

m0uH:
May be adding this to the `sensors.py` prevents using the component for non esp32:

`DEPENDENCIES = ["esp32"]`

See https://github.com/esphome/esphome/blob/dev/esphome/components/esp32_hall/sensor.py#L10

m0uH:
> Interesting! I am currently using a prox sensor to monitor the rotating metal on the water meter, this draws around 1-2 mA. Pulsing an LED and using an LDR to detect the rotating metal sounds like a great solution that would allow you to dramatically lower power consumption. Are you able to share any details on how you have physically setup the LED & LDR to the water meter?

@michaelt88 I've got this working now in a prototype state. I can pulse the LED and the detection module with a 2.5ms delay. To detect all pulses I need to check with a rate of 5Hz, approximately every 200ms. Having the LEDs enabled for about 3ms every 200ms results in an average power consumption of 0.08mAh during deep sleep in my case.

@brisk0 wondering if I should add such a feature to this MR? I could add an optional configuration setting which allows the user to set the pin to pulse and a delay in ms to leave the pin "On". What do you think?

brisk0:
@m0uH That sounds very useful. I would suggest it would be appropriate to be its own PR. Though of course if it depends on this PR, I can make no guarantees about when this could be merged.

m0uH:
I've added the corresponding changes to my fork of your branch, such that any interested can have a look at.

See here the diffs to this MR: https://github.com/brisk0/esphome/compare/pulse-counter-ulp...m0uH:esphome:pulse-counter-ulp

I've modified the pulse-counter-ulp component, so the changes don't make sense without it. But of course I can create a new PR once this here is merged.

RedFlash49:
Hi
Nice work!
Am i right that this code only supports one pulse pin? Im working on a weather station with windspeed pulses and rain amount pulses.
An ULP-double-counter would be perfect for this.
Any idea how to achieve this?

thanks

brisk0:
@RedFlash49 You are right that this only supports one pin, and you can't have multiples of the component either.
Making a trivial variant that has support for multiple pins probably wouldn't be too difficult, but making one with an arbitrary number would require substantial changes to the design of the assembly ULP program.

The design is very much geared towards one pin at the moment.

Hannes-Beckmann:
Hi, I wanted to use this component to count my electricity meter, which is in the basement. 
Sadly, esphome can't find the sensor platform. I am using esphome in Homeassistant with an esp32-s3.
This is my configuration, is there something wrong with it?:

```yaml
esphome:
  name: lora-electricity-counter
  friendly_name: lora_electricity_counter

esp32:
  board: esp32-s3-devkitc-1
  framework:
    type: esp-idf

external_components:
  - source: github://brisk0/esphome@pulse-counter-ulp
    components: [ pulse_counter_ulp ]


# Enable logging
logger:

# Enable Home Assistant API
api:
  encryption:
    key: "x"

ota:
  - platform: esphome
    password: "x"

wifi:
  ssid: !secret wifi_ssid

  ap:
    ssid: "Lora-Electricity-Counter"
    password: "x"

captive_portal:
    

sensor:
  - platform: pulse_counter_ulp
    name: "Number of Pulses"
    pin:
      number: GPIO7
      inverted: true
      mode:
        pullup: true
        input: true
    update_interval: 5s
```

brisk0:
@Hannes-Beckmann I haven't used External Components before, but this PR does include a change to a file outside of the component directory (const.py), which I can imagine being an issue.

What error are you seeing when you try to build the config?

Hannes-Beckmann:
@brisk0 Ahh I see, this is exactly it. So I guess it is not usable as an external component. 
I will fork your work and modify it to make it usable until this PR is merged. 
Thank you for your effort!

```
INFO ESPHome 2025.2.2
INFO Reading configuration /config/esphome/lora-electricity-counter.yaml...
ERROR Unable to import component pulse_counter_ulp.sensor:
  File "/esphome/esphome/loader.py", line 181, in _lookup_module
  File "/usr/lib/python3.11/importlib/__init__.py", line 126, in import_module
  File "/data/external_components/3c33d76f/esphome/components/pulse_counter_ulp/sensor.py", line 7, in <module>
    from esphome.const import (
ImportError: cannot import name 'CONF_EDGES_WAKEUP' from 'esphome.const' (/esphome/esphome/const.py)
Failed config

sensor.pulse_counter_ulp: [source /config/esphome/lora-electricity-counter.yaml:40]
  
  Platform not found: 'sensor.pulse_counter_ulp'.
  platform: pulse_counter_ulp
  name: Number of Pulses
  pin: 
    number: GPIO7
    inverted: True
    mode: 
      pullup: True
      input: True
  update_interval: 5s
```

alexd321:
is there a way to beta test this, as

```
external_components:
  - source: github://pr#7548
    components: 
      - pulse_counter_ulp
```

doesn't work?

brisk0:
@alexd321 There is a change to the repository outside of the component, so it won't work as an external component.

I recommended using the esphome "set up a development environment" instructions:

https://developers.esphome.io/contributing/development-environment/

Clone this repo, checkout this branch, set up the virtual environment, then run esphome from inside the virtual environment and it will include this component.

If that sounds like a bunch of nonsense I'm happy to clarify further.

pVer297:
Hello!
I've been using your component, as this fits exactly to what I was trying to do! It works pretty good, the installation and usage is straightforward and the documentation helps as well. Super greatful!

Now, it might just wrong usage on my end, but I have found an issue: the compilation tries to include the ULP code parts even when no ULP configuration is present in the yaml.

Steps to reproduce:
- Crete configuration without ULP component
    - Compilation OK
- Add ULP component
    - Compilation OK
- Remove ULP component
    - Compilation fails: `esp32ulp-elf-ld.exe: invalid length for memory region ram`
- Clean configuration
    - Compilation still fails
- Clean configuration + manually remove `.esphome/build/ulp` & `.esphome/build/src/CMakeLists.txt`
    - Compilation OK


Looks like project clean / recompilation does not take care of ULP dependencies when they are no longer required.

esphome/components/pulse_counter_ulp/pulse_counter_ulp_sensor.cpp
```diff
@@ -0,0 +1,205 @@
...
+  if (esp_sleep_get_wakeup_cause() == ESP_SLEEP_WAKEUP_UNDEFINED) {
+    return setup_priority::DATA;
+  } else {
+    /* In case of deep sleep wakup, we need to delay sending new values until we have a connection */
+    return setup_priority::AFTER_CONNECTION;
+  }
+}
+
+void PulseCounterUlpSensor::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up pulse counter '%s'...", this->name_.c_str());
```
kbx81:
```suggestion
  ESP_LOGCONFIG(TAG, "Running setup for '%s'", this->get_name().c_str());
```

esphome/components/pulse_counter_ulp/pulse_counter_ulp_sensor.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+void PulseCounterUlpSensor::dump_config() {
+  LOG_SENSOR("", "Pulse Counter", this);
+  LOG_PIN("  Pin: ", this->config_.pin_);
+  ESP_LOGCONFIG(TAG, "  Rising Edge: %s", to_string(this->config_.rising_edge_mode_));
+  ESP_LOGCONFIG(TAG, "  Falling Edge: %s", to_string(this->config_.falling_edge_mode_));
+  ESP_LOGCONFIG(TAG, "  Sleep Duration: %" PRIu32 " µs", this->config_.sleep_duration_ / microseconds{1});
+  ESP_LOGCONFIG(TAG, "  Debounce: %" PRIu16, this->config_.debounce_);
+  ESP_LOGCONFIG(TAG, "  Edges Wakeup: %" PRIu16, this->config_.edges_wakeup_);
+  LOG_UPDATE_INTERVAL(this);
```
kbx81, line 156: Please arrange this as described in [our developer docs](https://developers.esphome.io/architecture/logging/#best-practices-for-combined-logging).

brisk0:
@pVer297 

I did discover this issue when originally developing this PR, and at the time I was unable to find a decent solution.

At the moment I'm not able to look into it as my development environment appears to have bit-rotted away (possibly pending closure of https://github.com/platformio/platform-espressif32/pull/1599).

If you or anyone else are able to suggest how cleaning the ULP files can be done I'm all ears.

brisk0:
Converting to draft while I try to merge in the dev branch

pVer297:
I have no solution to the cleanup issue, as I'm not that familiar with Esphome development. On a different note, I came back with another finding (and solution) this time.

I have observed that the "pulse/min" metric is not 100% correct with longer sleep durations. In my setup I have 285s sleep with 15s wake time (5 min/period). In this frame the calculated pulse frequency is off by 1.5-5x depending on the real frequency. I guess due to the way the time is kept during sleep (calculating from execution time).

kép

instant is the pulse counter (multiplied by 60 with a filter)
total is the total counted pulses. This is correct over multiple days of data collection.
calc_instant is the calculated power by checking two adjacent **total** points and the time between them.

For me the solution was to use the time kept by the RTC directly instead of calculating the sleep time. Code example from ESP-IDF programming guide:
```
struct timeval tv_now;
gettimeofday(&tv_now, NULL);
int64_t time_us = (int64_t)tv_now.tv_sec * 1000000L + (int64_t)tv_now.tv_usec;
```

My changes:
```
RTC_DATA_ATTR int64_t PulseCounterUlpSensor::last_time_us = 0;  // add to header declaration as well

int64_t get_rtc_usec() {
  struct timeval now;
  gettimeofday(&now, NULL);
  return (int64_t)now.tv_sec * 1000000L + (int64_t)now.tv_usec;
}

[...]
  // instead of clock::now()
  int64_t now_us = get_rtc_usec();
  std::chrono::duration interval = microseconds(now_us - last_time_us);

  last_time_us = now_us;
[...]
```

brisk0:
Hey @pVer297, I really appreciate you testing the accuracy!

At first blush, using the RTC instead of estimating the sleep time is so obviously the right thing to do that I really hope I had a good reason to not do it like that in the first place. I still need to scrounge some more time to comfirm it on my end, but thank you for posting your solution.

michaelt88:
> Hey @pVer297, I really appreciate you testing the accuracy!
> 
> At first blush, using the RTC instead of estimating the sleep time is so obviously the right thing to do that I really hope I had a good reason to not do it like that in the first place. I still need to scrounge some more time to comfirm it on my end, but thank you for posting your solution.

My solution was to count the total ticks, get the difference in ticks, and then apply the transform to get ms

The transform to get the ms will change, so yes if you apply the transform at different times it will give incorrect results

> ticksNow = rtc_time_get() & 0xFFFFFFFF; // get lower 32-bits of current RTC clock ticks (48-bit)
> prevTicks = RTC_SLOW_MEM[PREV_TICKS]; // get previous lower 32-bits of RTC clock ticks (RTC memory is 32-bit)
> ticksDiff = ticksNow - prevTicks; // calc differnt in tics. this should handle overflow because the variables are 32-bit unsigned in.
>  timeDiff = rtc_time_slowclk_to_us(ticksDiff, esp_clk_slowclk_cal_get()); //note there is some variance in the esp_clk_slowclk_cal_get value. If the timeDiff is from timeNow minus prevTime, then this can lead to issues with different calibration values used for timeNow and prevTime. So that's why we store the ticks, and perform the conversion using the ticksDiff

Reference here
https://github.com/michaelt88/ULP_pulse_counter

brisk0:
The ESP32 C++ standard library implements chrono::system_clock as an interface to the RTC (allowing for smooth adjustments/corrections), so I've used that and ripped out the dodgy sleep estimation.

@pVer297 if you're willing or able to test it again using your setup, it would be good to see if it makes a difference and how much. I will actually get mine hooked up to a power meter as well - I've been testing to date with a breadboard and a tactile switch.

brisk0:
Will need https://github.com/esphome/esphome/pull/9428 merged into dev before I can get this branch up-to-date.

pVer297:
> @pVer297 if you're willing or able to test it again using your setup, it would be good to see if it makes a difference and how much. I will actually get mine hooked up to a power meter as well - I've been testing to date with a breadboard and a tactile switch.

I'll try to check it once I got the time. The changes look good, cleaner than mine. ~~At the first glance I think it makes more sense to update the `last_time_` value in the `update()` function instead of `setup()`. If there is no sleep for some time, the `last_time_` will not update.~~

Update:
Now I'm using your latest code and the **accuracy is good**. I've checked with 5 minutes sleep and without sleep as well.

brisk0:
Thanks @pVer297

TheOriginalMrWolf:
> Hello! I've been using your component, as this fits exactly to what I was trying to do! It works pretty good, the installation and usage is straightforward and the documentation helps as well. Super greatful!
> 
> Now, it might just wrong usage on my end, but I have found an issue: the compilation tries to include the ULP code parts even when no ULP configuration is present in the yaml.
> 
> Steps to reproduce:
> 
> * Crete configuration without ULP component
>   
>   * Compilation OK
> * Add ULP component
>   
>   * Compilation OK
> * Remove ULP component
>   
>   * Compilation fails: `esp32ulp-elf-ld.exe: invalid length for memory region ram`
> * Clean configuration
>   
>   * Compilation still fails
> * Clean configuration + manually remove `.esphome/build/ulp` & `.esphome/build/src/CMakeLists.txt`
>   
>   * Compilation OK
> 
> Looks like project clean / recompilation does not take care of ULP dependencies when they are no longer required.

First, thanks @brisk0 for making this component - it works really well!  I've been trialling this under a range of... challenging... scenarios for the last week and it hasn't skipped a beat - thank you! 👍

Re the issue above raised by @pVer297 - I'm very new to this whole ESPHome build process but managed to work out the following if it helps:

Before including the `ulp_pulse_count` in a yaml file the ESPhome `config/.esphome/build/<yaml-file-name>/src/CMakeLists.txt` file is very simple:
```
# This file was automatically generated for projects
# without default 'CMakeLists.txt' file.

FILE(GLOB_RECURSE app_sources ${CMAKE_SOURCE_DIR}/src/*.*)

idf_component_register(SRCS ${app_sources})
```
Once you add in the  `ulp_pulse_count` component, the `Master Control Program` \*heh\* creates a sibling directory (`ulp`) to `src` where it puts `pulse_cnt.S`, and it updates the `CMakeLists.txt` to the following:
```
FILE(GLOB_RECURSE app_sources ${CMAKE_SOURCE_DIR}/src/*.*)

idf_component_register(SRCS ${app_sources}
INCLUDE_DIRS ""
REQUIRES driver soc nvs_flash ulp)

##
## ULP support additions to component CMakeLists.txt.
##
## 1. The ULP app name must be unique (if multiple components use ULP).
set(ulp_app_name ulp_main)
##
## 2. Specify all assembly source files.
##    Files should be placed into a separate directory (in this case, ulp/),
##    which should not be added to COMPONENT_SRCS.
##TODO
set(ulp_s_sources "../ulp/pulse_cnt.S" "../ulp/wake_up.S")
##
## 3. List all the component source files which include automatically
##    generated ULP export file, ${ulp_app_name}.h:
set(ulp_exp_dep_srcs "components/pulse_counter/pulse_counter_sensor.cpp")
##
## 4. Call function to build ULP binary and embed in project using the argument
##    values above.
ulp_embed_binary(${ulp_app_name} "${ulp_s_sources}" "${ulp_exp_dep_srcs}")
#
target_compile_options(${COMPONENT_LIB} PRIVATE "-Wno-format")
```

The problem arises that, if/when you remove the `ulp_pulse_count` component from your yaml file, the `ulp` directory & contents remain, and the `CMakeLists.txt` remains the same as well - even if you do a `Clean Build Files` operation!  This of course means that on subesquent builds `pulse_cnt.S` gets unneccesarily recompiled and then things fall apart.

Unfortunately I have no idea where the fault lies or how to fix this, but maybe it points someone who _actually_ knows what they're doing in the right direction?

Cheers!

alexd321:
I tried this, but was compiling failed.

> @alexd321 There is a change to the repository outside of the component, so it won't work as an external component.
> 
> I recommended using the esphome "set up a development environment" instructions:
> 
> ........

An easier way is to install the ESPHome Device Builder (Dev), then setting Install ESPHome from a fork or branch to "brisk0:pulse-counter-ulp"

image

Then it works well :)

alexd321:
Still waiting for merge

moustic999:
Should it be possible to make this available as an external_component ?

brisk0:
@moustic999 The only thing stopping this from being used as an external component is the addition of `CONF_EDGES_WAKEUP = "edges_wakeup"` to `const.py` (because this change is outside the component directory).

This is only necessary as a part of quality and conformity of esphome. If I weren't hoping to merge, then that line could either be moved into the component or an existing configuration constant could be used in place of the new one. Following that it should be perfectly possible to use as an external component.

moustic999:
do you think it would be possible and lot of work to support esp32-c5 also ?  it seems ulp is different on these but should become the standard way for future esp

moustic999:
> @moustic999 The only thing stopping this from being used as an external component is the addition of `CONF_EDGES_WAKEUP = "edges_wakeup"` to `const.py` (because this change is outside the component directory).
> 
> This is only necessary as a part of quality and conformity of esphome. If I weren't hoping to merge, then that line could either be moved into the component or an existing configuration constant could be used in place of the new one. Following that it should be perfectly possible to use as an external component.

Seeing the time it take to merge and how it seems to be ignored, I would likely propose to make it available as a external components, at least we will not need to have a dev version of esp_home to use this

brisk0:
@moustic999 Point well taken, I will make this available as an external component.
I'm not likely to look into esp32-c5 myself simply because I don't have one and I don't need or use this component anymore

brisk0:
For those hoping for an external-component version of this, the following currently seems to work:

```yaml
external_components:
  - source:
      type: git
      url: https://github.com/brisk0/esphome.git
      ref: pulse-counter-ulp
    components: [const, pulse_counter_ulp]
```

brisk0:
External component version now has a permanent home (that doesn't require external `const` component) over here:
https://codeberg.org/brisk/esphome-pulse-counter-ulp

alexd321:
Please reopen

RedFlash49:
I dont get it why this is not merged allready. Im a fan of Esphome but this is ridiculous.

moustic999:
For me, as long as it is available as external component, I don't care
having it merged....

PR #7551, opened 2024-10-06
Enable support for low power protocol v0 of Tuya MCU

warehog:
There is speciall category of TuyaMCU based devices that are battery powered.
In such scenario, an modified version of v0 protocol is used, which is deeply incompatible with v3.

WARNING:
At this point, the code is in a POC state. I was able to report humidity and temperature values to HA, with the cost of completly breaking v3 protocol support.
image
Both protocols share some of the code - datapoint handling, localtime handling
Main differences:
* command IDs
* initial configuration flow (no config retrieve, no datapoint list retrieve)
* custom WIFI reporting state (MCU must know that Module has API connection)
* custom state loop
It might be ctually reasonable to make it into a dedicated component like `tuya_le` aka `tuya low energy`

There is excelleng guide on LE protocol:
https://www.elektroda.com/rtvforum/topic3975583.html

The basic operation looks like this:
Module -> Wifi Module with ESPHome loaded
MCU -> Microcontroller that operated re sensors

MCU wakes  up on event (sensor value change, timer etc)
MCU power up Module
Module initializes MCU with current date and networks status
MCU sends data points updates
Module ACKs updated
MCU powers down Module & itself

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Other

https://github.com/esphome/feature-requests/issues/497
https://github.com/esphome/feature-requests/issues/352

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ x ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: tmp-hum-3ft6

bk72xx:
  board: cb3s

logger:
  level: VERBOSE
  hardware_uart: UART2


api:
  password: ""

ota:
  - platform: esphome
    password: ""

web_server:

wifi:
  ssid: <ssid>
  password: <passwd>
  ap:
    ssid: tmp-hum-3ft6
    password: ""

captive_portal:

time:
  - platform: sntp
    id: timer

uart:
  rx_pin: P10
  tx_pin: P11
  baud_rate: 9600

sensor:
  - platform: tuya
    name: temperature
    sensor_datapoint: 1
  - platform: tuya
    name: humidity
    sensor_datapoint: 2

tuya:
  time_id: timer
```

## Checklist:
  - [ x ] The code change is tested and works locally.

### DIFF
--- a/esphome/components/tuya/tuya.cpp
+++ b/esphome/components/tuya/tuya.cpp
@@ -176,129 +176,29 @@ void Tuya::handle_command_(uint8_t command, uint8_t version, const uint8_t *buff
         this->product_ = R"({"p":"INVALID"})";
       }
       if (this->init_state_ == TuyaInitState::INIT_PRODUCT) {
-        this->init_state_ = TuyaInitState::INIT_CONF;
-        this->send_empty_command_(TuyaCommandType::CONF_QUERY);
+        this->init_state_ = TuyaInitState::INIT_WIFI;
+        this->send_raw_command_(TuyaCommand{.cmd = TuyaCommandType::WIFI_STATE, .payload = std::vector<uint8_t>{0x03}});
+        this->send_raw_command_(TuyaCommand{.cmd = TuyaCommandType::WIFI_STATE, .payload = std::vector<uint8_t>{0x04}});
       }
       break;
     }
-    case TuyaCommandType::CONF_QUERY: {
-      if (len >= 2) {
-        this->status_pin_reported_ = buffer[0];
-        this->reset_pin_reported_ = buffer[1];
-      }
-      if (this->init_state_ == TuyaInitState::INIT_CONF) {
-        // If mcu returned status gpio, then we can omit sending wifi state
-        if (this->status_pin_reported_ != -1) {
-          this->init_state_ = TuyaInitState::INIT_DATAPOINT;
-          this->send_empty_command_(TuyaCommandType::DATAPOINT_QUERY);
-          bool is_pin_equals =
-              this->status_pin_ != nullptr && this->status_pin_->get_pin() == this->status_pin_reported_;
-          // Configure status pin toggling (if reported and configured) or WIFI_STATE periodic send
-          if (is_pin_equals) {
-            ESP_LOGV(TAG, "Configured status pin %i", this->status_pin_reported_);
-            this->set_interval("wifi", 1000, [this] { this->set_status_pin_(); });
-          } else {
-            ESP_LOGW(TAG, "Supplied status_pin does not equals the reported pin %i. TuyaMcu will work in limited mode.",
-                     this->status_pin_reported_);
-          }
-        } else {
-          this->init_state_ = TuyaInitState::INIT_WIFI;
-          ESP_LOGV(TAG, "Configured WIFI_STATE periodic send");
-          this->set_interval("wifi", 1000, [this] { this->send_wifi_status_(); });
-        }
-      }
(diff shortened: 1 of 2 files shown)


bilogic:
~I like the idea of `platform: tuya_le`. We could eventually merge it back into `tuya` when better information comes in.~

Alternatively, this could also make sense
```YAML
sensor:
    - platform: tuya
      version: 0 # defaults to 3 if unspecified 
```

In my experience with protocol/interface versioning and after looking at the changes required, I think it is better to completely separate v3 and v0 in this case because they were not designed with backward compatiblity in mind. I think we can still park everything under `tuya` and add on `version`.

This approach should make the `tuya` component easier to maintain going forward.

bilogic:
@kudlatykaskader I think you are missing `#ifdef USE_TIME` here https://github.com/esphome/esphome/blob/42d65baabbed3aa28a6b04d1f0ab20f133c2e9a5/esphome/components/tuya/tuya.cpp#L185-L187

warehog:
I admit that this PR was made only to see if there is interest in such a
change (new component dedicated for LE Tuya devices).
The code there is a roughly modified/cut-down version of the original tuya
component, with modified state machine and entity autodiscovery totally
deleted. There is still a lot of work needed.
After some time I realized that tuya_le should be only used with a
companion of MQTT, since the Wifi module is only awake for a limited amount
of time, and there is no way to control that time window.

bilogic:
My intention was to help out where I can. Did you develop based on this document? https://developer.tuya.com/en/docs/iot/ble-door-lock-mcu-development-overview?id=K95l67wf9xopf

warehog:
I used a few different Tuya documents, but no one seemed to match a 100%. At a time I had only a USB<->TTL coverter, sniffing one channel at a time, which gave me a basic idea what how this protocol works. Perhaps there are multiple V0 sub protocols for different devices?
Now, I have a proper logic analyzer (8 channel). Give me some time - I'll bring back the setup. 

The device I was working on is:
462541610_1293836318709964_1648024437280885949_n

bilogic:
Sure, I used an ESP32 to sniff TX and RX. Now I'm thinking of how to replace the BK with ESP32. But one main observation is, the BK responds immediately to the pings, no delay.

warehog:
Original TUYA firmware dump
[tuya_white_clock_le_BK7231N_QIO_2024-19-11-23-45-50.zip](https://github.com/user-attachments/files/17822401/tuya_white_clock_le_BK7231N_QIO_2024-19-11-23-45-50.zip)

And The comm sniffs from different actions (https://sigrok.org/wiki/Downloads)

[tuya_white_clock_wifi_reset_dump.zip](https://github.com/user-attachments/files/17822711/tuya_white_clock_wifi_reset_dump.zip)

[tuya_white_clock_temp_change_trigger_dump.zip](https://github.com/user-attachments/files/17822766/tuya_white_clock_temp_change_trigger_dump.zip)


It seems to follow this protocol here:
https://developer.tuya.com/en/docs/iot/tuyacloudlowpoweruniversalserialaccessprotocol?id=K95afs9h4tjjh#title-0-Overview

@bilogic what do you think?

bilogic:
@kudlatykaskader 

Must the module (BK72xx) talk about wifi with the MCU? I suppose everything else should still work except the wifi icon will keep blinking right?

I was able to get my battery device to work with the main branch of ESPHome code with about 10 lines of new code. My device is the BK3432, which is a BLE only device, no wifi.

PR #7556, opened 2024-10-07
[online_image] Bugfix: Use std::string instead of const char *

### LINKED ISSUE #6309: Online Image URL Action is Unsafe
### The problem

The automation for [online_image.h](https://github.com/esphome/esphome/blob/dev/esphome/components/online_image/online_image.h#L160) uses a templatable argument value of a `char *`.  If this is populated via a non-static `std::string` the pointer may be garbage by the time it is converted back into a `std::string` 

In my case it's getting the picture entity  that is ~180 chars.  I see this in the log:  

```
[19:38:06][W][http_request.arduino:110]: HTTP Request failed; URL: http://homeassxV\xad\xba\xfd?eO; Error: connection refused
```

It would be better to use a `std::string` the entire way through.  Since this code is autogenerated it should be an invisible fix. 

2024.9.2

Home Assistant Add-on

all

ESP32

### Board

ESP32-D0WDQ5 rev1 on a wt32-sc01

online iage

### Example YAML snippet

```yaml
api:
  actions:
    - action: update_albumart_image
      variables:
        url: string
      then:
        - online_image.set_url:
            id: current_album_art
            url: !lambda |-
              return str_sprintf("http://homeassistant.local:8123%s", url.c_str()).c_str();
        - component.update: current_album_art
```

_No response_

### Additional information

I am using the new LVGL so that can suck up a bit of memory causing this to be aggravated vs when I am just downloading the image on a simple image.

clydebarrow:

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/6309

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/online_image/online_image.h
+++ b/esphome/components/online_image/online_image.h
@@ -170,7 +170,6 @@ template<typename... Ts> class OnlineImageSetUrlAction : public Action<Ts...> {
 template<typename... Ts> class OnlineImageReleaseAction : public Action<Ts...> {
  public:
   OnlineImageReleaseAction(OnlineImage *parent) : parent_(parent) {}
-  TEMPLATABLE_VALUE(const char *, url)
   void play(Ts... x) override { this->parent_->release(); }
 
  protected:
--- a/esphome/components/online_image/online_image.h
+++ b/esphome/components/online_image/online_image.h
@@ -157,7 +157,7 @@ class OnlineImage : public PollingComponent,
 template<typename... Ts> class OnlineImageSetUrlAction : public Action<Ts...> {
  public:
   OnlineImageSetUrlAction(OnlineImage *parent) : parent_(parent) {}
-  TEMPLATABLE_VALUE(const char *, url)
+  TEMPLATABLE_VALUE(std::string, url)
   void play(Ts... x) override {
     this->parent_->set_url(this->url_.value(x...));
     this->parent_->update();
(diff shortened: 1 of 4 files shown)


esphome/components/online_image/online_image.h
```diff
@@ -157,7 +157,7 @@ class OnlineImage : public PollingComponent,
 template<typename... Ts> class OnlineImageSetUrlAction : public Action<Ts...> {
  public:
   OnlineImageSetUrlAction(OnlineImage *parent) : parent_(parent) {}
-  TEMPLATABLE_VALUE(const char *, url)
+  TEMPLATABLE_VALUE(std::string, url)
   void play(Ts... x) override {
     this->parent_->set_url(this->url_.value(x...));
     this->parent_->update();
```
guillempages, line 163: Would you mind removing the `TEMPLATABLE_VALUE(const char *, url)` in the ` OnlineImageReleaseAction` (line 173) below as well?

IIRC that one is not needed at all, and since you are cleaning up anyway ;-)

guillempages (approved):
I think there was a reason why I used const char* originally, but I don't think that this reason is still valid :-)

There is another (unused AFAIR) mention fo a `const char *` some lines below your change; would you mind removing that as well? Or should I do a separate PR myself to clean that single line up?

esphome/components/online_image/online_image.h
clydebarrow: Sure, I'll deal with that tomorrow.

guillempages (approved):

kbx81 (approved):

PR #7557, opened 2024-10-07
[cst816] Allow skipping i2c probe

### LINKED ISSUE #6293: CST816S - component setup fails if the touchscreen controller is in standby state
### The problem

The CST816 touchscreen component, during setup() and particularly continue_setup(), reads a register from the I2C bus asking the touchsreen controller its chip ID register. If this read fails, the component is marked as failed and no further action can be done to recover it.

The problem is that the CST816S chip (the one I'm using) enters a **standby state** after 2 seconds of touch inactivity. When in this mode, the chip **does not** respond to I2C commands until it is awaken by a touch ([see datasheet](https://files.waveshare.com/upload/5/51/CST816S_Datasheet_EN.pdf)).

Thus, initialization of the component **always fails**, unless you **keep tapping on the screen until the component is loaded**. If the component is initialized correctly, then touch works fine indefinitely. Moreover, when the chip is in standby, it doesn't even appear in I2C bus scans.

I tried checking the source code of the component and it appears that the chip ID is not really used, so maybe it could be possible to accept it via config and, if provided, to avoid asking the chip ID on setup or maybe just issuing a warning without marking the component as failed. Unfortunately I'm not skilled enough to provide a MR.

I'm not sure if other variants of this chip have the same issue.

2024.9.2

Docker

ESP32

### Board

nodemcu-32s

cst816

### Example YAML snippet

```yaml
i2c:
  - id: i2c_bus0
    sda: GPIO21
    scl: GPIO4
    scan: true

touchscreen:
  platform: cst816
  id: touchscreen0
  i2c_id: i2c_bus0
  interrupt_pin: GPIO23
  reset_pin: GPIO32
  address: 0x15
```

clydebarrow:

## Types of changes

- [ ] Other

**Related issue or feature (if applicable):** fixes https://github.com/esphome/issues/issues/6293

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4312
## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/cst816/touchscreen/__init__.py
+++ b/esphome/components/cst816/touchscreen/__init__.py
@@ -1,11 +1,10 @@
-import esphome.codegen as cg
-import esphome.config_validation as cv
-
 from esphome import pins
+import esphome.codegen as cg
 from esphome.components import i2c, touchscreen
-from esphome.const import CONF_INTERRUPT_PIN, CONF_ID, CONF_RESET_PIN
-from .. import cst816_ns
+import esphome.config_validation as cv
+from esphome.const import CONF_ID, CONF_INTERRUPT_PIN, CONF_RESET_PIN
 
+from .. import cst816_ns
 
 CST816Touchscreen = cst816_ns.class_(
     "CST816Touchscreen",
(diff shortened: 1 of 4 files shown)


dsolera:
I tested this PR on my CST816S and the new flag works perfectly. The touchscreen initializes fine and it works immediately after boot without the need to repeatedly touch the screen.

clydebarrow:
@dsolera  I just cleaned up the code a bit - could you please test again just to be sure I didn't break anything? Thanks.

jesserockz (approved):

dsolera:
It seems to be working fine after the latest changes. Thanks!

PR #7564, opened 2024-10-08
Extending the bytebuffer Class. 

nielsnl68:

Introducing the following new options:
- Added a way to figure out the total used_space, not depending on the current position anymore. 
- `flip();` will now jump to the *used_space* marker instead of the current *position*. [BREAKING]
- added a getter/setter for byte arrays.
- Added a flag that will be set when the buffer is being changed.
- changing the parameter naming for byte array wrapper.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:
What's the use-case for these changes?

> Added a way to figure out the total used_space, not depending on the current position anymore

Why would the space used not be equal to the position?

Adding the extra call on every update to the buffer will slow it down significantly. This is supposed to be a low-overhead implementation, since it's intended to be used in encoding/decoding communication buffers.

nielsnl68:
> What's the use-case for these changes?
> 
> > Added a way to figure out the total used_space, not depending on the current position anymore
> 
> Why would the space used not be equal to the position?

The position could be moved to other positions within the bytebuffer like the example below:

```cpp




struct example {
   ByteBuffer: data(100);

   bool get_active() { return data[0].get_bool(); }
   uint16_t get_counter() { return data[1].get_uint16(); }

   uint8_t get_size() { return data[3].get_uint16(); }
   uint32_t get_id() { return data[4].get_uint32(); }
   void get_topic(uint8_t *topic, uint8_t size) { data[15].get_bytes(topic, size); }

   Void set_active() { return  data[0].put_bool(true); } 
   uint16_t add_counter() { return data[1].get_uint16(data[1].get_uint16()+1); }

   void set_size(uint8_t size) { data[3].put_uint8(size); }
   void set_id(uint32_t id) {data[4].put_uint32(id); }
   void set_topic(uint8_t *topic, uint8_t size) { data[15].put_bytes(topic, size); }
   void set_crc() { 
     uint8_t crc = calc_crc();
     data[15 + this->get_size()].put_uint8_t(crc); 
     data.resize();
   }
   /// etc. Missing the setters as this is just an example.

   uint8_t * get_data() { return data.get_data().data(); }
}

void change_topic (uint8_t *topic, uint8_t size) {
  example.set_topic(topic,size);
  example.set_size(size);
  example.data[0].put_bool(false);
  example.add_counter();
}


void main() {
  example.set_id(1324);
  change_topic ("abcsideg",8);
  set_active();
  send_example (example.get_data(), example.data.get_used_data());
  /// etc.

}

```




> 
> Adding the extra call on every update to the buffer will slow it down significantly. This is supposed to be a low-overhead implementation, since it's intended to be used in encoding/decoding communication buffers.

This makes sense, how could we speed this up?

clydebarrow:
```
return data[3].get_uint16();
```
Ah - your PR comment did not not mention the addition of the subscript operator. As implemented, this is a a Really Bad Idea - there should be no side-effects on the target when using subscription.

If you want the ability to read/write things from specific offsets, then you would want functions that do that without changing the position, e.g.

```
uint16_t ByteBuffer::get_uint16(size_t offset);
```

What I'd like to avoid is ByteBuffer becoming a Swiss Army knife - something that does lots of things, but none of them very well. Having said that I can see the use case for position-independent operations, so if done properly it could be a useful addition.

nielsnl68:
> ```
> return data[3].get_uint16();
> ```
> 
> Ah - your PR comment did not not mention the addition of the subscript operator. As implemented, this is a a Really Bad Idea - there should be no side-effects on the target when using subscription.'

Corrected. It is now in the list of changes.

> 
> If you want the ability to read/write things from specific offsets, then you would want functions that do that without changing the position, e.g.
> 
> ```
> uint16_t ByteBuffer::get_uint16(size_t offset);
> ```
This was my first troughs to do, do it this way i needed to do so much changes by add adding a lot of extra code per getter and putter inclusive adding an extra method's as well. In my opinion this would break to the idea of having a simple class. 
 
> What I'd like to avoid is ByteBuffer becoming a Swiss Army knife - something that does lots of things, but none of them very well. 

I agree on this, when you have an better solution that is easy to implement and use then i'm open to implement that in favor of what i did.

> Having said that I can see the use case for position-independent operations, so if done properly it could be a useful addition.

As i said above, please give me a suggestion on how to do it properly then i'm happy to implement it.

clydebarrow:
There's no quick and easy solution, the best option I can see is to rewrite all the put and get functions to take a specified offset and not update the position, then have the variants without a specified offset call those with the current position and update it.

That's a fair bit of reworking, but will give the flexibility you want, will not impact performance too much and will avoid duplicating the encode/decode stuff. E.g.
```
uint8_t ByteBuffer::get_uint8(size_t offset) {
  return this->data_[offset];
}

uint8_t ByteBuffer::get_uint8() {
  return this->get_uint8(this->position_++);
}

```

etc.

nielsnl68:
tried to change the branch name. It did not go well here.

nielsnl68:
> There's no quick and easy solution, the best option I can see is to rewrite all the put and get functions to take a specified offset and not update the position, then have the variants without a specified offset call those with the current position and update it.
> 
> That's a fair bit of reworking, but will give the flexibility you want, will not impact performance too much and will avoid duplicating the encode/decode stuff. E.g.
> 
> ```
> uint8_t ByteBuffer::get_uint8(size_t offset) {
>   return this->data_[offset];
> }
> 
> uint8_t ByteBuffer::get_uint8() {
>   return this->get_uint8(this->position_++);
> }
> ```
> 
> etc.

I have been looking to do it this way. But quite frankly it gives a lot of overhead. Just for something niche. The above means just that it double the code and from what i understood from earlier messages, you where against that, and i agree on this. When you do belief that the extra method call will slow down the normal use.
My proposal is that I make the `update_used_space_();` method an **inline** method. This way it will be added to each calling method. 
This way the class stay's working as it is without much overhead, and when someone like to out of sequence excess there are some way's to do so.

PR #7565, opened 2024-10-08
[automation] Implement all and any condition shortcuts

clydebarrow:

Currently the `if:` action takes a `condition` option, which can take a list of conditions, which are implicitly `and`ed.

This PR creates alternatives for `condition` that allow both `and` and `or` operations on lists. These are named `all` and `any`, which should be more intuitively named for non-programmers, and mimic functional equivalents in several languages including Python. This also provides more explicit semantics vs. `condition` where it is not necessarily obvious whether it ors or ands a list.

It also allows `all` and `any` to be used in place of `and` and `or` in more complex conditions for consistency.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4316
## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml
    then:
      - if:
          any:
            - lambda: return 0;
            - lambda: return 0;
            - lambda: return 0;
          then:
            lambda: assert(false);
      - if:
          all:
            - lambda: return 1;
            - lambda: return 1;
            - lambda: return 0;
          then:
            lambda: assert(false);
      - if:
          any:
            - lambda: return 0;
            - lambda: return 0;
            - lambda: return 1;
          else:
            lambda: assert(false);
      - if:
          all:
            - any:
                - lambda: return 1;
                - lambda: return 1;
            - lambda: return 1;
            - lambda: return 1;
          else:
            lambda: assert(false);
      - if:
          condition:
            any:
              - lambda: return 0;
              - lambda: return 0;
              - lambda: return 0;
          then:
            lambda: assert(false);
      - if:
          condition:
            all:
              - lambda: return 1;
              - lambda: return 1;
              - lambda: return 0;
          then:
            lambda: assert(false);
      - if:
          condition:
            any:
              - lambda: return 0;
              - lambda: return 0;
              - lambda: return 1;
          else:
            lambda: assert(false);
      - if:
          condition:
            all:
              - lambda: return 1;
              - lambda: return 1;
              - lambda: return 1;
          else:
            lambda: assert(false);
```

## Checklist:


esphome/const.py
```diff
@@ -250,6 +252,7 @@
 CONF_ECO2 = "eco2"
 CONF_EFFECT = "effect"
 CONF_EFFECTS = "effects"
+CONF_ELIF = "elif"
```
guillempages: I don't see the CONF_ELIF being used in the code you modified 🤔  Is this a preparation for an upcoming PR (would be great to have the elif)?

clydebarrow: That would have been nice, but it's not really implementable - you would want to be able to have multiple instances of `elif:` which isn't possible in YAML.

clydebarrow:
```suggestion
```

jesserockz: Potential config structure?
A list evaluated in order from top to bottom as if they were all elifs, stopping at the first to resolve to true.

(But not for this PR)

```yaml
- if:
    - condition: ... 
      then:
        - ...
    - condition: ... #elif
      then: 
        - ...
```

jesserockz (approved):

esphome/const.py
clydebarrow: Would not allow for an `else` - but maybe something like this would:

```
if:
	select:           # or maybe first_of or something else
		- condition:
			then:
		- condition:
			then:
	else:

guillempages: Maybe we should try doing something like ha is doing with "choice"? So having a specific keyword instead of if/elif? 

Like a switch in C , although I'd prefer to keep the same syntax as homeassistant

clydebarrow: > Maybe we should try doing something like ha is doing with "choice"?

Do you have a reference for that? I can't find anything in the HA docs, not that that means anything, it's a bit of a dog's breakfast, quite hard to navigate I find.

guillempages: Yes,  I also always have trouble finding that. What I meant is https://www.home-assistant.io/docs/scripts/#choose-a-group-of-actions

clydebarrow: Sure, `choose` and `default` would work. But not `conditions` since a) it's ambiguous, and b) we have `condition` and now `all` and `any`.

I'll work something up for review after this PR is merged, unless someone else wants to do it.

PR #7583, opened 2024-10-11
Enable text wrapping for printf on displays

davet2001:
For matrix style graphical displays, sometimes there is a need to display a block of text.
The exact newline points depend on the text, and this might vary depending on sensors.

This PR implements the ability to printf to a text box defined by its x, y, coordinates and width,height.

It automatically wraps words onto new lines.
When the text would vertically overspill the box provided, the final characters are replaced by '...'.

It has the ability to set the font and adjust the line spacing from the nominal value of 1.0.

image

Thanks to @alvesvaren who provided the initial version of this and agreed to allow it to be added to esphome core.
https://gist.github.com/alvesvaren/767d9585f0ecbef18ef1c7c0492c4332

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

display:
  - platform: st7735
    model: "INITR_MINI160X80"
    reset_pin: 1
    cs_pin: 4
    dc_pin: 2
    rotation: 270
    device_width: 82
    device_height: 161
    col_start: 0
    row_start: 0
    eight_bit_color: true
    update_interval: 5s
    invert_colors: True
    use_bgr: True
    lambda: |-
      int x = 0;
      int y = 0;
      int w = 120;
      int h = 80;
      it.fill(Color::BLACK);
      auto green = Color(0, 255, 0);
      it.rectangle(x, y, w, h, green);
      it.printf(1, 1, w-2, h-2, id(font_roboto), 1.0, "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.");
```

## Checklist:

### DIFF
--- a/esphome/components/display/display.cpp
+++ b/esphome/components/display/display.cpp
@@ -376,7 +376,7 @@ void Display::print(int x, int y, int width, int height, BaseFont *font, Color c
       this->get_text_bounds(line_x, line_y, w.c_str(), font, align, &word_x1, &word_y1, &word_width, &word_height);
 
       if (line_x + word_width >= x_max) {
-        // If the next line would overspill the height box, get ready to stop.
+        // If the next line would overspill the height box, get ready to stop
         if (line_y + 2 * static_cast<int>(word_height * line_height) > y_max) {
           last = true;
           size_t length = line.length();
--- a/esphome/components/display/display.cpp
+++ b/esphome/components/display/display.cpp
@@ -346,7 +346,7 @@ void Display::print(int x, int y, BaseFont *font, Color color, TextAlign align, 
 void Display::print(int x, int y, int width, int height, BaseFont *font, Color color, TextAlign align, const char *text,
                     Color background, float line_height, bool wrap) {
   int x_start, y_start;
-  if (wrap == false) {
+  if (!wrap) {
     this->get_text_bounds(x, y, text, font, align, &x_start, &y_start, &width, &height);
     font->print(x_start, y_start, this, color, text, background);
   } else {
--- a/esphome/components/display/display.cpp
+++ b/esphome/components/display/display.cpp
@@ -496,6 +496,7 @@ void Display::printf(int x, int y, int width, int height, BaseFont *font, float 
 
   char buffer[256];
   int ret = vsnprintf(buffer, sizeof(buffer), format, arg);
+  va_end(arg);
   if (ret == 0)
     return;
 
@@ -555,8 +556,6 @@ void Display::printf(int x, int y, int width, int height, BaseFont *font, float 
   if (!line.empty()) {
     this->printf(x, line_y, font, align, "%s", line.c_str());
   }
-
-  va_end(arg);
 }
 
 void Display::set_writer(display_writer_t &&writer) { this->writer_ = writer; }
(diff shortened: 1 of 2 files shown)


nielsnl68:
This is a very nice idea. Are you willing to add this to the other Print methods as well?

An alternative could be to use the clipping function to tell the size of the box and and add a property to set the wrapping mode for all print methods.

davet2001:
> This is a very nice idea. Are you willing to add this to the other Print methods as well?

@nielsnl68 Ok. I've just moved the wrap capability to the lowest level print function. If I have interpreted this correctly, it should now be possible to build up wrap capability into other functions by overloading them. Is that what you meant?

> An alternative could be to use the clipping function to tell the size of the box and and add a property to set the wrapping mode for all print methods.

Interesting, although for such a universal solution, I can see someone potentially wanting clipping and wrapping to be separate. For example, if someone was implementing a wipe/reveal using clipping, then they might want to have the text wrap to the full screen width, but the clipping box gradually increase width with successive frames.

Combining the clip/wrap box or forcing the user to choose between wrap or clip would prevent them doing this.

Happy to discuss further if I have misunderstood.

nielsnl68:
I was thinking something like:

```
   set_clip(10,10,100,100)
   print("this code is just clipped");
   set_wrap(true);
   print("this is wrapped inside the clipped region");
   set_wrap(false);
```
This is a bit of simplified code. But i'm sure you get what I mean here.
Not sure if i understand the case you addressed does that not fit the above idea?

davet2001:
Probably there are ways around most limitations, and we are talking about smaller and smaller edge cases, so potentially not worth stalling over.

The case I was trying to describe was a wrapped block of text in a box, which someone is then trying to print a smaller portion of using the clipping feature, perhaps during a transition. In that case the wrap box needs to be fixed at the full width, but the clip box needs to be adjusted each frame of the transition. When the transition is half way through it might look like this picture:
image
But that is impossible to do if clip box and wrap box are locked together.

nielsnl68:
Good point there.

Using my clipping suggestion is not a good idea.

davet2001:
But could work just as well with a `set_wrap(10,10,100,100)` that could be set outside of the print function.

I do think there is quite an advantage in not having to add lots of parameters to the print functions to enable wrapping.

Also having the wrap/clip bounds defined separately gives scope for them not being a rectangle in the future too. Potentially could wrap multiline text inside a circle for example.

I'll take another look in the next few days.

nielsnl68:
yeah that would work better, i think.

davet2001:
Not stale, still intending to finish this off.

davet2001:
Not stale

davet2001:
Not stale

davet2001:
Still wanted

PR #7585, opened 2024-10-11
Add a core audio struct

kahrendt:

Adds a new struct to the core to describe an audio stream. This will be used to help communicate the audio being passed from one component to another; e.g., a media player component will send audio to a speaker component, and this struct will let the speaker know what format the audio is in.

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/core/audio.h
+++ b/esphome/core/audio.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <cstdint>
+#include <stddef.h>
+
+namespace esphome {
+
+struct AudioStreamInfo {
+  bool operator==(const AudioStreamInfo &rhs) const {
+    return (channels == rhs.channels) && (bits_per_sample == rhs.bits_per_sample) && (sample_rate == rhs.sample_rate);
+  }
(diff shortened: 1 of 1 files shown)


tomaszduda23:
You could have put it in new component. core is hard to change since you cannot use external components.

kahrendt:
> You could have put it in new component. core is hard to change since you cannot use external components.

Multiple components may need this structure. This is still a draft, as I am still preparing an update to the ``i2s_audio`` speaker that uses this. The Nabu media player component (current version implemented [here](https://github.com/esphome/home-assistant-voice-pe/)) will be adjusted to write to any speaker component, including the updated ``i2s_audio`` version I'm preparing, and it will also rely on having this globally available.

kahrendt:
> You could have put it in new component. core is hard to change since you cannot use external components.

Thanks for the idea, sorry I misunderstood at first! Jesse explained to me that I could just auto load it in any other component that needs it, so that's the route I'm going to use.

PR #7591, opened 2024-10-12
[touchscreen] Fix coordinates when using rotation

clydebarrow:

When the display is rotated in software by 90 or 270 the x and y axes are swapped. This is managed by setting `transform` in the touchscreen config, and this works, however when the result is scaled to the display size, the scaling is currently done with the native (unrotated) screen dimensions whereas it should be done using the effective screen dimensions.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,7 +315,9 @@ jobs:
           key: platformio-${{ matrix.pio_cache_key }}
 
       - name: Install clang-tidy
-        run: sudo apt-get install clang-tidy-14
+        run: |
+          sudo apt-get update
+          sudo apt-get install clang-tidy-14
 
       - name: Register problem matchers
         run: |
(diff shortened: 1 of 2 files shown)


nielsnl68:
when you want to use the the effective screen dimensions make sure that the display is setup first. 
And make sure that all the possible configurations still work. ;) But knowing you you did that already.

clydebarrow:
> when you want to use the the effective screen dimensions make sure that the display is setup first

Good point but it actually only needs to be configured, not setup, since the dimensions and rotation (which are used to derive the effective dimensions) are set during configuration. Configuration of all components is guaranteed before setup of any so it will be fine.

All rotations have been tested.

jesserockz (approved):

PR #7602, opened 2024-10-14
[code-quality] udp component

tomaszduda23:

part of https://github.com/esphome/esphome/pull/7049. There is `DEPENDENCIES = ["network"]` so `USE_NETWORK`

```
/config/esphome/esphome/components/udp/udp_component.h:13:10: error: 'WiFiUdp.h' file not found [clang-diagnostic-error]
#include <WiFiUdp.h>
         ^~~~~~~~~~~
```

## Types of changes

- [x] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/udp/udp_component.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "esphome/core/component.h"
+#ifdef USE_NETWORK
```
jesserockz:
```suggestion
#include "esphome/core/defines.h"
#ifdef USE_NETWORK
```

clydebarrow (commented):
Is there any need for this given that network is a dependency of udp, so USE_NETWORK will always be defined? I mean, how does the file get included if it's not in use?

jesserockz:
> Is there any need for this given that network is a dependency of udp, so USE_NETWORK will always be defined? I mean, how does the file get included if it's not in use?

These files are causing CI failures in #7049 because there is no `network` when it is running clang on that env, but it still goes through all files.

clydebarrow:
Right. I guess a better scheme would be to use the exclusions set in the CONFIG_SCHEMA to constrain clang-tidy to files actually In use for a given platform, but that's a bit harder.

clydebarrow:
Would it not be better to do this?

```
#if defined(USE_SOCKET_IMPL_BSD_SOCKETS) || defined(USE_SOCKET_IMPL_LWIP_SOCKETS)
#include "esphome/components/socket/socket.h"
# endif
#ifdef USE_SOCKET_IMPL_LWIP_TCP
#include <WiFiUdp.h>
#endif

```

tomaszduda23:
> Would it not be better to do this?
> 
> ```
> #if defined(USE_SOCKET_IMPL_BSD_SOCKETS) || defined(USE_SOCKET_IMPL_LWIP_SOCKETS)
> #include "esphome/components/socket/socket.h"
> # endif
> #ifdef USE_SOCKET_IMPL_LWIP_TCP
> #include <WiFiUdp.h>
> #endif
> ```

I would keep it with USE_NETWORK. Global ifdef makes code a little bit cleaner. There is effort to add OpenThread https://github.com/esphome/esphome/pull/7506. It is applicable to nrf52840 also. One day we could have socket implementation for zephyr. Right now it makes no difference.

clydebarrow:
Sorry, I have a bad habit of presenting conclusions without showing the reasoning behind them. Let me explain:

* Your problem is that your work on adding a new platform is being stepped on by a rather clumsy CI check that, unlike the ESPHome application itself, can't currently determine what components are valid for a given platform;
* You've quite reasonably proposed a quick fix - comment out the entire file if the platform doesn't have network support at all (but titled the PR "Code Quality" which isn't strictly correct, yet.)
* In reviewing that I realised that the current ifdefs make the assumption that if the available network stack isn't BSD or LWIP, then it is LWIP_TCP. This is true right now but may not be true in the future.

So there *is* a code quality issue here, which should be fixed, i.e. any code dependent on the presence of a specific feature should be surrounded by ifdefs based on that feature, not assumed as a default.

Now as it happens fixing this particular code quality issue should also solve *your* problem as in the first point above and render the band-aid fix redundant. I'd see that as a better approach - solve your immediate problem and improve the code in one hit.

I'm happy to rework the PR with the necessary changes if you like, especially given I created the code-quality issue in the first place.

tomaszduda23:
updated

clydebarrow (approved):

PR #7604, opened 2024-10-14
Add log level env var

briancw:
Adds a LOG_LEVEL environment variable for setting the log level on the dashboard.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4341

## Checklist:

### DIFF
--- a/esphome/dashboard/web_server.py
+++ b/esphome/dashboard/web_server.py
@@ -52,6 +52,8 @@ if TYPE_CHECKING:
     from requests import Response
 
 
+LOG_LEVEL = os.getenv("LOG_LEVEL", "INFO").upper()
+logging.getLogger().setLevel(LOG_LEVEL)
 _LOGGER = logging.getLogger(__name__)
 
 ENV_DEV = "ESPHOME_DASHBOARD_DEV"
--- a/esphome/dashboard/web_server.py
+++ b/esphome/dashboard/web_server.py
@@ -52,7 +52,7 @@ if TYPE_CHECKING:
     from requests import Response
 
 
-LOG_LEVEL = os.getenv("LOG_LEVEL", "INFO").upper()
+LOG_LEVEL = os.getenv("ESPHOME_LOG_LEVEL", "INFO").upper()
 logging.getLogger().setLevel(LOG_LEVEL)
 _LOGGER = logging.getLogger(__name__)
 
--- a/esphome/dashboard/web_server.py
+++ b/esphome/dashboard/web_server.py
@@ -52,8 +52,6 @@ if TYPE_CHECKING:
     from requests import Response
 
 
-LOG_LEVEL = os.getenv("ESPHOME_LOG_LEVEL", "INFO").upper()
-logging.getLogger().setLevel(LOG_LEVEL)
 _LOGGER = logging.getLogger(__name__)
 
 ENV_DEV = "ESPHOME_DASHBOARD_DEV"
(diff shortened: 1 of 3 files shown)


esphome/dashboard/web_server.py
```diff
@@ -52,6 +52,8 @@
     from requests import Response
 
 
+LOG_LEVEL = os.getenv("LOG_LEVEL", "INFO").upper()
```
jesserockz: I think we should prefix this
```suggestion
LOG_LEVEL = os.getenv("ESPHOME_LOG_LEVEL", "INFO").upper()
```

jesserockz (changes requested):

esphome/dashboard/web_server.py
briancw: Makes perfect sense to me

esphome/dashboard/web_server.py
```diff
@@ -52,6 +52,8 @@
     from requests import Response
 
 
+LOG_LEVEL = os.getenv("ESPHOME_LOG_LEVEL", "INFO").upper()
+logging.getLogger().setLevel(LOG_LEVEL)
```
jesserockz: Just realised that this is probably not the correct file for this code. It should really be a part of `setup_log` in `esphome/log.py`

briancw: That makes sense. After diving in from there, I have a better idea of how logging is currently setup. Currently its not possible to set the log level to WARNING or ERROR.
Having both an environment variable and the existing flags would probably be somewhat confusing and overlapping. Would it be better to add new arguments, "-w", "--warning", "-e", "--error" to be able to target those log levels. Or to remove the existing "-q" and "-v" options and just have a single log level argument?

briancw: Personally I would opt for a single log level flag.

__main__.py
```suggestion
options_parser.add_argument(
    "-l",
    "--log-level",
    help="Set the log level.",
    default="INFO",
    action="store",
    choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
)
```

log.py
```suggestion
def setup_log(
    log_level = logging.INFO, include_timestamp: bool = False
) -> None:
    import colorama

    colorama.init()

    if log_level == logging.DEBUG:
        CORE.verbose = True
    elif log_level == logging.CRITICAL:
        CORE.quiet = True

    logging.basicConfig(level=log_level)

    logging.getLogger("urllib3").setLevel(logging.WARNING)

    logging.getLogger().handlers[0].setFormatter(
        ESPHomeLogFormatter(include_timestamp=include_timestamp)
    )
```

ptr727: Will this replace or work with `ESPHOME_VERBOSE` that sets `--verbose`?

briancw: Having  single flag --log-level, and a single env var ESPHOME_LOG_LEVEL would be the most intuitive in my mind. But going the other direction would also work. Keep ESPHOME_VERBOSE and add ESPHOME_INFO ESPHOME_WARN ESPHOME_ERROR and ESPHOME_CRITICAL.

ptr727: I'd vote for setting the log level, maybe retain `--verbose` as a verbose flag for CLI is pretty common, but `ESPHOME_VERBOSE` was just recently added, so disruptive but not the end of the world to deprecate. it in favor of a log level env.

briancw: I left the verbose and quiet flags in as well as the ESPHOME_VERBOSE env var so as to not be a breaking change. I added "--log-level" and "-l" as flags as well as the "ESPHOME_LOG_LEVEL" env var.

WagnerSystems:
Hello,
is there still work in progress here? I really would like to have such a feature.

jesserockz (approved):

PR #7605, opened 2024-10-14
[speaker, i2s_audio] I2S Speaker implementation using a ring buffer

kahrendt:

This is a near complete rewrite of the ``i2s_audio`` speaker component with improvements for simplification, stability, and supporting different audio sample rates and bits per sample. It adds to the ``speaker`` platform to support these new features while avoiding breaking changes. It adds a new ``audio`` component that is auto-loaded to handle passing different audio stream settings. Other audio components can auto-load it in the future if needed. Closes #7137, as this implements the core changes (the ESPHome ring buffer uses xStreamBuffers).

- Speaker platform supports writing audio with a FreeRTOS ``ticks_to_wait`` parameter for flexibility when using a separate task to write to the speaker.
- Speaker platform supports setting the volume with a new action. Each child speaker component needs to implement its own implementation.
  - The ``i2s_audio`` speaker component implements this in software.
- Transfers all audio data via a ring buffer instead of using a FreeRTOS queue.
  - The ring buffer deallocates when stopping, so there isn't an increase in memory use when idle.
- Uses a FreeRTOS Event Group for communicating between the speaker task and the main loop.
  - Simplifies the state and error reporting and commands.
  - Used to lock writing to the ring buffer, as FreeRTOS Stream buffers should only have 1 task writing at a time. Also ensures the task doesn't deallocate the ring buffer while a write is in progress.
  - Implements the existing stop logic: ``stop()`` ends it immediately, ``finish()`` ends the task after the ring buffer is emptied, and it will end the task with the configured timeout if no audio has been read after that time.
- Increases the default timeout to 500 ms to avoid occasional issues with delayed TTS responses.
- Simplifies the logic for writing to the I2S bus. 
  - Uses ``i2s_write_expand`` for converting between lower bits per sample to higher bits per sample.
  - Uses ``i2s_set_clk`` for handling mono and stereo audio
- Adds support for reconfiguring the I2S settings by the component sending audio with the ``set_audio_stream_info`` function.
  - This enables support for audio with more than 16 bits per sample, a sample rate higher than 16000 Hz, and more than 1 channel - the previous component always assumed that particular format.
  - Should give an error if the sending component tries to set an incompatible audio stream instead of playing distorted audio, though a component needs to implement this call if it is sending non-default setting audio.
- Documents the various new (and old) functions in detail. 

The component should be much more resilient to errors and should hopefully avoid ever getting in a stuck state that locks the device. It should stop the task and unlock the I2S port if it can't recover, but future playback should work without needing to reboot. My ATOM Echo is much more reliable and stable with this implementation. I additionally tested the changes to the parent speaker component with an S3 Box Lite using the ``esp_adf`` speaker with no issues. I would appreciate anyone able to test this on other hardware!

I'm keeping as a draft until I add documentation for the ``speaker.volume_set`` action.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4343

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

esphome:
  on_boot:
    - speaker.volume_set: 0.9


speaker:
  - platform: i2s_audio
    id: echo_speaker
    i2s_dout_pin: GPIO22
    dac_type: external
    bits_per_sample: 32bit
    channel: right
    timeout: 500ms

```

## Checklist:

### DIFF
--- a/esphome/components/audio/audio.h
+++ b/esphome/components/audio/audio.h
@@ -1,0 +1,21 @@
+#pragma once
+
+#include <cstdint>
+#include <stddef.h>
+
+namespace esphome {
+namespace audio {
+
+struct AudioStreamInfo {
+  bool operator==(const AudioStreamInfo &rhs) const {
+    return (channels == rhs.channels) && (bits_per_sample == rhs.bits_per_sample) && (sample_rate == rhs.sample_rate);
+  }
+  bool operator!=(const AudioStreamInfo &rhs) const { return !operator==(rhs); }
+  size_t get_bytes_per_sample() const { return bits_per_sample / 8; }
+  uint8_t channels = 1;
+  uint8_t bits_per_sample = 16;
+  uint32_t sample_rate = 16000;
+};
+
+}  // namespace audio
+}  // namespace esphome
--- a/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
+++ b/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
@@ -141,29 +141,6 @@ void I2SAudioSpeaker::loop() {
   }
 }
 
-void I2SAudioSpeaker::start() {
-  if (this->is_failed())
-    return;
-  if ((this->state_ == speaker::STATE_STARTING) || (this->state_ == speaker::STATE_RUNNING))
-    return;
-
-  if (this->speaker_task_handle_ == nullptr) {
-    xTaskCreate(I2SAudioSpeaker::speaker_task, "speaker_task", TASK_STACK_SIZE, (void *) this, TASK_PRIORITY,
-                &this->speaker_task_handle_);
-  }
-
-  if (this->speaker_task_handle_ != nullptr) {
-    xEventGroupSetBits(this->event_group_, SpeakerEventGroupBits::COMMAND_START);
-    this->task_created_ = true;
-  } else {
-    xEventGroupSetBits(this->event_group_, SpeakerEventGroupBits::ERR_TASK_FAILED_TO_START);
-  }
-}
-
-void I2SAudioSpeaker::stop() { this->stop_(false); }
-
-void I2SAudioSpeaker::finish() { this->stop_(true); }
-
 void I2SAudioSpeaker::set_volume(float volume) {
   this->volume_ = volume;
   ssize_t decibel_index = remap<ssize_t, float>(volume, 0.0f, 1.0f, 0, Q15_VOLUME_SCALING_FACTORS.size() - 1);
@@ -228,16 +205,9 @@ void I2SAudioSpeaker::speaker_task(void *params) {
 
   const size_t dma_buffers_size = FRAMES_IN_ALL_DMA_BUFFERS * bytes_per_sample * number_of_channels;
 
-  ExternalRAMAllocator<uint8_t> allocator(ExternalRAMAllocator<uint8_t>::ALLOW_FAILURE);
-  this_speaker->data_buffer_ = allocator.allocate(dma_buffers_size);
-
-  if (this_speaker->audio_ring_buffer_ == nullptr) {
-    // Allocate ring buffer
-    this_speaker->audio_ring_buffer_ = RingBuffer::create(RING_BUFFER_SAMPLES * bytes_per_sample);
-  }
-
-  if ((this_speaker->data_buffer_ == nullptr) || (this_speaker->audio_ring_buffer_ == nullptr)) {
-    // Failed to allocate a buffer
+  if (this_speaker->send_esp_err_to_event_group_(
+          this_speaker->allocate_buffers_(dma_buffers_size, RING_BUFFER_SAMPLES * bytes_per_sample))) {
+    // Failed to allocate buffers
     xEventGroupSetBits(this_speaker->event_group_, SpeakerEventGroupBits::ERR_ESP_NO_MEM);
     this_speaker->delete_task_(dma_buffers_size);
   }
@@ -262,7 +232,6 @@ void I2SAudioSpeaker::speaker_task(void *params) {
       event_group_bits = xEventGroupGetBits(this_speaker->event_group_);
 
       if (event_group_bits & SpeakerEventGroupBits::COMMAND_STOP) {
-        i2s_zero_dma_buffer(this_speaker->parent_->get_port());
         break;
       }
       if (event_group_bits & SpeakerEventGroupBits::COMMAND_STOP_GRACEFULLY) {
@@ -298,14 +267,17 @@ void I2SAudioSpeaker::speaker_task(void *params) {
 
       } else {
         // No data received
-        i2s_zero_dma_buffer(this_speaker->parent_->get_port());
 
         if (stop_gracefully) {
           break;
         }
+
+        i2s_zero_dma_buffer(this_speaker->parent_->get_port());
       }
     }
   }
+  i2s_zero_dma_buffer(this_speaker->parent_->get_port());
+
   xEventGroupSetBits(this_speaker->event_group_, SpeakerEventGroupBits::STATE_STOPPING);
 
   i2s_stop(this_speaker->parent_->get_port());
@@ -314,6 +286,29 @@ void I2SAudioSpeaker::speaker_task(void *params) {
   this_speaker->parent_->unlock();
   this_speaker->delete_task_(dma_buffers_size);
 }
+
+void I2SAudioSpeaker::start() {
+  if (this->is_failed())
+    return;
+  if ((this->state_ == speaker::STATE_STARTING) || (this->state_ == speaker::STATE_RUNNING))
+    return;
+
+  if (this->speaker_task_handle_ == nullptr) {
+    xTaskCreate(I2SAudioSpeaker::speaker_task, "speaker_task", TASK_STACK_SIZE, (void *) this, TASK_PRIORITY,
+                &this->speaker_task_handle_);
+  }
+
+  if (this->speaker_task_handle_ != nullptr) {
+    xEventGroupSetBits(this->event_group_, SpeakerEventGroupBits::COMMAND_START);
+    this->task_created_ = true;
+  } else {
+    xEventGroupSetBits(this->event_group_, SpeakerEventGroupBits::ERR_TASK_FAILED_TO_START);
+  }
+}
+
+void I2SAudioSpeaker::stop() { this->stop_(false); }
+
+void I2SAudioSpeaker::finish() { this->stop_(true); }
 
 void I2SAudioSpeaker::stop_(bool wait_on_empty) {
   if (this->is_failed())
@@ -348,6 +343,29 @@ bool I2SAudioSpeaker::send_esp_err_to_event_group_(esp_err_t err) {
       xEventGroupSetBits(this->event_group_, SpeakerEventGroupBits::ERR_ESP_FAIL);
       return true;
   }
+}
+
+esp_err_t I2SAudioSpeaker::allocate_buffers_(size_t data_buffer_size, size_t ring_buffer_size) {
+  if (this->data_buffer_ == nullptr) {
+    // Allocate data buffer for temporarily storing audio from the ring buffer before writing to the I2S bus
+    ExternalRAMAllocator<uint8_t> allocator(ExternalRAMAllocator<uint8_t>::ALLOW_FAILURE);
+    this->data_buffer_ = allocator.allocate(data_buffer_size);
+  }
+
+  if (this->data_buffer_ == nullptr) {
+    return ESP_ERR_NO_MEM;
+  }
+
+  if (this->audio_ring_buffer_ == nullptr) {
+    // Allocate ring buffer
+    this->audio_ring_buffer_ = RingBuffer::create(ring_buffer_size);
+  }
+
+  if (this->audio_ring_buffer_ == nullptr) {
+    return ESP_ERR_NO_MEM;
+  }
+
+  return ESP_OK;
 }
 
 esp_err_t I2SAudioSpeaker::start_i2s_driver_() {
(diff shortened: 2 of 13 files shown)


esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -4,291 +4,447 @@
...
 }
 
-void I2SAudioSpeaker::player_task(void *params) {
+bool I2SAudioSpeaker::has_buffered_data() const {
+  if (this->audio_ring_buffer_ != nullptr) {
+    return this->audio_ring_buffer_->available() > 0;
+  }
+  return false;
+}
+
```
nielsnl68, line 184:
```suggestion
size_t I2SAudioSpeaker::buffer_available() const {
  if (this->audio_ring_buffer_ != nullptr) {
    return this->audio_ring_buffer_->available();
  }
  return -1;
}
```

I love to have a method that tells the real available buffer space as well. so you do not need to send more data then available in the speaker buffer.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -4,291 +4,447 @@
...
+      xEventGroupWaitBits(this_speaker->event_group_,
+                          SpeakerEventGroupBits::COMMAND_START | SpeakerEventGroupBits::COMMAND_STOP |
+                              SpeakerEventGroupBits::COMMAND_STOP_GRACEFULLY,  // Bit message to read
+                          pdTRUE,                                              // Clear the bits on exit
+                          pdFALSE,                                             // Don't wait for all the bits,
+                          portMAX_DELAY);                                      // Block indefinitely until a bit is set
+
+  if (event_group_bits & (SpeakerEventGroupBits::COMMAND_STOP | SpeakerEventGroupBits::COMMAND_STOP_GRACEFULLY)) {
+    // Received a stop signal before the task was requested to start
+    this_speaker->delete_task_(0);
```
nielsnl68, line 197:
```suggestion
    this_speaker->delete_task_(0);
    return; // or a wait loop;
```
Just to be sure that the next code is not executed anymore. 
The same for all other places `this_speaker->delete_task_()` is called.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -4,291 +4,447 @@
...
+  } else {
+    xEventGroupSetBits(this->event_group_, SpeakerEventGroupBits::ERR_TASK_FAILED_TO_START);
   }
+}
 
-  xTaskCreate(I2SAudioSpeaker::player_task, "speaker_task", 8192, (void *) this, 1, &this->player_task_handle_);
-  this->task_created_ = true;
+void I2SAudioSpeaker::stop() { this->stop_(false); }
+
+void I2SAudioSpeaker::finish() { this->stop_(true); }
```
nielsnl68: Could you move this two above the `this->stop_();` for easy reading.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -4,291 +4,447 @@
...
-    event.type = TaskEventType::WARNING;
-    event.err = err;
-    xQueueSend(this_speaker->event_queue_, &event, 0);
-    event.type = TaskEventType::STOPPED;
-    xQueueSend(this_speaker->event_queue_, &event, 0);
-    while (true) {
-      delay(10);
-    }
+  const size_t dma_buffers_size = FRAMES_IN_ALL_DMA_BUFFERS * bytes_per_sample * number_of_channels;
+
```
nielsnl68, line 207: Should we not check if the buffer does not exist before we allocating new space here. Just to be sure that the last session finish properly?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -4,291 +4,447 @@
...
-  if (xQueueSend(this_speaker->event_queue_, &event, 10 / portTICK_PERIOD_MS) != pdTRUE) {
-    ESP_LOGW(TAG, "Failed to send STOPPED event");
-  }
+        if (bytes_written != bytes_read) {
+          xEventGroupSetBits(this_speaker->event_group_, SpeakerEventGroupBits::ERR_ESP_INVALID_SIZE);
+        }
+
+      } else {
+        // No data received
+        i2s_zero_dma_buffer(this_speaker->parent_->get_port());
```
nielsnl68: What happens when the speaker is still playing the last part of the data stream. Will this not stop it before it finished?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -4,291 +4,447 @@
...
+  } else if (audio_stream_info.channels == 2) {
+    return i2s_set_clk(this->parent_->get_port(), this->sample_rate_, this->bits_per_sample_, I2S_CHANNEL_STEREO);
   }
-  return index;
+
+  return ESP_ERR_INVALID_ARG;
 }
 
-bool I2SAudioSpeaker::has_buffered_data() const { return uxQueueMessagesWaiting(this->buffer_queue_) > 0; }
+void I2SAudioSpeaker::delete_task_(size_t buffer_size) {
```
nielsnl68, line 458: Would it not cleaner when we move the creation of the buffers  into `void I2SAudioSpeaker::create_task_(size_t buffer_size)' here as well? Instead of setting them up in the `speaker_task()` method?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -4,291 +4,447 @@
...
+        }
+      }
+    }
   }
-}
+  xEventGroupSetBits(this_speaker->event_group_, SpeakerEventGroupBits::STATE_STOPPING);
 
-void I2SAudioSpeaker::stop() { this->stop_(false); }
+  i2s_stop(this_speaker->parent_->get_port());
+  i2s_driver_uninstall(this_speaker->parent_->get_port());
```
nielsnl68:
```suggestion
  size_t bytes_written;
  uint32_t sample = 0;
  i2s_write(this->parent_->get_port(), &sample, 4, &bytes_written, portMAX_DELAY);

  i2s_stop(this_speaker->parent_->get_port());
  i2s_driver_uninstall(this_speaker->parent_->get_port());
```
to be make sure that the speaker output does not stay at an high/low level and burn away the speaker (it happened 3 times for me testing the RTTTL component).

nielsnl68 (changes requested):
Hi @kahrendt,

I like the new approach of the speaker component. 
I added a couple of suggestions below. 

Anyway it looks much better then what i tried to do with it.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: When the last bit of audio is copied out of the ring buffer into ``data_buffer_,`` it will then run the ``i2s_write`` or ``i2s_write_expand`` function. That has a delay of ``portMAX_DELAY``, meaning it will block indefinitely until all the data has been written to the I2S DMA buffers. Then on the next loop, no data is read from the ring buffer, so then this code executes. The ``i2s_zero_dma_buffer`` only adds zeros to the DMA buffer until it is full. It won't overwrite existing data in the DMA buffer, so the last bit of audio should play before outputting zeros.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: This is unnecessary, as deleting the task immediately stops it from executing. Espressif specifies that a task function should never return or it should be terminated using the vTaskDelete function. They even follow my exact approach in some of the [esp-idf code](https://github.com/espressif/esp-idf/blob/a97a7b0962da148669bb333ff1f30bf272946ade/examples/protocols/mqtt/ssl_ds/main/app_main.c#L102). There isn't a return function after vTaskDelete, even though there is more code after it.

nielsnl68:
@kahrendt could you show how to use the new futures in **CPP** components as well?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: I believe a better practice would be to use the ``ticks_to_wait`` parameter in the audio sending code. The sending code will block (let other tasks run) while it waits to be able to write all the data to the ring buffer. The sending code just has to handle internally that not all the data may be sent when calling the play function. I know the voice assistant component already handles this.

Overall I'm hesitant to add this because each speaker component would have to individually implement this, and I can't think of a safe way to provide a default function that wouldn't break any component that relies on it being correct. In contrast, the play function with ticks_to_wait and the finish function can all provide a safe default function if it isn't overridden by the component.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: I reorganized the code to follow this order.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: I added an ``allocate_buffers_`` function for this. It does simplify the ``speaker_task`` code, thanks!

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: I added this check to the ``allocate_buffers_`` function.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: I added an ``i2ss_zero_dma_buffer`` before stopping and uninstalling the driver, so it should effectively do this.

kahrendt:
> @kahrendt could you show how to use the new futures in **CPP** components as well?

I don't have an easy example readily available, but the nabu component in the ``kahrendt-i2s-audio-approach`` branch [home-assistant-voice-pe](https://github.com/esphome/home-assistant-voice-pe) calls the new play function in ``audio_mixer.cpp`` in the ``audio_mixer_task`` function. The ``set_audio_stream_info`` function is called in the ``nabu_media_player.cpp`` file in the same branch.

Thanks for all the feedback!

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
nielsnl68: Hmm, i see you have well toughed about it. I belief your approach is indeed better.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
nielsnl68: When you 💯% sure :D then it is okay by me.

esphome/components/audio/audio.h
```diff
@@ -0,0 +1,21 @@
...
+namespace esphome {
+namespace audio {
+
+struct AudioStreamInfo {
+  bool operator==(const AudioStreamInfo &rhs) const {
+    return (channels == rhs.channels) && (bits_per_sample == rhs.bits_per_sample) && (sample_rate == rhs.sample_rate);
+  }
+  bool operator!=(const AudioStreamInfo &rhs) const { return !operator==(rhs); }
+  size_t get_bytes_per_sample() const { return bits_per_sample / 8; }
+  uint8_t channels = 1;
```
nielsnl68, line 15: what happens when you configure the speaker with on speaker and the using component is set to 2 channels?
I hope i'm just missing something here?

kahrendt: It automatically writes the incoming one channel audio sample to both channels.

kahrendt: I realized you maybe meant the other way around. I tested this by sending stereo audio (set using the ``set_audio_stream_info``) and using the various channels options. It seems regardless of whether you set ``channel`` to ``mono``, ``left``, ``right``, or ``stereo``, it will still send the data to both channels. If you only have one channel hooked up, then it will still receive the audio.

Espressif's documentation isn't very clear on the exact behavior of the various channel formats and the ``i2s_set_clk`` function. In my tests though, whether sending stereo or mono channel audio to the speaker, it plays it back correctly; i.e., it doesn't play it double speed or anything like that.

nielsnl68: Cool, that is what i wanted to know. Now we know we do not be worried anymore.

nielsnl68 (approved):

nielsnl68:
Keven, about the speaker_task(), what you could do is the following:

```cpp
class i2s_audio_speaker .... {
  static void speaker_task(void *params);
protected:
  void run_speaker_task_();
}

}

static void i2s_audio_speaker::speaker_task(void *params) {
  I2SAudioSpeaker *this_speaker = (I2SAudioSpeaker *) params;
  this_speaker->run_speaker_task_();
}

```

This work very well on my espnow component.

jesserockz:
> Keven, about the speaker_task(), what you could do is the following

I actually think it is better to keep the task running code in the static function, it reminds you that by using `this_speaker` instead of just `this` that you are in a task and shouldn't do unsafe operations.

jesserockz (approved):

nielsnl68:
> > Keven, about the speaker_task(), what you could do is the following
> 
> I actually think it is better to keep the task running code in the static function, it reminds you that by using `this_speaker` instead of just `this` that you are in a task and shouldn't do unsafe operations.

Okay, then i will update my espnow component as well.

PR #7611, opened 2024-10-16
[lvgl] Revise code generation to allow early widget creation

clydebarrow:

The LVGL component relied on the code generation creating lambdas which contained all the code for creating widgets and which were executed during the LVGL `setup()` function. The reason for this was that the LVGL library could not be initialised until display parameters (dimensions and rotation) were known, which had to wait until after the display was setup.

A major issue with this was the occurrence of calls  to modify state of widgets (e.g. from a switch with a restore mode, linked to a widget, trying to set state during the switch `setup()`). This was worked-around by wrapping all such calls in tests for LVGL initialisation being complete to avoid null pointer references. However this meant some state synchronisation could be lost.

However It transpires that it is not necessary to wait until the display is setup, merely to wait until it is configured since all LVGL needs initially is to get the display parameters. This will be guaranteed by the code generation process with its coroutine structure.

Accordingly the basic initialisation of LVGL is now done in its constructor, allowing all widgets to be created in the ESPHome `main()` function before any component is setup and avoiding the need for the null pointer test wrappers.

## Types of changes

- [x] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/lvgl/__init__.py
```diff
@@ -200,7 +200,14 @@ async def to_code(config):
...
     add_define("LV_MEM_CUSTOM_INCLUDE", '"esphome/components/lvgl/lvgl_hal.h"')
 
-    add_define("LV_LOG_LEVEL", f"LV_LOG_LEVEL_{config[df.CONF_LOG_LEVEL]}")
+    add_define(
+        "LV_LOG_LEVEL", f"LV_LOG_LEVEL_{df.LV_LOG_LEVELS[config[df.CONF_LOG_LEVEL]]}"
+    )
+    # Doesn't seem to be a way to use cg.add_define to add a definition for a literal value
+    CORE.add_define(
+        Define("LVGL_LOG_LEVEL", f"ESPHOME_LOG_LEVEL_{config[df.CONF_LOG_LEVEL]}")
+    )
```
jesserockz: Maybe this?
```suggestion
    cg.add_define("LVGL_LOG_LEVEL", cg.RawExpression(f"ESPHOME_LOG_LEVEL_{config[df.CONF_LOG_LEVEL]}"))
```

jesserockz (approved):

PR #7625, opened 2024-10-17
extending the bytebuffer.

nielsnl68:

Introducing the following new options:
- Added a way to figure out the total used_space, not depending on the current position anymore. 
- `flip();` will now jump to the *used_space* marker instead of the current *position*. [BREAKING]
- added a getter/setter for byte arrays.
- Added a flag that will be set when the buffer is being changed.
- changing the parameter naming for byte array wrapper.

This PR is continuing from https://github.com/esphome/esphome/pull/7564

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


nielsnl68:
> There's no quick and easy solution, the best option I can see is to rewrite all the put and get functions to take a specified offset and not update the position, then have the variants without a specified offset call those with the current position and update it.
> 
> That's a fair bit of reworking, but will give the flexibility you want, will not impact performance too much and will avoid duplicating the encode/decode stuff. E.g.
> 
> ```
> uint8_t ByteBuffer::get_uint8(size_t offset) {
>   return this->data_[offset];
> }
> 
> uint8_t ByteBuffer::get_uint8() {
>   return this->get_uint8(this->position_++);
> }
> ```
> 
> etc.

I have been looking to do it this way. But quite frankly it gives a lot of overhead. Just for something niche. The above means just that it double the code and from what i understood from earlier messages, you where against that, and i agree on this. When you do belief that the extra method call will slow down the normal use.
My proposal is that I make the `update_used_space_();` method an **inline** method. This way it will be added to each calling method. 
This way the class stay's working as it is without much overhead, and when someone like to out of sequence excess there are some way's to do so.

clydebarrow:
The proposed change not only adds run-time overhead, but makes the code very hard to read (and therefore hard to ensure it's correct.)

I have reworked the class to use templates and all in-lineable functions, which will reduce the code size impact to zero when the class is not used, and minimise both time and space when anything is used, and added the position-independent puts and gets. See https://github.com/esphome/esphome/pull/7638

jesserockz:
Replaced by #7638. If you have any further enhancements, please make them towards the new component based code =)

PR #7629, opened 2024-10-18
Tcs34725: implement CIE1931 color output

RubenKelevra:
_This PR requires that https://github.com/esphome/esphome/pull/7508, https://github.com/esphome/esphome/pull/7517 and https://github.com/esphome/esphome/pull/7574 be accepted first._

This PR implements a matrix transformation to calculate the CIE1931 X, Y, and Z values from the raw RGB data provided by the TCS34725 sensor.

The method used is based on a vendor description for a similar color chip ([source](https://look.ams-osram.com/m/57eb503c1d12d800/original/TCS34xx-AN000517.pdf)). I have adapted this method for the more modern TCS34725, which has different color channel responsivities.

For the calculations and correction data specific to the TCS34725, please refer to my [side-repo](https://github.com/RubenKelevra/tcs34725-irradiance-correction).

This repo also contains a script to calculate the color gamut of the TCS34725. The TCS34725 demonstrates an impressive performance with nearly full coverage of the DCI-P3 color space, except in the deep red range, where the built-in IR filter appears to limit its range:

color_gamut_with_references

### New Sensors:

- `cie1931_x`
- `cie1931_y`
- `cie1931_z`

## Rationale

While  irradiance shows how much light the sensor is receiving (implemented in https://github.com/esphome/esphome/pull/7517), as an energy reading, irradiance alone isn't helpful in terms of what colors humans would perceive this light at, or at what brightness. 

The reason for this in simple terms is, that we got 3 color pixels in our eyes, which have different sensitivities than the sensor.

So irradiance is making the responses of each of the R, G, B channel the same, if the same amount of light is measured. 

While CIE1931 measures (very roughly) how humans perceive colors with three different sensitivities. 

This PR converts how the sensor is seeing light to how humans are seeing light, based on spectral measurements provided by the manufacturer. In addition manufacturer provided measurements for three different light sources are used to normalize how bright the sensor is seeing light, which were also used for the irradiance calculations.

The resulting CIE1931 colors can be used to determine the Lux better (as the current approach is just meant for light sources with 4000 K) - as the Y channel is equal to Lux. 

It can also be used to determine the color temperature better, as we can "cut off" the color temperature calculations, if the light is too far away from being white, instead of calculating weird values. 

## Limitations

The current approach uses simulated light source values of real measurements taken by the manufacturer in their calibration process. This is approach is likely not very accurate at the edges of the sensor's color spectrum. 

And indeed in my testing I found that the deep red section is having some troubles.

I'm looking into improving the color output in the future by comparing light source readings between this sensor and a calibrated color sensor.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

<details>

```yaml  
esphome:
  name: $devicename
  platform: ESP8266
  board: d1_mini
  friendly_name: $friendly_name

substitutions:

logger:

mdns:
  
time:
  - platform: homeassistant

api:
  encryption: 

ota:
  platform: esphome

wifi:
  networks:
    ssid: !secret wifi_ssid
    manual_ip:
      gateway: !secret wifi_gateway
      subnet: !secret wifi_subnet
      dns1: !secret wifi_dns1
  reboot_timeout: 5min

i2c:
  sda: D1
  scl: D2
  scan: true
  id: bus_a

sensor:
  - platform: tcs34725
    i2c_id: bus_a
    cie1931_x:
      name: "$friendly_name CIE1931 X"
    cie1931_y:
      name: "$friendly_name CIE1931 Y"
    cie1931_z:
      name: "$friendly_name CIE1931 Z"
    red_channel_irradiance:
      name: "$friendly_name Red Channel Irradiance"
    green_channel_irradiance:
      name: "$friendly_name Green Channel Irradiance"
    blue_channel_irradiance:
      name: "$friendly_name Blue Channel Irradiance"
    sensor_saturation:
    illuminance:
    color_temperature:
    integration_time: auto
    glass_attenuation_factor: 1.0
    address: 0x29
    update_interval: 1s    

output:
  - platform: gpio
    pin: D3
    id: white_led
    inverted: false
  
light:
  - platform: binary
    output: white_led
```

</details>

## Checklist:


RubenKelevra:
@swifty99 if you like to test this one as well, use this external component:

```
external_components:
  - source: github://RubenKelevra/esphome_esphome@tcs34725_add_cie_1931_colors_v0.14
    refresh: 1s
    components: [ tcs34725 ]  
```

swifty99:
Nice. Will observe how the real sky 2 CIE colors match.

RubenKelevra:
@swifty99 note that the calibration is **not good** at the moment. That was what the spec sheet provided, and just simulated based on converting color channel responses to the clear channel responses, after digitizing a graph.

So at least 3 steps which tainted the accuracies, unfortunately.

If I can I provide better calibration for this sensor in the future, but as implementing this with non-optimal calibration is a non-issue, I think this is fine for now.

It can however already provide new Lux values, because CIE1931's Y channel is equal to Lux, but isn't calibrated to a specific color temperature. So if the color temperature changes, it should provide a more consistent brightness response.

RubenKelevra:
Hi everyone! Here's my plan to improve the color correction profile for this sensor.

I’ve ordered seven additional sensors, so now I’ll be working with a total of nine. By averaging across multiple sensors, I hope to avoid the need for a single calibration profile, instead balancing the slight variations between sensors for more consistent results.

To cover a broad range of colors, I’ll be testing with different light sources I have on hand, including a photo-grade 200 W 6500 K CFM, high CRI LEDs (95+ RA), halogen lamps, standard 85+ RA lamps, and some typical light bulbs. This should give us a good spectrum across various lighting scenarios.

I’ll also use high-quality RGBCT LEDs, plus a display that’s rated for 99% DCI-P3 color coverage, which will help capture those extreme colors accurately.

For calibration, I’ll use a reference sensor that outputs calibrated CIE1931 data, as well as lux and white temperature. I’ll compare this data with the raw values from the TCS34725 sensors, then create a correction profile based on the average values across the nine sensors.

I’m also looking into re-enabling the clear channel for a “high-accuracy” mode. The clear channel has a similar response to the color channels but without the cross-talk interference, so including it might help correct cross-talk effects and improve accuracy in specific lighting setups (just a theory for now).

Finally, I’m working on integrating this calibrated color sensor with ESPHome—it's proving to be a bit more challenging than expected, so it might take a little time depending on how much I can put into it.

Let me know if you have any thoughts or suggestions!


In the end I like to point out, that this is no blocker: The current calibration is not optimal, but changing some numbers around once I get better once is a simple fix. :)

RubenKelevra:
@swifty99 have you found any bugs?

swifty99:
> @swifty99 have you found any bugs?

All good. Working stable since and still running this version.

RubenKelevra:
@swifty99 great! I organized funding for a spectrum analyzer which is kinda necessary to proceed with the calibration and hope it'll arrive here quickly, so I can proceed :)

PR #7631, opened 2024-10-19
[image][online_image][animation] Fix transparency in RGB565

clydebarrow:

When images were converted to RGB565 with transparency, especially from SVG, anti-aliasing was lost as the image format use chroma-keying instead of an alpha channel.

This PR changes the RGB565 with transparency format to one that has a proper alpha channel, so anti-aliasing works as expected.

## Types of changes

- [x] Other

* https://github.com/esphome/issues/issues/6047
* https://discord.com/channels/429907082951524364/1297013896099532810

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/image/image.h
```diff
@@ -3,12 +3,7 @@
 #include "esphome/components/display/display.h"
 
 #ifdef USE_LVGL
-// required for clang-tidy
-#ifndef LV_CONF_H
-#define LV_CONF_SKIP 1  // NOLINT
-#endif                  // LV_CONF_H
-
-#include <lvgl.h>
+#include "esphome/components/lvgl/lvgl_proxy.h"
```
guillempages: Shouldn't this part be a separate PR?
one PR to fix the transparency, and one PR to fix the host compilation?

esphome/components/image/image.h
```diff
@@ -49,6 +26,25 @@ class Image : public display::BaseImage {
   const uint8_t *get_data_start() const { return this->data_start_; }
   ImageType get_type() const;
 
+  int get_bpp() const {
```
guillempages, line 29: Having this as a static method (as it was), is slightly better performant...
It's not a big deal, but I've done some tests with a (very big -> 550x800) image, and I'm getting 4 ms difference in rendering.

With this PR it's averaging 1.486s per frame; without it was taking 1.482s.
As I said; not much, but since you are always trying to squeeze the most performance on graphics, it might be worth a second look...

esphome/components/online_image/online_image.h
```diff
@@ -86,13 +86,9 @@ class OnlineImage : public PollingComponent,
   Allocator allocator_{Allocator::Flags::ALLOW_FAILURE};
 
   uint32_t get_buffer_size_() const { return get_buffer_size_(this->buffer_width_, this->buffer_height_); }
-  int get_buffer_size_(int width, int height) const {
-    return std::ceil(image::image_type_to_bpp(this->type_) * width * height / 8.0);
-  }
+  int get_buffer_size_(int width, int height) const { return (this->get_bpp() * width + 7u) / 8u * height; }
```
guillempages, line 89:
```suggestion
  int get_buffer_size_(int width, int height) const { return (this->get_width_stride() * height; }
```

guillempages: Oh, no, the suggestion doesn't work, since it is not using the member width_ :-(

guillempages (commented):
Hm. I don't know why, but when testing this PR with an animation, the animation is now showing wrong.
Looks like there is something wrong with reading the pixels. (My animation is using RGB565 with transparency; maybe the drawing part of the animation also needs updating for the 3bytes per pixel somewhere?

Took a video: https://github.com/user-attachments/assets/232af92a-6b51-43ca-8dc2-eed7aaec9a21

Without the pr:
https://github.com/user-attachments/assets/199408d7-12ef-4ef3-9221-af625bc92340


(Ignore the cross/"HI" on the top left, that's a different issue (I hope))

clydebarrow:
> maybe the drawing part of the animation also needs updating for the 3bytes per pixel somewhere?

More than likely. I did make some changes to animation, but didn't test it :thinking:  I did test regular images and online images. Will do some testing and update.

esphome/components/image/image.h
clydebarrow: Perhaps, but I needed that for testing anyway.

esphome/components/image/image.h
clydebarrow: I'll have a look.

clydebarrow: I don't think `get_bpp()` has anything to do with it, it's not called in a loop anywhere that I can see. And that's a 0.25% difference, that's in the range of measurement error. You know that using LVGL to draw images is at least twice as fast as the display lambda?

clydebarrow:
> but when testing this PR with an animation, the animation is now showing wrong.

Fixed now - the python code to convert the animated GIF needed updating too.

jesserockz (approved):

PR #7635, opened 2024-10-19
[mlx90393] Fix inverted gain and resolution. Expose temperature_compensation and hallconf.

functionpointer:

I experienced integer overflows with MLX90393 three axis magnetometer. This fixes it.

I recommend reading the new documentation: [deploy preview](https://deploy-preview-4353--esphome.netlify.app/components/sensor/mlx90393)

The old gain options were copied from the arduino-MLX90393 library. However, the library actually does the opposite thing from the sensor hardware. Low hardware gain settings are multiplied with large numbers, while high gains are multiplied with small numbers. This is done in order to get consistent readings in µT independent of gain setting.

This is confusing from the user's perspective. The user configures the sensor, not the library.
Therefore, this PR inverts (actually 5/x) all gain settings. This reflects the analog gain used inside the sensor.

Furthermore, the resolution options are named more appropriately. The resolution is always 16 bit, we just choose which 16 bits we get.

Additionally, the default values are changed. The old defaults offered high sensitivity, but caused integer overflows in strong magnetic fields. This easily happened when a neodymium magnet was held within a few cm of the sensor.
The new default configuration prevents all integer overflows and can deal with strong fields easily.
Higher sensitivity must now be explicitly configured by the user. The new documentation informs about integer overflows, mitigating the risk of false readings that existed previously.

Finally, this PR exposes the already existing `hallconf` and `temperature_compensation` options.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4353

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [X] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: mlx90393
    gain: 1X
    x_axis:
      name: mlxxaxis
    y_axis:
      name: mlxyaxis
    z_axis:
      name: mlxzaxis
```

## Checklist:


esphome/components/mlx90393/sensor.py
```diff
@@ -58,27 +75,27 @@ def mlx90393_axis_schema(default_resolution: str):
...
+            cv.Optional(CONF_OVERSAMPLING, default=0): cv.int_range(min=0, max=3),
             cv.Optional(CONF_FILTER, default=6): cv.int_range(min=0, max=7),
-            cv.Optional(CONF_X_AXIS): mlx90393_axis_schema("19BIT"),
-            cv.Optional(CONF_Y_AXIS): mlx90393_axis_schema("19BIT"),
-            cv.Optional(CONF_Z_AXIS): mlx90393_axis_schema("16BIT"),
+            cv.Optional(CONF_X_AXIS): mlx90393_axis_schema(),
+            cv.Optional(CONF_Y_AXIS): mlx90393_axis_schema(),
+            cv.Optional(CONF_Z_AXIS): mlx90393_axis_schema(),
+            cv.Optional(CONF_TEMPERATURE_COMPENSATION, default=False): bool,
+            cv.Optional(CONF_HALLCONF, default=0xC): cv.one_of(0xC, 0x0),
```
jesserockz, line 98: You have not used these new variables in `to_code` below

jesserockz (changes requested):

esphome/components/mlx90393/sensor.py
functionpointer: Good catch, I have added it

functionpointer:
I have addressed the changes. The PR is ready for another review.

jesserockz (approved):

PR #7650, opened 2024-10-22
updating ESP32 board definitions

asolochek:

This adds the pin mappings for missing adafruit boards and a couple of missing pin definitions for wesp32. 

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esp32:
  board: adafruit_qtpy_esp32s3_nopsram

i2c:
  - id: bus_a
    sda: SDA1
    scl: SCL1
```

## Checklist:


esphome/components/esp32/boards.py
```diff
@@ -103,6 +103,217 @@
...
+        "MISO": 37,
+        "RX": 38,
+        "TX": 39,
+        "D13": 13,
+        "D12": 12,
+        "D11": 11,
+        "D10": 10,
+        "D9": 9,
+        "D6": 6,
+        "D5": 5,
```
jesserockz: Looking at the images of the board, the silkscreen does not label these as Dx so this is not needed.
```suggestion
```

jesserockz (changes requested):

esphome/components/esp32/boards.py
asolochek: I think you looked at the wrong board. This board does have those pins exposed and labelled. https://learn.adafruit.com/assets/110811

jesserockz: The pin definition we put in this file are supposed to be silkscreen to gpio pin number. The silkscreen on the physical board shows only 13, 12, etc. No `D`.

jesserockz: The Dx thing is something that is brought from the Arduino world, which we want to avoid as much as possible.

asolochek: Ok. I was trying to be consistent with the other boards, but at least for these adafruit boards the DXX pins correspond directly to GPIOXX, so I suppose it doesn't matter much. Although I'd think users might expect to be able to use DXX because of the adafruit documentation for the boards. 

Out of curiosity though, if the board had only numbers on the silk screen, and they didn't correspond to the same GPIO numbers, what convention would be used? DIOXX? IOXX?

jesserockz (approved):

PR #7529, opened 2024-10-02
Add OpenTherm component (part 2.1: sensor platform)

olegtarasov:
# What does this implement/fix?

This is an interim PR for OpenTherm component. It contains the code to support just the `sensor` platform. Original PR is here: #7463 

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** 

Part 1: #6645 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** 

esphome/esphome-docs#3800

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
esphome:
  name: thermostat-number-minimal

esp8266:
  board: d1_mini

opentherm:
  in_pin: 1
  out_pin: 2

sensor:
  - platform: opentherm
    rel_mod_level:
      name: "Boiler Relative modulation level"
    ch_pressure:
      name: "Boiler Water pressure in CH circuit"
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #7676, opened 2024-10-25
Add OpenTherm component (part 3: rest of the sensors)

olegtarasov:

This is the third (and final) part of OpenTherm component. It contains the code to support all the sensors, inputs and outputs. Sensors configured in YAML are transformed to a set of C++ macros, which in turn generate the final code that handles sensors.

## Types of changes

- [ ] Other

- Part 1: #6645 
- Part 2: #7529 

- esphome/esphome-docs#3800

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml

esphome:

esp8266:
  board: d1_mini

opentherm:
  in_pin: IN_PIN
  out_pin: OUT_PIN

number:
  - platform: opentherm
    t_set:
```

## Checklist:


olegtarasov:
Hi @jesserockz! The final part of OpenTherm integration is ready for review. Thanks!

jesserockz (commented):
How does the hub know if a switch (or other control type entity) has been changed by the user? I do not see the `add_on_state_callback`s being added.

I find it very difficult to follow and understand this component due to the heavy use of macros.

olegtarasov:
@jesserockz The hub doesn't need to know the exact moment when control entity changes because all sensors, switches and outputs are sent to OpenTherm device constantly in a loop. It's required by OpenTherm standard to communicate with device every 1 second even when there is no change in controls. So the approach when we send all the messages configured in yaml in a loop adds to component's robustness (a message can fail due to protocol error, and we don't need to use complicated retry logic — the same message will be sent again eventually).

> I find it very difficult to follow and understand this component due to the heavy use of macros.

I agree, but I see no other way to add class members dynamically, and adding > 100 class members for all the available sensors at once seems very wasteful. What's more, most of them likely won't be supported by end user's boiler.

tests/components/opentherm/common.yaml
```diff
@@ -74,3 +105,86 @@ sensor:
...
+      kp: 0.7
+      ki: 0.003
+  - platform: pid
+    name: "Central heating (Circuit 2)"
+    heat_output: t_set_ch2
+    default_target_temperature: 18
+    sensor: ch2_room_temperature
+    control_parameters:
+      kp: 0.4
+      ki: 0.004
```
jesserockz: Adding this here wont achieve anything and if the `pid` component has a breaking change, these files wont be run. They should be contained to only itself and dependencies
```suggestion
```

tests/components/opentherm/common.yaml
```diff
@@ -74,3 +105,86 @@ sensor:
...
+    entity_id: sensor.temperature
+    filters:
+      # Push room temperature every second to update PID parameters
+      - heartbeat: 1s
+  - platform: homeassistant
+    id: ch2_room_temperature
+    entity_id: sensor.other_temperature
+    filters:
+      # Push room temperature every second to update PID parameters
+      - heartbeat: 1s
```
jesserockz: Same with these
```suggestion
```

tests/components/opentherm/common.yaml
```diff
@@ -12,10 +12,41 @@ opentherm:
   cooling_enable: false
   otc_active: false
   ch2_active: true
+  t_room: ch_room_temperature
```
jesserockz: Just use one of the opentherm sensors here as this is a simple compile test and doesnt actually run

esphome/components/opentherm/switch/switch.cpp
```diff
@@ -0,0 +1,27 @@
...
+  } else {
+    ESP_LOGD("opentherm.switch", "Restored state for OpenTherm switch '%s': %d", this->get_name().c_str(),
+             restored.value());
+    state = restored.value();
+  }
+  this->write_state(state);
+}
+
+void OpenthermSwitch::dump_config() {
+  esphome::switch_::log_switch("opentherm.switch", "", "OpenTherm Switch", this);
```
jesserockz:
```suggestion
  LOG_SWITCH(TAG, "", "OpenTherm Switch", this);
```

esphome/components/opentherm/switch/switch.cpp
```diff
@@ -0,0 +1,27 @@
+#include "switch.h"
+
+namespace esphome {
+namespace opentherm {
+
```
jesserockz, line 5:
```suggestion

static const char *const TAG = "opentherm.switch";

```

esphome/components/opentherm/switch/switch.cpp
```diff
@@ -0,0 +1,27 @@
...
+void OpenthermSwitch::write_state(bool state) { this->publish_state(state); }
+
+void OpenthermSwitch::setup() {
+  auto restored = this->get_initial_state_with_restore_mode();
+  bool state = false;
+  if (!restored.has_value()) {
+    ESP_LOGD("opentherm.switch", "Couldn't restore state for OpenTherm switch '%s'", this->get_name().c_str());
```
jesserockz: Please update to add TAG to all the cpp files and use that for logging
```suggestion
    ESP_LOGD(TAG, "Couldn't restore state for OpenTherm switch '%s'", this->get_name().c_str());
```

esphome/components/opentherm/switch/switch.cpp
```diff
@@ -0,0 +1,27 @@
...
+  } else {
+    ESP_LOGD("opentherm.switch", "Restored state for OpenTherm switch '%s': %d", this->get_name().c_str(),
```
jesserockz:
```suggestion
    ESP_LOGD(TAG, "Restored state for OpenTherm switch '%s': %d", this->get_name().c_str(),
```

jesserockz (changes requested):

jesserockz (approved):

PR #7677, opened 2024-10-25
[speaker] Removing the Finish() method. It is not needed anymore.

nielsnl68:

After deliberation with @kahrendt we came to the conclusion that Finish() procedure is not needed anymore. You can now just wait until everything is played and the speaker stops automatically.

This PR fixes also the issue that not everything was played at the end. 

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
+++ b/esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
@@ -26,7 +26,6 @@ static const char *const TAG = "i2s_audio.speaker";
 enum SpeakerEventGroupBits : uint32_t {
   COMMAND_START = (1 << 0),                           // Starts the main task purpose
   COMMAND_STOP = (1 << 1),                            // stops the main task
-  COMMAND_STOP_GRACEFULLY = (1 << 2),                 // Stops the task once all data has been written
   MESSAGE_RING_BUFFER_AVAILABLE_TO_WRITE = (1 << 5),  // Locks the ring buffer when not set
   STATE_STARTING = (1 << 10),
   STATE_RUNNING = (1 << 11),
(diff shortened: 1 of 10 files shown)


esphome/components/speaker/__init__.py
```diff
@@ -81,9 +81,6 @@ async def speaker_play_action(config, action_id, template_arg, args):
 automation.register_action("speaker.stop", StopAction, SPEAKER_AUTOMATION_SCHEMA)(
     speaker_action
 )
-automation.register_action("speaker.finish", FinishAction, SPEAKER_AUTOMATION_SCHEMA)(
-    speaker_action
-)
```
jesserockz, line 86: Lets leave this in so its not a config breaking change and just change the class it uses.
```suggestion
automation.register_action("speaker.finish", StopAction, SPEAKER_AUTOMATION_SCHEMA)(
    speaker_action
)
```

jesserockz (changes requested):

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -261,26 +254,18 @@ void I2SAudioSpeaker::speaker_task(void *params) {
...
         }
-
-      } else {
-        // No data received
-
-        if (stop_gracefully) {
-          break;
-        }
-
-        i2s_zero_dma_buffer(this_speaker->parent_->get_port());
```
kahrendt, line 275: Can you add the ``i2s_zero_dma_buffer`` back in when no data is read? Some amps will generate harsh noises if you suddenly stop writing anything via I2S, so the zeroing out prevents that.

esphome/components/speaker/__init__.py
nielsnl68: @jesserockz, the Stop and finish are 2 different thinks. The first stops the speaker directly while finish should wait until the everything is send out and then close the speaker.
Changing the finish action by doing the full stop as you suggested is as a breaking change as removing. it is worse, imho, because the end-user will think it will wait and found that the end of the audio stream is not being played anymore after calling finish. By removing the action totally will make the end-user investigate what is the problem and will found out that he does not need to call finish anymore instead he could wait until the speaker is stopped playing. 

One solution could be that we make `finish` a wait action until the speaker stops.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
nielsnl68: Copy of discord message:

Sorry i was busy the last couple of days. 
I have a issue with this, because when added, not all the playable content is being output to the speaker. 
When we figure out how to resolve that then it can be back,
An other solution would be (which we talked before) is setting the timeout based on the data that is received.

nielsnl68:
This PR is not needed anymore PR https://github.com/esphome/esphome/pull/7749 fixes the need for it.

PR #7678, opened 2024-10-25
[POC] Adding audio locking to play streams to the speaker.

nielsnl68:

This PR is an attempt to lock speaker or mic to an audio components until it is released by this component. Instead of calling play directly you now need to call Start() first which return a streamer object that can be used to play the audio or get the audio back from the mic (not implemented yet).

Sadly this introduced a lot of changes. But i think it is worth it. Explicitly when we get more audio components that want to use the speaker or mic.

For the end user there will noting changed. But for audio component developers this will changes the way to talk to the speaker or listen to the microphone.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```cpp
# Example audio playback example in c++
void play(std::vector<uint8_t> vector_data) {
  while (true) {
    auto streamer this->speaker_ ->start();
    if (streamer != nullptr) break;
  }
  streamer.stream(vector_data);
  delete streamer;
}
```
or
```cpp
void play(std::vector<uint8_t> vector_data, const audio::AudioStreamInfo &audio_stream_info) {
  while (true) {
    auto streamer this->speaker_ ->start(audio_stream_info);
    if (streamer != nullptr) break;
  }
  streamer.stream(vector_data);
  delete streamer;
}
```

## Checklist:


nielsnl68:
this is not needed anymore

PR #7682, opened 2024-10-26
[esp32] Add custom partitions and refactor partition table generation

luar123:

Adds the ability to define custom partitions in the ESP32 partition table from YAML or programmatically via `add_partition()`. Also refactors the partition table generation to unify Arduino and IDF code paths.

**Custom partitions feature** (original work by @luar123):
- Components or YAML configs can add custom data/app partitions
- Custom partitions are appended at the end of flash and steal space from the OTA app partitions
- Validation ensures reserved names, valid types/subtypes, and 4KB alignment

**Partition table refactoring:**
- Replaced `APP_PARTITION_SIZES` lookup table with a formula: `app_size = (flash - 0x80000 - custom_total) / 2`
- Unified `get_arduino_partition_csv()` and `get_idf_partition_csv()` into a single `get_partition_csv()`
- Both frameworks now use auto-placed offsets (blank offset field) — the same `gen_esp32part.py` tool handles placement for both
- Added `phy_init` partition to Arduino (was IDF-only; both frameworks use ESP-IDF under the hood)
- Moved NVS to end of flash and increased its size to fill previously wasted space:
  - Arduino: 20KB → 384KB (was 384KB unused at end of flash)
  - IDF: 436KB → 448KB (was 12KB unused at end of flash)
- Extracted shared validation into `_validate_partition()` used by both YAML schema validation and the programmatic `add_partition()` API

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) — [policy](https://developers.esphome.io/contributing/code/#what-constitutes-a-c-breaking-change)
- [ ] Developer breaking change (an API change that could break external components) — [policy](https://developers.esphome.io/contributing/code/#what-is-considered-public-c-api)
- [ ] Undocumented C++ API change (removal or change of undocumented public methods that lambda users may depend on) — [policy](https://developers.esphome.io/contributing/code/#c-user-expectations)
- [ ] Other

**Breaking change details:**
- Partition layout changes for both frameworks (NVS moved to end, sizes changed, phy_init added to Arduino). Existing devices will get a new partition table on next OTA update. NVS data will be preserved as long as the NVS partition name remains `nvs` (the NVS library finds it by name, not position).

- fixes #11553

- esphome/esphome-docs#5538

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040/RP2350
- [ ] BK72xx
- [ ] RTL87xx
- [ ] LN882x
- [ ] nRF52840

```yaml
# Example config.yaml
esp32:
  partitions:
    - name: my_data
      type: data
      subtype: spiffs
      size: 0x4000
```

## Checklist:


luar123:
In case of a custom partition table, I have this simple final validation in my zigbee component to check if the needed partition is there:
```
def final_validate(config):
    esp_conf = fv.full_config.get()["esp32"]
    if CONF_PARTITIONS in esp_conf:
        with open(
            CORE.relative_config_path(esp_conf[CONF_PARTITIONS]), encoding="utf8"
        ) as f:
            partitions = f.read()
            if ("zb_storage" not in partitions):
                raise cv.Invalid(
                    "Add \n'zb_storage, data, fat,   , 16K,'\n to your custom partition table."
                )
    return config
```
Should I move this to the esp32 component, to check if all partitions added with `add_custom_partition()` are included in the custom partition table?

DFINITYManu (approved):

Hedda:
Any updates on this?

esphome/components/esp32/__init__.py
```diff
@@ -607,6 +614,21 @@ def final_validate(config):
                     path=[CONF_FRAMEWORK, CONF_ADVANCED, CONF_EXECUTE_FROM_PSRAM],
                 )
             )
+    if CONF_PARTITIONS in config and custom_partitions:
+        print("check part")
+        with open(
+            CORE.relative_config_path(config[CONF_PARTITIONS]), encoding="utf8"
+        ) as f:
+            partitions_tab = f.read()
+            print(partitions_tab)
```
tomaszduda23: debug print

luar123: Thanks

esphome/components/esp32/__init__.py
```diff
@@ -552,6 +556,9 @@ def _detect_variant(value):
     return value
 
 
+custom_partitions = {}
```
tomaszduda23: Shouldn't it be part of CORE.data?

esphome/components/esp32/__init__.py
```diff
@@ -1231,6 +1256,25 @@ def _set_default_framework(config):
                 *FLASH_SIZES, upper=True
             ),
             cv.Optional(CONF_PARTITIONS): cv.file_,
```
swoboda1337: Why not just reuse cv.Optional(CONF_PARTITIONS) allow a file or a definition?

swoboda1337 (changes requested):

swoboda1337:
Here is Claude's review, somethings may be wrong or not relevant but most look legitimate:
```
  Bugs

  1. No app partition size overflow check

  app_partition_size = APP_PARTITION_SIZES[flash_size] - _get_custom_partition_half_size()

  If custom partitions are too large, this goes negative (wraps to a huge value), producing a nonsensical partition table. Needs a
  check like:

  if app_partition_size < MINIMUM_APP_SIZE:
      raise cv.Invalid("Custom partitions too large for flash size")

  2. _get_custom_partition_half_size() halving logic is questionable

  The function sums all custom partition sizes, then divides by 2 and aligns to 64KB — assuming the space is stolen equally from
  app0 and app1. This works for the Arduino layout (where both app partitions shrink symmetrically), but is conceptually wrong:
  you're adding N bytes of custom partitions, so you need to steal N bytes total, not N/2 from each. The math happens to work out
  (N/2 from each = N total), but only because there are exactly 2 app partitions. It's confusing.

  3. Dead final_validate code path

  if CONF_PARTITIONS in config and KEY_CUSTOM_PARTITIONS in CORE.data:

  The has_at_most_one_key(CONF_PARTITIONS, CONF_CUSTOM_PARTITIONS) check at schema level means CONF_PARTITIONS (file) and
  CONF_CUSTOM_PARTITIONS (inline) can never coexist. So KEY_CUSTOM_PARTITIONS in CORE.data is only populated when
  CONF_CUSTOM_PARTITIONS is used, which means CONF_PARTITIONS is absent. This validation block is unreachable dead code.

  4. Duplicate partition names silently overwrite

  CORE.data[KEY_CUSTOM_PARTITIONS] is a dict keyed by name. Two partitions with the same name would silently merge, with the second
  overwriting the first. Should error on duplicates.

  Validation Gaps

  5. String subtype not validated

  When subtype is a string, any value is accepted via cv.string_strict. Invalid subtypes (e.g., subtype: "banana") would pass YAML
  validation but fail at flash time with a confusing ESP-IDF error. Should validate against known subtypes: ota, phy, nvs, nvs_keys,
   fat, spiffs, coredump, etc. (depending on the type).

  6. Reserved name list is incomplete

  if name in ["nvs", "app0", "app1", "otadata", "eeprom", "spiffs", "phy_init"]:

  Missing at least: coredump, nvs_keys, efuse, factory. The IDF partition table has more standard names that could conflict.

  7. No validation that type: app partitions have valid subtypes

  App partitions must have subtypes like ota_0, ota_1, factory, test. A type: app, subtype: spiffs would be invalid but would pass
  validation.

  Code Quality

  8. Floating-point alignment math

  custom_partition_start = ceil(custom_partition_start / 0x10000) * 0x10000

  Uses float division then ceil(). For large addresses this could lose precision. Use integer math instead:

  custom_partition_start = (custom_partition_start + 0xFFFF) & ~0xFFFF  # 64KB align
  custom_partition_start = (custom_partition_start + 0xFFF) & ~0xFFF    # 4KB align

  9. Side effects during validation

  validate_custom_partition() writes to CORE.data as a schema validator (inside cv.All). Validation should ideally be pure; side
  effects belong in to_code() or a final validation step. If validation is re-run (e.g., during config reload), the data
  accumulates.

  10. IDF partition CSV doesn't handle alignment for custom partitions

  The Arduino CSV generator correctly aligns custom partition offsets (64KB for app, 4KB for data). The IDF generator omits offsets
  entirely (relying on IDF auto-placement), which works for standard partitions but custom partitions with specific alignment
  requirements may not get placed correctly.

  Testing

  11. No Arduino test case

  Only test.esp32-s3-idf.yaml is updated. The Arduino partition CSV generation is more complex (explicit offsets), so it needs its
  own test to verify the offset math is correct.

  Summary

  The feature is needed (zigbee, custom data storage), but the implementation should:
  1. Adopt swoboda1337's suggestion: extend CONF_PARTITIONS instead of adding a new key
  2. Add overflow validation on app partition size
  3. Fix or remove the dead final_validate code
  4. Validate subtypes against known values per type
  5. Use integer alignment instead of float math
  6. Add an Arduino test case
  ```

**Comment by luar123** [comment]
Thanks for the feedback.
Have merged with CONF_PARTITIONS and refactored the validator. Includes fixes to Claudes comments 1, 3, 4, 5, 7, 8, 9. Rest seems irrelevant and there are no Arduino tests for esp32 at all, should I add one?

**Comment by luar123** [comment]
@swoboda1337 is there more I should look into?

**Comment by swoboda1337** [comment]
Sorry it was a busy week with the release. Will get to this and other PRs next week.

File: esphome/components/esp32/__init__.py [hunk 4]
```diff
@@ -1211,6 +1212,54 @@ def _set_default_framework(config):
...
+        and config["subtype"]
+        not in [
+            "nvs",
+            "nvs_keys",
+            "spiffs",
+            "coredump",
+            "efuse",
+            "fat",
+            "undefined",
+            "littlefs",
```
swoboda1337: Missing "ota" and "phy" from the valid data subtypes

esphome/components/esp32/__init__.py
```diff
@@ -1232,7 +1281,28 @@ def _set_default_framework(config):
...
+                                cv.Required(CONF_NAME): cv.string_strict,
+                                cv.Required(CONF_TYPE): cv.Any(
+                                    cv.string_strict,
+                                    cv.int_range(0x40, 0xFE),
+                                ),
+                                cv.Required(CONF_SUBTYPE): cv.Any(
+                                    cv.string_strict,
+                                    cv.int_range(0, 0xFE),
+                                ),
+                                cv.Required(CONF_SIZE): cv.int_,
```
swoboda1337: cv.int_ allows 0 and negative values, should validate size > 0

tests/components/esp32/test.esp32-s3-idf.yaml
```diff
@@ -1,5 +1,10 @@
 esp32:
   variant: esp32s3
+  partitions:
```
swoboda1337: Should also add an Arduino test for custom partitions

esphome/components/esp32/__init__.py
```diff
@@ -1790,6 +1869,35 @@ async def to_code(config):
         CORE.add_job(_write_arduino_libraries_sdkconfig)
 
 
+KEY_CUSTOM_PARTITIONS = "custom_partitions"
+
+
+def add_partition(name: str, p_type: str, subtype: str, size: int) -> None:
```
swoboda1337: Type hints say str but isinstance checks below also accept int. For the YAML path the isinstance checks are dead code since _validate_custom_partition already converts ints

esphome/components/esp32/__init__.py
```diff
@@ -1790,6 +1869,35 @@ async def to_code(config):
...
+        subtype_ = f"0x{subtype_:X}"
+    custom_partitions = CORE.data.setdefault(KEY_CUSTOM_PARTITIONS, {})
+    custom_partitions[name] = {"type": p_type_, "subtype": subtype_, "size": size}
+
+
+def _get_custom_partition_half_size() -> int:
+    size = 0
+    for partition in CORE.data.get(KEY_CUSTOM_PARTITIONS, {}).values():
+        if partition["type"] == "app":
+            size = (size + 0xFFFF) & ~0xFFFF  # align to 64KB
```
swoboda1337: Alignment logic is duplicated between here and get_arduino_partition_csv, could get out of sync

esphome/components/esp32/__init__.py
```diff
@@ -1790,6 +1869,35 @@ async def to_code(config):
...
+KEY_CUSTOM_PARTITIONS = "custom_partitions"
+
+
+def add_partition(name: str, p_type: str, subtype: str, size: int) -> None:
+    p_type_ = p_type
+    subtype_ = subtype
+    if name in CORE.data.get(KEY_CUSTOM_PARTITIONS, {}):
+        raise ValueError(f"Partition name {name} is already defined.")
+    if name in ["nvs", "app0", "app1", "otadata", "eeprom", "spiffs", "phy_init"]:
+        raise ValueError(f"Partition name {name} is reserved.")
```
swoboda1337: ValueError will give the user a traceback instead of a clean error message

esphome/components/esp32/__init__.py
```diff
@@ -1790,6 +1869,35 @@ async def to_code(config):
...
+
+def _get_custom_partition_half_size() -> int:
+    size = 0
+    for partition in CORE.data.get(KEY_CUSTOM_PARTITIONS, {}).values():
+        if partition["type"] == "app":
+            size = (size + 0xFFFF) & ~0xFFFF  # align to 64KB
+        else:
+            size = (size + 0xFFF) & ~0xFFF  # align to 4KB
+        size += partition["size"]
+    return (int(size / 2) + 0xFFFF) & ~0xFFFF  # align to 64KB
```
swoboda1337: nit: should be size // 2

tests/components/esp32/test.esp32-s3-idf.yaml
swoboda1337: Should also add an Arduino test

esphome/components/esp32/__init__.py
swoboda1337: `cv.int_` accepts 0 and negative values. Should validate minimum size (4KB) and alignment since ESP-IDF requires partition sizes to be 4KB aligned. Something like `cv.All(cv.int_range(min=0x1000), lambda v: v if v % 0x1000 == 0 else (_ for _ in ()).throw(cv.Invalid("size must be 4KB aligned")))`

Or add a simple helper to check alignment

esphome/components/esp32/__init__.py
swoboda1337:
```suggestion
    return (size // 2 + 0xFFFF) & ~0xFFFF  # align to 64KB
```
`int(size / 2)` uses float division. `size // 2` is cleaner

esphome/components/esp32/__init__.py
```diff
@@ -1790,6 +1869,35 @@ async def to_code(config):
...
+        raise ValueError(f"Partition name {name} is already defined.")
+    if name in ["nvs", "app0", "app1", "otadata", "eeprom", "spiffs", "phy_init"]:
+        raise ValueError(f"Partition name {name} is reserved.")
+    if isinstance(p_type_, int):
+        p_type_ = f"0x{p_type_:X}"
+    if isinstance(subtype_, int):
+        subtype_ = f"0x{subtype_:X}"
+    custom_partitions = CORE.data.setdefault(KEY_CUSTOM_PARTITIONS, {})
+    custom_partitions[name] = {"type": p_type_, "subtype": subtype_, "size": size}
+
```
swoboda1337: `add_partition()` is the public API for other components (e.g. zigbee calling it from their `to_code`) but it doesnt validate type/subtype at all. A component could pass invalid strings and silently produce a broken CSV. Should share the validation logic with `_validate_custom_partition` or at least validate the basics here

esphome/components/esp32/__init__.py
```diff
@@ -881,7 +883,6 @@ def final_validate(config):
                     path=[CONF_FRAMEWORK, CONF_ADVANCED, CONF_EXECUTE_FROM_PSRAM],
                 )
             )
-
```
swoboda1337, line 910: nit: unrelated whitespace change, should revert

swoboda1337 (dismissed):

esphome/components/esp32/__init__.py
luar123: Non app partitions could have arbitrary size, right? But as we need to align to 4KB anyway this would be a waste of space, so I think that makes sense.

luar123:
Good points, thanks.

luar123:
@swoboda1337 Any chance we could proceed with this?

swoboda1337:
> @swoboda1337 Any chance we could proceed with this?

Sorry for the delay, this one was not trivial. The pre-existing partition stuff was a mess which made adding custom partitions messy. 

I pushed some changes, please take a look. I unified the arduino and idf partition code and fixed a few preexisting issues.

swoboda1337:
Ok think it is good now. Please take a look.

esphome/components/esp32/__init__.py
```diff
@@ -1831,45 +1905,118 @@ async def to_code(config):
...
+                    f"Subtype '{subtype}' is invalid for data type."
+                    f" Only {', '.join(sorted(VALID_DATA_SUBTYPES))} are allowed."
+                    " Use numbers for custom subtypes."
+                )
+    if size % 0x1000 != 0:
+        raise ValueError("Partition size must be 4KB (0x1000) aligned.")
+
+
+def add_partition(name: str, p_type: str | int, subtype: str | int, size: int) -> None:
+    if name in CORE.data.get(KEY_CUSTOM_PARTITIONS, {}):
```
bdraco: we have been moving keys to the component namespace ie `CORE.data[KEY_ESP32][KEY_CUSTOM_PARTITIONS]` to better organize things

esphome/components/esp32/__init__.py
```diff
@@ -1831,45 +1905,118 @@ async def to_code(config):
...
 
+# Partition layout (offsets auto-placed by gen_esp32part.py):
+#   otadata(0x2000) + phy_init(0x1000) + [pad to 64KB] + app0 + app1
+#   + [eeprom(0x1000) + spiffs(0xF000)]  (Arduino only)
+#   + nvs (fixed: 0x60000 Arduino, 0x70000 IDF)
+#   + custom_partitions...
+#
+# PARTITION_OVERHEAD is the total non-app budget (system partitions + nvs + padding).
+# Custom partitions are appended at the end and steal from app.
+PARTITION_OVERHEAD = 0x80000  # 512 KB
```
bdraco: This seems a bit fragile. I wonder how we could make it safer

esphome/components/esp32/__init__.py
```diff
@@ -1831,45 +1905,118 @@ async def to_code(config):
...
+        raise ValueError(f"Partition name '{name}' is already defined.")
+    _validate_partition(name, p_type, subtype, size)
+    p_type_str = f"0x{p_type:X}" if isinstance(p_type, int) else p_type
+    subtype_str = f"0x{subtype:X}" if isinstance(subtype, int) else subtype
+    custom_partitions = CORE.data.setdefault(KEY_CUSTOM_PARTITIONS, {})
+    custom_partitions[name] = {"type": p_type_str, "subtype": subtype_str, "size": size}
+
+
+def _flash_size_to_bytes(flash_size: str) -> int:
+    return int(flash_size.rstrip("MB")) * 1024 * 1024
```
bdraco: Need to make it clear this is always flash size in MB

esphome/components/esp32/__init__.py
```diff
@@ -1831,45 +1905,118 @@ async def to_code(config):
...
+    size = 0
+    for partition in CORE.data.get(KEY_CUSTOM_PARTITIONS, {}).values():
+        if partition["type"] == "app":
+            size = (size + 0xFFFF) & ~0xFFFF  # align to 64KB
+        size += partition["size"]
+    return size
+
+
+def _get_app_partition_size(flash_size: str) -> int:
+    flash_bytes = _flash_size_to_bytes(flash_size)
```
bdraco: Need to make it clear this is always flash size in MB

bdraco:
looks good

```
[10:09:43.967][C][debug:114]: Partition table:
[10:09:43.968][C][debug:114]:   Name         Type Subtype  Address    Size      
[10:09:43.970][C][debug:121]:   otadata      1    0        0x00009000 0x00002000
[10:09:43.972][C][debug:121]:   phy_init     1    1        0x0000B000 0x00001000
[10:09:43.973][C][debug:121]:   app0         0    16       0x00010000 0x001A0000
[10:09:43.975][C][debug:121]:   app1         0    17       0x001B0000 0x001A0000
[10:09:43.977][C][debug:121]:   nvs          1    2        0x00350000 0x00070000
[10:09:43.978][C][debug:121]:   my_data      1    130      0x003C0000 0x00010000
[10:09:43.980][C][debug:121]:   my_storage   1    129      0x003D0000 0x00020000
```

bdraco (dismissed):

swoboda1337 (dismissed):

swoboda1337:
Thanks!

esphome/components/esp32/__init__.py
```diff
@@ -1836,45 +1905,170 @@ async def to_code(config):
...
+
+
+def _get_partition_overhead() -> int:
+    """Total non-app partition budget (system partitions + nvs + padding).
+
+    Custom partitions are appended at the end and steal from app.
+    """
+    # otadata + phy_init are followed by app0 which requires 64KB alignment,
+    # so pad up to the next 64KB boundary.
+    overhead = (OTADATA_SIZE + PHY_INIT_SIZE + 0xFFFF) & ~0xFFFF
```
luar123: OTADATA starts at 0x9000 offset
```suggestion
    # otadata + phy_init are followed by app0 which requires 64KB alignment,
    # so pad up to the next 64KB boundary.
    overhead = (0x9000 + OTADATA_SIZE + PHY_INIT_SIZE + 0xFFFF) & ~0xFFFF
```

esphome/components/esp32/__init__.py
```diff
@@ -1836,45 +1905,170 @@ async def to_code(config):
...
+    ]
+    if CORE.using_arduino:
+        partitions.append(
+            PartitionEntry(name="eeprom", type="data", subtype="0x99", size=EEPROM_SIZE)
+        )
+        partitions.append(
+            PartitionEntry(
+                name="spiffs", type="data", subtype="spiffs", size=SPIFFS_SIZE
+            )
+        )
```
luar123, line 2060: I don't know the internals, but are these two even needed? Afaik both arduino libs, eeprom and spiffs, are removed by default. And if components need to add the library they could also add these partitions.

luar123 (commented):
Awesome, thank you!

I wanted to keep it simple, but the refactor makes a lot of sense.

esphome/components/esp32/__init__.py
swoboda1337: Not a huge deal to include them only shrinks nvs a bit. If people want to enable later they would have to reflash.

esphome/components/esp32/__init__.py
swoboda1337: It was never placed at a fixed address on IDF:
```
otadata,  data, ota,     ,        0x2000,
phy_init, data, phy,     ,        0x1000,
app0,     app,  ota_0,   ,        0x{app_partition_size:X},
app1,     app,  ota_1,   ,        0x{app_partition_size:X},
nvs,      data, nvs,     ,        0x6D000,
```

swoboda1337: I assume the tool will take care of it

luar123: > I assume the tool will take care of it

Yes it does. But for the size calculation to be robust it should be considered. At the moment it won't make a difference because it is <64KB anyway.

First partition starts always at 0x9000. 0x8000 for bootloader + 0x1000 for partition table.

swoboda1337: Ok yeah could probably add defines for those.

swoboda1337 (approved):

bdraco (approved):

luar123:
Thanks!

swoboda1337:
Thanks!

PR #7689, opened 2024-10-28
[http_request] Always return defined server response status

### LINKED ISSUE #6010: http_request now produce an unspecified error and exit container instead of reporting the client code error
### The problem

Older versions of http request were reporting client error codes when connections was failing  such as " -1"  from the on_response trigger

This could be used for determining if a device was up or not and publish state for a sensor or a switch for example.
It now fails with an unspecified error  wich is cleared later. 

The container now  exits and no return code or other data are provided.

2024.6.6

pip

2024.7.0

ESP8266

### Board

esp01_1m

http_request

### Example YAML snippet

```yaml
esphome:
  name: ir_0
esp8266:
  board: esp01_1m
  framework:
    version: 3.0.2 
  restore_from_flash: true

http_request:
  id: http_request_data
  esp8266_disable_ssl_support: true
  #timeout: 0.5s

### ... sensor,  switches, etc 


script:
####### get fbr state ############
    - id: get_fbr_state
      mode: single
      then:
        - http_request.get:
             url: "http://192.168.0.253:54243/device.xml"
             capture_response: true
             on_response:
               then:
                 - lambda: |-
                    if (response->status_code == 200) {
                      id(fbx_oo).publish_state(true);
                    } else {
                      id(fbx_oo).publish_state(false);
                    };

# or 
    - id: get_lgtv_state
      mode: single
      then:
        - http_request.get:
            url: "http://192.168.0.251/binary_sensor/tv_state"
            #capture_response: true
            on_response:
               then:
                - logger.log:
                    format: "LGTV request Status code %d"
                    args:
                      - response->status_code
                - lambda: |-
                    if (response->status_code == 200) {
                      id(lg_oo).publish_state(true);
                    } else if (response->status_code == -1 ) {
                      id(lg_oo).publish_state(false);
                    };
```

```txt
[17:14:20][W][http_request.arduino:104]: HTTP Request failed; URL: http://192.168.0.253:54243/device.xml; Error: connection failed
[17:14:20][E][component:164]: Component http_request set Error flag: unspecified

[17:14:21][E][component:176]: Component http_request cleared Error flag
```

### Additional information

_No response_

clydebarrow:

When a server responds with a status code outside [200-299] the http_request code returns a null pointer, denying the calling code the opportunity to examine the response. Instead, return the container with the status code and body (which will likely contain an error message.)

This will mean calling code needs to check the status code, though it should already be doing so, and prior to 2024.6 release, had to do so anyway. But since it's possible some user code written since 2024.6 does not, this has been classed as a breaking change.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6010

- esphome/esphome-docs#4397

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
    - http_request.get:
        capture_response: true
        url: "http://192.168.1.131/bad"
        on_response:
          logger.log:
            format: "Bad Response code %d, body %s"
            args: [response->status_code, body.c_str()]
```

## Checklist:


jesserockz (changes requested):
This will require all existing code calling start/get/post etc that does
```c++
if (container == nullptr) {
```
to handle a bad status code in the same way as they do the nullptr.

clydebarrow:
> to handle a bad status code in the same way as they do the nullptr.

Correct, and in that sense it's a breaking change, but the code should *already* be checking the status code since it can already be in the range 200-299, and typically only 200 means there is valid data. This basically restores the functionality to what it was before release 2024.6.

I looked for usages that were not checking the status code and found a couple, have updated them.

clydebarrow:
It's important for user code to have the response and body returned even when the server returns a status other than 200, since even a 404 response will usually have some kind of error message in the body, as will other codes like 401 (unauthorised) and 500 (server error.)

jesserockz (approved):

PR #7711, opened 2024-11-01
handle bad pin schemas

ssieb:

If a pin schema is a list, the validation shouldn't crash.

e.g.
```yaml
pin:
  - number: 4
```

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/esp32/gpio.py
+++ b/esphome/components/esp32/gpio.py
@@ -69,6 +69,8 @@ def _translate_pin(value):
         )
     if isinstance(value, int):
         return value
+    if isinstance(value, list):
+        raise cv.Invalid("A pin schema can't be a list.")
     try:
         return int(value)
     except ValueError:
--- a/esphome/components/esp32/gpio.py
+++ b/esphome/components/esp32/gpio.py
@@ -67,10 +67,10 @@ def _translate_pin(value):
             "This variable only supports pin numbers, not full pin schemas "
             "(with inverted and mode)."
         )
-    if isinstance(value, int):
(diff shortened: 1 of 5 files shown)


ssieb:
"ruff" decided to fix all the imports as well.

esphome/components/esp32/gpio.py
```diff
@@ -69,6 +69,8 @@ def _translate_pin(value):
         )
     if isinstance(value, int):
         return value
+    if isinstance(value, list):
+        raise cv.Invalid("A pin schema can't be a list.")
```
clydebarrow: At this point the value should only be a string. Ditto for the other cases.
```suggestion
    if not isinstance(value, str):
        raise cv.Invalid(f"Invalid pin number: {value}")
```

clydebarrow (changes requested):

clydebarrow:
Another quirk - this passes validation:

```
spi:
  clk_pin: false
```

Apparently `isinstance(value, int)` is true when `value` is actually a bool! Evidently bool is a subtype of int.

So the first test should probably be:
```
    if type(value) == type(int):
        return value
```

I think that, with the subsequent test for `str`, will cover all possible types generated from yaml.

clydebarrow (approved):

PR #7241, opened 2024-08-10
[hlw8032] Single-phase metering IC

rici4kubicek:
# What does this implement/fix?

Implementation of Single-phase metering IC

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4124

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
sensor:
  - platform: hlw8032
    voltage_divider: 1.72
    voltage:
      name: HLW8032 Voltage
      id: hlw8032_voltage
    current:
      name: HLW8032 Current
      id: hlw8032_current
    power:
      name: HLW8032 Power
      id: hlw8032_power
    apparent_power:
      name: HLW8032 Apparent Power
      id: hlw8032_apparent_power
    power_factor:
      name: HLW8032 Power Factor
      id: hlw8032_power_factor

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #7720, opened 2024-11-04
[cse7766] Remove ``stream`` dependency

gabest11:

Shave off 200k from the firmware when you log the raw measurement data.

iomanip includes everything not just the kitchen sink. 

I tried debugging my smart plug (somehow only reports voltage), and with a barebone config, debug level none, the firmware is over half the flash size with this.

The fix is to use regular snprintf to dump those bytes.

## Types of changes

- [ ] Other

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/cse7766/cse7766.cpp
+++ b/esphome/components/cse7766/cse7766.cpp
@@ -71,12 +71,12 @@ bool CSE7766Component::check_byte_() {
 void CSE7766Component::parse_data_() {
 #if ESPHOME_LOG_LEVEL >= ESPHOME_LOG_LEVEL_VERY_VERBOSE
   {
-  	std::string s = "Raw data:";
-  	char buff[4] = {0};
-  	for (uint8_t i = 0; i <= 23; i++) {
-  		snprintf(buff, sizeof(buff), " %02X", this->raw_data_[i]);
-  		s += buff;
-  	}
+    std::string s = "Raw data:";
+    char buff[4] = {0};
+    for (uint8_t i = 0; i <= 23; i++) {
+      snprintf(buff, sizeof(buff), " %02X", this->raw_data_[i]);
+      s += buff;
+    }
     ESP_LOGVV(TAG, "%s", s.c_str());
   }
 #endif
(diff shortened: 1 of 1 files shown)


ssieb:
I don't know why that's taking space when it's not being used, but also remove the stream header as well.  And change the code to use https://github.com/esphome/esphome/blob/dev/esphome/core/helpers.cpp#L379 instead of doing the formatting itself.  This is actually a change I've wanted done for a while.  We shouldn't be using streams, it's too big.

jesserockz (dismissed):

gabest11:
Did the changes. Still fighting with my device, it's a hlw8032 but should be compatible. It used to work, then as cse7766 component was updated, it degraded, displayed low voltage, then suddenly nothing. Now swapped the chip to a new one from aliexpress and the voltage is back, but no current. Might be defective, but not sure.

gabest11:
> I don't know why that's taking space when it's not being use

Maybe I was not very clear, it does not when ifdef'ed out, but I needed to see those raw values and enabled that piece of code outside logging levels. Going veryveryverbose was not possible either, that's just even more data.

rici4kubicek:
> Still fighting with my device, it's a hlw8032 but should be compatible.

If you want specific driver for HLW8032, you can use a driver from PR #7241. You do not need to specific any post-process calculation. You can only define necessary resistor values in the driver setting.

gabest11:
It was my smart plug breaking down. Bought a different brand with this chip and it works fine. Any component. Even what I wrote myself. https://github.com/esphome/esphome/compare/dev...gabest11:esphome:hlw8032

DjordjeMandic (approved):

kbx81 (approved):
I talked with Jesse, we prefer to use `str_sprintf`, so I've pushed a minor change to do this (didn't feel like commenting on all those lines individually. Sorry, haha). Otherwise LGTM

PR #7722, opened 2024-11-05
[lvgl] Don't just throw key error if someone types a bad layout type

jesserockz:

From:
```
  File "/home/jesse/workspace/esphome/esphome/esphome/components/lvgl/schemas.py", line 394, in validator
    result = result.extend(LAYOUT_SCHEMAS[ltype.lower()])
                           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'blah'
```

to:

```
lvgl: [source config/dining-switch-display.yaml:124]
  ...
  pages: 
    - ...
      widgets: 
        - obj: 
            ...
            layout: 
              
              Unknown value 'blah', valid options are 'grid', 'flex', 'none'.
              type: blah
            widgets: 
              ...
```
## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/lvgl/schemas.py
+++ b/esphome/components/lvgl/schemas.py
@@ -391,7 +391,9 @@ def container_validator(schema, widget_type: WidgetType):
             add_lv_use(ltype)
         if value == SCHEMA_EXTRACT:
             return result
-        result = result.extend(LAYOUT_SCHEMAS.get(ltype.lower(), LAYOUT_SCHEMAS[df.TYPE_NONE]))
+        result = result.extend(
+            LAYOUT_SCHEMAS.get(ltype.lower(), LAYOUT_SCHEMAS[df.TYPE_NONE])
+        )
         return result(value)
 
     return validator
--- a/esphome/components/lvgl/schemas.py
+++ b/esphome/components/lvgl/schemas.py
@@ -391,6 +391,8 @@ def container_validator(schema, widget_type: WidgetType):
             add_lv_use(ltype)
         if value == SCHEMA_EXTRACT:
             return result
+        with cv.prepend_path([df.CONF_LAYOUT, CONF_TYPE]):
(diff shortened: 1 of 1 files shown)


esphome/components/lvgl/schemas.py
```diff
@@ -391,6 +391,8 @@ def validator(value):
             add_lv_use(ltype)
         if value == SCHEMA_EXTRACT:
             return result
+        with cv.prepend_path([df.CONF_LAYOUT, CONF_TYPE]):
+            ltype = cv.one_of(*LAYOUT_SCHEMAS.keys())(ltype)
         result = result.extend(LAYOUT_SCHEMAS[ltype.lower()])
```
clydebarrow:
```suggestion
        result = result.extend(LAYOUT_SCHEMAS.get(ltype.lower(), LAYOUT_SCHEMAS[df.TYPE_NONE]))
```

jesserockz: Config validation should be strict. If there is a fixed set of options, it should always be telling the user their choice is not valid and present the list of options and not just fallback to some other value.

clydebarrow: That's exactly what will happen, just in the normal config flow, rather than an explicit check. The config will still contain an invalid type, which will be flagged as expected.

clydebarrow: This part of the code is constructing the schema, not validating it. It looks for a layout type solely so it can choose the right schema to use for the `widgets` key. Validation of the `layout` key is already done by the pre-defined schema.

clydebarrow (approved):

PR #7725, opened 2024-11-06
Store yaml in firmware

gabest11:

This component can be used to store the flattened yaml in the firmware, to be retrieved in case the original was lost.

I was thinking about adding this into proto.api / DeviceInfoResponse, but that would have been a bit rude, maybe. And this way it works with MQTT too.

The needed size in the firmware is typically a few kilobytes. The user can decide if it's worth it.

About documentation. I can create a new page for the component, but someone should also modify that section about yaml recoverability in the FAQ.

Btw, this is not a sensor, because HA can only accept 255 characters in a text_sensor.

They way it works: `__init__.py` compresses `CORE.config` with a basic dictionary based compression into a global const byte array (`ESPHOME_YAML`) and it can be logged with an action. There is also an option to show it during `dump_config`, but currently this may trigger a watchdog reboot if the yaml could not be sent fast enough. 

## Types of changes

- [ ] Other

- fixes [Can Configuration Files Be Recovered From The Device?](https://esphome.io/guides/faq.html#can-configuration-files-be-recovered-from-the-device)

- esphome/esphome-docs#4425

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [x] BK72xx
- [x] RTL87xx

```yaml
# Example config.yaml
store_yaml:
  show_in_dump_config: True
  show_secrets: True
  http:
    url: /config

api:
  services:
    - service: 'store_yaml_log'
      then:
        - store_yaml.log
```

## Checklist:


TrentHouliston:
If the purpose is to store it for later retrieval, compressing it would save a lot of space

gabest11:
I was thinking about it, but every real compression library would take up more code space than the yaml itself. And the decompression algorithm should be something that python can compress.

This is the output of this nice progam, `elf-size-analyze -t /config/.esphome/platformio/packages/toolchain-xtensa/bin/xtensa-lx106-elf- -w 120 -HaF /config/.esphome/build/test/.pioenvs/test/firmware.elf -W` Notice the highlighted sizes. My test yaml with my basic includes. Also, INDEX_GZ, but captive portal does not have any decompression included, it just sends it to the web browser as-is.

Screenshot 2024-11-08 070420

DjordjeMandic:
Any way to store it in flash without pushing it into ram? I may be wrong but this code looks like whole string is in ram all the time. 

It could be served via new API call / mqtt topic?

For use, alloc memory required to store it, read it from flash to ram, dump it and free the memory, get rid of pointer.

gabest11:
> Any way to store it in flash without pushing it into ram? I may be wrong but this code looks like whole string is in ram all the time.

Going to check it with debug component. With and without.

> It could be served via new API call / mqtt topic?

As far I know, calls don't return anything. And HA can only take long strings in attributes, which esphome does not support. A web server or the proto.api could be used, but I would not go that far myself.

> For use, alloc memory required to store it, read it from flash to ram, dump it and free the memory, get rid of pointer.

Not sure how. There is no spiffs partition with esphome.

gabest11:
I can see a 7k difference in Heap Free with a large yaml. It's definitely in memory.

gabest11:
If I replace std::string with const char*, just to point to the string from main.cpp, which will never be deallocated, I can save that 7k.

gabest11:
Tried putting the string into a global const char[], same results.

DjordjeMandic:
> > For use, alloc memory required to store it, read it from flash to ram, dump it and free the memory, get rid of pointer.
> 
> Not sure how. There is no spiffs partition with esphome.

Take a look at webserver component, it stores web pages in flash as array. If codegen can be modified, py script could generate header file containing compressed yaml as array of bytes from program memory (flash).

gabest11:
I tried that. That's what I ment by storing it in a global variable. I think the string literal compiles to the same thing. Somewhere I read that PROGMEM makes no difference on ESP32.

gabest11:
And the web server does not have to decompress it, there is no code to do it, it exploits http compression. Browsers can accept gzipped pages.

gabest11:
Free heap as I was trying different approaches. Lower is with std::string. The more free memory is with a simple pointer to the argument or the global.

Screenshot 2024-11-08 185217

TrentHouliston:
Typically if I was trying to ensure that a symbol occupied a specific location in memory on another embedded platform I'd be playing with the linker script and potentially`__attribute__((section("name")))` to put it in a specific location. I'm not sure if something like that is possible within how esphome is set up though.

My other thought is that it's not actually the constant that's not getting stored in flash, but rather when you try to use the constant it's being copied into an std::string which would explain why the constant version is taking up more memory than the string version, you're storing both a string and a constant.

As for compression, I'm not sure that the device itself has any use for reading the yaml either, so compressing it and then storing the compressed `.yaml.gz` etc and just serving that would save space without requiring a compression/decompression library on the device itself.

gabest11:
I ran into a small problem.  !extend does not resolve with load_yaml+dump. I could fall back to CORE.config, but that has everything, all the default configurations, and auto-generated ids. At least there is this function called strip_default_ids.

gabest11:
Compression saves 50%-70%, but the data is also bigger now. My largest file is 8k, it cannot even list it in the debug output fast enough, the watchdog resets it. Maybe I could feed it in smaller chunks from loop(). Any ideas?

DjordjeMandic:
Where does wdt timeout happen? dump_config should be fast.

I know that audio and camera components create new task with xtaskcreate to prevent wdt timeouts. I might be wrong tho.

Could you time how much time does decompression take?

Also, i think it might be good time to mark this pr as draft because its work in progress.

nielsnl68:
Maybe a late question, but why did you want to have the compression/decompression inside the firmware. From your earlier messages i got the idea that you only wanted to store the yaml content. Why not compressing in `__init__.py` and just store the result. And when needed send the decompress it in the dashboard.

gabest11:
> Where does wdt timeout happen? dump_config should be fast.

There is one long yaml I have with 100+ entities, it reboots after a few seconds stuck in dump_config. The code runs fast, but the receiveing end is slow. In the browser it is very slow, command line (esphome log ...) moderately, and serial is almost fast enough. I moved the action response into loop() and every time it only prints a single line, now it survives nicely. The single run dump_config is out of the question for large yamls.
 
> I know that audio and camera components create new task with xtaskcreate to prevent wdt timeouts. I might be wrong tho.

That loop() solution works now, I'm satisfied with it.

> Could you time how much time does decompression take?

Instantly, it's very basic. Logging is slow as I said.

> Also, i think it might be good time to mark this pr as draft because its work in progress.

If we don't count the dump_config problem, it's usable as it is right now. Maybe I'm going to trim that output to a shorter version. First 100 lines or so.

gabest11:
> Maybe a late question, but why did you want to have the compression/decompression inside the firmware. From your earlier messages i got the idea that you only wanted to store the yaml content. Why not compressing in `__init__.py` and just store the result. And when needed send the decompress it in the dashboard.

That's exactly what's happening. You are not late, I just finished it yesterday.

DjordjeMandic:
I agree with suggestion from above. 

1) this now is fine, compress it during verification process, store it in flash. Loop() print works currently 

2) update api to support sending raw compressed data from flash back to dashboard only if authenticated

3) decompress in dashboard if needed

nagyrobi:
also see https://github.com/dentra/esphome-components/tree/master/components/backup @dentra

gabest11:
> also see https://github.com/dentra/esphome-components/tree/master/components/backup @dentra

Looks very similar to this one, just with a web server.

nagyrobi:
Yeah just no need to reinvent the wheel, maybe good for inspiration.

gabest11:
Components not in the main project don't exist for the user. And the FAQ just said it was not possible.

esphome/components/store_yaml/store_yaml.cpp
```diff
@@ -0,0 +1,88 @@
...
+void StoreYamlComponent::dump_config() {
+  if (this->show_in_dump_config_) {
+    ESP_LOGCONFIG(TAG, "YAML:");
+    RowDecompressor dec(ESPHOME_YAML, ESPHOME_YAML_SIZE);
+    std::string row;
+    while (dec.get_row(row)) {
+      ESP_LOGCONFIG(TAG, "%s", row.c_str());
+    }
+  }
+}
```
DjordjeMandic, line 24: Add more config info independent of `show_in_dump_config_`

```suggestion
void StoreYamlComponent::dump_config() {
  ESP_LOGCONFIG(TAG, "YAML:");
  ESP_LOGCONFIG(TAG, "  Compressed size: %zu", ESPHOME_YAML_SIZE);
  ESP_LOGCONFIG(TAG, "  Web server url: %s", (this->web_server_ != nullptr) ? this->web_server_url_ : "not configured!");
  if (this->show_in_dump_config_) {
    ESP_LOGCONFIG(TAG, "  Decompressed content:");
    RowDecompressor dec(ESPHOME_YAML, ESPHOME_YAML_SIZE);
    std::string row;
    while (dec.get_row(row)) {
      ESP_LOGCONFIG(TAG, "    %s", row.c_str());
    }
  }
}
```

esphome/components/store_yaml/store_yaml.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+void StoreYamlComponent::setup() {
+  if (this->web_server_ != nullptr) {
+    this->web_server_->init();
+    this->web_server_->add_handler(this);
+  }
```
DjordjeMandic, line 32:
```suggestion
  if (this->web_server_ != nullptr) {
    this->web_server_->init();
    this->web_server_->add_handler(this);
    ESP_LOGD(TAG, "Web server configured to serve at: %s", this->web_server_url_.c_str());
  }
```

esphome/components/store_yaml/store_yaml.cpp
```diff
@@ -0,0 +1,88 @@
...
+    if (this->dec_->get_row(row)) {
+      ESP_LOGI(TAG, "%s", row.c_str());
+    } else {
+      this->dec_ = nullptr;
+    }
+  }
+}
+
+void StoreYamlComponent::log() {
+  ESP_LOGI(TAG, "YAML:");
```
DjordjeMandic:
```suggestion
  ESP_LOGI(TAG, "Decompressed YAML:");
```

DjordjeMandic (changes requested):
Make logs more useful even if `show_in_dump_config_` is false

gabest11:
Does RP2040 really have no web server support? No captive portal, nothing. The automated test caught it. This will be a lot of ifdefs.

gabest11:
Why this script/ci-custom test cannot see CORE.target_platform. I have no problem compiling the firmware.

gabest11:
Just noticed the web server request does nothing the second time. Neighter way, arduino or esp-idf.

esphome/components/store_yaml/decompressor.h
DjordjeMandic, line 1: In c++ default access specifier is private. In esphome we dont use private. Make it protected.

esphome/components/store_yaml/decompressor.cpp
```diff
@@ -0,0 +1,125 @@
...
+  this->data_len_ = len;
+  this->reset();
+}
+
+void Decompressor::reset() {
+  this->pos_ = 0;
+  this->size_ = 0;
+  this->buff_ = 0;
+  this->codes_.clear();
+  for (uint32_t i = 0; i < 256; i++) {
```
DjordjeMandic:
```suggestion
  for (uint16_t i = 0; i < 256; i++) {
```

esphome/components/store_yaml/decompressor.cpp
```diff
@@ -0,0 +1,125 @@
...
+    this->pos_ = this->data_len_;  // error in input, set eof
+  }
+  return nullptr;
+}
+
+std::string Decompressor::get_string_(const Entry *entry) const {
+  std::string s(1, (char) entry->c);
+  while (entry->p != 0) {
+    if (entry->p >= this->codes_.size()) {
+      ESP_LOGE(TAG, "entry->p %d not found", entry->p);
```
DjordjeMandic:
```suggestion
      ESP_LOGE(TAG, "entry->p %" PRIu16 " not found", entry->p);
```
`entry->p` is `uint16_t` type right?

esphome/components/store_yaml/decompressor.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+bool Decompressor::is_eof() const { return this->pos_ >= this->data_len_; }
+
+uint32_t Decompressor::get_bits_(size_t bits) {
+  if (this->is_eof()) {
+    return UINT32_MAX;
+  }
+
+  while (this->size_ < bits) {
+    this->buff_ = (this->buff_ << 8) | progmem_read_byte(&this->data_ptr_[this->pos_++]);
```
DjordjeMandic, line 37: AFAIK progmem functions are part of avr-libc. They're ported to arduino as compatibility layer for code. Were trying to switch to esp-idf completely.

https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/pgmspace.h

Does this code compile when ESP-IDF framework is used without arduino wrapper?

esphome/components/store_yaml/decompressor.h
```diff
@@ -0,0 +1,49 @@
...
+  // uint32_t p : 24;
+  // uint32_t c : 8;
+  uint16_t p;  // save 1 byte, this reduces the number of code words to 65536, more than enough for configs
+  uint8_t c;
+} __attribute__((packed));
+
+class Decompressor {
+  const uint8_t *data_ptr_;
+  size_t data_len_;
+  std::vector<Entry> codes_;
```
DjordjeMandic, line 20: Is max size of `codes_` known or could be made known at compile time? If yes, reserve enough space during construction of Decompressor to prevent unnecessary memory reallocations. https://cplusplus.com/reference/vector/vector/reserve/

esphome/components/store_yaml/store_yaml.h
DjordjeMandic, line 1: Same as for `decompressor.h`. No private allowed.

DjordjeMandic (changes requested):

DjordjeMandic:
Could you test heap usage and heap fragmentation? Log them to home assistant, call log action multiple times. Im worried that it might fragment the heap too much.

gabest11:
> Could you test heap usage and heap fragmentation? Log them to home assistant, call log action multiple times. Im worried that it might fragment the heap too much.

The only memory issue I noticed was on a fully loaded esp8266 with lower than 10k free memory. I had to disable this, because once every day it stopped responding.

esphome/components/store_yaml/decompressor.cpp
gabest11: I don't really care about these, since smaller than int are promoted and passed as int.

esphome/components/store_yaml/decompressor.cpp
gabest11: progmem_read_byte just returns the pointer if the target board isn't esp8266. (or rp2040, I don't have that)

esphome/components/store_yaml/decompressor.cpp
DjordjeMandic: You're right, it will work for now. I just wanted to future-proof it. If entry->p is changed to uint32_t and the MSB gets set, it could print negative values. Similarly, on platforms where int is 2 bytes, the same issue would occur if the MSB of a uint16_t gets set. Since ESPHome is becoming more of a multi-platform project, I think it's good to account for these edge cases early.

DjordjeMandic: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/migration-guides/release-5.x/5.0/gcc.html#espressif-toolchain-changes

I dont know whats best practice here, this created unwanted headache in ESPHome as i understood it.

esphome/components/store_yaml/decompressor.cpp
DjordjeMandic: You're correct, my bad. I didn't see this https://github.com/esphome/esphome/blob/f7f8bf4da44b83796a301bc486d77fdb46d8ed82/esphome/core/hal.h#L46

esphome/components/store_yaml/decompressor.h
gabest11: That's a difficult one, I'll think about it. The beauty of LZW is that the dictionary is not transmitted, but dynamically built on both ends. The python compressor knows in the end, so it should be doable.

DjordjeMandic (approved):

esphome/components/store_yaml/decompressor.cpp
gabest11: This PRIu16 breaks compilation with the current Arduino framework. Will this be ever fixed?

DjordjeMandic: https://github.com/esphome/esphome/pull/8155

Does this fix it? Merge latest dev into gabest11/esphome#store_yaml

gabest11: That helped!

bdraco:
@gabest11 are you still interested in working on this? If so, I'll do a full review, as I've had users asking for something like this.

PR #7733, opened 2024-11-08
LD2410 remove NaN state in engineering mode

kbullet:

When engineering mode is disabled set state to 0 instead of NaN for move & still energy sensors, since Hass's MQTT integration doesn't seem to like them NaN values and its spamming the logs e.g.

> ValueError: Sensor sensor.ld2410_sensor_g8_move_energy has device class 'None', state class 'None' unit '%' and suggested precision 'None' thus indicating it has a numeric value; however, it has the non-finite value: 'nan'
> 2024-11-08 15:43:41.169 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved (None)
> Traceback (most recent call last):
>   File "/usr/src/homeassistant/homeassistant/components/mqtt/entity.py", line 856, in _async_process_discovery_update
>     await discovery_update(payload)
>   File "/usr/src/homeassistant/homeassistant/components/mqtt/entity.py", line 1137, in discovery_update
>     self.async_write_ha_state()
>   File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1005, in async_write_ha_state
>     self._async_write_ha_state()
>   File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1130, in _async_write_ha_state
>     self.__async_calculate_state()
>   File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1067, in __async_calculate_state
>     state = self._stringify_state(available)
>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1011, in _stringify_state
>     if (state := self.state) is None:
>                  ^^^^^^^^^^
>   File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 675, in state
>     raise ValueError(
> ValueError: Sensor sensor.ld2410_sensor_g8_still_energy has device class 'None', state class 'None' unit '%' and suggested precision 'None' thus indicating it has a numeric value; however, it has the non-finite value: 'nan'

## Types of changes

- [ ] Other

N/A

N/A

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:

### DIFF
--- a/esphome/components/ld2410/ld2410.cpp
+++ b/esphome/components/ld2410/ld2410.cpp
@@ -255,16 +255,16 @@ void LD2410Component::handle_periodic_data_(uint8_t *buffer, int len) {
   } else {
     for (auto *s : this->gate_move_sensors_) {
       if (s != nullptr && !std::isnan(s->get_state())) {
-        s->publish_state(NAN);
+        s->publish_state(0);
       }
     }
     for (auto *s : this->gate_still_sensors_) {
       if (s != nullptr && !std::isnan(s->get_state())) {
-        s->publish_state(NAN);
+        s->publish_state(0);
       }
     }
     if (this->light_sensor_ != nullptr && !std::isnan(this->light_sensor_->get_state())) {
-      this->light_sensor_->publish_state(NAN);
+      this->light_sensor_->publish_state(0);
     }
   }
 #endif


esphome/components/ld2410/ld2410.cpp
```diff
@@ -255,16 +255,16 @@ void LD2410Component::handle_periodic_data_(uint8_t *buffer, int len) {
   } else {
     for (auto *s : this->gate_move_sensors_) {
       if (s != nullptr && !std::isnan(s->get_state())) {
-        s->publish_state(NAN);
+        s->publish_state(0);
```
regevbr, line 258: Notice the check above... once you set it to 0 it will keep publishing the state and that is not desirable, please fix

kbullet: Thanks for the review. Correct me if I'm wrong, after further testing once engineering is disabled, it only publishes the state to 0 once, similarly like it did with the NaN state. 

Though if not a larger task is to extend the sensor/mqtt component to handle NaN to the liking of the changes of Hass's mqtt integration

regevbr: I honestly don't remember why I did that protection, but you can notice that before every publish we guard against publishing if value didn't change. Here you changed that logic

kbullet: Apologies if my explanation was unclear. 

In this engineering mode logic. When it's off, no more data is being published and those sensors are basically off/unavailable. So instead of publishing NaN to tell the sensor is unavailable. We use 0 instead to stop Homeassistant complaining about NaN values in the sensor state. If that makes any sense.

Though my best guess wrt to your guard there is to stop it from constantly spamming the NaN/Unavailable state when engineering mode is off.

regevbr: I understand yet the condition still needs to be fixed

jesserockz: NAN is the correct value to use here. The ESPHome native API handles this by setting the value in HA to `None`. seems the mqtt code expects it to be a literal `"None"` to achieve the same. So maybe the mqtt sensor code in ESPHome should send `"None"` if the value from the actual sensor is NAN.

https://github.com/esphome/esphome/blob/0b51ec2c88793191f8d28fecfe9b514b9195312f/esphome/components/mqtt/mqtt_sensor.cpp#L71-L74

kbullet: Yeah after further digging and trying out the API, NAN is the correct state and used throughout ESPHome. So you're absolutely right probably it is best to be handled on mqtt rather at individual components level. I will close this PR and open a new one

PR #7740, opened 2024-11-09
Updated dfplayer logging to be more user-friendly

solarkennedy:

I spent a number of hours troubleshooting a dfplayer, decoding hex output and stuff.

I think the logging can be improved to save time in the future. Specifically:

* LOGD for all actual dfplayer actions in human-readable form
* LOGV for any hex dump (was LOGD previously)
* LOGE for all errors, and expand the coverage for error codes

## Types of changes

- [ ] Other

- https://github.com/esphome/issues/issues/5174 is an example of an issue that would probably be easily debuggable with this PR in place.

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/dfplayer/dfplayer.cpp
+++ b/esphome/components/dfplayer/dfplayer.cpp
@@ -4,9 +4,8 @@
 namespace esphome {
 namespace dfplayer {
 
-static const char *const TAG = "dfplayer";
-
 void DFPlayer::play_folder(uint16_t folder, uint16_t file) {
+  ESP_LOGD(TAG, "Playing file %d in folder %d", file, folder);
   if (folder < 100 && file < 256) {
     this->ack_set_is_playing_ = true;
     this->send_cmd_(0x0F, (uint8_t) folder, (uint8_t) file);
@@ -29,7 +28,7 @@ void DFPlayer::send_cmd_(uint8_t cmd, uint16_t argument) {
 
   this->sent_cmd_ = cmd;
 
-  ESP_LOGD(TAG, "Send Command %#02x arg %#04x", cmd, argument);
+  ESP_LOGV(TAG, "Send Command %#02x arg %#04x", cmd, argument);
   this->write_array(buffer, 10);
 }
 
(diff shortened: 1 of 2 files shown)


solarkennedy:
```
Error: esphome/components/dfplayer/dfplayer.h:111:5: lint: Found reference to ESP_LOG in header file. Using ESP_LOG* in header files is currently not possible - please move the definition to a source file (.cpp)
```

Hmm. I'm surprised because it works locally? But I can move them.

esphome/components/dfplayer/dfplayer.cpp
```diff
@@ -4,9 +4,8 @@
 namespace esphome {
 namespace dfplayer {
 
-static const char *const TAG = "dfplayer";
```
DjordjeMandic, line 7: Keep tag inside CPP file. Line number in logs is generated based on line number inside that cpp file.

So do not delete it, also keep it inside namespace esphome::dfplayer like it was before

esphome/components/dfplayer/dfplayer.h
```diff
@@ -5,6 +5,7 @@
 #include "esphome/core/automation.h"
 
 const size_t DFPLAYER_READ_BUFFER_LENGTH = 25;  // two messages + some extra
+static const char *const TAG = "dfplayer";
```
DjordjeMandic: I'd move this to cpp file, dont log inside header file. Line number in logs is based on line number in cpp file.

esphome/components/dfplayer/dfplayer.h
```diff
@@ -23,62 +24,91 @@ enum Device {
   TF_CARD = 2,
 };
 
+// See the datasheet here:
+// https://github.com/DFRobot/DFRobotDFPlayerMini/blob/master/doc/FN-M16P%2BEmbedded%2BMP3%2BAudio%2BModule%2BDatasheet.pdf
 class DFPlayer : public uart::UARTDevice, public Component {
  public:
   void loop() override;
 
   void next() {
```
DjordjeMandic: Declare functions that require logging in header, define them in cpp (cpp file should contain body of function in this case since `ESP_LOG*` is not allowed in header file).

DjordjeMandic (changes requested):
Move tag const back to cpp since no logging is allowed inside header files. If you wish to keep it in header, at least move it inside dfplayer namespace

If function needs logging, declare it only in header, define it in cpp.

`core/log.h` uses `__LINE__` macro to get current line number in SOURCE file. Using header in different source files would result in different line numbers. I'd say that's only reason why ESP_LOG* is not allowed inside header files. It just makes line number in logs meaningless since it's going to change often.

DjordjeMandic:
> Hmm. I'm surprised because it works locally? But I can move them.

It works but compare line numbers in log entries with ESP_LOG* in header and in cpp

esphome/components/dfplayer/dfplayer.h
solarkennedy: got it, ty

esphome/components/dfplayer/dfplayer.cpp
```diff
@@ -6,7 +6,89 @@ namespace dfplayer {
 
 static const char *const TAG = "dfplayer";
 
+void DFPlayer::next() {
+  this->ack_set_is_playing_ = true;
+  ESP_LOGD(TAG, "Playing next track");
+  this->send_cmd_(0x01);
+}
+void DFPlayer::previous() {
```
DjordjeMandic: Maybe add a new line between methods? Not trivial but makes it more readable.

esphome/components/dfplayer/dfplayer.cpp
```diff
@@ -101,8 +183,26 @@ void DFPlayer::loop() {
...
+              ESP_LOGE(TAG, "Specified track is not found");
+              this->is_playing_ = false;
+            } else if (argument == 0x07) {
+              ESP_LOGE(TAG, "Insertion error (an inserting operation only can be done when a track is being played)");
+            } else if (argument == 0x08) {
+              ESP_LOGE(TAG, "SD card reading failed (SD card pulled out or damaged)");
+            } else if (argument == 0x09) {
+              ESP_LOGE(TAG, "Entered into sleep mode");
               this->is_playing_ = false;
             }
```
DjordjeMandic: Nothing wrong with it, if you wish you could switch thru argument instead of else if for each case.

DjordjeMandic (commented):
Looks good, maybe make it more readable and that's it.

DjordjeMandic (approved):

jesserockz (changes requested):
Agree with the points made above

DjordjeMandic (approved):

jesserockz (approved):

PR #7749, opened 2024-11-11
[i2s_audio] I2S speaker improvements

kahrendt:

- (Bugfix) Fixes a rare crash when the speaker tries to stop while another component is writing to the ring buffer. Uses a ``shared_ptr`` to manage the ring buffer to prevent this instead of using an event group bit to lock the ring buffer.
- (Bugfix) Properly detects DMA buffer underflows so that the ``finish`` function behaves correctly. Attaches a queue during the I2S driver installation to monitor this.
- (New feature) Adds a yaml configuration option to disable stopping the speaker due to a timeout.
- (New feature) Allocates the DMA buffers and ring buffers based on the configured audio stream information. Instead of using fixed sample numbers, it takes into account the number of channels, bytes per sample, and the sample rate for consistent behavior with different settings.

Tested successfully on an ATOM Echo, Voice PE, and S3 box.

## Types of changes

- [ ] Other

- closes #7677, as the finish function now behaves properly and will avoid a breaking change

- esphome/esphome-docs#4426

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

speaker:
  - platform: i2s_audio
    i2s_dout_pin: GPIO15
    dac_type: external
    sample_rate: 16000
    bits_per_sample: 16bit
    channel: stereo
    timeout: never

```

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/speaker/__init__.py
+++ b/esphome/components/i2s_audio/speaker/__init__.py
@@ -2,7 +2,7 @@ from esphome import pins
 import esphome.codegen as cg
 from esphome.components import esp32, speaker
 import esphome.config_validation as cv
-from esphome.const import CONF_CHANNEL, CONF_ID, CONF_MODE, CONF_TIMEOUT
+from esphome.const import CONF_CHANNEL, CONF_DISABLED, CONF_ID, CONF_MODE, CONF_TIMEOUT
 
 from .. import (
     CONF_I2S_DOUT_PIN,
@@ -72,9 +72,10 @@ BASE_SCHEMA = (
     )
     .extend(
         {
-            cv.Optional(
-                CONF_TIMEOUT, default="500ms"
-            ): cv.positive_time_period_milliseconds,
+            cv.Optional(CONF_TIMEOUT, default="500ms"): cv.Any(
+                cv.positive_time_period_milliseconds,
+                cv.one_of(CONF_DISABLED, lower=True),
+            ),
         }
     )
     .extend(cv.COMPONENT_SCHEMA)
@@ -116,4 +117,5 @@ async def to_code(config):
     else:
         cg.add(var.set_dout_pin(config[CONF_I2S_DOUT_PIN]))
         cg.add(var.set_i2s_comm_fmt(config[CONF_I2S_COMM_FMT]))
-    cg.add(var.set_timeout(config[CONF_TIMEOUT]))
+    if config[CONF_TIMEOUT] != CONF_DISABLED:
+        cg.add(var.set_timeout(config[CONF_TIMEOUT]))
--- a/esphome/components/i2s_audio/speaker/__init__.py
+++ b/esphome/components/i2s_audio/speaker/__init__.py
@@ -2,7 +2,7 @@ from esphome import pins
 import esphome.codegen as cg
 from esphome.components import esp32, speaker
 import esphome.config_validation as cv
-from esphome.const import CONF_CHANNEL, CONF_DISABLED, CONF_ID, CONF_MODE, CONF_TIMEOUT
+from esphome.const import CONF_CHANNEL, CONF_ID, CONF_MODE, CONF_TIMEOUT
 
 from .. import (
     CONF_I2S_DOUT_PIN,
@@ -27,6 +27,7 @@ I2SAudioSpeaker = i2s_audio_ns.class_(
 CONF_BUFFER_DURATION = "buffer_duration"
 CONF_DAC_TYPE = "dac_type"
 CONF_I2S_COMM_FMT = "i2s_comm_fmt"
+CONF_NEVER = "never"
 
 i2s_dac_mode_t = cg.global_ns.enum("i2s_dac_mode_t")
 INTERNAL_DAC_OPTIONS = {
@@ -77,7 +78,7 @@ BASE_SCHEMA = (
             ): cv.positive_time_period_milliseconds,
             cv.Optional(CONF_TIMEOUT, default="500ms"): cv.Any(
                 cv.positive_time_period_milliseconds,
-                cv.one_of(CONF_DISABLED, lower=True),
+                cv.one_of(CONF_NEVER, lower=True),
             ),
         }
     )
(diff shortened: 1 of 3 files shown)


esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -281,12 +281,18 @@ void I2SAudioSpeaker::speaker_task(void *params) {
         stop_gracefully = true;
       }
 
+      i2s_event_t i2s_event;
+      while (xQueueReceive(this_speaker->i2s_event_queue_, &i2s_event, 0)) {
+        if (i2s_event.type == I2S_EVENT_TX_Q_OVF) {
+          tx_dma_underflow = true;
+        }
+      }
+
```
nielsnl68, line 292: Be careful with this, switching to newer versions of edi_idf will break this construction.
V5.0 changes this to an isr callback. 
Maybe adding `#if ESP_IDF_VERSION == ESP_IDF_VERSION_VAL(4, 4)` check around it.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -246,10 +245,12 @@ void I2SAudioSpeaker::speaker_task(void *params) {
...
 
-  const size_t dma_buffers_size = FRAMES_IN_ALL_DMA_BUFFERS * bytes_per_sample * number_of_channels;
+  const size_t dma_buffers_size = DMA_BUFFERS_COUNT * DMA_BUFFER_DURATION_MS * this_speaker->sample_rate_ / 1000 *
+                                  bytes_per_sample * number_of_channels;
+  const size_t ring_buffer_size =
+      RING_BUFFER_DURATION_MS * this_speaker->sample_rate_ / 1000 * bytes_per_sample * number_of_channels;
 
-  if (this_speaker->send_esp_err_to_event_group_(
-          this_speaker->allocate_buffers_(dma_buffers_size, RING_BUFFER_SAMPLES * bytes_per_sample))) {
+  if (this_speaker->send_esp_err_to_event_group_(this_speaker->allocate_buffers_(dma_buffers_size, ring_buffer_size))) {
```
nielsnl68, line 255: Just wondering, why having the `this_speaker->send_esp_err_to_event_group_()` called here instead of doing this for example in the `this_speaker->allocate_buffers_()`?

Calling it in `this_speaker->allocate_buffers_()` makes the code look much cleaner imho.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -199,23 +207,14 @@ size_t I2SAudioSpeaker::play(const uint8_t *data, size_t length, TickType_t tick
...
-
-    // Lock the ring buffer, write to it, then unlock it
-    xEventGroupClearBits(this->event_group_, SpeakerEventGroupBits::MESSAGE_RING_BUFFER_AVAILABLE_TO_WRITE);
-    size_t bytes_written = this->audio_ring_buffer_->write_without_replacement((void *) data, length, ticks_to_wait);
-    xEventGroupSetBits(this->event_group_, SpeakerEventGroupBits::MESSAGE_RING_BUFFER_AVAILABLE_TO_WRITE);
-
-    return bytes_written;
+  size_t bytes_written = 0;
+  if (this->audio_ring_buffer_.use_count() == 1) {
+    // Temporarily share ownership of the ring buffer so it won't be deallocated while writing
```
nielsnl68: This is a very good idea. Just on suggestion, could you add some comments on the meaning why `use_count()` needs to be **1**, and which values could it be otherwise?

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: What part is the ISR callback? What version esp-idf are you getting this issue? I just tested building and running it on 5.1.2 and 5.3.0 with no issues.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
kahrendt: In general, I wanted the ``allocate_buffers_``, ``start_i2s_driver_``, and ``reconfigure_i2s_stream_info_`` to be as informative as possible when returning an error. These functions are very general and do not rely on the event group. 

In the task, which depends on the event group, I only need to know whether they are successful or not, and if not, it should send the error to the event group. So I call that helper function inside the task to centralize this operation for all 3 functions.

nielsnl68: thanks for the explanation. this can be closed.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
nielsnl68: top. I did not got an error i just checked the documentation: https://docs.espressif.com/projects/esp-idf/en/release-v5.0/esp32/api-reference/peripherals/i2s.html#id9
and
https://docs.espressif.com/projects/esp-idf/en/release-v5.0/esp32/api-reference/peripherals/i2s.html#_CPPv435i2s_channel_register_event_callback17i2s_chan_handle_tPK21i2s_event_callbacks_tPv

kahrendt: I'm pretty sure those refer to the new I2S driver for IDF v5. We use the legacy I2S driver, so it's still using the same the driver code from v4.4.

nielsnl68: check that was what i expected when you said that it worked on 5.1.2 and 5.3.0

nielsnl68 (approved):

kahrendt:
I temporarily set this back as a draft, as I want to make one small change so that the ring buffer duration is configurable. Having it fixed at 500 ms matches the current behavior for only 16 kHz audio streams. Unfortunately, it will cause issues with the speaker media player component by making the pause command take half a second to actually go into effect.

I'll update the PR tomorrow with that option added.

esphome/components/i2s_audio/speaker/__init__.py
```diff
@@ -73,8 +73,12 @@ def validate_esp32_variant(config):
     .extend(
         {
             cv.Optional(
-                CONF_TIMEOUT, default="500ms"
+                CONF_BUFFER_DURATION, default="500ms"
             ): cv.positive_time_period_milliseconds,
+            cv.Optional(CONF_TIMEOUT, default="500ms"): cv.Any(
```
jesserockz, line 79: This should be consistent with things like `wifi`/`api` -> `reboot_timeout`  where `0s` means never, or `never` like can be specified for `update_interval`

jesserockz (changes requested):
Just one comment about the timeout value, but otherwise looks good to me.

jesserockz (approved):

PR #7752, opened 2024-11-11
Fix DHT11 not reading any values

minomy13:
Makes the DHT11 read correct values if used correctly with an external pull-up resistor as [described in documentation](https://esphome.io/components/sensor/dht.html).
I disabled the internal pull-up resistor. With double pull-ups there is no chance of reading any values.

Credits to Reddit user [Responsible_Hat_6056](https://www.reddit.com/user/Responsible_Hat_6056/) who shared the fix in a [post](https://www.reddit.com/r/raspberrypipico/comments/1133qqr/comment/j91qmtt/) some time ago.

## Types of changes

- [ ] Other

None.

- esphome/esphome-docs#4434

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: dht
    pin: GPIO1
    temperature:
    humidity:
      name: "Living Room Humidity"
    update_interval: 60s
    model: DHT11
    use_internal_pullup: false
```

## Checklist:


DjordjeMandic:
Why would this be a problem? Internal pullup is anyways 10k and up. Do you have any captures from oscilloscope showing falling and rising edges?

This looks more like software fix for hardware specific problem caused due to sensor that's faulty or operating outside of recommended specs.

I don't have any dht11/dht22 sensors free to test dht component with rp2040 but i have tested them before without external pullup and they worked on esp32.

Is this rp2040 specific?

DjordjeMandic (changes requested):
Add to docs that in yaml config you can disable internal pullup with pin schema. Don't make it hard coded, your proposed change is breaking one. Some people don't use external pullup. 

Then add bool to dht.h that will store value of that pullup enable state and add method to set pullup enable state. 

Then in code generation, get value of pullup optional parameter from pin schema supplied for dht pin and use it in code gen for that method above. 

Enable pullup based on bool from dht.h

minomy13:
So, no, it's not RP2040 related. The most breakout boards just already have a pull-up resistor and a capacitor installed. I tested it with the original code and it just didn't work.

But, you're absolutely right - it would be breaking for those people who are not using an external pull-up. I will definitely implement the config option!

esphome/const.py
```diff
@@ -411,6 +411,7 @@
 CONF_INTERNAL = "internal"
 CONF_INTERNAL_FILTER = "internal_filter"
 CONF_INTERNAL_FILTER_MODE = "internal_filter_mode"
+CONF_INTERNAL_PULLUP = "use_internal_pullup"
```
DjordjeMandic: I'd move this to dht/sensor.py, let those with write access decide

minomy13: Okay, moved it.

esphome/components/dht/sensor.py
```diff
@@ -30,6 +30,7 @@
     "DHT22_TYPE2": DHTModel.DHT_MODEL_DHT22_TYPE2,
 }
 DHT = dht_ns.class_("DHT", cg.PollingComponent)
+CONF_INTERNAL_PULLUP = "use_internal_pullup"
```
DjordjeMandic:
```suggestion
CONF_USE_INTERNAL_PULLUP = "use_internal_pullup"
```

```suggestion
CONF_INTERNAL_PULLUP = "internal_pullup"
```

esphome/const.py
minomy13: @OttoWinter Could you please take a look if I should leave the constant where it is now or if it should be in const.py?

jesserockz:
Considering the `pin` schemas already have flags/mode, adding a `get_flags` to the generic `InternalGPIOPin` classes will allow component code to get the desired flags from the user and have defaults

esphome/components/dht/dht.cpp
```diff
@@ -24,6 +24,12 @@ void DHT::dump_config() {
     ESP_LOGCONFIG(TAG, "  Model: DHT22 (or equivalent)");
   }
 
+  if (this->use_internal_pullup_) {
+    ESP_LOGCONFIG(TAG, "  Internal pull-up resistor enabled.");
+  } else {
+    ESP_LOGCONFIG(TAG, "  Internal pull-up resistor disabled.");
+  }
```
DjordjeMandic, line 31:
```suggestion
    ESP_LOGCONFIG(TAG, "  Internal pull-up resistor: %s", ONOFF(this->pin_->get_flags() & gpio::Flags::FLAG_PULLUP));
```

Use one of these macros:
https://github.com/esphome/esphome/blob/f7f8bf4da44b83796a301bc486d77fdb46d8ed82/esphome/core/log.h#L161-L163

DjordjeMandic (changes requested):

kroimon:
Couldn't the `CONF_PIN` / `pin:` configuration just be changed to `internal_gpio_input_pullup_pin_schema` so that the pull-up is enabled by default like it was before, but the user could override it with `pullup: False`?

DjordjeMandic:
@kroimon same gpio is used for output and it's switched between input and output during runtime. Schema does set initial state, but during runtime you can't know what was the initial state. That's where get_flags become useful.

kroimon:
Oh I see! Sorry, didn't look too much into the component and wasn't aware of the direction switching.

esphome/components/dht/sensor.py
DjordjeMandic, line 1: Discard changes for this whole file

esphome/components/dht/dht.h
DjordjeMandic, line 1: Discard changes for this whole file

esphome/components/dht/dht.cpp
```diff
@@ -101,7 +107,12 @@ bool HOT IRAM_ATTR DHT::read_sensor_(float *temperature, float *humidity, bool r
   } else {
     delayMicroseconds(800);
   }
-  this->pin_->pin_mode(gpio::FLAG_INPUT | gpio::FLAG_PULLUP);
+
+  if (this->use_internal_pullup_) {
+    this->pin_->pin_mode(gpio::FLAG_INPUT | gpio::FLAG_PULLUP);
+  } else {
+    this->pin_->pin_mode(gpio::FLAG_INPUT);
+  }
```
DjordjeMandic, line 115:
```suggestion
  this->pin_->pin_mode(gpio::FLAG_INPUT | (this->pin_->get_state() & gpio::Flags::FLAG_PULLUP ? gpio::Flags::FLAG_PULLUP : 0));
```

DjordjeMandic (changes requested):
@minomy13  https://github.com/esphome/esphome/pull/8151 is merged into dev, update your branch and use get_flags()

DjordjeMandic:
Closed in favour of #8257

PR #7755, opened 2024-11-12
[opentherm] Message to string extend

FreeBear-nc:

Extends printed message ID strings to include latest additions defined in opentherm.h and use them when printing a list of repeating & non-repeating messages during config dump. Saves having to cross reference the ID numbers in the source code.


olegtarasov (approved):
These codes were indeed missed in `message_id_to_str`.

jesserockz (approved):

PR #7767, opened 2024-11-14
online_image: allow setting templatable headers on http request

afflux:

The HTTP requests by the `online_image` component did not give the user a config option to set the HTTP headers. This PR implements setting headers and supports templates.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4441

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
online_image:
  - id: radar0
    url: http://.../rain/radar0.png
    on_error:
      - online_image.release: radar0
    format: PNG
    type: RGBA
    placeholder: bad_radar

    headers:
      If-Modified-Since: !lambda |-
        char age_buf[64];
        auto now = id(rtc)->timestamp_now();
        auto time = ESPTime::from_epoch_utc(now - 10 * 60);
        auto age_len = time.strftime(age_buf, sizeof(age_buf), "%a, %d %b %Y %H:%M:%S %z");
        if (age_len != 0) {
          return age_buf;
        } else {
          return {};
        }


```

## Checklist:

### DIFF
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -104,10 +104,12 @@ void OnlineImage::update() {
   }
 
   std::list<http_request::Header> headers;
-  for (const auto &[key, value_template] : this->headers_) {
-    auto val = value_template();
-    if (val.has_value()) {
-      headers.push_back(http_request::Header{key, *val});
+  for (const auto &item : this->headers_) {
+    auto header_name = item.first;
+    auto header_value_template = item.second;
+    auto header_value = header_value_template();
+    if (header_value.has_value()) {
+      headers.push_back(http_request::Header{header_name, *header_value});
     }
   }
 
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -103,12 +103,11 @@ void OnlineImage::update() {
     ESP_LOGI(TAG, "Updating image");
   }
 
-
   std::list<http_request::Header> headers;
-  for (const auto &item : this->headers_) {
-    auto val = item.second();
+  for (const auto &[key, value_template] : this->headers_) {
+    auto val = value_template();
     if (val.has_value()) {
-      headers.push_back(http_request::Header{item.first, *val});
+      headers.push_back(http_request::Header{key, *val});
     }
   }
 
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -103,7 +103,16 @@ void OnlineImage::update() {
     ESP_LOGI(TAG, "Updating image");
   }
 
-  this->downloader_ = this->parent_->get(this->url_);
+
+  std::list<http_request::Header> headers;
+  for (const auto &item : this->headers_) {
+    auto val = item.second();
+    if (val.has_value()) {
+      headers.push_back(http_request::Header{item.first, *val});
+    }
+  }
+
+  this->downloader_ = this->parent_->get(this->url_, headers);
 
   if (this->downloader_ == nullptr) {
     ESP_LOGE(TAG, "Download failed.");
--- a/esphome/components/online_image/online_image.h
+++ b/esphome/components/online_image/online_image.h
@@ -62,6 +62,8 @@ class OnlineImage : public PollingComponent,
     }
   }
 
+  void add_header(const char *key, std::function<optional<const char*>()> &&f) { this->headers_.insert({key, f}); }
+
   /**
    * @brief Set the image that needs to be shown as long as the downloaded image
    *  is not available.
@@ -122,6 +124,7 @@ class OnlineImage : public PollingComponent,
   image::Image *placeholder_{nullptr};
 
   std::string url_{""};
+  std::map<const char *, std::function<optional<const char*>()>> headers_{};
 
   /** width requested on configuration, or 0 if non specified. */
   const int fixed_width_;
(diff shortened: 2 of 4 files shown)


esphome/components/online_image/online_image.cpp
```diff
@@ -103,7 +103,16 @@ void OnlineImage::update() {
     ESP_LOGI(TAG, "Updating image");
   }
 
-  this->downloader_ = this->parent_->get(this->url_);
+
+  std::list<http_request::Header> headers;
+  for (const auto &item : this->headers_) {
```
guillempages, line 107: Maybe use structured bindings? Should be available from C++17
```suggestion
  for (const auto &[key, value] : this->headers_) {
```
If this works, it would make the rest of the loop slightly better readable (without the ".first()" or ".second()" calls)

afflux:
```
src/esphome/components/online_image/online_image.cpp: In member function 'virtual void esphome::online_image::OnlineImage::update()':
src/esphome/components/online_image/online_image.cpp:107:20: warning: structured bindings only available with -std=c++17 or -std=gnu++17
   for (const auto &[key, value_template] : this->headers_) {
```
Do I set this somewhere or should I revert it?

afflux: CI doesn't like C++17 so I guess I go back

clydebarrow: C++11 is the base language level available on all ESPHome supported platforms.

tests/components/online_image/common.yaml
```diff
@@ -24,10 +24,26 @@ online_image:
...
+    format: PNG
+    type: RGBA
+    http_request_headers:
+      If-Modified-Since: !lambda |-
+        char age_buf[64];
+        auto now = id(sntp_time)->timestamp_now();
+        auto time = ESPTime::from_epoch_utc(now - 10 * 60);
+        auto age_len = time.strftime(age_buf, sizeof(age_buf), "%a, %d %b %Y %H:%M:%S %z");
+        if (age_len != 0) {
+          return age_buf;
```
clydebarrow, line 38: You can't return the address of a local variable from a lambda. Can I suggest the template use `std::string` instead of `const char *`? `ESPTime` has a `strftime()` method that yields `std::string`

esphome/components/online_image/online_image.h
```diff
@@ -122,6 +124,7 @@ class OnlineImage : public PollingComponent,
   image::Image *placeholder_{nullptr};
 
   std::string url_{""};
+  std::map<const char *, std::function<optional<const char *>()>> headers_{};
```
clydebarrow, line 127:
```suggestion
  std::map<const char *, TemplatableValue<std::string>> headers_{};
```
`TemplatableValue` is a utility class in ESPHome for just this application.

tests/components/online_image/common.yaml
afflux: oh man, rookie mistake. I guess it would work with a static const char * to avoid allocation, but I guess that's not really intuitive.

afflux: Actually, that raises another issue though. If I call c_str() here: https://github.com/esphome/esphome/blob/7365f5e219f96f9c4ef555900b082cedc1d52727/esphome/components/online_image/online_image.cpp#L112, the std::string must not be destroyed before calling `HttpRequestComponent::get()`. So do I need to keep the std::strings around in another std::list?

esphome/components/online_image/online_image.cpp
guillempages: Oh sorry. Somehow I thought I had read some discussions, that we were on C++17, and thinking about going to C++20, but looks like I was waaaay off :-(
Might have mistaken it with a different project.

tests/components/online_image/common.yaml
clydebarrow: Change your headers map as suggested above:

```
std::map<const char *, TemplatableValue<std::string>> headers_{}
```

The key (header name) is ok as a `const char *` since it is only set once. The value is mutable, and you save a lot of hassle by using `std::string` as well as maintaining consistency with the rest of ESPHome, where `std::string` is typically used when returning text from lambdas. Yes, there might be some allocation going on, but not necessarily - `std::string` stores short strings in itself. But it takes away any concerns about lifetime of data.

You would then call `c_str()` when setting the header - that's ok because the source string won't change until after the request. You'll also want to use `std::move` in `add_header()`.

clydebarrow:
Just looking at this again - would it not be better to make this facility part of http_request rather than specific to online_image?

tests/components/online_image/common.yaml
clydebarrow: Should have looked more closely - I see now you are leveraging existing stuff in http_request, but it is expecting a `const char *`.

Ideally the http_request component would be updated to provide a base schema for other components to extend that want to make http calls with headers etc. And refactored to use `std::string` to avoid these issues with data lifetime.

afflux: Will look into generalizing this. Base schema would be one option - another would be defining each request (URLs, headers, callbacks, timeouts, ...) below `http_request` and referencing them from http_request.get/post actions, online_images and whatever else needs them. Would break existing configs, though.
(The naming would make a little more sense, IMHO, as you'd actually define http requests in the `http_request:` yaml section)

afflux: @clydebarrow @guillempages thoughts on how it should look? Common schema for http.get Actions and online_image that both extend from (compatible with current configs, but I could see naming conflicts on potential future http use cases), or move the http request definitions into the http_request component and reference them from the use cases?

PR #7813, opened 2024-11-24
Add support for MCP3201 12-bit 1-channel A/D converter

zagor:

Add support for Microcip MCP3201 1-channel 12-bit A/D converter

This ADC is similar to the already supported MCP3204/8 chip. But since it only exposes a single channel it uses a slightly different wire protocol.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
mcp3201:
  cs_pin: GPIO15
  reference_voltage: 4.096V
```

## Checklist:

  https://github.com/esphome/esphome-docs/pull/4464

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -265,6 +265,7 @@ esphome/components/mcp23x08_base/* @jesserockz
 esphome/components/mcp23x17_base/* @jesserockz
 esphome/components/mcp23xxx_base/* @jesserockz
 esphome/components/mcp2515/* @danielschramm @mvturnho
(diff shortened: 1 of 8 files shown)


jesserockz (changes requested):
Since there is only a single channel, I would combine all of the parent and sensor code into just a sensor. There is only a need for a parent component if there can be multiple sensors attached to it.

PR #7836, opened 2024-11-25
[cse7761] Add Sonoff POW-CT support and frequency sensor

mick96:

- This PR implements the possibility to select the specific Sonoff device model to configure the cse7761 accordingly. This fixes issues on Sonoff POW Ring (POWCT).
Backward compatibility should be guaranteed by using default values on the new optional parameters.
- This PR implement optional frequency sensor available on cse7761.
- Improve code comments with cse7761 register details.

All the work is based on Tasmota implementation: https://github.com/arendst/Tasmota/blob/development/tasmota/tasmota_xnrg_energy/xnrg_19_cse7761.ino

**All credit to @arendst and Tasmota community.**

## Types of changes

- [ ] Other

- fixes https://community.home-assistant.io/t/sonoff-powct-pow-ring-power-not-accurate/789686

- esphome/esphome-docs#4469

**Pull request in [esphome-devices](https://github.com/esphome/esphome-devices) with documentation (if applicable):**

- esphome/esphome-devices#1007

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
 sensor: 
  - platform: cse7761
    model: 'SONOFF POWCT'
    update_interval: 2s
    voltage:
      name: Voltage
    frequency:
      name: Frequency
    current_1:
      name: Current
    active_power_1:
      name: Power

```

## Checklist:


mick96:
When and if this PR will be merged the template for Sonoff POW Ring must be updated as well.
- esphome/esphome-devices#1007

tinusb:
Tested this PR and can confirm that it fixes the problem with the cse7761 platform and the new Sonoff POW-CT.  Without the fix the Sonoff POW-CT only provides accurate readings up to around 2000W.  Above 2000W cse7761 sensor output reduces to about 1/3 of the actual usage.  After I added the PR to my yaml script the cse7761 sensor output value was correct and I verified the readings with an amp clamp meter.  The sensor also outputs the mains frequency which is a handy feature.

lightingman117:
Would love to see this merged when possible, thanks Devs!

333robbie333:
I'd also like to see this merged. Thanks Devs!

h1771t:
This PR fixes all my POWCT issues. Applied the patches, added to my config 
`sonoff_model: POWCT`
Now the sensor readings to HA looks good below and above 2kw load, and over 24 hours the HA daily usage matches my utility usage reading very closely.

I removed 2 of the below config as well (for current_1 and for active_power_1), initially take from https://devices.esphome.io/devices/SONOFF-POWCT

> filters:
        # Measurement divided by the PI number
        - lambda: return x / PI;

So I think a patch needs to also go with this PR to remove the filters from the config documentation at https://devices.esphome.io/devices/SONOFF-POWCT

h1771t:
Would also love for this to get merged, cc @jesserockz

mick96:
> This PR fixes all my POWCT issues. Applied the patches, added to my config `sonoff_model: POWCT` Now the sensor readings to HA looks good below and above 2kw load, and over 24 hours the HA daily usage matches my utility usage reading very closely.
> 
> I removed 2 of the below config as well (for current_1 and for active_power_1), initially take from https://devices.esphome.io/devices/SONOFF-POWCT
> 
> > filters:
> > # Measurement divided by the PI number
> > - lambda: return x / PI;
> 
> So I think a patch needs to also go with this PR to remove the filters from the config documentation at https://devices.esphome.io/devices/SONOFF-POWCT

Right, no more PI division needed.  I've also submitted a PR to esphome-devices to update the sample config.  The filter was just a workaround for the incorrect CSE7761 register config, which this PR now fixes.

h1771t:
Is this waiting on @berfenger @jesserockz ? The PR fixes work great for me since putting it in, would be nice to have it committed. Thanks.

h1771t:
I would like to upgrade esphome on my devices, but this one is holding me from doing so, it will break one of my main devices I depend on. Any word on merging @jesserockz ? Thanks

lightingman117:
> I would like to upgrade esphome on my devices, but this one is holding me from doing so, it will break one of my main devices I depend on. Any word on merging @jesserockz ? Thanks

Just do this, don't need to wait.

```
external_components:
  - source:
      type: git
      url: https://github.com/mick96/esphome.git
      ref: cse7761-sonoffpow-fix
    components: [ cse7761 ]
```

TimoPtr:
@jesserockz I tested the implementation and it works on my end on my device.

esphome/components/cse7761/cse7761.cpp
```diff
@@ -216,6 +392,10 @@ void CSE7761Component::get_data_() {
     this->voltage_sensor_->publish_state(voltage);
   }
 
+  float freq = (this->data_.frequency) ? ((float) CSE7761_FREF / 8 / this->data_.frequency) : 0;  // Hz
+  if (this->frequency_sensor_ != nullptr) {
```
jesserockz: No need to do maths if not going to be used
```suggestion
  if (this->frequency_sensor_ != nullptr) {
    float freq = (this->data_.frequency) ? ((float) CSE7761_FREF / 8 / this->data_.frequency) : 0;  // Hz
```

jesserockz (dismissed):

jebentancour:
Hi,
Just wanted to confirm I'm using @mick96's external component with my Sonoff POWCT and it works perfectly. Power readings are accurate across the full range.
Having this integrated into official ESPHome would be very helpful to avoid depending on external components. Thanks for the work on this PR!

ParanoiaCY:
I will like also to confirm accurate reading after manually implementing the fix to 2 separate devices in 2 different installation sites. Please add it to the official branch. Also [https://devices.esphome.io/devices/sonoff-powct/](url) will need to be updated as well.

esphome/components/cse7761/cse7761.cpp
```diff
@@ -12,30 +12,52 @@ static const char *const TAG = "cse7761";
...
-
-static const uint8_t CSE7761_SPECIAL_COMMAND = 0xEA;   // Start special command
-static const uint8_t CSE7761_CMD_RESET = 0x96;         // Reset command, after receiving the command, the chip resets
+// static const uint8_t CSE7761_REG_RMSIBC         = 0x71;     // (2) Channel B effective current conversion coefficient
+// static const uint8_t CSE7761_REG_RMSUC          = 0x72;     // (2) Effective voltage conversion coefficient
+// static const uint8_t CSE7761_REG_POWERPAC       = 0x73;     // (2) Channel A active power conversion coefficient
+// static const uint8_t CSE7761_REG_POWERPBC       = 0x74;     // (2) Channel B active power conversion coefficient
+// static const uint8_t CSE7761_REG_POWERSC        = 0x75;     // (2) Apparent power conversion coefficient
+// static const uint8_t CSE7761_REG_ENERGYAC       = 0x76;     // (2) Channel A energy conversion coefficient
+// static const uint8_t CSE7761_REG_ENERGYBC       = 0x77;     // (2) Channel B energy conversion coefficient
```
TimoPtr: Why all of these const are commented?

esphome/components/cse7761/cse7761.cpp
```diff
@@ -149,13 +171,17 @@ uint32_t CSE7761Component::read_(uint8_t reg, uint8_t size) {
 }
 
 uint32_t CSE7761Component::coefficient_by_unit_(uint32_t unit) {
+  uint32_t coeff = 1;
+  if (this->data_.model == CSE7761_MODEL_SONOFF_POWCT) {
+    coeff = 5;
```
TimoPtr, line 164: Why 5? It would be nice to have some documentation for it.

esphome/components/cse7761/cse7761.cpp
```diff
@@ -179,10 +205,157 @@ bool CSE7761Component::chip_init_() {
...
+      1001   Voltage channel zero-crossing signal output (Tasmota add zero-cross detection)
+      1010   Current channel A zero-crossing signal output
+      1011   Current channel B zero crossing signal output
+      1100   Voltage channel overvoltage indication signal output
+      1101   Voltage channel undervoltage indication signal output
+      1110   Current channel A overcurrent signal indication output
+      1111   Current channel B overcurrent signal indication output
+    */
+
+    // this->write_(CSE7761_REG_PULSE1SEL | 0x80, 0x3290);  // Enable zero crosing signal output on function pin
```
TimoPtr: dead code?

esphome/components/cse7761/cse7761.cpp
mick96: These register definitions (0x71–0x77) are kept as comments for documentation/reference purposes. They're not used directly in code because chip_init_() reads all 8 coefficients in a loop using CSE7761_REG_RMSIAC + i (register 0x70 through 0x77). Only CSE7761_REG_RMSIAC (0x70) is needed as the base address.

swoboda1337: Should remove them or uncomment them.

esphome/components/cse7761/cse7761.cpp
```diff
@@ -12,30 +12,52 @@ static const char *const TAG = "cse7761";
...
+//                                                             //   The channel indicated by the signal of power
+//                                                             overload is channel A
+// static const uint8_t CSE7761_CMD_CHAN_B_SELECT  = 0xA5;     // Current channel B setting command, which specifies the
+// current used to calculate apparent power,
+//                                                             //   Power factor, phase angle, instantaneous active
+//                                                             power, instantaneous apparent power and
+//                                                             //   The channel indicated by the signal of power
+//                                                             overload is channel B
 static const uint8_t CSE7761_CMD_CLOSE_WRITE = 0xDC;   // Close write operation
 static const uint8_t CSE7761_CMD_ENABLE_WRITE = 0xE5;  // Enable write operation
```
swoboda1337: Same here, remove commented out register/command definitions

esphome/components/cse7761/cse7761.cpp
```diff
@@ -179,10 +205,157 @@ bool CSE7761Component::chip_init_() {
...
+      configured first =1, turn on the overvoltage, overcurrent, and overload detection functions =0, turn off the
+      overvoltage, overcurrent, and overload detection functions (Sonoff Dual R3 / Pow CT) 2      ZxEN Zero-crossing
+      detection, phase angle, voltage frequency measurement enable signal =1, turn on the zero-crossing detection, phase
+      angle, and voltage frequency measurement functions (if frequency enable) =0, disable zero-crossing detection,
+      phase angle, voltage frequency measurement functions (Sonoff Dual R3 / Pow CT) 1      PeakEN             Peak
+      detect enable signal =1, turn on the peak detection function =0, turn off the peak detection function (Sonoff Dual
+      R3 / Pow CT) 0      NC                 Default is 1
+    */
+
+    this->write_(CSE7761_REG_EMUCON2 | 0x80, 0x0FE5);  // Sonoff Dual R3 / Pow CT + frequency measure enable
```
swoboda1337: This changed from `0x0FC1` to `0x0FE5` which enables zero-crossing (bit 2) and waveform output (bit 5) for all models, not just POWCT. Could this have side effects for existing Dual R3 users?

esphome/components/cse7761/cse7761.cpp
```diff
@@ -179,10 +205,157 @@ bool CSE7761Component::chip_init_() {
...
+      0111   Instantaneous value update interrupt output
+      1000   Average update interrupt output
+      1001   Voltage channel zero-crossing signal output (Tasmota add zero-cross detection)
+      1010   Current channel A zero-crossing signal output
+      1011   Current channel B zero crossing signal output
+      1100   Voltage channel overvoltage indication signal output
+      1101   Voltage channel undervoltage indication signal output
+      1110   Current channel A overcurrent signal indication output
+      1111   Current channel B overcurrent signal indication output
+    */
```
swoboda1337: The PULSE1SEL write was removed — old code wrote `0x3290`, now its commented out. This changes default behavior for existing Dual R3 users. Is this intentional?

esphome/components/cse7761/cse7761.cpp
```diff
@@ -179,10 +205,157 @@ bool CSE7761Component::chip_init_() {
...
+                                =0, enable U channel digital high pass filter (Sonoff Dual R3 / Pow CT)
+                                =1, turn off the U channel digital high-pass filter
+      3-2    NC                 -
+      1      PBRUN
+                                =1, enable PFB pulse output and active energy register accumulation; (Sonoff Dual R3 /
+      Pow CT) =0 (default), turn off PFB pulse output and active energy register accumulation. 0      PARUN =1, enable
+      PFA pulse output and active energy register accumulation; (Sonoff Dual R3 / Pow CT) =0 (default), turn off PFA
+      pulse output and active energy register accumulation.
+    */
+
```
swoboda1337: These register bit-field descriptions are copy-pasted datasheet content. A brief comment explaining the value chosen (e.g. `// PGA=1 for all channels, channel B enabled`) with a datasheet reference would be much cleaner than 100+ lines of register docs

esphome/components/cse7761/cse7761.cpp
```diff
@@ -216,6 +392,10 @@ void CSE7761Component::get_data_() {
     this->voltage_sensor_->publish_state(voltage);
   }
 
+  if (this->frequency_sensor_ != nullptr) {
+    float freq = (this->data_.frequency) ? ((float) CSE7761_FREF / 8 / this->data_.frequency) : 0;  // Hz
+    this->frequency_sensor_->publish_state(freq);
```
swoboda1337, line 260: nit: missing blank line before the for loop

swoboda1337 (changes requested):

swoboda1337:
There are also merge conflicts

esphome/components/cse7761/cse7761.cpp
mick96: The Dual R3 uses a shunt resistor with PGA=16×, while the POW CT uses a current transformer with PGA=1× (higher gain would clip the ADC). The chip's calibration coefficients are factory-burned by the supplier and identical across both products. The factor of 5 compensates for the net analog front-end gain difference. 
This value is taken from the Tasmota implementation and confirmed working.
I'll add a comment.

esphome/components/cse7761/cse7761.cpp
mick96: This is intentionally commented out. It would configure the PULSE1 pin to output the voltage zero-crossing signal (P1Sel=0b1001 = 0x9, visible in the register table above it). It's left as a reference for anyone who wants to enable zero-crossing detection on the function pin.

arendst:
Oh my. Users complaining over comments. The world just came to an end.

esphome/components/cse7761/cse7761.cpp
mick96: I'd prefer to keep them as comments for reference. These are register and command definitions from the CSE7761 datasheet — they document chip capabilities that aren't actively used in the current implementation but are useful context for anyone working on this driver (e.g., adding apparent power, energy accumulation, or channel B selection).

esphome/components/cse7761/cse7761.cpp
mick96: These are required for the frequency sensor to work — the UFREQ register (0x23) only produces valid data when zero-crossing detection is enabled, and ZxEN requires WaveEN as a prerequisite per the datasheet.

Regarding side effects for Dual R3 users: enabling these bits only activates internal measurement modules inside the chip. They don't change how voltage, current, or power are calculated — those paths are independent. If a Dual R3 user doesn't configure a frequency sensor, the value is read but never published, so there's no functional change for existing configurations.

esphome/components/cse7761/cse7761.cpp
swoboda1337: Then uncomment them. It wont get merged like this.

esphome/components/cse7761/sensor.py
```diff
@@ -60,6 +70,15 @@
...
             ),
+            cv.Optional(CONF_FREQUENCY): sensor.sensor_schema(
+                unit_of_measurement=UNIT_HERTZ,
+                accuracy_decimals=1,
+                device_class=DEVICE_CLASS_FREQUENCY,
+                state_class=STATE_CLASS_MEASUREMENT,
+            ),
+            cv.Optional(CONF_MODEL, default="SONOFF DUALR3"): cv.enum(
+                MODELS, upper=True
+            ),
```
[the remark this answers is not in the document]
mick96: PR description updated

esphome/components/cse7761/cse7761.cpp
mick96: Intentional. The 0x3290 value only changed what signal appears on the physical Pulse2 pin (zero-crossing instead of default PFB pulse). ESPHome doesn't monitor that pin, so it has no functional impact. Default 0x3210 is fine. Kept as a comment for reference.

Could you clarify the project's guidelines on commented-out code?

esphome/components/cse7761/cse7761.cpp
mick96: I'll add it

esphome/components/cse7761/cse7761.cpp
mick96: If we uncomment them we'll get unused symbol warnings in CI, since they're not referenced in the current implementation.

swoboda1337: Make them constexpr instead dont think CI will complain.

esphome/components/cse7761/cse7761.cpp
mick96: Ok, but consider that the only datasheet I could find is in Chinese.

mick96:
I am applying the suggested fixes.
@TimoPtr  what if we also include your fix for bidirectional power reading in this PR?

TimoPtr:
> I am applying the suggested fixes.
> @TimoPtr  what if we also include your fix for bidirectional power reading in this PR?

I think we can keep both separated, your PR already contains a lot of changes.

esphome/components/cse7761/cse7761.cpp
```diff
@@ -179,10 +192,30 @@ bool CSE7761Component::chip_init_() {
...
+    //           ZXD0=0 (positive), HPF enabled for all channels, PBRUN=1, PARUN=1
     this->write_(CSE7761_REG_EMUCON | 0x80, 0x1183);
-    this->write_(CSE7761_REG_EMUCON2 | 0x80, 0x0FC1);
-    this->write_(CSE7761_REG_PULSE1SEL | 0x80, 0x3290);
+
+    // EMUCON2 (0x13): See cse7761_registers.md for full bit-field reference.
+    // 0x0FE5 -- Energy not cleared after read (UART mode), DUPSEL=27.3Hz, CHS_IB=current,
+    //           PfactorEN=1, WaveEN=1, ZxEN=1 (required for frequency measurement)
+    this->write_(CSE7761_REG_EMUCON2 | 0x80, 0x0FE5);
+
```
bdraco: Please make the magic numbers constexpr

esphome/components/cse7761/cse7761.cpp
```diff
@@ -179,10 +192,30 @@ bool CSE7761Component::chip_init_() {
...
   if (sys_status & 0x10) {  // Write enable to protected registers (WREN)
-    this->write_(CSE7761_REG_SYSCON | 0x80, 0xFF04);
+
+    // SYSCON (0x00): See cse7761_registers.md for full bit-field reference.
+    // Dual R3:  0xFF04 -- ADC2ON=1, PGAIB=16x, PGAU=1x, PGAIA=16x (shunt resistor, needs amplification)
+    // POW CT:   0xFE00 -- ADC2ON=1, PGAIB=1x,  PGAU=1x, PGAIA=1x  (current transformer, no amplification)
+    if (this->data_.model == CSE7761_MODEL_SONOFF_POWCT) {
+      this->write_(CSE7761_REG_SYSCON | 0x80, 0xFE00);
+    } else {
+      this->write_(CSE7761_REG_SYSCON | 0x80, 0xFF04);
```
bdraco: I'd probably put a ternary in the 2nd arg to reduce the duplicate code

esphome/components/cse7761/cse7761.cpp
```diff
@@ -179,10 +192,30 @@ bool CSE7761Component::chip_init_() {
...
-    this->write_(CSE7761_REG_PULSE1SEL | 0x80, 0x3290);
+
+    // EMUCON2 (0x13): See cse7761_registers.md for full bit-field reference.
+    // 0x0FE5 -- Energy not cleared after read (UART mode), DUPSEL=27.3Hz, CHS_IB=current,
+    //           PfactorEN=1, WaveEN=1, ZxEN=1 (required for frequency measurement)
+    this->write_(CSE7761_REG_EMUCON2 | 0x80, 0x0FE5);
+
+    // PULSE1SEL (0x1D): Left at default 0x3210. Not written.
+    // Uncomment below to output voltage zero-crossing signal on Pulse2 pin (P2Sel=0x9):
+    // this->write_(CSE7761_REG_PULSE1SEL | 0x80, 0x3290);
```
bdraco: this is an undocumented behavior change

esphbot: Agreed. Affects all Dual R3 users. Gate EMUCON2 `0x0FE5` on `frequency_sensor_ != nullptr`. Otherwise keep `0x0FC1`. Keep `PULSE1SEL` write for Dual R3, or document the change.

esphome/components/cse7761/cse7761.cpp
esphbot: Still open at lines 199-213. Name SYSCON/EMUCON/EMUCON2 values as `static constexpr`. Add a POW CT gain constant too.

esphome/components/cse7761/cse7761.cpp
esphbot: Remove commented-out code. Maintainers won't merge it. Document zero-cross option in the PR, not source. Or make it a config option.

esphome/components/cse7761/cse7761.cpp
esphbot: Enabling ZxEN/WaveEN may be harmless. Still untested on Dual R3. Enable only with frequency sensor configured. Avoids the question entirely.

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Chip init changes for existing Dual R3 users are not documented

esphome/components/cse7761/sensor.py
```diff
@@ -27,6 +31,12 @@
 CONF_ACTIVE_POWER_1 = "active_power_1"
 CONF_ACTIVE_POWER_2 = "active_power_2"
 
+CSE7761Model = cse7761_ns.enum("CSE7761Model")
+MODELS = {
+    "SONOFF DUALR3": CSE7761Model.CSE7761_MODEL_SONOFF_DUALR3,
+    "SONOFF POWCT": CSE7761Model.CSE7761_MODEL_SONOFF_POWCT,
+}
```
bdraco: These chips exist in more than just Sonoff products.  The coupling of Sonoff models seems like its the wrong abstraction

bdraco (changes requested):
Please see comments above

PR #7892, opened 2024-11-29
[http_request] don't block loop with actions

### LINKED ISSUE #6493: watchdog non functional during stuck http request on rp2040
### The problem

I recently had a lockup after one or two weeks of uptime on my rp2040 based project. I'd normally expect it to reboot after 8.3s [due to the watchdog](https://github.com/esphome/esphome/blob/a3ef2ed7fd109fc6885de3dee7245d637d5ae2a9/esphome/components/rp2040/core.cpp#L22), but it didn't happen, stayed locked for roughly 10 minutes, when I noticed the problem and did a manual reset (can't really attach a debugger at the install location)

Last log message received is shown below, indicating a http_request. At precisely this time, state updates stop arriving on my home assistant instance.

Since the watchdog never triggered, I assume it was fed somewhere, so I look at all usages of `feed_wdt()` in `.esphome/build/project/src/`, and I find four (outside of `Application::setup()` and `Application::loop()`):
1. [i2c bus recovery](https://github.com/esphome/esphome/blob/a3ef2ed7fd109fc6885de3dee7245d637d5ae2a9/esphome/components/i2c/i2c_bus_arduino.cpp#L272) (only called during `setup()`)
2. various places in [`ESPHomeOTAComponent::handle_()`](https://github.com/esphome/esphome/blob/a3ef2ed7fd109fc6885de3dee7245d637d5ae2a9/esphome/components/esphome/ota/ota_esphome.cpp#L87)
3. display_buffer (yes, used a lot, but not in contexts that could loop forever)
4. [http_request action](https://github.com/esphome/esphome/blob/a3ef2ed7fd109fc6885de3dee7245d637d5ae2a9/esphome/components/http_request/http_request.h#L206-L208) and the [corresponding container implementation](https://github.com/esphome/esphome/blob/a3ef2ed7fd109fc6885de3dee7245d637d5ae2a9/esphome/components/http_request/http_request_arduino.cpp#L150)

Just from looking at the code, the http_request handling seems inconsistent:
1. [this `feed_wdt()`](https://github.com/esphome/esphome/blob/a3ef2ed7fd109fc6885de3dee7245d637d5ae2a9/esphome/components/http_request/http_request.h#L207) is unconditional, and the loop does not care if `container->read` returned 0 or even less than zero. If `container->read()` ends up returning -1, you'll get an endless loop. There is a second call to `feed_wdt()` in the `container->read()` implementation, but it will also not care if the http client returned 0 bytes, still causing an endless loop if the network had a hiccup.
2. Both the container and the action are tracking how many bytes were read, and the action is using both (`container->get_bytes_read()` and `read_index`). Due to 1., it can diverge - shouldn't this be unified?
3. Unlike [documented](https://github.com/esphome/esphome-docs/blob/5d624ed617a1e6a72052b884be12572d31c9e524/components/http_request.rst?plain=1#L25), the timeout set on the http_request actually sets the TCP connection timeout, and there is no timeout handling for the total http request (source for [rp2040](https://github.com/earlephilhower/arduino-pico/blob/e16c459598b86ce67d8b83ac5b1384dcb8c982ae/libraries/HTTPClient/src/HTTPClient.cpp#L448-L451) and [esp8266](https://github.com/esp8266/Arduino/blob/bb79e9076e8402c81c24efbabd19f1baab1c6bd1/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L305-L308))
4. [The call to `yield()` is not actually doing anything on rp2040](https://github.com/earlephilhower/arduino-pico/blob/e16c459598b86ce67d8b83ac5b1384dcb8c982ae/cores/rp2040/delay.cpp#L49), presumably because FreeRTOS is not linked in

Pretty sure that the combination of points 1, 3 and 4 caused lockup.

I'm happy to open a pull request to clean the http_request handling code, but I'm not familiar at all with FreeRTOS and platform.io, so I'm not sure what needs to happen to get `yield()` working on `rp2040` - but maybe that should be a second issue.

e819185de18f2f1d4f9b3e2c96190d1cefe53422

pip

_No response_

RP2040

### Board

Pico W

http_request

### Example YAML snippet

```yaml
script:
  - id: update_foo
    mode: single
    then:
      - http_request.post:
          url: http://foo/bar
          capture_response: true
          max_response_buffer_size: 6kB
time:
  - platform: sntp
    id: sntp_time
    timezone: Etc/UTC
    servers:
      - 192.168.1.1

    on_time:
      - seconds: 0
        minutes: /15
        then:
          - script.execute: update_foo
```

```txt
2024-11-24T17:15:00.042574+01:00 rpclock http_request.arduino <feff>[D][http_request.arduino:123]: Content-Length: 4736
[followed by silence]
```

### Additional information

_No response_

afflux:

Move HTTP receive loop into Component::loop implementation to prevent a slow server or lost data to permanently block.
Breaking behavior can be argued: previously, in a sequence of actions, the next action would only start after the HTTP request is finished, now it would start after the request is sent and status code is received.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6493

- esphome/esphome-docs#5627

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
script:
  - id: update_foo
    mode: single
    then:
      - http_request.post:
          url: http://foo/bar
          capture_response: true
          max_response_buffer_size: 6kB
time:
  - platform: sntp
    id: sntp_time
    timezone: Etc/UTC
    servers:
      - 192.168.1.1

    on_time:
      - seconds: 0
        minutes: /15
        then:
          - script.execute: update_foo
```

## Checklist:


afflux:
converted back to draft, still missing deallocations in the error cases

afflux:
Fix verified successfully with this testcase: https://github.com/esphome/issues/issues/6493#issuecomment-2508944465

afflux:
still relevant

esphome/components/http_request/http_request.h
```diff
@@ -188,54 +190,119 @@ template<typename... Ts> class HttpRequestSendAction : public Action<Ts...> {
...
-        // one so that modifications made in one trigger are not visible to
-        // the others
-        auto response_body_copy = std::string(response_body);
-        trigger->process(container, response_body_copy);
+    this->num_running_ = 0;
+    this->stop_next_();
+  }
+
+  void loop() override {
+    ExternalRAMAllocator<uint8_t> allocator(ExternalRAMAllocator<uint8_t>::ALLOW_FAILURE);
```
bdraco: We can use `enable_loop` and `disable_loop` here to only start the loop when a request is active

afflux:
This should be fairly easy to test with an integration test, I'll add some in the next few days.

bdraco:
Marking it as draft so we get a ping when its ready

afflux:
Not so easy after all: the host implementation of http_request uses [httplib](https://github.com/yhirose/cpp-httplib), which is blocking the request until the full response has been received. I see no way to work around that, except to drop httplib in favor of libcurl, but I have no idea how to get this working with platform io.

afflux:
@bdraco: what are your thoughts? Drop the tests, or switch to an async backend for http_requests on `host` platform?

bdraco:
Everything should be async on all platforms to avoid blocking the loop.  I haven't had a chance to full dig into each implementation on each platform yet though

PR #7897, opened 2024-11-30
add ID to wifi_info

Links2004:

this allows to use !remove on merge / include

one use case is if e.g. wifi is removed since one of your ESP32 is using ETH

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
packages:
  common: !include includes/test.yaml

text_sensor:
  - id: !remove wifi_info_id
  
# includes/test.yaml
....
text_sensor:
  - platform: wifi_info
    id: wifi_info_id
    bssid:
      name: BSSID
....
```

## Checklist:


DFINITYManu (approved):

Links2004:
waiting for merge

swoboda1337:
Thanks for the contribution @Links2004. However, after reviewing the implementation, I don't think this approach aligns with ESPHome's architecture.

The wifi_info platform is a "multi-sensor platform" that doesn't create a single component. Instead, it optionally creates
multiple independent text sensors (ip_address, ssid, bssid, mac_address, dns_address, scan_results). Each of these already
gets its own auto-generated ID through text_sensor.text_sensor_schema().

Adding a top-level id field to the platform configuration has two problems:
1. The ID is never used in the to_code() function, so it doesn't register a component that can be referenced or removed
2. Even if it did, you'd be trying to assign an ID to a platform configuration rather than a component instance

The use case is valid, but the solution needs to match ESPHome's component architecture patterns. I'm closing this PR, but
feel free to open a discussion issue if you'd like to explore alternative approaches.

PR #7898, opened 2024-11-30
Add to_ntc_resistance|temperature sensor filter (esphome/feature-requests#2967)

### LINKED ISSUE #2967: Component required to replace a real NTC with a device operated with ESPHome

I need an esphome component that simulates an NTC resistor, i.e. it receives a temperature and returns the corresponding resistance value.

I have a Buderus heater that uses an NTC resistor to detect the outside temperature. Over time, it turned out that the resistors used were unreliable (I was forced to replace them twice). Since I already have a Zigbee-based outdoor temperature sensor connected to Homeassistant, there is a desire to replace the real NTC with an ESPHome-powered device. For this use case, I need the reverse functionality of the current NTC component. 

**Additional context**

rhabacker:
Fixes https://github.com/esphome/feature-requests/issues/2967

Adds additional sensor filters ``to_ntc_resistance`` and ``to_ntc_temperature`` that can be used to convert temperature/resistance readings to resistance/temperature readings using an NTC characteristic.

## Types of changes

- [ ] Other

- fixes feature-requests/issues#2967

- esphome/esphome-docs#4485

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [x] BK72xx
- [x] RTL87xx

```yaml
    - platform: template
      id: to_ntc_resistance_sensor1
      unit_of_measurement: "Ohm"
      lambda: |-
        return id(some_sensor).state;
      update_interval: 1s
      filters:
        - to_ntc_resistance:
            calibration:
              - 10.0kOhm -> 25°C
              - 27.219kOhm -> 0°C
              - 14.674kOhm -> 15°C
```

```yaml
    - platform: template
      id: to_ntc_temperature_sensor1
      unit_of_measurement: "°C"
      lambda: |-
        return id(some_sensor).state;
      update_interval: 1s
      filters:
        - to_ntc_temperature:
            calibration:
              - 10.0kOhm -> 25°C
              - 27.219kOhm -> 0°C
              - 14.674kOhm -> 15°C
```

## Checklist:


esphome/components/ntcsim/sensor.py
```diff
@@ -0,0 +1,154 @@
...
+    sensor.sensor_schema(
+        NTCSimulation,
+        unit_of_measurement=UNIT_OHM,
+        accuracy_decimals=4,
+        device_class=DEVICE_CLASS_EMPTY,
+        state_class=STATE_CLASS_MEASUREMENT,
+    )
+    .extend(
+        {
+            cv.Required(CONF_SENSOR): cv.use_id(sensor.Sensor),
```
DjordjeMandic: I want to hear others what they think about this key, I think that "source_sensor" or "source_temperature_sensor" would be more appropriate to avoid confusion. https://esphome.io/components/copy.html

rhabacker: I searched the existing components for similar names for temperature-related sources and found `temperature_source` and used that.

DFINITYManu (approved):

rhabacker:
Anything left to do for this pull request ?

esphome/components/ntcsim/ntcsim.cpp
```diff
@@ -0,0 +1,48 @@
+#include "ntcsim.h"
+
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace ntcsim {
+
+static const char *const TAG = "ntcsimulation";
```
DjordjeMandic:
```suggestion
static const char *const TAG = "ntcsim";
```
```suggestion
static const char *const TAG = "ntc.simulation";
```

esphome/components/ntcsim/ntcsim.cpp
```diff
@@ -0,0 +1,48 @@
...
+  sensor_->add_on_state_callback([this](float value) { process_(value); });
+  if (sensor_->has_state())
+    process_(sensor_->state);
+}
+void NTCSimulation::dump_config() {
+  LOG_SENSOR("", "NTC Simulation", this);
+  ESP_LOGI(TAG, "  Coefficient a: %.8f", a_);
+  ESP_LOGI(TAG, "  Coefficient b: %.8f", b_);
+  ESP_LOGI(TAG, "  Coefficient c: %.8f", c_);
+}
```
DjordjeMandic:
```suggestion
void NTCSimulation::setup() {
  this->sensor_->add_on_state_callback([this](float value) { this->process_(value); });
  if (this->sensor_->has_state())
    this->process_(this->sensor_->state);
}
void NTCSimulation::dump_config() {
  LOG_SENSOR("", "NTC Simulation", this);
  ESP_LOGI(TAG, "  Coefficient a: %.8f", this->a_);
  ESP_LOGI(TAG, "  Coefficient b: %.8f", this->b_);
  ESP_LOGI(TAG, "  Coefficient c: %.8f", this->c_);
}
```

esphome/components/ntcsim/ntcsim.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+float NTCSimulation::get_setup_priority() const { return setup_priority::DATA; }
+
+void NTCSimulation::process_(float value) {
+  if (std::isnan(value)) {
+    publish_state(NAN);
+    return;
+  }
```
DjordjeMandic:
```suggestion
  if (!std::isfinite(value)) {
    this->publish_state(NAN);
    return;
  }
```

Use this if you don't accept -INF / +INF values and NAN.

esphome/components/ntcsim/ntcsim.cpp
```diff
@@ -0,0 +1,48 @@
...
+  double k = 273.15;
+  // https://de.wikipedia.org/wiki/Steinhart-Hart-Gleichung#cite_note-stein2_s4-3
+  double t = value + k;
+  double y = (a_ - 1 / (t)) / (2 * c_);
+  double x = sqrt(pow(b_ / (3 * c_), 3) + y * y);
```
DjordjeMandic:
```suggestion
  double y = (this->a_ - 1 / (t)) / (2 * this->c_);
  double x = sqrt(pow(this->b_ / (3 * this->c_), 3) + y * y);
```

esphome/components/ntcsim/ntcsim.h
```diff
@@ -0,0 +1,40 @@
...
+class NTCSimulation : public Component, public sensor::Sensor {
+ public:
+  void set_sensor(Sensor *sensor) { sensor_ = sensor; }
+#ifdef USE_OUTPUT
+  void set_output(output::FloatOutput *output) { output_ = output; }
+  void set_output_resistance(float resistance) { resistance_ = resistance; }
+#endif
+  void set_a(double a) { a_ = a; }
+  void set_b(double b) { b_ = b; }
+  void set_c(double c) { c_ = c; }
```
DjordjeMandic:
```suggestion
  void set_sensor(Sensor *sensor) { this->sensor_ = sensor; }
#ifdef USE_OUTPUT
  void set_output(output::FloatOutput *output) { this->output_ = output; }
  void set_output_resistance(float resistance) { this->resistance_ = resistance; }
#endif
  void set_a(double a) { this->a_ = a; }
  void set_b(double b) { this->b_ = b; }
  void set_c(double c) { this->c_ = c; }
```

tests/components/ntcsim/test.esp8266-ard.yaml
DjordjeMandic, line 1: https://esphome.io/guides/contributing.html#test-configurations

Please create common.yaml and a test for each framework/architecture combination

esphome/components/ntcsim/ntcsim.cpp
```diff
@@ -0,0 +1,48 @@
...
+  double resistance = exp(pow(x - y, 1 / 3.0) - pow(x + y, 1 / 3.0));
+  ESP_LOGI(TAG, "got resistor value %f for temperature %f", resistance, value);
+  ESP_LOGD(TAG, "'%s' - Resistance: %.1f Ohm", this->name_.c_str(), resistance);
+  publish_state(resistance);
+#ifdef USE_OUTPUT
+  if (output_) {
+    double normalized_resistance = resistance / resistance_;
+    ESP_LOGI(TAG, "send normalized resistor value %f to output device", normalized_resistance);
+    output_->set_level(normalized_resistance);
+  }
```
DjordjeMandic:
```suggestion
  this->publish_state(resistance);
#ifdef USE_OUTPUT
  if (this->output_) {
    double normalized_resistance = resistance / resistance_;
    ESP_LOGI(TAG, "send normalized resistor value %f to output device", normalized_resistance);
    this->output_->set_level(normalized_resistance);
  }
```

DjordjeMandic (changes requested):
> All uses of class members and member functions should be prefixed with this-> to distinguish them from global functions/variables.

> There should be at least one test for each framework/architecture combination.

Most of it looks good to me. Little things have to be addressed and that's all.

https://esphome.io/guides/contributing.html

DjordjeMandic (approved):

rhabacker:
Thanks for the approval - currently I see that “at least one positive review from reviewers with writing authorization is required” - what's the next step?

DjordjeMandic:
I guess only thing we can do is wait :)

https://esphome.io/guides/contributing.html#review-by-maintainers

Maintainers still need to review everything, share their opinions, and make the final decision. My review is just here to help catch common issues that often get flagged for changes. Reviews from those without write access serve as references but don’t directly impact the Maintainers' decisions. Even my reviews sometimes get corrected by Maintainers, but having additional reviews can still be helpful during the PR process.

jesserockz:
Just reading through the description and example YAML, this is not a `sensor` as it does not provide data to ESPHome, but is ESPHome providing data out. I think it should just be a top level component.

jesserockz (dismissed):
As above

rhabacker:
> I think it should only be a top level component.

Currently, this component must be defined as: 
```
 sensor:
      - platform: ntcsim
      ....
```
You are requesting to change the type to : 
```
 ???:
      - Platform: ntcsim
      ....
```

clydebarrow:
Why not make it a filter, remove the `output` specification and just use `on_value` to set a separate output?

DjordjeMandic:
> Why not make it a filter, remove the `output` specification and just use `on_value` to set a separate output?

This!!!

rhabacker:
> `on_value`

Is the value passed to the trigger also limited to a value between 0 and 1 as with the output component?

rhabacker:
> Why not make it a filter, remove the `output` specification and just use `on_value` to set a separate output?

Are there any limitations with this type of component compared to a sensor component?

clydebarrow:
> Are there any limitations with this type of component compared to a sensor component?

Not that I can think off. A filter gets applied to a sensor (which could be a real sensor, or a copy sensor, or a template) and the end result is the same - you have a sensor whose output looks like an NTC resistor value. Sending it to a destination, in your case an output, uses an existing feature. The advantage of using a filter is more flexibility and the use of building blocks to achieve the result rather than adding a new, lumpy, single purpose component.

clydebarrow:
> Is the value passed to the trigger also limited to a value between 0 and 1 as with the output component?

No, it can be scaled to whatever range you like e.g by using the offset and multiply filters.

rhabacker:
This component was created using the `ntc` component as a template. Is there a template where I can see how to create a corresponding filter ?

clydebarrow:
AFAIK currently all the sensor filters are defined in the sensor component. It's easy enough to add one.

rhabacker:
A look at  https://esphome.io/components/sensor/index.html#calibrate-linear shows this: 
``` 
# Example of a configuration entry
- platform: dht
  # ...
  temperature:
    name: “DHT22 Temperature”
    filters:
      - calibrate_linear:
         method: least_squares
         data points:
          # map 0.0 (from sensor) to 1.0 (true value)
          - 0.0 -> 1.0
          - 10.0 -> 12.1
```
An adjustment to the NTC simulation calculation would then look like this: 
``` 
# Example of a configuration entry
- platform: dht
  # ...
  temperature:
    name: “DHT22 Temperature”
    filters:
      - to_ntc_resistance:
         calibration:
           b_constant: 3950
           reference_temperature: 25°C
           reference_resistance: 10kOhm
```
or 
``` 
# Example of a configuration entry
- platform: dht
  # ...
  temperature:
    name: “DHT22 Temperature”
    filters:
      - to_ntc_resistance:
         calibration:
          - 10.0kOhm -> 25°C
          - 27.219kOhm -> 0°C
          - 14.674kOhm -> 15°C
```
There is still a question about the name of the filter, which I assumed above with `to_ntc_resistance`. It may also help to know that there is a reverse operation with the `ntc` component.

clydebarrow:
The name sounds fine to me.

rhabacker:
> Why not make it a filter, remove the `output` specification and just use `on_value` to set a separate output?

Do you have an example how to setup this ? 

```
    on_value:
     then:
       lambda: |-
        id(xxx).???(x);
```

clydebarrow:
> Do you have an example how to setup this ?

```
	on_value:
	  then:
	    - output.set_level:
	        id: output_id
	        level: !lambda return x;
```

rhabacker:
Related to the failed request https://github.com/esphome/esphome/actions/runs/13679997942/job/38249654359?pr=7898

I don't see any reference to the changes added by the commits included in this pull request. Does anyone have a clue ?

rhabacker:
> > Do you have an example how to setup this ?
> 
> ```
> 	on_value:
> 	  then:
> 	    - output.set_level:
> 	        id: output_id
> 	        level: !lambda return x;
> ```

Thanks - I will update the associated doc.

rhabacker:
> > ```
> > 	on_value:
> > 	  then:
> > 	    - output.set_level:
> > 	        id: output_id
> > 	        level: !lambda return x;
> > ```
> 
> Thanks - I will update the associated doc.

This requires additional information, as according to https://esphome.io/components/output/#output-set-level-action the values that are passed on to the output are limited to the range 0..1 (see also e.g. https://esphome.io/components/output/x9c.html) and require scaling with the actual resistance value, e.g. of the NTC:

```
    on_value:
      then:
        - output.set_level:
            id: output_id
            level: !lambda return x / 100000.0;
```
To see the actual value that is sent to the output, the following can be added: 
```
        - logger.log:
            format: "Send %.4f to output"
            args: [ x / 100000.0 ]
```

esphome/components/sensor/__init__.py
```diff
@@ -852,6 +856,138 @@ async def sensor_in_range_to_code(config, condition_id, template_arg, args):
...
+            raise cv.Invalid(
+                "Steinhart–Hart Calibration must consist of exactly three values"
+            )
+        value = cv.Schema([validate_ntc_calibration_parameter])(value)
+        a, b, c = ntc_calc_steinhart_hart(value)
+    else:
+        raise cv.Invalid(
+            f"Calibration parameter accepts either a list for steinhart-hart calibration, or mapping for b-constant calibration, not {type(value)}"
+        )
+    print("Coefficient:", a, b, c)
```
clydebarrow: Is this debug code?

clydebarrow (changes requested):

esphome/components/sensor/__init__.py
rhabacker: As the coefficients can be calculated, this should give the user the opportunity to check the structure. I have changed it so that it is displayed as a debug message.

clydebarrow: No change. Did you forget to push?

clydebarrow: It should be done as a logger INFO level message.

clydebarrow (changes requested):

esphome/components/sensor/__init__.py
```diff
@@ -852,6 +859,138 @@ async def sensor_in_range_to_code(config, condition_id, template_arg, args):
...
+                cv.Required(CONF_TEMPERATURE): cv.float_,
+                cv.Required(CONF_VALUE): cv.float_,
+            }
+        )(value)
+
+    value = cv.string(value)
+    parts = value.split("->")
+    if len(parts) != 2:
+        raise cv.Invalid("Calibration parameter must be of form 3000 -> 23°C")
+    voltage = cv.resistance(parts[0].strip())
```
clydebarrow:
```suggestion
    resistance = cv.resistance(parts[0].strip())
```

esphome/components/sensor/__init__.py
```diff
@@ -852,6 +859,138 @@ async def sensor_in_range_to_code(config, condition_id, template_arg, args):
     return var
 
 
+def validate_ntc_calibration_parameter(value):
+    if isinstance(value, dict):
+        return cv.Schema(
+            {
+                cv.Required(CONF_TEMPERATURE): cv.float_,
+                cv.Required(CONF_VALUE): cv.float_,
```
clydebarrow: Maybe use cv.temperature and cv.resistance here?

clydebarrow (changes requested):
Otherwise looks good.

clydebarrow:
Try to avoid force-pushing to a PR especially after review has started - it makes it harder to see what's been reviewed and what hasn't. If you committed by mistake, just revert that commit.

rhabacker:
> Try to avoid force-pushing to a PR especially after review has started - it makes it harder to see what's been reviewed and what hasn't. If you committed by mistake, just revert that commit.

I've seen [this](https://esphome.io/guides/contributing#your-local-copy-of-dev), but there was no mention of how to apply required changes to existing commits instead, which are normally incorporated through an interactive git rebase. In my opinion, there should be a note here that during an active review, changes are only allowed with additional commits (fixup commits) that must be squashed into the corresponding commit after a review is completed.

clydebarrow:
Those docs are a little out of date. All you need do to keep your branch up to date is `git pull upstream dev`. It's also not essential to keep it up to date constantly - as long as github is not flagging conflicts there's no big worry.

And don't worry about squashing commits - that gets done when merging anyway. It's fine to squash commits before pushing, but once it's pushed, leave the history alone.

rhabacker:
The way I see it, the CI error at https://github.com/esphome/esphome/actions/runs/14720500742/job/41313323302 is not related to changes from this pull request - how to proceed ?

clydebarrow (approved):

PR #7908, opened 2024-12-03
[image] Transparency changes; code refactor

### LINKED ISSUE #6446: When specifying AUTO_LOAD = ["animation"] the error `Failed config animation: None`
### The problem

When specifying in component:
```
AUTO_LOAD = ["animation"]
```
The error:
```
Failed config

animation: None
  
  'id' is a required option for [animation].
  []
required key not provided
```
If you add a file to the config that uses this string component:
```
animation:
  - file: images/1pixel.gif
    id: empty_animation
```
the build goes without problems. 
Moreover, if you specify `AUTO_LOAD = ["image"]` then there is no error and the `image` component gets into the config.

2024.10.3

Home Assistant Add-on

_No response_

ESP32

### Board

_No response_

animation

### Example YAML snippet

```yaml
AUTO_LOAD = ["image", "animation"]
```

```txt
INFO ESPHome 2024.10.3
INFO Reading configuration /config/esphome/test.yaml...
INFO Detected timezone 'Europe/Kiev'
Failed config

animation: None
  
  'id' is a required option for [animation].
  []
required key not provided
```

### Additional information

_No response_

### LINKED ISSUE #6498: RGB565 with Alpha-Data since #7631
### The problem

Hi,
since #7631 (@clydebarrow) images with type RGB565 support transparency in an "weird" (sorry i'm no nativ english speaker and don't if this wording is offending) way. There is a distinction between the type RGB24 (RGB without transparancy) and RGBA (with transparency) But RGB565 has a kind of hidden, implict way of handling the transparancy.
Also RGB565 is for saving memory introducing transparancy to it increases memory requirement to 33%.

2024.11.0

Home Assistant Add-on

all

ESP8266

### Board

UIlanzi TC001

image and animation

### Example YAML snippet

```yaml
not relevant
```

```txt
I think when converting to RGB565 the transparency of an image should be considered but but for storing and using in esphome the should be two image types RGB565 and RGB565A
```

### Additional information

see discussion here: https://github.com/lubeda/EspHoMaTriXv2/issues/255

### LINKED ISSUE #6522: local SVG images (non-mdi)
### The problem

ESPHome fails to build if image is SVG (pillow and cariosvg installed).

Build error:
```
  File "/usr/local/bin/esphome", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 1036, in main
    return run_esphome(sys.argv)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 1023, in run_esphome
    rc = POST_CONFIG_ACTIONS[args.command](args, config)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 503, in command_run
    exit_code = write_cpp(config)
                ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 212, in write_cpp
    generate_cpp_contents(config)
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 224, in generate_cpp_contents
    CORE.flush_tasks()
  File "/usr/local/lib/python3.12/site-packages/esphome/core/__init__.py", line 684, in flush_tasks
    self.event_loop.flush_tasks()
  File "/usr/local/lib/python3.12/site-packages/esphome/coroutine.py", line 246, in flush_tasks
    next(task.iterator)
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 204, in wrapped
    await coro(conf)
  File "/usr/local/lib/python3.12/site-packages/esphome/components/image/__init__.py", line 290, in to_code
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/PIL/Image.py", line 3498, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x7f86794039c0>

```

2024.11.2

pip

_No response_

ESP32-IDF

### Board

Guition display board (ESP-S3)

image

### YAML Config

```yaml
substitutions:
  name: "guition-esp32-s3-4848s040"
  friendly_name: "Guition480-basic"
  device_description: "Guition ESP32-S3-4848S040 480*480 Smart Screen"
  project_name: "Guition.ESP32_S3_4848S040"
  project_version: "1.0.0"

esphome:
  name: "${name}"
  friendly_name: "${friendly_name}"
  #name_add_mac_suffix: true
  project:
    name: "${project_name}"
    version: "${project_version}"
  platformio_options:
    board_build.flash_mode: dio

esp32:
  board: esp32-s3-devkitc-1
  variant: esp32s3
  flash_size: 16MB
  framework:
    type: esp-idf
    sdkconfig_options:
      COMPILER_OPTIMIZATION_SIZE: y
      CONFIG_ESP32S3_DEFAULT_CPU_FREQ_240: "y"
      CONFIG_ESP32S3_DATA_CACHE_64KB: "y"
      CONFIG_ESP32S3_DATA_CACHE_LINE_64B: "y"
      CONFIG_SPIRAM_FETCH_INSTRUCTIONS: y
      CONFIG_SPIRAM_RODATA: y

psram:
  mode: octal
  speed: 80MHz

logger:

wifi:
  ssid: !secret wifi_ssid

web_server:
  port: 80

#
#
# Images
#
#
image:
  - file: "/usr/src/esphome-configs-main/assets/forecast-icons/000.svg"
    id: idokep_fc_icon_000
    resize: 80x80

#
#-------------------------------------------
# LVGL Buttons
#-------------------------------------------
#
lvgl:
  displays:
    - my_display
  touchscreens:
    - touchscreen_id: my_touchscreen
  on_idle:
    - timeout: 10s
      then:
        - logger.log: idle timeout
        - if:
            condition:
              lvgl.is_idle:
                timeout: 5s
            then:
              - logger.log: LVGL is idle
    - timeout: 15s
      then:
        - logger.log: idle 15s timeout

      #- lvgl.pause:
      #- light.turn_off:
      #    id: display_backlight
      #    transition_length: 5s


#-------------------------------------------
# Internal outputs
#-------------------------------------------
output:
  # Backlight LED
  - platform: ledc
    pin: GPIO38
    id: GPIO38
    frequency: 100Hz

    # Built in 240v relay
  - id: internal_relay_1
    platform: gpio
    pin: 40

#-------------------------------------------
# Touchscreen gt911 i2c
#-------------------------------------------
i2c:
  - id: bus_a
    sda: GPIO19
    scl: GPIO45
    #frequency: 100kHz

touchscreen:
  platform: gt911
  transform:
    mirror_x: false
    mirror_y: false
  id: my_touchscreen
  display: my_display

  on_touch:
    - logger.log:
        format: Touch at (%d, %d)
        args: [touch.x, touch.y]
    - lambda: |-
            touch.x,
            touch.y,
            touch.x_raw,
            touch.y_raw
            );

#-------------------------------------------
# Display st7701s spi
#-------------------------------------------
spi:
  - id: lcd_spi
    clk_pin: GPIO48
    mosi_pin: GPIO47

display:
  - platform: st7701s
    id: my_display
    update_interval: never
    auto_clear_enabled: False
    spi_mode: MODE3
    data_rate: 2MHz
    color_order: RGB
    invert_colors: False
    dimensions:
      width: 480
      height: 480
    cs_pin: 39
    de_pin: 18
    hsync_pin: 16
    vsync_pin: 17
    pclk_pin: 21
    pclk_frequency: 12MHz
    pclk_inverted: False
    hsync_pulse_width: 8
    hsync_front_porch: 10
    hsync_back_porch: 20
    vsync_pulse_width: 8
    vsync_front_porch: 10
    vsync_back_porch: 10
    init_sequence:
      - 1
      # Custom sequences are an array, first byte is command, the rest are data.
      - [0xFF, 0x77, 0x01, 0x00, 0x00, 0x10] # CMD2_BKSEL_BK0
      - [0xCD, 0x00] # disable MDT flag
    data_pins:
      red:
        - 11 #r1
        - 12 #r2
        - 13 #r3
        - 14 #r4
        - 0 #r5
      green:
        - 8 #g0
        - 20 #g1
        - 3 #g2
        - 46 #g3
        - 9 #g4
        - 10 #g5
      blue:
        - 4 #b1
        - 5 #b2
        - 6 #b3
        - 7 #b4
        - 15 #b5
```

```txt
INFO Generating C++ source...
  File "/usr/local/bin/esphome", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 1036, in main
    return run_esphome(sys.argv)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 1023, in run_esphome
    rc = POST_CONFIG_ACTIONS[args.command](args, config)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 503, in command_run
    exit_code = write_cpp(config)
                ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 212, in write_cpp
    generate_cpp_contents(config)
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 224, in generate_cpp_contents
    CORE.flush_tasks()
  File "/usr/local/lib/python3.12/site-packages/esphome/core/__init__.py", line 684, in flush_tasks
    self.event_loop.flush_tasks()
  File "/usr/local/lib/python3.12/site-packages/esphome/coroutine.py", line 246, in flush_tasks
    next(task.iterator)
  File "/usr/local/lib/python3.12/site-packages/esphome/__main__.py", line 204, in wrapped
    await coro(conf)
  File "/usr/local/lib/python3.12/site-packages/esphome/components/image/__init__.py", line 290, in to_code
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/PIL/Image.py", line 3498, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x7f09a6080bd0>
```

### Additional information

SVG content:
```
<?xml version="1.0" encoding="UTF-8"?><svg xmlns="http://www.w3.org/2000/svg" width="1" height="1"/>
```

clydebarrow:

A previous PR changed the transparency type for some images from chroma-key (where a specific color flags transparency) to alpha-channel (where an additional 8 bits per pixel encodes the opacity value.)

Now there is a choice available for which type of transparency is used. Full list of changes:

0. `use_transparency` has options `none`, `chroma_key` and `alpha_channel`
1. Image type is a required field - the previous default was BINARY which is often not what is expected; But see 6.
2. The RGBA and TRANSPARENT_BINARY types are removed, instead use RGB or BINARY with an appropriate setting for `use_transparency`
3. `online_image` and `animation` code and schemas are streamlined and re-use `image` code where possible.
4. The  `image` code has been extensively refactored to facilitate maintenance.
5. Eliminate use of `puremagic` for checking if a file is of type `svg` - it turns out that `puremagic` merely checks that the file ends with `</svg>` which fails in the corner case of an empty svg file (which will end with ` />` only.) The code now simply checks that "<svg " appears early in the file.
6. Add a schema that permits images to be grouped by type and transparency to avoid repeating types.

## Testing

This PR needs testing before release since it contains breaking changes. Please test by adding this to your yaml:

```
external_components:
  - source: github://pr#7908
    components: [image,online_image,animation]
```

And make necessary changes to configs as above (or per the error messages when building.)

Please report any problems as comments to this PR.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6498
- fixes https://github.com/esphome/issues/issues/6522
- fixes https://github.com/esphome/issues/issues/6446

- esphome/esphome-docs#4506
## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml

image:
  rgb565:
    - file: cloudy.png
      id: cloudy_image
      resize: 50x50
      use_transparency: alpha_channel
    - file: sunny.png
      id: sunny_image
      resize: 50x50
  binary:
    opaque:
    - file: logo-text-trans.svg
      resize: 100x100
      id: alpha_binary

  grayscale:
    alpha_channel:
      - file: bluemax.png
        resize: 100x100
        id: esphome_png
```

## Checklist:


esphome/components/image/__init__.py
```diff
@@ -203,45 +375,68 @@ def validate_file_shorthand(value):
...
+    return cv.one_of(*TRANSPARENCY_TYPES, lower=True)(value)
+
 
+def validate_type(value):
+    value = cv.one_of(*IMAGE_TYPE, upper=True)(value)
+    return IMAGE_TYPE[value].validate(value)
+
+
+def validate_settings(value):
+    type = value[CONF_TYPE]
```
guillempages, line 396: Somehow the new validation lost the checks for MDI; was this on purpose?  In the old implementation, MDI images were set to transparent by default (because that's what the average user will expect), and only allowed binary types (transparent or not-transparent).

As far as I can see, now there is no distinction between mdi images or other images; so the user will manually have to specify the transparency, even if it does not make sense.

Also, mdi images are binary on the source, so it would be nice if the "mdi" type could still have the type set to binary by default, instead of having to explicitly add binary on each image.

esphome/components/image/__init__.py
```diff
@@ -203,45 +375,68 @@ def validate_file_shorthand(value):
...
-                "NONE", "FLOYDSTEINBERG", upper=True
-            ),
-            cv.GenerateID(CONF_RAW_DATA_ID): cv.declare_id(cg.uint8),
-        },
-        validate_cross_dependencies,
-    )
-)
+    {
+        cv.Required(CONF_ID): cv.declare_id(Image_),
+        cv.Required(CONF_TYPE): validate_type,
```
guillempages: As mentioned above; I think it would be better to keep this as optional, to allow mdi images to be configured with the least amount of parameters possible, and "require" the type for other images in the validation function

esphome/components/image/__init__.py
```diff
@@ -203,45 +375,68 @@ def validate_file_shorthand(value):
...
+        cv.GenerateID(CONF_RAW_DATA_ID): cv.declare_id(cg.uint8),
+    }
+).add_extra(validate_settings)
 
 CONFIG_SCHEMA = IMAGE_SCHEMA
 
 
-def load_svg_image(file: bytes, resize: tuple[int, int]):
+def load_svg_image(file, resize: tuple[int, int]):
     # Local imports only to allow "validate_pillow_installed" to run *before* importing it
```
guillempages: Is `validate_pillow_installed` still called at all? The comment is probably wrong

esphome/components/image/__init__.py
```diff
@@ -256,152 +451,77 @@ def load_svg_image(file: bytes, resize: tuple[int, int]):
...
     resize = config.get(CONF_RESIZE)
-    if "svg" in file_type:
-        image = load_svg_image(file_contents, resize)
+    if is_svg_file(path):
+        from cairosvg import svg2png
+
+        if not resize:
+            resize = (None, None)
+        with open(path, "rb") as file:
+            image = svg2png(
```
guillempages, line 506: Just out of curiosity; why open it as png, and then decoding into the buffer again, instead of keeping the old implementation that directly created the image from the svg?

esphome/components/image/image.cpp
```diff
@@ -61,20 +51,20 @@ void Image::draw(int x, int y, display::Display *display, Color color_on, Color
...
   }
 }
-Color Image::get_pixel(int x, int y, Color color_on, Color color_off) const {
+Color Image::get_pixel(int x, int y, const Color color_on, const Color color_off) const {
   if (x < 0 || x >= this->width_ || y < 0 || y >= this->height_)
     return color_off;
   switch (this->type_) {
     case IMAGE_TYPE_BINARY:
-      return this->get_binary_pixel_(x, y) ? color_on : color_off;
+      if (this->get_binary_pixel_(x, y))
```
guillempages, line 59: Any specific reason to not use the ternary operator here? It made the switch look more consistent (i.e. one return per case).

esphome/components/image/__init__.py
clydebarrow: MDI images (which are SVG) are not binary in source, they're vector drawings and `binary` for MDI (and any resized image with transparency) results in aliasing artefacts so that's the reason for not providing a default. Grayscale works much better for MDI. Before LVGL was merged (and the display component couldn't render transparency properly) it was not an issue.

Basically it comes down to the fact that a default type is going to be not what is wanted more often than not, so by requiring the type to be specified, the user has to at least think about what the desired result is, and if it's not as expected, there is already a starting point to tweak.

Transparency by default would make sense for MDI - I'll ponder that.

esphome/components/image/__init__.py
clydebarrow: As per comment above I disagree. If there were to be a default for MDI it should be grayscale, and since that would be a breaking change I think it's better to make the user be explicit.

esphome/components/image/image.h
```diff
@@ -12,59 +12,50 @@ namespace image {
...
-      case IMAGE_TYPE_RGB565:
-        return this->transparent_ ? 24 : 16;
-      case IMAGE_TYPE_RGB24:
-        return 24;
-      case IMAGE_TYPE_RGBA:
-        return 32;
-    }
-    return 0;
-  }
+  int get_bpp() const { return this->bpp_; }
```
guillempages, line 34: Nice one! Makes the checking of bpp once at construction instead of on every pixel.

esphome/components/image/image.cpp
clydebarrow: One of the linters didn't like it, not sure why but it was simpler just to rewrite it.

esphome/components/image/image.h
```diff
@@ -12,59 +12,50 @@ namespace image {
...
 
   /// Return the stride of the image in bytes, that is, the distance in bytes
   /// between two consecutive rows of pixels.
-  uint32_t get_width_stride() const { return (this->width_ * this->get_bpp() + 7u) / 8u; }
+  size_t get_width_stride() const { return (this->width_ * this->get_bpp() + 7u) / 8u; }
```
guillempages, line 38: maybe this could also be calculated at the constructor, instead of each time the method is called? width_ should be constant (over the image lifetime), and bpp as well.

esphome/components/online_image/__init__.py
```diff
@@ -57,48 +61,54 @@
...
     cv.All(
         ONLINE_IMAGE_SCHEMA,
-        validate_cross_dependencies,
         cv.require_framework_version(
             # esp8266 not supported yet; if enabled in the future, minimum version of 2.7.0 is needed
             # esp8266_arduino=cv.Version(2, 7, 0),
             esp32_arduino=cv.Version(0, 0, 0),
             esp_idf=cv.Version(4, 0, 0),
             rp2040_arduino=cv.Version(0, 0, 0),
+            host=cv.Version(0, 0, 0),
```
guillempages, line 132: Does this now work? IIRC it was not working because the http_request dependency didn't either; but this might have changed in the meantime

esphome/components/online_image/__init__.py
```diff
@@ -133,28 +143,27 @@ async def online_image_action_to_code(config, action_id, template_arg, args):
 
 async def to_code(config):
     format = config[CONF_FORMAT]
-    if format in [FORMAT_PNG]:
+    if format == FORMAT_PNG:
```
guillempages: Please, leave the "in [FORMAT_PNG]"; at some point I plan to add an "FORMAT_AUTO" as well, that should include all possible formats

esphome/components/image/__init__.py
clydebarrow: The old implementation did exactly the same. All I did was inline the function.

esphome/components/online_image/image_decoder.h
```diff
@@ -59,15 +58,15 @@ class ImageDecoder {
...
+   * @param color The fill color
    */
   void draw(int x, int y, int w, int h, const Color &color);
 
   bool is_finished() const { return this->decoded_bytes_ == this->download_size_; }
 
  protected:
   OnlineImage *image_;
-  // Initializing to 1, to ensure it is different than initial "decoded_bytes_".
+  // Initializing to 1, to ensure it is different to initial "decoded_bytes_".
```
guillempages: You are a native english speaker, so I won't insist much on that... But to me "different than" sounds better than "different to"... But it might be a British vs American vs Australian thing

esphome/components/online_image/online_image.cpp
```diff
@@ -71,24 +71,16 @@ bool OnlineImage::resize_(int width_in, int height_in) {
     return false;
   }
   auto new_size = this->get_buffer_size_(width, height);
-  ESP_LOGD(TAG, "Allocating new buffer of %d Bytes...", new_size);
+  ESP_LOGV(TAG, "Allocating new buffer of %d Bytes...", new_size);
```
guillempages: I'd like to keep at least this one as "debug"; memory is still an issue for some users, and having a log explicitly mentioning how much memory is being allocated helps with debugging user issues.
```suggestion
  ESP_LOGD(TAG, "Allocating new buffer of %d Bytes...", new_size);
```

esphome/components/online_image/online_image.cpp
```diff
@@ -71,24 +71,16 @@ bool OnlineImage::resize_(int width_in, int height_in) {
...
-#if defined(USE_ESP8266)
-    // NOLINTNEXTLINE(readability-static-accessed-through-instance)
-    int max_block = ESP.getMaxFreeBlockSize();
-#elif defined(USE_ESP32)
-    int max_block = heap_caps_get_largest_free_block(MALLOC_CAP_INTERNAL);
-#else
-    int max_block = -1;
-#endif
-    ESP_LOGE(TAG, "allocation failed. Biggest block in heap: %d Bytes", max_block);
+    ESP_LOGE(TAG, "allocation of %d bytes failed", new_size);
```
guillempages: Please, keep the old implementation with the actual checking of the available max block... There have been cases of complaints "I have xxx bytes free, but the allocation failed", and this helps debugging (or even avoiding the complain) that

esphome/components/online_image/png_image.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "image_decoder.h"
+#include "esphome/core/defines.h"
```
guillempages, line 4: why move the include here? If kept in the "image_decoder.h", it will be automatically included on all different image formats (e.g. jpeg) when implemented; instead of needing to explicitly add it to each format.

esphome/components/image/__init__.py
guillempages: I like my yaml files clean and DRY. Having to add an extra parameter for each MDI file I use is cumbersome.
You are a heavy defender of "make everything explicit that can have a consequence", whereas I'm a heavy defender of "put sensible defaults wherever possible, and allow overriding them for corner cases".
Changing the default to grayscale I could live with... But then this might again not work for binary eink displays (or at least waste memory)

esphome/components/image/__init__.py
guillempages: Making it explicit is also a breaking change

esphome/components/image/__init__.py
guillempages: Oops. My bad :-(

esphome/components/image/__init__.py
```diff
@@ -432,10 +432,6 @@ def validate_settings(value):
 
 
 def load_svg_image(file, resize: tuple[int, int]):
-    # Local imports only to allow "validate_pillow_installed" to run *before* importing it
```
guillempages, line 239: The mention of "local import of cairosvg" is still relevant IMHO. The esphome guideline is to use global imports, and this explicitly (and purposely) breaks this guideline IMHO a comment explaining why is needed, to avoid a future "let's clean this up" requiring cairosvg for everyone.

esphome/components/image/__init__.py
clydebarrow: I fully agree with sensible defaults where the only exceptions are indeed corner cases, I think since the expansion of the application of the image component it's not that clear-cut. This PR and the previous one are in response to real-world issues that have arisen due to unexpected results without obvious solutions. It's impossible to keep everybody happy so it's better IMO to at least keep everybody equally unhappy.

esphome/components/image/__init__.py
clydebarrow: > Making it explicit is also a breaking change

Sure, and that's the best kind of breaking change - it's obvious at build time, and the solution is proposed for you.

esphome/components/online_image/__init__.py
clydebarrow: It does work, I have a working http_request for host, not in PR yet. It was essential for testing.

esphome/components/online_image/image_decoder.h
clydebarrow: It just grated on me as I read it. I looked it up just now, and "different than" is an Americanism. The ROW uses "from" or "to".

esphome/components/image/__init__.py
clydebarrow: cairosvg is installed by default now. I'll move the import to the top level.

esphome/components/online_image/online_image.cpp
clydebarrow: Calculation of available heap should be abstracted out to the allocator, rather than having platform-dependent code here. That will need a separate PR to update the allocator code so I'll defer that for now.

esphome/components/online_image/png_image.h
clydebarrow: Simply because at moment there are no other formats so this is the only file where it's used. That can be updated when other formats are added.

esphome/components/image/image.h
clydebarrow: Can't `width_` change with online image?

guillempages: Oops, right. Forgot about the child classes :-(

esphome/components/online_image/image_decoder.h
guillempages: In that case, can we agree on "is different from"? Somehow I can't get "is different to" parsed right in my mind ;-)

esphome/components/online_image/online_image.cpp
```diff
@@ -71,37 +71,29 @@ bool OnlineImage::resize_(int width_in, int height_in) {
...
-#if defined(USE_ESP8266)
-    // NOLINTNEXTLINE(readability-static-accessed-through-instance)
-    int max_block = ESP.getMaxFreeBlockSize();
-#elif defined(USE_ESP32)
-    int max_block = heap_caps_get_largest_free_block(MALLOC_CAP_INTERNAL);
-#else
-    int max_block = -1;
+#if ESP_LOG_LEVEL >= ESPHOME_LOG_LEVEL_DEBUG
+  ESP_LOGD(TAG, "Allocating new buffer of %d bytes", new_size);
+  delay_microseconds_safe(2000);  // allow time for log line to appear
```
guillempages: IIRC the delay was not for the log, but because I had some strange crashes without it... But it was over a year ago, so I forgot the actual issue.

tests/components/animation/common.yaml
```diff
@@ -0,0 +1,23 @@
+animation:
```
guillempages, line 1: Maybe also port the comment explicitly mentioning to *not* load `image` manually (to test the auto-load). Just to make sure

tests/components/image/common.yaml
```diff
@@ -5,32 +5,36 @@ image:
...
-    type: RGBA
+    type: RGB
+    use_transparency: alpha_channel
     resize: 50x50
   - id: rgb24_image
     file: ../../pnglogo.png
-    type: RGB24
-    use_transparency: yes
+    type: RGB
+    use_transparency: alpha_channel
```
guillempages:
```suggestion
    use_transparency: chroma_key
```
The combination "RGB" with "alpha_channel" is already being tested in the image just above this one

esphome/components/online_image/online_image.cpp
clydebarrow: So remove it?

tests/components/animation/common.yaml
clydebarrow: Umm, why? Is a dev really going to look at that code and think "hmm, nothing is broken but I think I should add `image:` here".

clydebarrow:
@guillempages I have a plan to satisfy your DRY requirements while still avoiding unexpected results. Stand by...

clydebarrow:
@guillempages See here: https://deploy-preview-4506--esphome.netlify.app/components/image.html#grouping-images-by-type

esphome/components/online_image/image_decoder.h
clydebarrow: So now it reads
```
  // Initializing to 1, to ensure it is distinguishable from initial "decoded_bytes_".

```

requirements.txt
```diff
@@ -15,7 +15,6 @@ click==8.1.7
 esphome-dashboard==20241217.1
 aioesphomeapi==24.6.2
 zeroconf==0.132.2
-puremagic==1.27
```
jesserockz: Lets leave this in for now as there are external components that are using this library.

esphome/components/online_image/png_image.h
jesserockz: defines.h should always be included in any file it is needed. it is best not to rely on another file including it in case that is removed in the future.

esphome/components/image/image.h
```diff
@@ -12,59 +12,50 @@ namespace image {
...
   void draw(int x, int y, display::Display *display, Color color_on, Color color_off) override;
 
-  void set_transparency(bool transparent) { transparent_ = transparent; }
-  bool has_transparency() const { return transparent_; }
+  bool has_transparency() const { return transparency_ != TRANSPARENCY_OPAQUE; }
```
jesserockz:
```suggestion
  bool has_transparency() const { return this->transparency_ != TRANSPARENCY_OPAQUE; }
```

jesserockz (changes requested):

jesserockz (approved):

PR #7923, opened 2024-12-05
[http_request] on_response body must be owned

### LINKED ISSUE #5987: repeat action in http_request.get on_response handler fails to compile
### The problem

The repeat action can't be used inside the http_request `on_response` handler anymore because of a compile failure.

This is a regression from 2024.6.2 and the first bad commit is 0179358f9cbb212d30627999a2f7233ff18e12ea

2024.6.3+

pip

_No response_

ESP8266

### Board

d1_mini

http_request

### Example YAML snippet

```yaml
esphome:
  name: reproduce
  platform: ESP8266
  board: d1_mini
  on_boot:
    - priority: -100
      then:
        - http_request.get:
            url: "https://esphome.io/"
            on_response:
              then:
                - repeat:
                    count: 3
                    then:

wifi:
  ssid: "abcdefgh"
  password: "12345678"

http_request:
  useragent: esphome
  timeout: 2s
  id: http_request_data
  verify_ssl: false
```

```txt
src/main.cpp: In function 'void setup()':
src/main.cpp:149:99: error: use of deleted function 'esphome::RepeatAction<std::shared_ptr<esphome::http_request::HttpContainer>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&>::RepeatAction()'
  149 |   repeataction_id = new RepeatAction<std::shared_ptr<http_request::HttpContainer>, std::string &>();
      |                                                                                                   ^
In file included from src/esphome.h:18,
                 from src/main.cpp:3:
src/esphome/core/base_automation.h:273:32: note: 'esphome::RepeatAction<std::shared_ptr<esphome::http_request::HttpContainer>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&>::RepeatAction()' is implicitly deleted because the default definition would be ill-formed:
  273 | template<typename... Ts> class RepeatAction : public Action<Ts...> {
      |                                ^~~~~~~~~~~~
src/esphome/core/base_automation.h:273:32: error: no matching function for call to 'std::tuple<std::shared_ptr<esphome::http_request::HttpContainer>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&>::tuple()'
In file included from /home/me/.platformio/packages/toolchain-xtensa/xtensa-lx106-elf/include/c++/10.3.0/functional:54,
                 from /home/me/.platformio/packages/toolchain-xtensa/xtensa-lx106-elf/include/c++/10.3.0/pstl/glue_algorithm_defs.h:13,
                 from /home/me/.platformio/packages/toolchain-xtensa/xtensa-lx106-elf/include/c++/10.3.0/algorithm:74,
                 from /home/me/.platformio/packages/framework-arduinoespressif8266/cores/esp8266/Arduino.h:229,
                 from src/esphome/components/esp8266/gpio.h:6,
                 from src/esphome.h:4,
                 from src/main.cpp:3:
```

### Additional information

0179358f9cbb212d30627999a2f7233ff18e12ea is the first bad commit
commit 0179358f9cbb212d30627999a2f7233ff18e12ea

    Improve 'body' handling in http_request on_response triggers (#6968)

 esphome/codegen.py                             |  1 +
 esphome/components/http_request/__init__.py    |  2 +-
 esphome/components/http_request/http_request.h | 19 ++++++++++++++-----
 esphome/cpp_types.py                           |  1 +
 4 files changed, 17 insertions(+), 6 deletions(-)

afflux:

https://github.com/esphome/esphome/pull/6968 changed the response_body to a reference of string to avoid copies when passing the string by value. However, the owned string will be deallocated right after the call to process(), so actions such as delay and wait_until would be passed an invalid reference. ~Additionally, wait_until, while and repeat will not compile anymore, since they store their arguments in a default-constructed member - which fails for references.~ (fixed by #7972)

~This PR instead uses a shared_ptr, which will still avoid the copies between each action. However, users must now dereference the body argument in their on_response lambdas to get the std::string, therefore it is a breaking change.~ (fixed by #11704)

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/5987

N/A

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: reproduce
  platform: ESP8266
  board: d1_mini
  on_boot:
    - priority: -100
      then:
        - http_request.get:
            url: "https://example.com/"
            on_response:
              then:
                - repeat:
                    count: 3
                    then:

wifi:
  ssid: !secret wifi_ssid
  password: !secret wifi_psk

http_request:
  useragent: esphome
  timeout: 2s
  id: http_request_data
  verify_ssl: false
```

## Checklist:

### DIFF
--- a/esphome/components/http_request/__init__.py
+++ b/esphome/components/http_request/__init__.py
@@ -35,12 +35,6 @@ HttpContainer = http_request_ns.class_("HttpContainer")
 HttpRequestSendAction = http_request_ns.class_(
     "HttpRequestSendAction", automation.Action
 )
-HttpRequestResponseTrigger = http_request_ns.class_(
-    "HttpRequestResponseTrigger",
-    automation.Trigger.template(
-        cg.std_shared_ptr.template(HttpContainer), cg.std_string
-    ),
-)
 
 CONF_HTTP_REQUEST_ID = "http_request_id"
 
@@ -305,7 +299,7 @@ async def http_request_action_to_code(config, action_id, template_arg, args):
                 var.get_success_trigger_with_response(),
                 [
                     (cg.std_shared_ptr.template(HttpContainer), "response"),
-                    (cg.std_string_ref, "body"),
+                    (cg.std_string, "body"),
                     *args,
                 ],
                 response_conf,
--- a/esphome/components/http_request/http_request.h
+++ b/esphome/components/http_request/http_request.h
@@ -111,13 +111,6 @@ class HttpContainer : public Parented<HttpRequestComponent> {
   std::map<std::string, std::list<std::string>> response_headers_{};
 };
 
-class HttpRequestResponseTrigger : public Trigger<std::shared_ptr<HttpContainer>, std::string &> {
- public:
-  void process(const std::shared_ptr<HttpContainer> &container, std::string &response_body) {
-    this->trigger(container, response_body);
-  }
-};
-
 class HttpRequestComponent : public Component {
  public:
   void dump_config() override;
@@ -198,7 +191,7 @@ template<typename... Ts> class HttpRequestSendAction : public Action<Ts...> {
   void set_json(std::function<void(Ts..., JsonObject)> json_func) { this->json_func_ = json_func; }
 
 #ifdef USE_HTTP_REQUEST_RESPONSE
-  Trigger<std::shared_ptr<HttpContainer>, std::string &, Ts...> *get_success_trigger_with_response() const {
+  Trigger<std::shared_ptr<HttpContainer>, std::string, Ts...> *get_success_trigger_with_response() const {
     return this->success_trigger_with_response_;
   }
 #endif
@@ -259,7 +252,6 @@ template<typename... Ts> class HttpRequestSendAction : public Action<Ts...> {
           yield();
           read_index += read;
         }
-        response_body.reserve(read_index);
         response_body.assign((char *) buf, read_index);
         allocator.deallocate(buf, max_length);
       }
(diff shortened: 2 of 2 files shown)


afflux:
still relevant

bdraco:
@afflux would you please rebase

afflux:
Sure. I think with the change from https://github.com/esphome/esphome/pull/11464, I can skip the `shared_ptr` and just change it it back to a non-ref std::string...

afflux:
Well, I had another look at the disassembly and realized that all functions in the call tree below `Trigger::trigger()` will create copies of their arguments when passing them to the next function, so every std::string is copied at least three times before it even reaches the first action. I'm therefore now back to shared_ptr.

This has another caveat: [if allocation for the uint8 buffer failed](https://github.com/esphome/esphome/pull/7923/files#diff-1cf5bd82614da7ca2744ef3946b1ab6842d7652822ba1f00fc5c34ad6abd94bcR247), the shared_ptr will be empty, with UB when dereferencing it. So the breakage is even bigger, as every user has to check for validity and then dereference the body. We could just not call the trigger in this case and log an error.

Alternative to this, we could put some effort into reworking the automation trigger functions to properly use references and std::forward/std::move, to reduce the number of copies needed. (We could keep the current, slightly odd behavior of passing the empty string to the trigger when allocation failed.)

X-Ryl669:
Can't you use a const reference here instead ? It would avoid using the heap (and/or stack) to transmit the string argument.
If the pass by const ref doesn't work, what about moving the string to the levels below the first pass by value?

afflux:
See #11704 - it's a const reference already. To be perfectly honest, I'm not sure we still need this PR now (not much of a C++ wizard, no ideas what will happen with the double reference) but I guess it's cleaner this way.

X-Ryl669:
I don't know about the context. In your patch, you're changing a string reference to a value. This implies that any function in the callstack below this function will make a copy of the string (and thus, call the heap to allocate a buffer to copy the string content into). If the code works with a value, it's likely to work with a const reference (unless it's modifying the string's **content** in its scope). So instead of `string &` becoming `string`, make it `const string &` instead. That tells the compiler to only pass the address to the object below, which is fast and ensure there's no modification to the object. Most of the time, there isn't so it should be almost free win here.

In the worst case, if one of the function need to modify the string content, **and** none of the function above in the callstack use it afterward, you can pass the string by r-value (make it `string &&`) and move the string (with `std::move(variable)`) instead.

afflux:
I'm sorry, it doesn't make sense to argue without context. Please see the recently merged #11704, which is changing every argument to actions to const ref.

afflux:
@bdraco: let me know if you think we still need this PR. AFAICS, with #11704 merged, this now boils down to a minor cleanup

PR #7958, opened 2024-12-13
[rotary_encoder] Fix volatile increment/decrement deprecation warnings

edwardtfn:

Similar to the [pulse counter fix](https://github.com/esphome/esphome/pull/7954), this PR addresses warnings about deprecated use of increment/decrement operators on volatile-qualified types in the rotary encoder component. 

The warnings being fixed are:
```log
- warning: '++' expression of 'volatile'-qualified type is deprecated [-Wvolatile]
- warning: '--' expression of 'volatile'-qualified type is deprecated [-Wvolatile]
```

Changes:
- Keep the volatile qualifier as it's important for interrupt context
- Replace increment/decrement operations with load/store approach

This is a similar change to #7954 (pulse counter fix) but for the rotary encoder component.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/rotary_encoder/rotary_encoder.cpp
+++ b/esphome/components/rotary_encoder/rotary_encoder.cpp
@@ -92,14 +92,16 @@ void IRAM_ATTR HOT RotaryEncoderSensorStore::gpio_intr(RotaryEncoderSensorStore 
 
   int8_t rotation_dir = 0;
   uint16_t new_state = STATE_LOOKUP_TABLE[input_state];
+
+  esphome::InterruptLock lock;
   if ((new_state & arg->resolution & STATE_HAS_INCREMENTED) != 0) {
     if (arg->counter < arg->max_value)
-      arg->counter++;
+      arg->counter += 1;
     rotation_dir = 1;
   }
   if ((new_state & arg->resolution & STATE_HAS_DECREMENTED) != 0) {
     if (arg->counter > arg->min_value)
-      arg->counter--;
+      arg->counter -= 1;
     rotation_dir = -1;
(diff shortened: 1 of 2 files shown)


clydebarrow (changes requested):
* Using `InterruptLock` _inside_ an interrupt is pointless;
* The warnings about ++ are probably spurious in this context since on any of the target chips the increment should compile to an atomic operation.
* The `volatile` qualifier is important since it alerts the compiler that the value may change unexpectedly.

I would suggest reverting the changes and simply suppressing the warning with a pragma.

The other option would be to use `std::atomic` or similar but I don't know if that is supported across the platforms, and would be a performance hit.

There are discussions on the net about the inappropriateness of recent C++ changes deprecating compound operations on volatile objects, with a lack of appreciation for their use in embedded applications.

clydebarrow:
Alternatively (and more cleanly) load the value +1 into an auto variable then store it back. The load and store are each guaranteed to be atomic and will avoid the warning.
```
    if (arg->counter < arg->max_value) {
      auto x = arg->counter +1;
      arg->counter = x;
    }
```

edwardtfn:
Ok, I will look at that.
It should be the same for #7954 as that is exactly the same case. I will review both later today or tomorrow morning. 😉

edwardtfn:
Done!
On both this one and also #7954.

clydebarrow (approved):

PR #7901, opened 2024-12-01
[ble_client, bluetooth_proxy, esp32_ble_client, esp32_ble_tracker] fix ble proxy stop working

tomaszduda23:
# What does this implement/fix?

1. Added more logs to show what is going on
2. Create scan reboot timeout earlier to reboot device if ble stack fail to start ble scan
3. Init a few values
4. Mark ble client as failed if error in registration callback https://github.com/esphome/issues/issues/6553
5. Includes https://github.com/esphome/esphome/pull/7960

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- https://github.com/esphome/issues/issues/5119
- https://github.com/esphome/issues/issues/6270

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #7960, opened 2024-12-14
[esp32_ble] do not skip events if queue is blocked

tomaszduda23:

Do not skip events if queue is blocked. Before lock is taken with timeout. If semaphore is already taken it may happen that event is not added to queue.

## Types of changes

- [ ] Other

- https://github.com/esphome/issues/issues/5119
- https://github.com/esphome/issues/issues/6270

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/esp32_ble/queue.h
+++ b/esphome/components/esp32_ble/queue.h
@@ -26,6 +26,7 @@ template<class T> class Queue {
   void push(T *element) {
     if (element == nullptr)
       return;
+    // It is not called from main loop. Thus it won't block main thread.
     xSemaphoreTake(m_, portMAX_DELAY);
     q_.push(element);
     xSemaphoreGive(m_);
--- a/esphome/components/esp32_ble/queue.h
+++ b/esphome/components/esp32_ble/queue.h
@@ -26,10 +26,9 @@ template<class T> class Queue {
   void push(T *element) {
     if (element == nullptr)
       return;
-    if (xSemaphoreTake(m_, portMAX_DELAY)) {
-      q_.push(element);
-      xSemaphoreGive(m_);
-    }
+    xSemaphoreTake(m_, portMAX_DELAY);
+    q_.push(element);
+    xSemaphoreGive(m_);
   }
 
   T *pop() {
--- a/esphome/components/esp32_ble/queue.h
+++ b/esphome/components/esp32_ble/queue.h
@@ -26,7 +26,7 @@ template<class T> class Queue {
   void push(T *element) {
     if (element == nullptr)
       return;
-    if (xSemaphoreTake(m_, 5L / portTICK_PERIOD_MS)) {
+    if (xSemaphoreTake(m_, portMAX_DELAY)) {
       q_.push(element);
       xSemaphoreGive(m_);
     }


tomaszduda23:
@fightforlife please test it if possible.

fightforlife:
I am running this PR now on my main Bluetooth proxy. 
Will report when it fails. (Stops the scan)
If not I will comment with success in a few days.

fightforlife:
It happened again, the ESP32 rebooted and the BLE scanning is not running.
Currently I have no network loggin active. I have to find a way to do this without MQTT.

tomaszduda23:
Could you put your full config here? There is another issue https://github.com/esphome/issues/issues/6553. The error is visible only before wifi connection. It is present in serial log.

esphome/components/esp32_ble/queue.h
```diff
@@ -26,10 +26,9 @@ template<class T> class Queue {
   void push(T *element) {
     if (element == nullptr)
       return;
-    if (xSemaphoreTake(m_, 5L / portTICK_PERIOD_MS)) {
-      q_.push(element);
-      xSemaphoreGive(m_);
-    }
+    xSemaphoreTake(m_, portMAX_DELAY);
+    q_.push(element);
```
bdraco, line 31: If we don't hold the semaphore, I don't think its going to be thread-safe to push on to the queue.

We should add some type of logging if we can't obtain the semaphore, and maybe increase the timeout if needed.

tomaszduda23: It should never rich the queue without holding semaphore.
```
specifying the block time as 
portMAX_DELAY
 will cause the task to block indefinitely (without a timeout).
```

bdraco: I haven't seen the underlying code, but I'm guessing it can't fail if its `portMAX_DELAY` https://www.freertos.org/Documentation/02-Kernel/04-API-references/10-Semaphore-and-Mutexes/12-xSemaphoreTake

Won't that block the thread forever if it can't get the lock?

tomaszduda23: esphome already use similar approach in another places https://github.com/esphome/esphome/blob/1a89aa8fbfb5828f6c3bbf72343b72b8ce581187/esphome/core/helpers.cpp#L651
I can think of only about memory corruption to block here forever. If it happens reboot is only recovery.

tomaszduda23: btw I didn't change pop since it is fine to get event in next iteration if semaphore fails. There is no reason to block main loop.

bdraco: I think that makes sense. Would you please add a comment that this isn't ever called in the main loop and will not block the main thread.

tomaszduda23: done

bdraco:
Pushed to my production. I'll leave it running for a day or two and check it

bdraco:
So far so good overnight.  Will let it run another day before coming back to this

bdraco (approved):

bdraco:
Everything worked fine overnight again. 

Leaving this for a few days to get another opinion in case there is a problem I don't see before merging.

bdraco:
Still going well.

bdraco:
https://github.com/esphome/esphome/pull/7901 got a second opinion look with this change it it so good to merge now

bdraco:
Thanks @tomaszduda23

PR #7972, opened 2024-12-17
actions: fix loop re-entry

### LINKED ISSUE #6534: while/repeat/wait_until will corrupt arguments when run concurrently
### The problem

The actions are instantiated once per definition but they store the trigger arguments in their class member ([example](https://github.com/esphome/esphome/blob/444e162c92c3148a14493f66d3de256679a6eedf/esphome/core/base_automation.h#L356)), which gets overridden, causing corruption when they are called multiple times.

2024.12.0-dev

pip

n/a

RP2040

### Board

Pico W

core

### YAML Config

```yaml
esphome:
  name: test-2040

rp2040:
  board: rpipicow

wifi:
  ssid: Ulai
  fast_connect: true

  # Optional manual IP
  manual_ip:
    static_ip: 192.168.1.91
    gateway: 192.168.1.1
    subnet: 255.255.255.0
    dns1: 192.168.1.1

logger:
  level: ERROR

ota:
  - platform: esphome

api:
  reboot_timeout: 0s

globals:
   - id: global_counter
     type: size_t

time:
  - platform: sntp
    id: sntp_time
    timezone: Etc/UTC
    servers:
      - 192.168.1.1

    on_time:
      - seconds: /10
        then:
          - http_request.get:
              url: !lambda |-
                return "https://httpbin.org/anything?x=" + std::to_string(id(global_counter)++);
              capture_response: true
              max_response_buffer_size: 4kB
              on_response:
                then:
                  - lambda: |-
                      const char *const TAG = "TEST";
                      
                      json::parse_json(body, [TAG](JsonObject root) -> bool {

                        JsonObject args = root["args"];
                        if (!args) {
                          ESP_LOGE(TAG, "no args");
                          return false;
                        }
                        const char *arg = args["x"];
                        ESP_LOGE(TAG, "got response for request #%s", arg);
                        return true;
                      });

                  - wait_until:
                      condition:
                        lambda: return (id(global_counter) % 5) == 0;

                  - lambda: |-
                      const char *const TAG = "TEST_AFTER_WAIT";
                      
                      json::parse_json(body, [TAG](JsonObject root) -> bool {

                        JsonObject args = root["args"];
                        if (!args) {
                          ESP_LOGE(TAG, "no args");
                          return false;
                        }
                        const char *arg = args["x"];
                        ESP_LOGE(TAG, "got response for request #%s", arg);
                        return true;
                      });


http_request:
  verify_ssl: false
```

```txt
[09:58:41][E][TEST:068]: got response for request #0
[09:58:51][E][TEST:068]: got response for request #1
[09:59:00][E][TEST:068]: got response for request #2
[09:59:10][E][TEST:068]: got response for request #3
[09:59:22][E][TEST:068]: got response for request #4
[09:59:22][E][TEST_AFTER_WAIT:089]: got response for request #4
[09:59:22][E][TEST_AFTER_WAIT:089]: got response for request #3
[09:59:22][E][TEST_AFTER_WAIT:089]: got response for request #3
[09:59:22][E][TEST_AFTER_WAIT:089]: got response for request #3
[09:59:22][E][TEST_AFTER_WAIT:089]: got response for request #3
```

### Additional information

_No response_

afflux:

The actions are instantiated once per definition but they store the trigger arguments in their class member ([example](https://github.com/esphome/esphome/blob/444e162c92c3148a14493f66d3de256679a6eedf/esphome/core/base_automation.h#L356)), which gets overridden, causing corruption when they are called multiple times.

Note that I'm removing the stored arguments in the `while` and `repeat` actions because they still have the action arguments available in the call stack and there is no code path in the code base that can do any changes to them.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6534

- N/A

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esphome:
  name: test-2040

rp2040:
  board: rpipicow

wifi:
  ssid: !secret wifi_ssid
  fast_connect: true

  # Optional manual IP
  manual_ip:
    static_ip: 192.168.1.91
    gateway: 192.168.1.1
    subnet: 255.255.255.0
    dns1: 192.168.1.1

logger:
  level: ERROR

ota:
  - platform: esphome

api:
  reboot_timeout: 0s

globals:
   - id: global_counter
     type: size_t

script:
  - id: handle_response
    mode: parallel
    parameters:
      body: string
    then:
      - lambda: |-
          const char *const TAG = "TEST";

          json::parse_json(body, [TAG](JsonObject root) -> bool {

            JsonObject args = root["args"];
            if (!args) {
              ESP_LOGE(TAG, "no args");
              return false;
            }

            const char *arg = args["x"];
            ESP_LOGE(TAG, "got response for request #%s", arg);
            return true;
          });

      - wait_until:
          condition:
            lambda: return (id(global_counter) % 5) == 0;

      - lambda: |-
          const char *const TAG = "TEST_AFTER_WAIT";

          json::parse_json(body, [TAG](JsonObject root) -> bool {

            JsonObject args = root["args"];
            if (!args) {
              ESP_LOGE(TAG, "no args");
              return false;
            }
            const char *arg = args["x"];
            ESP_LOGE(TAG, "got response for request #%s", arg);
            return true;
          });

time:
  - platform: sntp
    id: sntp_time
    timezone: Etc/UTC
    servers:
      - 192.168.1.1

    on_time:
      - seconds: /10
        then:
          - http_request.get:
              url: !lambda |-

                return "https://httpbin.org/anything?x=" + std::to_string(id(global_counter)++);
              capture_response: true
              max_response_buffer_size: 4kB
              on_response:
                then:
                  - lambda: id(handle_response)->execute(body);

http_request:
  verify_ssl: false

```

## Checklist:


afflux:
still relevant

esphome/components/script/script.h
```diff
@@ -286,15 +286,19 @@ template<class C, typename... Ts> class ScriptWaitAction : public Action<Ts...>,
...
+    }
   }
 
   void play(Ts... x) override { /* ignore - see play_complex */
   }
 
  protected:
   C *script_;
-  std::tuple<Ts...> var_{};
+  std::queue<std::function<void()>> play_queue_;
```
bdraco: std::queue is backed by a std::deque which allocates 512 byte chunks.  With many script this could run an esp8266 out of RAM quickly.

afflux: what's your suggested alternative? std::list?

bdraco: `std::list` probably makes the most sense.

  | Approach              | Empty (never used) | Empty (after 1st use) | 1 element | 3 elements | Max capacity   |
  |-----------------------|--------------------|-----------------------|-----------|------------|----------------|
  | std::queue<T> (deque) | 24-512 bytes*      | 512 bytes ❌           | 512 bytes | 512 bytes  | ~512 per chunk |
  | StaticVector<T, 8>    | 260 bytes          | 260 bytes             | 260 bytes | 260 bytes  | 8 (fixed)      |
  | FixedVector<T>        | 12 bytes†          | 260 bytes‡            | 260 bytes | 260 bytes  | Set at init()  |
  | Ring buffer (lazy)    | 20 bytes           | 260 bytes‡            | 260 bytes | 260 bytes  | Set at init()  |
  | std::list<T> ✅        | 14 bytes           | 14 bytes              | 54 bytes  | 134 bytes  | Unbounded      |

Note that my table might not be perfectly accurate

bdraco: std::queue only starts to look good at large numbers..

bdraco: Also need to handle the unbounded grow. It needs a reasonable limit

bdraco (changes requested):
Thanks for the fix. The core approach looks correct once the memory issue is addressed.  Limit the fix to scripts that can run in parallel so we don't pay the memory cost unless its parrallel.

This PR changes `WhileAction`, `RepeatAction`, `WaitUntilAction`, and `ScriptWaitAction`. These are used heavily across all ESPHome configs. We need integration tests before merge.

Suggest splitting into two PRs:

**First PR:** Add `tests/integration/test_action_concurrent_reentry.py`
- Reproduce the parameter corruption from issue [#6534](https://github.com/esphome/issues/issues/6534)
- Parallel script with parameters
- wait_until that suspends execution
- Multiple concurrent invocations
- Assert each gets correct parameters
- Mark `@pytest.mark.xfail(reason="https://github.com/esphome/issues/issues/6534")`

**Second PR:** Apply this fix, remove the xfail marker, verify test passes.

This gives us regression coverage and proves the fix works.

esphome/components/script/script.h
afflux: How did you generate these numbers? How does std::forward_list compare? From its implementation, I see no reason for it to have a bigger cost than storing the default-initialized std::tuple before the change.

afflux:
> Limit the fix to scripts that can run in parallel so we don't pay the memory cost unless its parrallel.

On the `Action` level, I don't see a simple way to determine if the script or automation can run in parallel or not. I expect this to require a major refactoring, which I am not comfortable doing. In theory, the std::forward_list based approach should be cheap enough - let me know what you get from re-running the numbers.

> Also need to handle the unbounded grow. It needs a reasonable limit

Not sure how to handle this gracefully. My best idea is to check `Automation::num_running()` during trigger execution and reject new triggers when above a hardcoded maximum, but that is potentially a long walk through the ActionList for every trigger, and it's not clear to me, what a reasonable maximum would be.

> We need integration tests before merge.

done, see #11652

bdraco:
> > Also need to handle the unbounded grow. It needs a reasonable limit
> 
> Not sure how to handle this gracefully. My best idea is to check `Automation::num_running()` during trigger execution and reject new triggers when above a hardcoded maximum, but that is potentially a long walk through the ActionList for every trigger, and it's not clear to me, what a reasonable maximum would be.

I think we can punt on this for now as its also a problem for number of executions. I recently fixed this in #11308 for queued but didn't have the time to do parallel as well.

bdraco:
> > We need integration tests before merge.
> 
> done, see #11652

Thanks!

bdraco:
> > Limit the fix to scripts that can run in parallel so we don't pay the memory cost unless its parrallel.
> 
> On the `Action` level, I don't see a simple way to determine if the script or automation can run in parallel or not. I expect this to require a major refactoring, which I am not comfortable doing. In theory, the std::forward_list based approach should be cheap enough - let me know what you get from re-running the numbers.

We are going to need to come up with a way to split it. I'll think about that.  Maybe thats template classes with c++20 concepts or separate classes. It should be possible to figure out at Python codegen time separate classes, or with c++20 concepts at compile time if we can't figure it out at codegen time.

Downside to separate classes is 2x the bug fix surface. Templates are probably the best option

bdraco:
Separate classes might be the way to go. I started using templates and concepts to have the same `WaitUntilAction` and it became way too complex

afflux (commented):
What if the `script.wait_until` action is used in an automation trigger? Wouldn't this also be a potentially parallel context?

Edit: sorry, new GitHub UI is confusing me - this comment was supposed to go here:
https://github.com/esphome/esphome/pull/7972/files#diff-8fb488fbb00c19400b4b7888a5ffb32f83349ce9607e982f05717db5317605b7R54-R57

bdraco:
> What if the `script.wait_until` action is used in an automation trigger? Wouldn't this also be a potentially parallel context?

Than we have a correctness problem which is a good argument to always use the parallel versions.

esphome/components/script/script.h
bdraco: I'll check `std::forward_list`. They are a pain to benchmark because I have to write custom services, compile, upload to the device and than run

bdraco: Just thinking about it, should use 50% of the overhead of `std::list` so probably not worth benchmarking as its clearly the better choice

bdraco:
> > What if the `script.wait_until` action is used in an automation trigger? Wouldn't this also be a potentially parallel context?
> 
> Than we have a correctness problem which is a good argument to always use the parallel versions.

Now we need tests for concurrent execution of wait_until and script.wait actions

bdraco:
I think this is good to go. Let's see if the bot can find anything we missed

esphome/core/base_automation.h
```diff
@@ -268,32 +269,30 @@ template<typename... Ts> class WhileAction : public Action<Ts...> {
   void add_then(const std::initializer_list<Action<Ts...> *> &actions) {
     this->then_.add_actions(actions);
     this->then_.add_action(new LambdaAction<Ts...>([this](Ts... x) {
-      if (this->num_running_ > 0 && this->condition_->check_tuple(this->var_)) {
+      if (this->num_running_ > 0 && this->condition_->check(x...)) {
```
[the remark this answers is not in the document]
bdraco: 4486bdf77d1e971ed356d459ee83958b522535fa

esphome/components/script/script.h
```diff
@@ -286,15 +303,30 @@ template<class C, typename... Ts> class ScriptWaitAction : public Action<Ts...>,
...
       return;
 
-    this->play_next_tuple_(this->var_);
+    while (!this->param_queue_.empty()) {
+      auto &params = this->param_queue_.front();
+      this->play_next_tuple_(params, typename gens<sizeof...(Ts)>::type());
+      this->param_queue_.pop_front();
+    }
+    // Queue is now empty - disable loop until next play_complex
+    this->disable_loop();
```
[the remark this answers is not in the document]
bdraco: I don't think its correct. It should work either way. The question is if we want to process them all in one loop iteration or we should do one per loop to avoid blocking the loop for too long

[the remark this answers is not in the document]
bdraco: The only way we would care is if someone misconfigures this and we end up with a giant queue.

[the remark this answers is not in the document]
bdraco: I think we exhaust the memory on the device before this is a problem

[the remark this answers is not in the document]
bdraco: Yeah we are going to OOM before this is a problem

esphome/core/base_automation.h
```diff
@@ -359,39 +367,57 @@ template<typename... Ts> class WaitUntilAction : public Action<Ts...>, public Co
...
+      auto expired = timeout && (now - start) >= *timeout;
 
-    this->play_next_tuple_(this->var_);
+      if (!expired && !this->condition_->check_tuple(var)) {
+        return false;
+      }
+
+      this->play_next_tuple_(var);
+      return true;
+    });
```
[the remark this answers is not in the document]
bdraco: Same feedback as https://github.com/esphome/esphome/pull/7972/#discussion_r2485023962

esphome/components/script/script.h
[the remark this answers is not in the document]
afflux: It's not even correct:

1. num_running_ is incremented in play_complex(), not decremented
2. It will be decremented in play_next(), which is called exactly as many times, as play_complex() was called

bdraco (approved):
Thanks @afflux 

Good to go on my end if I didn't break anything with the refactoring

afflux:
Thank you @bdraco!

PR #7978, opened 2024-12-18
[helpers] Provide calls to get free heap and largest available block.

clydebarrow:

## Types of changes

- [x] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [x] BK72xx
- [x] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/core/helpers.h
+++ b/esphome/core/helpers.h
@@ -11,6 +11,14 @@
 
 #include "esphome/core/optional.h"
 
+#ifdef USE_ESP8266
+#include <Esp.h>
+#endif
+
+#ifdef USE_RP2040
+#include <Arduino.h>
+#endif
+
 #ifdef USE_ESP32
 #include <esp_heap_caps.h>
 #endif
@@ -684,7 +692,12 @@ template<class T> class RAMAllocator {
   };
 
   RAMAllocator() = default;
-  RAMAllocator(uint8_t flags) : flags_{flags} {}
+  RAMAllocator(uint8_t flags) {
+    // default is both external and internal
+    flags &= ALLOC_INTERNAL | ALLOC_EXTERNAL;
+    if (flags != 0)
+      this->flags_ = flags;
+  }
   template<class U> constexpr RAMAllocator(const RAMAllocator<U> &other) : flags_{other.flags_} {}
 
   T *allocate(size_t n) {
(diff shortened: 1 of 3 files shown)


esphome/core/helpers.h
```diff
@@ -710,8 +721,46 @@ template<class T> class RAMAllocator {
...
+        this->flags_ & ALLOC_INTERNAL ? heap_caps_get_free_size(MALLOC_CAP_8BIT | MALLOC_CAP_INTERNAL) : 0;
+    auto max_external =
+        this->flags_ & ALLOC_EXTERNAL ? heap_caps_get_free_size(MALLOC_CAP_8BIT | MALLOC_CAP_SPIRAM) : 0;
+    return max_internal + max_external;
+#elif defined(USE_RP2040)
+    return ::rp2040.getFreeHeap();
+#elif defined(USE_LIBRETINY)
+    return lt_heap_get_free();
+#else
+    return 100000;
```
guillempages, line 741: I'm not sure what will actually cause the else to be compiled; but maybe return some number where it's clear that it's not a right value? E.g. a 0, 1 or 42?

esphome/core/helpers.h
```diff
@@ -684,20 +692,23 @@ template<class T> class RAMAllocator {
...
       ptr = static_cast<T *>(heap_caps_malloc(size, MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT));
     }
-    // Fallback to internal allocation if explicitely requested or no flag is specified
-    if (ptr == nullptr && ((this->flags_ & Flags::ALLOC_INTERNAL) || (this->flags_ & Flags::ALLOC_EXTERNAL) == 0)) {
-      ptr = static_cast<T *>(malloc(size));  // NOLINT(cppcoreguidelines-owning-memory,cppcoreguidelines-no-malloc)
+    if (ptr == nullptr && this->flags_ & Flags::ALLOC_INTERNAL) {
+      ptr = static_cast<T *>(heap_caps_malloc(size, MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT));
     }
 #else
     // Ignore ALLOC_EXTERNAL/ALLOC_INTERNAL flags if external allocation is not supported
```
guillempages, line 714: Does the comment actually fit the code? The #ifdef is about ESP32, and not really about external allocation, is it?

guillempages (commented):
Wasn't there already another PR implementing the INTERNAL/EXTERNAL Flag meaning changes a while ago?

esphome/core/helpers.h
clydebarrow: Host platform. No way to get an actual value, it's just something large. MAX_INT is probably too large.

esphome/core/helpers.h
clydebarrow: ESP32 is the only one that supports external allocation, so it is, yes. There is no #define for external allocation as such.

clydebarrow:
> Wasn't there already another PR implementing the INTERNAL/EXTERNAL Flag meaning changes a while ago?

yes. This doesn't change the meaning, just adds the query calls. But the logic was a bit convoluted before - both flags set meant the same as no flags set, so that correspondence is moved to the constructor rather than having to have an obscure two-part test in multiple places.

jesserockz (approved):

PR #7770, opened 2024-11-14
[esp32_rmt] Updates for IDF 5+

swoboda1337:
# What does this implement/fix?

Newer versions of the framework have a new RMT driver:

    The legacy RMT driver is deprecated, please use driver/rmt_tx.h and/or driver/rmt_rx.h

One big difference is the user cannot choose the RMT channel. Will simplify the esphome code as there is no need to coordinate between components as its all done in the driver. 

Newer driver has a number of other advantages:
- Proper support for rx/tx on a single gpio, add "one_wire: true" in remote_transmitter to use this feature. 
- Previous driver would drop most rx data in high noise environments rendering it useless, new driver performs very well. 
- DMA support on newer chips like s3: add "with_dma: true" in remote_transmitter or remote_receiver. 
- Boards like C3 are limited to a single hw memory buffer, the previous driver can only read a max of 128 values. Even with all 8 buffers on other chips the max length was only 1024. The newer driver on newer chips reads continuously in a ping-pong fashion. To use this set max_length in remote_receiver to the desired length.

Remote code and led code have been updated to support both versions of the RMT driver. Arduino doesn't support the newer drivers yet. We have to support both drivers for a while. 

Note neopixelbus also used RMT but it doesn't work with IDF anyways so it wont be a problem.  

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- fixes esphome/issues#5357
- related #7706

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#4480

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
latest:

    external_components:
      - source: github://pr#7770
        components: [ remote_base, remote_receiver, remote_transmitter, esp32_rmt, esp32_rmt_led_strip ]
        refresh: 5min

if you are not on the lastest dev and get c6 config errors:

    external_components:
      - source: github://esphome/esphome@94befe126bc952785994ddf8dcfd15e9c3c3ff23
[...]

PR #8002, opened 2024-12-23
[esp32_rmt] IDF 5+ update fixes

### LINKED ISSUE #6589: (how) Does  one_wire work with the esp32_rmt.use_new_rmt_driver (remote_transmitter component)?
### The problem

Hi (@swoboda1337),
Is there documentation/an example/working case of how to use the one_wire with remote_transmitter with the new ESP-IDF platform for ESP32?  On the ESP8266, and ESP-32 with Arduino  and older RMT in older IDF, I have been able to use a single GPIO to transmit and receive on the TI-CC1101 using Radiolib (like [here](https://github.com/juanboro/esphome-radiolib-cc1101/blob/main/esp32-cc1101.yaml).  With the latest ESP-IDF platform release included with esphome -I can only get receive or transmit alone - not both.   With both enabled, and prioritizing receive component ahead of transmit, they can co-exist with receive working, but transmit will only transmit an idle carrier.  

2025.1.0-dev

pip

NA

ESP32-IDF

### Board

esp32dev

remote_receiver

### YAML Config

```yaml
https://github.com/juanboro/esphome-radiolib-cc1101/blob/main/esp32-cc1101.yaml

I test with: https://github.com/juanboro/esphome-radiolib-cc1101/blob/main/esp32-cc1101-udp-example.yaml
```

_No response_

### Additional information

_No response_

swoboda1337:

Fix a few issues:
1. Remove the digital_read call from remote receiver as setup is not run on that pin and the warning is confusing when RF is being used. 
2. Don't clear warnings after rmt_tx_wait_all_done as it may hide errors in rmt_transmit.
3. Fix inverted mode in remote transmitter on IDF 5+
4. When one_wire mode is enabled the gpio is both an input and an output in open drain mode. The pin is actively driven low or passively pulled high by pull up. After a transmit is complete it is important to release the pin by setting the level to high in order to not actively drive the pin low and fight the receiver. By default eot_level will now be true if one_wire is enabled. This may be inconvenient though as the pin will be high before, after or between transmits. Provide a way to override eot_level and provide a function to manually set the level. 

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6589
- related https://github.com/esphome/esphome/pull/7770 

- esphome/esphome-docs#4531

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Here is an example of how one_wire mode could be used while still keeping the pin low before/after/between transmits. The pin is manually driven/released between rf mode changes. 

```yaml
remote_receiver:
  id: rx_id
  pin:
    number: GPIO32
    allow_other_uses: true
  filter: 100us
  idle: 800us
  rmt_symbols: 256
  receive_symbols: 256
  filter_symbols: 50

remote_transmitter:
  id: tx_id
  pin:
    number: GPIO32
    allow_other_uses: true
  one_wire: true
  eot_level: false
  carrier_duty_percent: 100%
  on_transmit:
    then:
      - rfchip.set_mode_standby
      - lambda: 'id(tx_id)->digital_write(false);'
      - rfchip.set_mode_tx
  on_complete:
    then:
      - rfchip.set_mode_standby
      - lambda: 'id(tx_id)->digital_write(true);'
      - rfchip.set_mode_rx
```

## Checklist:

### DIFF
--- a/esphome/components/remote_receiver/remote_receiver_esp32.cpp
+++ b/esphome/components/remote_receiver/remote_receiver_esp32.cpp
@@ -154,16 +154,16 @@ void RemoteReceiverComponent::setup() {
 void RemoteReceiverComponent::dump_config() {
   ESP_LOGCONFIG(TAG, "Remote Receiver:");
   LOG_PIN("  Pin: ", this->pin_);
-  if (this->pin_->digital_read()) {
-    ESP_LOGW(TAG, "Remote Receiver Signal starts with a HIGH value. Usually this means you have to "
-                  "invert the signal using 'inverted: True' in the pin schema!");
-  }
 #if ESP_IDF_VERSION_MAJOR >= 5
   ESP_LOGCONFIG(TAG, "  Clock resolution: %" PRIu32 " hz", this->clock_resolution_);
   ESP_LOGCONFIG(TAG, "  RMT symbols: %" PRIu32, this->rmt_symbols_);
   ESP_LOGCONFIG(TAG, "  Filter symbols: %" PRIu32, this->filter_symbols_);
   ESP_LOGCONFIG(TAG, "  Receive symbols: %" PRIu32, this->receive_symbols_);
 #else
+  if (this->pin_->digital_read()) {
+    ESP_LOGW(TAG, "Remote Receiver Signal starts with a HIGH value. Usually this means you have to "
+                  "invert the signal using 'inverted: True' in the pin schema!");
+  }
   ESP_LOGCONFIG(TAG, "  Channel: %d", this->channel_);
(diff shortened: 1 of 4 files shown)


swoboda1337:
With this change the gpio will start high and finish high which is not ideal for many applications. Might need to provide more options.

swoboda1337:
@kbx81 I know you are probably off but there are a few small fixes needed for the rmt code. No rush as long as they get fixed before the next release. Thanks!

kbx81 (approved):
LGTM -- did a quick smoke test on my hardware here and it still appears to be working just fine. Thanks! 🍻

PR #8008, opened 2024-12-24
[psram] Improve total PSRAM display in logs by using rounded KB values

edwardtfn:

When displaying the total PSRAM size at the `dump_config`, round to the nearest KB value if within 0.05KB threshold. Otherwise show in bytes. This makes the logs clearer and more precise.

In a real case example, where `heap_caps_get_total_size(MALLOC_CAP_SPIRAM)` returns 2097151 bytes:

- Before: `Size: 2047 KB` (off by 1023 bytes)
- After: `Size: 2048 KB` (off by 1 byte)

## Types of changes

- [ ] Other

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/psram/psram.cpp
+++ b/esphome/components/psram/psram.cpp
@@ -27,7 +27,7 @@ void PsramComponent::dump_config() {
     if (abs(round(psram_total_size_kb) - psram_total_size_kb) < 0.05f) {
       ESP_LOGCONFIG(TAG, "  Size: %.0f KB", psram_total_size_kb);
     } else {
-      ESP_LOGCONFIG(TAG, "  Size: %.0f bytes", (float)psram_total_size_bytes);
+      ESP_LOGCONFIG(TAG, "  Size: %.0f bytes", (float) psram_total_size_bytes);
     }
   }
 #endif
--- a/esphome/components/psram/psram.cpp
+++ b/esphome/components/psram/psram.cpp
@@ -27,7 +27,7 @@ void PsramComponent::dump_config() {
     if (abs(std::round(psram_total_size_kb) - psram_total_size_kb) < 0.05f) {
       ESP_LOGCONFIG(TAG, "  Size: %.0f KB", psram_total_size_kb);
     } else {
-      ESP_LOGCONFIG(TAG, "  Size: %.0f bytes", (float) psram_total_size_bytes);
+      ESP_LOGCONFIG(TAG, "  Size: %zu bytes", psram_total_size_bytes);
     }
   }
 #endif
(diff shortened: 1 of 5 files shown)


DjordjeMandic (approved):

DjordjeMandic:
Technically, "KB" refers to 1000 bytes (base 10) in the SI system, while 1024 bytes is correctly termed "KiB" (base 2). For clarity, especially in user-facing outputs, it might be better to use "KiB" if precision matters. That said, what standard is actually followed for binary prefixes in embedded systems—IEC or JEDEC? It’s worth considering since JEDEC often sticks with "KB" for 1024 bytes, which might explain the choice here.

edwardtfn:
Well, you have a point...
I just kept the same units as it was before... My intention here was to improve the number itself, so I haven't thought about the units that much.
For consistency, I think it should be similar to how units are shown elsewhere in ESPHome (including documents) and, from memory, I can remember `flash_size` also using `MB` as unit, probably with a power of two meaning...
With that said, my suggestion woukd be to keep this PR as it is and possibly create another PR to review units, which should cover other components if needed.

DjordjeMandic:
Your PR looks fine to me, I just mentioned that because I'm confused too.

esphome/components/psram/psram.cpp
```diff
@@ -21,7 +21,14 @@ void PsramComponent::dump_config() {
...
 #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4, 1, 0)
   if (available) {
-    ESP_LOGCONFIG(TAG, "  Size: %d KB", heap_caps_get_total_size(MALLOC_CAP_SPIRAM) / 1024);
+    const size_t psram_total_size_bytes = heap_caps_get_total_size(MALLOC_CAP_SPIRAM);
+    const float psram_total_size_kb = psram_total_size_bytes / 1024.0f;
+
+    if (abs(std::round(psram_total_size_kb) - psram_total_size_kb) < 0.05f) {
+      ESP_LOGCONFIG(TAG, "  Size: %.0f KB", psram_total_size_kb);
+    } else {
+      ESP_LOGCONFIG(TAG, "  Size: %.0f bytes", (float) psram_total_size_bytes);
```
DjordjeMandic:
```suggestion
      ESP_LOGCONFIG(TAG, "  Size: %zu bytes", psram_total_size_bytes);
```

Because `psram_total_size_bytes` is `size_t`. Should be faster.

DjordjeMandic (changes requested):

DjordjeMandic (approved):

jesserockz (approved):

PR #8015, opened 2024-12-26
[remote_base] NEC Protocol implementation rewrite (NEC1/NEC2 repeat code support)

### LINKED ISSUE #6603: NEC protocol "command_repeats" implementation is incorrect
### The problem

Greetings, dear fellow esphome developers!
Thank you for the great tool that you all build. I think I found a bug in NEC protocol implementation, and I hope I was able to provide clear and complete issue description below.

According to the documentation of NEC protocol ([a good explanation can be found here https](https://techdocs.altium.com/display/FPGA/NEC%2bInfrared%2bTransmission%2bProtocol)) repeats of the command are not exactly the command itself, but a specific simple signal that is sent after the command. This specific signal being sent dictates the controlled device to repeat the command itself.

Current implementation in `esphome/components/remote_base/nec_protocol.cpp` ([link](https://github.com/esphome/esphome/blob/dev/esphome/components/remote_base/nec_protocol.cpp)) instead of it just repeats the bytes from the "command" part of the signal.

The same goes to the decoder, it can not decode more than one command because it awaits for the command itself to be repeated, not the specific "repeat" type of signal.

To encode repeated signal you need to add a pause (low signal) for 40.5ms after the address+command transmission, and then start sending "repeat codes".
Repeat code itself is `HEADER_HIGH_US` as 1 or "high" signal, then a space with a duration 1/2 of `HEADER_LOW_US` of 0 or "low", one `BIT_HIGH_US` as 1 or "high", and the last part is a long pause of 96.1875ms.

Each frame is 108ms.
Here is a picture to illustrate it
image

I myself don't know C++ good enough to resolve this issues in an elegant and laconic way, but I suggest this issue could be marked as "ready to code" or something, and I'm willing to test the exact implementation with my devices to prove that solution works correctly.

Also, there is a recommendation to set carrier frequency to 38222Hz, which should be more reliable.

2024.12.2 

Docker

Any

BK72XX

### Board

CBU

remote_transmitter.transmit_nec

### YAML Config

Currently I have a problem with Bluetooth reset command, which is "repeating" the Bluetooth button for 3 seconds.

```yaml
   - platform: template
     name: "Klipsh set to Bluetooth"
     on_press:
           address: 0x7F80
           command: 0xF20D
           command_repeats: 30
```

This is the output of the RAW decoder which reads repeats from the remote:
```txt
[I][remote.raw:028]: Received Raw: 9060, -4525, 549, -562, 563, -593, 531, -594, 562, -562, 563, -562, 562, -563, 561, -563, 562, -1687, 563, -1686, 563, -1687, 531, -1707, 542, -1687, 562, -1687, 562, -1687, 563, -1686, 563, -562, 562, -1687, 562, -563, 562, -1690, 559, 
[I][remote.raw:041]:   -1704, 545, -562, 563, -562, 562, -563, 562, -562, 563, -562, 562, -1687, 562, -563, 562, -562, 563, -1687, 562, -1690, 559, -1690, 559, -1687, 562
[I][remote.raw:041]: Received Raw: 9042, -2275, 553
[I][remote.raw:041]: Received Raw: 9044, -2246, 571
[I][remote.raw:041]: Received Raw: 9050, -2279, 531
[I][remote.raw:041]: Received Raw: 9049, -2279, 563
[I][remote.raw:041]: Received Raw: 7012, -4284, 556
[I][remote.raw:041]: Received Raw: 9044, -2249, 559
[I][remote.raw:041]: Received Raw: 9050, -2251, 559
[I][remote.raw:041]: Received Raw: 9056, -2282, 556
...
...
```

### Additional information

Just thank you for the great project, it is a real blessing to have it.

DjordjeMandic:

This PR updates the NEC implementation to achieve **almost**  full compliance with the official NEC specification. Added support for both NEC1 and NEC2. NEC2 is same as NEC1 but whole frame is sent again instead of NEC repeat code, see [here](http://www.hifi-remote.com/wiki/index.php/NEC) for more info.

Key improvements include:

1. **Corrected Timing Constants**  
   - Adjusts both frame and repeat code durations to match NEC timing more precisely.  

2. **Refined Encoding & Decoding Logic**  
   - Properly encodes and decodes frames and repeat codes. 
   - Introduces a dedicated enum class (`NECCodeType`) to simplify logic by clearly distinguishing between NEC1-style and NEC2-style sending behaviors, as well as internal frame and repeat types used during encoding.
   - Separates repeat codes from main frame decoding due to the low signal idle timeout (`10ms`), which makes processing impractical. This ensures that repeat codes can be detected and handled correctly in real-time without interfering with the main application.

3. **NEC Binary Sensor Enhancement**  
   - Adds support for long button presses (the sensor remains “on” while a button is held).  
   - Uses a 130 ms timeout to accommodate small variances in repeat code intervals (typically ~108 ms).  (configurable in yaml)

4. **User-Facing Schema Updates**  
   - Removes the old `command_repeats` option in favour of `repeat`->`times` parameter. 
   - Invalidates `repeat`->`wait_time` option for NEC protocol. (internally handled)
   - By default, a new `type: NEC1` is sent (for burst or longer presses).  
   - A new `type: NEC2` option enables sending full frames as repeat codes.  
   - Older configurations remain supported (`type` defaults to `NEC1` if unspecified).  

**Known Caveat**  
- Sending many repeat codes in quick succession can trigger a WDT reset on some devices because IR transmission is blocking and does not feed the watchdog.
- Decoding repeat-only signals: users must handle them manually if they need more advanced receive behaviors than what the new binary sensor provides.

**Examples**  
- **Simple Frame Only**  
  ```yaml
  on_press:
        address: 0xBF40
        command: 0xF40B
        # type defaults to "NEC1"
  ```  
- **NEC1: Single frame with multiple repeats**  
  ```yaml
  on_press:
        address: 0xBF40
        command: 0xF40B
        repeat:
          times: 10  # About 1 second of button hold, frame + 10 repeats
        # `type` defaults to `NEC1`
  ```  
- **NEC2: Frame sent multiple times**  
  ```yaml
  on_press:
        address: 0xBF40
        command: 0xF40B
        repeat:
          times: 10  # About 1 second of button hold, frame + 10 repeats
        type: NEC2
  ```  
- **Binary Sensor**  
  ```yaml
  - platform: remote_receiver
    name: "KWORLD Remote KW Button"
    nec:
      address: 0x6B86
      command: 0xF30C
      timeout: 130ms
  ```  

**Breaking Change**  
- `command_repeats` has been removed. `repeat`->`times` replaces it
- `repeat`->`wait_time` is invalid. Handled internally.

Overall, these changes aim to make NEC transmissions and receptions more robust, accurate, and user‑friendly.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6603

- esphome/esphome-docs#4841

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
external_components:
  - source: github://pr#8015
    components: [ remote_base ]
    refresh: always

esphome:
  name: esp32-test
  friendly_name: ESP32-Test

esp32:
  board: esp32dev
  framework:
    type: esp-idf

web_server:
  port: 80
  version: 3
  local: True

number:
  - platform: template
    id: my_number
    name: "Repeat count"
    optimistic: true
    min_value: 1
    max_value: 100
    step: 1

switch:
- platform: template
  name: "NEC1/NEC2"
  id: "type_switch"
  optimistic: True

button:
- platform: template
  name: "TX KWORLD Remote KW Button"
  on_press:
        address: 0x6B86
        command: 0xF30C
        repeat:
          times: !lambda "return static_cast<uint16_t>(id(my_number).state);"
        type: !lambda "return id(type_switch).state ? esphome::remote_base::NECCodeType::NEC2_FRAME_WITH_REPEATS : esphome::remote_base::NECCodeType::NEC1_FRAME_WITH_REPEATS;"
- platform: template
  name: "TX LG TV OK Button"
  on_press:
        address: 0xFB04
        command: 0xBB44
        repeat:
          times: !lambda "return static_cast<uint16_t>(id(my_number).state);"
        type: !lambda "return id(type_switch).state ? esphome::remote_base::NECCodeType::NEC2_FRAME_WITH_REPEATS : esphome::remote_base::NECCodeType::NEC1_FRAME_WITH_REPEATS;"

remote_transmitter:
  pin: GPIO26
  carrier_duty_percent: 100%

binary_sensor:
  - platform: remote_receiver
    name: "KWORLD Remote KW Button"
    nec:
      address: 0x6B86
      command: 0xF30C
      timeout: 130ms
  - platform: remote_receiver
    name: "LG TV OK Button"
    nec:
      address: 0xFB04
      command: 0xBB44
      timeout: 130ms

remote_receiver:
  pin:
    number: GPIO25
    inverted: true
    mode:
      input: true
      pullup: true
  dump:
    - nec
    - raw

# Enable logging
logger:
  baud_rate: 500000
  level: VERBOSE
```

## Checklist:

### Tasks Checklist

- [x] **TODO**
  - [x] Finalize renaming `command_repeats` to `repeats` for consistency and clarity.
  - [x] Ensure all references to `command_repeats` in the codebase are updated.

- [x] **Testing**
  - [x] Verify functionality with encoding and decoding methods.
  - [x] Tested RX and binary sensor with NEC IR remote
  - [x] Test with **IrScrutinizer** for proper NEC protocol compliance.

- [x] **Documentation**
  - [x] Update the NEC action documentation to reflect changes in the struct.

---
# TX captured with PulseView
---
NEC1:
image
image

NEC2:
image


esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,93 +6,138 @@ namespace remote_base {
...
-  ESP_LOGD(TAG, "Sending NEC: address=0x%04X, command=0x%04X command_repeats=%d", data.address, data.command,
-           data.command_repeats);
+  ESP_LOGD(TAG, "Sending NEC: address=0x%04X, command=0x%04X, command_repeats=%d", 
+           data.address, data.command, data.command_repeats);
 
-  dst->reserve(2 + 32 + 32 * data.command_repeats + 2);
-  dst->set_carrier_frequency(38000);
+  // Reserve: AGC Header (2) + Address bits (32) + Command bits (32) + Final mark (2) + Repeat codes (4 per repeat)
+  dst->reserve(2 + 32 + 32 + 2 + data.command_repeats * 4);
+  dst->set_carrier_frequency(38222);
```
DjordjeMandic, line 36: I switched the carrier frequency from 38000 Hz to 38222 Hz because docs say it’s better for NEC protocol. But since 38000 Hz is the usual standard, I’m not sure if this is actually an improvement. Anyone have thoughts?

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,93 +6,138 @@ namespace remote_base {
...
+  for (uint16_t repeats = 0; repeats < data.command_repeats; repeats++) {
+    // AGC Repeat header (shorter version of the initial AGC header)
+    dst->item(HEADER_HIGH_US, HEADER_LOW_US / 2);  // Shortened AGC header
+    dst->mark(BIT_HIGH_US);  // Mark to complete the repeat code
+
+    // Add space after repeat code, except after the final repeat
+    if (repeats < data.command_repeats - 1) {
+      dst->space(SPACE_AGC_REPEAT_US);
+    }
+  }
```
DjordjeMandic:
```suggestion
if (data.command_repeats > 0) {
  // Space between message frame and AGC repeat code
  dst->space(SPACE_INTER_FRAME_US);

  // Send AGC Repeat Codes
  for (uint16_t repeats = 0; repeats < data.command_repeats; repeats++) {
    // AGC Repeat header (shorter version of the initial AGC header)
    dst->item(HEADER_HIGH_US, HEADER_LOW_US / 2);  // Shortened AGC header
    dst->mark(BIT_HIGH_US);  // Mark to complete the repeat code

    // Add space after repeat code, except after the final repeat
    if (repeats < data.command_repeats - 1) {
      dst->space(SPACE_AGC_REPEAT_US);
    }
  }
}
```

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,93 +6,138 @@ namespace remote_base {
...
 }
+
+// TODO: command_repeats -> repeats (repeat as repeat code, not whole message frame)
 optional<NECData> NECProtocol::decode(RemoteReceiveData src) {
   NECData data{
       .address = 0,
       .command = 0,
-      .command_repeats = 1,
+      .command_repeats = 0, // Start with 0, as the first frame is counted explicitly
```
DjordjeMandic: The initialization of .command_repeats = 0 makes sense because if a valid NECData is returned, it guarantees that the code has been received at least once (the main frame). Any additional repeat codes are counted as repeats beyond this first frame. This approach ensures clarity and avoids double-counting the initial transmission.

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,93 +6,138 @@ namespace remote_base {
...
+  // Check for inter-frame space (only if repeats are expected)
+  if (src.peek_space(SPACE_INTER_FRAME_US)) {
+    src.expect_space(SPACE_INTER_FRAME_US); // Consume inter-frame space
+  }
+
+  // Decode AGC Repeat Codes
+  while (src.peek_item(HEADER_HIGH_US, HEADER_LOW_US / 2)) {
+    // Validate the repeat header
+    if (!src.expect_item(HEADER_HIGH_US, HEADER_LOW_US / 2)) {
+      break;
```
DjordjeMandic: Breaking the loop instead of returning invalid NEC data here makes sense because a failure to validate a repeat header doesn’t necessarily invalidate the entire transmission. The main frame has already been successfully decoded, so the NEC data is still valid even if some repeat codes are malformed or missing. This approach ensures partial but meaningful data can still be returned.

esphome/const.py
```diff
@@ -153,6 +153,7 @@
 CONF_COLORS = "colors"
 CONF_COMMAND = "command"
 CONF_COMMAND_REPEATS = "command_repeats"
+# TODO: CONF_REPEATS = "repeats"
```
DjordjeMandic: The existing CONF_REPEAT works, but CONF_REPEATS feels more natural and aligns better with the concept of multiple repetitions. If there's no strict need to stick with the singular form, renaming it to CONF_REPEATS could improve clarity and consistency. Open to opinions on this!

esphome/components/remote_base/nec_protocol.h
```diff
@@ -8,7 +8,7 @@ namespace remote_base {
 struct NECData {
   uint16_t address;
   uint16_t command;
-  uint16_t command_repeats;
+  uint16_t command_repeats; // TODO: command_repeats -> repeats (repeat as repeat code, not whole message frame)
```
DjordjeMandic: Renaming `command_repeats` to repeats makes sense for simplicity and readability, especially if it refers specifically to repeat codes and not the entire message frame. The term "command_repeats" might be slightly ambiguous without context. Would a more descriptive name like `repeat_codes` or `repeats` be clearer?

DjordjeMandic (commented):
I need opinions

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,93 +6,138 @@ namespace remote_base {
...
+  }
 
+  // Decode 16-bit Address
   for (uint16_t mask = 1; mask; mask <<= 1) {
     if (src.expect_item(BIT_HIGH_US, BIT_ONE_LOW_US)) {
-      data.address |= mask;
+      data.address |= mask;  // Logic '1'
     } else if (src.expect_item(BIT_HIGH_US, BIT_ZERO_LOW_US)) {
-      data.address &= ~mask;
+      data.address &= ~mask; // Logic '0'
```
DjordjeMandic: The use of `&= ~mask` for logic '0' in the decoding process is redundant because the bit is already cleared by default (`.address = 0` at initialization). Omitting this step would simplify the code without changing its behavior. However, leaving it in can make the logic more explicit and easier to understand. Opinions on this?

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,93 +6,138 @@ namespace remote_base {
...
   }
 
+  // Decode 16-bit Command
   for (uint16_t mask = 1; mask; mask <<= 1) {
     if (src.expect_item(BIT_HIGH_US, BIT_ONE_LOW_US)) {
-      data.command |= mask;
+      data.command |= mask;  // Logic '1'
     } else if (src.expect_item(BIT_HIGH_US, BIT_ZERO_LOW_US)) {
-      data.command &= ~mask;
+      data.command &= ~mask; // Logic '0'
```
DjordjeMandic: Same here as above on line 90

esphome/components/remote_base/nec_protocol.cpp
GRbit: I can't be sure about NEC implementation, but I was testing signals with RAW "codec" and with a long signal (3 seconds of repeats) it didn't work until I've set carrier_frequency to 38222 (as far as I know, the default one is 38000 for raw signals too). Thus, I would consider this an important thing to have correctly working signals.

esphome/components/remote_base/nec_protocol.cpp
GRbit: Technically, I 100% agree with this solution and this approach.

From the backward compatibility perspective, that might not work for some users how likes to specifically define all the arguments for their commands. I'm a complete newbie to the project, so I don't if that's ok. Looking at the version number, 1 means to me that backward compatibility should not be broken. Do you think it might make sense to make an exception and do not issue any repeat codes if the command_repeats value is 1?

esphome/components/remote_base/nec_protocol.h
GRbit: I think the original idea behind the "command_" prefix was the differentiate this from the general "repeat" structure wich is common to many other type of signals, while this exact value should specify not the exact repeat, but the NEC codec type of repeats with repeat code.

For me `command_repeats` might be confusing, since we have `address` and `command` fields here, and `command_repeats` can be seen as repeats of the second part (that's how it was implemented, indeed), so I would be ok with renaming it with something like `repeat_codes`, or some other name.

Actually, introducing a new field `repeat_codes` might help with backward compatibility, the previous configs with `command_repeats: 1` will still issue the same 1 command, and if users need real NEC repeat codes, they can add the new field to their command.

DjordjeMandic: Backwards compability can be handled during code generation using python component module. In c/c++ there is no need for 2 variables for repeat count since only type of repeat is repeat code, I've removed repeating of command byte

esphome/components/remote_base/nec_protocol.cpp
GRbit: Since you want to pay this part some attention, I want to suggest another way to simplify this. I don't know how common this practice is for C, because it's important to be clear for all other developers, but in Go, I often do something like this:
```
if (data.command_repeats == 0) {
  return
}

dst->space(SPACE_INTER_FRAME_US);

// Send AGC Repeat Codes if requested
for (uint16_t repeats = 0; repeats < data.command_repeats; repeats++) {
    // AGC Repeat header (shorter version of the initial AGC header)
    dst->item(HEADER_HIGH_US, HEADER_LOW_US / 2);  // Shortened AGC header
    dst->mark(BIT_HIGH_US);  // Mark to complete the repeat code
    // Add space after repeat code, except after the final repeat
    if (repeats < data.command_repeats - 1) {
      dst->space(SPACE_AGC_REPEAT_US);
    }
}
```

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,93 +6,138 @@ namespace remote_base {
...
+// NEC IR Protocol
+// Protocol Reference: https://techdocs.altium.com/display/FPGA/NEC%2bInfrared%2bTransmission%2bProtocol
+
+// Timing constants in microseconds
+static const uint32_t HEADER_HIGH_US = 9000;        // AGC burst: 9ms HIGH
+static const uint32_t HEADER_LOW_US = 4500;         // AGC burst: 4.5ms LOW
+static const uint32_t BIT_HIGH_US = 562;            // Bit HIGH duration: 562.5µs
+static const uint32_t BIT_ONE_LOW_US = 1687;        // Logic '1': 562.5µs HIGH + 1687.5µs LOW
+static const uint32_t BIT_ZERO_LOW_US = 562;        // Logic '0': 562.5µs HIGH + 562.5µs LOW
+static const uint32_t SPACE_INTER_FRAME_US = 40000;  // Inter-frame space: 40ms
```
GRbit: According to protocol reference, the pause should be 40500. It might work with 40k, but the raw signals emulation that works for me has a pause of 40500 which is 108ms - 67.5ms

esphome/components/remote_base/nec_protocol.cpp
GRbit: I'm all thumbs in bit operations, but when I'm in doubts, and if we don't have performance constraints, my vote is always for something more clear.

esphome/const.py
GRbit: I like the idea of renaming `command_repeats`, since the actual protocol doesn't repeat commands. yet, from the user perspective, this name might be clear enough. Like "I have a command, and I want it to be repeated", that's it. And from developers perspective, it might help to see something that will tell you that you don't actually repeat the command, but send repeat codes. If it's possible to somehow hit the both goals of being clear to user and helping developer to understand it, that would be just great.

GRbit (commented):
Amazing work!

I especially liked part with partially correct NEC codes, I think it is very user-friendly solution.

I don't have anything else to add about the code, I think the last part where I can help is to test it with my devices to see if everything works correctly. Unfortunately, that'll take some time. I've flashed firmware from github code only once, currently I use espdashboard in docker, which makes the code more difficult to access and change.

Thank you again for very quick reaction to my issue and this nice solution.

DjordjeMandic:
For development and testing, you can use the VS Code dev container. Unfortunately, I’m working on a Windows 10-based server with insufficient free RAM to run a dev container. Instead, I make code changes and test them as a custom_component on my existing ESPHome container running in Home Assistant OS (HAOS). For testing, I use IrScrutinizer along with my custom hardware board. Since I am currently working on my final paper about IR communication, this serves as a great example of real-world usage, ensuring thorough and practical testing.

esphome/components/remote_base/nec_protocol.cpp
DjordjeMandic: > Notice from Figure 1 that it takes:
> 
> 27ms to transmit both the 16 bits for the address (address + inverse) and the 16 bits for the command (command + inverse). This comes from each of the 16 bit blocks ultimately containing eight '0's and eight '1's - giving (8 * 1.125ms) + (8 * 2.25ms).
> 67.5ms to fully transmit the message frame (**`discounting the final 562.5µs pulse burst that signifies the end of message`**).

```cpp
  // Final mark to end the message frame
  dst->mark(BIT_HIGH_US);
```
This makes it roughly around 68ms. So in order to achieve 108ms, 40ms pause is needed.

Edit: now I need to do more research.
image


image


There is no final marker at end of message frame on this photo with repeat codes. 

You're right, I need to add 40500ms when repeat codes are needed.

But: 

> If the key on the remote controller is kept depressed, a repeat code will be issued, typically around 40ms after the pulse burst that signified the end of the message. A repeat code will continue to be sent out at 108ms intervals, until the key is finally released.

Soo.... more research is needed for this part :) (dw, i will do it)

esphome/components/remote_base/nec_protocol.cpp
DjordjeMandic: You're right, return would be better.

esphome/components/remote_base/nec_protocol.cpp
DjordjeMandic: I believe the user-facing aspect of NEC protocol integration lies primarily in the YAML configuration and code generation. When changes are made to the NECData struct, the only files affected are __init__.py (for the NEC transmit action) and nec_protocol.cpp. To accommodate these changes, backward compatibility can be ensured within the NEC_SCHEMA and the cg.StructInitializer(NECData) definition. A properly designed NEC_SCHEMA can support both the old and new configuration schemas, allowing for a smooth transition. In the future, support for the old schema can be deprecated and removed.

```py
# NEC
NECData, NECBinarySensor, NECTrigger, NECAction, NECDumper = declare_protocol("NEC")
NEC_SCHEMA = cv.Schema(
    {
        cv.Required(CONF_ADDRESS): cv.hex_uint16_t,
        cv.Required(CONF_COMMAND): cv.hex_uint16_t,
        cv.Optional(CONF_COMMAND_REPEATS, default=1): cv.uint16_t, # // TODO: command_repeats -> repeats (repeat as repeat code, not whole message frame)
    }
)


@register_binary_sensor("nec", NECBinarySensor, NEC_SCHEMA)
def nec_binary_sensor(var, config):
    cg.add(
        var.set_data(
            cg.StructInitializer(
                NECData,
                ("address", config[CONF_ADDRESS]),
                ("command", config[CONF_COMMAND]),
                ("command_repeats", config[CONF_COMMAND_REPEATS]), # // TODO: command_repeats -> repeats (repeat as repeat code, not whole message frame)
            )
        )
    )
```

DjordjeMandic:
This is on hold until I prepare hardware and software for testing.

esphome/components/remote_base/nec_protocol.cpp
DjordjeMandic: Commit hash: a9ae6ba05782fe6f332d6a645f3ea088ea8a2d37

image
108.03ms 

image
11.808ms

DjordjeMandic:
Commit hash: a9ae6ba05782fe6f332d6a645f3ea088ea8a2d37

1 repeat:
image
108.03ms 

image
11.813ms

image
67.474ms

0 repeats:
image


5 repeats:
image
108.01ms between repeat codes

63 repeats:
```
[02:16:03][D][remote.nec:022]: Sending NEC: address=0xBF40, command=0xF40B, repeats=63
[02:16:03][W][remote.nec:025]: High repeat count may cause WDT timeout.
[02:16:08]E (43191) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
[02:16:08]E (43191) task_wdt:  - loopTask (CPU 1)
[02:16:08]E (43191) task_wdt: Tasks currently running:
[02:16:08]E (43191) task_wdt: CPU 0: IDLE
[02:16:08]E (43191) task_wdt: CPU 1: IDLE
[02:16:08]E (43191) task_wdt: Aborting.
```

DjordjeMandic:
@GRbit 

Could you test encoding and decoding? On my end it works fine. Just schema needs to prevent combination of `type: repeat` and `repeats: 0`. (it results into empty remote transmitter data)


Here is log output while holding single button:
```
[05:57:43][I][remote.nec:146]: Received NEC Frame (16-bit address): address=0x6B86, command=0x0C, command#=0xF3, command_valid=YES, repeats=0
[05:57:43][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:43][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:43][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:43][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:43][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:43][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:44][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:44][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:44][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:44][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:44][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:44][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:44][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:45][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:45][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:45][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
[05:57:45][I][remote.nec:146]: Received NEC Repeat Code: repeats=1
```
> IDLE: The amount of time that a signal should remain stable (i.e. not change) for it to be considered complete

I don't know if it's good idea to decode everything and incrmenet repeats since default `idle` is `10ms`.

esphome/components/remote_base/nec_protocol.h
```diff
@@ -5,19 +5,66 @@
...
   optional<NECData> decode(RemoteReceiveData src) override;
   void dump(const NECData &data) override;
+
+  std::string get_protocol_type_and_fields(const NECData &data) const;
+  bool is_extended(const NECData &data) const {
+    return data.address_lower != static_cast<uint8_t>(~data.address_upper);
+  }
+  bool is_command_valid(const NECData &data) const {
+    return data.command_lower == static_cast<uint8_t>(~data.command_upper);
+  }
```
DjordjeMandic: Gotta move this into Necdata struct

esphome/components/remote_base/__init__.py
```diff
@@ -746,11 +751,20 @@ async def keeloq_action(var, config, args):
 
 # NEC
 NECData, NECBinarySensor, NECTrigger, NECAction, NECDumper = declare_protocol("NEC")
+nec_code_type_enum_class = ns.enum("NECCodeType", is_class=True)
+NEC_CODE_TYPES = {
+    TYPE_FRAME_WITH_REPEATS: nec_code_type_enum_class.FRAME_WITH_REPEATS,
+    TYPE_REPEATS_ONLY: nec_code_type_enum_class.REPEATS_ONLY,
+}
+
 NEC_SCHEMA = cv.Schema(
```
DjordjeMandic: I tried to make this typed_schema but I have failed.

esphome/components/remote_base/__init__.py
```diff
@@ -785,8 +800,10 @@ async def nec_action(var, config, args):
     cg.add(var.set_address(template_))
     template_ = await cg.templatable(config[CONF_COMMAND], args, cg.uint16)
     cg.add(var.set_command(template_))
-    template_ = await cg.templatable(config[CONF_COMMAND_REPEATS], args, cg.uint16)
-    cg.add(var.set_command_repeats(template_))
+    template_ = await cg.templatable(config[CONF_REPEATS], args, cg.uint16)
+    cg.add(var.set_repeats(template_))
+    template_ = await cg.templatable(config[CONF_TYPE], args, nec_code_type_enum_class)
+    cg.add(var.set_type(template_))
```
DjordjeMandic, line 880: If type is set to nec_code_type_enum_class.REPEATS_ONLY, protocol encode function ignores address and command automatically.

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,97 +6,231 @@ namespace remote_base {
...
+  ESP_LOGD(TAG, "Encoding %s", get_protocol_type_and_fields(data).c_str());
 
-  dst->reserve(2 + 32 + 32 * data.command_repeats + 2);
-  dst->set_carrier_frequency(38000);
+  if (data.repeats > 20) {
+    ESP_LOGW(TAG, "High repeat count may cause WDT timeout.");
+  }
 
-  dst->item(HEADER_HIGH_US, HEADER_LOW_US);
+  if (data.repeats == 0 && data.type != NECCodeType::FRAME_WITH_REPEATS) {
```
DjordjeMandic: This should be prevented with schema

esphome/components/remote_base/nec_protocol.cpp
kbx81: Just seeing this...sorry I'm late to the party.

I don't think going from 38000 to 38222 will matter. I have receivers that are "38 kHz" respond to signals with a carrier as low as 36 kHz and as high as 40 kHz. There's a good bit of margin because there has to be because most transmitters are battery powered (TV/AV/AC remotes) and don't have a crystal to save on power and cost...so the frequency generated by these devices can vary quite a bit.

That said, I'm curious what docs you're referring to that say 38222 is "better"...I've never seen that before. I usually just see "38 kHz". Got a link?

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,97 +6,231 @@ namespace remote_base {
...
-    if (command != data.command) {
-      return {};
-    }
+  // Validate the final stop bit (end of the message frame)
+  if (!src.expect_mark(BIT_HIGH_US)) {
+    return {};
+  }
 
-    data.command_repeats += 1;
+  // Message frame received, `data.type = NECCodeType::FRAME` is already set
```
DjordjeMandic: Gotta fix some comments
```suggestion
  // Message frame received, `data.type = NECCodeType::FRAME_WITH_REPEATS` is already set
```

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,97 +6,231 @@ namespace remote_base {
...
+  if (!this->is_command_valid(data)) {
+    ESP_LOGW(TAG, "Decoded command invalid: 0x%04X", data.command);
```
DjordjeMandic, line 132: Protocol command is 8 bits. There are also 8 invreted bits sent. That is always the case. Address can be 8bits (same implementaiton) or 16 bits. I think its good practice to NOT reject the data but just give a warning. Maybe some special cases use wider non-standard commands.

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,97 +6,231 @@ namespace remote_base {
...
+  return this->waiting_for_repeat_code_;
+}
+
+bool NECBinarySensor::on_receive(RemoteReceiveData src) {
+  if (!this->matches(src)) {
+    return false;
+  }
+
+  this->publish_state(true);
+  this->set_timeout("repeat", this->repeat_timeout_ms_, [this]() {
```
DjordjeMandic, line 221: This invalidates ON state after 130ms, timeout should be configurable in yaml.

tests/components/remote_transmitter/common-buttons.yaml
```diff
@@ -11,12 +13,29 @@ button:
...
+        repeats: 1
+  - platform: template
+    name: NEC Repeats Only
+    id: living_room_lights_off_repeats_only
+    on_press:
+      remote_transmitter.transmit_nec:
+        address: ${nec_address}
+        command: ${nec_comamnd}
+        repeats: 10
+        type: repeat
```
DjordjeMandic: How to merge this?

esphome/components/remote_base/nec_protocol.cpp
DjordjeMandic: The source file had a link, but unfortunately, it’s no longer valid.

https://files.doc.altium.com/sites/default/files/wiki_attachments/299480/Legacy%20-%20FPGA%20Design.pdf

There was link in source file but its not valid anymore.

This would be official documentation: https://www.renesas.com/ja/document/apn/1184-remote-control-ir-receiver-decoder

If you examine the number of carrier wave periods within common NEC timing durations, you'll notice a key difference depending on frequency. FPGA-based designs often use 38.222 kHz with 562.5 µs durations because it provides the cleanest integer division of the signal, minimizing rounding errors and jitter.

Using a whole or half-number period count within a given pulse duration ensures better timing accuracy, signal stability, and cleaner transitions.

### Cleaner Signal Cuts & Phase Alignment:
  * The NEC protocol uses specific pulse durations (e.g., 562.5 µs for a bit).
  * A frequency of 38.222 kHz results in 21.5 periods per 562.5 µs.
  * This means each cycle aligns neatly every other pulse, avoiding fractional cycles that introduce phase shifts and rounding errors over time.
  
| Frequency (kHz) | Duration (µs) | Periods  | Notes |
|----------------|--------------|---------|--------------------------------|
| 38.000        | 560.0        | 21.2800 | Not an exact division; may cause drift |
| 38.000        | 562.5        | 21.3750 | Slightly better, but still not ideal |
| 38.222        | 560.0        | 21.4043 | Closer to a half-period but still off |
| 38.222        | 562.5        | 21.5000 | Perfect half-period alignment ✅ |

In the end, both values work reliably with the RMT component and standard software decoding, as long as tolerance is accounted for.

esphome/components/remote_base/nec_protocol.h
DjordjeMandic: Decided not to since its part of protocol and not part of POD - plain old data

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,97 +6,234 @@ namespace remote_base {
...
+static const uint32_t BIT_HIGH_US = 562;             // Bit HIGH duration: 562.5µs
+static const uint32_t BIT_ONE_LOW_US = 1687;         // Logic '1': 562.5µs HIGH + 1687.5µs LOW
+static const uint32_t BIT_ZERO_LOW_US = 562;         // Logic '0': 562.5µs HIGH + 562.5µs LOW
+static const uint32_t SPACE_INTER_FRAME_US = 40000;  // Inter-frame space: 40ms
+static const uint32_t SPACE_AGC_REPEAT_US = 96187;   // AGC Repeat space: ~96.1875ms
 
 void NECProtocol::encode(RemoteTransmitData *dst, const NECData &data) {
-  ESP_LOGD(TAG, "Sending NEC: address=0x%04X, command=0x%04X command_repeats=%d", data.address, data.command,
-           data.command_repeats);
+  ESP_LOGD(TAG, "Encoding %s", get_protocol_type_and_fields(data).c_str());
```
DjordjeMandic:
```suggestion
  ESP_LOGD(TAG, "Encoding %s", this->get_protocol_type_and_fields(data).c_str());
```

esphome/components/remote_base/nec_protocol.cpp
kbx81: That makes perfect sense. Thanks for the explanation! 🍻

hmaleville:
Any hope that failing checks will be soon fixed?

DjordjeMandic:
@hmaleville Will be, I have some ideas on my mind for this but I'm currently busy. It's draft PR because it's still "work in progress".

esphome/components/remote_base/nec_protocol.cpp
DjordjeMandic: Tried to do it, schema.`add_extra` does not work...

DjordjeMandic:
Currently, I'm struggling with properly implementing the NEC config schema. While I have an initial implementation, it's not ideal and could greatly benefit from a more refined approach.

I've attempted to use a custom validator with schema.add_extra, but it seems to be completely ignored. At this point, I could really use some guidance—both in refining the schema and ensuring that the implementation follows the best approach.

Before finalizing the PR, I’d like to consult with someone to confirm the right way to structure the schema. Otherwise, there’s a risk of introducing another breaking change in the future on top of the existing one. A detailed review of python code and schemas would be incredibly helpful to address any potential issues.

GRbit:
Thank you @DjordjeMandic for your efforts, I see that you work hard and spend a lot of time for our common benefit. Thank you very much.

As I understand it, previously the NEC adapter was reading the signal correctly except in the case it had repeats after. Producing NEC format signals also worked correctly, except the case with repeats.

The current state, if I'm not mistaken, is that producing NEC signals is now 100% compliant with the protocol, but reading NEC signals is complicated. Since this allows users to manually set NEC signals correctly, I think this is already a big improvement. 
I have an idea to split PR into two parts: one for encoding and one for decoding. Then the first code can be merged and used, and the other one can wait for review by someone more experienced. What do you think about it?

Thanks again for your time and effort!

DjordjeMandic:
Thanks @GRbit, receiving with repeat codes is fully supported, but only when using the **NEC remote receiver binary sensor**.  

Signal is sent from [Universal IR Blaster](https://github.com/DjordjeMandic/Universal-IR-Blaster-TV-B-Gone) running [Girs infrared server](https://github.com/DjordjeMandic/GirsUIRB) with [IrScrutinizer PC app](https://github.com/bengtmartensson/IrScrutinizer)
image
Here’s an example from the logs:  

```
[23:09:02][V][remote.nec:154]: Decoded NEC Frame (16-bit address): address=0x6B86, command=0x0C, command#=0xF3, command_valid=YES, repeats=0, src_idx=67, src_sz=68
[23:09:02][D][binary_sensor:036]: 'KWORLD Remote KW Button': Sending state ON
[23:09:02][I][remote.raw:028]: Received Raw: 9111, -4510, 614, -558, 610, -1680, 610, -1662, 627, -558, 580, -567, 635, -533, 605, -538, 614, -1700, 585, -1705, 605, -1663, 657, -507, 606, -1705, 635, -511, 635, -1651, 609, -1680, 664, -499, 618, -559, 613, -525, 643, -1650, 635,
[23:09:02][I][remote.raw:041]:   -1654, 664, -504, 639, -533, 639, -503, 614, -558, 614, -1675, 635, -1650, 639, -533, 613, -529, 639, -1675, 589, -1700, 640, -1649, 636, -1650, 639
[23:09:02][V][remote.nec:111]: Decoded repeat code only
[23:09:02][I][remote.raw:041]: Received Raw: 9139, -2222, 639
[23:09:02][D][binary_sensor:036]: 'KWORLD Remote KW Button': Sending state OFF
```

As you can see, this was a **single button press** from the user's perspective, but the **remote receiver component treats it as two separate receptions** (two instances of `[I][remote.raw:028]: Received Raw:`).  

### **Why Repeat Codes Are Treated Separately**  
The issue with handling repeat codes comes from hardware limitations and the way **ESPHome processes IR signals**.  

1. When a reception starts, the raw timing data is stored in memory.  
2. The system waits until no signal is received for a duration exceeding the [`[idle](https://esphome.io/components/remote_receiver.html)`](https://esphome.io/components/remote_receiver.html) period.  
3. Only then is the reception considered complete, and the data is passed to the decoder.  
4. **Because of this behavior, repeat codes must be treated as separate IR receptions.**  

### **Current State of the PR**  
- The **C/C++ implementation is stable and ready.**  
- There may be **room for improvement in TX/encoding**, but that should wait for **callback-based encoding**.  
- For now, users should refer to the **example config.yaml** in the PR description, which should be sufficient for most real-world use cases.  

### **Future Improvements (out of scope for this PR)**  
The **remote transmitter component** will need:  
- Callback-based encoding  
- Asynchronous transmission

Once those are implemented, **more advanced protocol handling** can be added on top. For now, users must stay within the **short TX limit** and use the **remote receiver binary sensor**. Fortunately, the binary sensor is asynchronous, so users can **hold down the button as long as they want**—they just can’t **transmit for an extended period** without triggering the **watchdog timer (WDT)**.  

There are also **many filters available** for the binary sensor, so users can customize their setup as needed.  

### **Final Thoughts**  
I believe the current **RX implementation with the remote receiver binary sensor is sufficient**. If you have any concerns, feel free to **test it yourself!** 😊

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,97 +6,243 @@ namespace remote_base {
...
+// // NOLINT https://www.renesas.com/ja/document/apn/1184-remote-control-ir-receiver-decoder
+
+// Timing constants in microseconds
+static const uint32_t AGC_HIGH_US = 9000;            // AGC burst: 9ms HIGH
+static const uint32_t LONG_PAUSE_LOW_US = 4500;      // After AGC burst: 4.5ms LOW
+static const uint32_t SHORT_PAUSE_LOW_US = 2250;     // After AGC burst: 2.25ms LOW
+static const uint32_t BIT_HIGH_US = 562;             // Bit HIGH duration: 562.5µs
+static const uint32_t BIT_ONE_LOW_US = 1687;         // Logic '1': 562.5µs HIGH + 1687.5µs LOW
+static const uint32_t BIT_ZERO_LOW_US = 562;         // Logic '0': 562.5µs HIGH + 562.5µs LOW
+static const uint32_t SPACE_INTER_FRAME_US = 40000;  // Inter-frame space: 40ms
```
kbx81, line 22: Maybe these could be `uint16_t` types? :)

esphome/components/remote_base/nec_protocol.cpp
```diff
@@ -6,97 +6,243 @@ namespace remote_base {
...
+  }
+
+  debug_message += str_sprintf(" repeats=%u", data.repeats);
+
+  return debug_message;
+}
+
+void NECBinarySensor::dump_config() {
+  LOG_BINARY_SENSOR("", "Remote Receiver NEC Binary Sensor", this);
+  ESP_LOGCONFIG(TAG, "  Repeat Timeout: %u ms", this->repeat_timeout_ms_);
```
kbx81: Nitpicking a bit, but...No Reason For These To Be Title Case. haha
```suggestion
  LOG_BINARY_SENSOR("", "Remote receiver NEC binary sensor", this);
  ESP_LOGCONFIG(TAG, "  Repeat timeout: %u ms", this->repeat_timeout_ms_);
```

esphome/components/remote_base/nec_protocol.cpp
kbx81: Maybe these could be `uint16_t` types? :)

GRbit:
> Thanks @GRbit, receiving with repeat codes is fully supported, but only when using the **NEC remote receiver binary sensor**.

Oh, that's amazing! I just misunderstood your previous message. Great job!

ximex:
> Commit hash: [a9ae6ba](https://github.com/esphome/esphome/commit/a9ae6ba05782fe6f332d6a645f3ea088ea8a2d37)
> 
> 1 repeat: image 108.03ms
> 
> image 11.813ms
> 
> image 67.474ms
> 
> 0 repeats: image
> 
> 5 repeats: image 108.01ms between repeat codes
> 
> 63 repeats:
> 
> ```
> [02:16:03][D][remote.nec:022]: Sending NEC: address=0xBF40, command=0xF40B, repeats=63
> [02:16:03][W][remote.nec:025]: High repeat count may cause WDT timeout.
> [02:16:08]E (43191) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
> [02:16:08]E (43191) task_wdt:  - loopTask (CPU 1)
> [02:16:08]E (43191) task_wdt: Tasks currently running:
> [02:16:08]E (43191) task_wdt: CPU 0: IDLE
> [02:16:08]E (43191) task_wdt: CPU 1: IDLE
> [02:16:08]E (43191) task_wdt: Aborting.
> ```

@DjordjeMandic what software is this that you are using to visual the bits?

DjordjeMandic:
> @DjordjeMandic what software is this that you are using to visual the bits?

[PulseView](https://sigrok.org/wiki/PulseView) logic analyzer GUI for sigrok with IR NEC decoder and Timing decoder. List of [supported hardware](https://sigrok.org/wiki/Supported_hardware#Logic_analyzers)

esphome/components/remote_base/nec_protocol.cpp
DjordjeMandic: That's what I wanted but then I decided to copy https://github.com/esphome/esphome/blob/58fe8b39b20c8fe9767107a6136ca30873b82056/esphome/components/remote_base/remote_base.cpp#L126 and stick with it. I agree that it looks ugly title case.

DjordjeMandic:
Defining `USE_REMOTE_TRANSMITTER_NEC_FULL_LEN` ensures that the NEC encoding function always appends a space after the frame or repeat code, rounding the total transmission time to 108ms. This guarantees a consistent delay from the start of each transmission, preventing the next IR code from being sent until the full 108ms period has elapsed.

While this effectively enforces the required IR silence time at the end of a transmission, it also blocks the loop for that duration. In most cases, this blocking behavior isn't an issue, as IR silence naturally occurs unless a new transmission starts immediately. However, since this may not be suitable for all use cases, it is disabled by default. Users can enable it by defining USE_REMOTE_TRANSMITTER_NEC_FULL_LEN if needed.

A practical use case for this is when sending multiple NEC transmissions sequentially without adding a delay between them. This define ensures the proper timing between transmissions, preventing overlapping signals.

I'm unsure where the best place to configure this in YAML would be, should it go under the remote transmitter component or somewhere else?

DjordjeMandic:
I decided to reuse `times` config option. Ditched the `repeats` config option idea but left `repeats` in `NECData` reserved for future implementations.

Emrvb:
I'm a bit conflicted whether NEC2 should be integrated into the existing implementation or just be defined as another protocol. 

If you want to adjust timings to match what is seen in the wild:

FREQUENCY: 38400
BIT_HIGH_US: 564
BIT_ONE_LOW_US: 1692
BIT_ZERO_LOW_US: 564

Sources:
- https://github.com/bengtmartensson/IrpTransmogrifier 
- http://hifi-remote.com/forums/index.php

DjordjeMandic:
> I'm a bit conflicted whether NEC2 should be integrated into the existing implementation or just be defined as another protocol.

They're the same, only difference is in a way which repeat codes are sent. Same encoder can be used, only difference is that instead of sending repeat frames, whole data frame is sent again. I don't think that its necessary to make whole new protocol (in source) just because of that. Source(encoder/decoder) allows it, yaml config should just do the diff between them. But IMO NEC/NEC1/NEC2 are wayyy to similar to have separate actions / sensors / triggers for each.

> If you want to adjust timings to match what is seen in the wild:
> 
> FREQUENCY: 38400 BIT_HIGH_US: 564 BIT_ONE_LOW_US: 1692 BIT_ZERO_LOW_US: 564
> 
> Sources:
> 
> * https://github.com/bengtmartensson/IrpTransmogrifier
> * http://hifi-remote.com/forums/index.php

I took it from altiums fpga design guide for nec protocol. I'd say that those 2us won't change much. Most battery remotes dont actually have that precise timings and stuff still works. 38.222kHz makes most sense at 562.5us. Take a look at https://github.com/esphome/esphome/pull/8015/files#r1960821448

DjordjeMandic:
Closing this because of nature of PR (wayy to long for single review). Will do it in multiple PRs per feature/mod when I get enough free time.

PR #8020, opened 2024-12-27
add bus voltage over limit for ina2xx sensors

asssaf:

INA228/229 have the ability to set a threshold that triggers the alert pin when the bus voltage exceeds the provided limit.
This change adds the ability to set the threshold.

## Types of changes

- [ ] Other

- N/A

- esphome/esphome-docs#4555

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: ina2xx_i2c
    id: my_charge_sensor
    i2c_id: bus_a
    model: INA228
    address: 0x40
    shunt_resistance: 0.015 ohm
    max_current: 10 A
    adc_range: 0
    adc_averaging: 1
    adc_time: 1052us
    update_interval: 1s
    current: 
      name: "INA228 Current"
      id: current
    bus_voltage: 
      name: "INA228 Bus Voltage"
      id: bus_voltage
    charge: 
      name: "INA228 Charge"
      id: charge
    energy: 
      name: "INA228 Energy"
      id: energy
    power:
      name: "INA228 Power"
      id: power
    shunt_voltage:
      name: "INA228 Shunt Voltage"
      id: shunt_voltage
      unit_of_measurement: "mV"
    temperature:
      name: "INA228 Temperature"
      id: temperature
    bus_voltage_over_limit: 3v

```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs). https://github.com/esphome/esphome-docs/pull/4555


esphome/components/ina2xx_base/ina2xx_base.cpp
```diff
@@ -374,6 +378,14 @@ bool INA2XX::configure_shunt_tempco_() {
   return true;
 }
 
+bool INA2XX::configure_bovl_() {
+  // Only for 228/229
+  if ((this->ina_model_ == INAModel::INA_228 || this->ina_model_ == INAModel::INA_229) && this->bovl_v_ > 0) {
+    return this->write_unsigned_16_(RegisterMap::REG_BOVL, (uint16_t) (this->bovl_v_ * 1000 / 3.125));
+  }
+  return true;
```
DjordjeMandic, line 387: IMO i would return false here but rest of the code (`configure_shunt_tempco_`) already does the same. idk why

asssaf: this is my first ESPHome PR so i'm not sure about the conventions. My guess is that it returns true because this setting is not relevant, so is just being ignored (validation is done at the schema level)?

DjordjeMandic: > My guess is that it returns true because this setting is not relevant

Might be, anyways it will work

latonita:
@asssaf looks okay to me

esphome/components/ina2xx_base/ina2xx_base.cpp
```diff
@@ -374,6 +378,14 @@ bool INA2XX::configure_shunt_tempco_() {
   return true;
 }
 
+bool INA2XX::configure_bovl_() {
```
jesserockz: Should it be bvol everywhere instead? The acronym doesnt match the config string.

jesserockz (changes requested):

esphome/components/ina2xx_base/ina2xx_base.cpp
asssaf: This is an inconsistency from the [data sheet](https://www.ti.com/lit/ds/symlink/ina228.pdf). The registers are called [BOVL](https://github.com/esphome/esphome/blob/9e40d4cf45269e052bd62a78a5619ad8dc863594/esphome/components/ina2xx_base/ina2xx_base.h#L24), BUVL, SOVL, SUVL.
I'll jut rename it to the full name to avoid confusion.

DjordjeMandic: That could be mentioned in doxygen comment

latonita: @asssaf at another look - there is undervoltage and overvoltage conditions
in my world it would make senesto rename all the functions and variables properly.
from `configure_bus_voltage_over_limit_` to `configure_bus_over_voltage_limit_` and so on.
its not 'voltage over' but it is 'over voltage'

PR #8026, opened 2024-12-31
Added VERY_VERBOSE dfplayer printing

solarkennedy:
This adds VERY_VERBOSE logging for the defplayer component.

This is useful to me because there are still messages it is sending that I'm cross-referencing with the datasheet.

For example, the USB drive version of finished playback is different than with the SD card.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/dfplayer/dfplayer.cpp
+++ b/esphome/components/dfplayer/dfplayer.cpp
@@ -159,6 +159,15 @@ void DFPlayer::loop() {
         }
         break;
       case 9:  // End byte
+#ifdef ESPHOME_LOG_HAS_VERY_VERBOSE
+        char byte_sequence[100];
+        byte_sequence[0] = '\0';
+        for (size_t i = 0; i < this->read_pos_ + 1; ++i) {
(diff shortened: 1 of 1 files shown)


glmnet:
Even though new case handler for usb is ok, have you tested enabling the uart debugger? https://esphome.io/components/uart.html#uart-debugging

solarkennedy:
Ah interesting, no I hadn't thought to add debugging at that layer. Thanks for the tip!
I think it would need to have a special delimiter, because in the dfplayer.cpp we know when we have hit the special length and when a message is complete, where we can pretty print the whole thing. 

All I really want is the serial data to be printed in exactly the same form that I read on the datasheet.

DjordjeMandic (approved):

jesserockz (approved):

PR #8027, opened 2024-12-31
[nextion] Brightness control tweaks

kbx81:

I noticed recently that my displays would "pop" their brightness to 100% and then immediately reduce it back down (as configured) shortly after boot. I realized that this is because we were *always* setting the brightness as part of start-up. Since I'm managing display brightness...another way...this behavior is not desirable. This PR only sets the display brightness at start-up if `brightness` is in the config -- otherwise, it leaves it alone.

In addition, this PR adds an action which can be used to set the brightness (as opposed to having to use a lambda).

I've labeled this as a breaking change only because it *may* result in a change of behavior when `brightness` is not set in config. That said, I believe the displays default to 100% brightness unless the `dims` command was used on the display at some point to change it.

## Types of changes

- [x] Other

- fixes <link to issue>

- esphome/esphome-docs#4550

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/nextion/automation.h
+++ b/esphome/components/nextion/automation.h
@@ -49,6 +49,23 @@ class TouchTrigger : public Trigger<uint8_t, uint8_t, bool> {
   }
 };
 
+template<typename... Ts> class NextionSetBrightnessAction : public Action<Ts...> {
+ public:
+  explicit NextionSetBrightnessAction(Nextion *component) : component_(component) {}
+
+  TEMPLATABLE_VALUE(float, brightness)
+
+  void play(Ts... x) override {
+    this->component_->set_brightness(this->brightness_.value(x...));
+    this->component_->set_backlight_brightness(this->brightness_.value(x...));
+  }
+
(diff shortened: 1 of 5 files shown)


edwardtfn:
Looks good.
Also during the connection this cmponent sends a commands to restart the display, which is a bit annoying... I will see if I find some time to look at that later (in another PR) as I think it should do the bare minimum changes when establishing a connection
Any thoughts on that?

kbx81:
> Looks good.

Thanks for taking a look! If it's good with you, as a code owner, feel free to mark "approved". 😄

> Also during the connection this cmponent sends a commands to restart the display, which is a bit annoying... I will see if I find some time to look at that later (in another PR) as I think it should do the bare minimum changes when establishing a connection Any thoughts on that?

I tend to agree -- or at least make it a flag so the user can enable/disable the behavior to fit their use case.

edwardtfn (approved):

jesserockz (approved):

PR #8030, opened 2025-01-01
Add support for the DAC on the S2

### LINKED ISSUE #4123: I2S Media Player not working for ESP32 S2 based boards
Because the ESP32 S2 does not contain an internal DAC the mediaplayer will fail to compile even when a external DAC is being used  because the i2s_dac_mode_t struct isn't available.

See: https://github.com/espressif/esp-idf/blob/b04e33256df3a6149e75a4704393322b9149df6c/components/driver/deprecated/driver/i2s_types_legacy.h#L130

Example of it being used regardless of the target IC:
https://github.com/esphome/esphome/blob/72b294333287658ecf8577de5cafd05a901d7921/esphome/components/i2s_audio/i2s_audio_media_player.h#L28

maraid:

Added support for DAC on the S2.
Start using dac_oneshot.h instead of the deprecated dac.h

## Types of changes

- [ ] Other

- fixes [4123](https://github.com/esphome/esphome/issues/4123)

- esphome/esphome-docs#4552

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esp32:
  board: esp32-s2-saola-1
  framework:
    type: esp-idf
output:
  - platform: esp32_dac
     pin: GPIO17
     id: dac1_output

```

## Checklist:

### DIFF
--- a/esphome/components/esp32_dac/output.py
+++ b/esphome/components/esp32_dac/output.py
@@ -5,7 +5,7 @@ import esphome.codegen as cg
 from esphome.core import CORE
 from esphome.components.esp32 import get_esp32_variant
 from esphome.components.esp32.const import VARIANT_ESP32S2
-from esphome.const import CONF_ID, CONF_NUMBER, CONF_PIN, KEY_CORE, KEY_TARGET_FRAMEWORK
+from esphome.const import CONF_ID, CONF_NUMBER, CONF_PIN
 
 DEPENDENCIES = ["esp32"]
 
--- a/esphome/components/esp32_dac/output.py
+++ b/esphome/components/esp32_dac/output.py
@@ -1,18 +1,27 @@
+import esphome.codegen as cg
+import esphome.config_validation as cv
 from esphome import pins
 from esphome.components import output
-import esphome.config_validation as cv
-import esphome.codegen as cg
 from esphome.components.esp32 import get_esp32_variant
-from esphome.components.esp32.const import VARIANT_ESP32S2
+from esphome.components.esp32.const import VARIANT_ESP32, VARIANT_ESP32S2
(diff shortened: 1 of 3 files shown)


esphome/components/esp32_dac/output.py
```diff
@@ -2,14 +2,17 @@
...
+from esphome.components.esp32 import get_esp32_variant
+from esphome.components.esp32.const import VARIANT_ESP32S2
 from esphome.const import CONF_ID, CONF_NUMBER, CONF_PIN
 
 DEPENDENCIES = ["esp32"]
 
 
 def valid_dac_pin(value):
+    valid_pins = (17, 18) if get_esp32_variant() == VARIANT_ESP32S2 else (25, 26)
     num = value[CONF_NUMBER]
```
jesserockz: Something like this would be a little better and easier to understand
```suggestion
DAC_PINS = {
  VARIANT_ESP32: (25, 26),
  VARIANT_ESP32S2: (17, 18),
}

def valid_dac_pin(value):
    variant = get_esp32_variant()
    if (valid_pins := DAC_PINS.get(variant)) is None:
        raise cv.Invalid(f"DAC is not supported on {variant})
    num = value[CONF_NUMBER]
```

jesserockz (changes requested):

esphome/components/esp32_dac/output.py
maraid: Added, thanks

maraid:
https://github.com/esphome/issues/issues/3787

jesserockz (approved):

PR #8031, opened 2025-01-01
[climate/template]: New template climate component

### LINKED ISSUE #2783: Template climate platform
There are many cases where directly unsupported HVAC devices could be added to ESPHome if a Template climate platform would exist.
- for unsupported IR devices lambdas could be used to call directly various library functions (see Hagakurje's example below)
- standard ESPHome triggers and actions could be used to call ModBUS or even Remote Transmitter stuff
- some simpler devices could be controlled with switches etc.

Here somebody implements a `custom` climate to control an unsupported `gree` model by calling in directly `IRremoteESP8266`: https://github.com/Hagakurje/ESPHome_Gree_AC (it is known that `custom` platforms are to be deprecated/removed)

Here someone is sending captured IR commands in RAW or other likesome protocols using `remote_transmitter` platform, and integrates the climate functionality as template fan: https://github.com/esphome/feature-requests/issues/2240 (there are many cases when this would be enough).

Here some people have HVAC devices accessible via ModBUS and would like to implement it as a climate: https://github.com/esphome/feature-requests/issues/2752, https://github.com/esphome/feature-requests/issues/2582, https://github.com/esphome/feature-requests/issues/2687 

Also applicable: https://github.com/esphome/feature-requests/issues/2323, https://github.com/esphome/feature-requests/issues/2265, https://github.com/esphome/feature-requests/issues/2149, https://github.com/esphome/feature-requests/issues/2105, https://github.com/esphome/feature-requests/issues/2070, https://github.com/esphome/feature-requests/issues/1909, https://github.com/esphome/feature-requests/issues/1855, https://github.com/esphome/feature-requests/issues/1797 (again, with `custom`), https://github.com/esphome/feature-requests/issues/1743, https://github.com/esphome/feature-requests/issues/1521, https://github.com/esphome/feature-requests/issues/1223, https://github.com/esphome/feature-requests/issues/689, https://github.com/esphome/feature-requests/issues/478 and so on...

**Additional context**
The concept behind Tasmota's IRHVAC is nice: https://tasmota.github.io/docs/Tasmota-IR/#sending-irhvac-commands
It's basically a wrapper around `IRremoteESP8266` where you just select a protocol and the commands you'd like to send.

A kind of template climate component would be very flexible in ESPHome, where anyone could mix and match the features supported by `IRremoteESP8266` and his device. Eg. some devices only have vertical swing, then let the component not expose the horizontal swing. Advanced functions could be called with template switches.

polyfloyd:

This patch introduces a new template climate component.

The goal is to make the climate interface available for a system where all state is external. The implementation works by combining the state of other components such as Selects, Sensors and Numbers.

## Types of changes

- fixes esphome/feature-requests#2783

- esphome/esphome-docs#4554

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
select:
  - platform: template
    id: climate_mode
    lambda: !lambda |-
        return id(lg_therma_active).state
             ? id(lg_therma_mode).state
             : std::string("OFF");
    set_action:
      - if:
          condition:
            lambda: 'return x != "OFF";'
          then:
            - switch.turn_on: lg_therma_active
            - select.set:
                id: lg_therma_mode
                option: !lambda return x;
          else:
            - switch.turn_off: lg_therma_active
    options:
      - "OFF"
      - "COOL"
      - "HEAT"
      - "AUTO"

number:
  - platform: modbus_controller
    id: true_setpoint
    icon: mdi:thermometer
    address: 2
    min_value: 16
    max_value: 30
    step: 1
    multiply: 10
    device_class: "temperature"
    unit_of_measurement: "°C"

climate:
  - platform: template_climate
    name: "LG Therma"
    target_temperature_id: true_setpoint
    current_temperature_id: climate_current_temperature
    mode_id: climate_mode
    fan_mode_id: climate_fan_mode
    action_id: climate_action
    temperature_step: 0.1C
```

A full configuration file using this component can be found here: https://git.polyfloyd.net/polyfloyd/esphome-config/src/branch/main/lg-therma.yaml

## Checklist:

### DIFF
--- a/esphome/components/template/climate/__init__.py
+++ b/esphome/components/template/climate/__init__.py
@@ -1,0 +1,65 @@
+import esphome.codegen as cg
+from esphome.components import climate
+from esphome.components.number import Number
+from esphome.components.select import Select
+from esphome.components.sensor import Sensor
+from esphome.components.text_sensor import TextSensor
+import esphome.config_validation as cv
+from esphome.const import CONF_ACTION_ID, CONF_ID
+
+from .. import template_ns
+
+CONF_AUTO_ACTION = "auto_action"
+CONF_CURRENT_TEMPERATURE_ID = "current_temperature_id"
+CONF_TARGET_TEMPERATURE_ID = "target_temperature_id"
+CONF_MODE_ID = "mode_id"
+CONF_FAN_MODE_ID = "fan_mode_id"
+CONF_SWING_MODE_ID = "swing_mode_id"
+CONF_PRESET_ID = "preset_id"
+
+TemplateClimate = template_ns.class_("TemplateClimate", climate.Climate, cg.Component)
+
+CONFIG_SCHEMA = climate.CLIMATE_SCHEMA.extend(
+    {
+        cv.GenerateID(CONF_ID): cv.declare_id(TemplateClimate),
+        cv.Required(CONF_CURRENT_TEMPERATURE_ID): cv.use_id(Sensor),
+        cv.Required(CONF_TARGET_TEMPERATURE_ID): cv.use_id(Number),
+        cv.Required(CONF_MODE_ID): cv.use_id(Select),
+        cv.Optional(CONF_FAN_MODE_ID): cv.use_id(Select),
+        cv.Optional(CONF_SWING_MODE_ID): cv.use_id(Select),
+        cv.Optional(CONF_PRESET_ID): cv.use_id(Select),
+        cv.Optional(CONF_ACTION_ID): cv.use_id(TextSensor),
+    }
+)
+
+
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await climate.register_climate(var, config)
+
+    cg.add(
+        var.set_current_temperature(
+            await cg.get_variable(config[CONF_CURRENT_TEMPERATURE_ID])
+        )
+    )
+
+    cg.add(
+        var.set_target_temperature(
+            await cg.get_variable(config[CONF_TARGET_TEMPERATURE_ID])
+        )
+    )
+
+    cg.add(var.set_mode(await cg.get_variable(config[CONF_MODE_ID])))
+
+    if v := config.get(CONF_FAN_MODE_ID):
+        cg.add(var.set_fan_mode(await cg.get_variable(v)))
+
+    if v := config.get(CONF_SWING_MODE_ID):
+        cg.add(var.set_swing_mode(await cg.get_variable(v)))
+
+    if v := config.get(CONF_PRESET_ID):
+        cg.add(var.set_preset(await cg.get_variable(v)))
+
+    if v := config.get(CONF_ACTION_ID):
+        cg.add(var.set_action(await cg.get_variable(v)))
--- a/esphome/components/template/climate/template_climate.cpp
+++ b/esphome/components/template/climate/template_climate.cpp
@@ -1,0 +1,210 @@
+#include "template_climate.h"
+#include <functional>
+
+namespace esphome {
+namespace template_ {
+
+static const char *const TAG = "template_climate";
+
+ClimateMode mode_from_string(const std::string &s) {
+  if (str_equals_case_insensitive(s, "OFF")) {
+    return CLIMATE_MODE_OFF;
+  } else if (str_equals_case_insensitive(s, "HEAT_COOL")) {
+    return CLIMATE_MODE_HEAT_COOL;
+  } else if (str_equals_case_insensitive(s, "COOL")) {
+    return CLIMATE_MODE_COOL;
+  } else if (str_equals_case_insensitive(s, "HEAT")) {
+    return CLIMATE_MODE_HEAT;
+  } else if (str_equals_case_insensitive(s, "FAN_ONLY")) {
+    return CLIMATE_MODE_FAN_ONLY;
+  } else if (str_equals_case_insensitive(s, "DRY")) {
+    return CLIMATE_MODE_DRY;
+  } else if (str_equals_case_insensitive(s, "AUTO")) {
+    return CLIMATE_MODE_AUTO;
+  } else {
+    ESP_LOGW(TAG, "Unrecognized mode %s", s.c_str());
+    return CLIMATE_MODE_OFF;
+  }
+}
+
+ClimateFanMode fan_mode_from_string(const std::string &s) {
+  if (str_equals_case_insensitive(s, "ON")) {
+    return CLIMATE_FAN_ON;
+  } else if (str_equals_case_insensitive(s, "OFF")) {
+    return CLIMATE_FAN_OFF;
+  } else if (str_equals_case_insensitive(s, "AUTO")) {
+    return CLIMATE_FAN_AUTO;
+  } else if (str_equals_case_insensitive(s, "LOW")) {
+    return CLIMATE_FAN_LOW;
+  } else if (str_equals_case_insensitive(s, "MEDIUM")) {
+    return CLIMATE_FAN_MEDIUM;
+  } else if (str_equals_case_insensitive(s, "HIGH")) {
+    return CLIMATE_FAN_HIGH;
+  } else if (str_equals_case_insensitive(s, "MIDDLE")) {
+    return CLIMATE_FAN_MIDDLE;
+  } else if (str_equals_case_insensitive(s, "FOCUS")) {
+    return CLIMATE_FAN_FOCUS;
+  } else if (str_equals_case_insensitive(s, "DIFFUSE")) {
+    return CLIMATE_FAN_DIFFUSE;
+  } else if (str_equals_case_insensitive(s, "QUIET")) {
+    return CLIMATE_FAN_QUIET;
+  } else {
+    ESP_LOGW(TAG, "Unrecognized fan mode %s", s.c_str());
+    return CLIMATE_FAN_AUTO;
+  }
+}
+
+ClimateSwingMode swing_mode_from_string(const std::string &s) {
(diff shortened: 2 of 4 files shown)


polyfloyd:
@jesserockz May I tag you for a review?

jzucker2:
I'm also very interested in this! Thanks for the effort!

uinbytes:
I am also interested in this component.

ssieb:
A template climate component is a good thing, but why are you combining it with numbers and selects?  The climate component already has its own interface.  You just need to provide triggers like all the other template components.

polyfloyd:
Because triggers are a means to copy state from one component to another. The device that I have written this for already has its own statekeeping that I did not want to replicate.

The state also has to be propagated in both ways. So from Climate to underlying and from underlying to climate.  I could make a template climate component that uses a trigger to persist, e.g. its temperature setpoint. But if the setpoint is altered by the underlying device it would need to propagate it to the Climate via a trigger too? That would cause an event-storm.

My first iteration of this component did use triggers along with a kind of function to retrieve the current state. But I quickly realized that I was duplicating the exact interface of template selects and numbers and thought it would be cleaner to delegate to those instead.

This also removes the need for cumbersome value mapping from C++ enums to strings and vice versa.

ssieb:
You should start a thread in the devs channel on the esphome discord to discuss this.

polyfloyd:
Why Discord? I think GitHub is a suitable medium for discussing this

LukasGrebe:
thanks for your work on this. found this googling how to get my esphome interface via modbus to my heatpump to show as climate control in home assistant.

uinbytes:
Up

devildant:
Hi i'am interesting to this feature, but question why fix the string for the function bellow?  it's realy limited
I have a lot of additional options that aren't in these lists for my air conditioner. As it stands, this implementation is unusable for me. It's possible to enter any values, and it doesn't cause any issues on the HA side.

I currently use the Climate Template add-on to do this (https://github.com/litinoveweedle/hass-template-climate), but I would have liked a solution directly in the ESP.

```
ClimateFanMode fan_mode_from_string(const std::string &s) {
  if (str_equals_case_insensitive(s, "ON")) {
    return CLIMATE_FAN_ON;
  } else if (str_equals_case_insensitive(s, "OFF")) {
    return CLIMATE_FAN_OFF;
  } else if (str_equals_case_insensitive(s, "AUTO")) {
    return CLIMATE_FAN_AUTO;
  } else if (str_equals_case_insensitive(s, "LOW")) {
    return CLIMATE_FAN_LOW;
  } else if (str_equals_case_insensitive(s, "MEDIUM")) {
    return CLIMATE_FAN_MEDIUM;
  } else if (str_equals_case_insensitive(s, "HIGH")) {
    return CLIMATE_FAN_HIGH;
  } else if (str_equals_case_insensitive(s, "MIDDLE")) {
    return CLIMATE_FAN_MIDDLE;
  } else if (str_equals_case_insensitive(s, "FOCUS")) {
    return CLIMATE_FAN_FOCUS;
  } else if (str_equals_case_insensitive(s, "DIFFUSE")) {
    return CLIMATE_FAN_DIFFUSE;
  } else if (str_equals_case_insensitive(s, "QUIET")) {
    return CLIMATE_FAN_QUIET;
  } else {
    ESP_LOGW(TAG, "Unrecognized fan mode %s", s.c_str());
    return CLIMATE_FAN_AUTO;
  }
}

ClimateSwingMode swing_mode_from_string(const std::string &s) {
  if (str_equals_case_insensitive(s, "OFF")) {
    return CLIMATE_SWING_OFF;
  } else if (str_equals_case_insensitive(s, "BOTH")) {
    return CLIMATE_SWING_BOTH;
  } else if (str_equals_case_insensitive(s, "VERTICAL")) {
    return CLIMATE_SWING_VERTICAL;
  } else if (str_equals_case_insensitive(s, "HORIZONTAL")) {
    return CLIMATE_SWING_HORIZONTAL;
  } else {
    ESP_LOGW(TAG, "Unrecognized swing mode %s", s.c_str());
    return CLIMATE_SWING_BOTH;
  }
}

ClimatePreset preset_from_string(const std::string &s) {
  if (str_equals_case_insensitive(s, "NONE")) {
    return CLIMATE_PRESET_NONE;
  } else if (str_equals_case_insensitive(s, "HOME")) {
    return CLIMATE_PRESET_HOME;
  } else if (str_equals_case_insensitive(s, "AWAY")) {
    return CLIMATE_PRESET_AWAY;
  } else if (str_equals_case_insensitive(s, "BOOST")) {
    return CLIMATE_PRESET_BOOST;
  } else if (str_equals_case_insensitive(s, "COMFORT")) {
    return CLIMATE_PRESET_COMFORT;
  } else if (str_equals_case_insensitive(s, "ECO")) {
    return CLIMATE_PRESET_ECO;
  } else if (str_equals_case_insensitive(s, "SLEEP")) {
    return CLIMATE_PRESET_SLEEP;
  } else if (str_equals_case_insensitive(s, "ACTIVITY")) {
    return CLIMATE_PRESET_ACTIVITY;
  } else {
    ESP_LOGW(TAG, "Unrecognized preset %s", s.c_str());
    return CLIMATE_PRESET_NONE;
  }
}

ClimateAction action_from_string(const std::string &s) {
  if (str_equals_case_insensitive(s, "OFF")) {
    return CLIMATE_ACTION_OFF;
  } else if (str_equals_case_insensitive(s, "COOLING")) {
    return CLIMATE_ACTION_COOLING;
  } else if (str_equals_case_insensitive(s, "HEATING")) {
    return CLIMATE_ACTION_HEATING;
  } else if (str_equals_case_insensitive(s, "IDLE")) {
    return CLIMATE_ACTION_IDLE;
  } else if (str_equals_case_insensitive(s, "DRYING")) {
    return CLIMATE_ACTION_DRYING;
  } else if (str_equals_case_insensitive(s, "FAN")) {
    return CLIMATE_ACTION_FAN;
  } else {
    ESP_LOGW(TAG, "Unrecognized action %s", s.c_str());
    return CLIMATE_ACTION_OFF;
  }
}
```

polyfloyd:
@devildant Those strings reflect the respective enums defined by Esphome, which are fixed by themselves

devildant:
> @devildant Those strings reflect the respective enums defined by Esphome, which are fixed by themselves

It's a shame because it loses a lot of its interest.
But thx for the dev

ssieb:
Custom presets can be added.  But if you know how to control your AC, you should make a proper component for it.

devildant:
> Custom presets can be added.  But if you know how to control your AC, you should make a proper component for it.

in my case I'm trying to replace the tuya module of my air conditioner, I have the dp and the values, I have 5 shutter positions, in addition to the auto and off mode, for the ventilation speed I have 5 speeds, as well as 3 modes: auto, mute and turbo... I was just saying that if it was more open it would allow this kind of case to be mapped without reinventing the wheel, cpp it's been more than 15 years since I didn’t it it's going to be complicated ^^ whereas if we could simply map key value without control it would be much simpler ^^

ssieb:
Is there a reason that https://esphome.io/components/climate/tuya won't work for you?  It would better to extend that one if necessary instead of using something completely different.

devildant:
> Is there a reason that https://esphome.io/components/climate/tuya won't work for you? It would better to extend that one if necessary instead of using something completely different.

yes, i will create my own post, I'll stop polluting this topic

schmutzgeier:
This is very interesting for everyone with modbus controlled systems. Thank you for the effort. Would be great to see this merged.

esphome/components/template/climate/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+from esphome.components.number import Number
+from esphome.components.select import Select
+from esphome.components.sensor import Sensor
+from esphome.components.text_sensor import TextSensor
+import esphome.config_validation as cv
+from esphome.const import CONF_ACTION_ID, CONF_ID
+
+from .. import template_ns
+
+DEPENDENCIES = ["climate"]
```
jesserockz: Not required for platforms
```suggestion
```

esphome/components/template/climate/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+CONFIG_SCHEMA = climate.CLIMATE_SCHEMA.extend(
+    {
+        cv.GenerateID(CONF_ID): cv.declare_id(TemplateClimate),
+        cv.Required(CONF_CURRENT_TEMPERATURE_ID): cv.use_id(Sensor),
+        cv.Required(CONF_TARGET_TEMPERATURE_ID): cv.use_id(Number),
+        cv.Required(CONF_MODE_ID): cv.use_id(Select),
+        cv.Optional(CONF_FAN_MODE_ID): cv.use_id(Select),
+        cv.Optional(CONF_SWING_MODE_ID): cv.use_id(Select),
+        cv.Optional(CONF_PRESET_ID): cv.use_id(Select),
+        cv.Optional(CONF_ACTION_ID): cv.use_id(TextSensor),
```
jesserockz, line 31: This is not the way to build a template entity. Granted this is the most complex type of entity we have in ESPHome so it needs to be slightly different to the other types, but This is not flexible and relies on a certain structure. 

For example the target temperature should not be a Number, the climate has an internal variable for the target and if the user wants to use a `number` then they can with automations or a lambda that returns the state from another entity.

Same goes for all of these.

jesserockz (changes requested):
As SSieb pointed out, this component needs more work to be considered ready to merge into ESPHome.

SzczepanLeon:
Great add-on. Works OK with my Gree XK76.
Please add possibility to define by user other fan modes (for example TURBO).

PR #8033, opened 2025-01-03
New Component: Dallas 2406

### LINKED ISSUE #1099: Support for 1-wire  DS2408  binary_sensor and switch

I would like to use communication between 1-wire binary_sensor and switch using DS2408 8-ch or DS2406 2-ch. It can be used in all directions, such as Relay control, status LED control via switch and binary sensor.
https://datasheets.maximintegrated.com/en/ds/DS2408.pdf

The problem is that I only have 3-wire attached to the buttons on the wall. I tried to use ADC 1115 as recommended, but this solution only works when using 3 - 6 buttons (I have 13 of them), above 6 I already have a problem using the "on_multi_click" function and I also have a problem with ADC when using ttp223 touch sensor . With 13 buttons, the voltage difference cannot be perfectly distinguished (fluctuates when the touch sensor is pressed). DALLAS ds2408 would solve this problem.
The ADC can be used with analog buttons. When using the ttp223 touch sensor, there is a high voltage fluctuation, especially for "on_multi_click". This is not very applicable to "on_multi_click", because only 6 buttons worked on 100% from using 13 buttons. If the switch could also be integrated into Dallas, I would be able to send 1-wire status using an LED. When using Dallas, it is possible to push 2 buttons at the same time, which is not possible with ADC.

**Additional context**

The integration of this solution into ESPhome would facilitate work over 1-wire with the used binary sensor and switch, which would probably be appreciated by several users who have few cables or it could also be used for the old 3-wire electrical installation.
circuit

autoantwort:

New Component for the Dallas 2406: https://www.analog.com/media/en/technical-documentation/data-sheets/ds2406.pdf

## Types of changes

- part of fixes https://github.com/esphome/feature-requests/issues/1099 

- https://github.com/esphome/esphome-docs/pull/4569

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
ds2406:  
  - address: 0x99000000c695bd12
    update_interval: 1s

switch:
  - platform: ds2406
    channel: 1
    name: "Channel 1"

binary_sensor:
  - platform: ds2406
    channel: 2
    name: "Channel 2"

```

## Checklist:

  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs): https://github.com/esphome/esphome-docs/pull/4569


ssieb:
This should be renamed to something like `ds240x`.
You've implemented binary sensors under `sensor:`.  That's not going to work.
What you need to do is implement them as gpios.
```yaml
ds240x:
  id: switches
  address: <address here>

binary_sensor:
  - platform: gpio
    pin:
      ds240x: switches
      channel: 1

switch:
  - platform: gpio
    pin:
      ds240x: switches
      channel: 0
```

autoantwort:
> You've implemented binary sensors under sensor:. That's not going to work. 

Can you elaborate? It already works for me locally without problems. 

 > What you need to do is implement them as gpios.
 
 Where would I define the address and the update interval in your example?

ssieb:
> > You've implemented binary sensors under sensor:. That's not going to work.
> 
> Can you elaborate? It already works for me locally without problems.

You must have a `binary_sensor:` configured somewhere else.  But it will not be accepted like that.

> > What you need to do is implement them as gpios.
> 
> Where would I define the address and the update interval in your example?

I forgot about the address.  I've edited the previous comment for that.

A binary sensor does not have an update interval.  But yes, this is a slow device, so it probably won't work as a general gpio.  I guess you'll need something more like this:

```yaml
ds240x:
  id: switches
  address: <address here>
  update_interval: 500ms

binary_sensor:
  - platform: ds240x
    id: channel_1
    channel: 1

switch:
  - platform: ds240x
    id: channel_0
    channel: 0
```

autoantwort:
Ok so I guess the config will look something like
```yml
ds240x:
  - id: dev1
    address: <address here>
    update_interval: 500ms

binary_sensor:
  - platform: ds240x
    ds240x_id: dev1
    channel: 1

switch:
  - platform: ds240x
    ds240x_id: dev1
    channel: 0
```

Is there a similar construct at which I can take a look to have inspiration while coding? And afaics all the ds240x devices don't share any implementation details (except from being a 1-Wire device).

ssieb:
> Is there a similar construct at which I can take a look to have inspiration while coding?

There are a lot of kind of similar examples, but I can't think of one off-hand that has a switch as well.  Come to the #devs channel on the esphome discord server and we can help you with the details.

> And afaics all the ds240x devices don't share any implementation details (except from being a 1-Wire device).

Ok.  And I guess they're simple enough that there's not much code to share.

ssieb:
binary sensor also needs to be in a subdirectory.  And please rename to `ds2406`.

autoantwort:
> And please rename to `ds2406`.

Everywhere?

autoantwort:
Screenshot 2025-01-05 at 03 24 44
And in all the other components the file is also not in a subdir?

ssieb:
Everywhere except the descriptive log message.

ssieb:
The `status` one is only a binary sensor.
New components with sub-components must have this structure.  The `udp` one is a special case.

autoantwort:
So it gets `binary_sensor/init.py` and not `binary_sensor/binary_sensor.py` right ?

ssieb:
Just like you did with the switch part.  `__init__.py`

autoantwort:
Why should this be renamed to `ds_2406`, the other components are also called `dallas_xxx` ?

ssieb:
`ds2406`.  The only `dallas_*` one is `dallas_temp` because it's for multiple Dallas temperature sensors with different names.  The normal naming scheme for components is to use the chip name.

PR #8037, opened 2025-01-05
[lc709203f] Add battery monitor

### LINKED ISSUE #2040: support for "LiPoly battery monitor - LC709203"

I would like support for reading battery status with this sensor [Adafruit LC709203F LiPoly / LiIon Fuel Gauge and Battery Monitor - STEMMA JST PH & QT / Qwiic](https://www.adafruit.com/product/4712)

I have an [Adafruit ESP32-S2 Feather with BME280 Sensor - STEMMA QT - 4MB Flash + 2 MB PSRAM](https://www.adafruit.com/product/5303) that I am using for environmental sensing and it has that chip onboard for battery status monitoring.

**Additional context**

Logs esp32-s2-feather.yaml
```
INFO Reading configuration /config/esphome/esp32-s2-feather.yaml...
INFO Starting log output from esp32-s2-feather.local using esphome API
INFO Successfully connected to esp32-s2-feather.local
[14:13:52][I][app:102]: ESPHome version 2022.12.3 compiled on Dec 26 2022, 13:35:08
[14:13:52][C][wifi:504]: WiFi:
[14:13:52][C][wifi:362]:   Local MAC: 84:F7:03:D8:61:F0
[14:13:52][C][wifi:363]:   SSID: [redacted]
[14:13:52][C][wifi:364]:   IP Address: 192.168.33.173
[14:13:52][C][wifi:366]:   BSSID: [redacted]
[14:13:52][C][wifi:367]:   Hostname: 'esp32-s2-feather'
[14:13:52][C][wifi:369]:   Signal strength: -46 dB ▂▄▆█
[14:13:52][C][wifi:373]:   Channel: 11
[14:13:52][C][wifi:374]:   Subnet: 255.255.255.0
[14:13:52][C][wifi:375]:   Gateway: 192.168.33.1
[14:13:52][C][wifi:376]:   DNS1: 192.168.33.1
[14:13:52][C][wifi:377]:   DNS2: 0.0.0.0
[14:13:52][C][logger:293]: Logger:
[14:13:52][C][logger:294]:   Level: DEBUG
[14:13:52][C][logger:295]:   Log Baud Rate: 115200
[14:13:52][C][logger:296]:   Hardware UART: UART0
[14:13:52][C][i2c.arduino:052]: I2C Bus:
[14:13:52][C][i2c.arduino:053]:   SDA Pin: GPIO3
[14:13:52][C][i2c.arduino:054]:   SCL Pin: GPIO4
[14:13:52][C][i2c.arduino:055]:   Frequency: 50000 Hz
[14:13:52][C][i2c.arduino:058]:   Recovery: bus successfully recovered
[14:13:52][I][i2c.arduino:068]: Results from i2c bus scan:
[14:13:52][I][i2c.arduino:074]: Found i2c device at address 0x0B
[14:13:52][I][i2c.arduino:074]: Found i2c device at address 0x77
```
https://learn.adafruit.com/i2c-addresses/the-list says that that `0x0B` is the `LC709203` (`0x77` is the onboard `bme280`, an already supported sensor)

https://www.adafruit.com/product/5303 says
> Features:
> LiPoly battery monitor - LC709203 chip actively monitors your battery for voltage and state of charge / percentage reporting over I2C

ilikecake:

Implements the [LC709203F](https://www.adafruit.com/product/4712) battery monitor sensor component. I will also submit a PR to update the documentation with details.

## Types of changes

- [ ] Other

- resolves https://github.com/esphome/feature-requests/issues/2040

- esphome/esphome-docs#4800

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: lc709203f
    size: 2000
    voltage: 3.7
    battery_voltage:
      name: "Battery Voltage"
    battery_level:
      name: "Battery"
    temperature:
      name: "Pack Temperature"
      thermistor_b_constant: 0xA5A5
```

## Checklist:


ilikecake:
> Hey there @ilikecake, Thanks for submitting this pull request! Can you add yourself as a codeowner for this integration? This way we can notify you if a bug report for this integration is reported. In `__init__.py` of the integration, please add:
> 
> ```python
> CODEOWNERS = ["@ilikecake"]
> ```
> 
> And run `script/build_codeowners.py`
> 
> (message by NeedsCodeownersLabel)

If anyone is reading this, I could use help figuring out how to do this. I have added my handle to the `__init__` file, but I can't figure out how to run the script.

Also, I could use some info on what you want for the tests. I added a yaml file to the tests directory, but I don't think it is quite right.

tests/components/lc709203f/test.esp32-ard.yaml
```diff
@@ -0,0 +1,16 @@
+i2c:
+  - id: i2c_lc709203f
+    scl: 16
+    sda: 17
+
+sensor:
```
swoboda1337: Tests need to be refactored. Should have one common file and then substitutions for pins. Look at the current tests for examples.

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,328 @@
...
+      {
+          // A register was read from a device at the correct address, but it does not match
+          //  what we expect for the LC709203F
+          ESP_LOGE(TAG, "Device does not appear to be a LC709203F");
+          this->status_set_error("unrecognised");
+          this->mark_failed();
+          return;
+      }
+  }
+  */
```
swoboda1337: Why have dead/commented code, its not formatted correctly either.

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,328 @@
...
+    for (int i = 8; i; --i) {
+      crc = (crc & 0x80) ? (crc << 1) ^ polynomial : (crc << 1);
+    }
+  }
+  return crc;
+}
+
+void Lc709203f::set_pack_size(uint16_t pack_size) {
+  uint16_t pack_size_array[6] = {100, 200, 500, 1000, 2000, 3000};
+  uint16_t apa_array[6] = {0x08, 0x0B, 0x10, 0x19, 0x2D, 0x36};
```
swoboda1337: These two should be static const?

esphome/components/lc709203f/sensor.py
```diff
@@ -0,0 +1,98 @@
...
+    "3.8": LC709203FBatteryVoltage.LC709203F_BATTERY_VOLTAGE_3_8,
+}
+
+lc709203f = lc709203f_ns.class_("Lc709203f", cg.PollingComponent, i2c.I2CDevice)
+
+CONFIG_SCHEMA = (
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(lc709203f),
+            cv.Optional(CONF_SIZE, default="500"): cv.int_range(200, 3000),
```
swoboda1337: So you cant set values bewteen100mAh and 200mAh??

esphome/components/lc709203f/sensor.py
```diff
@@ -0,0 +1,98 @@
...
+    cg.add(var.set_pack_size(config.get(CONF_SIZE)))
+    cg.add(var.set_pack_voltage(BATTERY_VOLTAGE_OPTIONS[config[CONF_VOLTAGE]]))
+
+    if voltage_config := config.get(CONF_BATTERY_VOLTAGE):
+        sens = await sensor.new_sensor(voltage_config)
+        cg.add(var.set_voltage_sensor(sens))
+
+    if CONF_BATTERY_LEVEL in config:
+        sens = await sensor.new_sensor(config[CONF_BATTERY_LEVEL])
+        cg.add(var.set_battery_remaining_sensor(sens))
```
swoboda1337, line 88: Why is this if not consistent with the other two?
```
if CONF_BATTERY_LEVEL in config:
``` 
Vs 
```
if voltage_config := config.get(CONF_BATTERY_VOLTAGE):
```

esphome/const.py
```diff
@@ -874,6 +874,7 @@
 CONF_TEXT = "text"
 CONF_TEXT_SENSORS = "text_sensors"
 CONF_THEN = "then"
+CONF_THERMISTOR_B_CONSTANT = "thermistor_b_constant"
```
swoboda1337: Not sure its worth adding this to const can just define this in your python file. It is unlikely to be reused by anyone else.

tests/components/lc709203f/test.esp32-ard.yaml
ilikecake: Thanks for the review. Can you point me in the direction of an example of the current tests? I tried to follow the formatting from some of the other tests.

swoboda1337: > Thanks for the review. Can you point me in the direction of an example of the current tests? I tried to follow the formatting from some of the other tests.

https://github.com/esphome/esphome/tree/dev/tests/components/am2315c

You will see there is now a common and all tests include that common. This was changed not too long ago.

esphome/components/lc709203f/sensor.py
ilikecake: Good catch. That should allow values between 100 and 3000. TBH, even pack sizes above that range are probably okay, but the datasheet only provides calibration data for that range of sizes.

esphome/components/lc709203f/sensor.py
ilikecake: I suspect that is a remnant of the examples that I was using to write this. Is one of those methods preferred?

swoboda1337: Just make it consistent

ilikecake:
@swoboda1337, thanks for looking at this. I have made updates based on your review. Please let me know if you have any other comments.

ilikecake:
Note that this PR would close this [issue](https://github.com/esphome/feature-requests/issues/2040).

swoboda1337:
> N/A

If you add `- resolves https://github.com/esphome/feature-requests/issues/2040` to `Related issue or feature (if applicable):` it will link them and auto close the issue when merged.

esphome/components/lc709203f/sensor.py
kbx81: [The walrus is preferred.](https://developers.esphome.io/contributing/code/#python) Please change them all to use the walrus operator.

esphome/components/lc709203f/sensor.py
```diff
@@ -0,0 +1,99 @@
...
+                unit_of_measurement=UNIT_CELSIUS,
+                accuracy_decimals=2,
+                device_class=DEVICE_CLASS_TEMPERATURE,
+                state_class=STATE_CLASS_MEASUREMENT,
+                entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
+            ).extend(
+                {
+                    cv.Optional(
+                        CONF_THERMISTOR_B_CONSTANT, default="0x0D34"
+                    ): cv.int_range(0, 0xFFFF),
```
kbx81: Any reason this is in hex? Seems a bit...weird...here. 😄

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+  LOG_UPDATE_INTERVAL(this);
+  ESP_LOGCONFIG(TAG, "  Pack Size: %d mAH", this->pack_size_);
+  ESP_LOGCONFIG(TAG, "  Pack APA: 0x%02X", this->apa_);
+  if (this->pack_voltage_ == 0x0000) {
+    ESP_LOGCONFIG(TAG, "  Pack Rated Voltage: 3.8V");
+  } else {
+    // This is only true if the pack_voltage_ is 0x0001. The config validator should
+    //  have already made sure that this value can be only 0 or 1.
+    ESP_LOGCONFIG(TAG, "  Pack Rated Voltage: 3.7V");
+  }
```
kbx81: Instead of storing two copies of the same string:
```suggestion
  // This is only true if the pack_voltage_ is 0x0001. The config validator should
  //  have already made sure that this value can be only 0 or 1.
  ESP_LOGCONFIG(TAG, "  Pack Rated Voltage: 3.%sV", this->pack_voltage_ == 0x0000 ? "8" : "7");
```

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+uint8_t Lc709203f::get_register_(uint8_t register_to_read, uint16_t *register_value) {
+  i2c::ErrorCode return_code;
+  uint8_t read_buffer[6];
+
+  read_buffer[0] = (this->address_) << 1;
+  read_buffer[1] = register_to_read;
+  read_buffer[2] = ((this->address_) << 1) | 0x01;
+
+  for (int i = 0; i <= LC709203F_I2C_RETRY_COUNT; i++) {
```
kbx81:
```suggestion
  for (uint8_t i = 0; i <= LC709203F_I2C_RETRY_COUNT; i++) {
```

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  // Note: We don't actually send byte[0] of the buffer. We include it because it is
+  //  part of the CRC calculation.
+  write_buffer[0] = (this->address_) << 1;
+  write_buffer[1] = register_to_set;
+  write_buffer[2] = value_to_set & 0xFF;         // Low byte
+  write_buffer[3] = (value_to_set >> 8) & 0xFF;  // High byte
+  write_buffer[4] = this->crc8_(write_buffer, 4);
+
+  for (int i = 0; i <= LC709203F_I2C_RETRY_COUNT; i++) {
```
kbx81:
```suggestion
  for (uint8_t i = 0; i <= LC709203F_I2C_RETRY_COUNT; i++) {
```

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+  //  does not record power usage. If there is significant power consumption during sleep mode,
+  //  the pack RSOC will likely no longer be correct. Because of that, I do not implement
+  //  sleep mode on this device.
+  if (this->set_register_(LC709203F_IC_POWER_MODE, LC709203F_POWER_MODE_ON) == i2c::NO_ERROR) {
+    if (this->set_register_(LC709203F_APA, this->apa_) == i2c::NO_ERROR) {
+      if (this->set_register_(LC709203F_CHANGE_OF_THE_PARAMETER, this->pack_voltage_) == i2c::NO_ERROR) {
+        this->state_ = LC709203F_STATE_RSOC;
+      }
+    }
+  }
```
kbx81: Nothing specifically wrong here, but...how many levels deep are you going to stack these? 😅
```suggestion
  if (this->set_register_(LC709203F_IC_POWER_MODE, LC709203F_POWER_MODE_ON) != i2c::NO_ERROR) {
    return;
  }
  if (this->set_register_(LC709203F_APA, this->apa_) != i2c::NO_ERROR) {
    return;
  }
  if (this->set_register_(LC709203F_CHANGE_OF_THE_PARAMETER, this->pack_voltage_) != i2c::NO_ERROR) {
    return;
  }
  this->state_ = LC709203F_STATE_RSOC;
```

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+  //  until a device reset.
+  //
+  //  This device can be picky about I2C communication and can error out occasionally. The
+  //  get/set register functions impelment retry logic to retry the I2C transactions. The
+  //  initialization code checks the return code from those functions. If they don't return
+  //  NO_ERROR (0x00), that part of the initialization aborts and will be retried on the next
+  //  call to update().
+  ESP_LOGCONFIG(TAG, "Setting up LC709203F...");
+
+  this->state_ = LC709203F_STATE_INIT;
```
kbx81: Why not just initialize this in the header?

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+  } else if (this->state_ == LC709203F_STATE_INIT) {
+    // We should only get here if the init sequence failed during the setup() function.
+    //  This would likely occur because of a repeated failure on the I2C bus.
+    if (this->set_register_(LC709203F_IC_POWER_MODE, LC709203F_POWER_MODE_ON) == i2c::NO_ERROR) {
+      if (this->set_register_(LC709203F_APA, this->apa_) == i2c::NO_ERROR) {
+        if (this->set_register_(LC709203F_CHANGE_OF_THE_PARAMETER, this->pack_voltage_) == i2c::NO_ERROR) {
+          this->state_ = LC709203F_STATE_RSOC;
+        }
+      }
+    }
```
kbx81: Same thought as I have in `setup()`

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+    } else {
+      // The device expects to get updates to the temperature in this mode.
+      //  I am not doing that now. The temperature register defaults to 25C.
+      //  In theory, we could have another temperature sensor and have ESPHome
+      //  send updated temperature to the device occasionally, but I have no idea
+      //  how to make that happen.
+      if (this->set_register_(LC709203F_STATUS_BIT, 0x0000) == i2c::NO_ERROR) {
+        this->state_ = LC709203F_STATE_NORMAL;
+      }
+    }
```
kbx81:
```suggestion
    } else if (this->set_register_(LC709203F_STATUS_BIT, 0x0000) == i2c::NO_ERROR) {
      // The device expects to get updates to the temperature in this mode.
      //  I am not doing that now. The temperature register defaults to 25C.
      //  In theory, we could have another temperature sensor and have ESPHome
      //  send updated temperature to the device occasionally, but I have no idea
      //  how to make that happen.
      this->state_ = LC709203F_STATE_NORMAL;
    }
```

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+    //  transaction. This is bad in this case and will result in reading nothing but 0xFFFF
+    //  from the registers.
+    return_code = this->read_register(register_to_read, &read_buffer[3], 3, false);
+    if (return_code != i2c::NO_ERROR) {
+      // Error on the i2c bus
+      snprintf(this->error_code_buffer_, 50, "Error code %d when reading from register 0x%02X", return_code,
+               register_to_read);
+      this->status_set_warning(this->error_code_buffer_);
+    } else {
+      if (this->crc8_(read_buffer, 5) != read_buffer[5]) {
```
kbx81: Another if/else I think you can combine. 🙂

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+}
+
+uint8_t Lc709203f::crc8_(uint8_t *byte_buffer, uint8_t length_of_crc) {
+  uint8_t crc = 0x00;
+  const uint8_t polynomial(0x07);
+
+  for (int j = length_of_crc; j; --j) {
+    crc ^= *byte_buffer++;
+
+    for (int i = 8; i; --i) {
```
kbx81:
```suggestion
  for (uint8_t j = length_of_crc; j; --j) {
    crc ^= *byte_buffer++;

    for (uint8_t i = 8; i; --i) {
```

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+  static const uint16_t PACK_SIZE_ARRAY[6] = {100, 200, 500, 1000, 2000, 3000};
+  static const uint16_t APA_ARRAY[6] = {0x08, 0x0B, 0x10, 0x19, 0x2D, 0x36};
+  float slope;
+  float intercept;
+
+  this->pack_size_ = pack_size;  // Pack size in mAH
+
+  // The size is used to calculate the 'Adjustment Pack Application' number.
+  // Here we assume a type 01 or type 03 battery and do a linear curve fit to find the APA.
+  for (int i = 0; i < sizeof(PACK_SIZE_ARRAY) / sizeof(PACK_SIZE_ARRAY[0]); i++) {
```
kbx81: I think you can use `uint8_t` here, too, since, if I'm understanding this correctly, you're basically just doing `for (uint8_t i = 0; i < 6; i++)`...
```suggestion
  for (uint8_t i = 0; i < sizeof(PACK_SIZE_ARRAY) / sizeof(PACK_SIZE_ARRAY[0]); i++) {
```

kbx81 (dismissed):

esphome/components/lc709203f/sensor.py
ilikecake: Hmm, not really. I just matched how they called out the default in the datasheet. Should I make this a required variable? If the user is using a thermistor, this value needs to be set, and I don't really know if that constant would be suitable for any generic thermistor. It might be better to make that value required if the user is using the thermistor.

kbx81: For the [NTC sensor](https://github.com/esphome/esphome/blob/1c5dc63eb4df01935d19652ab5e9b74dbe329486/esphome/components/ntc/sensor.py#L92), it is required.
I don't think we should assume any defaults here if the user is providing this part.

kbx81: ...which reminds me -- please use the constant from `const.py` here.

ilikecake: I will change it to required. I don't see a config variable in const.py for this, which one do you mean? Should I use the same `CONF_B_CONSTANT = "b_constant"` from the NTC sensor you linked?

kbx81: > Should I use the same `CONF_B_CONSTANT = "b_constant"` from the NTC sensor you linked?

Yes please. 🙂

esphome/components/lc709203f/lc709203f.h
```diff
@@ -0,0 +1,57 @@
...
+#include "esphome/components/sensor/sensor.h"
+#include "esphome/components/i2c/i2c.h"
+
+namespace esphome {
+namespace lc709203f {
+
+static const uint8_t LC709203F_STATE_INIT = 0x01;
+static const uint8_t LC709203F_STATE_RSOC = 0x02;
+static const uint8_t LC709203F_STATE_TEMP_SETUP = 0x03;
+static const uint8_t LC709203F_STATE_NORMAL = 0x00;
```
kbx81: Any reason these aren't an `enum`? That might make more sense since you're tracking state -- there should only be one state at a time.

esphome/components/lc709203f/lc709203f.h
```diff
@@ -0,0 +1,57 @@
...
+  sensor::Sensor *temperature_sensor_{nullptr};
+  uint16_t pack_size_;
+  uint16_t apa_;
+  uint16_t b_constant_;
+  uint8_t state_ = LC709203F_STATE_INIT;
+  uint16_t pack_voltage_;
+
+  // A buffer to store error code messages. We put this here so as to not have to
+  //  duplicate buffers in the different functions
+  char error_code_buffer_[50];
```
kbx81: Don't do this here. You're tying up 50 bytes of RAM per instance of the component all the time. You really only need this memory when you're printing an error which I see being done in only two methods. Just allocate it there as needed.

esphome/components/lc709203f/lc709203f.h
```diff
@@ -0,0 +1,57 @@
...
+  void dump_config() override;
+
+  void set_pack_size(uint16_t pack_size);
+  void set_thermistor_b_constant(uint16_t b_constant);
+  void set_pack_voltage(LC709203FBatteryVoltage pack_voltage);
+  void set_voltage_sensor(sensor::Sensor *voltage_sensor) { voltage_sensor_ = voltage_sensor; }
+  void set_battery_remaining_sensor(sensor::Sensor *battery_remaining_sensor) {
+    battery_remaining_sensor_ = battery_remaining_sensor;
+  }
+  void set_temperature_sensor(sensor::Sensor *temperature_sensor) { temperature_sensor_ = temperature_sensor; }
```
kbx81, line 36: Please ensure you have the `this->` prefix on all member variable accesses & calls to class methods.
```suggestion
  void set_voltage_sensor(sensor::Sensor *voltage_sensor) { this->voltage_sensor_ = voltage_sensor; }
  void set_battery_remaining_sensor(sensor::Sensor *battery_remaining_sensor) {
    this->battery_remaining_sensor_ = battery_remaining_sensor;
  }
  void set_temperature_sensor(sensor::Sensor *temperature_sensor) { this->temperature_sensor_ = temperature_sensor; }
```

kbx81:
Bumping https://github.com/esphome/esphome/pull/8037/files#r2092316064

esphome/components/lc709203f/lc709203f.h
ilikecake: Sorry, I am used to my microcontrollers having a static memory map. If I move the declaration of that array to the start of the `get_register_` function, does that help? Wont it still get allocated every time the function is called? What is the preferred way to handle this?

kbx81: Just do something like [this](https://github.com/esphome/esphome/blob/9dd404598464d103392c29d13e946e65fd2dedd3/esphome/components/ch422g/ch422g.cpp#L97)

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+  //  Assuming no errors, the whole process should occur during the setup() function and
+  //  the first two calls to update(). After that, the part should remain in normal mode
+  //  until a device reset.
+  //
+  //  This device can be picky about I2C communication and can error out occasionally. The
+  //  get/set register functions impelment retry logic to retry the I2C transactions. The
+  //  initialization code checks the return code from those functions. If they don't return
+  //  NO_ERROR (0x00), that part of the initialization aborts and will be retried on the next
+  //  call to update().
+  ESP_LOGCONFIG(TAG, "Setting up LC709203F...");
```
kbx81:
```suggestion
  ESP_LOGCONFIG(TAG, "Running setup");
```

esphome/components/lc709203f/lc709203f.cpp
```diff
@@ -0,0 +1,302 @@
...
+}
+
+void Lc709203f::dump_config() {
+  ESP_LOGCONFIG(TAG, "LC709203F:");
+  LOG_I2C_DEVICE(this);
+  if (this->is_failed()) {
+    // Nothing in this code actually sets the device to failed, so this should never trigger.
+    //  I am leaving this in incase I want to implement a check during init to verify the
+    //  correct device is on the bus.
+    ESP_LOGCONFIG(TAG, "  I2C Device at 0x0B does not appear to be a LC709203F");
```
kbx81:
```suggestion
    ESP_LOGCONFIG(TAG, ESP_LOG_MSG_COMM_FAIL);
```

kbx81: @ilikecake will you implement this check now? If not, let's remove this block of code and I'll merge the PR. You can always do a follow-up to add enhancements.

ilikecake: I don't know a good way to check this for this device. If there were some sort of ID register that we could read to verify the device is what we think, that would be great. In this case, I don't think there is any reliable way to do that. There is an 'IC Version' register, but it is unclear if that is specific to an individual chip or the same for every chip. I will remove this code for now.

ilikecake: On second thought, while I don't think there is a good way to verify device ID during init, there might be a use for `mark_failed()`. The device can get confused if you poll it too often and respond incorrectly over i2c. That is why I implement 10 retries and then return an error code during register read and write operations. The way the code works now, the higher level functions will ignore i2c transactions that don't return the `NO_ERROR` code, but they will try again at the next poll time. Do you think it would be worthwhile to use `mark_failed()` if the device has 10 failed i2c transactions in a row? FWIW, I have never seen more than 2-3 failed transactions consecutively in my testing.

kbx81: I think, in general, we only use `mark_failed()` for unrecoverable failures. If you can recover communication by retrying reads/writes, then there's probably no reason to `mark_failed()` and instead maybe just set the "warning" status/flag.

ilikecake: ok, that makes sense. The code will already send warnings on i2c bus errors, so I removed that block of code.

kbx81 (approved):
LGTM -- thanks for working on this! 🍻

PR #8041, opened 2025-01-06
[esp32_rmt] Use USE_ESP_IDF & USE_ARDUINO to match config code

swoboda1337:

Switch from ESP_IDF_VERSION_MAJOR to USE_ESP_IDF & USE_ARDUINO to match config code and avoid confusion about IDF 4 support. 

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/esp32_rmt_led_strip/led_strip.cpp
+++ b/esphome/components/esp32_rmt_led_strip/led_strip.cpp
@@ -42,7 +42,7 @@ void ESP32RMTLEDStripLightOutput::setup() {
     return;
   }
 
-#if ESP_IDF_VERSION_MAJOR >= 5
+#ifdef USE_ESP_IDF
   RAMAllocator<rmt_symbol_word_t> rmt_allocator(this->use_psram_ ? 0 : RAMAllocator<rmt_symbol_word_t>::ALLOC_INTERNAL);
 
   // 8 bits per byte, 1 rmt_symbol_word_t per bit + 1 rmt_symbol_word_t for reset
@@ -145,7 +145,7 @@ void ESP32RMTLEDStripLightOutput::write_state(light::LightState *state) {
 
   ESP_LOGVV(TAG, "Writing RGB values to bus...");
 
-#if ESP_IDF_VERSION_MAJOR >= 5
+#ifdef USE_ESP_IDF
   esp_err_t error = rmt_tx_wait_all_done(this->channel_, 1000);
 #else
   esp_err_t error = rmt_wait_tx_done(this->channel_, pdMS_TO_TICKS(1000));
@@ -162,7 +162,7 @@ void ESP32RMTLEDStripLightOutput::write_state(light::LightState *state) {
   size_t size = 0;
   size_t len = 0;
   uint8_t *psrc = this->buf_;
-#if ESP_IDF_VERSION_MAJOR >= 5
+#ifdef USE_ESP_IDF
   rmt_symbol_word_t *pdest = this->rmt_buf_;
 #else
   rmt_item32_t *pdest = this->rmt_buf_;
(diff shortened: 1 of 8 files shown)


gnumpi:
If you want to end the support for IDF 4 I suggest adding a validation to the yaml config check which notifies the user if he tries using that component with IDF 4.  Also, for developers it would be more clear if you checked against the Arduino framework instead of not using the new rmt driver, which is implicitly the case when not supporting IDF 4 anymore and using the IDF framework, right?(https://github.com/esphome/esphome/blob/dc5b4087486c8d707c7b89379c7bf8bdd367c86a/esphome/components/esp32_rmt_led_strip/light.py#L68C5-L68C78)

An alternative to this PR would be to just fix the config check and keep supporting IDF4, see PR #8035.

@jesserockz , has with setting the default framework version to IDF 5.1 the support for IDF4 officially ended? If so, it should be stated as a breaking change in the release notes. Are you aware of any other component that will work with IDF 5.1 only, or would this be the first one if this PR gets accepted?

kbx81:
> @jesserockz , has with setting the default framework version to IDF 5.1 the support for IDF4 officially ended? If so, it should be stated as a breaking change in the release notes. Are you aware of any other component that will work with IDF 5.1 only, or would this be the first one if this PR gets accepted?

I'm leaning against specifically _breaking_ things in terms of IDF 4 right now--if only because we only just made the switch--but do understand that, in the longer term, continuing _any_ support for IDF 4 is going to become more and more complex/difficult. Espressif is officially no longer supporting IDF 4 and continuing to use it will present more and more issues, particularly as it relates to support for ESP32 variants.

Jesse can/will correct me if I'm incorrect with this, but ESPHome as a project will only officially support _one_ IDF version at a time. That's not to say that other versions _won't/can't work,_ but even just the basic compile tests we have as a part of our CI right now take _several hours_ to run, and testing against multiple IDF versions would increase that run time by multiples -- so supporting multiple IDF versions is just not really feasible as things are right now.

swoboda1337:
How long would we keep supporting 4x? With the January release dropped there will be two months warning before 4x stops working for a few components. 

I am fine with whatever is decided however there are over 400 open prs. Personally would rather spend time moving things forward.

gnumpi:
I completely agree that, in the long term, only a single framework version should be supported and maintained. However, to ensure a smooth transition and avoid breaking changes, I propose the following steps:

**New Integrations:** 
New integrations should no longer require support for IDF 4. The recommended version was set to IDF 5.1 starting from December 2024. No additional effort should be invested in integrating specifically for IDF 4.

**Integration Updates:** 
When updating integrations to support or leverage IDF 5 features, a fallback mechanism should be implemented for compatibility with IDF 4. Since the existing IDF 4 code is already functional, this approach avoids breaking changes while allowing developers to benefit from IDF 5 advancements.

**Bug Fixes:** 
Developers should have the flexibility to decide whether to address bugs in both IDF 4 and IDF 5 versions, depending on the complexity and effort involved. If fixing both is not feasible, it is acceptable to fix only the IDF 5 variant, leaving IDF 4 users with the current (possibly buggy) state of the integration. This approach also avoids introducing breaking changes.

To ensure clarity and consistency, a specific deadline can be set—such as six months from now or the release date of the next IDF version—after which all IDF 4-specific code can be removed entirely.

kbx81:
Part of the problem is that we don't have a _documented_ way to approach this...but that's coming...Jesse and I will be working on some developer docs over the next few weeks/months. Our approach/policies for this type of stuff will inevitably be outlined there. 😄

PR #8051, opened 2025-01-08
Add multicast support to udp component

HeMan:

This gives multicast support to udp component

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4576

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:
This should be split into two PRs - one to implement the new validation functions, then one for the UDP component.

HeMan:
> This should be split into two PRs - one to implement the new validation functions, then one for the UDP component.

The validator is now in https://github.com/esphome/esphome/pull/8072

esphome/components/udp/sensor.py
```diff
@@ -24,4 +25,7 @@ async def to_code(config):
     var = await new_sensor(config)
     comp = await cg.get_variable(config[CONF_UDP_ID])
     remote_id = str(config.get(CONF_REMOTE_ID) or config.get(CONF_ID))
-    cg.add(comp.add_remote_sensor(config[CONF_PROVIDER], remote_id, var))
+    listen_address = str(config.get(CONF_LISTEN_ADDRESS))
+    cg.add(
+        comp.add_remote_sensor(config[CONF_PROVIDER], remote_id, listen_address, var)
+    )
```
clydebarrow: Why is this being set in the sensor config, when it's a singleton property of the hub component?

clydebarrow (changes requested):

esphome/components/udp/__init__.py
```diff
@@ -62,6 +63,9 @@ def sensor_validation(cls: MockObjClass):
 PROVIDER_SCHEMA = cv.Schema(
     {
         cv.Required(CONF_NAME): cv.valid_name,
+        cv.Optional(
+            CONF_LISTEN_ADDRESS, default="255.255.255.255"
+        ): cv.ipv4address_multi_broadcast,
```
clydebarrow: Looking at the code some more, there is only one listen socket in use, with potentially multiple multicast addresses added to it. That's fine, but there's nothing limiting those addresses to a particular provider - any packet received on the socket will be processed in exactly the same way irrespective of the destination address. So the listen addresses shouldn't be configured under the provider list, they should be configured directly on the `usb:` element.

clydebarrow (changes requested):
More to come...

esphome/components/udp/udp_component.cpp
```diff
@@ -256,6 +256,21 @@ void UDPComponent::setup() {
...
+      ESP_LOGVV(TAG, "Join multicast %s", listen_address.str().c_str());
+      err = this->listen_socket_->setsockopt(IPPROTO_IP, IP_ADD_MEMBERSHIP, &imreq, sizeof(struct ip_mreq));
+      if (err < 0) {
+        ESP_LOGE(TAG, "Failed to set IP_ADD_MEMBERSHIP. Error %d", errno);
+        this->mark_failed();
+        this->status_set_error("Failed to set IP_ADD_MEMBERSHIP");
+        return;
+      }
+    }
+#endif
```
clydebarrow: If using a multicast listen address, should not bind to the broadcast address, use one of the multicast addresses instead. This works fine on host. Some other minor changes. I can't propose this as a comment since it involves code you didn't touch. I can just push if you like.

```
    server.sin_family = AF_INET;
    server.sin_addr.s_addr = ESPHOME_INADDR_ANY;
    server.sin_port = htons(this->port_);

    for (const auto &listen_address : this->listen_addresses_) {
      struct ip_mreq imreq = {};
      imreq.imr_interface.s_addr = ESPHOME_INADDR_ANY;
      inet_aton(listen_address.str().c_str(), &imreq.imr_multiaddr);
      server.sin_addr.s_addr = imreq.imr_multiaddr.s_addr;
      ESP_LOGV(TAG, "Join multicast %s", listen_address.str().c_str());
      err = this->listen_socket_->setsockopt(IPPROTO_IP, IP_ADD_MEMBERSHIP, &imreq, sizeof(imreq));
      if (err < 0) {
        ESP_LOGE(TAG, "Failed to set IP_ADD_MEMBERSHIP. Error %d", errno);
        this->mark_failed();
        this->status_set_error("Failed to set IP_ADD_MEMBERSHIP");
        return;
      }
    }
    err = this->listen_socket_->bind((struct sockaddr *) &server, sizeof(server));
    if (err != 0) {
      ESP_LOGE(TAG, "Socket unable to bind: errno %d", errno);
      this->mark_failed();
      this->status_set_error("Unable to bind socket");
      return;
    }
```

clydebarrow (changes requested):
With these changes, plus some test code for the listen addresses, it should be good to go. Tested here on ESP32 and host.

esphome/components/udp/udp_component.cpp
HeMan: It does however not work when you try to listen to multiple multicast addresses, it's only listening to the bound address. Unicast is also disable with this.

esphome/components/udp/udp_component.h
```diff
@@ -143,6 +145,7 @@ class UDPComponent : public PollingComponent {
   std::map<std::string, std::map<std::string, binary_sensor::BinarySensor *>> remote_binary_sensors_{};
 #endif
 
+  std::vector<network::IPAddress> listen_addresses_{};
```
clydebarrow: Now that only one listen address is allowed, this should be an `optional` rather than a vector.

clydebarrow (approved):

PR #8060, opened 2025-01-09
Enable udp to work (on ipv4) when ipv6 is enabled

HeMan:

UDP component didn't work when IPv6 was enabled, not even on IPv4.

NOTE: UDP component listens to broadcast address 255.255.255.255 and IPv6 doesn't have the concept of broadcast at all. To be able to use anything similar UDP component need to have multicast support, which is not part of this PR.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/udp/udp_component.cpp
+++ b/esphome/components/udp/udp_component.cpp
@@ -245,7 +245,9 @@ void UDPComponent::setup() {
     }
     struct sockaddr_in server {};
 
-    socklen_t sl = socket::set_sockaddr((struct sockaddr *) &server, sizeof(server), "0.0.0.0", this->port_);
+    server.sin_family = AF_INET;
+    server.sin_addr.s_addr = ESPHOME_INADDR_ANY;
+    server.sin_port = htons(this->port_);
     if (sl == 0) {
       ESP_LOGE(TAG, "Socket unable to set sockaddr: errno %d", errno);
       this->mark_failed();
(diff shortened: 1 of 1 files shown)


esphome/components/udp/udp_component.cpp
```diff
@@ -245,7 +245,7 @@ void UDPComponent::setup() {
     }
     struct sockaddr_in server {};
 
-    socklen_t sl = socket::set_sockaddr_any((struct sockaddr *) &server, sizeof(server), this->port_);
+    socklen_t sl = socket::set_sockaddr((struct sockaddr *) &server, sizeof(server), "0.0.0.0", this->port_);
```
clydebarrow: ~This just hard-codes the IPV4 broadcast address. What if you actually want to use IPV6? I didn't do any testing of the udp component with IPV6 and I'm sure it doesn't work right now, but it would be a better fix than this. Of course it may be impossible or too hard, in which case this change would be a reasonable work-around.~

I see you addressed this in the description above.
```suggestion
    server.sin_family = AF_INET;
    server.sin_addr.s_addr = ESPHOME_INADDR_ANY;
    server.sin_port = htons(this->port_);
```

HeMan: Yeah, I'll address IPv6 unicast and multicast in a separate PR, but for now I just like to be able to have it at least use IPv4 in a dual stack network.

clydebarrow (dismissed):

clydebarrow (approved):

PR #8064, opened 2025-01-10
Only warn if the component blocked for a longer time than the last time

numo68:

Some time ago the ESPHome code was changed so that every time the component blocks for more than 50ms, a warning is output. The rationale was that one should not have to bump the warning level to debug, as it slows the device too much.

Unfortunately sometimes one has a device that works that way, but wants to develop something else. The warnings from this code are then buried inside the component blocking ones.

This PR changes the behavior of `Component` and `WarnIfComponentBlockingGuard` so that a warning is only output if the time the component has blocked is more than 10ms above when the last warning was output. This still gives the information needed, but does not produce that much output.

In case the component developer wants to reset the limit - for example because some operation is knowingly long - the `warn_if_blocking_over_` member can be simply reset.

## Types of changes

- [x] Other

- relates to [https://github.com/esphome/esphome/issues/9517]

N/A

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

N/A

## Checklist:


DjordjeMandic:
I like the idea but i would still limit it. If another upper limit const value of block time is passed, inform always. Maybe limit it to 300ms? 

Also maybe dump config should show current value of warn_if_blocking_over_ which would be limited to eg 300ms

numo68:
Yes, there are many possibilities, per-component limits, time-based silencing etc, tweaking from the configuration, etc. I wanted to keep it simple, with little memory added per component, and perhaps start the discussion how people are actually using the information right now.

Fact is, this *is* debugging-level information that was made to warning just because it was hard to use it otherwise. It is the *component developer* that is interested in it, and mostly for that component only. The developer of an ESPHome-based solution usually does not care much if at all, and cannot do anything about it anyway. What he/she cares about is that his own `ESP_LOGW` scattered over the lambdas are not buried in irrelevant information. 

My patch is thus necessarily a hack of a hack :) The correct solution would be to selectively enable/disable debugging features, per feature and per component. Of course the cost/benefit is debatable here and most probably not worth the effort.

FWIW, the problem is not unique; I do embedded development for living and we also had to deal with spammy but nevertheless needed log information. In our case we solved the problem with having an in-memory log for them, keeping the last message and the count, with the first one also being logged to the standard log. Of course that is out of question with the memory we have here.

@jesserockz could you please comment as well?

DjordjeMandic:
You're right. Since `warn_if_blocking_over_` is available in all derived classes, it can be set by each component individually. I like the idea.

DjordjeMandic (approved):

esphome/core/component.cpp
```diff
@@ -231,10 +241,11 @@ void PollingComponent::set_update_interval(uint32_t update_interval) { this->upd
 WarnIfComponentBlockingGuard::WarnIfComponentBlockingGuard(Component *component)
     : started_(millis()), component_(component) {}
 WarnIfComponentBlockingGuard::~WarnIfComponentBlockingGuard() {
-  uint32_t now = millis();
-  if (now - started_ > 50) {
+  uint32_t blocking_time = millis() - started_;
+  if ((component_ != nullptr) ? component_->should_warn_of_blocking(blocking_time)
+                              : (blocking_time > WARN_IF_BLOCKING_OVER_MS)) {
```
jesserockz: Please split the ternary outside of this condition

jesserockz (changes requested):

esphome/core/component.cpp
numo68: Sorry, I don't understand - split how? Do you mean computing the bool before, then have just one term in the if? Thanks.

esphome/core/component.cpp
```diff
@@ -231,10 +241,11 @@ void PollingComponent::set_update_interval(uint32_t update_interval) { this->upd
 WarnIfComponentBlockingGuard::WarnIfComponentBlockingGuard(Component *component)
     : started_(millis()), component_(component) {}
 WarnIfComponentBlockingGuard::~WarnIfComponentBlockingGuard() {
-  uint32_t now = millis();
-  if (now - started_ > 50) {
+  uint32_t blocking_time = millis() - started_;
```
DjordjeMandic:
```suggestion
  uint32_t blocking_time = millis() - this->started_;
```

esphome/core/component.cpp
DjordjeMandic:
```suggestion
  bool should_warn = blocking_time > WARN_IF_BLOCKING_OVER_MS;
  if (this->component_ != nullptr) {
    should_warn = this->component_->should_warn_of_blocking(blocking_time);
  } 
  if (should_warn) {
```

```suggestion
  bool should_warn;
  if (this->component_ != nullptr) {
    should_warn = this->component_->should_warn_of_blocking(blocking_time);
  } else {
    should_warn = blocking_time > WARN_IF_BLOCKING_OVER_MS;
  } 
  if (should_warn) {
```
Give it another name if you don't like it :)

Coding from phone is hard 😂

DjordjeMandic (changes requested):

DjordjeMandic (approved):

numo68:
fwiw, I have rebased to fix the "This branch must not contain merge commits", no other changes. However, the message is still there, I don't know what to do about it...

esphome/core/component.cpp
```diff
@@ -39,6 +39,9 @@ const uint32_t STATUS_LED_OK = 0x0000;
 const uint32_t STATUS_LED_WARNING = 0x0100;
 const uint32_t STATUS_LED_ERROR = 0x0200;
 
+const uint32_t WARN_IF_BLOCKING_OVER_MS = 50U;       ///< Initial blocking time allowed without warning
+const uint32_t WARN_IF_BLOCKING_INCREMENT_MS = 10U;  ///< How long the blocking time must be larger to warn again
```
oliv3r, line 43: While I think it's good to add documentation, this isn't done anywhere else within this file; just a thought ...

esphome/core/component.h
```diff
@@ -65,6 +65,8 @@ extern const uint32_t STATUS_LED_ERROR;
 
 enum class RetryResult { DONE, RETRY };
 
+extern const uint32_t WARN_IF_BLOCKING_OVER_MS;
```
oliv3r, line 68: this looks a little hacky, why not make it constexpr uint32_t here; this is the header after all, and juts 'use' it in the source?

oliv3r (approved):

oliv3r:
I was also thinking if we cannot also configure the timeout on the component part, but didn't see easily how to pass anything from the component to the WarnIfComponentBlockingGuard, they just share the same cpp file. This is very similar to your suggestion of doing `warn_if_blocking_over_`. Regardless, I think it makes sense that if a component is known to be slow (bit-banged read/pin-wait for example), spamming the log is not useful, your change helps in writing it once, but may still end up in spam if you get variations (though you'd end up clamping to the max, so maybe not so loud). Setting a timeout for your component, that tends to be ~100ms, might be also useful.

esphome/core/component.h
numo68: well, I just used the style that the source uses for a dozen of other constants. Not sure what is nicer, to keep it or to change it...

esphome/core/component.cpp
oliv3r: @numo68 I think that's what @jesserockz means :)

oliv3r:
> fwiw, I have rebased to fix the "This branch must not contain merge commits", no other changes. However, the message is still there, I don't know what to do about it...

If you do a `git rebase  upstream/dev` (assuming upstream is the name of the remote linking to esphome, can also be `origin` of course), followed by a push, you should be good to go

numo68:
> If you do a `git rebase upstream/dev` (assuming upstream is the name of the remote linking to esphome, can also be `origin` of course), followed by a push, you should be good to go

I am pretty sure I did just that, `git fetch --all`, `git rebase upstream/dev`, `git push --force`, I even still have that in the shell history and the git graph looks like I expect it to. Seems like some github glitch. I'll now try to do it through the button here at github.

numo68:
> I'll now try to do it through the button here at github.

Well, nope...

jesserockz (approved):

domoticasalve:
How can i implement this change into my esphome builder in home assistant? I have a problem with the actual build and my display tents to flick because of this problem

DjordjeMandic:
You should have beta and dev variants of hassio addons. Install beta or dev or wait until 2025.5.0 release

PR #8067, opened 2025-01-11
[lvgl] Fix keyboard module not compiling

Valcob:

While compiling keyboard module an error is thrown:

Error: 'mode'
image

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
- keyboard:
    id: keyboardID
    x: 0
    y: 0
    align: LV_ALIGN_BOTTOM_MID

```

## Checklist:

### DIFF
--- a/esphome/components/lvgl/widgets/keyboard.py
+++ b/esphome/components/lvgl/widgets/keyboard.py
@@ -41,7 +41,8 @@ class KeyboardType(WidgetType):
         lvgl_components_required.add("KEY_LISTENER")
         lvgl_components_required.add(CONF_KEYBOARD)
         add_lv_use("btnmatrix")
-        await w.set_property(CONF_MODE, await KEYBOARD_MODES.process(config[CONF_MODE]))
+        if mode := config.get(CONF_MODE):
+            await w.set_property(CONF_MODE, await KEYBOARD_MODES.process(mode))
         if ta := await get_widgets(config, CONF_TEXTAREA):
             await w.set_property(CONF_TEXTAREA, ta[0].obj)
 


clydebarrow:
What is this supposed to fix? The sample yaml provided does throw an error, but it's because of the invalid LV_ALIGN, not mode (which defaults to TEXT_UPPER).

Valcob:
> What is this supposed to fix? The sample yaml provided does throw an error, but it's because of the invalid LV_ALIGN, not mode (which defaults to TEXT_UPPER).

sorry for wrong config,  my bad copy-paste was not copypasting :). Also I've added a screenshot of the issue

clydebarrow:
What version of ESPHome are you using? That error does not happen with the current version (and cannot, since mode has a default value.)

Valcob:
> What version of ESPHome are you using? That error does not happen with the current version (and cannot, since mode has a default value.)

it's the latest release 
image

clydebarrow:
Well, I can't reproduce a problem with that version either. Do you have a complete yaml that demonstrates it? Using any external components?

clydebarrow:
And BTW instead of screenshots, copy and paste the actual text.

Valcob:
> Well, I can't reproduce a problem with that version either. Do you have a complete yaml that demonstrates it? Using any external components?

sure here is the minimal code to reproduce the issue:
````
lvgl:
  bg_color: 0 
  buffer_size: 1 #0.25

  displays:
    - main_display
  touchscreens:
    - main_touch
  pages:
    - id: main_page
      widgets:
         - label:
            text: 'test'
         - keyboard:
            id: keyboardID2
            on_ready:
              then:
                - lvgl.widget.update:
                    id: keyboardID2
                    hidden: true
````


easy-peasy  on_ready is breaking stuff

clydebarrow:
Finally! It would have saved a great deal of time if you had provided that example in the first place. But your fix isn't correct, it just masks the underlying problem. I need to look into it further.

Can I suggest that in future when encountering a problem you raise an issue first - with the actual yaml demonstrating the problem! - so there can be a proper discussion about it.

clydebarrow:
~I'm gonna re-open this because~ the change you suggested is needed, but for a slightly different issue.

So the problem you ran into is that `lvgl.widget.update` should only set widget style properties and flags, but it's also calling the `to_code()` function for the widget type. That is generally harmless because the schema for `lvgl.widget.update` doesn't include any widget-specific properties, but it's still wrong. Fixing that alone would solve your problem, and I'll raise a separate PR for that.

But the other problem this highlights is that the keyboard config includes a default for the mode, which is fine for initial setup, but it's also used for `lvgl.keyboard.update` so unless a mode is provided it will default to `TEXT_UPPER` which may be wrong. So it needs a different schema for the modify operation. That in turn means your suggested change *is* needed since the mode may be absent when an update is done.

~So I'm going to complete that fix here since you've done half the work already.~ Oh, wait I can't - you deleted the branch already. Never mind, I'll include it in the other.

So, thanks for raising this, it's been useful. The process could just have been a bit more streamlined.

PR #8073, opened 2025-01-12
[esp32_rmt] Disable CONFIG_RMT_ISR_IRAM_SAFE when PSRAM is used

swoboda1337:

Add RMT IRAM flag to led_strip when use_psram is false as well remote_transmitter, will hopefully resolve some rare glitching. 

When use_psram is true RMT interrupts must be deferred as psram cannot be accessed in interrupt context. Add an if statement in esp32 to handle multiple components setting the same config. If one component disables it the value should be false regardless of the order components are processed in. 

Change use_psram default to false since it has a fairly large negative effect on RMT performance in every RMT component not just the led_strip code. If no psram component is configured and use_psram is true (the current default) memory will be allocated from DRAM however interrupts will still be deferred (not a good default). 
 
## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4579

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


swoboda1337:
Note, with 5.3.1 there is a callback encoder that could be used to eliminate the large rmt buffer.

PR #8077, opened 2025-01-13
add strong pullup functionality

### LINKED ISSUE #6207: DS18B20 freaking out / continuously reporting 85°C
### The problem

More parasitic powered sensors on bus, 1k pull up. Seems like every time after reset it publishes 85. And sometimes repeated bus reset, so one of the sensors has 85˚C all the time.
In older versions worked fine (don't know which but some from 2023).

2024.8.0

pip

_No response_

ESP8266

### Board

_No response_

_No response_

### Example YAML snippet

```yaml
one_wire:
 - platform: gpio
   pin: 27

sensor:
 - platform: dallas_temp
   address: 0x724b847d0a646128
   name: rain water temperature
   icon: mdi:thermometer-water
   update_interval: 1s
   filters:
         window_size: 256
         send_every: 16

 - platform: dallas_temp
   address: 0xe36168730a646128
   name: main pipe temperature
   update_interval: 1s
   filters:
         window_size: 256
         send_every: 16

 - platform: dallas_temp
   address: 0x2a4361700a646128
   name: tank air temperature
   id: tank_air_temperature
   update_interval: 1s
   filters:
         window_size: 256
         send_every: 16
```

_No response_

### Additional information

Had quick look to source code and seems to me there is no first reading throw nor scratchpad 0x0c test.
Also upon datasheet for parasitic power there should be “strong pull up” during measurement what could be done by GPIO set OUTPUT an HIGH.

prchal:
Adds strong pullup on one wire bus required if parasite powered devices. As explained in https://www.analog.com/media/en/technical-documentation/data-sheets/DS18B20.pdf, page 7, fig. 6.
In function "strung_pullup" it changes gpio to output and sets to 1. Turning back is done in original function "reset".
This is the 1st part - bus part usable for all devices., dalas_temp part will follow.
## Types of changes

- [ ] Other

- fixes [#6207](https://github.com/esphome/issues/issues/6207)

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


prchal:
v. 2:
- fixed tabs to spaces
- added dalas_temp part

DjordjeMandic:
While the current implementation works by toggling the same pin between input and output for strong pullup, it presents several risks. High inrush current through the GPIO during mode switching can lead to voltage drops, potential GPIO damage, or even system instability. A more critical concern is the possibility of the bus being externally pulled low at the wrong moment, which could result in high-side switch failure.

To mitigate these risks, I strongly recommend using a dedicated GPIO pin to control an external switch (such as a FET) for powering the one-wire bus, as outlined in the datasheet example. This approach isolates transient currents from the microcontroller, offering improved reliability and protection.

Furthermore, I suggest implementing a safeguard mechanism in `gpio_one_wire.h`/`gpio_one_wire.cpp` to ensure that external power to the bus is always turned off before driving the bus on the data pin. This would prevent potential conflicts and protect the MCU from unintended current paths.

Lastly, it’s advisable to update the documentation with a recommendation to include a resistor in series with the data pin. The resistor should be sized to limit the current to a safe level for the GPIO, offering additional protection in case of external switch failure and reducing the risk of excessive current draw through the GPIO.

image

image

Your approach is also valid but not fool proof.
image

prchal:
>While the current implementation works by toggling the same pin between input and output for strong pullup, it presents several risks. High inrush current through the GPIO during mode switching can lead to voltage drops, potential GPIO damage, or even system instability. A more critical concern is the possibility of the bus being externally pulled low at the wrong moment, which could result in high-side switch failure.

I think your worries are superfluous since the strong pullup is switched on only after "take measurement" command which is issued on properly recognized temperature sensor. I mean: one wire bus works fine, there is DS18B20, so the risk of short circuit is very low, same as with any other GPIO. 
> To mitigate these risks, I strongly recommend using a dedicated GPIO pin to control an external switch (such as a FET) for powering the one-wire bus, as outlined in the datasheet example. This approach isolates transient currents from the microcontroller, offering improved reliability and protection.

Yes, but I've got a lot of hw without this circuit. And yes, this will isolate current from mcu, but not from power supply what can result in unintended reset and damage of external FET as well. Moreover, high side switch is allways a problem, needs P-channel FET or more complex circuit.
> Furthermore, I suggest implementing a safeguard mechanism in `gpio_one_wire.h`/`gpio_one_wire.cpp` to ensure that external power to the bus is always turned off before driving the bus on the data pin. This would prevent potential conflicts and protect the MCU from unintended current paths.

This is implemented by bus reset which is in every start of communication.
> 
> Lastly, it’s advisable to update the documentation with a recommendation to include a resistor in series with the data pin. The resistor should be sized to limit the current to a safe level for the GPIO, offering additional protection in case of external switch failure and reducing the risk of excessive current draw through the GPIO.

This is good idea, 82 Ohm (or 100 Ohm). BTW as EMC protection for longer buses.

DjordjeMandic:
> Yes, but I've got a lot of hw without this circuit. And yes, this will isolate current from mcu, but not from power supply what can result in unintended reset and damage of external FET as well. Moreover, high side switch is allways a problem, needs P-channel FET or more complex circuit.

Let's say an option to allow another gpio pin for driving power was added:

> This is implemented by bus reset which is in every start of communication.

If pin for driving power was set to be the same one as data pin, everything is implemented already. If it's different then it needs a gpio guard.

But okay for now your implementation is practical. Let's leave adding a different gpio for power control to be dealt with in future.

DjordjeMandic (approved):

ssieb:
The current one wire component already does this.

PR #8083, opened 2025-01-14
Allow overriding GPIO pin validations for specific ESP32 boards

### LINKED ISSUE #3538: Allow GPIO8 for ESP32-Pico-V3-02
### The problem

The  ESP32 QT PY PICO has a neopixel power on GPIO8?  Can we allow GPIO8 on ESP32-Pico-V3-02?  I recognize that ESP32 cannot use it, but "ESP32-Pico-V3-02" can.

See also https://github.com/esphome/issues/issues/3417

ESPHome 2022.8.0 - 17th August 2022

Home Assistant Add-on

Home Assistant 2022.8.7

ESP32

### Board

adafruit_qtpy_esp32, ESP32-Pico-V3-02

platform: gpio

### Example YAML snippet

```yaml
#esp32:
#  board: adafruit_qtpy_esp32
#  variant: esp32
#  framework:
#    type: arduino

esp32:
  board: esp32dev
  framework:
    type: arduino

output:
  - platform: gpio
    id: "neopixel_power"
    pin: GPIO8 # NEOPIXEL_POWER
    inverted: true
```

_No response_

### Additional information

_No response_

edave:

Adds a scoped override to the conf's GPIO pins validation which normally prevents certain pins from being used. This override is scoped to three boards which use GPIO pins 7 & 8 for TX/RX (which are normally reserved for flashing), but is based on implementing a dict mapping board names to pins to be exempted from the reserved validation to allow future boards to be easily added.

Users of ESPHome do not need to make any changes to take advantage of this feature, as it is activated when they specify one of (currently) three known boards in the conf file. For all other boards, the GPIO pin validation is still used.

## Types of changes

- [ ] Other

- fixes esphome/issues#3538 and esphome/issues#6022 

- 

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esp32:
  board: adafruit_feather_esp32_v2
  framework:
    type: esp-idf

uart:
  id: HP_UART
  baud_rate: 2400
  tx_pin: GPIO8
  rx_pin: GPIO7
```

## Checklist:


jesserockz:
I don't think we want to add a list of specific boards like this. We already have `ignore_pin_validation_error` which allows use of the pin in these cases.


```yaml
esphome:
  name: feather

esp32:
  board: adafruit_feather_esp32_v2

binary_sensor:
  - platform: gpio
    id: some_pin
    pin:
      number: GPIO7
      ignore_pin_validation_error: true
```

edave:
Thanks for the feedback! I think I'm a bit confused about about a few things which led to this PR, if you could answer them I think it'd help the community better understand how to avoid this in the future.

1) Where is the documentation for `ignore_pin_validation_error`? I cannot find it if I try searching the docs (and it's immediately apparent in the relevant method that this is possible), and it seems others were not aware there was a way to ignore the validation error.
2) Your [comment](https://github.com/esphome/esphome/pull/5615#issuecomment-1784199603) for the `ignore_pin_validation_error` PR suggests that the best solution in this case is to update `esp32/boards.py` rather than use `ignore_pin_validation_error` as that will cause the validation to be ignored. 

However, `esp32/boards.py` already contains the board `adafruit_feather_esp32_v2` with TX/RX correctly mapped to pin 7/8 that are causing the validation error. If I understand your comment correctly, though, this mapping in `esp32/boards.py` should cause the validation error to be ignored in the first place without any additional configuration. To me this suggests that the logic mentioned in the comment may not be working as intended and should be fixed? I'm happy to investigate further and submit a fix if so.

Thanks for helping me understand this!

robotechnology1977:
> ```yaml
> ignore_pin_validation_error
> ```
Looking in the esphome project, I see that **ignore_pin_validation_warning** is only an attribute for the ESP32/**GPIO component**. If an application/board uses pins like the ones in the related issues, but for other components, the ignore approach will not help.

In my case, for one project, I use the UART component and, if I insert the ignore attribute....
```yaml
uart:
  - id: uart_MBUS
    tx_pin: GPIO8
    rx_pin: GPIO7
    ignore_pin_validation_error: true
    baud_rate: 2400
    rx_buffer_size: 2048
    parity: EVEN
    data_bits: 8
    stop_bits: 1
```

... I get the (expected) error: _[ignore_pin_validation_error] is an invalid option for [uart]._

I recommend considering the fix with the boards' pin-definitions (reserved for validation).
Ideally, the boards should have the exact 1:1 matching with the reality of pins available for usage, without the user having to create "exception lists" ....

jesserockz:
@robotechnology1977 Your yaml is incorrect.

It should be:
```yaml
uart:
  tx_pin: 
    number: GPIOXX
    ignore_pin_validation_error: true
  rx_pin:
    number: GPIOXX
    ignore_pin_validation_error: true
```

robotechnology1977:
@jesserockz - Thank you for pointing out this option for configuring pins as part of an UART component . I tried it in the .yaml file and works :+1: 

I have previously looked at the doc. of the UART Component: https://esphome.io/components/uart.html
... and also at the Pin Configuration: https://esphome.io/guides/configuration-types#config-pin ,

None of them indicate the option "ignore_pin_validation_error", like @edave also pointed out - although can be a very useful one.  I recommend adding this option to _all the components_ where it is supported, it can help others like me, when they use boards not 100% matching with the pin configs of pre-defined boards in esphome. 

Also, updating (or adding other board-IDs) to **esp32/boards.py** for such boards, would be the ideal way to deal with them, in my opinion.

swoboda1337:
I have added the docs for `ignore_pin_validation_error` in https://github.com/esphome/esphome-docs/pull/5563 as it was missing. 

Thanks for working on this PR. However, I'm going to close this in favor of documenting the existing `ignore_pin_validation_error` feature, which already provides the functionality needed for boards like these.

PR #8096, opened 2025-01-16
[sensor] Add confirmation filter

swoboda1337:

This filter stores the last value and only passes incoming values through if the incoming value is sufficiently similar to the previously one, basically an inverse delta filter. 

Useful to filter out corrupted values, e.g. a false CRC pass from a wireless sensor. Corrupted values tend to vary significantly and happen infrequently, easy to filter out in slower moving data or wireless sensors that repeat the same value multiple times. 

I had implemented this in a lambda filter but I think others would find it useful. 

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4586

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:
  - [x The code change is tested and works locally.


jesserockz (changes requested):
I feel like this could just be an `inverse` bool flag on the `delta` filter instead of duplicating it?

PR #8032, opened 2025-01-03
[modbus] Fix timing bugs and better adhere to spec

exciton:
This PR is part of a series of fixes for modbus and related components:
1. #8032 
2. #14172 (merge after 1)
3. #11969 (merge after 2)
4. #11781 (merge after 3)
5. #11987 (merge after 3)
6.  #12376 (merge after 3)
7.  #12421 (merge after 4 & 6)
8. #12612 (merge after 7)
9.  #12614 (merge after 8)

# The issues (see https://github.com/esphome/issues/issues/6633)
- Frame delay timeout (3.5 characters from spec) is not honoured, incorrectly merging unrelated bytes
- Turnaround delay (application specific) is not present, allowing tx too soon after rx or tx
- UART buffer (120 bytes by default) causes timeouts for long responses which get broken into pieces
- TX is synchronous (with flush) causing long component loop execution times
- Waiting for response didn't care if response was from correct address
- Waiting for response is used for server mode which cannot receive responses
- waiting_for_response() only blocks when waiting, doesn't block for frame delay or turnaround delay
- Multiple components inherit from ModbusDevice but don't block for waiting_for_response() -- sdm_meter, havells_solar, kuntze, pzemac, pzemdc, selec_meter

# The fixes:
- Adds a standard 3.5 character timeout for receiving frames (per spec)
- Adds a configurable turnaround delay to prevent sending too quickly after a previous message
- Adds an additional receive timeout delay when long messages are received, to workarond UART buffering
- <strike>Forces RxFIFOFull to 1 for Arduino and uart_set_rx_full_threshold to 1 for ESP-IDF.</strike> This change is taken care of in #8103 with new defaults. 
- Makes TX asynchronous (with set_timeout) to avoid long loops
- Makes on_modbus_data and other callbacks asynchronous to avoid long loops
- Adds a TX buffer queue for ModbusDevices which don't pay attention to waiting_for_response().
- Modifies !waiting_for_response (and renames to ready_for_immediate_send) to ensure send can be immediate (ModbusController will behave the same, TX buffer will never fill up)

To test this PR add the following to your configuration.

```yaml
external_components:
  - source: github://pr#8032
    components: [modbus, modbus_controller, growatt_solar]
    refresh: 1h
```

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[...]

PR #8103, opened 2025-01-17
[uart] Multiple ESP32 features and fixes

brambo123:

This PR adds 3 new configuration options to the uart component for ESP32:
- `flow_control_pin` can be used for ~~hardware~~ driver level RS485 flow control. Reducing the need for blocking software flow control.
- `rx_full_threshold` determines how many bytes are needed te be received before data is available for processing.
- `rx_timeout` determines how long (in byte-lengths) the bus can be idle before data becomes available for processing.

In addition, in the IDF platform version, some code was placed in the setup() function instead of load_settings().
This made it impossible to change some parameters during runtime.

The current fixed `rx_full_threshold` seems to be 120.
Newer Arduino framework versions seems to switch to 1 at low baudrates.
The current `rx_timeout` seems to be 10 for ESP32-IDF and 2 for Arduino, although Arduino seems to have opted for 1 in newer versions.
A `rx_timeout` of 10 byte-lengths adds a minimum of about 10.5ms delay at 9600 bauds before data becomes available. Arduino's 2 seems a more suitable default value.

## Types of changes

- [ ] Other

- https://github.com/esphome/feature-requests/issues/421

- esphome/esphome-docs#5352

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
uart:
  flow_control_pin: GPIO22
  rx_full_threshold: 8
  rx_timeout: 1

```

## Checklist:


brambo123:
For information, the current fixed values for rx_timeout and rx_full_threshold:
[Arduino rx_timeout: 2](https://github.com/espressif/arduino-esp32/blob/2.0.5/cores/esp32/HardwareSerial.cpp#L143)
[Arduino rx_full_threshold: 112](https://github.com/espressif/arduino-esp32/blob/2.0.5/cores/esp32/HardwareSerial.h#L109)
[ESP-IDF rx_timeout: 10](https://github.com/espressif/esp-idf/blob/v5.1.5/components/driver/uart/uart.c#L56)
[ESP-IDF rx_full_threshold: 120](https://github.com/espressif/esp-idf/blob/v5.1.5/components/driver/uart/uart.c#L55)

esphome/components/uart/__init__.py
```diff
@@ -239,8 +243,17 @@ def validate_port(value):
...
+            cv.Optional(CONF_FLOW_CONTROL_PIN): cv.All(
+                cv.only_on_esp32, pins.internal_gpio_output_pin_schema
+            ),
             cv.Optional(CONF_PORT): cv.All(validate_port, cv.only_on(PLATFORM_HOST)),
             cv.Optional(CONF_RX_BUFFER_SIZE, default=256): cv.validate_bytes,
+            cv.SplitDefault(CONF_RX_FULL_THRESHOLD, esp32=120): cv.All(
+                cv.only_on_esp32, cv.validate_bytes, cv.uint8_t
+            ),
+            cv.SplitDefault(CONF_RX_TIMEOUT, esp32=2): cv.All(
+                cv.only_on_esp32, cv.validate_bytes, cv.int_range(min=0, max=92)
```
jesserockz, line 235: Shouldn't this be a time_period validator?

exciton: No - this timeout is set in bytes. The actual time is calculated internally by the framework depending on the baud rate and accounting for start/stop/parity bits.

esphome/components/uart/__init__.py
```diff
@@ -239,8 +243,17 @@ def validate_port(value):
             cv.Required(CONF_BAUD_RATE): cv.int_range(min=1),
             cv.Optional(CONF_TX_PIN): pins.internal_gpio_output_pin_schema,
             cv.Optional(CONF_RX_PIN): validate_rx_pin,
+            cv.Optional(CONF_FLOW_CONTROL_PIN): cv.All(
+                cv.only_on_esp32, pins.internal_gpio_output_pin_schema
+            ),
             cv.Optional(CONF_PORT): cv.All(validate_port, cv.only_on(PLATFORM_HOST)),
             cv.Optional(CONF_RX_BUFFER_SIZE, default=256): cv.validate_bytes,
+            cv.SplitDefault(CONF_RX_FULL_THRESHOLD, esp32=120): cv.All(
```
exciton: Can we emulate the recent library behaviour of setting this to 120 for high bauds and 1 for <= 9600)?

brambo123: I was thinking about what a good default value is.
Something that doesn't cause an unnecessary amount of interrupts, but does provide good throughput.

What if we calculate a default value based on a maximum time?
For example limited to 10ms.
Then you have good throughput at any baudrate.

exciton: That could work, but it would definitely be different from what the frameworks currently use, and may change behaviour for folks unexpectedly. 
It may be better to try to emulate what the frameworks do (as a default), while making it super clear to people what the default is (currently it's hidden deep in framework code). 
e.g. Maybe like this: https://github.com/espressif/arduino-esp32/blob/b63f947037bece7a499faf0043636dd8531b885e/cores/esp32/HardwareSerial.cpp#L414

brambo123: Only newer framework versions adjust the default value.
If I'm correct, the recommended Arduino version is 2.0.5, which has a fixed value.

The frameworks already use different default values.
And depending on the framework version there is also a difference.
For example, newer Arduino versions also use an rx_timeout of 1 instead of 2.

So what is the sensible thing to do?
I think it is wise to choose a default value for ESPHome that does not change across different frameworks and versions.
We must not forget that ESP32 is not the only platform within ESPHome.
Other platforms work differently, if a component relies on the high rx_full_threshold, it will not work on other platforms.
So I think it is safe to lower the default values.
Raising it later is not wise.

nielsnl68:
Just wondering; Are this changes you made also useful for the esp32s3 etc. with UART over USB? I haven't looked into this jet but in the near future i wanted to use this option. To create a dongle for a bridge to the ESPNOW protocol.

brambo123:
> Just wondering; Are this changes you made also useful for the esp32s3 etc. with UART over USB? I haven't looked into this jet but in the near future i wanted to use this option. To create a dongle for a bridge to the ESPNOW protocol.

No, this is only about the 'real' UART.
For usb, USB_CDC is used, which is not (yet) available as a uart port.

esphome/components/uart/__init__.py
clydebarrow: So maybe the option should be named differently?

esphome/components/uart/__init__.py
exciton: Fair enough. Setting it to 120 / 2 seems good.

esphome/components/uart/__init__.py
exciton: The naming matches the framework naming, so that's helpful, and it _is_ a timeout, just set in units of bytes. If the documentation is updated it shouldn't be confusing to people.

clydebarrow: Step out of the developer mindset for a moment; the average user will know nothing about the framework and won't read the docs unless forced to. And it already confused Jesse :-)
IMO the framework function is confusingly named, there is a brief opportunity here to not repeat that mistake.

brambo123:
I came across a small thing.
The get_rx_full_threshold() (and other functions) should of course also indicate a good value on other platforms.
I will adjust it a bit.

And of course the tests need to be updated.
Which I think contains an error anyway.
Someone once thought that ESP32-S3 only has 2 uart ports, should be 3 as far as I know.

esphome/components/uart/__init__.py
brambo123: It is just a timeout, just not with second scales.
Calling it rx_timeout_in_byte_lengths is also a bit weird.
A better name is welcome.

By the way, I expect that a average user only sets this value if a component indicates that a certain value is required  ;)

esphome/components/uart/__init__.py
brambo123: I do think 120 is a bit high for lower baudrates.
Maybe other developers can also let us know what they think is wise.

Keeping the current value is nice.
But at a baudrate of 9600 a byte lasts about 1ms.
So with long data streams you would only get data every 120ms, a bit annoying with timeout detection.
And for general users it's a bit unfriendly to have to set a weird threshold all the time.

That's why I thought a default based on a safe 10ms would be useful.
Or we could go all the way to time in ms.

exciton:
I've already written code to work around any value that's set here for the
modbus component at least, so the default isn't super important. Simpler is
always better in my opinion.

brambo123:
I was thinking about this and then I thought of something.
Both rx_full_threshold and rx_timeout are purely for internal timing.
Shouldn't we just give components the freedom to adjust (and read) it?
And so users don't have to adjust it.

exciton:
Probably should let users set it, given the number of configurations that
are possible. Components can always ensure it's set correctly.

esphome/components/uart/uart_component.cpp
```diff
@@ -20,5 +20,19 @@ bool UARTComponent::check_read_timeout_(size_t len) {
...
 }
 
+void UARTComponent::set_rx_full_threshold_ms(uint8_t time) {
+  uint8_t bytelength = this->data_bits_ + this->stop_bits_ + 1;
+  if (this->parity_ != UARTParityOptions::UART_CONFIG_PARITY_NONE)
+    bytelength += 1;
+  uint8_t val = (this->baud_rate_ / (bytelength * 1000 / time)) - 1;
+  if (val <= 1) {
+    val = 1;
+    this->set_rx_timeout(0);
```
exciton: This disables the timeout interrupt, which may still be desired even if the full_threshold is 1. Better to let folks set this manually, since they may want to use the interrupt

exciton (approved):

esphome/components/uart/__init__.py
DjordjeMandic: `rx_byteout` (:
`rx_timeout_bytes`

exciton:
This looks good to me. Once merged I will update #8032 to utilize this. @jesserockz

esphome/components/uart/__init__.py
jesserockz: Indeed, Its confusing to call something a **time**out when it has nothing to do with time.

It is not exactly bytes too, so I don't have a good alternative name

```c
 * @param tout_thresh  This parameter defines timeout threshold in uart symbol periods. The maximum value of threshold is 126.
 *        tout_thresh = 1, defines TOUT interrupt timeout equal to transmission time of one symbol (~11 bit) on current baudrate.
 *        If the time is expired the UART_RXFIFO_TOUT_INT interrupt is triggered. If tout_thresh == 0,
 *        the TOUT feature is disabled.
```

patagonaa:
Please merge this and ideally set a good default for the rx_full_threshold (lower than common timeouts).

I just spent hours of my life staring at my logic analyzer and the modbus component code, wondering why it would send requests before the last response is received.

Turns out, of course, the receive timeout elapses because no bytes "arrive" (which would prevent the timeout from elapsing), allowing the next request to be sent.

esphome/components/uart/__init__.py
kbx81: Confusing naming -- I also agree.
If I'm understanding this correctly, the timeout is measured in intervals of something ("uart symbol periods").
Maybe `CONF_RX_TIMEOUT_INTERVALS` or `CONF_RX_TIMEOUT_INTERVAL_COUNT` would be better?

esphome/components/uart/uart_component.cpp
```diff
@@ -20,5 +20,18 @@ bool UARTComponent::check_read_timeout_(size_t len) {
...
+void UARTComponent::set_rx_full_threshold_ms(uint8_t time) {
+  uint8_t bytelength = this->data_bits_ + this->stop_bits_ + 1;
+  if (this->parity_ != UARTParityOptions::UART_CONFIG_PARITY_NONE)
+    bytelength += 1;
+  int32_t val = (this->baud_rate_ / (bytelength * 1000 / time)) - 1;
+  if (val < 1) {
+    val = 1;
+  } else if (val > 120) {
+    val = 120;
+  }
```
kbx81: `#include "esphome/core/helpers.h"` and use `clamp` here:
```suggestion
  int32_t val = clamp<int32_t>((this->baud_rate_ / (bytelength * 1000 / time)) - 1, 1, 120);
```

exciton:
This looks ready to merge to me. When this is merged I can update #8032 to use this feature directly.

esphome/components/uart/uart_component_esp32_arduino.cpp
```diff
@@ -161,6 +171,16 @@ void ESP32ArduinoUARTComponent::dump_config() {
   this->check_logger_conflict();
 }
 
+void ESP32ArduinoUARTComponent::set_rx_full_threshold(size_t rx_full_threshold) {
+  this->rx_full_threshold_ = rx_full_threshold;
+  this->hw_serial_->setRxTimeout(this->rx_full_threshold_);
```
exciton: this should be setRxFIFOFull

ngehrsitz (approved):

esphome/components/uart/__init__.py
patagonaa: Both approaches (setting the default to a fixed time based on the baud rate or setting it to 120 by default and to 2 when baud rate is low) sound okay, I think I personally would prefer a fixed time, so time-based timeouts (like there are in the modbus component) always work the same no matter the baudrate.

Either way, even if left in the current state, this should definitely be merged, as without this, modbus doesn't work properly (and can't be configured to) when the response is long and the baudrate is low.

patagonaa (approved):
Code looks great and I've been running this for months now (to fix issues with long modbus responses) without any problems.

esphome/components/uart/uart_component_esp32_arduino.cpp
```diff
@@ -151,8 +157,11 @@ void ESP32ArduinoUARTComponent::dump_config() {
   ESP_LOGCONFIG(TAG, "UART Bus %d:", this->number_);
   LOG_PIN("  TX Pin: ", tx_pin_);
   LOG_PIN("  RX Pin: ", rx_pin_);
+  LOG_PIN("  Flow Control Pin: ", flow_control_pin_);
   if (this->rx_pin_ != nullptr) {
     ESP_LOGCONFIG(TAG, "  RX Buffer Size: %u", this->rx_buffer_size_);
+    ESP_LOGCONFIG(TAG, "  RX Full Threshold: %u", this->rx_full_threshold_);
+    ESP_LOGCONFIG(TAG, "  RX Timeout: %u", this->rx_timeout_);
```
kbx81: `ESP_LOGCONFIG` lines need to be consolidated into a single call. See our [logging best practices](https://developers.esphome.io/architecture/logging/#the-solution-combine-related-log-messages) (or look at any other components; they've all been updated accordingly).

kbx81 (dismissed):

patagonaa:
So, as far as I can tell, the open points are:
- default value for `rx_full_threshold` (I would propose 1 for baudrate <= 9600, 120 for everything else, as Arduino uses and should be good for most cases)
- naming for `rx_timeout` (maybe `rx_timeout_bytes`? Either way, a short description in the docs should clear up confusion that comes up)

@brambo123 are you planning to keep working on this PR as there hasn't been communication from you in a while?

To the ESPHome maintainers:
If brambo123 doesn't want to keep working on this, is there a way for me to complete this PR? I have a bunch of free time right now and am somewhat familiar with the ESPHome codebase.  
Also, what do you think about my proposals above?

exciton:
> So, as far as I can tell, the open points are:
> - default value for `rx_full_threshold` (I would propose 1 for baudrate <= 9600, 120 for everything else, as Arduino uses and should be good for most cases)

I support this. Will make for good defaults for modbus.

jesserockz:
@patagonaa Given the way GitHub works with PRs and branches/permissions etc, the best way for you to continue this PR would be to work on the branch in your own fork, then comment back here again and a maintainer is able to pull your commits into this branch. 

Either that, or @brambo123 needs to give you write access to their fork so you can push directly into the branch/PR.

brambo123:
> > So, as far as I can tell, the open points are:
> > 
> > * default value for `rx_full_threshold` (I would propose 1 for baudrate <= 9600, 120 for everything else, as Arduino uses and should be good for most cases)
> 
> I support this. Will make for good defaults for modbus.

If we're going to use a different default value anyway...
Wouldn't it be better to go with a more accurately calculated value, like 10ms?
That way, you'll always have a reasonable threshold for any baudrate.
We can use the new set_rx_full_threshold_ms function for that.

brambo123:
> * naming for `rx_timeout` (maybe `rx_timeout_bytes`? Either way, a short description in the docs should clear up confusion that comes up)

No one has a good alternative.
Adding _bytes doesn't automatically make it correct.
And then, in all other cases where time is involved, you'd also have to add things like _s or _ms.

> @brambo123 are you planning to keep working on this PR as there hasn't been communication from you in a while?
> 

Yes, I was just waiting for any reply

bigretromike:
@clydebarrow @jesserockz any chance to review this

patagonaa:
Regarding `rx_full_threshold`:
> If we're going to use a different default value anyway...
> Wouldn't it be better to go with a more accurately calculated value, like 10ms?
> That way, you'll always have a reasonable threshold for any baudrate.

Fair. Thinking about it again, going with the defaults Arduino uses doesn't have a lot of advantages compared to just going with a fixed time like 10ms (while still setting the value in bytes/symbols if one wants to override it).

---

Regarding `rx_timeout`:
> No one has a good alternative.
> Adding _bytes doesn't automatically make it correct.

Maybe changing this timeout to actual time instead would be a good solution then (rounding up to 1 byte if it is below that), as I don't think it is ever really necessary to set a byte-exact timeout.

---

> Yes, I was just waiting for any reply

Thank you, sadly this seems to be one of these cases where there isn't a "right" or "wrong" solution and someone just has to make a decision.

---

To anyone who has the power to merge this:  
Are there any solutions you wouldn't be fine with merging?
Nitpicks aside, getting this PR merged is important as right now, somewhat short timeouts just don't work on slow baudrates, breaking actual ESPHome use cases (like modbus).

exciton:
The timeout setting can be done either way (by time or bytes) - and either will work. Just note that some protocols (e.g. modbus) specify this in bytes, and the underlying frameworks also receive it in bytes too, so setting it in time in the UART component will require two calculations for those protocols (e.g. bytes to time in the modbus component, then time back to bytes in the UART component). Very simple calculations of course, but adds a bit of complexity. 

Ultimately can make it work either way.

patagonaa:
> setting it in time in the UART component will require two calculations for those protocols

Is it planned to validate / set the UART settings (timeout and full threshold) from the modbus component to make sure it is set correctly? 

If so, it's probably best to keep timeout in bytes IMO.

brambo123:
> Regarding `rx_timeout`:
> 
> > No one has a good alternative.
> > Adding _bytes doesn't automatically make it correct.
> 
> Maybe changing this timeout to actual time instead would be a good solution then (rounding up to 1 byte if it is below that), as I don't think it is ever really necessary to set a byte-exact timeout.

`rx_timeout` shouldn't be in time.
Changing this to time actually causes problems because the value becomes too high.
The default value of 2 is almost always fine.
Only in very specific cases may adjustment be desirable.

It is only the `rx_full_threshold` that is currently giving problems, and only for this value could you actually use time.
I feel like it might be better if the modbus component itself called set_rx_full_threshold_ms().
Creating a default value based on baud rate or time is a bit tricky to do in a neat way, mainly because the default is handled in the Python config verification.

exciton:
> > Regarding `rx_timeout`:
> > 
> > > No one has a good alternative.
> > > Adding _bytes doesn't automatically make it correct.
> > 
> > Maybe changing this timeout to actual time instead would be a good solution then (rounding up to 1 byte if it is below that), as I don't think it is ever really necessary to set a byte-exact timeout.
> 
> `rx_timeout` shouldn't be in time.
> Changing this to time actually causes problems because the value becomes too high.
> The default value of 2 is almost always fine.
> Only in very specific cases may adjustment be desirable.

Agreed.

> 
> It is only the `rx_full_threshold` that is currently giving problems, and only for this value could you actually use time.
> I feel like it might be better if the modbus component itself called set_rx_full_threshold_ms().
> Creating a default value based on baud rate or time is a bit tricky to do in a neat way, mainly because the default is handled in the Python config verification.
> 

The way I handled this in #8032 is to set `rx_full_threshold` in the UART configuration, and read it in the modbus component to account for it. There isn't a "correct" value for this for modbus, it's an implementation detail of the UART buffer that the modbus component needs to work around. 

The configuration belongs in the UART component IMO, since it affects everything that uses UART, not just modbus.

The defaults I proposed in #8032 should work fine - they basically match what the frameworks use for defaults anyway.

brambo123:
I've now made it so that the default value for `rx_full_threshold` is calculated to 10ms.
This should give the following default values:
| Baudrate | rx_full_threshold |
| -------- | ------- |
| 2400 | 1 |
| 9600 | 8 |
| 19200 | 18 |
| 38400 | 37 |
| 57600 | 56 |
| 115200 | 114 |

This provides data to the components fairly quickly without unnecessary hardware interrupts.

esphome/components/uart/uart_component_esp32_arduino.cpp
```diff
@@ -162,6 +173,16 @@ void ESP32ArduinoUARTComponent::dump_config() {
   this->check_logger_conflict();
 }
 
+void ESP32ArduinoUARTComponent::set_rx_full_threshold(size_t rx_full_threshold) {
+  this->rx_full_threshold_ = rx_full_threshold;
+  this->hw_serial_->setRxFIFOFull(this->rx_full_threshold_);
```
exciton: This line should be deleted, or changed to guard against null hw_serial_. As-is this causes a default configuration to crash at start-up because it's called before hw_serial_ is initialised.

esphome/components/uart/uart_component_esp32_arduino.cpp
```diff
@@ -162,6 +173,16 @@ void ESP32ArduinoUARTComponent::dump_config() {
...
+}
+
+void ESP32ArduinoUARTComponent::set_rx_timeout(size_t rx_timeout) {
+  this->rx_timeout_ = rx_timeout;
+  this->hw_serial_->setRxTimeout(this->rx_timeout_);
```
exciton: Likewise this needs to be deleted or guarded against a null hw_serial_.

exciton (changes requested):
Couple of changes to make this actually work with arduino framework. As is, it crashes on boot.

asund:
Thank you all for the effort. This is going to make communications with my 1200 baud device much quicker.

brambo123:
> Couple of changes to make this actually work with arduino framework. As is, it crashes on boot.

Good find.
I think it worked in the past, but a newer Arduino framework has been adopted, so perhaps something has changed.
The ESP-IDF firmware already had a check for this, and the same check is now included in the Arduino Framework.

exciton (approved):
Tested with latest fix for arduino platform, and works well. Default rx_full_threshold of 10ms makes sense for esphome - this should ensure bytes are always available between component loops if any have been received. 
Testing was done along with #8032 (skipping the uart changes currently in that PR), and everything works together very nicely!

exciton:
@kbx81 - this PR is good to go, and implements all the UART features needed for #8032.

esphome/components/uart/__init__.py
kbx81: Are we agreeing to stick with `CONF_RX_TIMEOUT` for the name here?

kbx81: Left some suggestions on the docs; in particular, [this suggestion](https://github.com/esphome/esphome-docs/pull/5352#discussion_r2356202639) is related to this discussion and I wanted to call it out for feedback. 🙂

kbx81:
Looks like a couple of merge conflicts have crept in 🙈

brambo123:
> Looks like a couple of merge conflicts have crept in 🙈

ESP32 Arduino uart is gone.
Hopefully I pressed the right buttons to fix it 😐

kbx81 (approved):
Thanks for working on this! 🍻

kbx81:
Yes, some changes were merged which rework how we are dealing with Arduino configs. Is there anything we should do further testing on before merging?

brambo123:
> Is there anything we should do further testing on before merging?

Just to be sure, I did another test with the Arduino platform, works fine.
So I think it's ready to merge.

bdraco:
clang-tidy failure looks transient. We can re-run once its done

kbx81:
Installed on a few devices I have here which use the UART component, just for an additional smoke check...seems to be working fine. Thanks for all your work! 🍻

DavidAntonin:
With this PR and ESPHome 9.1 I am getting compile error:
src/esphome/components/uart/uart_component_esp_idf.cpp: In member function 'virtual void esphome::uart::IDFUARTComponent::setup()':
src/esphome/components/uart/uart_component_esp_idf.cpp:78:30: error: 'class esphome::logger::Logger' has no member named 'get_uart_num'; did you mean 'get_uart'?
   78 |       logger::global_logger->get_uart_num() == next_uart_num) {
      |                              ^~~~~~~~~~~~
      |                              get_uart
*** [.pioenvs/esp32-s2-bazen/src/esphome/components/uart/uart_component_esp_idf.cpp.o] Error 1

brambo123:
> With this PR and ESPHome 9.1 I am getting compile error: src/esphome/components/uart/uart_component_esp_idf.cpp: In member function 'virtual void esphome::uart::IDFUARTComponent::setup()': src/esphome/components/uart/uart_component_esp_idf.cpp:78:30: error: 'class esphome::logger::Logger' has no member named 'get_uart_num'; did you mean 'get_uart'? 78 | logger::global_logger->get_uart_num() == next_uart_num) { | ^~~~~~~~~~~~ | get_uart *** [.pioenvs/esp32-s2-bazen/src/esphome/components/uart/uart_component_esp_idf.cpp.o] Error 1

I think this is due to other changes; I had the same issue.
If you use the full esphome dev version, it works.
`pip3 install https://github.com/esphome/esphome/archive/dev.zip`

DavidAntonin:
I temporarily solved it by:
```
external_components:
  - source: github://brambo123/esphome@uart_flow_control
    components: uart
```

ccutrer:
Heh, you merged _just_ before I could comment. My only concern is naming the config option "flow_control_pin", since "flow control" with a UART typically refers to hardware flow control, with RTS and CTS signals, which do _not_ map exactly onto the "driver enable" semantics used by MAX485. Though if in the future RTS/CTS become supported, it would be reasonable to add an additional option "flow_control: hardware" to mean RTS/CTS style, and not MAX485 style.

brambo123:
> Heh, you merged _just_ before I could comment. My only concern is naming the config option "flow_control_pin", since "flow control" with a UART typically refers to hardware flow control, with RTS and CTS signals, which do _not_ map exactly onto the "driver enable" semantics used by MAX485. Though if in the future RTS/CTS become supported, it would be reasonable to add an additional option "flow_control: hardware" to mean RTS/CTS style, and not MAX485 style.

It was already called `flow_control_pin` in the modbus component.
We have to use consistent naming.

PR #8107, opened 2025-01-17
Move CI from Python 3.9 to Python 3.11

stellar-aria:

Brings internal Python references (namely CI and dev requirements) up to parity with the current Docker image version.

Note: depends on #8108

### DIFF
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5.3.0
         with:
-          python-version: "3.9"
+          python-version: "3.11"
       - name: Set up Docker Buildx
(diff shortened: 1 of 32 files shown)


stellar-aria:
Ugh. That's annoying. I'm not sure the best way to go about this now. The CI will need to be moved from Black to Ruff, to match the general migration, that might be easier to do before any version upgrades...

jesserockz:
We are now using python 3.11 as the minimum so this PR can be cleaned up a bit to only include specific component changes.

bdraco:
Thanks for putting this work together! We appreciate the effort you put into modernizing the codebase to Python 3.11.

It looks like we ended up getting to Python 3.11 through other means as well, including #8850, so we won't be moving forward with merging this PR.

Thanks again for your contributions to the project!

PR #8116, opened 2025-01-21
[online_image] Add binary bmp support

jesserockz:

This adds basic support for binary BMP images to be downloaded and displayed. This PR will **not** add more functionality. If anyone wants colour support, it can be done in a further PR after this is merged.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4600

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+static const char *const TAG = "online_image.bmp";
+
+int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
+  size_t index = 0;
+  this->decoded_bytes_ += size;
+  if (this->current_index_ == 0 && this->decoded_bytes_ > 14) {
+    // Check if the file is a BMP image
+    if (buffer[0] != 'B' || buffer[1] != 'M') {
+      ESP_LOGE(TAG, "Not a BMP file");
+      return 0;
```
guillempages: Errors in online_image decoders are normally signaled with negative values
```suggestion
      return -1;
```

esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace online_image {
+
+static const char *const TAG = "online_image.bmp";
+
+int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
+  size_t index = 0;
+  this->decoded_bytes_ += size;
```
guillempages: It somehow feels wrong to increase decoded_bytes before actually decoding them 🤔

esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+  if (this->current_index_ == 0 && this->decoded_bytes_ > 14) {
+    // Check if the file is a BMP image
+    if (buffer[0] != 'B' || buffer[1] != 'M') {
+      ESP_LOGE(TAG, "Not a BMP file");
+      return 0;
+    }
+
+    this->total_size_ = encode_uint32(buffer[5], buffer[4], buffer[3], buffer[2]);
```
guillempages: Might be nice to add some comments about what the magic numbers represent / how the header in the file is formatted

esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+    this->data_offset_ = encode_uint32(buffer[13], buffer[12], buffer[11], buffer[10]);
+
+    this->current_index_ = 14;
+    index = 14;
+  }
+  if (this->current_index_ == 14 && this->decoded_bytes_ > data_offset_) {
```
guillempages:
```suggestion
  if (this->current_index_ == 14 && this->decoded_bytes_ > this->data_offset_) {
```

esphome/components/online_image/online_image.cpp
```diff
@@ -120,9 +124,14 @@ void OnlineImage::update() {
 
 #ifdef USE_ONLINE_IMAGE_PNG_SUPPORT
   if (this->format_ == ImageFormat::PNG) {
-    this->decoder_ = esphome::make_unique<PngDecoder>(this);
-  }
+    this->decoder_ = make_unique<PngDecoder>(this);
+  } else
```
guillempages:
```suggestion
  }
```
this->format can only have one value, so it should behave exactly the same without an "else" chain.

With the #ifdef combination it might cause issues with the else not having a block, depending on which combination of defines is actually compiled in.

This is also only called once per image download, so the extra few cycles that the additional if's would cost should be negligible.

esphome/components/online_image/online_image.cpp
```diff
@@ -120,9 +124,14 @@ void OnlineImage::update() {
...
 #endif  // ONLINE_IMAGE_PNG_SUPPORT
+#ifdef USE_ONLINE_IMAGE_BMP_SUPPORT
+      if (this->format_ == ImageFormat::BMP) {
```
guillempages:
```suggestion
  if (this->format_ == ImageFormat::BMP) {
```
If the else is removed, no need to indent this further

tests/components/online_image/common.yaml
```diff
@@ -26,6 +26,10 @@ online_image:
     format: PNG
     type: RGB
     transparency: chroma_key
+  - id: online_binary_bmp
+    url: http://www.libpng.org/pub/png/img_png/pnglogo-blk-tiny.png
```
guillempages: It's not really executed, so it's not that important, but it would feel more consistent, if the URL would point to a bmp image ;-)

esphome/components/online_image/bmp_image.h
```diff
@@ -0,0 +1,41 @@
...
+   *
+   * @param display The image to decode the stream into.
+   */
+  BmpDecoder(OnlineImage *image) : ImageDecoder(image) {}
+
+  int HOT decode(uint8_t *buffer, size_t size) override;
+
+ protected:
+  size_t current_index_{0};
+  ssize_t width_{0};
```
guillempages, line 27: what's the difference between `size_t` (above) and `ssize_t` (here)?

esphome/components/online_image/bmp_image.h
```diff
@@ -0,0 +1,41 @@
...
+  ssize_t height_{0};
```
guillempages, line 28: Are width and height really necessary here? they are already used (and set) in the destination image; can't those be used instead?

esphome/components/online_image/bmp_image.h
```diff
@@ -0,0 +1,41 @@
...
+  uint16_t bits_per_pixel_{0};
```
guillempages, line 29: Same here; the destination image already should have/get this information. Or is this about the "source" bpp? Is there any conversion happening if the destination image uses a different format?

esphome/components/online_image/bmp_image.h
```diff
@@ -0,0 +1,41 @@
...
+  uint32_t compression_method_{0};
+  uint32_t image_data_size_{0};
+  uint32_t color_table_entries_{0};
+  size_t width_bytes_{0};
+  size_t total_size_{0};
```
guillempages: what's the difference to the "download_size_" of the superclass? Can't that one be used/updated?

esphome/components/online_image/bmp_image.h
```diff
@@ -0,0 +1,41 @@
...
+namespace esphome {
+namespace online_image {
+
+/**
+ * @brief Image decoder specialization for PNG images.
+ */
+class BmpDecoder : public ImageDecoder {
+ public:
+  /**
+   * @brief Construct a new PNG Decoder object.
```
guillempages:
```suggestion
   * @brief Construct a new BMP Decoder object.
```

esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+    this->image_data_size_ = encode_uint32(buffer[37], buffer[36], buffer[35], buffer[34]);
+    this->color_table_entries_ = encode_uint32(buffer[49], buffer[48], buffer[47], buffer[46]);
+
+    switch (this->bits_per_pixel_) {
+      case 1:
+        this->width_bytes_ = (this->width_ % 8 == 0) ? (this->width_ / 8) : (this->width_ / 8 + 1);
+        break;
+      default:
+        ESP_LOGE(TAG, "Unsupported bits per pixel: %d", this->bits_per_pixel_);
+        return 0;
```
guillempages:
```suggestion
        return -2;
```
Maybe we should define some constants for possible errors in the "ImageDecoder" class

esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "esphome/core/helpers.h"
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace online_image {
+
+static const char *const TAG = "online_image.bmp";
+
+int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
+  size_t index = 0;
```
guillempages, line 15: I might have missed that, but it looks like `index` is not really used (besides being set at different places), and only `this->current_index_` is actually being read.

Is this variable really needed?

guillempages (commented):
Adding only partial support for (binary) BMP might cause false expectations if people then start wildly downloading BMPs.
OTOH, it's a start, and additional BMP encodings/formats can be added later.

esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+      case 1:
+        this->width_bytes_ = (this->width_ % 8 == 0) ? (this->width_ / 8) : (this->width_ / 8 + 1);
+        break;
+      default:
+        ESP_LOGE(TAG, "Unsupported bits per pixel: %d", this->bits_per_pixel_);
+        return 0;
+    }
+
+    if (!this->set_size(this->width_, this->height_)) {
+      return 0;
```
guillempages:
```suggestion
      return -3;
```

esphome/components/online_image/bmp_image.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    switch (this->bits_per_pixel_) {
+      case 1:
+        this->width_bytes_ = (this->width_ % 8 == 0) ? (this->width_ / 8) : (this->width_ / 8 + 1);
+        break;
+      default:
+        ESP_LOGE(TAG, "Unsupported bits per pixel: %d", this->bits_per_pixel_);
+        return 0;
+    }
+
```
guillempages, line 74: Maybe also return an error "unsupported format" if the compression_method is not "uncompressed"?

esphome/components/online_image/bmp_image.cpp
jesserockz: It is used to know which byte in the current buffer it is up to.
https://github.com/esphome/esphome/pull/8116/files#diff-6400dadccb239aaadf18aad140f9f795cb6dc755e1b05171203a161b09da779fR56

esphome/components/online_image/bmp_image.cpp
jesserockz: I don't disagree. I can rework this

esphome/components/online_image/bmp_image.cpp
jesserockz: I can add something, but isn't the variable name the explanation/description of what is in the (4) bytes?

esphome/components/online_image/bmp_image.cpp
jesserockz: I like this idea

esphome/components/online_image/bmp_image.h
jesserockz: :see_no_evil:

esphome/components/online_image/bmp_image.h
jesserockz: ssize_t is signed size_t, the BMP format allows for negative numbers, mostly negative height which indicates a top down image vs the standard bottom up image.

esphome/components/online_image/bmp_image.h
jesserockz: This is decoded in the first iteration from the header, but will be used later when more bpp are supported to know how to decode the individual bytes which happens in later loops where a local variable wont exist.

esphome/components/online_image/bmp_image.h
jesserockz: I don't believe they would be any different, We could probably update that. If a server doesnt present a Content-Length, then this variable from the header would server as a good replacement of the size of the entire file.

esphome/components/online_image/online_image.cpp
jesserockz: That was the formatter =)

tests/components/online_image/common.yaml
jesserockz: Haha yes. I just didnt know any and was being lazy at the end.

esphome/components/online_image/bmp_image.h
jesserockz: No conversion in this code, but it would be possible in a future PR to say I want greyscale, but the source is RGB etc.

esphome/components/online_image/bmp_image.h
jesserockz: I have removed `total_size_` and assigned the value read from the header to `download_size_`

esphome/components/online_image/bmp_image.cpp
jesserockz: It is now used to loop the buffer instead of current_index (which is used to get the paint_index)

esphome/components/online_image/bmp_image.h
jesserockz: They are required to know the x and y of the image we are drawing, for example if someone tried to download an image larger than the "container" it will attempt to draw outside the bounds (which is just ignored and logged)

esphome/components/online_image/bmp_image.cpp
guillempages: It's just a reflex. I can't stand "magic numbers", but I agree that defining constants here everywhere for the indexes would be overkill.
But having a nice comment with an explanation about the format and how it works might help later on, when implementing additional compression methods or bpps.

esphome/components/online_image/bmp_image.cpp
```diff
@@ -13,21 +13,42 @@ static const char *const TAG = "online_image.bmp";
...
-  this->decoded_bytes_ += size;
-  if (this->current_index_ == 0 && this->decoded_bytes_ > 14) {
+  if (this->current_index_ == 0 && index == 0 && size > 14) {
+    /**
+     * BMP file format:
+     * 0-1: Signature (BM)
+     * 2-5: File size
+     * 6-9: Reserved
+     * 10-13: Pixel data offset
+     */
```
guillempages, line 25:
```suggestion
     *
     * Integer values are stored in little-endian format.
     */
```
Probably obvious to anyone working with BMP, but IMHO it doesn't hurt to point this out ;-)

esphome/components/online_image/image_decoder.h
```diff
@@ -4,6 +4,12 @@
 namespace esphome {
 namespace online_image {
 
+enum DecodeError : int {
```
guillempages, line 7: Hm. I think the project maintainers might consider this to be worthy of a separate PR. Jesse often mentions that smaller separate PRs are easier to review 😜 

Jokes apart, this looks good. Once this is merged I'll update the PNG decoder to make use of those values as well.

esphome/components/online_image/bmp_image.h
guillempages: Actually this out-of-bounds checking should already be checked in the draw method in the ImageDecoder superclass. It even takes care of resizing, if the container is smaller than the image.
So using "setSize()" and the actual image width and height should be ok.

tests/components/online_image/common.yaml
guillempages:
```suggestion
    url: https://samples-files.com/samples/images/bmp/480-360-sample.bmp
```
As mentioned, feel free to ignore (nobody knows, how long the link will be valid); but it feels better to have an actual BMP link ;-)

guillempages (commented):
Looking good.
Just a final set of suggestions.

esphome/components/online_image/image_decoder.h
jesserockz: :rofl:

guillempages (approved):
LGTM

PR #8125, opened 2025-01-22
[spi] Fix data type in bitbash transfer_()

oliv3r:

## Types of changes

- [ ] Other

### DIFF
--- a/esphome/components/spi/spi.cpp
+++ b/esphome/components/spi/spi.cpp
@@ -88,7 +88,7 @@ void SPIDelegateBitBash::write(uint16_t data, size_t num_bits) { this->transfer_
 uint16_t SPIDelegateBitBash::transfer_(uint16_t data, size_t num_bits) {
   // Clock starts out at idle level
   this->clk_pin_->digital_write(clock_polarity_);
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -43,7 +43,8 @@ RUN \
     && rm -rf \
         /tmp/* \
         /var/{cache,log}/* \
-        /var/lib/apt/lists/*
+        /var/lib/apt/lists/* \
+    && usermod -U root -a -G staff
```
clydebarrow: Why is this here? If it's needed, should be in a separate PR with some justification.

clydebarrow (changes requested):

docker/Dockerfile
oliv3r: sorry, yes it has its own PR, and it accidentally stuck :) https://github.com/esphome/esphome/pull/7098 was that PR, but i'll obviously remove that patch from this PR.

Thanks @clydebarrow for noticing

clydebarrow (approved):

jesserockz (approved):

PR #8127, opened 2025-01-22
[online_image] Add JPEG support to online_image

guillempages:

Add JPEG support to online_image, in addition to the already existing PNG support

## Types of changes

- [ ] Other

- esphome/esphome-docs#4617

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
http_request:
  verify_ssl: false

online_image:
  - url: https://www.w3.org/MarkUp/Test/xhtml-print/20050519/tests/jpeg444.jpg
    id: jpeg_image
    format: JPEG
    type: RGB565

display:
  - platform: xyz
    # <...>
    lambda: |-
      it.image(0, 0, id(jpeg_image));
  
```

## Checklist:


esphome/components/online_image/__init__.py
```diff
@@ -69,11 +78,11 @@ def actions(self):
...
 
-# New formats can be added here.
 IMAGE_FORMATS = {
     x.image_type: x
     for x in (
         BMPFormat(),
+        JPEGFormat(),
         PNGFormat(),
     )
 }
```
clydebarrow, line 88: Allowing `format: jpg` as well will avoid some swearing.

```suggestion
}
IMAGE_FORMATS.update({"JPG": IMAGE_FORMATS["JPEG"]})
```

esphome/components/online_image/online_image.h
```diff
@@ -119,6 +126,12 @@ class OnlineImage : public PollingComponent,
 
   uint8_t *buffer_;
   DownloadBuffer download_buffer_;
+  /**
+   * This is the *initial* size of the download buffer, not the current size.
+   * The download buffer can be resized at runtime; the download_buffer_size_ will
```
clydebarrow:
```suggestion
   * The download buffer can be resized at runtime; the download_buffer_initial_size_ will
```

Maybe the constructor parameter should also be renamed?

clydebarrow (changes requested):

esphome/components/online_image/__init__.py
guillempages: Done. I've also added a test for that.

esphome/components/online_image/online_image.h
guillempages: Thanks for finding this one.

W.r.t. renaming the constructor parameter; I'd rather not. The constructor parameter is actually used to set the size of the download buffer (as the name says); just in case of Jpeg decoder, the image needs to be fully downloaded before decoding (because of the way the decoder works), so the initial size value is still needed later on.

clydebarrow (approved):

PR #8157, opened 2025-01-30
[sensor] Fix bug in percentage based delta filter

swoboda1337:

If a percentage based delta filter is used and the sensor returns all zeros the zeros will not be filtered.

Also remove a useless piece of code, no point calculating the delta on NAN. 

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#5147

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


ssieb:
It's true that the docs say that, but I would rather fix the docs than change the code.  The way the code works makes more sense to me.  It just needs a test for the delta being 0.

swoboda1337:
> It's true that the docs say that, but I would rather fix the docs than change the code. The way the code works makes more sense to me. It just needs a test for the delta being 0.

Here are my thoughts:
- I always thought you should be able to do `delta: 0.0` and have any difference pass through. 
- An exception could be added to check for a delta of 0, this exception would also have to be documented.
- Most people won't notice this change as sensor values are floats and whether its 1.00 or 1.01 it doesn't really matter. 

To me it makes more sense to implement it the way the docs are currently written. Code and docs will be simpler and more natural. Curious what others think?

swoboda1337:
Interested to know what others think. If the consensus is >= that is fine but personally I think > makes more sense.

swoboda1337:
@kbx81 thoughts on this one? Would be good to get this fixed.

swoboda1337:
@bdraco thoughts on this one? Change it to > or keep it >= and add exceptions to handle 0?

People are using it as an external component: https://discord.com/channels/429907082951524364/1396610420038439034

bdraco:
I agree that it now aligns with the documentation, and while sensor values are floating-point numbers so exact equality with the threshold is extremely rare in practice, its technically a breaking change. I'm not sure its worth documenting it as such though as it might be noise.

bdraco:
I'm inclined to document it as a bug fix and not a breaking change since the documentation consistently uses "greater than"

bdraco (dismissed):

bdraco:
Since there are some comments that indicate this PR doesn't have broad alignment, this should get another approval before merging.

ssieb (approved):

esphome/components/sensor/filter.cpp
```diff
@@ -340,13 +340,11 @@ optional<float> DeltaFilter::new_value(float value) {
...
     } else {
-      if (this->percentage_mode_) {
-        this->current_delta_ = fabsf(value * this->delta_);
-      }
       return this->last_value_ = value;
     }
   }
-  if (std::isnan(this->last_value_) || fabsf(value - this->last_value_) >= this->current_delta_) {
+  float diff = fabsf(value - this->last_value_);
+  if (std::isnan(this->last_value_) || (diff > 0.0f && diff >= this->current_delta_)) {
```
jesserockz, line 382: Is this gonna work if last_value_ is nan? Its gonna try calculate using it before checking if its nan now...

swoboda1337: Shouldn't it? Diff would be nan as well but that is fine (wont be used).

ssieb: The diff isn't used after this point.  If the last_value is nan then the delta succeeds.

bdraco (approved):

bdraco (approved):

PR #8158, opened 2025-01-30
[bl0940] extend configuration options of bl0940 device

dan-s-github:

It seems that different versions of the bl0940 chip are used in smart plugs.
The initial implementation used non documented read and write commands to communicate with the device, while working on an ESPHome firmware for the [Rowi2 Smart Plug](https://www.kiwi-warmer.co.nz/shop/product/826130/rowi-2-smart-plug-with-api/) I discovered that this device is using a bl0940 that uses the documented read and write commands.

I also added some additional configuration options which would allow to fine tune the readings from the sensor similiar to what is done for the bl0942.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4744

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
sensor:
  - platform: bl0940
    uart_id: uart_bus
    tuya_mode: false
    voltage:
      name: 'Voltage'
    current:
      name: 'Current'
    power:
      name: 'Power'
    energy:
      name: 'Energy'
    internal_temperature:
      name: 'bl0940 Internal temperature'
      entity_category: diagnostic
    update_interval: 60s

```

## Checklist:


dwmw2:
LGTM.

dan-s-github:
converted back to draft as I am adding runtime calibration options

dwmw2:
Perhaps do those in a new PR? This one looks ready and has been for some time.

I'd be interested in how you're planning to do the runtime calibration, and could look at matching it in bl0942.

dan-s-github:
added external calibration option, will leave as draft until docs are updated

dwmw2:
How does the external calibration work (should I just wait for the docs update)?

Does it make sense to try to share this code between this and bl0942, and other similar devices? Or just cut/paste it...

dan-s-github:
documentation has now been updated. It basically uses a Template Number component to tweak references in %.
+/- 5% should be more than required if initial calibration isn't to far of.
These changes have been implemented to allow creation of [this esphome firmware ](https://github.com/vt-vaio/rowi2)

esphome/components/bl0940/sensor.py
```diff
@@ -69,6 +89,21 @@
...
+            cv.Optional(CONF_RESISTOR_ONE): cv.float_,
+            cv.Optional(CONF_RESISTOR_TWO): cv.float_,
+            cv.Optional(CONF_CURRENT_REFERENCE): cv.float_,
+            cv.Optional(CONF_ENERGY_REFERENCE): cv.float_,
+            cv.Optional(CONF_POWER_REFERENCE): cv.float_,
+            cv.Optional(CONF_VOLTAGE_REFERENCE): cv.float_,
+            cv.Optional(CONF_CURRENT_CALIBRATION): cv.use_id(number.Number),
+            cv.Optional(CONF_VOLTAGE_CALIBRATION): cv.use_id(number.Number),
+            cv.Optional(CONF_POWER_CALIBRATION): cv.use_id(number.Number),
+            cv.Optional(CONF_ENERGY_CALIBRATION): cv.use_id(number.Number),
```
jesserockz: Instead of referencing a number here, there should be a `number` platform for this component to allow controlling of these data points.

Example yaml:
```yaml
sensor:
  - platform: bl0940
    id: some_id
    ...

number:
  - platform: bl0940
    bl0940_id: some_id
    current_calibration:
      name: Current Calibration
    voltage_calibration:
      name: Voltage Calibration
    ...
```

jesserockz (changes requested):

esphome/components/bl0940/sensor.py
dan-s-github: looking Into this and try to find some docs or another component that does similar stuff. Where is the best place to get some guidance for this?

jesserockz: Sorry for not coming back to this, but looks like you found what you needed?

esphome/components/bl0940/number/calibration_number.cpp
```diff
@@ -0,0 +1,43 @@
...
+namespace bl0940 {
+
+static const char *const TAG = "bl0940.number";
+
+void CalibrationNumber::setup() {
+  if (this->f_.has_value())
+    return;
+
+  float value;
+  this->pref_ = global_preferences->make_preference<float>(this->get_object_id_hash());
```
jesserockz: Using preferences should always be a user decision with config, please add `CONF_RESTORE_VALUE` to the schemas and use it to decide if to load/save or not.

esphome/components/bl0940/sensor.py
dan-s-github: indeed, the rain weekend helped. 
adding also a button to reset the calibration settings, then I remove the draft status
```
button:
  - platform: bl0940
    bl0940_id: test_id
    name: Cal Reset
```

esphome/components/bl0940/bl0940.cpp
```diff
@@ -7,36 +7,52 @@ namespace bl0940 {
...
+    {BL0940_REG_I_FAST_RMS_CTRL, 0x1C, 0x18, 0x00, 0x1B}};
 
 void BL0940::loop() {
   DataPacket buffer;
   if (!this->available()) {
     return;
   }
   if (read_array((uint8_t *) &buffer, sizeof(buffer))) {
-    if (validate_checksum(&buffer)) {
+    if (validate_checksum_(&buffer)) {
```
kbx81: Please make sure all calls such as this include the `this->` prefix. The lines above and below this one need it added, as well. 🙂 Please add it as necessary anywhere else it's missing.
```suggestion
    if (this->validate_checksum_(&buffer)) {
```

esphome/components/bl0940/bl0940.cpp
```diff
@@ -87,18 +275,25 @@ float BL0940::update_temp_(sensor::Sensor *sensor, ube16_t temperature) const {
...
+  }
+  this->prev_cf_cnt_ = cf_cnt;
+
+  float v_rms = (uint24_t) data->v_rms / voltage_reference_cal_;
+  float i_rms = (uint24_t) data->i_rms / current_reference_cal_;
+  float watt = (int24_t) data->watt / power_reference_cal_;
+  float total_energy_consumption = cf_cnt / energy_reference_cal_;
 
   float tps1 = update_temp_(internal_temperature_sensor_, data->tps1);
   float tps2 = update_temp_(external_temperature_sensor_, data->tps2);
```
kbx81: More lines missing `this->`. 🙂

esphome/components/bl0940/bl0940.h
```diff
@@ -84,26 +114,74 @@ class BL0940 : public PollingComponent, public uart::UARTDevice {
...
+  float voltage_reference_cal_;
+  bool voltage_reference_set_ = false;
+  // Divide by this to turn into Ampere
+  float current_reference_;
+  float current_reference_cal_;
+  bool current_reference_set_ = false;
+  // Divide by this to turn into kWh
+  float energy_reference_;
+  float energy_reference_cal_;
+  bool energy_reference_set_ = false;
```
kbx81: Why initialize some but not others? 🙂

kbx81 (dismissed):

esphome/components/bl0940/bl0940.cpp
dan-s-github: added missing this-> pointers

esphome/components/bl0940/bl0940.cpp
dan-s-github: also here, hopefully I got them all 🙏

esphome/components/bl0940/bl0940.h
dan-s-github: I don't usually initialize variables with values if the initial values is not actually used. Should they always be initialized in a cpp context?

esphome/components/bl0940/number/calibration_number.cpp
dan-s-github: added CONF_RESTORE_VALUE and also fixed calibration reset

esphome/components/bl0940/__init__.py
```diff
@@ -1 +1,10 @@
...
+import esphome.codegen as cg
+from esphome.components import uart
+
+CODEOWNERS = ["@tobias-", "@dan-s-github"]
+
+DEPENDENCIES = ["uart"]
+
+CONF_BL0940_ID = "bl0940_id"
+bl0940_ns = cg.esphome_ns.namespace("bl0940")
+BL0940 = bl0940_ns.class_("BL0940", cg.PollingComponent, uart.UARTDevice)
```
jesserockz: This one line should stay inside the `sensor.py` file

esphome/components/bl0940/sensor.py
```diff
@@ -69,6 +80,17 @@
                 device_class=DEVICE_CLASS_TEMPERATURE,
                 state_class=STATE_CLASS_MEASUREMENT,
             ),
+            cv.Optional(CONF_TUYA_MODE, default=True): cv.boolean,
+            cv.Optional(CONF_READ_COMMAND): cv.hex_uint8_t,
+            cv.Optional(CONF_WRITE_COMMAND): cv.hex_uint8_t,
+            cv.Optional(CONF_REFERENCE_VOLTAGE): cv.float_,
+            cv.Optional(CONF_RESISTOR_SHUNT): cv.float_,
+            cv.Optional(CONF_RESISTOR_ONE): cv.float_,
+            cv.Optional(CONF_RESISTOR_TWO): cv.float_,
```
jesserockz, line 200: If I am understanding correctly, these four values are not used if `tuya_mode` is enabled as it uses a constant instead of calculating?
If so, there should be a validation blocking the use of them if `tuya_mode` is set to true please.

esphome/components/bl0940/sensor.py
```diff
@@ -69,6 +80,17 @@
...
+            cv.Optional(CONF_CURRENT_REFERENCE): cv.float_,
+            cv.Optional(CONF_ENERGY_REFERENCE): cv.float_,
+            cv.Optional(CONF_POWER_REFERENCE): cv.float_,
+            cv.Optional(CONF_VOLTAGE_REFERENCE): cv.float_,
```
jesserockz, line 204: Similarly, if these values are set, the above four used for calculations should be blocked from being used simultaneously in config

esphome/components/bl0940/sensor.py
```diff
@@ -69,6 +80,17 @@
                 device_class=DEVICE_CLASS_TEMPERATURE,
                 state_class=STATE_CLASS_MEASUREMENT,
             ),
+            cv.Optional(CONF_TUYA_MODE, default=True): cv.boolean,
```
jesserockz: Can we name this something other thatn `tuya`? We would prefer not to refer to `Tuya` in ESPHome where possible.

esphome/components/bl0940/bl0940.cpp
```diff
@@ -122,6 +317,26 @@ void BL0940::received_package_(const DataPacket *data) const {
...
 void BL0940::dump_config() {  // NOLINT(readability-function-cognitive-complexity)
   ESP_LOGCONFIG(TAG, "BL0940:");
+  ESP_LOGCONFIG(TAG, "  TUYA MODE: %s", TRUEFALSE(this->tuya_mode_enabled_));
+  ESP_LOGCONFIG(TAG, "  READ  CMD: 0x%02X", this->read_command_);
+  ESP_LOGCONFIG(TAG, "  WRITE CMD: 0x%02X", this->write_command_);
+  ESP_LOGCONFIG(TAG, "  ------------------");
+  ESP_LOGCONFIG(TAG, "  Vref: %f", this->vref_);
+  ESP_LOGCONFIG(TAG, "  R shunt: %f", this->r_shunt_);
+  ESP_LOGCONFIG(TAG, "  R One: %f", this->r_one_);
+  ESP_LOGCONFIG(TAG, "  R Two: %f", this->r_two_);
```
jesserockz: As above, if `tuya_mode` is enabled, these should not be logged

esphome/components/bl0940/bl0940.cpp
```diff
@@ -46,35 +62,207 @@ void BL0940::loop() {
...
+    }
+  }
+
+  if (!this->write_command_set_) {
+    if (this->tuya_mode_enabled_) {
+      this->write_command_ = BL0940_TUYA_WRITE_COMMAND;
+    } else {
+      this->write_command_ = BL0940_WRITE_COMMAND;
+    }
+  }
```
jesserockz: This should be done in a custom python validator that sets the default based on the config items.

esphome/components/bl0940/bl0940.h
```diff
@@ -84,26 +114,74 @@ class BL0940 : public PollingComponent, public uart::UARTDevice {
   sensor::Sensor *internal_temperature_sensor_{nullptr};
   sensor::Sensor *external_temperature_sensor_{nullptr};
 
+#ifdef USE_NUMBER
+  // calibration helpers
+  number::Number *voltage_calibration_{nullptr};
+  number::Number *current_calibration_{nullptr};
+  number::Number *power_calibration_{nullptr};
+  number::Number *energy_calibration_{nullptr};
```
jesserockz: Best to be explicit, and matches the sensors above and button below.
```suggestion
  number::Number *voltage_calibration_number_{nullptr};
  number::Number *current_calibration_number_{nullptr};
  number::Number *power_calibration_number_{nullptr};
  number::Number *energy_calibration_number_{nullptr};
```

esphome/components/bl0940/bl0940.cpp
```diff
@@ -46,35 +62,207 @@ void BL0940::loop() {
...
+  this->power_reference_cal_ = this->power_reference_ / this->power_cal_;
+
+  if (!this->energy_reference_set_) {
+    if (this->tuya_mode_enabled_) {
+      this->energy_reference_ = BL0940_TUYA_EREF;
+    } else {
+      this->energy_reference_ = this->calculate_energy_reference_();
+    }
+  }
+  this->energy_reference_cal_ = this->energy_reference_ / this->energy_cal_;
```
jesserockz: his could all technically be calculated in python during codegen so that the device just has the values baked into the firmware

esphome/components/bl0940/bl0940.cpp
```diff
@@ -122,6 +317,26 @@ void BL0940::received_package_(const DataPacket *data) const {
 
 void BL0940::dump_config() {  // NOLINT(readability-function-cognitive-complexity)
   ESP_LOGCONFIG(TAG, "BL0940:");
+  ESP_LOGCONFIG(TAG, "  TUYA MODE: %s", TRUEFALSE(this->tuya_mode_enabled_));
```
jesserockz: https://developers.esphome.io/architecture/logging/

esphome/components/bl0940/bl0940.cpp
```diff
@@ -46,35 +62,207 @@ void BL0940::loop() {
...
+      this->energy_calibration_callback_(this->energy_calibration_->state);
+    }
+  }
+#endif
+
+#ifdef USE_BUTTON
+  if (this->reset_calibration_button_ != nullptr) {
+    this->reset_calibration_button_->add_on_press_callback([this]() { this->reset_calibration_callback_(); });
+  }
+#endif
```
jesserockz: Rather than the main component needing to know about the existence of the button, change `reset_calibration_callback_` to be public and call it from the button class `on_press` function.

The same can be done for the `number` entities and keeps the main class cleaner.

jesserockz (changes requested):

esphome/components/bl0940/bl0940.cpp
```diff
@@ -7,37 +7,53 @@ namespace bl0940 {
...
 
-static const uint8_t BL0940_WRITE_COMMAND = 0xA0;  // 0xA8 according to documentation
 static const uint8_t BL0940_REG_I_FAST_RMS_CTRL = 0x10;
 static const uint8_t BL0940_REG_MODE = 0x18;
 static const uint8_t BL0940_REG_SOFT_RESET = 0x19;
 static const uint8_t BL0940_REG_USR_WRPROT = 0x1A;
 static const uint8_t BL0940_REG_TPS_CTRL = 0x1B;
 
-const uint8_t BL0940_INIT[5][6] = {
+const uint8_t BL0940_INIT[5][5] = {
```
kbx81: Static, too?

esphome/components/bl0940/bl0940.h
kbx81: I don't think we have a preference in terms of ESPHome. 😅 Probably fine as long as it doesn't cause a problem. 🙂

esphome/components/bl0940/sensor.py
dan-s-github: These actually allow to override the a single reference value that has been calculated as they take precedence over the calculated values. This allows to fine tune a single reference value as required at build time.

esphome/components/bl0940/sensor.py
dan-s-github: would legacy_mode be a better option?

esphome/components/bl0940/__init__.py
dan-s-github: moved as suggested

esphome/components/bl0940/sensor.py
dan-s-github: added validation check to handle this case

esphome/components/bl0940/sensor.py
dan-s-github: renamed to legacy_mode

esphome/components/bl0940/bl0940.cpp
dan-s-github: this no longer exists in cpp code

esphome/components/bl0940/bl0940.cpp
dan-s-github: defaults are now set in sensor.py

esphome/components/bl0940/bl0940.cpp
dan-s-github: calculations based on configuration settings are now done during config validation

esphome/components/bl0940/bl0940.cpp
dan-s-github: reduced number of log statements base on recommendations

esphome/components/bl0940/bl0940.cpp
dan-s-github: reset is no longer using the callback but I kept the number callbacks so that I can reset the UI values when reset button is pressed

esphome/components/bl0940/bl0940.cpp
dan-s-github: made it static

esphome/components/bl0940/bl0940.cpp
```diff
@@ -87,41 +194,70 @@ float BL0940::update_temp_(sensor::Sensor *sensor, ube16_t temperature) const {
...
   }
 
   ESP_LOGV(TAG, "BL0940: U %fV, I %fA, P %fW, Cnt %" PRId32 ", ∫P %fkWh, T1 %f°C, T2 %f°C", v_rms, i_rms, watt, cf_cnt,
            total_energy_consumption, tps1, tps2);
 }
 
 void BL0940::dump_config() {  // NOLINT(readability-function-cognitive-complexity)
-  ESP_LOGCONFIG(TAG, "BL0940:");
+  ESP_LOGCONFIG(TAG,
+                "\nBL0940:\n"
```
kbx81: Not necessary
```suggestion
                "BL0940:\n"
```

esphome/components/bl0940/bl0940.cpp
```diff
@@ -87,41 +194,70 @@ float BL0940::update_temp_(sensor::Sensor *sensor, ube16_t temperature) const {
...
+                "  Current reference: %f\n"
+                "  Energy reference: %f\n"
+                "  Power reference: %f\n"
+                "  Voltage reference: %f\n"
+                "  ------------------\n",
+                TRUEFALSE(this->legacy_mode_enabled_), this->read_command_, this->write_command_,
+                this->current_reference_, this->energy_reference_, this->power_reference_, this->voltage_reference_);
+#ifdef USE_NUMBER
+  ESP_LOGCONFIG(TAG,
+                "\nBL0940:\n"
```
kbx81: Same
```suggestion
                "BL0940:\n"
```

esphome/components/bl0940/bl0940.h
```diff
@@ -1,58 +1,37 @@
...
+  uint24_t RESERVED1;      // reserved
+  int24_le_t watt;         // 0x08
+  uint24_t RESERVED2;      // reserved
+  uint24_le_t cf_cnt;      // 0x0A
+  uint24_t RESERVED3;      // reserved
+  uint16_le_t tps1;        // 0x0c
+  uint8_t RESERVED4;       // value of 0x00
+  uint16_le_t tps2;        // 0x0c
+  uint8_t RESERVED5;       // value of 0x00
+  uint8_t checksum;        // checksum
```
kbx81: I've got no experience using `uint24` types, but, in general, you'll likely see slightly lower memory usage by arranging from widest to narrowest:
```suggestion
  uint24_le_t i_fast_rms;  // 0x00
  uint24_le_t i_rms;       // 0x04
  uint24_t RESERVED0;      // reserved
  uint24_le_t v_rms;       // 0x06
  uint24_t RESERVED1;      // reserved
  int24_le_t watt;         // 0x08
  uint24_t RESERVED2;      // reserved
  uint24_le_t cf_cnt;      // 0x0A
  uint24_t RESERVED3;      // reserved
  uint16_le_t tps1;        // 0x0c
  uint16_le_t tps2;        // 0x0c
  uint8_t RESERVED4;       // value of 0x00
  uint8_t RESERVED5;       // value of 0x00
  uint8_t checksum;        // checksum
  uint8_t frame_header;    // value of 0x58 according to en docs but 0x55 in cn docs and Tasmota real world tests
```

esphome/components/bl0940/bl0940.h
```diff
@@ -84,26 +84,59 @@ class BL0940 : public PollingComponent, public uart::UARTDevice {
...
+  number::Number *energy_calibration_{nullptr};
+#endif
+
   // Max difference between two measurements of the temperature. Used to avoid noise.
   float max_temperature_diff_{0};
+
+  bool legacy_mode_enabled_ = true;
+
+  uint8_t read_command_;
+  uint8_t write_command_;
```
kbx81: Move these below all these `float` types to reduce padding, lowering memory use

esphome/components/bl0940/bl0940.h
```diff
@@ -84,26 +84,59 @@ class BL0940 : public PollingComponent, public uart::UARTDevice {
...
   // Divide by this to turn into kWh
-  float energy_reference_ = BL0940_EREF;
+  float energy_reference_;
+  float energy_reference_cal_;
 
-  float update_temp_(sensor::Sensor *sensor, ube16_t packed_temperature) const;
+  float update_temp_(sensor::Sensor *sensor, uint16_le_t packed_temperature) const;
 
-  static uint32_t to_uint32_t(ube24_t input);
+  uint32_t prev_cf_cnt_ = 0;
```
kbx81: Move this above the floats in the list above

esphome/components/bl0940/bl0940.h
```diff
@@ -84,26 +84,59 @@ class BL0940 : public PollingComponent, public uart::UARTDevice {
...
-  static int32_t to_int32_t(sbe24_t input);
+  bool validate_checksum_(DataPacket *data);
+  void received_package_(DataPacket *data);
 
-  static bool validate_checksum(const DataPacket *data);
+  // HA calibration options
+  float current_cal_{1};
+  float voltage_cal_{1};
+  float power_cal_{1};
+  float energy_cal_{1};
```
kbx81: Move these up with the other floats above

esphome/components/bl0940/bl0940.h
dan-s-github: don't think this can be changed as it is a representation of the datapacket received from the sensor

esphome/components/bl0940/bl0940.h
dan-s-github: moved

esphome/components/bl0940/bl0940.h
dan-s-github: moved

esphome/components/bl0940/bl0940.h
dan-s-github: moved

esphome/components/bl0940/bl0940.cpp
dan-s-github: removed initial \n

esphome/components/bl0940/bl0940.cpp
dan-s-github: removed initial \n

esphome/components/bl0940/bl0940.cpp
```diff
@@ -46,35 +44,144 @@ void BL0940::loop() {
...
+  this->recalibrate_();
+}
+void BL0940::energy_calibration_callback_(float state) {
+  this->energy_cal_ = this->calculate_calibration_value_(state);
+  ESP_LOGD(TAG, "update energy calibration state: %f", this->energy_cal_);
+  this->recalibrate_();
+}
+
+void BL0940::recalibrate_() {
+  ESP_LOGD(TAG, "Recalibrating reference values");
```
kbx81: The logging in these methods should really be verbose with `ESP_LOGV`; it's not necessary for every method to log something outside of troubleshooting.

dan-s-github: changed log levels on the callback logs but added a more useful log after recalibration. Which would be useful when updating the default calibration.

kbx81 (dismissed):
LGTM, thanks! 👍🏻
Just left a couple comments on the docs. 🙂

esphome/components/bl0940/__init__.py
```diff
@@ -1 +1,8 @@
-CODEOWNERS = ["@tobias-"]
+import esphome.codegen as cg
+
+CODEOWNERS = ["@tobias-", "@dan-s-github"]
+
+DEPENDENCIES = ["uart"]
```
jesserockz: I don't think this works as expected as the init file is not being loaded as a component, therefore this `DEPENDENCIES` is not being read.

dan-s-github: moved DEPENDENCIES declaration back to sensor.py

esphome/components/bl0940/number/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # Process all calibration types
+    for cal_type, setter_method in [
+        (CONF_CURRENT_CALIBRATION, "set_current_calibration_number"),
+        (CONF_VOLTAGE_CALIBRATION, "set_voltage_calibration_number"),
+        (CONF_POWER_CALIBRATION, "set_power_calibration_number"),
+        (CONF_ENERGY_CALIBRATION, "set_energy_calibration_number"),
+    ]:
+        if cal_type in config:
+            conf = dict(config[cal_type])
```
jesserockz:
```suggestion
        if conf := config.get(cal_type):
```

esphome/components/bl0940/number/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            conf[CONF_MODE] = number.NumberMode.NUMBER_MODE_BOX
```
jesserockz: We don't force settings like this.
You can set a default similar to min/max and step
```suggestion
```

esphome/components/bl0940/number/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            var = await number.new_number(
+                conf,
+                min_value=conf.get(CONF_MIN_VALUE),
+                max_value=conf.get(CONF_MAX_VALUE),
+                step=conf.get(CONF_STEP),
+            )
+            await cg.register_component(var, conf)
+
+            if CONF_RESTORE_VALUE in config:
+                cg.add(var.set_restore_value(config[CONF_RESTORE_VALUE]))
```
jesserockz:
```suggestion
            if restore_value := config.get(CONF_RESTORE_VALUE):
                cg.add(var.set_restore_value(restore_value))
```

esphome/components/bl0940/number/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    )
+    .extend(cv.polling_component_schema("never")),
+    validate_min_max,
+)
+
+# Configuration schema for BL0940 numbers
+CONFIG_SCHEMA = cv.Schema(
+    {
+        cv.GenerateID(): cv.declare_id(BL0940Number),
+        cv.Required(CONF_BL0940_ID): cv.use_id(BL0940),
```
jesserockz: We can auto link if there is only one
```suggestion
        cv.GenerateID(CONF_BL0940_ID): cv.use_id(BL0940),
```

esphome/components/bl0940/number/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+CalibrationNumber = bl0940_ns.class_(
+    "CalibrationNumber", number.Number, cg.PollingComponent
+)
+
+
+def validate_min_max(config):
+    if config[CONF_MAX_VALUE] <= config[CONF_MIN_VALUE]:
+        raise cv.Invalid("max_value must be greater than min_value")
+    if config[CONF_MIN_VALUE] < -50 or config[CONF_MAX_VALUE] > 50:
+        raise cv.Invalid("max calibration range is +/-50 percent")
```
jesserockz: This should be a `cv.Range` validator on the individual config items
```suggestion
```

esphome/components/bl0940/number/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+CALIBRATION_SCHEMA = cv.All(
+    number.number_schema(
+        CalibrationNumber,
+        entity_category=ENTITY_CATEGORY_CONFIG,
+        unit_of_measurement=UNIT_PERCENT,
+    )
+    .extend(
+        {
+            cv.Optional(CONF_MAX_VALUE, default=10): cv.float_,
```
jesserockz:
```suggestion
            cv.Optional(CONF_MAX_VALUE, default=10): cv.All(cv.float_, cv.Range(max=50)),
```

esphome/components/bl0940/number/calibration_number.h
```diff
@@ -0,0 +1,29 @@
...
+  void setup() override;
+  void update() override;
+  void dump_config() override;
+  float get_setup_priority() const override { return setup_priority::HARDWARE; }
+
+  void set_restore_value(bool restore_value) { this->restore_value_ = restore_value; }
+
+ protected:
+  void control(float value) override;
+  float initial_value_{0};
```
jesserockz: Not used?

esphome/components/bl0940/number/calibration_number.h
```diff
@@ -0,0 +1,29 @@
...
+  bool restore_value_{true};
+  optional<std::function<optional<float>()>> f_;
```
jesserockz: Copy pasta from template number?

esphome/components/bl0940/number/calibration_number.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "esphome/components/number/number.h"
+#include "esphome/core/component.h"
+#include "esphome/core/preferences.h"
+
+namespace esphome {
+namespace bl0940 {
+
+class CalibrationNumber : public number::Number, public PollingComponent {
```
jesserockz: Does not need to be polling
```suggestion
class CalibrationNumber : public number::Number, public Component {
```

esphome/components/bl0940/number/calibration_number.cpp
```diff
@@ -0,0 +1,52 @@
...
+void CalibrationNumber::update() {
+  if (!this->f_.has_value())
+    return;
+
+  auto val = (*this->f_)();
+  if (!val.has_value())
+    return;
+
+  this->publish_state(*val);
+}
```
jesserockz:
```suggestion
```

esphome/components/bl0940/number/calibration_number.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "esphome/core/log.h"
+
+namespace esphome {
+namespace bl0940 {
+
+static const char *const TAG = "bl0940.number";
+
+void CalibrationNumber::setup() {
+  if (this->f_.has_value())
+    return;
```
jesserockz:
```suggestion
```

esphome/components/bl0940/number/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            cv.Optional(CONF_MIN_VALUE, default=-10): cv.float_,
+            cv.Optional(CONF_STEP, default=0.1): cv.positive_float,
+            cv.Optional(CONF_RESTORE_VALUE): cv.boolean,
+        }
+    )
+    .extend(cv.polling_component_schema("never")),
```
jesserockz:
```suggestion
    .extend(cv.COMPONENT_SCHEMA),
```

esphome/components/bl0940/button/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    "CalibrationResetButton", button.Button, cg.Component
+)
+
+CONFIG_SCHEMA = cv.All(
+    button.button_schema(
+        CalibrationResetButton,
+        entity_category=ENTITY_CATEGORY_CONFIG,
+        icon=ICON_RESTART,
+    )
+    .extend({cv.Required(CONF_BL0940_ID): cv.use_id(BL0940)})
```
jesserockz:
```suggestion
    .extend({cv.GenerateID(CONF_BL0940_ID): cv.use_id(BL0940)})
```

esphome/components/bl0940/button/calibration_reset_button.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include "esphome/core/component.h"
+#include "esphome/components/button/button.h"
+
+namespace esphome {
+namespace bl0940 {
+
+class BL0940;  // Forward declaration of BL0940 class
+
+class CalibrationResetButton : public button::Button, public Component {
```
jesserockz:
```suggestion
class CalibrationResetButton : public button::Button, public Component, public Parented<BL0940> {
```

esphome/components/bl0940/button/calibration_reset_button.h
```diff
@@ -0,0 +1,23 @@
...
+ public:
+  void dump_config() override;
+  void set_bl0940(BL0940 *bl0940) { bl0940_ = bl0940; }
```
jesserockz:
```suggestion
```

esphome/components/bl0940/button/calibration_reset_button.h
```diff
@@ -0,0 +1,23 @@
...
+
+  void press_action() override;
+
+ protected:
+  BL0940 *bl0940_{nullptr};
```
jesserockz:
```suggestion
```

esphome/components/bl0940/button/calibration_reset_button.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+static const char *const TAG = "calibration_reset.button";
+
+void CalibrationResetButton::dump_config() { LOG_BUTTON("", "Calibration Reset Button", this); }
+
+void CalibrationResetButton::press_action() {
+  ESP_LOGI(TAG, "Resetting calibration defaults...");
+  if (bl0940_ != nullptr) {
+    bl0940_->reset_calibration();
+  }
```
jesserockz: With the above change to parent, plus not possible to be null due to validations
```suggestion
  this->parent_->reset_calibration();
```

esphome/components/bl0940/button/calibration_reset_button.cpp
```diff
@@ -0,0 +1,22 @@
+#include "calibration_reset_button.h"
+#include "../bl0940.h"
+#include "esphome/core/hal.h"
+#include "esphome/core/log.h"
+#include "esphome/core/application.h"
+
+namespace esphome {
+namespace bl0940 {
+
+static const char *const TAG = "calibration_reset.button";
```
jesserockz: Tag should always start with the component name
```suggestion
static const char *const TAG = "bl0940.button.calibration_reset";
```

esphome/components/bl0940/button/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+async def to_code(config):
+    bl0940 = await cg.get_variable(config[CONF_BL0940_ID])
+
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await button.register_button(var, config)
+
+    cg.add(var.set_bl0940(bl0940))
```
jesserockz:
```suggestion
    var = await button.new_button(config)
    await cg.register_component(var, config)
    await cg.register_parented(var, config[CONF_BL0940_ID])
```

jesserockz (changes requested):

esphome/components/bl0940/number/calibration_number.h
dan-s-github: was used in setup but simplified that code and this is no no longer required

esphome/components/bl0940/number/calibration_number.h
dan-s-github: off course, removed now

jesserockz (approved):

jesserockz:
Sorry for the delays @dan-s-github. Please rebase the docs PR and we can then get this merged

dwmw2:
While you're working on bl0940, do you feel like fixing the problem that was reported in https://github.com/esphome/esphome/issues/10136 for bl0942, which I'm trying to fix with #10646

dan-s-github:
recreated docs pr after converting rst to md. https://github.com/esphome/esphome-docs/pull/5340

Unfortunately the devcontainer is not working on my m2 macbook

dan-s-github:
> While you're working on bl0940, do you feel like fixing the problem that was reported in #10136 for bl0942, which I'm trying to fix with #10646

I haven't noticed this issue so far, might be good to get the current version out, then it can be fixed in a followup pr

PR #8164, opened 2025-01-31
[audio, i2s_audio, speaker] Media Player Components PR2

kahrendt:

This is pull request 2 of 9 to add a new speaker media player component with several other helper components. It is built off #8163.

PR2 adds:

- A new PlatformIO library ``esp-audio-libs`` ([GitHub](https://github.com/esphome/esp-audio-libs), [PlatformIO](https://registry.platformio.org/libraries/esphome/esp-audio-libs)) that handles audio decoding, resampling, and other miscellaneous audio functions.
- Adds ``AUDIO_DEVICE_SCHEMA`` and helper functions to enable speaker components to auto-configure based on the parent speakers settings. If users override the auto-configured settings, their changes are validated to ensure compatibility.
- Rewrites the ``AudioStreamInfo`` as a class with many helper functions to convert between bytes, samples, frames, and duration.
- Adds pausing to the parent ``speaker`` component (and implements it in the ``i2s_audio`` speaker component). This will reduce latency when pausing the media player when several virtual speakers are stacked.
- Adds a playback duration callback to the parent ``speaker`` component (and implements it in the ``i2s_audio`` speaker component). This precisely tracks the audio duration written to the DMA buffers to potentially enable synced playback in a future update.
- Adds some helper functions for ``AudioFile`` to log audio file types to be used in future PRs.

Note that this requires changes to the core const.py file from #8163, so it cannot be easily tested as an external component until that is merged! Once merged, you should be able to test if you build with the ESPHome dev branch.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/audio/__init__.py
```diff
@@ -1,9 +1,121 @@
...
 audio_ns = cg.esphome_ns.namespace("audio")
 
+AudioFile = audio_ns.struct("AudioFile")
+AudioFileType = audio_ns.enum("AudioFileType", is_class=True)
+AUDIO_FILE_TYPE_ENUM = {
+    "NONE": AudioFileType.NONE,
+    "WAV": AudioFileType.WAV,
+    "MP3": AudioFileType.MP3,
+    "FLAC": AudioFileType.FLAC,
+}
```
nielsnl68, line 16: Please make this a registry system so we can add more options on the fly. See for an example the filter option etc.

Thanks

esphome/components/audio/audio.h
```diff
@@ -1,21 +1,139 @@
...
+enum class AudioFileType : uint8_t {
+  NONE = 0,
+#ifdef USE_AUDIO_FLAC_SUPPORT
+  FLAC,
+#endif
+#ifdef USE_AUDIO_MP3_SUPPORT
+  MP3,
+#endif
+  WAV,
+};
```
nielsnl68, line 117: Instead of creating an fixed enum, Create a separate class for every Audio type, so that extending the audio types more easily in the future without the need to modify this file again and again.

nielsnl68 (changes requested):

esphome/components/audio/__init__.py
kahrendt: For now I'm not going to implement this, but I would welcome a PR down the road implementing it.

esphome/components/audio/audio.h
kahrendt: For now I'm not going to implement this, but I would welcome a PR down the road implementing it.

jesserockz (approved):

PR #8169, opened 2025-01-31
[resampler] Media Player Components PR7

kahrendt:
# What does this implement/fix?

This is pull request 7 of 9 to add a new speaker media player component with several other helper components. It is built off #8168.

PR7 adds a new ``resampler`` speaker component. It creates a virtual speaker where it resamples any audio played on the virtual speaker before sending it on to the configured output speaker. It can convert sample rates and bits per sample. By default, it inherits the output speaker's sample rate and bits per sample, but users may override those settings. These settings are validated against the output speaker's capabilities to ensure compatibility.

Volume control and pausing is passed onto to the output speaker. The audio played duration callback adjusts the received duration from the parent speaker to account for slightly mismatched timings due to the resampling process, so the resampler callback function has duration in reference to the input audio.

Note that this requires changes to the core ``const.py`` file from #8163, so it cannot be easily tested as an external component until that is merged! Once merged, you should be able to test if you build with the ESPHome dev branch.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):** not applicable

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#4628

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

speaker:
  - platform: resampler
    id: resampler_speaker_input_id
    output_speaker: i2s_speaker_id
    sample_rate: 48000
    bits_per_sample: 16

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8170, opened 2025-01-31
[mixer] Media Player Components PR8

kahrendt:

This is pull request 8 of 9 to add a new speaker media player component with several other helper components. It is built off #8169.

PR8 adds a new ``mixer`` speaker component. It creates a virtual speaker where it mixes the audio from all the source speakers. All source speakers must have the same sample rate to mix, so using the ``resampler`` speaker from #8169 is beneficial. If the user wants to avoid resampling, they can enable ``queue_mode``. This plays the audio from the highest listed speaker first. It attempts to reset the output speakers sample rate to match the inpu audio before playing from a new source speaker. This assumes the output speaker can handle different sample rates (which is usually possible in the case if it an ``i2s_audio`` speaker in primary mode). Individual source speakers can be ducked with the ``apply_ducking`` action.

Volume control is passed onto the output speaker. Pausing is handled directly in the ``mixer`` speaker and is not passed onto to the output speaker. The audio played duration callback applies the elapsed duration only to source speakers that have sent audio, so each source speaker's callback is in reference to its input audio.

Note that this requires changes to the core ``const.py`` file from #8163, so it cannot be easily tested as an external component until that is merged! Once merged, you should be able to test if you build with the ESPHome dev branch.

## Types of changes

- [ ] Other

- esphome/esphome-docs#4629

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

speaker:
  - platform: mixer
    id: mixer_speaker_id
    output_speaker: i2s_speaker_id
    queue_mode: false
    source_speakers:
      - id: announcement_spk_mixer_input
        timeout: never
        sample_rate: 16000
      - id: media_spk_mixer_input
        timeout: never

```

## Checklist:


esphome/components/mixer/speaker/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    PLATFORM_ESP32,
+)
+from esphome.core.entity_helpers import inherit_property_from
+import esphome.final_validate as fv
+
+AUTO_LOAD = ["audio"]
+CODEOWNERS = ["@kahrendt"]
+
+mixer_speaker_ns = cg.esphome_ns.namespace("mixer_speaker")
+MixerSpeaker = mixer_speaker_ns.class_("MixerSpeaker", cg.Component, speaker.Speaker)
```
jesserockz: It does not derive from Speaker:
```c++
class MixerSpeaker : public Component {
```

```suggestion
MixerSpeaker = mixer_speaker_ns.class_("MixerSpeaker", cg.Component)
```

jesserockz (changes requested):

jesserockz (approved):

PR #8171, opened 2025-01-31
[speaker] Media Player Components PR9

### LINKED ISSUE #1848: Support more audio codecs, not just lossy MP3
Media Player currently seems to support only MP3 streams. MP3 is a lossy format, sound quality is degraded.
[Home Assistant supports](https://www.youtube.com/watch?v=SEH-DxOsywg) the [Music Assistant](https://github.com/music-assistant/hass-music-assistant) integration which by default renders the audio streams in FLAC, which is a lossless format.
Would be nice to enable support for other codecs too:
- lossless FLAC
- ncompressed PCM up to 48k stereo
- open source OPUS (which is lossy, but way better than MP3)

Since only MP3 is supported, one needs to use other firmware than ESPHome, like squeezelite. The goal would be not to have to rely on third party stuff, but remain in the ESPHome ecosystem.

**Additional context**
Adding more codecs of course needs PSRAM for ESP32, which is not a problem as plenty of such modules are available. This just needs to be pointed out properly in the documentation.

### LINKED ISSUE #1864: Port i2s_audio to idf framework

It would reduce flash size if the `i2s_audio` integration could be ported to use the `idf` framework.

Combining `i2s_audio` with `bluetooth_proxy` fills up the flash, resulting in crashes.
See [this](https://github.com/esphome/issues/issues/3481) issue.

**Additional context**

### LINKED ISSUE #1898: Media player volume up/down step size
The current volume up/down actions only step by 10%. For different speaks the 10% step size can either be too big or too small. Would be nice to have an option to customize the step size.

**Additional context**

### LINKED ISSUE #2429: add esp-idf media_player support

Please add support for media_player component on esp-idf framework.  Many esphome features now recommend using esp-idf framework (i.e. bluetooth proxy) however media player still only work on arduino framework.

Currently media player is only available on arduino framework and when trying to compile this code:
````
i2s_audio:
  - id: i2s_out
    i2s_lrclk_pin: 44
    i2s_bclk_pin: 36

media_player:
  - platform: i2s_audio
    name: Media Player
    id: media_out
    i2s_audio_id: i2s_out
    dac_type: external
    i2s_dout_pin: 35
    mode: mono
````
I get: This feature is only available with frameworks [‘arduino’].

**Additional context**

### LINKED ISSUE #2704: Playing a embedded audio file

Currently media player is only able to play media from an url. I'd like to play media from an embedded file (e.g. in a byte array). 

I've built a door bell based on esphome with i2s_audio and media player and and amplifier. It works fine as long as the device has network connection. But i need a fallback, to play a sound even if there is no network connection. 

As a workarround i will try to use speaker play to play raw audio. But in that case i assume i will loose the codec support of media player. So the audio will propably be very limited by the space available.

**Additional context**

### PR DESCRIPTION
**Pull request by kahrendt**

This is pull request 9 of 9 to add a new speaker media player component with several other helper components. It is built off #8170.

PR9 adds a new speaker media player component that sends audio to another ESPHome speaker component. It supports two pipelines, one for announcements and one for regular media. Each pipeline requires a unique output speaker. Use a ``mixer`` speaker from #8170 to create two source speakers that are mixed and sent to another speaker component.

To increase the number of audio formats available and to minimize the load on the ESP32, you can use HA to transcode audio automatically before it is sent to the device. By default, it transcodes to ``FLAC``, which uses the CPU the least. You can set it to transcode to ``MP3``, ``WAV``, or disabled. You can specify the sample rate and number of channels for the transcoding, but by default it will match the output speaker's setting. If you disable HA transcoding, then the media player only supports FLAC, MP3, and WAV files.

This component may run without external memory, but it is not recommended. To minimize memory usage, only define the announcement pipeline, decrease the buffer size, and transcode to ``WAV`` with only 1 channel and a low sample rate.

Since this relies on the ``AudioReader`` class from #8166 , this component is only available for esp-idf.

Note that this requires changes to the core ``const.py`` file from #8163, so it cannot be easily tested as an external component until that is merged! Once merged, you should be able to test if you build with the ESPHome dev branch.

## Types of changes

- [ ] Other

- Closes esphome/feature-requests#2704
- Closes esphome/feature-requests#2429
- Closes esphome/feature-requests#1864
- Closes esphome/feature-requests#1898
- Closes esphome/feature-requests#1848

- esphome/esphome-docs#4630

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

speaker:
  - platform: i2s_audio
    id: speaker_id
    dac_type: external
    i2s_dout_pin: GPIOXX
    sample_rate: 48000
    channel: stereo
    timeout: never
    buffer_duration: 100ms
  - platform: mixer
    id: mixer_speaker_id
    output_speaker: speaker_id
    source_speakers:
      - id: announcement_spk_mixer_input
      - id: media_spk_mixer_input
  - platform: resampler
    id: media_spk_resampling_input
    output_speaker: media_spk_mixer_input
  - platform: resampler
    id: announcement_spk_resampling_input
    output_speaker: announcement_spk_mixer_input

media_player:
  - platform: speaker
    name: "Speaker Media Player"
    id: speaker_media_player_id
    buffer_size: 2000000
    media_pipeline:
        speaker: media_spk_resampling_input
        sample_rate: 48000
        num_channels: 2
        format: FLAC
    announcement_pipeline:
        speaker: announcement_spk_resampling_input
        sample_rate: 48000
        num_channels: 1
        format: FLAC

```

## Checklist:


esphome/components/audio/__init__.py
```diff
@@ -1,9 +1,121 @@
...
 CODEOWNERS = ["@kahrendt"]
 audio_ns = cg.esphome_ns.namespace("audio")
 
+AudioFile = audio_ns.struct("AudioFile")
+AudioFileType = audio_ns.enum("AudioFileType", is_class=True)
+AUDIO_FILE_TYPE_ENUM = {
+    "NONE": AudioFileType.NONE,
+    "WAV": AudioFileType.WAV,
+    "MP3": AudioFileType.MP3,
+    "FLAC": AudioFileType.FLAC,
```
nielsnl68: Please make this a registry system so we can add more options on the fly. See for an example the filter option etc.

Thanks

esphome/components/audio/__init__.py
```diff
@@ -1,9 +1,121 @@
...
+CONF_MAX_SAMPLE_RATE = "max_sample_rate"
+
+
 CONFIG_SCHEMA = cv.All(
     cv.Schema({}),
 )
+
+AUDIO_COMPONENT_SCHEMA = cv.Schema(
+    {
+        cv.Optional(CONF_BITS_PER_SAMPLE): cv.int_range(8, 32),
```
nielsnl68: I would make this an enum for 8,16,24,32  All values in between are invalid anyway,

esphome/components/audio/audio.cpp
```diff
@@ -0,0 +1,67 @@
...
+  *total_frames = unprocessable_frames;
+  return playback_ms;
+}
+
+bool AudioStreamInfo::operator==(const AudioStreamInfo &rhs) const {
+  return (this->bits_per_sample_ == rhs.get_bits_per_sample()) && (this->channels_ == rhs.get_channels()) &&
+         (this->sample_rate_ == rhs.get_sample_rate());
+}
+
+const char *audio_file_type_to_string(AudioFileType file_type) {
```
nielsnl68: Make this more flexible so new audio types can easily added without the need of changing the function.

esphome/components/audio/audio.h
```diff
@@ -1,21 +1,139 @@
...
+enum class AudioFileType : uint8_t {
+  NONE = 0,
+#ifdef USE_AUDIO_FLAC_SUPPORT
+  FLAC,
+#endif
+#ifdef USE_AUDIO_MP3_SUPPORT
+  MP3,
+#endif
+  WAV,
+};
```
nielsnl68: As above making the audio component future proof.
Instead of creating a fixed enum list here create 3 extra components that are specific for the chosen audio type. And make it easy to implement new types as well.

esphome/components/audio/audio_decoder.h
```diff
@@ -0,0 +1,135 @@
...
+// Only used within the AudioDecoder class; conveys the state of the particular file type decoder
+enum class FileDecoderState : uint8_t {
+  MORE_TO_PROCESS,     // Successsfully read a file chunk and more data is available to decode
+  IDLE,                // Not enough data to decode, waiting for more to be transferred
+  POTENTIALLY_FAILED,  // Decoder encountered a potentially recoverable error if more file data is available
+  FAILED,              // Decoder encoutnered an uncrecoverable error
+  END_OF_FILE,         // The specific file decoder knows its the end of the file
+};
+
+class AudioDecoder {
```
nielsnl68: Instead of putting every audio type decoding option inside one class, please create a decode class per audio type. So extending is more flexible.

esphome/components/audio/audio_reader.cpp
nielsnl68, line 1: From the looks of it this reader only from http, so my suggestion is to change the filename to: **audion_http_reader.\***.

esphome/components/audio/audio_reader.h
nielsnl68, line 1: From the looks of it this reader only from http, so my suggestion is to change the filename to:
**audion_http_reader.\***

nielsnl68 (changes requested):
Great work Kevin,
My only important request is to rework the audio type option structure so we can easily add new audio types without modify the whole audio class files for them.

esphome/components/audio/audio_reader.cpp
kahrendt: It also reads files compiled into the firmware as an AudioFile struct.

esphome/components/audio/audio_reader.h
kahrendt: It also reads files compiled into the firmware as an AudioFile struct.

esphome/components/audio/__init__.py
kahrendt: Technically they are valid as long as they are aligned to the byte. For example, if  you have 10 bits per sample, then so long as you use 2 bytes to store the sample where the 10 most significant bits represent the audio and the 6 least significant bits are zeros, then it will play fine.

The ``i2s_audio`` speaker already does restrict the configuration further to the byte since that's required by the driver, but it will still play audio with lower bits per sample if they are aligned to the byte.

esphome/components/audio/__init__.py
kahrendt: As discussed in PR2, I'm holding off on changing this now, but will welcome a future PR with it!

esphome/components/media_player/__init__.py
```diff
@@ -21,6 +20,17 @@
 
 MediaPlayer = media_player_ns.class_("MediaPlayer")
 
+media_player_ns = cg.esphome_ns.namespace("media_player")
```
jesserockz: Duplicate from 4 lines above
```suggestion
```

esphome/components/speaker/media_player/__init__.py
```diff
@@ -0,0 +1,469 @@
...
+        )
+    if on_volume := config.get(CONF_ON_VOLUME):
+        await automation.build_automation(
+            var.get_volume_trigger(),
+            [(cg.float_, "x")],
+            on_volume,
+        )
+
+    if files_list := config.get(CONF_FILES):
+        for file_config in files_list:
```
jesserockz: Will need indentation adjustment below too, but you can compact this to a single line
```suggestion
    for file_config in config.get(CONF_FILES, []):
```

esphome/components/speaker/media_player/__init__.py
```diff
@@ -0,0 +1,469 @@
...
+                (
+                    "file_type",
+                    media_file_type,
+                ),
+            )
+
+            cg.new_Pvariable(
+                file_config[CONF_ID],
+                media_files_struct,
+            )
```
jesserockz: In combination with the above
```suggestion
        data, media_file_type = _read_audio_file_and_type(file_config)

        rhs = [HexInt(x) for x in data]
        prog_arr = cg.progmem_array(file_config[CONF_RAW_DATA_ID], rhs)

        media_files_struct = cg.StructInitializer(
            audio.AudioFile,
            (
                "data",
                prog_arr,
            ),
            (
                "length",
                len(rhs),
            ),
            (
                "file_type",
                media_file_type,
            ),
        )

        cg.new_Pvariable(
            file_config[CONF_ID],
            media_files_struct,
        )
```

esphome/components/speaker/media_player/__init__.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+@automation.register_action(
+    "speaker_media_player.play_on_device_media_file",
```
jesserockz: We format the action name like this for some other components too
```suggestion
    "media_player.speaker.play_on_device_media_file",
```

esphome/components/speaker/media_player/__init__.py
```diff
@@ -0,0 +1,469 @@
...
+            raise cv.Invalid(
+                "The announcement and media pipelines cannot use the same speaker. Use the `mixer` speaker component to create two source speakers."
+            )
+
+    return config
+
+
+def _validate_supported_local_file(config):
+    if files_list := config.get(CONF_FILES):
+        for file_config in files_list:
```
jesserockz:
```suggestion
    for file_config in config.get(CONF_FILES, []):
```

esphome/components/speaker/media_player/__init__.py
```diff
@@ -0,0 +1,469 @@
...
+            _, media_file_type = _read_audio_file_and_type(file_config)
+            if str(media_file_type) == str(audio.AUDIO_FILE_TYPE_ENUM["NONE"]):
+                raise cv.Invalid("Unsupported local media file.")
+            if not config[CONF_CODEC_SUPPORT_ENABLED] and str(media_file_type) != str(
+                audio.AUDIO_FILE_TYPE_ENUM["WAV"]
+            ):
+                # Only wav files are supported
+                raise cv.Invalid(
+                    f"Unsupported local media file type, set {CONF_CODEC_SUPPORT_ENABLED} to true or convert the media file to wav"
+                )
```
jesserockz:
```suggestion
        _, media_file_type = _read_audio_file_and_type(file_config)
        if str(media_file_type) == str(audio.AUDIO_FILE_TYPE_ENUM["NONE"]):
            raise cv.Invalid("Unsupported local media file.")
        if not config[CONF_CODEC_SUPPORT_ENABLED] and str(media_file_type) != str(
            audio.AUDIO_FILE_TYPE_ENUM["WAV"]
        ):
            # Only wav files are supported
            raise cv.Invalid(
                f"Unsupported local media file type, set {CONF_CODEC_SUPPORT_ENABLED} to true or convert the media file to wav"
            )
```

esphome/components/speaker/media_player/__init__.py
```diff
@@ -0,0 +1,469 @@
...
+
+        file_type: str = puremagic.from_string(data)
+    except ImportError:
+        try:
+            from magic import Magic
+
+            magic = Magic(mime=True)
+            file_type: str = magic.from_buffer(data)
+        except ImportError as exc:
+            raise cv.Invalid("Please install puremagic") from exc
```
jesserockz: No need to test for this in internal components
```suggestion
    import puremagic

    file_type: str = puremagic.from_string(data)
```

esphome/components/speaker/media_player/__init__.py
```diff
@@ -0,0 +1,469 @@
...
+from esphome.core import CORE, HexInt
+from esphome.core.entity_helpers import inherit_property_from
+from esphome.external_files import download_content
+
+_LOGGER = logging.getLogger(__name__)
+
+AUTO_LOAD = ["audio", "psram"]
+
+CODEOWNERS = ["@kahrendt", "@synesthesiam"]
+DOMAIN = "file"
```
jesserockz: Should this be "media"/"media_player" or similar?

jesserockz (changes requested):

esphome/components/speaker/media_player/__init__.py
kahrendt: To be honest, I'm not sure what the purpose of ``DOMAIN`` is. The ``file`` domain was just a copy-pasta issue. I changed it to ``media_player``.

jesserockz: It becomes the name of the folder inside `.esphome` where the files are downloaded to

jesserockz (approved):

PR #8175, opened 2025-02-01
fix: Add support for PAS CO2 sensor.

ozonejunkieau:

Adds Support for Infineon XENSIV PASCO2 CO2 Sensor.

This is largely based on the existing #5883 code with minor changes to remove any excessive delay calls. If this is not the proper way to contribute these changes please let me know and I'll update. 

## Types of changes

- [ ] Other

- esphome/esphome-docs#3427

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
sensor:
  - platform: pasco2
    co2:
      name: co2_name
    update_interval: 60s
```

## Checklist:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -306,6 +306,7 @@ esphome/components/online_image/* @guillempages
 esphome/components/opentherm/* @olegtarasov
 esphome/components/ota/* @esphome/core
 esphome/components/output/* @esphome/core
+esphome/components/pasco2/* @circuitvalley @ozonejunkieau
 esphome/components/pca6416a/* @Mat931
 esphome/components/pca9554/* @clydebarrow @hwstar
 esphome/components/pcf85063/* @brogon
--- a/esphome/components/pasco2/__init__.py
+++ b/esphome/components/pasco2/__init__.py
@@ -1,0 +1,1 @@
+CODEOWNERS = ["@circuitvalley", "@ozonejunkieau"]
--- a/esphome/components/pasco2/automation.h
+++ b/esphome/components/pasco2/automation.h
@@ -1,0 +1,22 @@
+#pragma once
+
+#include "esphome/core/component.h"
+#include "esphome/core/automation.h"
+#include "pasco2.h"
+
+namespace esphome {
+namespace pasco2 {
+
+template<typename... Ts> class PerformForcedCalibrationAction : public Action<Ts...>, public Parented<PASCO2Component> {
+ public:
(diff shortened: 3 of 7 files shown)


mgiacomelli:
Is forcing a calibration actually working?  Testing on my device it does not calibrate, and looking at the code, perform_forced_calibration() sets the appropriate bit in XENSIV_PASCO2_REG_MEAS_CFG to start calibration, but then calls start_measurement_ which immediately overwrites the value in the XENSIV_PASCO2_REG_MEAS_CFG register.  It looks to me like the calibration sequence is never actually started.  

I spent a while trying to get it to work following the Arduino example (code attached), but I never get it to set BOC_CFG = 0, indicating that calibration is finished.  Maybe someone else can spot what I'm missing?  

[changes.patch](https://github.com/user-attachments/files/19239652/changes.patch)

mgiacomelli:
The issue was that I used write_byte rather than write_byte_16 and so wasn't setting the lower byte of a register.  I forked this PR and committed a working calibration function here if you want to take a look at it:  

https://github.com/mgiacomelli/esphome/tree/pr/8175

mgiacomelli:
While this could be a problem with my sensor, I have noticed a few time points that give a CO2 level of either 2^15 -1 or very close to 2^16:
image

Unfortunately I didn't have the logs recording, so I'm not sure if a warning was generated but presumably this is just corrupted data.  Maybe we should sanity check values since the valid range for the sensor is only up to 30,000 PPM?  Could reread any values that come out above 30,000 PPM before sending them.

davidkna:
@mgiacomelli The co2 sensor values are actually signed integers.

mgiacomelli:
@davidkna You're right.  Tested and indeed it can return negative values depending on calibration (or apparently random glitching in my case).  I pushed a fix to my branch.

mgiacomelli:
Pushed a fix for the strings of missing data that tended to happen if you used a fast update rate.  The issue was that you can't read out the sensor while its performing a measurement, so if the update rate was 5s and you polled every 5s, you could accidentally align with the update period and miss long runs of data.  Tested with both single shot and periodic mode and this seems much more reliable.

mgiacomelli:
@ozonejunkieau Are you still working on this PR?  I was going to fix a few more things related to polling and then try and get it merged into git, hopefully soonish.

PR #8151, opened 2025-01-28
Add virtual get_flags() to GPIOPin and implementation in InternalGPIOPin derivatives

DjordjeMandic:
# What does this implement/fix?

As suggested in https://github.com/esphome/esphome/pull/7752#issuecomment-2481434707

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [ ] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8178, opened 2025-02-02
[esp32_rmt] Set pull-up and open-drain modes based on pin schema

swoboda1337:

The RMT drivers in IDF 5+ now enable pullups by default in both rx and tx, this didnt happen in the legacy RMT drivers. Multiple people have reported rmt code not working in dev, these issues have have been traced to pullups being enabled. Add code to enable / disable pullups based on the pin schema. Pullups will be disabled be default to keep it the same as IDF 4 & Arduino. 

Get open drain mode from pin schema as well instead of using a separate config variable. 

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4634

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


swoboda1337:
This PR depends on https://github.com/esphome/esphome/pull/8151

esphome/components/remote_transmitter/__init__.py
```diff
@@ -74,11 +73,12 @@ async def to_code(config):
                 cg.add(var.set_clock_resolution(config[CONF_CLOCK_RESOLUTION]))
             if CONF_USE_DMA in config:
                 cg.add(var.set_with_dma(config[CONF_USE_DMA]))
-            if CONF_ONE_WIRE in config:
-                cg.add(var.set_one_wire(config[CONF_ONE_WIRE]))
             if CONF_EOT_LEVEL in config:
                 cg.add(var.set_eot_level(config[CONF_EOT_LEVEL]))
-            elif CONF_ONE_WIRE in config and config[CONF_ONE_WIRE]:
+            elif (
+                CONF_OPEN_DRAIN in config[CONF_PIN]
```
jesserockz: This is inside the `CONF_MODE` not directly on the pin

swoboda1337:
It isn't really breaking none of this existed in 2024.12, it hasn't been released yet

esphome/components/remote_transmitter/remote_transmitter_esp32.cpp
```diff
@@ -109,7 +114,7 @@ void RemoteTransmitterComponent::configure_rmt_() {
       this->mark_failed();
       return;
     }
-    this->digital_write(this->one_wire_ || this->inverted_);
+    this->digital_write(this->eot_level_);
```
jesserockz: Why not checking if pin is inverted anymore?

jesserockz (dismissed):

esphome/components/remote_transmitter/remote_transmitter_esp32.cpp
swoboda1337: > Why not checking if pin is inverted anymore?

So eot_level is set inside init.py, it is based on inverted and od mode and can be overridden, this was a minor bug, should have been using eot_level.

swoboda1337: In init.py this->eot_level_ is actually set based on inverted, however it is better to leave it the way it was.

swoboda1337:
Sorry I should have tested the eot_level stuff better, it is correct now. Thanks

esphome/components/remote_transmitter/__init__.py
```diff
@@ -74,13 +74,11 @@ async def to_code(config):
...
                 cg.add(var.set_with_dma(config[CONF_USE_DMA]))
-            if CONF_ONE_WIRE in config:
-                cg.add(var.set_one_wire(config[CONF_ONE_WIRE]))
             if CONF_EOT_LEVEL in config:
                 cg.add(var.set_eot_level(config[CONF_EOT_LEVEL]))
-            elif CONF_ONE_WIRE in config and config[CONF_ONE_WIRE]:
+            elif config[CONF_PIN][CONF_MODE][CONF_OPEN_DRAIN]:
                 cg.add(var.set_eot_level(True))
-            elif CONF_INVERTED in config[CONF_PIN] and config[CONF_PIN][CONF_INVERTED]:
+            elif config[CONF_PIN][CONF_INVERTED]:
```
kbx81: Could this and the lines above be combined to use an `or`?:
```suggestion
            elif config[CONF_PIN][CONF_MODE][CONF_OPEN_DRAIN] or config[CONF_PIN][CONF_INVERTED]:
```

swoboda1337: I could but the formatter tool would change that into an ugly 4 line if, making it longer than before:
```
+            elif (
+                config[CONF_PIN][CONF_MODE][CONF_OPEN_DRAIN]
+                or config[CONF_PIN][CONF_INVERTED]
+            ):
```

kbx81: I think that's fine, I'm used to seeing it do that.

swoboda1337: Ok think this looks the least ugly :)

kbx81 (approved):
WOMMs -- thanks again! 🍻

PR #8181, opened 2025-02-02
[i2s_audio, i2s_audio_microphone, i2s_audio_speaker] Add basic support for new esp-idf 5.x.x i2s driver.

luar123:
This PR adds basic support for the new i2s driver. This driver removes the deprecated warning when compiling with esp-idf 5.x.x and could add additional functionality like independent tx/rx channels (full duplex) on newer esp versions and better DMA. Additionally, I am working on a snapclient component that relies on the new driver. Old and new driver can not be used in parallel.

Changes:
- Most names changed, but I still kept the same config parameters
- new parameter: `use_legacy`, allows to use the old driver even with esp-idf 5. Some components like speaker, mediaplayer and external components still need the legacy driver.
- i2s microphone and speaker: normal external dac is ported to the new driver, internal dac not
- i2s mediaplayer: added final validate to make sure legacy driver is used.

Todo:
- internal ADC/DAC is not supported by the new driver, needs to be implemented by adc/dac drivers or use the legacy driver.
- independent rx/tx channels would allow tx and rx at the same time for esp32 c3 (only one i2s controller) and two tx and two rx channels for esp32-s3. Only MCLK needs to be shared by tx/rx, but is often not needed for both.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4638

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
esp32:
  board: esp32dev
  framework:
    type: esp-idf

i2s_audio:
  - id: a1 
    i2s_lrclk_pin: GPIO17
    i2s_bclk_pin: GPIO16
    use_legacy: false

microphone:
  - platform: i2s_audio
    id: mic
    adc_type: external
    i2s_din_pin: GPIO15
    i2s_audio_id: a1

```

## Checklist:


luar123:
Added documentation PR.
I would suggest to implement the open items in separate PRs. Or maybe I could add pdm mode here, shouldn't be too much work.
I could also make `use_legacy` default to `true`, then this would not be a breaking change. We could change it when pdm mode and speaker are ported (after #8171 is merged).

jhbruhn:
A proper implementation of the duplex mode, which would be enabled by this PR, is greatly needed for te Onju-Voice project to be able to integrate the upcoming media_player component with it. Looking forward on what this PR brings!

luar123:
> A proper implementation of the duplex mode, which would be enabled by this PR, is greatly needed for te Onju-Voice project to be able to integrate the upcoming media_player component with it. Looking forward on what this PR brings!

The duplex mode I discussed here would use different clk and gpio settings for rx and tx, therefore I am not sure if that would help. Full duplex with shared clk signals would be possible also with the old driver, however, the whole esphome implementation would need to be changed.

luar123:
Added rx pdm and speaker external dac support.
For speaker I found two issues:
* what is `stand_max` format? Couldn't find a definition, but could be implemented.
* I couldn't find a replacement for `i2s_write_expand()` so audiostream and speaker need to have same bits per sample.

Hedda:
> A proper implementation of the duplex mode, which would be enabled by this PR, is greatly needed for te Onju-Voice project to be able to integrate the upcoming media_player component with it. Looking forward on what this PR brings!

I understand the [FutureProofHomes Satellite1 hardware ](https://github.com/FutureProofHomes/Satellite1-Hardware)also uses single duplex i2s channel for both microphone and speaker?

* https://github.com/orgs/FutureProofHomes/discussions/6

...so they might have the same issue too without this new i2s driver with full duplex support?

* https://github.com/FutureProofHomes/Satellite1-ESPHome/issues/256

RJPoelstra:
Unfortunately, I can't get this PR to work.
I tried the following yaml:
```
external_components:
  - source: github://pr#8181
    components: [ i2s_audio, microphone ]

i2s_audio:
  - id: i2s_audio_bus
    i2s_lrclk_pin: GPIO33
    i2s_bclk_pin: GPIO19

microphone:
  - platform: i2s_audio
    id: echo_microphone
    i2s_din_pin: GPIO23
    adc_type: external
    pdm: true
```

This is for an Atom Echo, but it seems I don't get audio. The full config works when I target IDF 4.4.8. I no longer get the deprecation warnings, so it looks like this new driver does get used.

luar123:
Thanks for testing. I don't have a pdm microphone so I can't test myself.
Could you try:
1. set `bits_per_sample: 16` in microphone, seems like the default is 32bit and that won't work with pdm.
2. add `use_legacy: true` to i2s_audio. This will use the old driver with idf5 and should show the deprecation warnings.

rpoelstra:
> 1. set `bits_per_sample: 16` in microphone, seems like the default is 32bit and that won't work with pdm.

This doesn't help

> 2. add `use_legacy: true` to i2s_audio. This will use the old driver with idf5 and should show the deprecation warnings.

This indeed makes the warnings return, but doesn't solve the problem. I only can get audio by targeting IDF 4.

luar123:
Is there anything useful in the logs?

> This indeed makes the warnings return, but doesn't solve the problem. I only can get audio by targeting IDF 4.

That's odd. Does it work with idf 5 if you don't use this PR? There shouldn't be any change.

rpoelstra:
I'm sorry, I should have been more clear. Since IDF 5 audio hasn't worked at all. I was hoping that the new drivers from this PR would improve the situation.
The only way I can get audio working is to use IDF 4.

There is nothing in the logs that might explain the non-working audio.

luar123:
ok, that makes sense then.
There are a few more things you could try:
* clk needs to be connected to the i2s_lrclk_pin
* try to set channel to left or stereo
* Try to modify settings in this [section](https://github.com/esphome/esphome/blob/f0014b3666ef69296f92f985e8737e05a5d78cd4/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp#L147), specially `I2S_PDM_DSR_8S` and `I2S_MCLK_MULTIPLE_256`

In general I am a bit unsure about the data format in mono mode. For PDM it is normal but for STD mode TX:
> Besides, for 8-bit and 16-bit mono modes, the real data on the line is swapped. To get the correct data sequence, the writing buffer needs to swap the data every two bytes.

and for RX:
> The receive case is a little bit complicated on ESP32. Firstly, when the data width is 8-bit or 24-bit, the received data will still align with two bytes or four bytes, which means that the valid data are put in the high 8 bits in every two bytes and high 24 bits in every four bytes. For example, the received data will be 0x5A00 when the data on the line is 0x5A in 8-bit width, and 0x0000 5A00 if the data on the line is 0x00 005A. Secondly, for the 8-bit or 16-bit mono case, the data in buffer is swapped every two data, so it may be necessary to manually swap the data back to the correct order.

Couldn't find much information how this was handled by the legacy driver.

rpoelstra:
Hi, I'm pursuing a slightly differ route: I've checked the I2S peripheral register contents from IDF 4 to IDF 5. They do indeed differ, and if I write the working register values into the registers even after the microphone component is running I get a working PDM mic. It turns out the important differences are here:

| Reg name | non-working contents (IDF5) | working contents (IDF4) | differences |
| --- | --- | --- | --- |
I2S_FIFO_CONF_REG |  0x00111820 | 0x00131820 | I2S_RX_FIFO_MOD 1 > 3
I2S_CONF_CHAN_REG | 0x00000000 | 0x00000008 | I2S_RX_CHAN_MOD 0>1
I2S_SAMPLE_RATE_CONF_REG |  0x00410206 | 0x00810206 | I2S_RX_BITS_MOD differs

I2S_CONF_REG also differs (it configures Philips I2S mode), but this seems overruled by the PDM configuration.

I've yet to find out why these bits are so important and how to get them set right using the drivers. I'm now fixing them in this way:
```
                REG_WRITE(0x3FF4F020, 0x00131820); //I2S_FIFO_CONF_REG
                REG_WRITE(0x3FF4F02C ,0x00000008); //I2S_CONF_CHANNEL_REG
                REG_WRITE(0x3FF4F0B0  ,0x00810206); //I2S_SAMPLE_RATE_CONF_REG
```

luar123:
Great, thanks for investigating.

The new driver supports only 16bit and I don't think the old driver had such limitation. Could you set `bits_per_sample: 16` and compare idf4 and idf5? Default is 32bit. I would be interested if idf4 works with 16bit at all and whats the register difference.

rpoelstra:
I will check when I've more time to spare.
Please note that I had the table headers the wrong way around. I've fixed this.

luar123:
I2S_RX_FIFO_MOD 1 > 3: 16 vs 32 bit
I2S_RX_CHAN_MOD 0>1: left/right channel, 0 might be the issue
I2S_RX_BITS_MOD differs: some data field length, could be related to 16 vs 32 bit

luar123:
Fixed a bug related to the channel setting. Please try again.
If it is still not working try to set channel to left.

rmeissn:
May I ask a question about duplex mode:
Based on the [esp-idf docs for i2s](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/i2s.html#full-duplex), full duplex mode is supported in the standard communication mode, that is one of the two supported modes (STD and PDM) of this driver (right?).
The only issue with duplex mode seems to be, that the TX and RX channels need to be allocated at the same time, which this driver won't, as a shared i2s configuration is needed (between e.g. mic and speaker). Currently, the mic and speaker will initiate i2s on their own, through `i2s_new_channel(&chan_cfg, NULL, &this->rx_handle_); or i2s_new_channel(&chan_cfg, &this->tx_handle_, NULL);` asking for one of the channels only ([here](https://github.com/esphome/esphome/pull/8181/files#diff-3f8e2f825cf0fd65e8648cd62f90f4eb145e4c1b514c1b2c9f3260e2df68ec3eR131) and [here](https://github.com/esphome/esphome/pull/8181/files#diff-b96ba3f970e563013595617089c4d066548c86af9c2d72077bface3115f7bc34R594)).

Based on the [esp-idf docs for i2s](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/i2s.html#full-duplex), duplex simply requires to create both channels at the same time through `i2s_new_channel(&chan_cfg, &this->tx_handle_, &this->rx_handle_);` with a shared configuration (shared `clk_cfg` and `bit_width`, while din and dout must be known at that time).

May I ask for a workaround to use duplex mode with the current component implementation, while improving error reporting at the same time?

**Current Situation (esphome stable, haven't tested this PR yet):**
Currently, when e.g. a speaker creates the i2s bus first, it seems to be "silently" overwritten by the microphone creating it afterwards without raising a proper error message that this configuration can't possibly work.

**Workaround:**
My proposal is to add optional `din` and `dout` parameters to the i2s_audio interface - if one is given, the other one is also required, which translates to duplex mode. Example:
```
i2s_audio:
  - id: i2s_in
    i2s_lrclk_pin: GPIO13
    i2s_bclk_pin: GPIO18
    i2s_din_pin: GPIO17
    i2s_dout_pin: GPIO12
```
Whenever a component (mic or speaker) initializes the i2s interface first, the rx_handle and tx_handle are created at the same time using the config of the respective component and i2s_audio din and dout pins, while these information (including the handles) are saved to a global variable.
Whenever another component (mic or speaker) tries to initialize the i2s interface again for the same i2s_id, its i2s config (especially clock and bit_width) are compared to the already created i2s interface (global variable). 
If these match, the respective tx or rx handle is returned without reinitializing the i2s bus -> full duplex mode
If these don't match, an error is raised, e.g. that this request is impossible and destroys the already configured component -> improvement over the current situation
If a different i2s_id is targeted, it is normally initialized.

I image that this shouldn't be too much work (mainly a global variable and a few if/else statements in both, the mic and speaker components).
Unfortunately, I'm no c++ developer and thus can't provide an implemented proposal.

luar123:
I think your proposal would work without too much work, however, there is more to it:
* the i2s_audio locking mechanism needs to be refactored, maybe use lock_rx and lock_tx instead?
* likely this will introduce braking changes
* could be implemented for both new and legacy driver (more work?) or only for the new driver (more duplicated code?) 
* I think there are plans to re-write the microphone component anyway.

To conclude, I like the idea, but would not include it in this PR.

rmeissn:
So just for summary: This PR will not support i2s duplex mode and it needs a separate feature request and pull request for it.

rpoelstra:
I can confirm that these latest commits work for PDM.
A remark: Can you make 16 bits the default for PDM mode? It feels a bit redundant to have to explicitly specify it. I think it's good to generate the error when a user specifies anything else than 16 bits, but leaving the option out should default to a workable value.

rpoelstra:
Also I'm not sure why you want to keep the legacy drivers in. They are deprecated, so they have to be removed in the near future anyway.

kahrendt:
Thanks for working on this, it is a big help! I haven't done a full review yet, but I've been experimenting with the speaker side of this quite a bit and have notes that can simplify the configuration and minimize the number of breaking changes.

With the current component, we configure the bits per sample with what we actually send to the DAC. So if I configure 32 bits per sample, the I2S driver should send 32 bits per sample no matter what. This corresponds to the ``slot_bit_width`` parameter in the new I2S driver. When we decode audio in ESPHome to send to the speaker, it has its own bits per sample corresponding to the actual file we are decoding. This one corresponds to the ``data_bit_width`` parameter in the new driver. So in my view, the new driver should use the existing bits per sample configuration option, and that should set the ``slot_bit_width``. The speaker's current ``AudioStreamInfo``'s bits per sample should set the data bit width in the driver initialization code; e.g., use ``I2S_STD_PHILIPS_SLOT_DEFAULT_CONFIG((i2s_data_bit_width_t) audio_stream_info.get_bits_per_sample(), slot_mode);``.

The channel configuration is similar. If I configure stereo channels in yaml, the driver should always output audio for 2 speakers. This corresponds to the slot mask, whereas the current ``AudioStreamInfo`` should specify the slot mode (which you seem to be doing in your code already). So instead of creating a new configuration parameters for the slot mode and slot mask, just use the current channel parameter to set the slot mask and keep setting the slot mode based on the current ``AudioStreamInfo``.

I haven't looked at the microphone side a whole lot, but I would guess similar simplifications could be made. Though it is a bit trickier, as we currently manually convert 32 bits per sample to 16 by shifting the most significant bits. I want to redo the microphone code in the near future, but in the mean time it could, perhaps, work to just set both the slot width and data width to the current bits per sample setting, and rely on the manual conversion in the existing code to make sure it outputs 16 bits.

Please let me know if have any questions about this. Again, I really appreciate you working on this as it is making some of my stuff a lot easier to build!

luar123:
@kahrendt thanks for the feedback.

>So in my view, the new driver should use the existing bits per sample configuration option, and that should set the slot_bit_width. The speaker's current AudioStreamInfo's bits per sample should set the data bit width in the driver initialization code

Fully agree, I guess this would also solve the missing `i2s_write_expand()`.

>So instead of creating a new configuration parameters for the slot mode and slot mask, just use the current channel parameter to set the slot mask and keep setting the slot mode based on the current AudioStreamInfo.

This is already implemented, no new config parameters. It might be a bit confusing because some of the logic is in `__init__.py` and some in `i2s_audio_speaker.cpp`, with the latter overwriting some settings.

>but in the mean time it could, perhaps, work to just set both the slot width and data width to the current bits per sample setting, and rely on the manual conversion in the existing code to make sure it outputs 16 bits.

I think that is how it is implemented, but need to double check.

Open questions:
* I translated all formats except `stand_max`. Do you know how it is defined?
* ~~I am a bit unsure about the data format in mono mode. For PDM it is normal but for STD mode TX:~~
>~~For 8-bit and 16-bit mono modes, the real data on the line is swapped. To get the correct data sequence, the writing buffer needs to swap the data every two bytes.~~

~~and for RX:~~

>~~The receive case is a little bit complicated on ESP32. Firstly, when the data width is 8-bit or 24-bit, the received data will still align with two bytes or four bytes, which means that the valid data are put in the high 8 bits in every two bytes and high 24 bits in every four bytes. For example, the received data will be 0x5A00 when the data on the line is 0x5A in 8-bit width, and 0x0000 5A00 if the data on the line is 0x00 005A. Secondly, for the 8-bit or 16-bit mono case, the data in buffer is swapped every two data, so it may be necessary to manually swap the data back to the correct order.~~

~~Was this the same with the legacy driver or do we need implement some data swapping?~~

luar123:
@kahrendt I tried to make the suggested changes regarding the bit width, but I have some questions:

speaker:
* Setting `data_bit_width` to `audio_stream_info.get_bits_per_sample()` makes sense.
* Now for `slot_bit_width` we could use the setting from `bits_per_sample` or `bits_per_channel`. I guess `bits_per_sample` is usually used so this would not break anything. However, `bits_per_channel` has the default `auto` parameter which could make more sense in many cases. Maybe the best solution would be if we could specify bit widths that are supported by a DAC.

microphone:
* How about setting `data_bit_width` always to 16 bit (or 8bit if `slot_bit_width=8bit`), then no manual conversion is needed.
* `slot_bit_width` can then use the value from `bits_per_sample` parameter. Or we could also use `bits_per_channel`?

nielsnl68:
When i may add my opinion to this.

To me the *bits_per_channel* option makes much more sense then the *bits_per_sample* option.
Mostly i think because the first is more explainable when using mono vs stereo. I know it would be a radical change. But i think it is worth it.
Also you can always multiple the value internal to set the *bits_per_sample* when needed.

luar123:
Thanks for the feedback, for now I used `bits_per_channel`. This could be changed easily, though.

Also implemented the bit swapping for ESP32 standard mono modes. This is not needed for any other ESP32 variants.
**Important:** All but the original ESP32 expect 8bit/24bit buffers for those bit widths (not 16/32 as ESP32 or with the old driver). I implemented this for 8bit microphone, but not for speaker. Speaker mediaplayer always uses 16 bit, right? This might simplify audio streams with `bit_per_sample != 16/32`.

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -545,6 +588,92 @@ esp_err_t I2SAudioSpeaker::start_i2s_driver_(audio::AudioStreamInfo &audio_strea
...
+  const i2s_event_callbacks_t callbacks = {
+      .on_send_q_ovf = i2s_overflow_cb,
+  };
+
+  i2s_channel_register_event_callback(this->tx_handle_, &callbacks, this);
+
+  /* Before reading data, start the RX channel first */
+  i2s_channel_enable(this->tx_handle_);
+  if (err != ESP_OK) {
+    // ESP_LOGW(TAG, "Error enabling I2S Speaker: %s", esp_err_to_name(err));
```
kahrendt: Remove commented code

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -545,6 +588,92 @@ esp_err_t I2SAudioSpeaker::start_i2s_driver_(audio::AudioStreamInfo &audio_strea
...
+  i2s_std_config_t std_cfg = {
+      .clk_cfg = clk_cfg,
+      .slot_cfg = std_slot_cfg,
+      .gpio_cfg = pin_config,
+  };
+  /* Initialize the channel */
+  err = i2s_channel_init_std_mode(this->tx_handle_, &std_cfg);
+
+  if (err != ESP_OK) {
+    // ESP_LOGW(TAG, "Error initializing I2S channel: %s", esp_err_to_name(err));
```
kahrendt: Remove commented code

esphome/components/i2s_audio/speaker/i2s_audio_speaker.cpp
```diff
@@ -545,6 +588,92 @@ esp_err_t I2SAudioSpeaker::start_i2s_driver_(audio::AudioStreamInfo &audio_strea
...
+      .id = this->parent_->get_port(),
+      .role = this->i2s_role_,
+      .dma_desc_num = DMA_BUFFERS_COUNT,
+      .dma_frame_num = dma_buffer_length,
+      .auto_clear = true,
+  };
+  /* Allocate a new RX channel and get the handle of this channel */
+  esp_err_t err = i2s_new_channel(&chan_cfg, &this->tx_handle_, NULL);
+  if (err != ESP_OK) {
+    // ESP_LOGW(TAG, "Error creating new I2S channel: %s", esp_err_to_name(err));
```
kahrendt: Remove commented code

kahrendt:
> Speaker mediaplayer always uses 16 bit, right? This might simplify audio streams with `bit_per_sample != 16/32`.

Currently it always uses 16 bits, but its possible in the future to support higher bit depths like 24 bits. There are a few blockers before that can happen though (the flac decoder needs some modifications to support it, and the audio mixer component only works with 16 bits per sample)

kahrendt (changes requested):
Very nice work! All the audio C++ code looks good, and I've been testing it without issues on a variety of devices (ATOM Echo, S3 Box Lite, and Voice PE, with the nabu microphone custom component adapted for this). How you implemented things seems fine to me, but I still miss things, especially on the Python side, with how to do things in an ESPHome way, so I'll defer to Jesse or Keith if they have any suggestions.

kahrendt:
Is there anything you still want to add to your PR that's keeping it in draft?

luar123:
Thanks for the review, I removed the comments. If there is anything in the Python part I should change, please let me know. I could also remove the `use_legacy` parameter and logic and deprecate i2s mediaplayer as both microphone and speaker are supported now.

kahrendt (approved):

PR #8187, opened 2025-02-03
[packet_transport] Extract packet encoding functionality

clydebarrow:

The packet encoding routines in the `udp` component have been hoisted out into a dedicated, transport-independent, platform component that can be used by multiple communication mechanisms.

The `packet_transport` component provides all sensor encode/decode, encryption etc. The UDP component just provides the send/receive functionality.

the `udp.sensor` and `udp.binary_sensor` components are now `packet_transport.sensor` and `packet_transport.binary_sensor`.

UDP component also now has a receive trigger and write action.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4643

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
# Example config.yaml

udp:
  listen_address: 239.0.60.53
  addresses: ["239.0.60.53"]

packet_transport:
  platform: udp
  update_interval: 5s
  encryption: "our key goes here"
  rolling_code_enable: true
  ping_pong_enable: true
  binary_sensors:
    - binary_sensor_id1
    - id: binary_sensor_id1
      broadcast_id: other_id
  sensors:
    - sensor_id1
    - id: sensor_id1
      broadcast_id: other_id
  providers:
    - name: some-device-name
      encryption: "their key goes here"

sensor:
  - platform: template
    id: sensor_id1
  - platform: packet_transport
    provider: some-device-name
    id: our_id
    remote_id: some_sensor_id

binary_sensor:
  - platform: packet_transport
    provider: unencrypted-device
    id: other_binary_sensor_id
  - platform: template
    id: binary_sensor_id1

```

## Checklist:


esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -0,0 +1,460 @@
...
+void PacketTransport::process_(std::vector<uint8_t> &data) {
+  auto ping_key_seen = !this->ping_pong_enable_;
+  auto len = data.size();
+  if (len < 8) {
+    ESP_LOGV(TAG, "Bad length %zu", len);
+    return;
+  }
+  char namebuf[256]{};
+  uint8_t byte;
+  auto buf = data.data();
```
clydebarrow:
```suggestion
  auto *buf = data.data();
```

nagyrobi:
Testing...

esphome/components/udp/packet_transport/udp_transport.h
```diff
@@ -0,0 +1,33 @@
...
+#include "../udp_component.h"
+#include "esphome/core/component.h"
+#include "esphome/components/packet_transport/packet_transport.h"
+#include "esphome/components/network/ip_address.h"
+#if defined(USE_SOCKET_IMPL_BSD_SOCKETS) || defined(USE_SOCKET_IMPL_LWIP_SOCKETS)
+#include "esphome/components/socket/socket.h"
+#endif
+#ifdef USE_SOCKET_IMPL_LWIP_TCP
+#include <WiFiUdp.h>
+#endif
```
jesserockz: These are not required right? This part should be abstracted away by the UDPComponent parent.

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -0,0 +1,461 @@
...
+#endif
+#ifdef USE_BINARY_SENSOR
+    if (byte == BINARY_SENSOR_KEY && binary_sensors.count(namebuf) != 0)
+      binary_sensors[namebuf]->publish_state(rdata.u32 != 0);
+#endif
+  }
+}
+
+void PacketTransport::dump_config() {
+  ESP_LOGCONFIG(TAG, "UDP:");
```
mcuelenaere: Probably you meant something like this:

```suggestion
  ESP_LOGCONFIG(TAG, "packet_transport:");
```

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -0,0 +1,462 @@
...
+    return;
+  }
+
+  auto hlen = *buf++;
+  if (hlen > len - 3) {
+    ESP_LOGV(TAG, "Bad hostname length %u > %zu", hlen, len - 3);
+    return;
+  }
+  memcpy(namebuf, buf, hlen);
+  if (strcmp(this->name_, namebuf) == 0) {
```
j9brown, line 376: While it appears name is null terminated because it was fully zeroed when initialized, it’s possible that it also contains embedded nulls that would make strcmp() think it’s looking at a shorter string than was provided.  It seems fairly benign in this context but I’d suggest using a different string comparison function to be rigorous.

Incidentally, it doesn’t look like the names need to be copied out of the packet within this function.  They could possibly be referenced in-situ using std::string_view or similar.  The performance gains would be negligible but it would solve the problem of embedded nulls (because the comparison operators consider the length).

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -0,0 +1,462 @@
...
+        ESP_LOGV(TAG, "Binary sensor key requires at least 3 more bytes");
+        return;
+      }
+      rdata.u32 = *buf++;
+    } else if (byte == SENSOR_KEY) {
+      if (end - buf < 6) {
+        ESP_LOGV(TAG, "Sensor key requires at least 6 more bytes");
+        return;
+      }
+      rdata.u32 = get_uint32(buf);
```
j9brown: Similar to how the encoder has a nice add() method for appending data to the packet, it might be nice to refactor how the decoder pulls data from the packet to make all of the length checking and conversions happen in the same place.

For example, you could make a class that wraps the buffer, current index, and length.  Then it can have some methods that try to read a field and indicate failure if they cannot.

For example:

```
bool read_byte(uint8_t* out);
bool read_float(float* out);
bool read_string(std::string_view* out);
```

Then a bunch of this code can become: `if (!reader.read_thing(&var)) { /*oops*/ }`

esphome/components/udp/__init__.py
```diff
@@ -1,164 +1,162 @@
...
+                    cv.Required(CONF_LISTEN_PORT): cv.port,
+                    cv.Required(CONF_BROADCAST_PORT): cv.port,
+                }
+            ),
+        ),
+        cv.Optional(
+            CONF_LISTEN_ADDRESS, default="255.255.255.255"
+        ): cv.ipv4address_multi_broadcast,
+        cv.Optional(CONF_ADDRESSES, default=["255.255.255.255"]): cv.ensure_list(
+            cv.ipv4address,
```
dwmw2, line 78: We shouldn't be adding Legacy IP only code, surely?

clydebarrow: Well, firstly IPV4 isn't legacy, it's actually mainstream. IPV6 has not taken over the internet, largely because the apparently imminent exhaustion of V4 addresses has been indefinitely deferred. Secondly, this is not being added, it's already there - the UDP component was merged some time ago and is in use by quite a few people who find it very useful.

However, IPV6 support would be nice to have - no argument there. To make that really usable we'd also need a DNS resolver (because who wants to type in IPV6 addresses?) That doesn't exist in ESPHome as a component right now, and when it does, it's going to have an asynchronous API, so there would be some changes needed in the UDP component to deal with that. Again, that would be nice to have, but it's not there yet. And then there's the testing required, it takes time and effort.

So there's quite a bit to do to make including IPV6 feasible in this context; now the good news is that as an OSS project anybody can contribute that work; the bad news is that it requires someone who has the motivation to do so, which typically means someone who has all of several attributes:

1. The need for that feature
2. The skills to implement it
3. The time to do so.

So "we" can only add what is possible right now. If you have the attributes listed above (and clearly you have a need, so that's a good start) and can build on this to add IPV6 support, that would be absolutely terrific and very welcome. 

It's best however to make changes incrementally, and the ESPHome admins are very clear that small PRs get merged more quickly, and indeed will often request large ones to be broken up, so if IPV6 is going to be added, it will need to be in a follow-up PR, not this one. No reason you can't start on it now though.

esphome/components/packet_transport/packet_transport.cpp
clydebarrow: If the string contains a null, we don't want to match it since that would be some kind of error. The name _does_ need to be copied since it is used later in a context that requires it to be a null-terminated string. And `std::string_view` is not available on all the target platforms since we can only target C++11.

esphome/components/packet_transport/packet_transport.cpp
clydebarrow: I'll have a look at that.

esphome/components/udp/__init__.py
dwmw2: No, we absolutely don't need DNS as a dependency of IPv6. If we can hard-code Legacy IP addresses we can hard-code IPv6 addresses too.

DNS is nice to have, sure. And works fine for MQTT. But certainly isn't a hard dependency for basic UDP connectivity. That's just a silly thing to say.

clydebarrow: In any case I'm not going to delay the review and merge of this PR by adding new functionality. Happy to fix bugs and address code quality, but IPV6 can wait.

dala318:
In lack of linked feature_request I'm commenting in this PR instead for a question and request of addition.
At least with use of ping_pong between nodes, would it be possible to add a `connected` binary_sensor?
Would be nice to see on a provider basis if connection is alive.

EDIT: I've started to play around and drafted an addition for this in https://github.com/dala318/esphome/pull/7

clydebarrow:

```yaml
external_components:
  - source: github://pr#8187
    components: [packet_transport, udp]
    refresh: 1h
```

---
(Added by the PR bot)

wildekek:
This is such an amazing feature. This has the potential to standardize unstructured serial data like CANBus/RS485/UART and also Wireless LoRa encoding. This is the biggest adoption barrier for these protocols and this can solve it. Brilliant.

jesserockz (approved):

PR #8192, opened 2025-02-03
added deep sleep check for partial updates to inkplate6

karlg100:
partial updates now work by overriding the block if we are waking up from a deep sleep.

this simple check will identify when wakeup is from a deep sleep, and using the block_partial bool, allow the display to resume where it left off. 

Albeit, the user will have to configure the default page to be no updates, and force an update with the desired page after all objects are synced from sources and trigger a page update.

Note all the counters an all state will still reset on wakeup, so the user will have to manage at a lower level to force a full refresh when desired, perhaps using home assistant to decide when this should occur.

no direct issue opened, but similar to the feature request https://github.com/esphome/feature-requests/issues/1555

allow deep sleep wakeup to avoid full display refresh, if configured for partial updates

## Types of changes

- [X] Code quality improvements to existing code or addition of tests
- [ ] Other

- fixes <link to issue>

https://github.com/esphome/feature-requests/issues/3043

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

deep_sleep:
  run_duration: 1min
  id: deep_sleep_1
  sleep_duration: 10s

display:
  - platform: inkplate6
    id: inkplate_display
    model: inkplate_10
    greyscale: False
    partial_updating: True
    auto_clear_enabled: False
    update_interval: never
    #update_interval: 1min
    # must manually trigger full updates
    full_update_every: 0
    #full_update_every: 480

    ckv_pin: 32
    sph_pin: 33
    gmod_pin:
      pca6416a: pca6416a_hub
      number: 1
    gpio0_enable_pin:
      pca6416a: pca6416a_hub
      number: 8
    oe_pin:
      pca6416a: pca6416a_hub
      number: 0
    spv_pin:
      pca6416a: pca6416a_hub
      number: 2
    powerup_pin:
      pca6416a: pca6416a_hub
      number: 4
    wakeup_pin:
      pca6416a: pca6416a_hub
      number: 3
    vcom_pin:
      pca6416a: pca6416a_hub
      number: 5
    pages:
      - id: page1
        lambda: |-
        // your code here
```

## Checklist:

### DIFF
--- a/esphome/components/inkplate6/inkplate.cpp
+++ b/esphome/components/inkplate6/inkplate.cpp
@@ -65,6 +65,13 @@ void Inkplate6::initialize_() {
   if (buffer_size == 0)
     return;
 
+  // check if we came out of deep sleep
+  int resetInfo = esp_sleep_get_wakeup_cause();
+  if (resetInfo >= 1) {
+    ESP_LOGD(TAG, "Woke up from deep sleep, disabling full refresh by turning off partial update blocking by default");
+    this->block_partial_ = false;
+  }
+
   if (this->partial_buffer_ != nullptr)
     allocator.deallocate(this->partial_buffer_, buffer_size);
   if (this->partial_buffer_2_ != nullptr)


GraffJosh:
Watching this issue, as it would be very useful to me presently. 

The specific configurations you've outlined here aren't so clear though, would it make more sense to have a native parameterization for a deep sleep epaper? Something that does a partial refresh on wake and then follows the parameters for deep sleep?

PR #8197, opened 2025-02-04
[scheduler] Properly handle millis() overflow

### LINKED ISSUE #5375: Secheduler "stops" on second "Incrementing scheduler major" - after ~100d
### The problem

As soon my devices reached ~99d10h uptime:
* they can't go over - continue reporting same uptime for an hours;
* they start to work strange - stop reporting measurements, stop react on buttons...

image
image

Unfortunately no DEBUG configured on devices side.

image

2023.9.3

Home Assistant Add-on

2024.1.3

ESP8266

### Board

sonoff 4chror3/dualr3/..., d1 mini

_No response_

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

clydebarrow:

Some users have reported a failure to update sensors when the device uptime hits around 100 days (second overflow of a 32 bit count in milliseconds.)

The scheduler does have code to manage this overflow, and it works most of the time. In the particular case tested the factor that triggered a failure was setting the `flash_write_interval` to 0s, presumably in the belief that this disabled the flash sync. In fact it doesn't, it does cause the sync function to be called continually, since that literally sets the interval to 0s.

Exactly why this triggers the failure is unclear, but the effect is that no other scheduled callbacks ever happen once the millisecond time has overflowed twice. The code handling this is rather opaque, since it evidently was written at a time when `uint64_t` was either unavailable or not preferred, and consequently it's quite hard to check for correctnesss by examination. Testing confirms the failure in the above scenario however.

Consequently I have refactored the scheduler code to remove several pieces of convoluted code. The result should actually be more efficient than the original code since it calculates the next execution time for a task once, rather than each time the scheduler runs.

Testing of this confirms correct operation even after multiple `millis()` overflows.

In addition the scheduler debug code has been revised to add additional reporting, and allowing that debug code to be enabled with a yaml config option.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/5375

- esphome/esphome-docs#4641

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

esphome:
  # enable scheduler debug logging 
  debug_scheduler: true
```

## Checklist:


tests/components/esphome/common.yaml
```diff
@@ -0,0 +1,20 @@
...
+  platformio_options:
+    board_build.flash_mode: dio
+  area: testing
+  on_boot:
+    logger.log: on_boot
+  on_shutdown:
+    logger.log: on_shutdown
+  on_loop:
+    logger.log: on_loop
+  compile_process_limit: 10
```
clydebarrow:
```suggestion
  compile_process_limit: 1
```

esphome/core/scheduler.cpp
```diff
@@ -254,13 +250,7 @@ void HOT Scheduler::call() {
...
 
       if (item->type == SchedulerItem::INTERVAL) {
-        if (item->interval != 0) {
-          const uint32_t before = item->last_execution;
-          const uint32_t amount = (now - item->last_execution) / item->interval;
-          item->last_execution += amount * item->interval;
-          if (item->last_execution < before)
-            item->last_execution_major++;
-        }
+        item->next_execution_ = now + item->interval;
```
jesserockz, line 253: The reason this code was how it was is to get an item back onto schedule if it was late.
interval: 100ms
- Exec 1: 100ms since boot (not exactly 100ms as the scheduler gives random offset when setting up to each item)
- (Something here took too long and held loop up for 150ms)
- Exec 2: 250ms since boot
- Exec 3: 300ms since boot

With your change, it would be:
interval: 100ms
- Exec 1: 100ms since boot
- (Something here took too long and held loop up for 150ms)
- Exec 2: 250ms since boot
- Exec 3: 350ms since boot

clydebarrow: Yes, I was aware of that. The problem with the current implementation is that if the unexpected delay is a multiple of of an interval, e.g. say there is a 10ms interval, and there is a loop block that lasts for 1000ms, the event will be rescheduled 100 times in immediate succession, which is not a good outcome IMO since it doesn't match the expected schedule anyway and will block everything else in the meantime, so the problem cascades.

There is no perfect solution, because in your example the execution at 250ms is the earliest possible, so the one expected at 200ms is forever lost. IMO it's better to get over the unexpected delay and move on.

There's no particular need to re-align the interval to 100ms boundaries since boot, because that was never guaranteed - indeed it is guaranteed that won't happen because of the random offset that is applied to an interval (but not a timeout.)

ssieb: It's not about aligning with boot, but aligning with the initial execution.  I agree with Jesse that it should continue at a consistent interval.  Skipping any overdue ones is expected, so don't make up the missed ones.

clydebarrow: So are you suggesting not executing the overdue call as soon as possible but deferring it to a multiple of the original interval? The problem with that will be that if further delays occur it might never be executed.

Or are you suggesting executing it now anyway, but scheduling the next one at a the next available interval after now that it should have been scheduled at? That would be doable, but it would likely result in one over-long interval (which is unavoidable), and one short one. Since the only real guarantee offered by `set_interval()` is that the callback will not be called at intervals *less* than requested, I'm not sure that's going to be helpful.

Do you have any use cases in mind where it would make a difference?

clydebarrow: I did a quick review of all the places `set_interval()` is used in ESPHome, and nothing jumped out as being sensitive to the alignment of calls. There are some that would be sensitive to delays, like the `throttle_average` filter, but nothing can be done about that since the problem of loop blocking can't be solved here.

Also on further reflection to try to re-align the intervals (to something that was random in the first place) would result in behaviour that is hard to predict. As is it offers the guarantee that the interval will not be called more often than requested. No other guarantee is possible, and trying (but sometimes failing) to meet some other objective at the expense of not guaranteeing anything would just muddy the waters I think.

DAVe3283: The new implementation should not cause the interval to skew except in cases where the scheduled execution was delayed. While this could cause a slightly longer than asked for interval, the simplicity and reliability of the implementation is worth it IMO.

If the desire is to try and keep aligned to the original schedule, there are only 2 safe ways to do it IMO. Using the example above of a 100ms interval:
- Exec 1: 100ms since boot
- (Something here took too long and held loop up for 150ms)
- Exec 2: 250ms since boot

To keep things aligned, we can really only schedule execution 3 a couple ways:
1. 300ms since boot - execute in 50ms instead of 100ms, but gets us back on track as quickly as possible
2. 400ms since boot - add another delay to get on track, but maintain the 100ms minimum guarantee

I can see the utility of both options above. Option 1 would make things like moving average filters more accurate than what is proposed in this PR, but could cause failures when dealing with external devices that only have data ready to read X amount of time after the last read (many ADCs for example).
Option 2 would ensure external devices are always ready and keep things aligned, but at the expense of the average interval actually being much longer than expected. And the PR as written provides the same guarantee for external devices without the extra delay.

I don't know what the impacts to the rest of the codebase would be, but for the integrations I maintain, I would prefer either the PR as written or option 1 above. Intentionally adding _more_ delay (option 2) seems counterproductive.

ssieb: There's no suggestion to add more delay.  The alternative suggestion would be to exec at 250 and then 300, 400, etc.  But to be clear also, that first 100 is not correct.  The first one will be at a random start time and then steps of 100 after that.
Another part of this issue is that if you count from the current time, there will always be small delays every time you calculate, so it will continuously drift.

clydebarrow: The API for `set_interval()` has this comment:

>    * IMPORTANT: Do not rely on this having correct timing. This is only called from
>    * loop() and therefore can be significantly delay. If you need exact timing please
>    * use hardware timers.

IOW there never have been any guarantees. The implementation as proposed is simple, readily verifiable, and should perform as expected unless something breaches the `loop()` contract by excessive delay. Adding complexity would not provide any other guarantees so I'm going to declare this conversation resolved.

PS: further comments added to clarify the `set_interval()` semantics.

kbx81 (approved):

septillion-git:
Isn't this THE classic example of why never to do ``previousMillis + interval < millis()`` but always do ``millis() - previousMillis > interval``? So instead of setting ``item->next_execution_ = now + offset;`` we should set ``item->last_execution_ = now;`` and change the check to break to ``if (now  - item->last_execution_ < item->interval)``

That way you are immune to overflows as long as you keep millis related variables as unsigned.

clydebarrow:
> `if (item->last_execution_ - now < item->interval)`

Working in 16 bit for example,

* interval == 0x4000
* last_execution == 0xFFFF
* now == 1

So actual elapsed time is 2, but last_execution - now evaluates to FFFE, which is larger than interval, so there would be a spurious additional execution.

septillion-git:
@clydebarrow Oops, now I did it wrong myself. Swap last_execution and now of course. Other way around makes no sense :sweat_smile:

1 - 0xFFFF = 2

But instead of trying to calculate the new moment, this will always work. No matter an overflow. As long as the variables are unsigned and larger than interval.

But in the current situation if we do a callback @ 0xF000, this will set next_execution_ = 0xF000 + 0x4000  = 0x3000

So next time we do a ``call()`` to check if we need to do the callback it could be that now = 0xFF00
And thus [this](https://github.com/clydebarrow/esphome/blob/a87b4beba30b7678097daa719de4401df383e356/esphome/core/scheduler.cpp#L208-L211)
```c++
      if (item->next_execution_ > now) {
        // Not reached timeout yet, done for this call
        break;
      }
````
Will not break and thus do a callback even though just 0xFF00 - 0xF000 = 0x0F00 time passed...

As if you would do ``now  - item->last_execution_ < item->interval`` 
0xFF00 - 0xF000 < 0x4000 => true and thus does break.

DAVe3283:
@septillion-git while I agree the new code does appear to have a bug when `this->millis_()` wraps around, it was changed to a `uint64_t`, which will wrap every 2^64-1 ms, or 584,942,417 years. So not actually a problem in practice, as far as I can tell.

septillion-git:
@DAVe3283 So we rather waste memory than to fix a bug?

I would not dare to call that proper handling.... It is after all [Arduino 101](https://docs.arduino.cc/built-in-examples/digital/BlinkWithoutDelay/)

I'm willing to change the code to really fix it. But I don't have a dev setup to test it...

DAVe3283:
> @DAVe3283 So we rather waste memory than to fix a bug?
> 
> I would not dare to call that proper handling.... It is after all [Arduino 101](https://docs.arduino.cc/built-in-examples/digital/BlinkWithoutDelay/)
> 
> I'm willing to change the code to really fix it. But I don't have a dev setup to test it...

I don't believe this new method actually uses any more memory. It used to be two separate uint32_t values stored per instance, but it is now rolled up to a single uint64_t with a function. Or maybe I'm misreading it, this isn't a part of the codebase I'm familiar with.

I'm just happy to have the previous very real bug at 100 days resolved. I do agree, making the scheduler theoretically perfect would be nice. But I don't have a test framework like Clyde has to verify the change works as expected by speeding up execution or however he did it.

septillion-git:
@DAVe3283 True for the fix! But it's a bit shameful it was present to start with...

And yeah, does use more memory. 
- 2 bytes for the major
- 4 bytes to store last millis
- 2 extra bytes to store the moment

Know it's not major amount. But it's really such a basic bug with such a simple solution. It's a bit comical this required this much to "fix"...

Let me see if a have some time to rewrite it :)

PR #8207, opened 2025-02-05
[sdl] Implement binary sensors from keystrokes

bdm310:

Add an SDL platform to binary sensor to enable generic keyboard inputs. Without this, encoder and keypad based        GUIs are difficult to evaluate on the host platform.

## Types of changes

- [ ] Other

- esphome/esphome-docs#4644

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

```yaml
esphome:
  name: sdl

host:

logger:

display:
  - platform: sdl
    dimensions:
      width: 320
      height: 320
    auto_clear_enabled: false 

binary_sensor:
  - platform: sdl
    id: key_up
    key: SDLK_a
  - platform: sdl
    id: key_down
    key: SDLK_d
  - platform: sdl
    id: key_enter
    key: SDLK_s

lvgl:
  encoders:
    enter_button: key_enter
    sensor:
      left_button: key_up
      right_button: key_down

```

## Checklist:


esphome/components/sdl/__init__.py
```diff
@@ -1 +1,237 @@
 CODEOWNERS = ["@clydebarrow"]
+
+SDL_KEYMAP = {
```
clydebarrow: Since this is only used in the binary_sensor platform, no real need for it to be here - move it to `binary_sensor/__init__.py`

esphome/components/sdl/binary_sensor/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+SdlBinarySensor = sdl_ns.class_(
+    "SdlBinarySensor", binary_sensor.BinarySensor, cg.Component
+)
+
+CONFIG_SCHEMA = (
+    binary_sensor.binary_sensor_schema(SdlBinarySensor)
+    .extend(
+        {
+            cv.Required(CONF_KEY): cv.enum(SDL_KEYMAP),
+            cv.GenerateID(CONF_SDL_ID): cv.use_id(Sdl),
```
clydebarrow: Need to document this option.

esphome/components/sdl/binary_sensor/sdl_binary_sensor.h
```diff
@@ -0,0 +1,22 @@
...
+#ifdef USE_HOST
+#include "../sdl_esphome.h"
+#include "esphome/components/binary_sensor/binary_sensor.h"
+
+namespace esphome {
+namespace sdl {
+
+class SdlBinarySensor : public binary_sensor::BinarySensor, public Component, public Parented<Sdl> {
+ public:
+  void loop() override { this->publish_state(this->parent_->key_state[this->key_]); }
```
clydebarrow: This isn't the right approach. Instead, add to the parent component a map of key listeners (indexed by key) which will be called when a key event is detected, publish the appropriate state when the listener is called. CallBackManager is a utility class that will help with this - allows multiple listeners for the same thing.

You can achieve this without even extending `BinarySensor` - just have the `to_code` function create a plain `BinarySensor`, then create a lambda calling its `publish_state` method, add that lambda as a listener for the corresponding key. So `sdl_binary_sensor.h` can be removed.

esphome/components/sdl/sdl_esphome.cpp
```diff
@@ -87,6 +87,14 @@ void Sdl::loop() {
         }
         break;
 
+      case SDL_KEYDOWN:
+        this->key_state[e.key.keysym.sym] = true;
```
clydebarrow: This would need range-checking. But if you do what I suggested it becomes moot.

clydebarrow (changes requested):

esphome/components/sdl/__init__.py
bdm310: I placed this here as it is a higher level SDL concept. Doesn't matter to me either way, will move.

esphome/components/sdl/binary_sensor/__init__.py
bdm310: Documentation should be in the linked PR?

esphome/components/sdl/binary_sensor/sdl_binary_sensor.h
bdm310: I agree that sounds like a much cleaner approach. I don't know that I can commit the hours to understanding the overall system enough here to make that happen. Is this PR dead without the architecture change?

bdm310: I had not considered multiple displays yet, which I think means a deeper look makes sense. Will submit changes when I have something.

clydebarrow: Don't see any issues with multiple displays. If you like I can rework it - won't take me long. The idea is great BTW.

bdm310: If you're willing to, I would appreciate it. I hate to push work on to others but it has been a couple hours and I've not made much progress in figuring out the details of "create a lambda calling publish_state". Conceptually I'm completely with you and I should have looked at the LVGL widget sensors for inspiration to start with.

Even if you just want to get it most of the way done, I'm happy to test and wrap up the details.

clydebarrow (approved):

PR #8209, opened 2025-02-05
[seeed_mr60bha2] Improve stability

limengdu:

## Types of changes

- [ ] Other

- fixes <https://github.com/limengdu/MR60BHA2_ESPHome_external_components/pull/7>

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/seeed_mr60bha2/seeed_mr60bha2.cpp
+++ b/esphome/components/seeed_mr60bha2/seeed_mr60bha2.cpp
@@ -129,7 +129,7 @@ bool MR60BHA2Component::validate_message_() {
 }
 
 void MR60BHA2Component::process_frame_(uint16_t frame_id, uint16_t frame_type, const uint8_t *data, size_t length) {
-  if (this->has_target_binary_sensor_ != nullptr && this->has_target_binary_sensor_->state == false &&
+  if (this->has_target_binary_sensor_ != nullptr && !this->has_target_binary_sensor_->state &&
       frame_type != PEOPLE_EXIST_TYPE_BUFFER) {
     // Do not process other frames while people exists sensor is still false
     return;
--- a/esphome/components/seeed_mr60bha2/seeed_mr60bha2.cpp
+++ b/esphome/components/seeed_mr60bha2/seeed_mr60bha2.cpp
@@ -129,7 +129,8 @@ bool MR60BHA2Component::validate_message_() {
 }
 
 void MR60BHA2Component::process_frame_(uint16_t frame_id, uint16_t frame_type, const uint8_t *data, size_t length) {
-  if (this->has_target_binary_sensor_ != nullptr && this->has_target_binary_sensor_->state == false && frame_type != PEOPLE_EXIST_TYPE_BUFFER) {
+  if (this->has_target_binary_sensor_ != nullptr && this->has_target_binary_sensor_->state == false &&
+      frame_type != PEOPLE_EXIST_TYPE_BUFFER) {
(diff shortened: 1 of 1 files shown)


ssieb:
What is the purpose of these changes?  Some of them have no effect and some change the normal functioning of a sensor.

limengdu:
Reducing duplicate data refreshes

ssieb:
That might be an ok change, but that is the normal way sensors work.  If you don't like the updates, use a filter.
What about all the changes in `validate_message`?

limengdu:
This is a user-submitted revision, and the logic hasn't changed much.

ssieb:
I prefer the current logic and I don't see any reason to change it.

bobvandevijver:
@ssieb I'm the author of these changes. The current implementation is unusable and broken, this PR solves it.

This fixes a couple of things:

- Optimises memory usage: only allocate memory when actually used, and reuse the same memory when possible
- No longer crash when the binary sensor is enabled, but not all other sensors are
- No longer flood the ESP with unnecessary state changes when the state did not change at all
- When the binary sensor is enable and its state is false, do not handle any other frames until it has been set to true. This fixes my problem where the person count kept toggling between 0 and something else, from the log in [this comment](https://github.com/limengdu/MR60BHA2_ESPHome_external_components/issues/6#issuecomment-2614412318). While this is an issue with the mmWave module as it sends conflicting data, solving it here is easiest.

jesserockz:
> @ssieb I'm the author of these changes. The current implementation is unusable and broken, this PR solves it.
> 
> This fixes a couple of things:
> 
> * Optimises memory usage: only allocate memory when actually used, and reuse the same memory when possible

1 byte (`new_byte`) is not a big saving to worry about IMO. Everything in `validate_message_` is gonna be freed at the end so its not going to cause any issues with anything else.

> * No longer crash when the binary sensor is enabled, but not all other sensors are

This is fine, but it is mashed in with a != 0 check for the sensor which doesn't need to exist. (See next answer)

> * No longer flood the ESP with unnecessary state changes when the state did not change at all

ESPHome core code should always relay data that comes from an external sensor as is. We have `filters` to allow users to deduplicate the data and you are welcome to add them to the examples on the documentation for this sensor. The problem with doing deduplication in the code means that a user cannot see if the sensor is working correctly or not if the data is not changing.

> * When the binary sensor is enable and its state is false, do not handle any other frames until it has been set to true. This fixes my problem where the person count kept toggling between 0 and something else, from the log in [this comment](https://github.com/limengdu/MR60BHA2_ESPHome_external_components/issues/6#issuecomment-2614412318). While this is an issue with the mmWave module as it sends conflicting data, solving it here is easiest.

We should not read the state from the binary sensor like this. If a user doesn't have it configured, this feature wont work. Keep a protected boolean for this value if it is needed across loops

jesserockz (changes requested):
See above

Rudd-O:
I would like this to be merged.

Furthermore, the current implementation is broken in another way that should be fixed as per

https://github.com/limengdu/MR60BHA2_ESPHome_external_components/commit/f875eba14198a61f13d39182ba6c2558b7401511

Thanks.

koszta:
I would not recommend merging this until sending out all the sensor data is restored, see:
https://github.com/limengdu/MR60BHA2_ESPHome_external_components/issues/12

PR #8212, opened 2025-02-05
[online_image] Improve error handling

guillempages:
JPEG images need to resize the download buffer to the full size before start downloading. If the resize fails (e.g. because of not enough memory), abort the downloading, instead of trying to download into a too small buffer.
This prevents crashing in such cases.

Also improved the same situation in case of PNG images (i.e. bail out earlier)

## Types of changes

- [ ] Other

N/A

N/A

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -152,7 +152,6 @@ void OnlineImage::update() {
   }
   auto prepare_result = this->decoder_->prepare(total_size);
   if (prepare_result < 0) {
-    ESP_LOGE(TAG, "Error while preparing image download: %d", prepare_result);
     this->end_connection_();
     this->download_error_callback_.call();
     return;
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -150,7 +150,13 @@ void OnlineImage::update() {
     this->download_error_callback_.call();
     return;
   }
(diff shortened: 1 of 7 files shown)


esphome/components/online_image/online_image.cpp
```diff
@@ -150,7 +150,13 @@ void OnlineImage::update() {
     this->download_error_callback_.call();
     return;
   }
-  this->decoder_->prepare(total_size);
+  auto prepare_result = this->decoder_->prepare(total_size);
+  if (prepare_result < 0) {
+    ESP_LOGE(TAG, "Error while preparing image download: %d", prepare_result);
```
clydebarrow: Is this needed since the `prepare()` method will log an error anyway?

guillempages: I had it as a generic "catch-all" in case the decoders do not log the error, but I can remove it.

guillempages:
Looks like I still have not caught all the crashes yet; I'm still trying to debug an additional crash. Was working yesterday, it's crashing today 🤔

guillempages:
I think I got the bug. Working stable now on my setup.

jesserockz (approved):

PR #8216, opened 2025-02-06
[online_image] Set Accept header

candrews:

The HTTP Accept request header indicates which content types, expressed as MIME types, the sender is able to understand. In requests, the server uses content negotiation to select one of the proposals.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

## Types of changes

- [ ] Other

n/a

n/a

## Test Environment

- [ ] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Any example would work, such as the one shown in https://esphome.io/components/online_image.html

```yaml
online_image:
  - url: "https://example.com/example.png"
    format: png
    id: my_online_image
```

## Checklist:

### DIFF
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -100,7 +100,27 @@ void OnlineImage::update() {
   }
   ESP_LOGI(TAG, "Updating image %s", this->url_.c_str());
 
-  this->downloader_ = this->parent_->get(this->url_);
+  std::list<http_request::Header> headers = {};
+
+  http_request::Header accept_header;
+  accept_header.name = "Accept";
+  switch (this->format_) {
+    case ImageFormat::BMP:
+      accept_header.value = "image/bmp;q=0.9,*/*;q=0.8";
+      break;
+    case ImageFormat::JPEG:
+      accept_header.value = "image/jpeg;q=0.9,*/*;q=0.8";
+      break;
+    case ImageFormat::PNG:
+      accept_header.value = "image/png;q=0.9,*/*;q=0.8";
+      break;
+    default:
+      accept_header.value = "*/*";
+  }
+
+  headers.push_back(accept_header);
+
+  this->downloader_ = this->parent_->get(this->url_, headers);
 
   if (this->downloader_ == nullptr) {
     ESP_LOGE(TAG, "Download failed.");
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -105,15 +105,21 @@ void OnlineImage::update() {
   http_request::Header accept_header;
   accept_header.name = "Accept";
   switch (this->format_) {
+#ifdef USE_ONLINE_IMAGE_BMP_SUPPORT
     case ImageFormat::BMP:
       accept_header.value = "image/bmp;q=0.9,*/*;q=0.8";
       break;
+#endif  // ONLINE_IMAGE_BMP_SUPPORT
+#ifdef USE_ONLINE_IMAGE_JPEG_SUPPORT
     case ImageFormat::JPEG:
       accept_header.value = "image/jpeg;q=0.9,*/*;q=0.8";
       break;
+#endif  // USE_ONLINE_IMAGE_JPEG_SUPPORT
+#ifdef USE_ONLINE_IMAGE_PNG_SUPPORT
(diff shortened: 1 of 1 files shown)


esphome/components/online_image/online_image.cpp
```diff
@@ -100,7 +100,27 @@ void OnlineImage::update() {
   }
   ESP_LOGI(TAG, "Updating image %s", this->url_.c_str());
 
-  this->downloader_ = this->parent_->get(this->url_);
+  std::list<http_request::Header> headers = {};
+
+  http_request::Header accept_header;
+  accept_header.name = "Accept";
+  switch (this->format_) {
```
guillempages, line 108: Why not guard the different formats with the "#ifdef ONLINE_IMAGE_USE_XYZ" defines, to not compile them in, if the format is not used at all? Not really using a big amount of space anyway, but for consistency with the rest of the format-dependent code.

esphome/components/online_image/online_image.cpp
```diff
@@ -100,7 +100,27 @@ void OnlineImage::update() {
...
+      accept_header.value = "image/bmp;q=0.9,*/*;q=0.8";
+      break;
+    case ImageFormat::JPEG:
+      accept_header.value = "image/jpeg;q=0.9,*/*;q=0.8";
+      break;
+    case ImageFormat::PNG:
+      accept_header.value = "image/png;q=0.9,*/*;q=0.8";
+      break;
+    default:
+      accept_header.value = "*/*";
```
guillempages: Maybe set "image/\*" as a first type, and only "\*/\*" as a fallback?

guillempages (commented):
Have you tested this with the arduino framework as well? AFAIK, the http_request implementation is very different for both esp_idf and arduino.

esphome/components/online_image/online_image.cpp
```diff
@@ -100,7 +100,27 @@ void OnlineImage::update() {
...
   ESP_LOGI(TAG, "Updating image %s", this->url_.c_str());
 
-  this->downloader_ = this->parent_->get(this->url_);
+  std::list<http_request::Header> headers = {};
+
+  http_request::Header accept_header;
+  accept_header.name = "Accept";
+  switch (this->format_) {
+    case ImageFormat::BMP:
+      accept_header.value = "image/bmp;q=0.9,*/*;q=0.8";
```
guillempages: Why explicitly limit the "main" format? Why not just
```suggestion
      accept_header.value = "image/bmp,*/*;q=0.8";
```
?

esphome/components/online_image/online_image.cpp
candrews: Makes sense, I've updated the MR.

guillempages:
Is there a use case that you have encountered for this? If the accept is still falling back to accepting everything anyway, and the user already needs to know which format the downloaded image is in, what's the point of adding the accept header at all? In which way does this help? (Just trying to understand; not trying to reject the change)

esphome/components/online_image/online_image.cpp
candrews: I like it! I've updated this MR accordingly.

esphome/components/online_image/online_image.cpp
candrews: I like it! I've updated this MR accordingly.

candrews:
> Is there a use case that you have encountered for this? If the accept is still falling back to accepting everything anyway, and the user already needs to know which format the downloaded image is in, what's the point of adding the accept header at all? In which way does this help? (Just trying to understand; not trying to reject the change)

Servers use the accept header to respond with the correctly formatted image. For example, I'm generating webp or png based on the `Accept` header in my image generating service.

By having `online_image` send the appropriate `Accept` header, when `online_image` gains support for a new format, the server will automatically return the image in that new format.

guillempages (approved):

guillempages:
Ok, I wouldn't have expected that for the esphome devices this is actually a use case. I would have expected people to just download existing images, or have a server explicitly generating the image for ESPHome, so there would be no need for the accept headers.

Nevertheless, this *is* an improvement so I have no objection to getting this in.

jesserockz (approved):

PR #8224, opened 2025-02-10
[http_request] Ability to get response headers

candrews:

[http_request] Ability to get response headers

This is a breaking change because `headers` is renamed to `request_headers`

## Types of changes

- [ ] Other

- esphome/esphome-docs#4654

## Test Environment

- [X] ESP32 Arduino
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
external_components:
  - source: github://pr#8224
    components: [ http_request ]
    refresh: 0s

http_request:
  verify_ssl: false

time:
  - platform: homeassistant
    id: homeassistant_time
    timezone: America/New_York
    on_time: 
      - seconds: '*'
        then:
          - logger.log: "About to perform request"
          - http_request.get:
              url: https://www.integralblue.com/logo.png
              collect_headers:
                - etag
                - last-modified
              on_response:
                then:
                  - logger.log:
                      format: 'Response status: %d, Duration: %u ms, etag: %s, last-modified: %s'
                      args:
                        - response->status_code
                        - response->duration_ms
                        - response->get_response_header("etag").c_str()
                        - response->get_response_header("last-modified").c_str()
              on_error:
                  then:
                  - logger.log:
                      format: 'Error!'
          - logger.log: "Request completed"
```

## Checklist:


candrews:
I'm still working on testing this.

I'd love to get feedback from the developer(s) sooner rather than later so I can start addressing TODOs.

Thank you so much for your attention and participation!

esphome/components/http_request/__init__.py
```diff
@@ -269,6 +271,12 @@ async def http_request_action_to_code(config, action_id, template_arg, args):
         )
         cg.add(var.add_header(key, template_))
 
+    for key in config.get(CONF_COLLECT_HEADER_NAMES, []):
```
clydebarrow: Is there much to be gained by filtering the response headers? Why not just collect them all?

esphome/components/http_request/__init__.py
```diff
@@ -269,6 +271,12 @@ async def http_request_action_to_code(config, action_id, template_arg, args):
...
+        template_ = await cg.templatable(
+            config[CONF_COLLECT_HEADER_NAMES][key], args, cg.const_char_ptr
```
clydebarrow:
```suggestion
            key, args, cg.const_char_ptr
```

esphome/components/http_request/http_request.h
```diff
@@ -95,9 +95,26 @@ class HttpContainer : public Parented<HttpRequestComponent> {
...
+  /**
+   * @brief Get response headers.
+   *
+   * @return The key is the lower case response header name, the value is the header value.
+   */
+  std::map<std::string, std::list<std::string>> get_response_headers() const { return this->response_headers_; }
+
+  std::string get_response_header(const std::string &header_name) {
+    auto header_name_lower_case = str_lower_case(header_name);
+    if (this->response_headers_.count(header_name_lower_case) == 0) {
```
clydebarrow: I'm pretty sure this should be `!= 0`

esphome/components/http_request/http_request.h
```diff
@@ -95,9 +95,26 @@ class HttpContainer : public Parented<HttpRequestComponent> {
...
+      return this->response_headers_[header_name_lower_case].front();
```
clydebarrow: Why only return the first one?

esphome/components/http_request/http_request.h
```diff
@@ -239,6 +274,7 @@ template<typename... Ts> class HttpRequestSendAction : public Action<Ts...> {
   void encode_json_func_(Ts... x, JsonObject root) { this->json_func_(x..., root); }
   HttpRequestComponent *parent_;
   std::map<const char *, TemplatableValue<const char *, Ts...>> headers_{};
+  std::list<const char *> collect_header_names_{};
```
clydebarrow:
```suggestion
  std::list<std::string> collect_header_names_{};
```

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -121,6 +129,18 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
     // Still return the container, so it can be used to get the status code and error message
   }
 
+  container->response_headers_ = {};
+  for (int i = 0; i < container->client_.headers(); i++) {
+    std::string header_name = str_lower_case(container->client_.headerName(i).c_str());
```
clydebarrow: Rather than creating lower-cased copies of strings, just use `str_equals_case_insensitive` from helpers.h.

clydebarrow (changes requested):

clydebarrow:
And I would recommend splitting this into two PRs, one to update `http_request` and one for `online_image`.

esphome/components/http_request/__init__.py
candrews: Unfortunately, the Arduino API filters the response headers, requiring a list of what response headers to return before making the request.

https://github.com/esp8266/Arduino/blob/1a13ab95fb520e294ebffa4ece6b3a1466762604/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L726

https://forum.arduino.cc/t/getting-http-get-response-header/1014971/5

It isn't very pleasant... but it is what it is.

esphome/components/http_request/http_request.h
candrews: I agree! Thank you!

esphome/components/http_request/http_request.h
candrews: Because that's what callers (usually) want. Most of the time, there's only one header of a given name, and that caller wants that single header's value.

If we have `get_response_header` return a list of strings, that puts more work on the caller.

This approach is taken by other libraries, too, such as Java's servlet: https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html#getHeader(java.lang.String)

If the caller really wants all the headers, they can use `std::map<std::string, std::list<std::string>> get_response_headers()`.

esphome/components/http_request/http_request.h
candrews: Thank you!

esphome/components/http_request/http_request_arduino.cpp
candrews: I hadn't noticed that!

I've updated the MR accordingly, thank you!

esphome/components/http_request/http_request.h
```diff
@@ -119,19 +136,28 @@ class HttpRequestComponent : public Component {
...
+  }
+  std::shared_ptr<HttpContainer> post(std::string url, std::string body, std::list<Header> headers,
+                                      std::list<std::string> collect_header_names) {
+    return this->start(std::move(url), "POST", std::move(body), std::move(headers), std::move(collect_header_names));
   }
 
   virtual std::shared_ptr<HttpContainer> start(std::string url, std::string method, std::string body,
-                                               std::list<Header> headers) = 0;
+                                               std::list<Header> headers,
+                                               std::list<std::string> collect_header_names) = 0;
```
guillempages: isn't this a breaking change? If some code is directly using `start` instead of `get` or `post`? E.g. for a `delete` or `update` request?

esphome/components/http_request/__init__.py
```diff
@@ -264,11 +266,12 @@ async def http_request_action_to_code(config, action_id, template_arg, args):
                 template_ = await cg.templatable(json_[key], args, cg.std_string)
                 cg.add(var.add_json(key, template_))
     for key in config.get(CONF_HEADERS, []):
-        template_ = await cg.templatable(
-            config[CONF_HEADERS][key], args, cg.const_char_ptr
-        )
+        template_ = await cg.templatable(key, args, cg.const_char_ptr)
```
guillempages: Hm. Now that I see that, I'm not sure if it should be changed to `cg.std_string` instead; since the actual types are now strings and not char*.

Have you tried setting the headers with a lambda in the yaml code?

esphome/components/http_request/http_request.h
```diff
@@ -119,19 +136,28 @@ class HttpRequestComponent : public Component {
...
+    return this->post(std::move(url), std::move(body), std::move(headers), {});
+  }
+  std::shared_ptr<HttpContainer> post(std::string url, std::string body, std::list<Header> headers,
+                                      std::list<std::string> collect_header_names) {
+    return this->start(std::move(url), "POST", std::move(body), std::move(headers), std::move(collect_header_names));
   }
 
   virtual std::shared_ptr<HttpContainer> start(std::string url, std::string method, std::string body,
-                                               std::list<Header> headers) = 0;
+                                               std::list<Header> headers,
```
guillempages: maybe "headers" should now be renamed to "request_headers" or something similar, to distinguish them even better from the "collect_headers"?
Otherwise it feels "unbalanced" to call one parameter a generic "headers", and the other a specific "collect_headers"

guillempages: OTOH, the yaml configuration is called "headers", and it would be a breaking change to rename that as well 🤔

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -100,7 +101,14 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
   }
 
   // returned needed headers must be collected before the requests
-  static const char *header_keys[] = {"Content-Length", "Content-Type"};
+  const char *header_keys[collect_header_names.size() + 2];
```
guillempages: Since you are reworking this part anyway; wouldn't it be cleaner to just add the two headers "Content-Length" and "Content-Type" in the constructor (or even the python file)?
So that you then don't need to explicitly take care of getting a bigger array and adding them? 
This would also make the arduino and esp_idf versions behave the same (i.e. both request length and type; AFAIK it is currently only done for arduino).

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -100,7 +101,14 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
...
+  int index = 0;
+  for (auto const &header_name : collect_header_names) {
+    header_keys[index++] = header_name.c_str();
+  }
+  // header names must be lowercase
+  header_keys[index++] = "content-length";
+  header_keys[index++] = "content-type";
   static const size_t HEADER_COUNT = sizeof(header_keys) / sizeof(header_keys[0]);
```
guillempages: Wouldn't `index` actually have the same value as `HEADER_COUNT`?

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -121,6 +129,18 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
     // Still return the container, so it can be used to get the status code and error message
   }
 
+  container->response_headers_ = {};
+  for (int i = 0; i < container->client_.headers(); i++) {
```
guillempages: Is the compiler smart enough to move the `client_.headers()` call outside of the loop?
Or would this be considered a micro-optimization?

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -121,6 +129,18 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
...
+    const auto *header_name = container->client_.headerName(i).c_str();
```
guillempages: does it really need to be a c_str()? Or is it because the function returns an "arduino" `String`?
```suggestion
    const auto header_name = container->client_.headerName(i);
```

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,43 @@ namespace http_request {
 
 static const char *const TAG = "http_request.idf";
 
+struct UserData {
+  const std::list<std::string> &collect_header_names;
+  const std::map<std::string, std::list<std::string>> &response_headers;
+};
+
 void HttpRequestIDF::dump_config() {
   HttpRequestComponent::dump_config();
```
guillempages, line 28: maybe the "collect headers" (and even the "request headers") should also be dumped in the logs here?

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -89,6 +118,10 @@ std::shared_ptr<HttpContainer> HttpRequestIDF::start(std::string url, std::strin
 
   const int body_len = body.length();
 
+  config.event_handler = http_event_handler;
+  auto user_data = UserData{collect_header_names, {}};
+  config.user_data = static_cast<void *>(&user_data);
```
guillempages: Is this really needed? Wouldn't just setting `this` as user_data allow access to the headers as well? Without the need of copying the headers around?

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,43 @@ namespace http_request {
...
+esp_err_t HttpRequestIDF::http_event_handler(esp_http_client_event_t *evt) {
+  UserData *user_data = (UserData *) evt->user_data;
+  auto response_headers = user_data->response_headers;
+  auto collect_header_names = user_data->collect_header_names;
+
+  switch (evt->event_id) {
+    case HTTP_EVENT_ON_HEADER: {
+      auto *header_name = evt->header_key;
+      for (const auto &collect_header_name : collect_header_names) {
+        if (str_lower_case(collect_header_name) == header_name) {
```
guillempages: wouldn't it be more efficient to store the lowercase name when compiling the "collect_header_name" list instead of converting to lower case on each received header?

This would probably also avoid the need for the `str_equals_case_insensitive` call in the arduino version. (unless the server can return CamelCase headers; which would then also fail here as it is currently implemented)

esphome/components/http_request/http_request_idf.h
```diff
@@ -9,6 +9,8 @@
 #include <esp_netif.h>
 #include <esp_tls.h>
 
+#include <utility>
```
guillempages: Just out of curiosity; what does this header actually include? std::map, std::list & co? Woudln't it be better to explicitly add the needed headers instead, to reduce the amount of files to the preprocessor?

esphome/components/online_image/online_image.cpp
```diff
@@ -126,9 +132,17 @@ void OnlineImage::update() {
...
   accept_header.value = accept_mime_type + ",*/*;q=0.8";
 
+  if (!this->etag_.empty()) {
+    headers.push_back(http_request::Header{IF_NONE_MATCH_HEADER_NAME, this->etag_});
+  }
+
+  if (!this->last_modified_.empty()) {
+    headers.push_back(http_request::Header{IF_MODIFIED_SINCE_HEADER_NAME, this->last_modified_});
+  }
+
```
guillempages: I agree that this makes sense, to avoid downloading a file that hasn't changed on the server in case etag is not supported by it.

Nevertheless, there was a discussion a while ago with someone, that had actually the opposite use case; he actually wanted to return an error if the image *hadn't* been changed.
Or for debugging purposes, it might be desireable to re-download an image that hasn't changed.

It might be a good idea to make this (and the etag header) configurable. I have no problem in enabling it by default (i.e. avoiding unnecessary redownloads), but it would be nice to have an option to actually *force* a redownload.

esphome/components/online_image/online_image.cpp
```diff
@@ -126,9 +132,17 @@ void OnlineImage::update() {
...
   headers.push_back(accept_header);
 
-  this->downloader_ = this->parent_->get(this->url_, headers);
+  this->downloader_ = this->parent_->get(this->url_, headers, {ETAG_HEADER_NAME, LAST_MODIFIED_HEADER_NAME});
```
guillempages: IMHO the code would be slightly better readable, if the collect headers were extracted to a named constant:
```suggestion
  const auto collect_headers{ETAG_HEADER_NAME, LAST_MODIFIED_HEADER_NAME};
  this->downloader_ = this->parent_->get(this->url_, headers, collect_headers);
```

esphome/components/online_image/online_image.cpp
```diff
@@ -126,9 +132,17 @@ void OnlineImage::update() {
   }
   accept_header.value = accept_mime_type + ",*/*;q=0.8";
 
+  if (!this->etag_.empty()) {
```
guillempages: Maybe add a comment what the etag header is, for people that aren't that familiar with the http protocol

esphome/components/online_image/online_image.cpp
```diff
@@ -126,9 +132,17 @@ void OnlineImage::update() {
...
+    headers.push_back(http_request::Header{IF_NONE_MATCH_HEADER_NAME, this->etag_});
+  }
+
+  if (!this->last_modified_.empty()) {
+    headers.push_back(http_request::Header{IF_MODIFIED_SINCE_HEADER_NAME, this->last_modified_});
```
guillempages: Same here; a comment on how the "MODIFIED_SINCE" works would be nice. It's already pretty clear from the name, but nevertheless, for consistency ;-)

esphome/components/online_image/online_image.h
```diff
@@ -160,6 +160,14 @@ class OnlineImage : public PollingComponent,
...
    */
   int buffer_height_;
+  /**
+   * The value of the Last-Modified HTTP header provided in the last response.
+   */
+  std::string last_modified_;
+  /**
+   * The value of the ETag HTTP header provided in the last response.
+   */
+  std::string etag_;
```
guillempages: The rest of the code has `etag` before `last_modified` ;-)
```suggestion
  /**
   * The value of the ETag HTTP header provided in the last response.
   */
  std::string etag_;
  /**
   * The value of the Last-Modified HTTP header provided in the last response.
   */
  std::string last_modified_;
```

guillempages: Also; should they be explicitly initialized to the empty string?

esphome/components/online_image/image_decoder.cpp
```diff
@@ -25,15 +25,6 @@ void ImageDecoder::draw(int x, int y, int w, int h, const Color &color) {
   }
 }
 
-DownloadBuffer::DownloadBuffer(size_t size) : size_(size) {
```
guillempages: Any specific reason why you are reverting those fixes?

guillempages (changes requested):

guillempages:
Why did you revert #8212?

Also, I've learned the hard way that once someone started doing reviews, it's bad practice to do force-pushes; it's much easier for the reviewers to just add additional commits.

In the end, everything will be squashed together into a single commit when finally merging

esphome/components/http_request/http_request_arduino.cpp
candrews: Yes it would! Thank you for pointing that out.

esphome/components/http_request/http_request.h
candrews: Yes it would be, that's not good. I added `start` function so there will be no breaking change. I updated the PR accordingly.

esphome/components/http_request/http_request.h
candrews: I didn't rename it initially as I didn't want to make a breaking change :/

If you want it renamed, let me know and I'll do it.

esphome/components/online_image/online_image.cpp
candrews: :+1:

esphome/components/online_image/online_image.h
candrews: I did the reordering and initialized them both to empty string.

candrews:
> Why did you revert #8212?

I was testing. I undid the revert.

> Also, I've learned the hard way that once someone started doing reviews, it's bad practice to do force-pushes; it's much easier for the reviewers to just add additional commits.
> 
> In the end, everything will be squashed together into a single commit when finally merging

I won't be doing any more force pushes :)

I was just working on a project that requires force pushes so I was in that mindset. I'll be now happily leaving that mindset :)

esphome/components/http_request/http_request_arduino.cpp
candrews: > add the two headers "Content-Length" and "Content-Type" in the constructor

Can you clarify what constructor / where exactly you're suggesting?

I like the idea of improving consistency so I'm eager to make this improvement.

esphome/components/http_request/http_request.h
guillempages: It's not my component, so I can't make that call. Jus tpointing out options so that Jesse or whoever feels responsible for the http_request component can decide ;-)

esphome/components/http_request/http_request_arduino.cpp
guillempages: I meant, the "collect_header_names" array is being filled somewhere; either in the class constructor as an empty array, or it is being filled by the generated code when calling `cg.add(var.add_collect_header_name(value))` in the loop. So the two headers could be hard-coded before or after the loop in python. 
If the headers are stored in an std::set instead of an std::list, you don't even need to worry about a header being set twice (once explicitly in yaml, and once implicitly in the code generation).

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -15,7 +15,8 @@ namespace http_request {
 static const char *const TAG = "http_request.arduino";
 
 std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::string method, std::string body,
-                                                         std::list<Header> headers) {
+                                                         std::list<Header> headers,
+                                                         std::list<std::string> collect_header_names) {
```
guillempages: maybe change this to an `std::set` instead, so that duplicates are automatically filtered out?

esphome/components/online_image/online_image.cpp
candrews: Comments added in 904571c74

esphome/components/online_image/online_image.cpp
candrews: Comments added in 904571c74

esphome/components/http_request/http_request_arduino.cpp
candrews: Changed to a set :)

esphome/components/http_request/http_request_idf.h
candrews: The include isn't necessary at all so I removed it.

esphome/components/http_request/__init__.py
candrews: Yes, I have and it works. I've provided the example YAML I used for testing in the PR description.

esphome/components/online_image/online_image.cpp
candrews: I see what you're saying and I'll come up with something.

I've separated out the `online_image` component changes, which I'll submit in a future MR once this one is in, as requested

esphome/components/online_image/image_decoder.cpp
candrews: I was testing to see if those changes were causing errors. They weren't. I reverted the reverted.

esphome/components/http_request/http_request_arduino.cpp
candrews: I have made this change. Thank you!

esphome/components/http_request/http_request_arduino.cpp
candrews: Great point, I implemented this change in 4338161deb3e43f95d48c45a1e0b99f3420a31f3

esphome/components/http_request/http_request_idf.cpp
candrews: The collect header names are a property of the request in progress (inside `start`), not of the `HttpRequestComponent`, and therefore they're not available in `HttpRequestComponent::dump_config`.

candrews:
I tested this MR with both `arduino` and `esp-idf` framework types successfully.

I think it's ready for another round of review.

esphome/components/http_request/http_request.cpp
```diff
@@ -20,5 +20,27 @@ void HttpRequestComponent::dump_config() {
...
+  ESP_LOGD(TAG, "Entered get_response_header");
+  auto response_headers = this->get_response_headers();
+  auto header_name_lower_case = str_lower_case(header_name);
+  if (response_headers.count(header_name_lower_case) == 0) {
+    ESP_LOGD(TAG, "No header with name %s found", header_name_lower_case.c_str());
+    return "";
+  } else {
+    auto values = response_headers[header_name_lower_case];
+    if (values.empty()) {
+      ESP_LOGD(TAG, "header with name %s returned an empty list, this shouldn't happen",
```
guillempages:
```suggestion
      ESP_LOGE(TAG, "header with name %s returned an empty list, this shouldn't happen",
```
If this shouldn't happen, the log should probably be an Error log then, or at least a warning

esphome/components/http_request/http_request.cpp
```diff
@@ -20,5 +20,27 @@ void HttpRequestComponent::dump_config() {
...
+               header_name_lower_case.c_str());
+      return "";
+    } else {
+      ESP_LOGD(TAG, "Header with name %s found", header_name_lower_case.c_str());
+      auto header_value = values.front();
+      ESP_LOGD(TAG, "Header with name %s found with value %s", header_name_lower_case.c_str(), header_value.c_str());
```
guillempages: Do you really need those two logs so close to each other? The seocnd one has the same information as the first one anyway...
```suggestion
      auto header_value = values.front();
      ESP_LOGD(TAG, "Header with name %s found with value %s", header_name_lower_case.c_str(), header_value.c_str());
```

esphome/components/http_request/http_request.cpp
```diff
@@ -20,5 +20,27 @@ void HttpRequestComponent::dump_config() {
...
+      return values.front();
```
guillempages: Since you are already packing that into a variable; why call the "front()" method again?
```suggestion
      return header_value;
```

esphome/components/http_request/http_request.h
```diff
@@ -119,19 +130,34 @@ class HttpRequestComponent : public Component {
   void set_follow_redirects(bool follow_redirects) { this->follow_redirects_ = follow_redirects; }
   void set_redirect_limit(uint16_t limit) { this->redirect_limit_ = limit; }
 
-  std::shared_ptr<HttpContainer> get(std::string url) { return this->start(std::move(url), "GET", "", {}); }
-  std::shared_ptr<HttpContainer> get(std::string url, std::list<Header> headers) {
-    return this->start(std::move(url), "GET", "", std::move(headers));
+  std::shared_ptr<HttpContainer> get(const std::string &url) { return this->start(url, "GET", "", {}); }
+  std::shared_ptr<HttpContainer> get(const std::string &url, const std::list<Header> &headers) {
```
guillempages: If we go for the breaking renaming of headers, this should be renamed as well.
And even if we don't rename the yaml options, renaming the parameter names here would make the code clearer.

esphome/components/http_request/__init__.py
```diff
@@ -47,6 +47,7 @@
 CONF_MAX_RESPONSE_BUFFER_SIZE = "max_response_buffer_size"
 CONF_ON_RESPONSE = "on_response"
 CONF_HEADERS = "headers"
```
guillempages, line 49: Since we now have `COLLECT_HEADER_NAMES`, IMHO we should rename `HEADERS` to `REQUEST_HEADERS` or similar, to make clearer what is what.
But I'd like @jesserockz to comment on that, and not base the breaking change only on my opinion ;-)

esphome/components/http_request/__init__.py
```diff
@@ -47,6 +47,7 @@
...
+CONF_COLLECT_HEADER_NAMES = "collect_header_names"
```
guillempages: Maybe rename to just `collect_headers`, for consistency with the already available headers option?
```suggestion
CONF_COLLECT_HEADERS = "collect_headers"
```

esphome/components/http_request/http_request.h
```diff
@@ -119,19 +130,34 @@ class HttpRequestComponent : public Component {
...
+    return this->start(url, "GET", "", headers);
+  }
+  std::shared_ptr<HttpContainer> get(std::string url, std::list<Header> headers,
```
guillempages: same here

esphome/components/http_request/http_request.h
```diff
@@ -119,19 +130,34 @@ class HttpRequestComponent : public Component {
...
+                                     std::set<std::string> collect_header_names) {
+    return this->start(std::move(url), "GET", "", std::move(headers), std::move(collect_header_names));
+  }
+  std::shared_ptr<HttpContainer> post(const std::string &url, const std::string &body) {
+    return this->start(url, "POST", body, {});
   }
-  std::shared_ptr<HttpContainer> post(std::string url, std::string body) {
-    return this->start(std::move(url), "POST", std::move(body), {});
+  std::shared_ptr<HttpContainer> post(const std::string &url, const std::string &body,
+                                      const std::list<Header> &headers) {
```
guillempages: and here

esphome/components/http_request/http_request.h
```diff
@@ -119,19 +130,34 @@ class HttpRequestComponent : public Component {
...
+    return this->start(url, "POST", body, headers);
   }
-  std::shared_ptr<HttpContainer> post(std::string url, std::string body, std::list<Header> headers) {
-    return this->start(std::move(url), "POST", std::move(body), std::move(headers));
+  std::shared_ptr<HttpContainer> post(std::string url, std::string body, std::list<Header> headers,
```
guillempages: and here

esphome/components/http_request/http_request.h
```diff
@@ -119,19 +130,34 @@ class HttpRequestComponent : public Component {
...
+                                      std::set<std::string> collect_header_names) {
+    return this->start(std::move(url), "POST", std::move(body), std::move(headers), std::move(collect_header_names));
+  }
+
+  std::shared_ptr<HttpContainer> start(const std::string &url, const std::string &method, const std::string &body,
+                                       const std::list<Header> &headers) {
```
guillempages: yet another one

esphome/components/http_request/http_request.h
guillempages: Let's hope that Jesse comments on the renaming; the more I think about it, the more convinced I am that we should do the breaking change renaming

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -15,7 +15,8 @@ namespace http_request {
 static const char *const TAG = "http_request.arduino";
 
 std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::string method, std::string body,
-                                                         std::list<Header> headers) {
+                                                         std::list<Header> headers,
```
guillempages: another possible `headers` rename

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -100,9 +101,13 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
   }
 
   // returned needed headers must be collected before the requests
-  static const char *header_keys[] = {"Content-Length", "Content-Type"};
-  static const size_t HEADER_COUNT = sizeof(header_keys) / sizeof(header_keys[0]);
-  container->client_.collectHeaders(header_keys, HEADER_COUNT);
+  const char *header_keys[collect_header_names.size() + 2];
```
guillempages: Why the +2?
```suggestion
  const char *header_keys[collect_header_names.size()];
```

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -100,9 +101,13 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
...
+  int index = 0;
+  for (auto const &header_name : collect_header_names) {
+    header_keys[index++] = header_name.c_str();
+  }
+  // header names must be lowercase
```
guillempages: Does the comment match the code?

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -121,6 +126,20 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
     // Still return the container, so it can be used to get the status code and error message
   }
 
+  container->response_headers_ = {};
+  auto header_count = container->client_.headers();
+  for (int i = 0; i < header_count; i++) {
+    const std::string header_name = str_lower_case(container->client_.headerName(i).c_str());
+    for (const auto &collect_header_name : collect_header_names) {
```
guillempages: Instead of the nested `for`; would it be possible to use the `count()` method on the `std::set`, to make the code simpler?

esphome/components/http_request/http_request_arduino.h
```diff
@@ -30,8 +30,8 @@ class HttpContainerArduino : public HttpContainer {
 
 class HttpRequestArduino : public HttpRequestComponent {
  public:
-  std::shared_ptr<HttpContainer> start(std::string url, std::string method, std::string body,
-                                       std::list<Header> headers) override;
+  std::shared_ptr<HttpContainer> start(std::string url, std::string method, std::string body, std::list<Header> headers,
```
guillempages: maybe rename `headers`?

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,47 @@ namespace http_request {
...
+};
+
 void HttpRequestIDF::dump_config() {
   HttpRequestComponent::dump_config();
   ESP_LOGCONFIG(TAG, "  Buffer Size RX: %u", this->buffer_size_rx_);
   ESP_LOGCONFIG(TAG, "  Buffer Size TX: %u", this->buffer_size_tx_);
 }
 
+esp_err_t HttpRequestIDF::http_event_handler(esp_http_client_event_t *evt) {
+  ESP_LOGD(TAG, "Entered http_event_handler");
```
guillempages: If you really want to keep the log, it should be made verbose; this is being logged very often otherwise
```suggestion
  ESP_LOGV(TAG, "Entered http_event_handler");
```

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,47 @@ namespace http_request {
...
+          break;
+        }
+      }
+      break;
+    }
+    default: {
+      break;
+    }
+  }
+  ESP_LOGD(TAG, "Exit http_event_handler");
```
guillempages: This is logged very often; if you really need to keep it, at least make it verbose.
```suggestion
  ESP_LOGV(TAG, "Exit http_event_handler");
```

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,47 @@ namespace http_request {
...
   ESP_LOGCONFIG(TAG, "  Buffer Size TX: %u", this->buffer_size_tx_);
 }
 
+esp_err_t HttpRequestIDF::http_event_handler(esp_http_client_event_t *evt) {
+  ESP_LOGD(TAG, "Entered http_event_handler");
+  UserData *user_data = (UserData *) evt->user_data;
+
+  switch (evt->event_id) {
+    case HTTP_EVENT_ON_HEADER: {
+      ESP_LOGD(TAG, "Entered HTTP_EVENT_ON_HEADER");
```
guillempages: I'd say this was useful when developing the change, but won't be useful at all for debugging later on
```suggestion
```

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,47 @@ namespace http_request {
...
+      ESP_LOGD(TAG, "header name: %s", evt->header_key);
```
guillempages: Maybe also make this verbose? But not so sure about this like in the above cases.

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,47 @@ namespace http_request {
...
+      const std::string header_name = str_lower_case(evt->header_key);
+      for (const auto &collect_header_name : user_data->collect_header_names) {
+        if (str_equals_case_insensitive(collect_header_name, header_name)) {
```
guillempages: This could probably also be handled with the `set::count()` method
```suggestion
      if (user_data->collect_header_names.count(header_name)) {
```
(You might need to either make sure that all "collect_header_names" are always stored as lowercase, or that the comparator in the set is case insensitive. I'd vote for the first option.

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -19,14 +19,47 @@ namespace http_request {
...
+      break;
+    }
+  }
+  ESP_LOGD(TAG, "Exit http_event_handler");
+  return ESP_OK;
+}
+
 std::shared_ptr<HttpContainer> HttpRequestIDF::start(std::string url, std::string method, std::string body,
-                                                     std::list<Header> headers) {
+                                                     std::list<Header> headers,
```
guillempages: If renaming `headers`, here's another instance

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -118,14 +155,21 @@ std::shared_ptr<HttpContainer> HttpRequestIDF::start(std::string url, std::strin
     esp_http_client_cleanup(client);
     return nullptr;
   }
+  ESP_LOGD(TAG, "HTTP Request ok");
 
   container->feed_wdt();
+  ESP_LOGD(TAG, "About to esp_http_client_fetch_headers");
```
guillempages: Do you still need *all* of these debug logs? Or were they only needed during development?

esphome/components/http_request/http_request_idf.cpp
```diff
@@ -118,14 +155,21 @@ std::shared_ptr<HttpContainer> HttpRequestIDF::start(std::string url, std::strin
...
   container->feed_wdt();
+  ESP_LOGD(TAG, "About to esp_http_client_get_status_code");
   container->status_code = esp_http_client_get_status_code(client);
   container->feed_wdt();
+  ESP_LOGD(TAG, "About to set_response_headers, size: %d", user_data.response_headers.size());
+  container->set_response_headers(user_data.response_headers);
+  ESP_LOGD(TAG, "Done with set_response_headers");
   if (is_success(container->status_code)) {
     container->duration_ms = millis() - start;
+    ESP_LOGD(TAG, "HTTP Request successful");
```
guillempages: Isn't this also logged anyway on the calling method?

esphome/components/http_request/http_request_idf.h
```diff
@@ -29,8 +33,8 @@ class HttpRequestIDF : public HttpRequestComponent {
  public:
   void dump_config() override;
 
-  std::shared_ptr<HttpContainer> start(std::string url, std::string method, std::string body,
-                                       std::list<Header> headers) override;
+  std::shared_ptr<HttpContainer> start(std::string url, std::string method, std::string body, std::list<Header> headers,
```
guillempages: yet another `headers` instance

esphome/components/http_request/http_request_idf.cpp
candrews: I removed it.

esphome/components/http_request/http_request_idf.cpp
candrews: I removed it.

esphome/components/http_request/http_request_idf.cpp
candrews: I removed it

esphome/components/http_request/http_request_idf.cpp
candrews: I agree they're unnecessary - I've removed them.

esphome/components/http_request/http_request_idf.cpp
candrews: Yes, I've removed it.

candrews:
@jesserockz can you please take a look at this PR?

I'm particularly curious in your thoughts on naming. Should we do a breaking change and rename the existing `headers` field to `request_headers` as raised at https://github.com/esphome/esphome/pull/8224#discussion_r1951520976 ?

clydebarrow:
Jesse is on parental leave right now, and while I don't speak for him I do know he has previously been quite willing to accept breaking changes that result in a worthwhile improvement in the code quality.

IMO it makes sense for it to be clear which headers are being referred to. `request_headers` and `response_headers` would be consistent with usage elsewhere and clearly identify what they are associated with.

candrews:
> Jesse is on parental leave right now, and while I don't speak for him I do know he has previously been quite willing to accept breaking changes that result in a worthwhile improvement in the code quality.

Congratulations to Jesse! :baby: are awesome!

> IMO it makes sense for it to be clear which headers are being referred to. `request_headers` and `response_headers` would be consistent with usage elsewhere and clearly identify what they are associated with.


I'll update this MR accordingly, thank you for the direction

esphome/components/http_request/__init__.py
candrews: I've renamed "headers" to "request_headers"

esphome/components/http_request/__init__.py
candrews: That sounds like a good idea - I have done so.

esphome/components/http_request/http_request.h
candrews: I agree and have done so.

esphome/components/http_request/http_request.h
candrews: Done!

esphome/components/http_request/http_request.h
candrews: Done!

esphome/components/http_request/http_request.h
candrews: I have done the renaming.

esphome/components/http_request/http_request_arduino.cpp
candrews: Done!

esphome/components/http_request/http_request_arduino.h
candrews: Renamed to `request_headers`

esphome/components/http_request/http_request_idf.cpp
candrews: Renamed.

esphome/components/http_request/http_request_idf.h
candrews: Done!

esphome/components/http_request/http_request_arduino.cpp
candrews: I removed the comment; arduino handles case insensitive header name matching.

esphome/components/http_request/http_request_idf.cpp
candrews: I ensured all names are lower case and made the suggested change.

esphome/components/http_request/http_request.cpp
```diff
@@ -20,5 +20,25 @@ void HttpRequestComponent::dump_config() {
   }
 }
 
+std::string HttpContainer::get_response_header(const std::string &header_name) {
+  auto response_headers = this->get_response_headers();
+  auto header_name_lower_case = str_lower_case(header_name);
+  if (response_headers.count(header_name_lower_case) == 0) {
+    ESP_LOGD(TAG, "No header with name %s found", header_name_lower_case.c_str());
```
guillempages: If the code explicitly requests a header, and this is not available; shouldn't this be considered at least a warning? Debug logs might be easy to oversee.

Of course you have done more checks; maybe this *is* a frequent case which then would be OK as debug.
```suggestion
    ESP_LOGW(TAG, "No header with name %s found", header_name_lower_case.c_str());
```

esphome/components/http_request/http_request.h
```diff
@@ -176,16 +216,21 @@ template<typename... Ts> class HttpRequestSendAction : public Action<Ts...> {
...
       Header header;
       header.name = item.first;
       header.value = val.value(x...);
-      headers.push_back(header);
+      request_headers.push_back(header);
+    }
+    std::set<std::string> collect_headers;
+    for (const auto &item : this->collect_headers_) {
+      collect_headers.insert(item);
     }
```
guillempages: Why do you need to create a new copy of `collect_headers`? Doesn't it work if you just pass `this->collect_headers_` to the `start` call? `collect_headers_` is also an `std::set<std::string>`, isn't it?

esphome/components/http_request/http_request_arduino.cpp
```diff
@@ -121,6 +125,20 @@ std::shared_ptr<HttpContainer> HttpRequestArduino::start(std::string url, std::s
     // Still return the container, so it can be used to get the status code and error message
   }
 
+  container->response_headers_ = {};
+  auto header_count = container->client_.headers();
+  for (int i = 0; i < header_count; i++) {
+    const std::string header_name = str_lower_case(container->client_.headerName(i).c_str());
+    for (const auto &collect_header : collect_headers) {
+      if (str_equals_case_insensitive(collect_header, header_name)) {
```
guillempages: `header_name` is already lowercase (you converted it two lines above). If you make sure that `collect_header` is also lowercase (e.g. by converting it directly in the `add` method), you don't need to do a case insensitive comparison, since everything is already lowercase.

And then you could also use the `if collect_headers.count(header_name) >0)` trick, instead of the explicit `for` and `if` (like in the idf case)

candrews: That's definitely an improvement, I made that change: 2e6bbea7748621bb8e92a564c4fd701723242c89

esphome/components/http_request/http_request.h
candrews: I think that's a holdover from an earlier approach. I removed the unnecessary copy: dda3628cdcb10799f425da2eff3c4c36ac041045 :bowing_man:

guillempages (approved):
LGTM

candrews:
> LGTM

What are the next steps?

guillempages:
You need the approval of someone with enough rights. 

I think the PR is good to go, but don't have enough privileges to actually merge it.

clydebarrow (approved):

jant90:
@candrews I was very happy to see this feature in ESPHome, however at this time the implementation seems buggy, I opened an issue (https://github.com/esphome/issues/issues/7083) for it, maybe you have an idea what's going wrong here?

PR #8228, opened 2025-02-11
[i80] Initial implementation

clydebarrow:
# What does this implement/fix?

## Types of changes

Adding a new bus parallel type - `i80` - used for interfacing  LCD display controllers. This is used in a similar way to the `spi` component, and will be utilised by the `ili9xxx` display driver, enabling support for the Seeed [W32-SC01-Plus](https://www.seeedstudio.com/WT32-3-5-Inch-Display-p-5542.html) and the [LilyGo T-Display S3](https://www.lilygo.cc/products/t-display-s3?variant=42589373268149).

This PR builds on #8227 and adds the i80 bus component. Further PRs will add functionality to the SPI and ili9xxx components.

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes 
- https://github.com/esphome/feature-requests/issues/2387
- https://github.com/esphome/feature-requests/issues/901
- 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3761

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
i80:
  dc_pin: 7
  wr_pin: 8
  rd_pin: 9
  data_pins:
    - 39
    - 40
    - 41
    - 42
    -
      ignore_strapping_warning: true
      number: 45
    -
      ignore_strapping_warning: true
      number: 46
    - 47
    - 48
    
display:
  - platform: ili9xxx
    bus_type: i80
    cs_pin: 6
    reset_pin: 5
    model: st7789v
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8227, opened 2025-02-11
[io_bus] Initial implementation

clydebarrow:
# What does this implement/fix?

## Types of changes

In preparation for adding a new bus parallel type - `i80` - used for interfacing  LCD display controllers. This is used in a similar way to the `spi` component, and will be utilised by the `ili9xxx` display driver, enabling support for the Seeed [W32-SC01-Plus](https://www.seeedstudio.com/WT32-3-5-Inch-Display-p-5542.html) and the [LilyGo T-Display S3](https://www.lilygo.cc/products/t-display-s3?variant=42589373268149).

This PR contains only the `io_bus`  components. Further PRs will add the `i80` bus component and functionality to the SPI and ili9xxx components.

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

**Related issue or feature (if applicable):** fixes 
- https://github.com/esphome/feature-requests/issues/2387
- https://github.com/esphome/feature-requests/issues/901
- 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3761

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
i80:
  dc_pin: 7
  wr_pin: 8
  rd_pin: 9
  data_pins:
    - 39
    - 40
    - 41
    - 42
    -
      ignore_strapping_warning: true
      number: 45
    -
      ignore_strapping_warning: true
      number: 46
    - 47
    - 48
    
display:
  - platform: ili9xxx
    bus_type: i80
    cs_pin: 6
    reset_pin: 5
    model: st7789v
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8230, opened 2025-02-11
[ili9xxx] Use io_bus for communication

clydebarrow:

## Types of changes

Final stage of adding a new bus parallel type - `i80` - used for interfacing  LCD display controllers. This is used in a similar way to the `spi` component, and is utilised by the `ili9xxx` display driver, enabling support for the Seeed [W32-SC01-Plus](https://www.seeedstudio.com/WT32-3-5-Inch-Display-p-5542.html) and the [LilyGo T-Display S3](https://www.lilygo.cc/products/t-display-s3?variant=42589373268149).

This PR builds on #8227, #8228 #8229 and updates the `ili9xxx` driver to use the `io_bus` interface via either SPI or I80.

- [ ] Other

- https://github.com/esphome/feature-requests/issues/2387
- https://github.com/esphome/feature-requests/issues/901
- 

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#3761

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
i80:
  dc_pin: 7
  wr_pin: 8
  rd_pin: 9
  data_pins:
    - 39
    - 40
    - 41
    - 42
    -
      number: 45
    -
      number: 46
    - 47
    - 48
    
display:
  - platform: ili9xxx
    bus_type: i80
    cs_pin: 6
    reset_pin: 5
    model: st7789v
```

## Checklist:


esphome/components/ili9xxx/display.py
```diff
@@ -144,51 +146,71 @@ def _validate(config):
...
-            cv.Optional(CONF_COLOR_ORDER): cv.one_of(*COLOR_ORDERS.keys(), upper=True),
-            cv.Exclusive(CONF_ROTATION, CONF_ROTATION): validate_rotation,
-            cv.Exclusive(CONF_TRANSFORM, CONF_ROTATION): cv.Schema(
+            TYPE_I80: BASE_SCHEMA.extend(i80.i80_client_schema()).extend(
                 {
-                    cv.Optional(CONF_SWAP_XY, default=False): cv.boolean,
-                    cv.Optional(CONF_MIRROR_X, default=False): cv.boolean,
-                    cv.Optional(CONF_MIRROR_Y, default=False): cv.boolean,
+                    cv.Optional(CONF_DC_PIN): cv.invalid(
+                        "DC pin should be specified in the i80 component only"
```
guillempages, line 207: Why is that? Couldn't it be specified here instead, so that the usage is consistent within SPI and i80?
I'm not familiar with the i80 bus or display drivers at all; just speaking from a "generic" view.
IMHO it would be better if all displays would use some consistent/similar configurations whenever possible (of course taking the specifics of every controller when required)

esphome/components/ili9xxx/ili9xxx_init.h
```diff
@@ -6,6 +6,7 @@
 namespace esphome {
 namespace ili9xxx {
 
+#define PROGMEM
```
guillempages: shouldn't this be guarded by an `#ifndef`? or an `#ifdef USE_ESP32` or similar? I.e. do esp8266 or rp2040 also work if "PROGMEM" is replaced by the empty string? I would think there is a reason why the macro is used in this file.

esphome/components/ili9xxx/display.py
clydebarrow: It would be better to have more consistency, but the i80 bus libraries require the DC pin specified on setup and manage it, whereas the SPI bus doesn't know or care about DC.

esphome/components/ili9xxx/ili9xxx_init.h
clydebarrow: Not sure why that's there, will check.

clydebarrow: So the problem is that the ili9xxx code doesn't use PROGMEM read functions anymore - that has been the case for many months. So on ESP8266 the driver will likely not work if the init tables are in code memory. On the other chips PROGMEM is already a no-op so it makes no difference. I'll remove it here, but this will need addressing if support for the ESP8266 is to be restored. I do plan to merge the ili9xxx and qspi_dbi drivers into one `mipi_dbi` driver after this PR is merged, and at that time the init tables can be moved into Python code so only those actually in use take up any space, so it won't matter too much if they are in RAM.

I suspect I added that define when testing on ESP8266 to get it to work. Nobody else has complained AFAIK.

clydebarrow:
Closed in favour of #8383

clydebarrow:
To use the changes in this PR:

   ```bash
   # Clone the repository:
   git clone https://github.com/esphome/esphome
   cd esphome

   # Checkout the PR branch:
   git fetch origin pull/8230/head:i80-ili
   git checkout i80-ili

   # Install the development version:
   script/setup

   # Activate the development version:
   source venv/bin/activate
   ```

Now you can run `esphome` as usual to test the changes in this PR.


---
(Added by the PR bot)

PR #8236, opened 2025-02-11
[online_image]Fix reset if buffer not allocated

guillempages:

When (re)allocating the image buffer, the "out of memory" case was being ignored by some decoders.
Make sure that decoders don't try to decode to an empty buffer.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -64,28 +64,28 @@ void OnlineImage::release() {
   }
 }
 
-ssize_t OnlineImage::resize_(int width_in, int height_in) {
+size_t OnlineImage::resize_(int width_in, int height_in) {
   int width = this->fixed_width_;
   int height = this->fixed_height_;
-  if (this->auto_resize_()) {
+  if (this->is_auto_resize_()) {
     width = width_in;
     height = height_in;
     if (this->width_ != width && this->height_ != height) {
       this->release();
     }
   }
+  size_t new_size = this->get_buffer_size_(width, height);
   if (this->buffer_) {
     // Buffer already allocated => no need to resize
-    return 0;
+    return new_size;
   }
-  ssize_t new_size = this->get_buffer_size_(width, height);
   ESP_LOGD(TAG, "Allocating new buffer of %zu bytes", new_size);
   this->buffer_ = this->allocator_.allocate(new_size);
   if (this->buffer_ == nullptr) {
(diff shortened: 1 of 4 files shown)


esphome/components/online_image/online_image.cpp
```diff
@@ -75,22 +75,23 @@ bool OnlineImage::resize_(int width_in, int height_in) {
     }
   }
   if (this->buffer_) {
-    return false;
+    // Buffer already allocated => no need to resize
+    return 0;
```
clydebarrow: Why not return the buffer size here instead of 0?

clydebarrow (changes requested):

jesserockz (dismissed):

esphome/components/online_image/online_image.cpp
guillempages: Because I want to explicitly give the information that no resize actually happened.
Before this change, it was a bool, and both "not changed" and "out of memory" were returning `false`; which is what caused the problem.
So "0" means "not changed", "-1" means "error", and ">0" means "resized"

clydebarrow: So those return values would be better as an enum. Returning the size implies more information that just "resized".

guillempages: I'll have to think about it. Now it's time to go to bed.

clydebarrow: Sure. Just looking at the code I found it confusing that the value returned was sometimes the size and sometimes not, but the calling code didn't seem to distinguish between the two. An enum would make it explicit what information is being returned - if it's actually necessary and I can't see that it is. Otherwise just true on success and false on error would work.

guillempages: Ok, Since the information was actually not really being used, and I don't have a use case in mind for the "has not been resized, but no error" case, I changed it to return the actual size of the buffer; regardless of whether resized or not. And 0 in case of error; which is also the "size" of the buffer in that case 😆 

I also added additional logs in some corner cases, to aid in debugging.

clydebarrow (approved):

PR #8252, opened 2025-02-14
LVGL input group switching and page handling

bdm310:

This makes some changes to how input groups are handled to deal with the underlying LVGL behavior during page switching. Core issue is described in https://github.com/esphome/issues/issues/6725

The approach here is to extend the `CONF_DEFAULT_GROUP` concept from the top level LVGL component down to pages as well. As objects are added, the LVGL default group is changed so that subsequent widgets are added to the right group for where they are in the hierarchy. This method has the benefit of keeping the automatic widget type filtering that LVGL does on the default group so that only widgets that are actually inputs get added. Additionally, there is now no need to specify groups for input devices when multiple LVGL instances are defined.

When pages are shown, all the input devices of that particular LVGL instance that do not have a specific group configured for them are switched to the page-specific default group.

This should result in the same behavior for users that have configured specific groups on input devices and widgets while also giving a more sensible and expected behavior when groups aren't specifically defined.

I've categorized this as non-breaking because the prior behavior was really non-functional for multipage setups so I can't imagine anyone was intentionally using it.

I've tested this with multiple displays/multiple LVGL configs, multiple pages, no pages, touchscreen, encoder, keypad, and multiple input devices per LVGL instance.

No docs or test changes yet, I can work on that if this concept looks good to others.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [X] Host

```yaml
esphome:
  name: sdl

host:

logger:

display:
  - platform: sdl
    id: disp0
    dimensions:
      width: 320
      height: 320
    auto_clear_enabled: false 
  - platform: sdl
    id: disp1
    dimensions:
      width: 640
      height: 640
    auto_clear_enabled: false 

touchscreen:
  - platform: sdl
    id: touch0
    display: disp0
    sdl_id: disp0

  - platform: sdl
    id: touch1
    display: disp1
    sdl_id: disp1

binary_sensor:
  - platform: sdl
    id: key_up1
    key: SDLK_a
    sdl_id: disp0
  - platform: sdl
    id: key_down1
    key: SDLK_d
    sdl_id: disp0
  - platform: sdl
    id: key_enter1
    key: SDLK_s
    sdl_id: disp0

  - platform: sdl
    id: keypad_left
    key: SDLK_q
    sdl_id: disp0
  - platform: sdl
    id: keypad_right
    key: SDLK_e
    sdl_id: disp0
  - platform: sdl
    id: keypad_enter
    key: SDLK_w
    sdl_id: disp0

  - platform: sdl
    id: key_up2
    key: SDLK_j
    sdl_id: disp0
  - platform: sdl
    id: key_down2
    key: SDLK_l
    sdl_id: disp0
  - platform: sdl
    id: key_enter2
    key: SDLK_k
    sdl_id: disp0

lvgl:
  #log_level: INFO
  - id: lvgl_0
    displays: disp0

    encoders:
      group: disp0_group
      enter_button: key_enter1
      sensor:
        left_button: key_up1
        right_button: key_down1

    keypads:
      next: keypad_left
      prev: keypad_right
      enter: keypad_enter

    touchscreens:
      touchscreen_id: touch0

    theme:
      button:
        focused:
          border_width: 5
          border_color: 0x00FF00
      label:
        focused:
          border_width: 5
          border_color: 0x00FF00

    pages:
      - id: page1
        layout:
          type: flex
        widgets:
          - label:
              text: "Hello Page 1"
          - button:
              widgets:
                - label: 
                    text: "Next"
              on_click:
                - lvgl.page.show:
                    lvgl_id: lvgl_0
                    id: page2
          - button:
              widgets:
                - label: 
                    text: "Prev"
              on_click:
                then:
                  - lvgl.page.show:
                      lvgl_id: lvgl_0
                      id: page2
      - id: page2
        layout:
          type: flex
        widgets:
          - label:
              text: "Hello Page 2"
          - button:
              group: disp0_group
              widgets:
                - label: 
                    text: "Next"
              on_click:
                then:
                  - lvgl.page.show:
                      lvgl_id: lvgl_0
                      id: page1
          - button:
              group: disp0_group
              widgets:
                - label: 
                    text: "Prev"
              on_click:
                then:
                  - lvgl.page.show:
                      lvgl_id: lvgl_0
                      id: page1

  - id: lvgl_1
    displays: disp1

    encoders:
#      group: disp1_group
      enter_button: key_enter2
      sensor:
        left_button: key_up2
        right_button: key_down2

    touchscreens:
      touchscreen_id: touch1

    layout:
      type: flex

    widgets:
      - label:
          text: "Hello 2"
      - button:
#          group: disp1_group
          widgets:
            - label: 
                text: "Button"
      - button:
#          group: disp1_group
          widgets:
            - label: 
                text: "Button"
```

## Checklist:


esphome/components/lvgl/widgets/page.py
```diff
@@ -166,7 +170,9 @@ async def add_pages(lv_component, config):
     for pconf in config.get(CONF_PAGES, ()):
         id = pconf[CONF_ID]
         skip = pconf[CONF_SKIP]
-        var = cg.new_Pvariable(id, skip)
+        default_group = cg.Pvariable(pconf[CONF_DEFAULT_GROUP], lv_expr.group_create())
+        cg.add(lv.group_set_default(default_group))
```
bdm310, line 174: This line results in double generation of the call to `lv_group_set_default` ?

clydebarrow:
Rather than implicitly creating hidden groups, I think it would be better to allow a page to have a named group specified that becomes the default group for widgets within that page. Switching of encoder groups will still be needed but that could be done automatically if a single encoder is configured, but require configuration per-page if there are multiple.

bdm310:
I would add documentation to describe the DEFAULT_GROUP concept, in that way they wouldn't be hidden per se and can be changed by the user if desired. Very similar to IDs, they aren't required and are then generated for you. This gives exactly the behavior I expect by default even for multiple encoders and easily allows for overriding and customizing in any way desired.

Perhaps I'm not seeing all the use cases for pages? I expect only one page to be visible at a time and I expect my input to just work on that visible thing by default.

clydebarrow:
> I would add documentation to describe the DEFAULT_GROUP concept, in that way they wouldn't be hidden per se and can be changed by the user if desired.

How would it be changed by the user if there is no config option to do so? With your proposal as it stands the group name could be changed, but there is no way to turn off the assignment of a per-page default group. While that might suit the obvious use-case it shouldn't close off other possibilities, so that a possible future requirement won't need a breaking change to accommodate it.

bdm310:
`default_group` can be specified at the page and lvgl levels. My intent was for the lvgl level to then be the default for every page, if specified. Then if you desire the existing behavior, you just have to specify `default_config` once at the top level. I appear to have missed actually implementing this and also it isn't easy to do. I will add a config flag to disable the automatic switching behavior.

I missed testing with a top layer, some modifications needed for that to work as well. I'll drop this back to draft for now and make some changes.

I think it is at least a little important to acknowledge that the current implementation simply does not work for multi-page configurations and also that only one page can be visible at a time. Definitely agree with considering the future but LVGL doesn't even do anything if you hit enter while one of the off-screen widgets has focus.

PR #8260, opened 2025-02-17
Fix unexpected widget update behavior

### LINKED ISSUE #6759: LVGL widget updates result in unexpected behavior
### The problem

It looks like the widget specific update actions calls the `to_code` of the widget again with the new config. This new config may or may not have the same items as the initial config and they may or may not have a default specified that is different from what the widget was initially configured with. A spot check of some `to_code` seems to indicate that they weren't written to account for the possibility of being called without a full config set specified. Some of the `MODIFY_SCHEMA` were written with defaults, which will then invisibly flow in to the update.

Looking at slider, for defaults, if you update the value of the slider you are also inadvertently passing in a default `animated: true` unless you specify the `animated` property in every `lvgl.slider.update` block.

slider.py
```python
SLIDER_MODIFY_SCHEMA = cv.Schema(
    {
        cv.Optional(CONF_VALUE): lv_float,
        cv.Optional(CONF_ANIMATED, default=True): animated,
    }
)
```

Arc's `to_code` doesn't handle missing optional configs correctly. If you update the value you cannot also pass in a `group` config. This then results in the arc getting moved to the default lvgl input group regardless of whether or not you had specified a different group initially.

arc.py
```python
elif CONF_GROUP not in config:
    # For some reason arc does not get automatically added to the default group
    lv.group_add_obj(lv_expr.group_get_default(), w.obj)
```

I've confirmed both of these cases in the generated code.

The docs don't mention this possibility at all and in fact say that you can update all of the properties of these widgets even though often value is the only thing you can change.

2025.2.0-dev

Home Assistant Add-on

_No response_

ESP8266

### Board

_No response_

lvgl

### YAML Config

```yaml

```

```txt

```

### Additional information

_No response_

bdm310:

- animimg would have an unintended default `repeat_count` of `forever` when updated.
- arc would be added to whatever the current default LVGL input group is when updated.
- dropdown would have an unintended default `direction` of `BOTTOM` when updated.
- image had invalid default values for `pivot_x/y`, which would also be unintended when updating.

## Types of changes

- fixes esphome/issues/issues/6759

- esphome/esphome-docs/pull/4677

## Test Environment

- [X] Host

## Checklist:

### DIFF
--- a/esphome/components/lvgl/widgets/animimg.py
+++ b/esphome/components/lvgl/widgets/animimg.py
@@ -23,12 +23,12 @@ def lv_repeat_count(value):
 
 ANIMIMG_BASE_SCHEMA = cv.Schema(
     {
-        cv.Optional(CONF_REPEAT_COUNT, default="forever"): lv_repeat_count,
         cv.Optional(CONF_AUTO_START, default=True): cv.boolean,
     }
 )
 ANIMIMG_SCHEMA = ANIMIMG_BASE_SCHEMA.extend(
     {
+        cv.Optional(CONF_REPEAT_COUNT, default="forever"): lv_repeat_count,
         cv.Required(CONF_DURATION): lv_milliseconds,
         cv.Required(CONF_SRC): lv_image_list,
     }
@@ -36,6 +36,7 @@ ANIMIMG_SCHEMA = ANIMIMG_BASE_SCHEMA.extend(
 
 ANIMIMG_MODIFY_SCHEMA = ANIMIMG_BASE_SCHEMA.extend(
     {
+        cv.Optional(CONF_REPEAT_COUNT): lv_repeat_count,
         cv.Optional(CONF_DURATION): lv_milliseconds,
         cv.Optional(CONF_SRC): lv_image_list,
     }
--- a/esphome/components/lvgl/widgets/animimg.py
+++ b/esphome/components/lvgl/widgets/animimg.py
@@ -23,12 +23,12 @@ def lv_repeat_count(value):
 
 ANIMIMG_BASE_SCHEMA = cv.Schema(
     {
+        cv.Optional(CONF_REPEAT_COUNT, default="forever"): lv_repeat_count,
         cv.Optional(CONF_AUTO_START, default=True): cv.boolean,
     }
 )
 ANIMIMG_SCHEMA = ANIMIMG_BASE_SCHEMA.extend(
     {
-        cv.Optional(CONF_REPEAT_COUNT, default="forever"): lv_repeat_count,
         cv.Required(CONF_DURATION): lv_milliseconds,
         cv.Required(CONF_SRC): lv_image_list,
     }
@@ -36,7 +36,6 @@ ANIMIMG_SCHEMA = ANIMIMG_BASE_SCHEMA.extend(
 
 ANIMIMG_MODIFY_SCHEMA = ANIMIMG_BASE_SCHEMA.extend(
     {
-        cv.Optional(CONF_REPEAT_COUNT): lv_repeat_count,
         cv.Optional(CONF_DURATION): lv_milliseconds,
         cv.Optional(CONF_SRC): lv_image_list,
     }
--- a/esphome/components/lvgl/widgets/arc.py
+++ b/esphome/components/lvgl/widgets/arc.py
@@ -67,10 +67,11 @@ class ArcType(NumberType):
             lv.arc_set_mode(w.obj, literal(config[CONF_MODE]))
             lv.arc_set_change_rate(w.obj, config[CONF_CHANGE_RATE])
 
-        if config.get(CONF_ADJUSTABLE) is False:
+        adjustable = config.get(CONF_ADJUSTABLE)
+        if adjustable is False:
             lv_obj.remove_style(w.obj, nullptr, literal("LV_PART_KNOB"))
             w.clear_flag("LV_OBJ_FLAG_CLICKABLE")
-        elif CONF_GROUP not in config:
+        elif adjustable is True and CONF_GROUP not in config:
             # For some reason arc does not get automatically added to the default group
             lv.group_add_obj(lv_expr.group_get_default(), w.obj)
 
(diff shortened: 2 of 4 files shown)


esphome/components/lvgl/widgets/animimg.py
```diff
@@ -23,19 +23,20 @@ def lv_repeat_count(value):
 
 ANIMIMG_BASE_SCHEMA = cv.Schema(
     {
-        cv.Optional(CONF_REPEAT_COUNT, default="forever"): lv_repeat_count,
```
clydebarrow: The provision of a default repeat count for both creation and update was deliberate and I'm not convinced anything needs to change except perhaps a documentation update to make that clear.

esphome/components/lvgl/widgets/arc.py
```diff
@@ -67,10 +67,11 @@ async def to_code(self, w: Widget, config):
...
 
-        if config.get(CONF_ADJUSTABLE) is False:
+        adjustable = config.get(CONF_ADJUSTABLE)
+        if adjustable is False:
             lv_obj.remove_style(w.obj, nullptr, literal("LV_PART_KNOB"))
             w.clear_flag("LV_OBJ_FLAG_CLICKABLE")
-        elif CONF_GROUP not in config:
+        elif adjustable is True and CONF_GROUP not in config:
             # For some reason arc does not get automatically added to the default group
             lv.group_add_obj(lv_expr.group_get_default(), w.obj)
```
clydebarrow:
```
        if CONF_ADJUSTABLE in config:
            if not config[CONF_ADJUSTABLE]:
                lv_obj.remove_style(w.obj, nullptr, literal("LV_PART_KNOB"))
                w.clear_flag("LV_OBJ_FLAG_CLICKABLE")
            elif CONF_GROUP not in config:
                # For some reason arc does not get automatically added to the default group
                lv.group_add_obj(lv_expr.group_get_default(), w.obj)
```

esphome/components/lvgl/widgets/img.py
```diff
@@ -22,8 +22,8 @@
 
 BASE_IMG_SCHEMA = cv.Schema(
     {
-        cv.Optional(CONF_PIVOT_X, default="50%"): size,
-        cv.Optional(CONF_PIVOT_Y, default="50%"): size,
```
clydebarrow, line 26: A default is needed here since the LVGL default is not what most people would expect. Again I think clear documentation would be sufficient.

esphome/components/lvgl/widgets/dropdown.py
```diff
@@ -36,20 +36,21 @@
         cv.Optional(CONF_SYMBOL): lv_text,
         cv.Exclusive(CONF_SELECTED_INDEX, CONF_SELECTED_TEXT): lv_int,
         cv.Exclusive(CONF_SELECTED_TEXT, CONF_SELECTED_TEXT): lv_text,
-        cv.Optional(CONF_DIR, default="BOTTOM"): DIRECTIONS.one_of,
```
clydebarrow, line 39: As with animimg I'm not convinced there's a strong argument to change this.

clydebarrow (changes requested):

esphome/components/lvgl/widgets/img.py
bdm310: Sorry, I could've added more detail to the initial message. This does not work as is, I get invalid conversion from const char* to lv_coord_t when specifying an angle without x/y in the yaml. I think this could get fixed a different way to allow % to work but also the docs already say the units are specifically pixels. Additionally, the LVGL default is already the center of the image?

I've verified all of this, the proposed change results in the image rotating about the center as I would expect when I give an angle with no x/y.

esphome/components/lvgl/widgets/dropdown.py
bdm310: I configure the direction as RIGHT. Dropdown opens to the right.
I update the list. Dropdown now opens to the bottom even though I've said nothing about changing the direction.

Documentation could describe this but I don't think anyone would expect their configured direction to just go away?

esphome/components/lvgl/widgets/animimg.py
bdm310: Interesting. If I configure an object with a specific repeat count I would not expect the next update to make the repeat count infinite? Can definitely make it a doc clarification but that's not the behavior I would want/expect.

I expect when I call an "update" with specific parameters that only those parameters are getting touched.

esphome/components/lvgl/widgets/img.py
clydebarrow: Ah, there are clearly some bugs in there. The other problem is that in LVGL 8.x the angle on an image doesn't seem to work properly anyway, at least in some cases.

esphome/components/lvgl/widgets/dropdown.py
clydebarrow: There's certainly more of an argument in this case than for animimg. And looking at my code that was probably intended.

esphome/components/lvgl/widgets/img.py
bdm310: I'm testing purely on whatever the dev branch pulls in here. Angles work fine in my testing, albeit the input angle is 10x and the docs need clarification. I do forget if I put that in the doc PR.

esphome/components/lvgl/widgets/dropdown.py
bdm310: We can disagree on animimg, I can roll that one back.

esphome/components/lvgl/widgets/img.py
bdm310: Ignore my 10x comment, that was from when I enabled lambda for angles in testing. Docs and function are fine.

esphome/components/lvgl/widgets/animimg.py
bdm310: Reverted

bdm310:
@clydebarrow , any additional feedback here?

clydebarrow (approved):

PR #8263, opened 2025-02-18
dashboard: Implement automatic ping fallback

bdraco:

Automatically fallback to ping when mDNS is broken or the device has mdns disabled instead of having the user have to manually configure this.  As before, ping only happens when the dashboard is open so mDNS is still preferred since we are always browsing for devices when the dashboard is running.

This change also improves the situation for the mixed use case where some devices are reachable different methods:

- mDNS
- Ping
- MQTT

Previously users would have to give up mDNS to enable ping since it was one or the other.

Once we see a device as reachable via a specific method we will keep using that method to check device until it becomes unreachable by that method, at which point we will accept state from any method that can reach it.

# Breaking change

It is no longer necessary to configure `status_use_ping` as fallback to ping will automatically happen when mDNS is not available or disabled for a device.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/dashboard/dns.py
+++ b/esphome/dashboard/dns.py
@@ -18,7 +18,7 @@ RESOLVE_TIMEOUT = 2.5
 async def _async_resolve_wrapper(hostname: str) -> list[str] | Exception:
     """Wrap the icmplib async_resolve function."""
     with suppress(ValueError):
-        return ip_address(hostname)
+        return [ip_address(hostname)]
     try:
         async with async_timeout(RESOLVE_TIMEOUT):
             return await async_resolve(hostname)
--- a/esphome/dashboard/dns.py
+++ b/esphome/dashboard/dns.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import asyncio
+from contextlib import suppress
+from ipaddress import ip_address
 import sys
 
 from icmplib import NameLookupError, async_resolve
@@ -10,11 +12,15 @@ if sys.version_info >= (3, 11):
 else:
     from async_timeout import timeout as async_timeout
 
+RESOLVE_TIMEOUT = 2.5
+
 
 async def _async_resolve_wrapper(hostname: str) -> list[str] | Exception:
     """Wrap the icmplib async_resolve function."""
+    with suppress(ValueError):
+        return ip_address(hostname)
     try:
-        async with async_timeout(2):
+        async with async_timeout(RESOLVE_TIMEOUT):
             return await async_resolve(hostname)
     except (asyncio.TimeoutError, NameLookupError, UnicodeError) as ex:
         return ex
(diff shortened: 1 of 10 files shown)


esphome/dashboard/entries.py
```diff
@@ -119,6 +137,55 @@ async def _async_set_state(self, entry: DashboardEntry, state: EntryState) -> No
         """Set the state for an entry."""
         self.async_set_state(entry, state)
 
+    def set_state_if_online_or_source(
```
bdraco, line 140: All of the checks must happen in the event loop so we end up with sync versions here for the mqtt status thread

esphome/dashboard/status/mdns.py
```diff
@@ -58,13 +74,26 @@ async def async_refresh_hosts(self):
                 result = bool(address_list)
                 host_mdns_state[name] = result
                 for entry in poll_names[name]:
-                    entries.async_set_state(entry, bool_to_entry_state(result))
+                    self._async_set_state(entry, result)
+
+    def _async_set_state(self, entry: DashboardEntry, result: bool | None) -> None:
+        """Set the state of an entry."""
+
```
bdraco:
```suggestion
```

esphome/zeroconf.py
```diff
@@ -168,11 +174,8 @@ def update_device_mdns(self, node_name: str, version: str):
 
 def _make_host_resolver(host: str) -> HostResolver:
     """Create a new HostResolver for the given host name."""
-    name = host.partition(".")[0]
-    info = HostResolver(
```
bdraco, line 172: HostResolver is no longer used

esphome/dashboard/dns.py
```diff
@@ -10,11 +12,15 @@
 else:
     from async_timeout import timeout as async_timeout
 
+RESOLVE_TIMEOUT = 2.5
```
bdraco:
```suggestion
RESOLVE_TIMEOUT = 3.0
```

esphome/zeroconf.py
```diff
@@ -16,15 +22,6 @@
 _BACKGROUND_TASKS: set[asyncio.Task] = set()
 
 
-class HostResolver(ServiceInfo):
```
bdraco, line 19: zeroconf now has a built-in resolver class called `AddressResolver`. Better to use that since it works around the issue fixed by https://github.com/espressif/esp-protocols/pull/758 which will likely still be present in many devices

balloob (approved):

PR #8268, opened 2025-02-18
Use the process CPU count to determine how many children to create

bdraco:

Python 3.13+ introduced os.process_cpu_count() which is better aligned with want we want here. For older Python we will use `os.cpu_count` which is what `multiprocessing` called under the hood anyways.

https://docs.python.org/3/library/os.html#os.process_cpu_count

Additionally this avoids bring in multiprocessing which previously increased the startup time

reduction in `importtime-waterfall esphome.__main__`

```
        multiprocessing (82)
          multiprocessing.context (188)
            multiprocessing.process (221)
              signal (195)
            multiprocessing.reduction (116)
              pickle (293)
                struct (45)
                  _struct (209)
                _compat_pickle (101)
                _pickle (241)
              socket (405)
                _socket (320)
                array (205)
```

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/core/config.py
+++ b/esphome/core/config.py
@@ -94,6 +94,10 @@ def valid_project_name(value: str):
 
 
 def get_usable_cpu_count() -> int:
+    """Return the number of CPUs that can be used for processes.
+    On Python 3.13+ this is the number of CPUs that can be used for processes.
+    On older Python versions this is the number of CPUs.
+    """
     return (
         os.process_cpu_count() if hasattr(os, "process_cpu_count") else os.cpu_count()
     )
--- a/esphome/core/config.py
+++ b/esphome/core/config.py
@@ -1,5 +1,4 @@
 import logging
-import multiprocessing
 import os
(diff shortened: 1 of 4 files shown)


esphome/core/config.py
```diff
@@ -94,10 +93,15 @@ def valid_project_name(value: str):
     return value
 
 
+def get_usable_cpu_count() -> int:
+    return (
```
bdraco, line 101:
```suggestion
    """Return the number of CPUs that can be used for processes.
    On Python 3.13+ this is the number of CPUs that can be used for processes.
    On older Python versions this is the number of CPUs.
    """
    return (
```

jesserockz (approved):

PR #8108, opened 2025-01-18
Add Ruff to CI

stellar-aria:
Adds a ruff-format check to the CI, as an alternative to the black formatter as part of the general migration to Ruff

Not sure if this is the best way to go about this, as ideally we want file-based granularity over whole-repo (i.e. a _file_ passes either black or ruff). There's probably some way to do this, but I'm not entirely sure how...

EDIT: I've tried to implement this granularity by running either black or ruff on each file via parallel in the action

CAVEATS: We can no longer display the changes to be made, but since Ruff is running via the pre-commit hook anyways, this shouldn't be a huge issue?

PR #8276, opened 2025-02-19
Ruff format for CI

stellar-aria:
Switch the CI formatter from Black-format to Ruff. Supercedes #8108

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,6 +1,6 @@
 pylint==3.2.7
 flake8==7.0.0  # also change in .pre-commit-config.yaml when updating
-black==24.4.2  # also change in .pre-commit-config.yaml when updating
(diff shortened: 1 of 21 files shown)


bdraco (approved):

bdraco:
Thanks @stellar-aria

requirements_test.txt
```diff
@@ -1,6 +1,6 @@
 pylint==3.2.7
 flake8==7.0.0  # also change in .pre-commit-config.yaml when updating
-black==24.4.2  # also change in .pre-commit-config.yaml when updating
```
syssi, line 3: FYI: https://github.com/esphome/esphome/blob/dev/script/lint-python#L72 doesn't work anymore if `requirements_test.txt` is used.

stellar-aria: I've made a new PR at #8294 if you'd like to check that over

PR #8277, opened 2025-02-19
[fan] fix initial FanCall to properly set speed

dhewg:

According to https://developers.home-assistant.io/docs/core/entity/fan/#preset-modes preset modes and manual speed are mutually exclusive:
`Manually setting a speed must disable any set preset mode`.

So if a preset mode is active, don't set the manual speed to 100%, as that disables the preset mode (e.g. `Sleep`) and puts the device in full power (e.g. opposite of sleep ;)

Noticed while developing a fan component locally, and this fixes it.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


dhewg:
@bdraco It's been a while, but you pointed out the very same thing on the initial preset mode PR:
`The one big thing that looks missing is that when the user manually sets a speed, it should override and clear the preset mode`
https://github.com/esphome/esphome/pull/5694#issuecomment-1814558963

So maybe you have some feedback for this PR?

The comment just after that one reads:
`From HA, any manual change to speed, direction or oscillation causes a call to be made with an empty preset_mode which clears the preset.`

From my testing, HA doesn't do that, and neither does the esphome fan component.
But my fan component does. And then you're running into what this issue tries to solve

dhewg:
Updated the fixes to be more complete.

The clang-format check fails, but the suggestions decrease the readability imho, let me know which way to go.

dhewg:
Anyone? Bueller?

bdraco:
You'll need to get the CI passing before anyone can start reviewing the PR. I know clang-format can be a bit picky sometimes, but accepting its changes is the easiest way to move things forward.

esphome/components/fan/fan.cpp
```diff
@@ -41,16 +41,35 @@ void FanCall::perform() {
...
-      this->speed_ = traits.supported_speed_count();
+    // https://developers.home-assistant.io/docs/core/entity/fan/#preset-modes
+    // "Manually setting a speed must disable any set preset mode"
+    this->preset_mode_.clear();
+  }
+
+  if (!this->preset_mode_.empty()) {
+    const auto &preset_modes = traits.supported_preset_modes();
+    if (preset_modes.find(this->preset_mode_) == preset_modes.end()) {
+      ESP_LOGW(TAG, "'%s' - This fan does not support preset mode '%s'!", this->parent_.get_name().c_str(),
```
bdraco: https://developers.esphome.io/architecture/logging/

esphome/components/fan/fan.cpp
```diff
@@ -41,16 +41,35 @@ void FanCall::perform() {
...
+  // when turning on...
+  if (!this->parent_.state && this->binary_state_.has_value() &&
+      *this->binary_state_
+      // ..,and no preset mode will be active...
+      && this->preset_mode_.empty() &&
+      this->parent_.preset_mode.empty()
+      // ...and neither current nor new speed is available...
+      && traits.supports_speed() && this->parent_.speed == 0 && !this->speed_.has_value()) {
+    // ...set speed to 100%
+    this->speed_ = traits.supported_speed_count();
```
bdraco, line 69: turning on to max speed might not be desirable. Usually fan integration turn on to 50% by default when no speed is specified

bdraco: Thats existing behavior though....

bdraco (approved):

bdraco:
@dhewg I wrote some tests for this. Please double check that the behavior works as intended

dhewg:
Nice, thanks for the fixups @bdraco !

Confirmed, the fixes are still working as intended.

Tested with:
- turn off the device (a purifier in my case)
- enable the 'sleep' preset mode
- turn on the device

Without these patches the fan will go to 100% (i.e. loud).
With these patches the fan will be on, but correctly in the chosen 'sleep' preset (i.e. silent).

Thanks!

PR #8282, opened 2025-02-19
Add support for deep sleep on ESP32-H2 and add ext1 wakeup reason to make the H2 useable.

### LINKED ISSUE #6752: deep sleep component not working with esp32-h2
### The problem

Esphome does not compile with esp-idf when using the ESP32-H2 as target and using the deep sleep component. It looks like there are some definitions missing in `deep_sleep_esp32.cpp` that cause incompatibility with esp-idf code.

`src/esphome/components/deep_sleep/deep_sleep_esp32.cpp: In member function 'void esphome::deep_sleep::DeepSleepComponent::deep_sleep_()':
src/esphome/components/deep_sleep/deep_sleep_esp32.cpp:76:5: error: 'esp_sleep_enable_ext0_wakeup' was not declared in this scope; did you mean 'esp_sleep_enable_ext1_wakeup'?
   76 |     esp_sleep_enable_ext0_wakeup(gpio_num_t(this->wakeup_pin_->get_pin()), level);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     esp_sleep_enable_ext1_wakeup
src/esphome/components/deep_sleep/deep_sleep_esp32.cpp:83:5: error: 'esp_sleep_enable_touchpad_wakeup' was not declared in this scope; did you mean 'esp_sleep_enable_uart_wakeup'?
   83 |     esp_sleep_enable_touchpad_wakeup();
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     esp_sleep_enable_uart_wakeup
src/esphome/components/deep_sleep/deep_sleep_esp32.cpp:84:25: error: 'ESP_PD_DOMAIN_RTC_PERIPH' was not declared in this scope; did you mean 'ESP_PD_DOMAIN_RC_FAST'?
   84 |     esp_sleep_pd_config(ESP_PD_DOMAIN_RTC_PERIPH, ESP_PD_OPTION_ON);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~
      |                         ESP_PD_DOMAIN_RC_FAST
*** [.pioenvs/zb-example-h2/src/esphome/components/deep_sleep/deep_sleep_esp32.cpp.o] Error 1
========================= [FAILED] Took 13.18 seconds =========================`

2024.12.4

Home Assistant Add-on

2025.2

ESP32

### Board

esp32-h2-devkitm-1

deep_sleep

### YAML Config

```yaml
esphome:
  name: zb-example-h2

deep_sleep:
  run_duration: 10s
  sleep_duration: 10min


esp32:
  board: esp32-h2-devkitm-1
  framework:
    type: esp-idf
    version: latest
```

```txt

```

### Additional information

_No response_

baal86:

Add support for deep sleep on the ESP32-H2. Since the H2 variant does not support GPIO wakeup but only EXT1, EXT1 wakeup reason is added as well so that the user can implement an IO triggered wakeup.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6752

- esphome/esphome-docs#4686

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/deep_sleep/deep_sleep_component.h
+++ b/esphome/components/deep_sleep/deep_sleep_component.h
@@ -48,7 +48,7 @@ struct WakeupCauseToRunDuration {
   uint32_t touch_cause;
   // Run duration if woken up by GPIO pins.
   uint32_t gpio_cause;
-  // Run duration if woken up by ext1 pins.
+  // Run duration if woken up by EXT1 pins.
   uint32_t ext1_cause;
 };
 
--- a/esphome/components/deep_sleep/deep_sleep_component.h
+++ b/esphome/components/deep_sleep/deep_sleep_component.h
@@ -48,6 +48,8 @@ struct WakeupCauseToRunDuration {
   uint32_t touch_cause;
   // Run duration if woken up by GPIO pins.
   uint32_t gpio_cause;
+  // Run duration if woken up by ext1 pins.
+  uint32_t ext1_cause;
 };
 
 #endif
--- a/esphome/components/deep_sleep/deep_sleep_esp32.cpp
+++ b/esphome/components/deep_sleep/deep_sleep_esp32.cpp
@@ -69,7 +69,7 @@ bool DeepSleepComponent::prepare_to_sleep_() {
 }
 
 void DeepSleepComponent::deep_sleep_() {
-#if !defined(USE_ESP32_VARIANT_ESP32C3) && !defined(USE_ESP32_VARIANT_ESP32C6) && !defined(USE_ESP32_VARIANT_ESP32H2) 
+#if !defined(USE_ESP32_VARIANT_ESP32C3) && !defined(USE_ESP32_VARIANT_ESP32C6) && !defined(USE_ESP32_VARIANT_ESP32H2)
   if (this->sleep_duration_.has_value())
     esp_sleep_enable_timer_wakeup(*this->sleep_duration_);
   if (this->wakeup_pin_ != nullptr) {
@@ -88,7 +88,7 @@ void DeepSleepComponent::deep_sleep_() {
     esp_sleep_pd_config(ESP_PD_DOMAIN_RTC_PERIPH, ESP_PD_OPTION_ON);
   }
 #endif
-#if defined(USE_ESP32_VARIANT_ESP32H2) 
+#if defined(USE_ESP32_VARIANT_ESP32H2)
   if (this->sleep_duration_.has_value())
     esp_sleep_enable_timer_wakeup(*this->sleep_duration_);
 
--- a/esphome/components/deep_sleep/deep_sleep_esp32.cpp
+++ b/esphome/components/deep_sleep/deep_sleep_esp32.cpp
@@ -13,6 +13,7 @@ optional<uint32_t> DeepSleepComponent::get_run_duration_() const {
     switch (wakeup_cause) {
       case ESP_SLEEP_WAKEUP_EXT0:
       case ESP_SLEEP_WAKEUP_EXT1:
+        return this->wakeup_cause_to_run_duration_->ext1_cause;
       case ESP_SLEEP_WAKEUP_GPIO:
         return this->wakeup_cause_to_run_duration_->gpio_cause;
       case ESP_SLEEP_WAKEUP_TOUCHPAD:
@@ -30,7 +31,9 @@ void DeepSleepComponent::set_wakeup_pin_mode(WakeupPinMode wakeup_pin_mode) {
 
 #if !defined(USE_ESP32_VARIANT_ESP32C3) && !defined(USE_ESP32_VARIANT_ESP32C6)
 void DeepSleepComponent::set_ext1_wakeup(Ext1Wakeup ext1_wakeup) { this->ext1_wakeup_ = ext1_wakeup; }
+#endif
 
+#if !defined(USE_ESP32_VARIANT_ESP32C3) && !defined(USE_ESP32_VARIANT_ESP32C6) && !defined(USE_ESP32_VARIANT_ESP32H2)
 void DeepSleepComponent::set_touch_wakeup(bool touch_wakeup) { this->touch_wakeup_ = touch_wakeup; }
 #endif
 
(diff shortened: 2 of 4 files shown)


esphome/components/deep_sleep/deep_sleep_esp32.cpp
```diff
@@ -13,6 +13,7 @@ optional<uint32_t> DeepSleepComponent::get_run_duration_() const {
     switch (wakeup_cause) {
       case ESP_SLEEP_WAKEUP_EXT0:
       case ESP_SLEEP_WAKEUP_EXT1:
+        return this->wakeup_cause_to_run_duration_->ext1_cause;
```
DjordjeMandic, line 16:
```suggestion
#if defined(USE_ESP32_VARIANT_ESP32H2)
        return this->wakeup_cause_to_run_duration_->ext1_cause;
#endif
```
I believe that without this it would be breaking change.

esphome/components/deep_sleep/deep_sleep_esp32.cpp
```diff
@@ -47,6 +50,7 @@ void DeepSleepComponent::dump_config_platform_() {
                   this->wakeup_cause_to_run_duration_->default_cause);
     ESP_LOGCONFIG(TAG, "  Touch Wakeup Run Duration: %" PRIu32 " ms", this->wakeup_cause_to_run_duration_->touch_cause);
     ESP_LOGCONFIG(TAG, "  GPIO Wakeup Run Duration: %" PRIu32 " ms", this->wakeup_cause_to_run_duration_->gpio_cause);
+    ESP_LOGCONFIG(TAG, "  EXT1 Wakeup Run Duration: %" PRIu32 " ms", this->wakeup_cause_to_run_duration_->ext1_cause);
```
DjordjeMandic, line 53:
```suggestion
#if defined(USE_ESP32_VARIANT_ESP32H2)
    ESP_LOGCONFIG(TAG, "  EXT1 Wakeup Run Duration: %" PRIu32 " ms", this->wakeup_cause_to_run_duration_->ext1_cause);
#endif
```

DjordjeMandic (changes requested):

tests/component_tests/deep_sleep/test_deep_sleep.py
```diff
@@ -40,5 +40,6 @@ def test_deep_sleep_run_duration_dictionary(generate_main):
         "    .default_cause = 10000,\n"
         "    .touch_cause = 10000,\n"
         "    .gpio_cause = 30000,\n"
+        "    .ext1_cause = 10000,\n"
```
DjordjeMandic, line 43:
```suggestion
```

DjordjeMandic:
Hey @baal86
Could you take a look at https://github.com/baal86/esphome/pull/1

esphome/components/deep_sleep/deep_sleep_component.h
```diff
@@ -48,6 +48,8 @@ struct WakeupCauseToRunDuration {
   uint32_t touch_cause;
   // Run duration if woken up by GPIO pins.
   uint32_t gpio_cause;
+  // Run duration if woken up by ext1 pins.
```
DjordjeMandic:
```suggestion
  // Run duration if woken up by EXT1 pins.
```

DjordjeMandic (changes requested):

DjordjeMandic (commented):
This introduces breaking change. It will work for H2 but OLD GPIO time becomes useless because in deep sleep GPIO is not available, only in light sleep. Only EXT and timer.

(that's what i read from espressif idf docs).

Sadly this whole component needs breaking rework in order to comply with all esp32 variants correctly.

Switch statement used to fall thru EXT0/EXT1 to GPIO. GPIO actually had user provided run time.

Existing configurations that get woken up by EXT0 or EXT1 will default to EXT1 run time instead of GPIO.

esphome/components/deep_sleep/deep_sleep_esp32.cpp
DjordjeMandic, line 16: This is breaking change IMO

baal86: I guess the H2 variant has limited usefulness for esphome as a whole at the moment since ZIgbee support is only through external components. Once/if this gains traction it is very valuable due to it's much lower power consumption compared to C6.

I'm a hardware person so I have a hard time judging the quality of the current deep sleep component. My changes certainly are hacky since the H2 differs so much from the S and C series parts. In the end it is up to people more involved with esphome to decide if the H2 can be added or if a more sensible approach would be to consider a more in-depth rework of deep_sleep first (I'm not qualified to take that on) and for me to create an external component, deep_sleep_h2, to have a working solution for that part.

DjordjeMandic: You can always import deep sleep component from this pull request as external component. Just I would do the following:
```suggestion
#if defined(USE_ESP32_VARIANT_ESP32H2)
        return this->wakeup_cause_to_run_duration_->ext1_cause;
#endif
```
and also add whole EXT1 addition under same if defined preprocessor directive and most importatly mention in docs that EXT1 is used only on ESP32-H2.

baal86: I think I get your point now! If EXT1 wakeup duration is only available for H2 this does not break anything else. As I read the Espressif docs at least the C3 and C6 also support EXT1 wakeup so the wakeup cause would be useful which was my original thought. The day job calls but I'll get to it this weekend. TODO for me:

- Make EXT1 wakeup duration only available to H2 by preprocessor.
- Update docs that EXT1 wakeup only on H2.
- Make a new PR for the docs (github can't change source branch, only destination)
- Maybe have another look at the __init__.py

I am already using deep sleep on H2 as external component from my fork, so not in a hurry from my side.

DjordjeMandic: Problem is that EXT0 wakeup is currently(in dev) being used on supported variants with GPIO run time. Your change would result into using EXT1 run time. 

You need to do the following to make it non breaking.
```cpp
      case ESP_SLEEP_WAKEUP_EXT0:
#if defined(USE_ESP32_VARIANT_ESP32H2)
        return this->wakeup_cause_to_run_duration_->gpio_cause;
#endif
      case ESP_SLEEP_WAKEUP_EXT1:
#if defined(USE_ESP32_VARIANT_ESP32H2)
        return this->wakeup_cause_to_run_duration_->ext1_cause;
#endif
      case ESP_SLEEP_WAKEUP_GPIO:
        return this->wakeup_cause_to_run_duration_->gpio_cause;
```

baal86: I think I might just not add the ext1_cause and use the gpio_cause instead.

I mean it is ugly from the hardware perspective since EXT0 and EXT1 are quite different (EXT1 is much more constrained) and hiding this difference from the user could cause problems later but I will experiment with it and maybe then this PR (or a new one since scope will be different) might be a trivial patch just fixing H2.

baal86:
Closed this pull request. Thank you @DjordjeMandic for the very valuable discussion. I will re-submit a new PR which does the bare minimum to support the H2 using EXT1 wakeup as gpio_cause.

PR #8284, opened 2025-02-20
[online_image] Add WEBP animation support

lanrat:

Adds support for both static and animated WEBP images to `online_image`.

## Types of changes

- [ ] Other

- esphome/esphome-docs#4688

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
online_image:
  - url: "https://example.com/example.webp"
    format: webp
    id: online_image_webp
    type: RGB
```

## Checklist:


esphome/components/online_image/online_image.cpp
```diff
@@ -90,7 +94,10 @@ size_t OnlineImage::resize_(int width_in, int height_in) {
   this->buffer_width_ = width;
   this->buffer_height_ = height;
   this->width_ = width;
-  ESP_LOGV(TAG, "New size: (%d, %d)", width, height);
+  this->animation_frame_count_ = frames;
+  this->buffer_frame_size_ = new_size / frames;
+  this->current_frame_ = 0;
+  ESP_LOGV(TAG, "New size: (%d, %d, %d)", width, height, frames);
```
guillempages:
```suggestion
  ESP_LOGV(TAG, "New size: (%d, %d). Frames: %d", width, height, frames);
```

esphome/components/online_image/webp_image.cpp
```diff
@@ -0,0 +1,125 @@
...
+  static int ix_a = 3;
+  static int channels = 4;
+
+  for (unsigned int y = 0; y < height; y++) {
+    for (unsigned int x = 0; x < width; x++) {
+      const uint8_t *p = &pix[(y * width + x) * channels];
+      uint8_t r = p[ix_r];
+      uint8_t g = p[ix_g];
+      uint8_t b = p[ix_b];
+      Color color(r, g, b, p[ix_a]);
```
guillempages: Any specific reason, why you define variables for r,g and b, but not for a?

esphome/components/online_image/webp_image.cpp
```diff
@@ -0,0 +1,125 @@
...
+  webp_data.size = size;
+
+  WebPAnimDecoderOptions decoder_options;
+  WebPAnimDecoderOptionsInit(&decoder_options);
+  decoder_options.color_mode = MODE_RGBA;
+
+  this->decoder_ = WebPAnimDecoderNew(&webp_data, &decoder_options);
+  if (this->decoder_ == NULL) {
+    ESP_LOGE(TAG, "Could not create WebP decoder");
+    return DECODE_ERROR_UNSUPPORTED_FORMAT;
```
guillempages, line 71: Just asking: If the "WevPAnimDecoderNew" returns null; does it mean it is the wrong format, or not enough memory? Or a different error? Is there a way of checking the actual error and return it accordingly?

esphome/components/online_image/webp_image.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  if (animation_.frame_count == 0) {
+    ESP_LOGE(TAG, "Animation has 0 frames");
+    WebPAnimDecoderDelete(this->decoder_);
+    this->decoder_ = NULL;
+    return 0;
+  }
+
+  // iterate over all frames
+  for (uint frame = 0; frame < animation_.frame_count; frame++) {
```
guillempages: This could cause the watchdog to trigger...

Is there a way of splitting this loop, so that the decode method is being called repeatedly by the outer loop (i.e. the component's loop), instead of adding an explicit loop here?

E.g. like the PNG decoder is doing

esphome/components/online_image/__init__.py
```diff
@@ -25,7 +28,7 @@
     CONF_URL,
 )
 
-AUTO_LOAD = ["image"]
+AUTO_LOAD = ["image", "animation"]
```
guillempages, line 29: I'm not sure that forcing autoload of animation for all online_images is a good idea :-(

The "offline" image and animation components are separate components, that can't be used interchangeably...
I'd rather keep "online_image" for single images and not animations.

esphome/components/online_image/__init__.py
```diff
@@ -83,12 +95,15 @@ def actions(self):
     for x in (
         BMPFormat(),
         JPEGFormat(),
+        WEBPFormat(),
         PNGFormat(),
```
guillempages:
```suggestion
        PNGFormat(),
        WEBPFormat(),
```

esphome/components/online_image/__init__.py
```diff
@@ -69,6 +72,15 @@ def actions(self):
         cg.add_library("JPEGDEC", None, "https://github.com/bitbank2/JPEGDEC#ca1e0f2")
 
 
+class WEBPFormat(Format):
```
guillempages: Please, keep them sorted alphabetically

esphome/components/online_image/online_image.cpp
```diff
@@ -12,6 +12,9 @@ static const char *const TAG = "online_image";
 #ifdef USE_ONLINE_IMAGE_JPEG_SUPPORT
 #include "jpeg_image.h"
 #endif
+#ifdef USE_ONLINE_IMAGE_WEBP_SUPPORT
+#include "webp_image.h"
+#endif
 #ifdef USE_ONLINE_IMAGE_PNG_SUPPORT
 #include "png_image.h"
 #endif
```
guillempages:
```suggestion
#ifdef USE_ONLINE_IMAGE_PNG_SUPPORT
#include "png_image.h"
#endif
#ifdef USE_ONLINE_IMAGE_WEBP_SUPPORT
#include "webp_image.h"
#endif
```

esphome/components/online_image/online_image.cpp
```diff
@@ -117,6 +124,11 @@ void OnlineImage::update() {
       accept_mime_type = "image/jpeg";
       break;
 #endif  // USE_ONLINE_IMAGE_JPEG_SUPPORT
+#ifdef USE_ONLINE_IMAGE_WEBP_SUPPORT
```
guillempages: Please keep the alphabetic sorting

esphome/components/online_image/online_image.cpp
```diff
@@ -166,6 +178,12 @@ void OnlineImage::update() {
     this->decoder_ = esphome::make_unique<JpegDecoder>(this);
   }
 #endif  // USE_ONLINE_IMAGE_JPEG_SUPPORT
+#ifdef USE_ONLINE_IMAGE_WEBP_SUPPORT
```
guillempages: webp after png

guillempages (commented):
I'm not sure I like the idea of mixing images and animations in a single component for "online", but keeping them separate for "offline".

esphome/components/online_image/webp_image.cpp
lanrat: Not particularly. I'll update this without the extra variables.

lanrat:
@guillempages thanks for taking the time to look at this PR.

In regards to extending `online_image` to support animations, what are your suggestions if you think it should be approached another way?

I was thinking I could add a yaml property `animation: true/false` that when set to false (default) would not load any of the animation classes, but when set to true does.

lanrat:
Tangential to this PR, I was wondering if it might be wise to break image parsing out of  `online_image` into its own component to be used by `online_image` or others. The new image parsing component could handle generating an image object from a buffer of bytes. When used together you would get the same effect as `online_image` today.

That way images could be generated from other sources in the future, such as from the HASS API, or an SD card.

clydebarrow:
> I'm not sure I like the idea of mixing images and animations in a single component for "online", but keeping them separate for "offline".

So you're thinking to merge the `animation` component into `image`? That makes a lot of sense. Maybe `online_image` should also be merged in, so `image` could represent single or multiple image frames, static or online.

esphome/components/online_image/webp_image.cpp
lanrat: `WebPAnimDecoderNew` can return NULL due to a memory or parsing error.

There does not appear to be a straightforward way to determine the difference between the two errors.

documentation: https://developers.google.com/speed/webp/docs/container-api#webpanimdecodernew

esphome/components/online_image/webp_image.cpp
lanrat: This is a good idea and I'd like to implement it.

Looking over the PNG decoder code, it looks like its using the pngle library's built in support for using a callback. Unfortunately the current libwebp library does not offer that. Is there a straightforward way to break it out into an ESPHome task?

lanrat:
> > I'm not sure I like the idea of mixing images and animations in a single component for "online", but keeping them separate for "offline".
> 
> So you're thinking to merge the `animation` component into `image`? That makes a lot of sense. Maybe `online_image` should also be merged in, so `image` could represent single or multiple image frames, static or online.

`animation` currently extends `image` and just adds some logic to seek into the image's `data_start` buffer for different frames. If an `animation` has 1 frame, it acts identically to `image`.

esphome/components/online_image/__init__.py
lanrat: Is there a downside other than some additional overhead at compile time?

lanrat: If there is, I can add a toggle to extend `animation` in the yaml as proposed here: https://github.com/esphome/esphome/pull/8284#issuecomment-2677004162

esphome/components/online_image/webp_image.cpp
lanrat: I added `decode_loop()` to `ImageDecoder` to facilitate image decoders getting called by the components `loop()` as needed.

guillempages:
> Tangential to this PR, I was wondering if it might be wise to break image parsing out of `online_image` into its own component to be used by `online_image` or others. The new image parsing component could handle generating an image object from a buffer of bytes. When used together you would get the same effect as `online_image` today.
> 
> That way images could be generated from other sources in the future, such as from the HASS API, or an SD card.

I'm not sure how big a change this would be. My *very* long term plan was to just reuse `online_image` for images read from SD Card or similar, but haven't really started looking into that (since there is no `storage` component implemented yet).

Might be worth splitting, but I'd propose to defer that until there is really a need for it. The downloading is already offloaded to the `http_request` component anyway.

guillempages:
> @guillempages thanks for taking the time to look at this PR.
> 
> In regards to extending `online_image` to support animations, what are your suggestions if you think it should be approached another way?
> 
> I was thinking I could add a yaml property `animation: true/false` that when set to false (default) would not load any of the animation classes, but when set to true does.

I'm not decided on that :-/ Adding the property (defaulting to false) might be a good idea. But this might also set wrong expectations: people might then expect other formats to also support animations (e.g. M-JPEG)

My gut feeling tells me it would be clearer/safer to just stick to static images for the online_image component, but seeing that the needed changes to support animations are not very big, I don't really have meaningful arguments to oppose the animation support ;-)

esphome/components/online_image/__init__.py
guillempages: > Is there a downside other than some additional overhead at compile time?

Well, it's including an additional component, so it's more files to compile, and a slightly bigger binary (since more code is included).
Nevertheless it's not a huge increase anyway.

> If there is, I can add a toggle to extend animation in the yaml as proposed here: #8284 (comment)

This is probably a good idea, so that if someone really needs the animations, it can be explicitly configured, but for most people that just use simple images, there is no overhead.

esphome/components/online_image/webp_image.cpp
```diff
@@ -0,0 +1,131 @@
...
+    this->decoder_ = NULL;
+    return 0;
+  }
+
+  this->decoded_bytes_ = 0;
+  this->next_frame_ = 0;
+  return size;
+}
+
+enum DecodeError WebpDecoder::decode_loop() {
```
guillempages, line 103: Do you really need to add this decode_loop method? 
Couldn't this be handled in the normal `decode` method (returning a size smaller than the total size will cause the decode method to be called again from the online_image component anyway).

esphome/components/online_image/image_decoder.h
```diff
@@ -49,15 +50,24 @@ class ImageDecoder {
    */
   virtual int decode(uint8_t *buffer, size_t size) = 0;
 
+  /**
+   * @brief For images that may take a while to decode, such as animations
+   * call decode_loop() from the component's main loop() after decode()
+   * until is_finished() returns true
+   * @return any decoding error encountered or DECODE_ERROR_NONE
+   */
+  virtual enum DecodeError decode_loop() { return DECODE_ERROR_NONE; }
```
guillempages, line 59: See comment in the webp_image.cpp implementation of this method; but I'd rather not add this one, and make the `decode` method be called more than once. E.g. PNG images already do that; they decode in chunks while downloading.

esphome/components/online_image/webp_image.cpp
lanrat: I looked into using just the `decode()` method. But in order to do that, I would need to make the `decode()` method lie. Currently its given chunks of a download buffer and returns how much was used in order to free the buffer when done. However, in this case, after receiving the entire download buffer, we need to continue calling decode for the processing of each frame, but no more of the download buffer will be consumed. This is a problem as that is not how the logic in `online_image` works.

As is currently implemented, each invocation of `decode()` is meant to consume a part of the download buffer. In the webp case, the entire buffer will already be consumed, we just need to process the frames. In your png example, each invocation consumes more of the download buffer, which works for png, but not for webp.

And if we "consume" the entire download buffer before parsing each frame in `decode()` then `online_image` will not call `decode()` after the buffer is fully read. In order to use `decode()` this way, webp_image would need to return false information about not consuming the entire buffer so that online_image continues to call decode and hold onto the download buffer longer than needed. I do not think this is a good solution. Hence why I added `decode_loop()` which adds a lot more flexibility.

What do you think?

esphome/components/online_image/__init__.py
lanrat: I'm not convinced this change would be worth it. The animation logic is minimal. It just facilitates moving a pointer to the correct frame for the image.

esphome/components/online_image/image_decoder.h
lanrat: See thread: https://github.com/esphome/esphome/pull/8284#discussion_r1986110071

lanrat:
@guillempages where do you stand on merging this?

Are there any other changes you would like?

guillempages:
Sorry about the delay; RL has higher prio at the moment ;-)

I'm still not convinced about the animation support...
Yes, I see that the changes to support it are not big, but they do add some (small) overhead, and also would make online_image inconsistent with the "static" image/animation components (they have explicit separate components).

Also, not sure how this will play together with LVGL. AFAIK, it has its own way for defining animations, and this then would be an additional source for inconsistency.

So I'd rather only add support for single frame WEBP images at the moment, and add the animation support generically later on in a consistent way for all three "types" (LVGL, static and online).

That said, I'm the code owner of the online_image component, but not of the whole project. If @clydebarrow or Jesse are OK with this, I'd withdraw my objection.

jesserockz:
I agree with @guillempages on adding single frame WEBP images in by itself first, then looking at how we handle online_animations (new component name?) in another PR.

PR #8250, opened 2025-02-14
[sht4x] Reduce warn spam, added communication check in setup

DjordjeMandic:
# What does this implement/fix?

> Sensor should be marked offline at I2C scan without attempts to read or write to the device in the "loop" when sensor is not connected and not found in boot scan.

SSIA

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- fixes https://github.com/esphome/issues/issues/6732
- fixes https://github.com/esphome/issues/issues/6708

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8293, opened 2025-02-21
Add CUBIC CM1106 Single Beam NDIR CO2 Sensor Module

andrewjswan:

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4697

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
uart:
  - id: cm1106_uart
    rx_pin: 16
    tx_pin: 17
    baud_rate: 9600

sensor:
  - platform: cm1106
    co2:
      name: "${name} CO2"
      id: co2sensor
    id: cm1106_sensor
```

## Checklist:

## Test this PR:

```Yaml
external_components:
  - source: github://pr#8293
    components: [cm1106]
    refresh: 60s
```


esphome/components/cm1106/cm1106.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "esphome/core/component.h"
+#include "esphome/core/automation.h"
+#include "esphome/components/sensor/sensor.h"
+#include "esphome/components/uart/uart.h"
+
+namespace esphome {
+namespace cm1106 {
+
+static const char *const TAG = "cm1106";
```
DjordjeMandic: Move tag to source if you're not using it in multiple source files.

esphome/components/cm1106/cm1106.h
```diff
@@ -0,0 +1,46 @@
...
+  float get_setup_priority() const override { return esphome::setup_priority::DATA; }
+
+  void update() override;
+  void dump_config() override;
+
+  void calibrate_zero(uint16_t ppm);
+
+  void set_co2_sensor(sensor::Sensor *co2_sensor) { co2_sensor_ = co2_sensor; }
+
+ private:
```
DjordjeMandic: We don't use private. Make it protected.

DjordjeMandic (changes requested):
Use `this->` when accessing members.

You should maybe check in setup if component is accessible, if read fails, mark component as failed. If it was disconnected during boot, there would be huge log spam that could be prevented.

andrewjswan:
It works on the stand without any complaints or regressions.

DjordjeMandic:
> It works on the stand without any complaints or regressions.

What happens when esphome boots without sensor connected?

DjordjeMandic:
> You should maybe check in setup if component is accessible, if read fails, mark component as failed. If it was disconnected during boot, there would be huge log spam that could be prevented.

image

andrewjswan:
> What happens when esphome boots without sensor connected?

I don't know, but judging by the picture above, the uart component will have problems, it won't be able to read the data. The same problem will be with all components that are based on UARTComponent. For example MHZ-19
https://github.com/esphome/esphome/blob/96682f5cbefe9cd38d76adac1dcef779ee25ff7e/esphome/components/mhz19/mhz19.cpp#L43

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,97 @@
...
+                                   sizeof(response))) {
+    ESP_LOGW(TAG, "Reading data from CM1106 failed!");
+    this->status_set_warning();
+    return;
+  }
+
+  if (response[0] != 0x16 || response[1] != 0x05 || response[2] != 0x01) {
+    ESP_LOGW(TAG, "Got wrong UART response from CM1106: %02X %02X %02X %02X...", response[0], response[1], response[2],
+             response[3]);
+    this->status_set_warning();
```
andrewjswan, line 31: When there are problems reading, I set `this->status_set_warning();` as other components do.

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,97 @@
...
+  }
+  return crc;
+}
+
+void CM1106Component::update() {
+  uint8_t response[8] = {0};
+  if (!this->cm1106_write_command_(this->c_m1106_cmd_get_co2_, sizeof(this->c_m1106_cmd_get_co2_), response,
+                                   sizeof(response))) {
+    ESP_LOGW(TAG, "Reading data from CM1106 failed!");
+    this->status_set_warning();
```
andrewjswan, line 36: When there are problems reading, I set `this->status_set_warning();` as other components do.

andrewjswan:
This should prevent spam to log.
https://github.com/esphome/esphome/blob/96682f5cbefe9cd38d76adac1dcef779ee25ff7e/esphome/core/component.cpp#L151-L158

andrewjswan:
The only method I found was `available()`, but it doesn't seem to solve this problem.
https://github.com/esphome/esphome/blob/96682f5cbefe9cd38d76adac1dcef779ee25ff7e/esphome/components/uart/uart.h#L41

DjordjeMandic:
I apologize for the misunderstanding—I assumed you were already familiar with how component.setup() works.

Every component has a setup() method that runs once at boot. After setup, the component enters a loop, which is called continuously unless it's unnecessary for your use case. Instead, the scheduler periodically calls update() at intervals defined in the configuration.

Each component has its own status, which can be normal, warning, error, or failed. If a component is marked as failed, the scheduler disables it, preventing update() from being called until the next reboot.

In your case, you should perform a single UART check in setup() to verify that your component is responding correctly. If you don't receive a valid response (such as a header or acknowledgment), log an error and mark the component as failed. Since the component is in a failed state, the scheduler won't enable it to call update().

Additionally, in the dump_config() function, it's good practice to check the component’s status. If it has failed, log an error to make debugging easier.

For reference, you might find this example helpful:
https://github.com/esphome/esphome/pull/8250

andrewjswan:
> For reference, you might find this example helpful:

I'm not sure if this is correct, we may get a temporary problem during `setup`, which we can "solve" at best by rebooting.

DjordjeMandic:
> > For reference, you might find this example helpful:
> 
> I'm not sure if this is correct, we may get a temporary problem during `setup`, which we can "solve" at best by rebooting.

The approach taken in the [DHT12 component](https://github.com/esphome/esphome/blob/dev/esphome%2Fcomponents%2Fdht12%2Fdht12.cpp) is a good example of how to handle failed hardware initialization.

During setup(), the component attempts to communicate with the hardware. If it fails, it is marked as failed, preventing update() or loop() from running until the next reboot.

Consider a scenario where a user is running ESPHome firmware but has not physically connected a specific component. In such cases, it's completely reasonable to log an error in setup(), mark the component as failed, and also report its status in dump_config(). If the component didn’t work at boot, it’s assumed to be either not connected or faulty, so there’s no need to continue attempting communication while the system is running.

This approach is similar to how operating systems handle faulty USB devices. If a USB device is detected but isn’t working correctly, the system flags it as failed and disables it. This prevents repeated error messages or notifications.

In ESPHome, the setup() function should verify communication with the component at boot. If the check fails, the component should be marked as failed, and setup() should return early. This ensures that update() is never called until the next reboot. This is particularly useful for cases where the same firmware is used across multiple devices, but the component might not always be physically present or populated.

andrewjswan:
@DjordjeMandic Thanks for the tip, I added sensor failing check on startup

andrewjswan:
> In ESPHome, the setup() function should verify communication with the component at boot. If the check fails, the component should be marked as failed, and setup() should return early. This ensures that update() is never called until the next reboot. This is particularly useful for cases where the same firmware is used across multiple devices, but the component might not always be physically present or populated.

I agree that this is correct and good, but I relied on ready-made components that did not have this, and I also thought that the warning state would not spam the log.
Thank you, now I understand a little more how the ESPHome component system works.

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+  return crc;
+}
+
+void CM1106Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up CM1106...");
+  uint8_t response[8] = {0};
+  if (!this->cm1106_write_command_(this->c_m1106_cmd_get_co2_, sizeof(this->c_m1106_cmd_get_co2_), response,
+                                   sizeof(response))) {
+    ESP_LOGW(TAG, "Reading data from CM1106 failed!");
```
DjordjeMandic:
```suggestion
    ESP_LOGE(TAG, "Reading data from CM1106 failed!");
```
This is actually error, it happened in setup.

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  this->status_clear_warning();
+  ESP_LOGD(TAG, "CM1106 Successfully calibrated sensor to %uppm", ppm);
+}
+
+bool CM1106Component::cm1106_write_command_(uint8_t *command, size_t command_len, uint8_t *response,
+                                            size_t response_len) {
+  // Empty RX Buffer
+  while (this->available())
```
DjordjeMandic, line 90: Unrelated comment, this is not part of requested change.

Nasa has nice rule I like:
> All loops must have a fixed upper-bound. It must be trivially possible for a checking tool to prove statically that a preset upper-bound on the number of iterations of a loop cannot be exceeded. If the loop-bound cannot be proven statically, the rule is considered violated.

It's not really needed for this component and it's also not really used in esphome but here is example of implementation of that rule:

```suggestion
  // 1000ms timeout
  uint32_t startMillis = millis();
  while (this->available() && (millis() - startMillis < 1000UL))
```
In your case, you would also need to return false if timeout occurred.

DjordjeMandic (changes requested):
Just change log level from warning to error on communication check in setup.

esphome/components/cm1106/cm1106.h
jesserockz: Each source file should have its own tag anyway so we can identify which file the log comes from

DjordjeMandic: Makes sense, I thought that tags are related to part of component, not the source file.

esphome/components/cm1106/cm1106.cpp
andrewjswan: > It's not really needed for this component and it's also not really used in esphome but here is example of implementation of that rule

I thought about it, but in the case of this sensor it is more of a protection against garbage in case of failure of the sensor itself, because in normal operation, the buffer will always be empty and this cycle will pass in a single check cycle.

DjordjeMandic: Yeah, I agree. That's why I said that its unrelated to requested changes 😁

andrewjswan: > In your case, you would also need to return false if timeout occurred.


This is not a good option for using `millis()`, we will have a problem with resetting `millis()` to 0 every ~50 days.
I think I will abandon this change.

esphome/components/cm1106/cm1106.cpp
DjordjeMandic: This is fine

esphome/components/cm1106/cm1106.cpp
DjordjeMandic: > Unrelated comment, this is not part of requested change.

As I said, you don't have to implement this.

Since `millis()` and `startMillis` are both **unsigned integers**, subtraction inherently handles overflow (rollover) correctly due to **modular arithmetic**. In an unsigned integer, when `millis()` rolls over from `0xFFFFFFFF` to `0x00000000`, the difference calculation:

```cpp
elapsedTime = millis() - startMillis;
```

still yields the correct time difference. This works because unsigned subtraction in C/C++ wraps around naturally. Therefore, no special handling for overflow is needed when comparing elapsed time.

Here are some examples:
```cpp
bool condition = millis() - startMillis < 1000UL; // GOOD
```

```cpp
bool condition = millis() < startMillis + 1000UL; // BAD, will cause problem when millis overflows
```

DjordjeMandic (approved):

esphome/components/cm1106/sensor.py
```diff
@@ -0,0 +1,78 @@
...
+        {
+            cv.GenerateID(): cv.declare_id(CM1106Component),
+            cv.Required(CONF_CO2): sensor.sensor_schema(
+                unit_of_measurement=UNIT_PARTS_PER_MILLION,
+                icon=ICON_MOLECULE_CO2,
+                accuracy_decimals=0,
+                device_class=DEVICE_CLASS_CARBON_DIOXIDE,
+                state_class=STATE_CLASS_MEASUREMENT,
+            ),
+            cv.Optional(CONF_AUTOMATIC_BASELINE_CALIBRATION): cv.boolean,
```
DjordjeMandic:
```suggestion
```

esphome/components/cm1106/sensor.py
```diff
@@ -0,0 +1,78 @@
...
+
+async def to_code(config) -> None:
+    """Code generation entry point."""
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+    await uart.register_uart_device(var, config)
+
+    if CONF_CO2 in config:
+        sens = await sensor.new_sensor(config[CONF_CO2])
+        cg.add(var.set_co2_sensor(sens))
```
DjordjeMandic:
```suggestion
    await uart.register_uart_device(var, config)
    sens = await sensor.new_sensor(config[CONF_CO2])
    cg.add(var.set_co2_sensor(sens))
```

DjordjeMandic (changes requested):
Since CONF_CO2 is required, there is no need for if CONF_CO2 in config. Also `CONF_AUTOMATIC_BASELINE_CALIBRATION` is not implemented.

andrewjswan:
```
bool condition = millis() - startMillis < 1000UL; // GOOD
```
I meant that let's say we have some garbage in the buffer and reading it takes 2ms, and `millis()` is at its maximum value, then before reading we will have something like this:
```
millis() = 4 294 967 295
startMillis = 4 294 967 295
```
But at the moment of reading, we will exit the cycle after 1ms, and will not be able to read all the data, this is not critical, but it can lead to a failure of reading data in this update cycle.
```
(millis() =  0 - startMillis = 4 294 967 295) < 1000UL
```

DjordjeMandic (approved):

andrewjswan:
@DjordjeMandic Thank you very much for your patience and time.

esphome/components/cm1106/sensor.py
```diff
@@ -0,0 +1,73 @@
...
+    automation.Action,
+)
+CM1106ABCEnableAction = cm1106_ns.class_("CM1106ABCEnableAction", automation.Action)
+CM1106ABCDisableAction = cm1106_ns.class_("CM1106ABCDisableAction", automation.Action)
+
+CONFIG_SCHEMA = (
+    cv.Schema(
+        {
+            cv.GenerateID(): cv.declare_id(CM1106Component),
+            cv.Required(CONF_CO2): sensor.sensor_schema(
```
jesserockz: No sensors are required in ESPHome. If this is the only sensor for this component, then the entire `CONFIG_SCHEMA` should be an extension of `sensor.sensor_schema`.

Example:
https://github.com/esphome/esphome/blob/6f801fec66b7b2b5e4f7749afa14eaf271abebb4/esphome/components/a01nyub/sensor.py#L18-L26

```suggestion
            cv.Optional(CONF_CO2): sensor.sensor_schema(
```

jesserockz (changes requested):

esphome/components/cm1106/cm1106.h
```diff
@@ -0,0 +1,44 @@
...
+  void dump_config() override;
+
+  void calibrate_zero(uint16_t ppm);
+
+  void set_co2_sensor(sensor::Sensor *co2_sensor) { this->co2_sensor_ = co2_sensor; }
+
+ protected:
+  uint8_t c_m1106_cmd_get_co2_[4] = {0x11, 0x01, 0x01, 0xED};
+  uint8_t c_m1106_cmd_set_co2_calib_[6] = {0x11, 0x03, 0x03, 0x00, 0x00, 0x00};
+  uint8_t c_m1106_cmd_set_co2_calib_response_[4] = {0x16, 0x01, 0x03, 0xE6};
```
jesserockz: These should be static const in the cpp file

jesserockz (changes requested):

esphome/components/cm1106/cm1106.h
andrewjswan: @jesserockz if i change this to static i have this error:
```
src/esphome/components/cm1106/cm1106.cpp: In member function 'virtual void esphome::cm1106::CM1106Component::setup()':
src/esphome/components/cm1106/cm1106.cpp:26:36: error: invalid conversion from 'const uint8_t*' {aka 'const unsigned char*'} to 'uint8_t*' {aka 'unsigned char*'} [-fpermissive]
   if (!this->cm1106_write_command_(c_m1106_cmd_get_co2_, sizeof(c_m1106_cmd_get_co2_), response,
                                    ^~~~~~~~~~~~~~~~~~~~
In file included from src/esphome/components/cm1106/cm1106.cpp:1:
src/esphome/components/cm1106/cm1106.h:26:39: note:   initializing argument 1 of 'bool esphome::cm1106::CM1106Component::cm1106_write_command_(uint8_t*, size_t, uint8_t*, size_t)'
   bool cm1106_write_command_(uint8_t *command, size_t command_len, uint8_t *response, size_t response_len);
                              ~~~~~~~~~^~~~~~~
src/esphome/components/cm1106/cm1106.cpp: In member function 'virtual void esphome::cm1106::CM1106Component::update()':
src/esphome/components/cm1106/cm1106.cpp:36:36: error: invalid conversion from 'const uint8_t*' {aka 'const unsigned char*'} to 'uint8_t*' {aka 'unsigned char*'} [-fpermissive]
   if (!this->cm1106_write_command_(c_m1106_cmd_get_co2_, sizeof(c_m1106_cmd_get_co2_), response,
                                    ^~~~~~~~~~~~~~~~~~~~
In file included from src/esphome/components/cm1106/cm1106.cpp:1:
src/esphome/components/cm1106/cm1106.h:26:39: note:   initializing argument 1 of 'bool esphome::cm1106::CM1106Component::cm1106_write_command_(uint8_t*, size_t, uint8_t*, size_t)'
   bool cm1106_write_command_(uint8_t *command, size_t command_len, uint8_t *response, size_t response_len);
                              ~~~~~~~~~^~~~~~~
```
I can't change the cm1106_write_command_ call, I don't see the point in creating another procedure.
What should I do?

andrewjswan: I can't convert `cm1106_write_command_` to static, because the array changes, for example, here:

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L62-L66

or here:

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L91

andrewjswan:
@jesserockz Any news?

andrewjswan:
@jesserockz , does it make sense to constantly update branch, or is it better to wait for your decision and then update?
Thanks in advance for your answer.

esphome/components/cm1106/cm1106.h
DjordjeMandic: > I can't convert `cm1106_write_command_` to static, because the array changes, for example, here:
> 
> https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L62-L66

You're already copying it into local array, it can be const there.

> 
> or here:
> 
> https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L91

Do something like this:
```cpp
uint8_t checksum = cm1106_checksum(command, command_len);
this->write_array(command, command_len - 1);
this->write_byte(checksum);
```

Then it can be const

andrewjswan: @DjordjeMandic 
Then there will be an error here:


https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L62-L66

DjordjeMandic: I don't see where and why, memcpy takes pointer to const source in memory, memcmp also does not modify the memory. You have your local cmd array.

andrewjswan: @DjordjeMandic Maybe I didn't explain it right, or I didn't understand your proposal right. 
The problem is that if I make arrays:

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.h#L24-L26

as `const`, I get an error in the `cm1106_write_command_ function` - `invalid conversion from 'const uint8_t*' {aka 'const unsigned char*'} to 'uint8_t*' {aka 'unsigned char*'}`.

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L86

There are two ways out of this, either copy `const` into an array and pass it to `cm1106_write_command_` But this is an extra step for each call, or duplicate the `cm1106_write_command_` call function, one `with const` and the `other without`.

Because two call options are used:

1. roughly speaking, a constant:
 https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L32

2. mutable array:
 https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L62-L68

That's why I preferred not to do unnecessary transformations when querying data, and to use non-const arrays. This way we waste a little memory, but we increase the speed.
Maybe I'm wrong, but in my opinion this solution is better.

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,111 @@
...
+    this->read();
+  command[command_len - 1] = cm1106_checksum(command, command_len);
+  this->write_array(command, command_len);
```
DjordjeMandic:
```suggestion
  this->write_array(command, command_len - 1);
  this->write_byte(cm1106_checksum(command, command_len));
```

esphome/components/cm1106/cm1106.h
DjordjeMandic:
```suggestion
```

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,111 @@
+#include "cm1106.h"
+#include "esphome/core/log.h"
+
+#include <cinttypes>
+
+namespace esphome {
+namespace cm1106 {
+
+static const char *const TAG = "cm1106";
+
```
DjordjeMandic, line 13:
```suggestion
static const uint8_t c_m1106_cmd_get_co2_[4] = {0x11, 0x01, 0x01, 0xED};
static const uint8_t c_m1106_cmd_set_co2_calib_[6] = {0x11, 0x03, 0x03, 0x00, 0x00, 0x00};
static const uint8_t c_m1106_cmd_set_co2_calib_response_[4] = {0x16, 0x01, 0x03, 0xE6};
```

Also I would make this all uppercase

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,111 @@
...
+             response[3]);
+    this->status_set_warning();
+    return;
+  }
+
+  this->status_clear_warning();
+  ESP_LOGD(TAG, "CM1106 Successfully calibrated sensor to %uppm", ppm);
+}
+
+bool CM1106Component::cm1106_write_command_(uint8_t *command, size_t command_len, uint8_t *response,
```
DjordjeMandic:
```suggestion
bool CM1106Component::cm1106_write_command_(const uint8_t *command, size_t command_len, uint8_t *response,
```

Make this first arg (pointer to command) read only. That will fix your issue

esphome/components/cm1106/cm1106.h
```diff
@@ -0,0 +1,44 @@
...
+
+  sensor::Sensor *co2_sensor_{nullptr};
+
+  bool cm1106_write_command_(uint8_t *command, size_t command_len, uint8_t *response, size_t response_len);
```
DjordjeMandic:
```suggestion
  bool cm1106_write_command_(const uint8_t *command, size_t command_len, uint8_t *response, size_t response_len);
```

DjordjeMandic (changes requested):
Take a look at suggested changes and try to implement them. It should work with read only pointer..

andrewjswan:
> Take a look at suggested changes and try to implement them. It should work with read only pointer..

What to do with this block?

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L62-L68

Here the target CO2 value is added for calibration, i.e. the **target CO2 value is added to the array**, and then the new array is passed to the `cm1106_write_command_` procedure.

DjordjeMandic:
> > Take a look at suggested changes and try to implement them. It should work with read only pointer..
> 
> What to do with this block?
> 
> https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L62-L68
> 
> Here the target CO2 value is added for calibration, i.e. the **target CO2 value is added to the array**, and then the new array is passed to the `cm1106_write_command_` procedure.

---

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L63 

```cpp
void* memcpy( void* dest, const void* src, std::size_t count );
```

`src` is not modified, no problem there.

---

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L75

```cpp
int memcmp( const void* lhs, const void* rhs, std::size_t count );
```

`lhs` and `rhs` are not modified, no problem there.

---

https://github.com/andrewjswan/esphome/blob/cb7441ae77d062988b1e24685fb338cb3bf69202/esphome/components/cm1106/cm1106.cpp#L68

Here you have local cmd array, that also wont be a problem. Try it

andrewjswan:
> Here you have local cmd array, that also wont be a problem. Try it

I did, I hope the assembly will pass without errors, I will try to check at home, and if it works, I will update the status of PR.
Thanks!

andrewjswan:
I tested on the device, the readings are transmitted, problems are not noticed.

bbluvgg:
Hello, I have tested the external_components:[cm1106] and confirmed that the ESP32 can successfully read data. However, I noticed an issue with the CO2 calibration function id(cm1106_sensor).calibrate_zero(). I created a button to implement the calibration feature, but after calibration, the CO2 value remains stuck at 400ppm. Could you please explain why this is happening and provide the official standard usage guidelines for the calibration function? We greatly appreciate your work and guidance.
`button:
  - platform: factory_reset
    id: reset
    internal: true
  - platform: template
    name: "DAIKIN CO2 Calibration" 
    entity_category: "diagnostic"
    on_press:
      then:
        - switch.turn_on: co2led
        - delay: 10min
        - lambda: |-
            id(cm1106_sensor).calibrate_zero(400);
        - switch.turn_off: co2led`

andrewjswan:
> the CO2 value remains stuck at 400ppm

This function calibrates the indicators to 400ppm, i.e. the minimum measured value becomes 400, let's say it was 300 (which does not happen in nature), after calibration it will become 400. This is how all NDIR CO2 sensors work.

https://github.com/andrewjswan/esphome/blob/2d3d83e0f35614d6cc59e2145eb8c0c6c40ca466/esphome/components/cm1106/cm1106.h#L33

andrewjswan:
```Yaml
    on_...:
      then:
        - cm1106.calibrate_zero: my_cm1106_id
```

esphome/components/cm1106/cm1106.h
andrewjswan: > These should be static const in the cpp file

Changed to static ...

andrewjswan:
@DjordjeMandic Thanks for Help! I hope all the changes are made and this PR will receive a review and will be added to ESPHome.

schellekensonline:
@andrewjswan, thank you very much for all the work on the CM1106 sensor component. While testing I noticed that the following errors and warnings are thrown when calibrating the sensor. 

[07:47:31][E][uart:015]: Reading from UART timed out at byte 0!
[07:47:31][W][cm1106:070]: Reading data from CM1106 failed!
[07:47:31][W][component:157]: Component cm1106.sensor set Warning flag: unspecified
[07:47:31][D][switch:016]: 'CM1106 zero calibration' Turning OFF.
[07:47:31][W][component:237]: Component api took a long time for an operation (134 ms).
[07:47:31][W][component:238]: Components should block for at most 30 ms.
[07:47:47][W][component:170]: Component cm1106.sensor cleared Warning flag
[07:47:47][D][cm1106:057]: CM1106 Received CO₂=643ppm DF3=00 DF4=A1

Maybe that the calibration messes up the normal sensor reading resulting in a timeout? I is not a big deal, the calibration works but throwing errors during normal operation is not nice of course. Next to that is it possible to pass the calibration value to the calibration function. 400 ppm is already on the low side nowadays and will increase over the years to 500ppm so hard coding may be not the most handy method.  

The complete yaml configuration I use I have added below, also for others so that they can use the sensor more easily. 

Thanks again for all the work, hope this helps! 

```yaml
    external_components:
        - source: github://pr#8293
          components: [cm1106]
          refresh: 60s
    uart:
      - id: cm1106_uart
        rx_pin: 17
        tx_pin: 16
        baud_rate: 9600
    sensor:
      - platform: cm1106
        co2:
          name: "CO2 Sensor"
          id: co2index
          icon: "mdi:molecule-co2"
        id: cm1106_sensor
    switch:
      - platform: template
        id: cm1106_zero_calibration
        name: "CM1106 zero calibration"
        turn_on_action:
          - cm1106.calibrate_zero: cm1106_sensor
          - switch.turn_off: cm1106_zero_calibration
```

andrewjswan:
> Maybe that the calibration messes up the normal sensor reading resulting in a timeout?

This is possible if the calibration and polling times coincide, if I'm not mistaken, the sensor does not send data during calibration. I did it like this:

```Yaml
  - platform: template
    internal: true
    id: co2_calibration
    lambda: |-
      if (id(co2sensor).state < 400)
      {
        return true;
      }
      else
      {
        return false;
      }
    filters:
      - delayed_on: 15min
    on_press:
      then:
        - cm1106.calibrate_zero: cm1106_sensor
```

schellekensonline:
Your method does not prevent the commutation issue right? But again it is not a big deal. I prefer a button in HA but that is personal. Is it possible to pass the calibration value as mentioned? 400ppm is already on the low side and it will increase over the years with the current political situation.

andrewjswan:
> Is it possible to pass the calibration value as mentioned?

Through lambda yes.

esphome/components/cm1106/cm1106.cpp
```diff
@@ -0,0 +1,112 @@
...
+  uint8_t checksum = cm1106_checksum(response, sizeof(response));
+  if (response[7] != checksum) {
+    ESP_LOGW(TAG, "CM1106 Checksum doesn't match: 0x%02X!=0x%02X", response[7], checksum);
+    this->status_set_warning();
+    return;
+  }
+
+  this->status_clear_warning();
+
+  int16_t ppm = response[3] << 8 | response[4];
```
DjordjeMandic: Shouldn't this be positive only? You're using `%u` in log function, it will still work since reported values are 0-5000ppm.
```suggestion
  uint16_t ppm = response[3] << 8 | response[4];
```

esphome/components/cm1106/sensor.py
```diff
@@ -0,0 +1,74 @@
...
+cm1106_ns = cg.esphome_ns.namespace("cm1106")
+CM1106Component = cm1106_ns.class_(
+    "CM1106Component", cg.PollingComponent, uart.UARTDevice
+)
+CM1106CalibrateZeroAction = cm1106_ns.class_(
+    "CM1106CalibrateZeroAction",
+    automation.Action,
+)
+CM1106ABCEnableAction = cm1106_ns.class_("CM1106ABCEnableAction", automation.Action)
+CM1106ABCDisableAction = cm1106_ns.class_("CM1106ABCDisableAction", automation.Action)
```
patagonaa:
```suggestion
```
I assume these are leftovers from the MHZ19 component (they aren't used here as far as I can tell)

andrewjswan:
@patagonaa Thanks!

esphome/components/cm1106/sensor.py
```diff
@@ -0,0 +1,72 @@
...
+                icon=ICON_MOLECULE_CO2,
+                accuracy_decimals=0,
+                device_class=DEVICE_CLASS_CARBON_DIOXIDE,
+                state_class=STATE_CLASS_MEASUREMENT,
+            ),
+        },
+    )
+    .extend(cv.polling_component_schema("60s"))
+    .extend(uart.UART_DEVICE_SCHEMA)
+)
```
DjordjeMandic, line 44: It's a bit meaningless to have optional sensor when it's the only sensor.

I'd change this config schema to be sensor schema which is extended with poling component schema and uart device schema. Might require some C/C++ changes

As example you can take a look at schema for dallas temp sensor:
https://github.com/esphome/esphome/blob/219ba6152cc033b5da99566537664d842b3faee9/esphome/components/dallas_temp/sensor.py#L20-L35

DjordjeMandic (changes requested):

esphome/components/cm1106/sensor.py
andrewjswan: > It's a bit meaningless to have optional sensor when it's the only sensor.

This is done on purpose to separate the device ID (for calibration) and the sensor ID for use in lambda, if i remove the device ID, we will lose the ability to manually start calibration.

DjordjeMandic (approved):

andrewjswan:
@jesserockz any news?

esphome/components/cm1106/sensor.py
jesserockz: > > It's a bit meaningless to have optional sensor when it's the only sensor.
> 
> This is done on purpose to separate the device ID (for calibration) and the sensor ID for use in lambda, if i remove the device ID, we will lose the ability to manually start calibration.

This is not true, they just become the same ID. The `CM1106Component` class can just implement `sensor::Sensor` directly and you can still call any custom functions or actions with that ID.

jesserockz (changes requested):

andrewjswan:
@jesserockz , I looked at the ESPHome source code, and sensors similar to `CM1106` have the same scheme:

`Senseair`

https://github.com/esphome/esphome/blob/b82666002dc55b60b3e7f3f2549c0bf8a0203f91/esphome/components/senseair/sensor.py#L37-L52

`PM1006`

https://github.com/esphome/esphome/blob/b82666002dc55b60b3e7f3f2549c0bf8a0203f91/esphome/components/pm1006/sensor.py#L24-L40

`T6615`

https://github.com/esphome/esphome/blob/b82666002dc55b60b3e7f3f2549c0bf8a0203f91/esphome/components/t6615/sensor.py#L18-L32

It turns out that this scheme does not contradict the internal requirements of ESPHome, and if you do not mind, I would stay within its limits.

andrewjswan:
@jesserockz @DjordjeMandic Any decision? Feedback?

DjordjeMandic:
> @jesserockz , I looked at the ESPHome source code, and sensors similar to `CM1106` have the same scheme... 
>
> It turns out that this scheme does not contradict the internal requirements of ESPHome, and if you do not mind, I would stay within its limits.

I would then at least make CONF_CO2 required, not optional.

andrewjswan:
> I would then at least make CONF_CO2 required, not optional.

Well, if that's the only problem, then I'll do it, it's simple.

andrewjswan:
> I would then at least make CONF_CO2 required, not optional.

Ready ...

esphome/components/cm1106/sensor.py
jesserockz: I see there are some Required sensors in ESPHome, this is wrong and I am going to change them. Please alter this to be Optional.

I am going to also update https://developers.esphome.io with this requirement. Currently it has only been specified adhoc in PR reviews when it came up (and we didnt have developer docs)

https://github.com/esphome/esphome/pull/8803

jesserockz (changes requested):

DjordjeMandic:
Sorry for extra work @andrewjswan .

@jesserockz I recommended to make it required based on senseair and t6615 component. 
> The CM1106Component class can just implement sensor::Sensor directly and you can still call any custom functions or actions with that ID.

Since there is only one sensor, does it make sense to have optional sensor ptr instead of implementing sensor::Sensor directly? I see no use for this component without optional sensor supplied in yaml.

jesserockz:
That is why my original comment said to just make it top level and don't put the key underneath.
But if there is a key for the sensor, it should be optional.

andrewjswan:
> Sorry for extra work

No problem ... I make changes ...

andrewjswan:
> That is why my original comment said to just make it top level and don't put the key underneath.

Is this mandatory? In my opinion, the current implementation is logical, and will allow adding quality sensors or other CO2-based sensors in the future.

If this is not a mandatory requirement, I would leave the implementation in its current version (like other sensors in ESPHome). https://github.com/esphome/esphome/pull/8293#issuecomment-2818250315

If it is mandatory, tell me, I will try to rework it.

jesserockz:
@andrewjswan No, you can have it the way you have it now, but change the sensor to Optional, then I have no issues =)

DjordjeMandic (approved):

jesserockz (dismissed):

esphome/components/cm1106/sensor.py
```diff
@@ -0,0 +1,72 @@
...
+    await cg.register_component(var, config)
+    await uart.register_uart_device(var, config)
+    if co2_config := config.get(CONF_CO2):
+        sens = await sensor.new_sensor(co2_config)
+        cg.add(var.set_co2_sensor(sens))
+
+
+CALIBRATION_ACTION_SCHEMA = maybe_simple_id(
+    {
+        cv.Required(CONF_ID): cv.use_id(CM1106Component),
```
jesserockz: Optional change to allow the user to call the action without any id at all if you only have one of this platform configured, ESPHome can figure it out automatically
```suggestion
        cv.GenerateID(): cv.use_id(CM1106Component),
```

andrewjswan:
> but change the sensor to Optional

Ready ...

> Optional change to allow the user to call the action without any id at all if you only have one of this platform configured

Ready ...

jesserockz (approved):

AbianTang:
Platform not found: 'sensor.cm1106'.
  platform: cm1106
  co2: 
    name: CM1106 CO2 Value
  update_interval: 15s
  id: co2sensor

andrewjswan:
> Platform not found: 'sensor.cm1106'.

Try

```Yaml
external_components:
  - source:
      type: git
      url: https://github.com/esphome/esphome
      ref: dev
    components: [cm1106]
    refresh: 60s
```

PR #8316, opened 2025-02-24
[uptime][text_sensor] Allow to specify the type

### LINKED ISSUE #137: V2/V3 - Uptime type timestamp shown in integer scientific notation
The uptime sensor when used as type timestamp shows as scientific notation in the web browser display.  

Example:
```
Uptime    2e+09
```

Desired display may be debated (readable static date/time of boot time as represented by the timestamp, or a changing "ago" duration).  Perhaps a `display_type` attribute could be added.

RFDarter:

Allows the user to specify the type (seconds, timestamp) like on the `uptime` `sensor`.
Also allows the user to set the format for the strftime. default is `%F %T`

Bildschirmfoto vom 2025-02-24 23-41-56

usecase: https://github.com/esphome/esphome-webserver/issues/137

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/esphome-webserver/issues/137

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml

esphome:
  name: webserver

esp32:
  board: wemos_d1_mini32
  framework:
    type: arduino

wifi:
  ssid: !secret wifi_ssid

ota:
  platform: esphome

logger:
  level: DEBUG


web_server:
  version: 3

time:
  - platform: sntp
    id: sntp_time

text_sensor:
  - platform: uptime
    name: Uptime
    type: timestamp
    format: "Date: %F  Time: %T"

```

## Checklist:

### DIFF
--- a/esphome/components/uptime/text_sensor/__init__.py
+++ b/esphome/components/uptime/text_sensor/__init__.py
@@ -18,14 +18,6 @@ UptimeTimestampTextSensor = uptime_ns.class_(
     "UptimeTimestampTextSensor", text_sensor.TextSensor, cg.Component
 )
 
-
-CONFIG_SCHEMA = text_sensor.text_sensor_schema(
-    UptimeDurationTextSensor,
-    icon=ICON_TIMER,
-    entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
-).extend(cv.polling_component_schema("30s"))
-
-
 CONFIG_SCHEMA = cv.typed_schema(
     {
         CONF_DURATION: text_sensor.text_sensor_schema(
--- a/esphome/components/uptime/text_sensor/__init__.py
+++ b/esphome/components/uptime/text_sensor/__init__.py
@@ -49,7 +49,7 @@ CONFIG_SCHEMA = cv.typed_schema(
         )
         .extend(cv.COMPONENT_SCHEMA),
     },
-    default_type="seconds",
+    default_type=CONF_DURATION,
 )
 
 
--- a/esphome/components/uptime/text_sensor/__init__.py
+++ b/esphome/components/uptime/text_sensor/__init__.py
@@ -2,13 +2,13 @@ import esphome.codegen as cg
 from esphome.components import text_sensor, time
 import esphome.config_validation as cv
 from esphome.const import (
-    CONF_DURATION,
     CONF_TIME_ID,
     ENTITY_CATEGORY_DIAGNOSTIC,
     ICON_TIMER,
 )
 
-CONF_TIMESTAMP = "timestamp"
+TYPE_DURATION = "duration"
+TYPE_TIMESTAMP = "timestamp"
 
 uptime_ns = cg.esphome_ns.namespace("uptime")
 UptimeDurationTextSensor = uptime_ns.class_(
@@ -20,28 +20,28 @@ UptimeTimestampTextSensor = uptime_ns.class_(
 
 CONFIG_SCHEMA = cv.typed_schema(
     {
-        CONF_DURATION: text_sensor.text_sensor_schema(
+        TYPE_DURATION: text_sensor.text_sensor_schema(
             UptimeDurationTextSensor,
             icon=ICON_TIMER,
             entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
         ).extend(cv.polling_component_schema("30s")),
-        CONF_TIMESTAMP: text_sensor.text_sensor_schema(
-            UptimeTimestampTextSensor,
-            icon=ICON_TIMER,
-            entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
+        TYPE_TIMESTAMP: cv.Schema(
+            {
+                cv.GenerateID(CONF_TIME_ID): cv.All(
+                    cv.requires_component("time"), cv.use_id(time.RealTimeClock)
+                ),
+            }
         )
         .extend(
-            cv.Schema(
-                {
-                    cv.GenerateID(CONF_TIME_ID): cv.All(
-                        cv.requires_component("time"), cv.use_id(time.RealTimeClock)
-                    ),
-                }
+            text_sensor.text_sensor_schema(
+                UptimeTimestampTextSensor,
+                icon=ICON_TIMER,
+                entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
             )
         )
         .extend(cv.COMPONENT_SCHEMA),
     },
-    default_type=CONF_DURATION,
+    default_type=TYPE_DURATION,
 )
 
 
--- a/esphome/components/uptime/text_sensor/__init__.py
+++ b/esphome/components/uptime/text_sensor/__init__.py
@@ -1,11 +1,18 @@
 import esphome.codegen as cg
 from esphome.components import text_sensor, time
 import esphome.config_validation as cv
-from esphome.const import CONF_TIME_ID, ENTITY_CATEGORY_DIAGNOSTIC, ICON_TIMER
+from esphome.const import (
+    CONF_DURATION,
+    CONF_TIME_ID,
+    ENTITY_CATEGORY_DIAGNOSTIC,
+    ICON_TIMER,
+)
+
+CONF_TIMESTAMP = "timestamp"
 
 uptime_ns = cg.esphome_ns.namespace("uptime")
-UptimeSecondsTextSensor = uptime_ns.class_(
-    "UptimeSecondsTextSensor", text_sensor.TextSensor, cg.PollingComponent
+UptimeDurationTextSensor = uptime_ns.class_(
+    "UptimeDurationTextSensor", text_sensor.TextSensor, cg.PollingComponent
 )
 UptimeTimestampTextSensor = uptime_ns.class_(
     "UptimeTimestampTextSensor", text_sensor.TextSensor, cg.Component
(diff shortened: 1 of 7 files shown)


esphome/components/uptime/text_sensor/__init__.py
```diff
@@ -1,19 +1,55 @@
...
-from esphome.components import text_sensor
+from esphome.components import text_sensor, time
 import esphome.config_validation as cv
-from esphome.const import ENTITY_CATEGORY_DIAGNOSTIC, ICON_TIMER
+from esphome.const import CONF_TIME_ID, ENTITY_CATEGORY_DIAGNOSTIC, ICON_TIMER
 
 uptime_ns = cg.esphome_ns.namespace("uptime")
-UptimeTextSensor = uptime_ns.class_(
-    "UptimeTextSensor", text_sensor.TextSensor, cg.PollingComponent
+UptimeSecondsTextSensor = uptime_ns.class_(
```
clydebarrow: `UptimeDurationTextSensor` would be better, it's not always expressed in seconds. Or just leave it as `UptimeTextSensor` since "uptime" implies duration anyway - that would avoid many of the small changes. I'd suggest the latter.

esphome/components/uptime/text_sensor/__init__.py
```diff
@@ -1,19 +1,55 @@
...
+        "timestamp": text_sensor.text_sensor_schema(
+            UptimeTimestampTextSensor,
+            icon=ICON_TIMER,
+            entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
+        )
+        .extend(
+            cv.Schema(
+                {
+                    cv.GenerateID(CONF_TIME_ID): cv.All(
+                        cv.requires_component("time"), cv.use_id(time.RealTimeClock)
```
clydebarrow: ~This should be specific to the timestamp type as no time component is needed for the duration sensor.~
Also at first glance it looks like the `time` component is required for both types - can the formatting be altered to avoid that impression?

esphome/components/uptime/text_sensor/__init__.py
```diff
@@ -1,19 +1,55 @@
...
-    UptimeTextSensor,
+    UptimeSecondsTextSensor,
     icon=ICON_TIMER,
     entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
 ).extend(cv.polling_component_schema("30s"))
 
 
+CONFIG_SCHEMA = cv.typed_schema(
+    {
+        "seconds": text_sensor.text_sensor_schema(
```
clydebarrow:
```suggestion
        "duration": text_sensor.text_sensor_schema(
```

Also consider defining constants `TYPE_TIMESTAMP` and `TYPE_DURATION` rather than embedding strings - not an absolute requirement when only used once, but is consistent with the style of other components.

esphome/components/uptime/text_sensor/uptime_timestamp_text_sensor.cpp
```diff
@@ -0,0 +1,36 @@
...
+  this->time_->add_on_time_sync_callback([this]() {
+    if (this->has_state_)
+      return;  // No need to update the timestamp if it's already set
+
+    auto now = this->time_->now();
+    const uint32_t ms = millis();
+    if (!now.is_valid())
+      return;  // No need to update the timestamp if the time is not valid
+
+    std::string now_string = this->time_->now().strftime("%F %T");
```
clydebarrow: You had better make the format configurable since somebody will ask for that anyway.

esphome/components/uptime/text_sensor/uptime_timestamp_text_sensor.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+    this->publish_state(now_string);
+  });
+}
+
+float UptimeTimestampTextSensor::get_setup_priority() const { return setup_priority::HARDWARE; }
+void UptimeTimestampTextSensor::dump_config() { LOG_TEXT_SENSOR("", "Uptime Text Sensor", this); }
```
clydebarrow:
```suggestion
void UptimeTimestampTextSensor::dump_config() { LOG_TEXT_SENSOR("", "Uptime Timestamp Text Sensor", this); }
```
Or something like that.

esphome/components/uptime/text_sensor/uptime_timestamp_text_sensor.h
```diff
@@ -0,0 +1,30 @@
...
+ public:
+  void dump_config() override;
+  void setup() override;
+
+  float get_setup_priority() const override;
+
+  void set_time(time::RealTimeClock *time) { this->time_ = time; }
+
+ protected:
+  time::RealTimeClock *time_;
```
clydebarrow:
```suggestion
  time::RealTimeClock *time_{};
```
Or even better, make it a constructor parameter so it's always initialised.

clydebarrow (changes requested):

esphome/components/uptime/text_sensor/uptime_timestamp_text_sensor.cpp
RFDarter: makes sense.

esphome/components/uptime/text_sensor/__init__.py
RFDarter: from my testing it does seem to be requiered only of `timestamp` is specified. I copied the `uptime` `sensor` schema since it seems to work there

clydebarrow: I believe it is correct as you have it, it just doesn't look right :-) Maybe extend the schema with TIME_ID by the text_sensor_schema rather than the other way around. That will make it obvious I think.

esphome/components/uptime/text_sensor/uptime_timestamp_text_sensor.h
RFDarter: if I make it a constructor parameter how would I hand over the the pointer to the rtc? 
All the components seem to allways use a setter to do stuff like that. 
Even the `uptime` `sensor` is doing it the same way.

esphome/components/uptime/text_sensor/__init__.py
```diff
@@ -1,19 +1,54 @@
...
 import esphome.config_validation as cv
-from esphome.const import ENTITY_CATEGORY_DIAGNOSTIC, ICON_TIMER
+from esphome.const import (
+    CONF_DURATION,
+    CONF_TIME_ID,
+    ENTITY_CATEGORY_DIAGNOSTIC,
+    ICON_TIMER,
+)
+
+CONF_TIMESTAMP = "timestamp"
```
clydebarrow:
```suggestion
TYPE_TIMESTAMP = "timestamp"
```
It's a value, not a config key name. Same for CONF_DURATION.

esphome/components/uptime/text_sensor/__init__.py
RFDarter: like so?

```py
        CONF_DURATION: text_sensor.text_sensor_schema(
            UptimeDurationTextSensor,
            icon=ICON_TIMER,
            entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
        ).extend(cv.polling_component_schema("30s")),
        CONF_TIMESTAMP: cv.Schema(
            {
                cv.GenerateID(CONF_TIME_ID): cv.All(
                    cv.requires_component("time"), cv.use_id(time.RealTimeClock)
                ),
            }
        )
        .extend(
            text_sensor.text_sensor_schema(
                UptimeTimestampTextSensor,
                icon=ICON_TIMER,
                entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
            )
        )
        .extend(cv.COMPONENT_SCHEMA),
```

clydebarrow:
Also need to add test case, and docs.

esphome/components/uptime/text_sensor/__init__.py
clydebarrow: yeah, that's probably better. I would like the `.extend` on the same line as the preceding parenthesis, but the formatter doesn't like it. Don't sweat it, I'm nit-picking here.

esphome/components/uptime/text_sensor/uptime_timestamp_text_sensor.cpp
RFDarter: validating a format provided the the user seems to be tricky. Can you think of an easy solution?

RFDarter: or maybe we dont need to validate it at all?

clydebarrow: This will do it (obviously replace the prints with `raise cv.Invalid()`)
```
import re

allowed = set("aAbBdejmUWwxyYHIlMNPpSXZ")

def check(string):
    indexes = [x.start() for x in re.finditer('%', string)]
    if not indexes:
        print("no format specifiers in ", string)
        return

    if indexes[-1]+1 == len(string):
        print("format ends with % in ", string)
        return
    wrong = set(string[i+1] for i in indexes).difference(allowed)
    if wrong:
        print(f"Invalid format specifiers: {",".join('%'+c for c in wrong)}")
        return
    print("Success with ", string)
```

clydebarrow:
You'll need to resolve the conflicts before it's ready for review.

PR #8317, opened 2025-02-25
[tmp1075] fix component for TMP1075N

### LINKED ISSUE #5387: TMP1075: device ID checks prevent use with TMP1075N
### The problem

Per the [TMP1075 datasheet](https://www.ti.com/lit/ds/symlink/tmp1075.pdf) Figure 9-2 (page 16), the device ID register is not available on the TMP1075N devices. The current code checks the device ID and marks the component failed if a valid device ID is not read out, which means it fails when used with the TMP1075N.

I did a quick'n'dirty modification of the TMP1075 component code to remove the device ID checks and was able to read temperature from two TMP1075Ns on the bus. There's probably a more elegant solution than just completely removing the device ID checks, perhaps a configuration entry to specify the device variant, or to directly skip the device ID check.

I can implement / PR one of the above if desired.

2023.10.4

pip

_No response_

ESP32

### Board

custom ESP32-S3

tmp1075

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

ssieb:

The component tries to read and compare a register that doesn't exist on the TMP1075**N**.  It also should only be doing byte operations on the config register.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/5387

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/tmp1075/tmp1075.cpp
+++ b/esphome/components/tmp1075/tmp1075.cpp
@@ -18,14 +18,9 @@ static uint16_t temp2regvalue(float temp);
 static float regvalue2temp(uint16_t regvalue);
 
 void TMP1075Sensor::setup() {
-  uint8_t die_id;
-  if (!this->read_byte(REG_DIEID, &die_id)) {
-    ESP_LOGW(TAG, "'%s' - unable to read ID", this->name_.c_str());
-    this->mark_failed();
-    return;
-  }
-  if (die_id != EXPECT_DIEID) {
-    ESP_LOGW(TAG, "'%s' - unexpected ID 0x%x found, expected 0x%x", this->name_.c_str(), die_id, EXPECT_DIEID);
+  uint8_t cfg;
+  if (!this->read_byte(REG_CFGR, &cfg)) {
+    ESP_LOGE(TAG, "'%s' - unable to read", this->name_.c_str());
     this->mark_failed();
     return;
   }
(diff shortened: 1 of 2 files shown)


jesserockz (approved):

sybrenstuvel (changes requested):
I think it would be better to check against the known IDs, rather than skipping the ID check altogether. That will make mix-ups of I2C addresses a lot easier to debug.

Also this PR changes more than what was described as the solution in #5387. Not only is the ID check removed, but also the size of the configuration register is changed. I'm fine accepting that change as an optimization, to reduce data transferred in configuration changes. But that should IMO be in a separate commit, with a separate reasoning as to why this is done.

sybrenstuvel:
> I think it would be better to check against the known IDs, rather than skipping the ID check altogether. That will make mix-ups of I2C addresses a lot easier to debug.

Actually, I checked the datasheet, and it says:

> this register only available on the TMP1075 non-N orderables

So the entire DeviceID register is unavailable on the TMP1075N. So yeah, 👍 on the removal of the ID check. I can confirm the PR works on a TMP1075 IC.

Still, IMO the optimization for the config register size should be in a separate commit.

ssieb:
If you read the datasheet, you are only supposed to byte read the config register for the N as well.

antsundq:
Tested this and can confirm it works with a TMP1075N using an ESP32.

PR #8325, opened 2025-02-25
Fix MQTT blocking main loop for multiple seconds at a time

### LINKED ISSUE #6810: MQTT component causes loop watchdog resets, measurement errors
### The problem

Originally commented on #4717 but I think this deserves its own separate bug as it's causing real problems, not just a warning.

With poor WiFi connectivity, I see the MQTT component taking so long in the loop that I've had to disable the watchdog timer on the loop thread, as it was causing resets. Even with the watchdog disabled, the mqtt component still sometimes spends over 10 seconds in its `call_loop()`, and it's causing other sensors to break (in my case a rotary encoder is missing full revolutions as discussed in https://github.com/joshr120/PD-Stepper/issues/23#issuecomment-2678120522).

With some added debugging, I've seen the call to `this->send_initial_state()` in `MQTTComponent::call_loop()` taking over 20 seconds.
```diff
--- a/esphome/components/mqtt/mqtt_component.cpp
+++ b/esphome/components/mqtt/mqtt_component.cpp
@@ -261,7 +261,10 @@ void MQTTComponent::call_loop() {
   if (this->is_internal())
     return;
 
+  uint32_t times[4];
+  times[0] = millis();
   this->loop();
+  times[1] = millis();
 
   if (!this->resend_state_ || !this->is_connected_()) {
     return;
@@ -273,9 +276,16 @@ void MQTTComponent::call_loop() {
       this->schedule_resend_state();
     }
   }
+  times[2] = millis();
   if (!this->send_initial_state()) {
     this->schedule_resend_state();
   }
+  times[3] = millis();
+  if (times[3] > times[0] + 30) {
+	 ESP_LOGE(TAG, "call_loop() took too long: %d ms (%d + %d + %d)",
+		  times[3] - times[0], times[1] - times[0],
+		  times[2] - times[1], times[3] - times[2]);
+  }
 }
 void MQTTComponent::call_dump_config() {
   if (this->is_internal())

```

```
Feb 24 13:13:01 landing-blind-3 mqtt.component ﻿[E][mqtt.component:284]: call_loop() took too long: 389 ms (0 + 388 + 1)
Feb 24 13:13:01 landing-blind-3 component ﻿[W][component:237]: Component mqtt took a long time for an operation (392 ms).
Feb 24 13:13:01 landing-blind-3 component ﻿[W][component:238]: Components should block for at most 30 ms.
Feb 24 13:13:20 landing-blind-3 esp-idf ﻿[I][esp-idf:000]: #033[0;33mW (515582) transport_base: Poll timeout or error, errno=No more processes, fd=55, timeout_ms=10000#033[0m
Feb 24 13:13:20 landing-blind-3 esp-idf ﻿[I][esp-idf:000]: #033[0;31mE (525587) mqtt_client: Writing didn't complete in specified timeout: errno=11#033[0m
Feb 24 13:13:20 landing-blind-3 esp-idf ﻿[I][esp-idf:000]: #033[0;33mW (525589) mqtt_client: Publish: Losing qos0 data when client not connected#033[0m
Feb 24 13:13:20 landing-blind-3 esp-idf ﻿[I][esp-idf:000]: #033[0;33mW (525591) mqtt_client: Publish: Losing qos0 data when client not connected#033[0m
Feb 24 13:13:20 landing-blind-3 component ﻿[W][component:157]: Component mqtt set Warning flag: unspecified
Feb 24 13:13:20 landing-blind-3 mqtt.component ﻿[E][mqtt.component:284]: call_loop() took too long: 20019 ms (0 + 3 + 20016)
Feb 24 13:13:20 landing-blind-3 component ﻿[W][component:237]: Component mqtt took a long time for an operation (20022 ms).
Feb 24 13:13:20 landing-blind-3 component ﻿[W][component:238]: Components should block for at most 30 ms.
```

That log seems to show `this->send_discovery_()` taking 388ms, followed by `this->send_initial_state()` taking 20016ms. It seems the latter is actually waiting for a network write to complete, which we should never do synchronously from the loop.

2025.3.0-dev

pip

_No response_

ESP32-IDF

### Board

 https://github.com/joshr120/PD-Stepper

mqtt

### YAML Config

```yaml

```

```txt

```

### Additional information

_No response_

dwmw2:
The ESP-IDF MQTT component is fairly unusable for low-latency setups such as ESPHome. By default, ESPHome calls esp_mqtt_client_publish() directly from the MQTT component's loop() function, or when publishing status updates for other components. This may block for up to 20 seconds(!!) in adverse network conditions.

With the `idf_send_async` option, subscribe and unsubscribe requests can still block the loop thread for multiple seconds, but publishing sensor updates is queued for esp-mqtt's own thread to actually send them. Which it does very slowly, no more than one per second, as discussed in https://github.com/espressif/esp-idf/issues/15458

And to top it all off, even with `idf_send_async` set, the so-called 'asynchronous' send can still block for ten seconds because it takes the same MQTT_API_LOCK that the esp-mqtt thread holds while it runs its loop and a network send is timing out. This is reported in https://github.com/espressif/esp-idf/issues/13078

The only way I can see to use esp-mqtt sanely is to use a thread of our own, queueing all sub/unsub/publish requests and invoking the esp-mqtt APIs from that thread.

So replace the existing `idf_send_async` behaviour, using a new thread of our own and a queue.

Fixes: https://github.com/esphome/issues/issues/6810

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/6810

- https://github.com/esphome/esphome-docs/pull/4714

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


dwmw2:
~~Obviously this just hard-codes the new mode and drops *both* the old behaviours either with or without `idf_send_async`. We'll want to decide if we actually want to do that, or whether we want to keep either or both of the existing (broken in different ways) modes.~~

This changes the behaviour of the `idf_send_async` option.

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -160,6 +197,31 @@ class MQTTBackendESP32 final : public MQTTBackend {
...
+    if (payload && len) {
+      elem.payload = (char *) malloc(len);  // NOLINT
+      elem.payload_len = len;
+      memcpy(elem.payload, payload, len);
+    } else {
+      elem.payload = NULL;
+      elem.payload_len = 0;
+    }
+
+    return xQueueSend(this->mqtt_queue_, &elem, 20);
```
patagonaa: Just looked at this code again and found two things here:
1. the timeout should probably be given in ms explicitly by using `pdMS_TO_TICKS()`
2. in the case where the queue is full and the timeout is exceeded, the topic and payload are never freed.

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -160,6 +198,37 @@ class MQTTBackendESP32 final : public MQTTBackend {
   optional<std::string> cl_certificate_;
   optional<std::string> cl_key_;
   bool skip_cert_cn_check_{false};
+#if defined(USE_MQTT_IDF_ENQUEUE)
+  static void esphome_mqtt_task(void *params);
+  QueueHandle_t mqtt_queue_;
+  TaskHandle_t task_handle_;
+  bool enqueue_(esp_mqtt_event_id_t type, const char *topic, int qos = 0, bool retain = false,
+                const char *payload = NULL, size_t len = 0) {
```
jesserockz: Body should be in the cpp file please, not the h file

jesserockz (dismissed):

esphome/components/mqtt/mqtt_backend_esp32.h
DjordjeMandic: Deconstructor should also be moved then, right?

dwmw2: ... and the publish/subscribe/unsubscribe methods? Or are those OK because they're smaller?

jesserockz: Ideally everything that is not a simple one-line setter should be moved.

esphome/components/mqtt/mqtt_backend_esp32.cpp
```diff
@@ -94,6 +94,14 @@ bool MQTTBackendESP32::initialize_() {
   } else {
     mqtt_cfg_.broker.address.transport = MQTT_TRANSPORT_OVER_TCP;
   }
+#endif
+#if defined(USE_MQTT_IDF_ENQUEUE)
+  this->mqtt_queue_ = xQueueCreate(MQTT_QUEUE_LENGTH, sizeof(struct QueueElement));
+  xTaskCreate(esphome_mqtt_task, "esphome_mqtt", TASK_STACK_SIZE, (void *) this, TASK_PRIORITY, &this->task_handle_);
```
bdraco: Long running task here will use another 4KB of memory

bdraco:
A lock-free **SPSC (Single Producer Single Consumer)** queue using atomics would be a better fit here because:

1. **Perfect pattern fit**: The architecture is a classic SPSC setup. The main loop is the sole producer, and the MQTT task is the sole consumer.  
2. **Lock-free performance**: Avoids mutexes entirely, relying only on atomic operations for synchronization.  
3. **Lower latency**: No blocking behavior during enqueue/dequeue operations.  

In contrast, the current FreeRTOS `xQueue` implementation introduces unnecessary overhead:

- Every queue operation involves mutex/semaphore handling  
- Higher cost due to system calls and potential task switches  
- Additional latency that could be avoided with a lock-free structure

bdraco:
Also for benchmarking options I use something like this
```yaml
api:
  services:
    - service: measure_component_memory
      then:
        - lambda: |-
            // Measure Component error_message_ memory impact
            ESP_LOGI("memory_test", "=== Component Error Message Memory Test ===");
            
            // Count total components - estimate based on this config
            // This config has approximately 200+ sensors plus other components
            size_t component_count = 250;  // Rough estimate for this config
            ESP_LOGI("memory_test", "Estimated components: %zu", component_count);
            
            // Get heap before forcing errors
            size_t heap_before = heap_caps_get_free_size(MALLOC_CAP_INTERNAL);
            ESP_LOGI("memory_test", "Free heap before errors: %zu bytes", heap_before);
            
            // Force some components to have errors (testing string storage)
            // Note: We can't actually set errors on arbitrary components safely,
            // but we can measure the struct sizes
            
            // Measure Component base size
            ESP_LOGI("memory_test", "sizeof(Component): %zu bytes", sizeof(Component));
            ESP_LOGI("memory_test", "sizeof(const char*): %zu bytes", sizeof(const char*));
            ESP_LOGI("memory_test", "sizeof(std::string): %zu bytes", sizeof(std::string));
            ESP_LOGI("memory_test", "sizeof(uint16_t): %zu bytes", sizeof(uint16_t));
            ESP_LOGI("memory_test", "sizeof(uint32_t): %zu bytes", sizeof(uint32_t));
            
            // Theoretical savings calculation
            size_t old_error_size = sizeof(std::string);  // Was std::string
            size_t new_error_size = sizeof(const char*);  // Now const char*
            size_t savings_per_component = old_error_size - new_error_size;
            size_t total_savings = savings_per_component * component_count;
            
            ESP_LOGI("memory_test", "Memory savings per component: %zu bytes", savings_per_component);
            ESP_LOGI("memory_test", "Total theoretical savings: %zu bytes", total_savings);
            
            // Get heap after
            size_t heap_after = heap_caps_get_free_size(MALLOC_CAP_INTERNAL);
            ESP_LOGI("memory_test", "Free heap after: %zu bytes", heap_after);
            ESP_LOGI("memory_test", "=== End Memory Test ===");
    - service: test_error_memory_usage
      then:
        - lambda: |-
            // Test actual memory usage with error messages
            ESP_LOGI("memory_test", "=== Testing Error Message Memory Usage ===");
            
            // Test with ESP32's std::string implementation
            class TestOldComponent {
            public:
              std::string error_message_{};
            };
            
            class TestNewComponent {
            public:
              const char *error_message_{nullptr};
            };
            
            ESP_LOGI("memory_test", "Old implementation (std::string): %zu bytes", sizeof(TestOldComponent));
            ESP_LOGI("memory_test", "New implementation (const char*): %zu bytes", sizeof(TestNewComponent));
            ESP_LOGI("memory_test", "Savings per component: %zu bytes", 
                     sizeof(TestOldComponent) - sizeof(TestNewComponent));
            
            // Test heap usage with actual string assignment
            size_t heap_start = heap_caps_get_free_size(MALLOC_CAP_INTERNAL);
            
            // Old way - allocates heap memory
            TestOldComponent old_comp;
            old_comp.error_message_ = "Connection timeout error";
            size_t heap_after_old = heap_caps_get_free_size(MALLOC_CAP_INTERNAL);
            
            // New way - no heap allocation for string literals
            TestNewComponent new_comp;
            new_comp.error_message_ = "Connection timeout error";
            size_t heap_after_new = heap_caps_get_free_size(MALLOC_CAP_INTERNAL);
            
            ESP_LOGI("memory_test", "Heap used by old method: %zu bytes", 
                     heap_start - heap_after_old);
            ESP_LOGI("memory_test", "Heap used by new method: %zu bytes", 
                     heap_after_old - heap_after_new);
            
            ESP_LOGI("memory_test", "=== End Test ===");
    - service: measure_socket
      then:
        - lambda: |-
            #include "lwip/sockets.h"
            #include <netinet/in.h>
            #include <fcntl.h>
            #include <unistd.h>
            #include <errno.h>

            const int sock_count = 5;
            const int base_port = 6000;
            int listen_fds[sock_count];

            for (int i = 0; i < sock_count; ++i) {
              int fd = lwip_socket(AF_INET, SOCK_STREAM, 0);
              if (fd < 0) {
                ESP_LOGE("benchmark", "socket() failed");
                return;
              }

              fcntl(fd, F_SETFL, O_NONBLOCK);

              struct sockaddr_in addr = {
                .sin_family = AF_INET,
                .sin_port = htons(base_port + i),
                .sin_addr = { .s_addr = INADDR_ANY }
              };

              if (lwip_bind(fd, (struct sockaddr *)&addr, sizeof(addr)) < 0) {
                ESP_LOGE("benchmark", "bind() failed on port %d: %s", base_port + i, strerror(errno));
                for (int j = 0; j <= i; ++j) lwip_close(listen_fds[j]);
                return;
              }

              if (lwip_listen(fd, 1) < 0) {
                ESP_LOGE("benchmark", "listen() failed on port %d", base_port + i);
                for (int j = 0; j <= i; ++j) lwip_close(listen_fds[j]);
                return;
              }

              listen_fds[i] = fd;
              ESP_LOGI("benchmark", "Listening on port %d", base_port + i);
            }

            // === Benchmark lwip_select() with timeout = 0 ===
            {
              fd_set rfds;
              FD_ZERO(&rfds);
              int maxfd = 0;
              for (int i = 0; i < sock_count; ++i) {
                FD_SET(listen_fds[i], &rfds);
                if (listen_fds[i] > maxfd) maxfd = listen_fds[i];
              }

              struct timeval timeout = { .tv_sec = 0, .tv_usec = 0 };
              unsigned long start = micros();
              int ret = lwip_select(maxfd + 1, &rfds, nullptr, nullptr, &timeout);
              unsigned long end = micros();

              ESP_LOGI("benchmark", "[lwip_select] ret=%d, duration=%lu us", ret, end - start);
            }

            // === Benchmark lwip_recv(sock, NULL, 0) on each socket ===
            {
              ESP_LOGI("benchmark", "Calling lwip_recv(sock, NULL, 0) on each socket...");
              unsigned long start = micros();
              for (int i = 0; i < sock_count; ++i) {
                char dummy;
                int ret = lwip_recv(listen_fds[i], &dummy, 0, 0);  // will fail
                if (ret >= 0) {
                  ESP_LOGW("benchmark", "unexpected recv success on port %d", base_port + i);
                }
              }
              unsigned long end = micros();
              ESP_LOGI("benchmark", "[recv(0)] duration=%lu us", end - start);
            }

            // Cleanup
            for (int i = 0; i < sock_count; ++i) {
              lwip_close(listen_fds[i]);
            }
    - service: measure_idle_loop
      then:
        - lambda: |-
            const int iterations = 10000;
            uint32_t start = micros();
            for (int i = 0; i < iterations; ++i) {
              App.loop();  // this is the ESPHome core loop
            }
            uint32_t end = micros();
            float avg_us = (end - start) / float(iterations);
            ESP_LOGI("benchmark", "Average App.loop() call: %.3f µs", avg_us);    
    - service: run_timer_benchmark
      then:
        - lambda: |-
            ESP_LOGI("benchmark", "Running timer benchmark...");

            const int iterations = 100000;
            volatile uint32_t dummy = 0;

            struct Timer {
              uint32_t cached_value;
              uint32_t get() const { return cached_value; }
            } timer;

            // Warm-up
            timer.cached_value = millis();
            dummy = timer.get();
            dummy = millis();

            // millis() timing
            int64_t start = micros();
            for (int i = 0; i < iterations; ++i) {
              dummy = millis();
            }
            int64_t end = micros();
            double millis_us = (end - start) / static_cast<double>(iterations);

            // Cached member access timing
            timer.cached_value = 123456789;
            start = micros();
            for (int i = 0; i < iterations; ++i) {
              dummy = timer.get();
            }
            end = micros();
            double cached_us = (end - start) / static_cast<double>(iterations);

            ESP_LOGI("benchmark", "millis():                   %.3f µs per call", millis_us);
            ESP_LOGI("benchmark", "Cached class member access: %.6f µs per call", cached_us);
    - service: check_atomic_lock_free
      then:
        - lambda: |-
            #include <atomic>
            ESP_LOGI("atomic_test", "=== Atomic Lock-Free Status Check on ESP32 ===");
            ESP_LOGI("atomic_test", "");
            
            // Platform-specific constants (compile-time)
            ESP_LOGI("atomic_test", "Platform constants:");
            ESP_LOGI("atomic_test", "  ATOMIC_BOOL_LOCK_FREE: %d", ATOMIC_BOOL_LOCK_FREE);
            ESP_LOGI("atomic_test", "  ATOMIC_CHAR_LOCK_FREE: %d", ATOMIC_CHAR_LOCK_FREE);
            ESP_LOGI("atomic_test", "  ATOMIC_SHORT_LOCK_FREE: %d", ATOMIC_SHORT_LOCK_FREE);
            ESP_LOGI("atomic_test", "  ATOMIC_INT_LOCK_FREE: %d", ATOMIC_INT_LOCK_FREE);
            ESP_LOGI("atomic_test", "  ATOMIC_LONG_LOCK_FREE: %d", ATOMIC_LONG_LOCK_FREE);
            ESP_LOGI("atomic_test", "  ATOMIC_LLONG_LOCK_FREE: %d", ATOMIC_LLONG_LOCK_FREE);
            ESP_LOGI("atomic_test", "  ATOMIC_POINTER_LOCK_FREE: %d", ATOMIC_POINTER_LOCK_FREE);
            ESP_LOGI("atomic_test", "");
            ESP_LOGI("atomic_test", "Note: 0 = never lock-free, 1 = sometimes lock-free, 2 = always lock-free");
            ESP_LOGI("atomic_test", "");
            
            // Check specific atomic types used in ESPHome based on size
            ESP_LOGI("atomic_test", "Atomic type analysis (based on ATOMIC_*_LOCK_FREE macros):");
            
            // uint8_t atomics (used in BLE queue head/tail)
            ESP_LOGI("atomic_test", "  std::atomic<uint8_t>: %s (ATOMIC_CHAR_LOCK_FREE = %d)",
                     ATOMIC_CHAR_LOCK_FREE == 2 ? "ALWAYS LOCK-FREE" : 
                     ATOMIC_CHAR_LOCK_FREE == 1 ? "SOMETIMES LOCK-FREE" : "NEVER LOCK-FREE",
                     ATOMIC_CHAR_LOCK_FREE);
            
            // uint16_t atomics (used in BLE queue dropped_count and logger message_counter)
            ESP_LOGI("atomic_test", "  std::atomic<uint16_t>: %s (ATOMIC_SHORT_LOCK_FREE = %d)",
                     ATOMIC_SHORT_LOCK_FREE == 2 ? "ALWAYS LOCK-FREE" : 
                     ATOMIC_SHORT_LOCK_FREE == 1 ? "SOMETIMES LOCK-FREE" : "NEVER LOCK-FREE",
                     ATOMIC_SHORT_LOCK_FREE);
            
            // size_t atomics (used in BLE tracker ring indices)
            // size_t is typically 32-bit on ESP32
            ESP_LOGI("atomic_test", "  std::atomic<size_t> (32-bit): %s (ATOMIC_INT_LOCK_FREE = %d)",
                     ATOMIC_INT_LOCK_FREE == 2 ? "ALWAYS LOCK-FREE" : 
                     ATOMIC_INT_LOCK_FREE == 1 ? "SOMETIMES LOCK-FREE" : "NEVER LOCK-FREE",
                     ATOMIC_INT_LOCK_FREE);
            
            // int atomics (used in HTTP request socket_t)
            ESP_LOGI("atomic_test", "  std::atomic<int>: %s (ATOMIC_INT_LOCK_FREE = %d)",
                     ATOMIC_INT_LOCK_FREE == 2 ? "ALWAYS LOCK-FREE" : 
                     ATOMIC_INT_LOCK_FREE == 1 ? "SOMETIMES LOCK-FREE" : "NEVER LOCK-FREE",
                     ATOMIC_INT_LOCK_FREE);
            
            // bool atomics (used in HTTP request flags)
            ESP_LOGI("atomic_test", "  std::atomic<bool>: %s (ATOMIC_BOOL_LOCK_FREE = %d)",
                     ATOMIC_BOOL_LOCK_FREE == 2 ? "ALWAYS LOCK-FREE" : 
                     ATOMIC_BOOL_LOCK_FREE == 1 ? "SOMETIMES LOCK-FREE" : "NEVER LOCK-FREE",
                     ATOMIC_BOOL_LOCK_FREE);
            
            // pointer atomics
            ESP_LOGI("atomic_test", "  std::atomic<void*>: %s (ATOMIC_POINTER_LOCK_FREE = %d)",
                     ATOMIC_POINTER_LOCK_FREE == 2 ? "ALWAYS LOCK-FREE" : 
                     ATOMIC_POINTER_LOCK_FREE == 1 ? "SOMETIMES LOCK-FREE" : "NEVER LOCK-FREE",
                     ATOMIC_POINTER_LOCK_FREE);
            
            ESP_LOGI("atomic_test", "");
            ESP_LOGI("atomic_test", "ESP32 Architecture info:");
            ESP_LOGI("atomic_test", "  sizeof(size_t): %zu bytes", sizeof(size_t));
            ESP_LOGI("atomic_test", "  sizeof(int): %zu bytes", sizeof(int));
            ESP_LOGI("atomic_test", "  sizeof(void*): %zu bytes", sizeof(void*));
            
            ESP_LOGI("atomic_test", "=== End Atomic Test ===");
    - service: benchmark_ringbuffer_comparison
      then:
        - lambda: |-
            #include <atomic>
            #include <freertos/FreeRTOS.h>
            #include <freertos/ringbuf.h>
            
            ESP_LOGI("benchmark", "=== Ring Buffer Performance Comparison ===");
            
            // Simplified test structure
            struct TestItem {
                uint8_t data[64];  // Similar size to BLE scan result
                uint32_t timestamp;
            };
            
            const int iterations = 10000;
            const int buffer_items = 32;
            
            uint32_t atomic_time = 0;
            uint32_t rtos_time = 0;
            
            // Test 1: Current atomic-based approach
            {
                class AtomicBuffer {
                    TestItem buffer_[32];
                    std::atomic<uint8_t> write_idx_{0};
                    std::atomic<uint8_t> read_idx_{0};
                public:
                    bool push(const TestItem& item) {
                        uint8_t write = write_idx_.load(std::memory_order_relaxed);
                        uint8_t next = (write + 1) % 32;
                        uint8_t read = read_idx_.load(std::memory_order_acquire);
                        if (next != read) {
                            buffer_[write] = item;
                            write_idx_.store(next, std::memory_order_release);
                            return true;
                        }
                        return false;
                    }
                    bool pop(TestItem& item) {
                        uint8_t read = read_idx_.load(std::memory_order_relaxed);
                        uint8_t write = write_idx_.load(std::memory_order_acquire);
                        if (read != write) {
                            item = buffer_[read];
                            read_idx_.store((read + 1) % 32, std::memory_order_release);
                            return true;
                        }
                        return false;
                    }
                };
                
                AtomicBuffer atomic_buf;
                TestItem test_item = {};
                test_item.timestamp = 12345;
                
                uint32_t start = micros();
                for (int i = 0; i < iterations; i++) {
                    atomic_buf.push(test_item);
                    TestItem item;
                    atomic_buf.pop(item);
                }
                atomic_time = micros() - start;
                
                ESP_LOGI("benchmark", "Atomic Ring Buffer:");
                ESP_LOGI("benchmark", "  Total time: %u µs", atomic_time);
                ESP_LOGI("benchmark", "  Per push+pop: %.2f µs", atomic_time / (float)iterations);
            }
            
            // Test 2: FreeRTOS ring buffer
            {
                RingbufHandle_t ring_buf = xRingbufferCreate(
                    buffer_items * sizeof(TestItem), 
                    RINGBUF_TYPE_NOSPLIT
                );
                
                if (ring_buf == nullptr) {
                    ESP_LOGE("benchmark", "Failed to create FreeRTOS ring buffer");
                    return;
                }
                
                TestItem test_item = {};
                test_item.timestamp = 12345;
                
                uint32_t start = micros();
                for (int i = 0; i < iterations; i++) {
                    xRingbufferSend(ring_buf, &test_item, sizeof(test_item), 0);
                    size_t item_size;
                    TestItem* item = (TestItem*)xRingbufferReceive(ring_buf, &item_size, 0);
                    if (item) {
                        vRingbufferReturnItem(ring_buf, item);
                    }
                }
                rtos_time = micros() - start;
                
                vRingbufferDelete(ring_buf);
                
                ESP_LOGI("benchmark", "FreeRTOS Ring Buffer:");
                ESP_LOGI("benchmark", "  Total time: %u µs", rtos_time);
                ESP_LOGI("benchmark", "  Per push+pop: %.2f µs", rtos_time / (float)iterations);
            }
            
            ESP_LOGI("benchmark", "");
            ESP_LOGI("benchmark", "Performance ratio: %.2fx", 
                     (float)rtos_time / (float)atomic_time);
            ESP_LOGI("benchmark", "Atomic is %s", 
                     atomic_time < rtos_time ? "FASTER" : "SLOWER");
            
            ESP_LOGI("benchmark", "=== End Benchmark ===");

```

dwmw2:
> A lock-free **SPSC (Single Producer Single Consumer)** queue using atomics would be a better fit here because:

Thanks. Do we have an implementation of this in ESPHome already or should I create/import one from somewhere?

esphome/components/mqtt/mqtt_backend_esp32.cpp
dwmw2: Yes, but unless we're going to ditch the esp-idf MQTT implementation completely, it seems unavoidable.

bdraco:
esp32_ble_tracker and esp32_ble have example implementations

esphome/components/mqtt/mqtt_backend_esp32.cpp
bdraco: You might be able to reduce the task stack size a bit with some careful testing

dwmw2:
Converted to LockFreeQueue in commit [ba05cac](https://github.com/esphome/esphome/commit/ba05cac936f9c3f57526425394098ca9184c54b8) but untested as I'm travelling this week.

I also now want to double-check that all the subscribe/unsubscribe/publish events *are* coming from the main thread, as we now depend on that. I might even implement a sanity check for it.

dwmw2:
We should also think about making the LockFreeQueue and the EventPool more generic and sharing them...

bdraco:
If you didn't need to make any changes, it might make sense to make it a core helper

dwmw2:
> If you didn't need to make any changes, it might make sense to make it a core helper

I turned the event pool into a *template* but other than that, no changes.

bdraco:
> > If you didn't need to make any changes, it might make sense to make it a core helper
> 
> 
> 
> I turned the event pool into a *template* but other than that, no changes.

If you want to do a PR to make it a core helper ahead of this one, that would be great as it will reduce the review burden here

dwmw2:
Sure.. but I'll actually test this first, and make sure it's even going to be useful. As I came to actually write down the part about MQTT messages only being queued from the main thread, I found I wanted to *prove* that first. If *not*, then it was actually OK using the FreeRTOS queue before.

bdraco:
> Sure.. but I'll actually test this first, and make sure it's even going to be useful. As I came to actually write down the part about MQTT messages only being queued from the main thread, I found I wanted to *prove* that first. If *not*, then it was actually OK using the FreeRTOS queue before. 

No worries. Take your time.

esphome/core/lock_free_queue.h
```diff
@@ -37,6 +39,12 @@ template<class T, uint8_t SIZE> class LockFreeQueue {
 
     buffer_[current_tail] = element;
     tail_.store(next_tail, std::memory_order_release);
+
+    // Notify the consumer task if set (optional)
+    if (task_to_notify_ != nullptr) {
```
dwmw2: You don't need to notify *every* time; only when the consumer had actually stopped (because the cons pointer had reached the previous value of the prod pointer before you updated the latter?)

bdraco: We can't reliably know the consumer's state without additional synchronization. 

The consumer might be processing the last item, about to check the queue again, or just entering sleep. Any attempt to track this would create race conditions and complexity that defeats the purpose of a lock-free design. FreeRTOS task notifications are specifically designed for high-frequency producer-consumer patterns like this as they're just an atomic bit operation that's essentially free when the task is already awake, and multiple notifications safely coalesce.

Optimization would require state tracking that's likely more expensive than the notification itself, while risking missed wake-ups in edge cases. The current approach of notifying on every push is both the simplest and most reliable pattern.

bdraco:
> Sure.. but I'll actually test this first, and make sure it's even going to be useful. As I came to actually write down the part about MQTT messages only being queued from the main thread, I found I wanted to _prove_ that first. If _not_, then it was actually OK using the FreeRTOS queue before.

AFAIK the only thing that generates messages in other tasks is logging, and we fixed that in the logger itself with a ring buffer a few versions back to make sure the logging happens in the main loop since it was a problem for API as well. I'd expect that if its safe for API now it should be safe for MQTT as well or we have a bigger problem.

esphome/core/lock_free_queue.h
dwmw2: It's OK to have false positives with the wakeup, just not false negatives. So the way to think about the optimisation is to ask when you can be *sure* you don't need to send the wakeup.

You can read `head_` before incrementing `tail_`. If they match (i.e. `empty()` would return true) then you'll have to do the wakeup because the consumer thread might be on the way into sleep.

After you increment `tail_`, read `head_` again. If it didn't match before, and if it still hasn't reached the *previous* version of `head_` then there is no chance that the consumer thread is going into sleep. It's still at least one element behind and there's no chance of it thinking that it was caught up. So you don't need the wakeup in that case.

bdraco: Thats bit more complex but manageable. I pushed a change that implements that.  I'm not sure if its better or worse performance wise though since we have to do another .load to know

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -97,7 +103,7 @@ class MQTTBackendESP32 final : public MQTTBackend {
   static const size_t MQTT_BUFFER_SIZE = 4096;
   static const size_t TASK_STACK_SIZE = 4096;
   static const ssize_t TASK_PRIORITY = 5;
-  static const uint32_t MQTT_QUEUE_LENGTH = 20;
+  static const uint8_t MQTT_QUEUE_LENGTH = 30;  // Was 20*24 bytes = 480, now 30*16 bytes = 480
```
dwmw2: You make it sound like I didn't just pull that number out of my posterior in the first place... :)

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -50,14 +50,20 @@ struct QueueElement {
   esp_mqtt_event_id_t type;
   char *topic;
   char *payload;
-  uint32_t payload_len;
-  int qos;
-  bool retain;
+  uint16_t payload_len;  // MQTT max payload is 64KB
```
dwmw2: 64KiB

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -50,14 +50,20 @@ struct QueueElement {
...
+  uint8_t qos : 2;       // QoS only needs values 0-2
+  uint8_t retain : 1;
+  uint8_t reserved : 5;  // Reserved for future use
```
dwmw2: You can probably put `type` in here too; we only need about three values for that and it doesn't *have* to actually match the `esp_mqtt_event_id_t`; that was just the easy thing to do.

dwmw2:
Seems to be working nicely now with the lock free queue. Thanks for the help!

bdraco:
extracted in https://github.com/esphome/esphome/pull/9238

bdraco:
looks like everything is going to pass. I'll backmerge the other branch here to reduce conflicts later

esphome/components/mqtt/mqtt_backend_esp32.cpp
bdraco: I think we could go to 2048 on the task stack here as the task isn't doing anything too complex.

It will need some testing though

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -42,9 +46,71 @@ struct Event {
...
+  uint8_t qos : 2;  // QoS only needs values 0-2
+  uint8_t retain : 1;
+  uint8_t reserved : 3;  // Reserved for future use
+
+  QueueElement() : topic(nullptr), payload(nullptr), payload_len(0), qos(0), retain(0), reserved(0) {}
+
+  // Helper to set topic/payload (handles malloc)
+  bool set_data(const char *topic_str, const char *payload_data, size_t len) {
+    // Check payload size limit (MQTT max is 64KiB)
+    if (len > UINT16_MAX) {
```
bdraco:
```suggestion
    if (len > std::numeric_limits<uint16_t>::max()) {
```

bdraco: will need #include <limits>

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -42,9 +46,71 @@ struct Event {
...
+      payload = nullptr;
+    }
+    payload_len = 0;
+  }
+};
+
 class MQTTBackendESP32 final : public MQTTBackend {
  public:
   static const size_t MQTT_BUFFER_SIZE = 4096;
+  static const size_t TASK_STACK_SIZE = 4096;
```
bdraco:
```suggestion
  static const size_t TASK_STACK_SIZE = 2048;
```

I think we can get away with 2048

bdraco:
I think this is good to go once that above comments are addressed

bdraco:
Looks good. Let me do a final review

esphome/components/mqtt/mqtt_backend_esp32.h
```diff
@@ -42,9 +46,71 @@ struct Event {
...
+      payload = nullptr;
+    }
+    payload_len = 0;
+  }
+};
+
 class MQTTBackendESP32 final : public MQTTBackend {
  public:
   static const size_t MQTT_BUFFER_SIZE = 4096;
+  static const size_t TASK_STACK_SIZE = 2048;
```
bdraco, line 119: Just thought about MQTT might use TLS. need to check if thats supported. We might need a larger stack for that 😢

bdraco (dismissed):

esphome/components/mqtt/mqtt_backend_esp32.cpp
```diff
@@ -205,16 +205,6 @@ void MQTTBackendESP32::esphome_mqtt_task(void *params) {
     // Wait for notification indefinitely
     ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
 
-    // Log dropped messages periodically
-    uint16_t dropped = this_mqtt->mqtt_queue_.get_and_reset_dropped_count();
-    if (dropped > 0) {
-      ESP_LOGW(TAG, "Dropped %u outbound MQTT messages due to buffer overflow", dropped);
```
dwmw2: Hm, I moved that here because logging from the producer side would cause more messages which would cause more logging.... I believe that's why I was seeing a huge number of "1 message dropped" logs followed by a watchdog reset...

bdraco: Ideally we want to log in the main loop since if we log in the task it has to transfer the message back to the main loop and log it there via a ring buffer in the task logger.

dwmw2: We could rate-limit it to avoid causing a new warning message for each dropped (warning) message.

Or we could avoid emitting the warning when the topic is the log topic?

bdraco (dismissed):

esphome/components/mqtt/mqtt_backend_esp32.cpp
bdraco: Should be fixed in 9c9f0d1749596551458ffa4244e73fc59f78bcc2

dwmw2: LGTM, thanks.

bdraco (approved):

bdraco:
I thought about this some more overnight and made a few more tweaks

bdraco:
I'm mildly worried about the heap fragmentation from the heap allocations, but given the sizes of the messages vary so much its probably still the best choice

bdraco:
We could do a ring buffer like logger but than we would have to pre-allocate for worst case..  this seems like the most practical solution for MQTT's highly variable message sizes.  We can always revisit later if needed.

bdraco (approved):

randybb:
What a change! MQTT is not blocking on boot anymore, connects instantly, no more "Component mqtt took a long time for an operation" :) thanks!

jesserockz (approved):

dwmw2:
```
[11:16:24][D][sensor:104]: 'StallGuard result': Sending state 128.00000  with 0 decimals of accuracy
[11:16:24][I][main:369]: Reed switch off at -561073.000000 in direction 0, travel IDLE!
[11:16:24][D][binary_sensor:026]: 'Reed Switch': New state is OFF
[11:16:25][W][component:255]: Component mqtt cleared Warning flag
[11:16:25][D][status_led:030]: Restoring light state ON
[11:16:31][W][mqtt.idf:149]: Dropped 255 messages (10s)
[11:16:32][D][sensor:104]: 'VBUS Voltage': Sending state 5.23057 V with 2 decimals of accuracy

[11:16:33]***ERROR*** A stack overflow in task esphome_mqtt has been detected.


[11:16:33]Backtrace: 0x40376d56:0x3fcc00b0 0x4037ce41:0x3fcc00d0 0x4037dd96:0x3fcc00f0 0x4037ec33:0x3fcc0170 0x4037dec8:0x3fcc01a0 0x4037debe:0x00000001 |<-CORRUPTED
WARNING Found stack trace! Trying to decode it
WARNING Decoded 0x40376d56: panic_abort at /home/ANT.AMAZON.COM/dwmw/.platformio/packages/framework-espidf@src-c44434145e05010467d5d5a727b42ef9/components/esp_system/panic.c:463
WARNING Decoded 0x4037ce41: esp_system_abort at /home/ANT.AMAZON.COM/dwmw/.platformio/packages/framework-espidf@src-c44434145e05010467d5d5a727b42ef9/components/esp_system/port/esp_system_chip.c:92
WARNING Decoded 0x4037dd96: vApplicationStackOverflowHook at /home/ANT.AMAZON.COM/dwmw/.platformio/packages/framework-espidf@src-c44434145e05010467d5d5a727b42ef9/components/freertos/FreeRTOS-Kernel/portable/xtensa/port.c:563
WARNING Decoded 0x4037ec33: vTaskSwitchContext at /home/ANT.AMAZON.COM/dwmw/.platformio/packages/framework-espidf@src-c44434145e05010467d5d5a727b42ef9/components/freertos/FreeRTOS-Kernel/tasks.c:3701 (discriminator 7)
WARNING Decoded 0x4037dec8: _frxt_dispatch at /home/ANT.AMAZON.COM/dwmw/.platformio/packages/framework-espidf@src-c44434145e05010467d5d5a727b42ef9/components/freertos/FreeRTOS-Kernel/portable/xtensa/portasm.S:451
WARNING Decoded 0x4037debe: _frxt_int_exit at /home/ANT.AMAZON.COM/dwmw/.platformio/packages/framework-espidf@src-c44434145e05010467d5d5a727b42ef9/components/freertos/FreeRTOS-Kernel/portable/xtensa/portasm.S:246




[11:16:35]ELF file SHA256: cf6da77a1

[11:16:35]Rebooting...
[11:16:35]ESP-ROM:esp32s3-20210327
[11:16:35]Build:Mar 27 2021
[11:16:35]rst:0xc (RTC_SW_CPU_RST),boot:0x28 (SPI_FAST_FLASH_BOOT)
[11:16:35]Saved PC:0x40376c95
WARNING Decoded 0x40376c95: esp_restart_noos at /home/ANT.AMAZON.COM/dwmw/.platformio/packages/framework-espidf@src-c44434145e05010467d5d5a727b42ef9/components/esp_system/port/soc/esp32s3/system_internal.c:158
[11:16:35]SPIWP:0xee

```

dwmw2:
```
[11:36:29][W][component:255]: Component mqtt cleared Warning flag
[11:36:29][D][status_led:030]: Restoring light state ON
[11:36:31][D][sensor:104]: 'VBUS Voltage': Sending state 5.22209 V with 2 decimals of accuracy
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (106997) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (106998) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (106999) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107000) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107001) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107001) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107002) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107003) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107003) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107004) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107005) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107006) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107006) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107007) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107008) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107009) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107009) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107010) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107011) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107012) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107012) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107013) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107014) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107014) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107015) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107016) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (107017) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][light:036]: 'Status LED' Setting:
[11:36:33][D][light:047]:   State: OFF
[11:36:33][W][mqtt.idf:149]: Dropped 150 messages (10s)
[11:36:33][W][mqtt:352]: Disconnected: TCP disconnected
[11:36:33][W][component:240]: Component mqtt set Warning flag: unspecified
[11:36:33][I][mqtt:262]: Connecting
[11:36:33][D][esp-idf:000]: I (107030) mqtt_client: Client asked to disconnect
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (106996) transport_base: Poll timeout or error, errno=Success, fd=56, timeout_ms=10000
[11:36:33][D][esp-idf:000][esphome_mqtt]: E (106996) mqtt_client: Writing didn't complete in specified timeout: errno=0
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (106997) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (106997) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:33][D][esp-idf:000][esphome_mqtt]: W (106997) mqtt_client: Publish: Losing qos0 data when client not connected
[11:36:41][D][sensor:104]: 'VBUS Voltage': Sending state 5.20514 V with 2 decimals of accuracy
[11:36:48][D][light:036]: 'Status LED' Setting:
[11:36:48][D][light:047]:   State: ON
[11:36:48][W][component:255]: Component mqtt cleared Warning flag
[11:36:48][I][mqtt:303]: Connected
```

dwmw2:
Added a log when the loop drops a message because `!this_mqtt->is_connected_`
```
[12:17:06][D][cover:186]:   Current Operation: IDLE
[12:17:06][W][component:240]: Component mqtt set Warning flag: unspecified
[12:17:07][W][component:255]: Component mqtt cleared Warning flag
[12:17:07][D][status_led:030]: Restoring light state ON
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][D][light:036]: 'Status LED' Setting:
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:09][D][light:047]:   State: OFF
[12:17:09][W][mqtt.idf:149]: Dropped 13 messages (10s)
[12:17:09][W][mqtt:352]: Disconnected: TCP disconnected
[12:17:09][W][component:240]: Component mqtt set Warning flag: unspecified
[12:17:09][I][mqtt:262]: Connecting
[12:17:09][D][esp-idf:000]: I (214296) mqtt_client: Client asked to disconnect
[12:17:09][D][esp-idf:000][mqtt_task]: E (214277) mqtt_client: No PING_RESP, disconnected
[12:17:09][D][esp-idf:000][esphome_mqtt]: W (214279) mqtt_client: Publish: Losing qos0 data when client not connected
[12:17:09][D][esp-idf:000][esphome_mqtt]: W (214280) mqtt_client: Publish: Losing qos0 data when client not connected
[12:17:09][D][esp-idf:000][esphome_mqtt]: W (214280) mqtt_client: Publish: Losing qos0 data when client not connected
[12:17:09][D][esp-idf:000][esphome_mqtt]: W (214280) mqtt_client: Publish: Losing qos0 data when client not connected
[12:17:09][E][mqtt.idf:266][esphome_mqtt]: Drop when not connected
[12:17:16][D][sensor:104]: 'VBUS Voltage': Sending state 5.23057 V with 2 decimals of accuracy
[12:17:19][D][light:036]: 'Status LED' Setting:
[12:17:19][D][light:047]:   State: ON
[12:17:19][W][component:255]: Component mqtt cleared Warning flag
[12:17:19][I][mqtt:303]: Connected

```

dwmw2:
With fixes in #9273 this seems to be working now.

I think there are still issues with MQTT reconnection but those were always there. I'll get to those at some point... I think we should set a flag on the components whose state we failed to send, and resend the *current* state after a reconnect or when the queue becomes empty again.

PR #8214, opened 2025-02-06
[uart] Add packet_transport platform

clydebarrow:
# What does this implement/fix?

Building on #8187 this adds a `packet_transport` platform implementation to the UART component. This allows ESPHome devices to send sensor and binary_data sensor to each other via a serial link, including multi-drop RS485.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#4645

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx
- [x] host

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

    packet_transport:
      platform: uart
      sensors:
        - dht_temp

    uart:
      tx_pin: GPIOXX
      rx_pin: GPIOXX
      baud_rate: 9600

    sensor:
      - platform: dht
          id: dht
          pin: GPIOXX
          temperature:
            name: "Temperature"
            id: dht_temp

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8334, opened 2025-02-27
[usb_uart] Implement USB Host mode UART

clydebarrow:

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4724

## Test Environment

- [x] ESP32-S2
- [x] ESP32-S3
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

usb_uart:
  - id: uart_1
    type: esp_jtag
    channels:
      - id: channel_1
        debug: true
        dummy_receiver: true

```

## Checklist:


functionpointer:
This means ESP32 becomes a USB host? So we can connect a usb-to-serial adapter to it and access it in ESPHome?

clydebarrow:
> So we can connect a usb-to-serial adapter to it and access it in ESPHome?

Yes, that is exactly the purpose. By using a multi-port adapter you can access up to 3 serial ports via USB, in addition to the ESP32's 3 inbuilt UARTs.

functionpointer:
Thats so awesome! I might actually have a use for this, for interfacing with my iCharger X8. Altough that one might be using USB HID.

By multi port adapter you mean a standard usb hub, right?

clydebarrow:
> By multi port adapter you mean a standard usb hub, right?

No, hubs are not supported by ESP-IDF. You need a multi-channel chip like a CH344.

nielsnl68:
finally a PR that i like.

Could you also create a component that could be used as USB_UART client side?

clydebarrow:
> Could you also create a component that could be used as USB_UART client side?

It's certainly possible, but I don't have a need. The main use for USB serial as peripheral is for flashing and logging and that already works.

rubenbaldewsing:
> finally a PR that i like.
> 
> 
> 
> Could you also create a component that could be used as USB_UART client side?

I think you mean this right?: https://github.com/esphome/feature-requests/issues/2924

SeByDocKy:
> CH344

Hasn't the CH344 4 UART ports ?

clydebarrow:
> Hasn't the CH344 4 UART ports ?

Yes, but the ESP32-Sx USB hardware can't make use of more than 3 at one time.

martijnbuts:
Can the USB host mode be combined with https://github.com/esphome/esphome/pull/8214 so you can control 3 extra esp32 over USB to UART and not wifi?

clydebarrow:
> Can the USB host mode be combined with #8214 so you can control 3 extra esp32 over USB to UART and not wifi?

Sure, though with the packet transport over UART you could use RS485 to have multiple devices on a single shared UART bus as well.

p1ngb4ck:
what about FTDI chip support for this ? would that be of interest ? Have some of those lying around here and would be willing to put some effort in.

clydebarrow:

```yaml
external_components:
  - source: github://pr#8334
    components: [usb_host, usb_uart]
    refresh: 1h
```

---
(Added by the PR bot)

clydebarrow:
> what about FTDI chip support for this ? would that be of interest ? Have some of those lying around here and would be willing to put some effort in.

Yeah, sure. Not sure if there is an FTDI doc describing the descriptors.

esphome/components/usb_host/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+        schema = schema.extend({cv.Required(CONF_PID): cv.uint16_t})
+    return schema
+
+
+CONFIG_SCHEMA = cv.COMPONENT_SCHEMA.extend(
+    {
+        cv.GenerateID(): cv.declare_id(USBHost),
+        cv.Optional(CONF_DEVICES): cv.ensure_list(usb_device_schema()),
+    }
+)
```
jesserockz, line 50: Needs to have `cv.only_with_esp_idf` and `esp32.only_on_variant(supported=[VARIANT_ESP32S3, VARIANT_ESP32S2])`

esphome/components/usb_host/__init__.py
```diff
@@ -0,0 +1,55 @@
+import esphome.codegen as cg
+from esphome.components.esp32 import add_idf_sdkconfig_option
+import esphome.config_validation as cv
+from esphome.const import CONF_ID
+from esphome.cpp_types import Component
+
+AUTO_LOAD = ["bytebuffer"]
+CODEOWNERS = ["@clydebarrow"]
```
jesserockz, line 13:
```suggestion
CODEOWNERS = ["@clydebarrow"]
DEPENDENCIES = ["esp32"]
```

esphome/components/usb_host/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+        }
+    )
+    if vid:
+        schema = schema.extend({cv.Optional(CONF_VID, default=vid): cv.uint16_t})
+    else:
+        schema = schema.extend({cv.Required(CONF_VID): cv.uint16_t})
+    if pid:
+        schema = schema.extend({cv.Optional(CONF_PID, default=pid): cv.uint16_t})
+    else:
+        schema = schema.extend({cv.Required(CONF_PID): cv.uint16_t})
```
jesserockz: Purely visual for codegen so it shows `0x1234` instead of `4660` inside `main.cpp`

```suggestion
        schema = schema.extend({cv.Optional(CONF_VID, default=vid): cv.hex_uint16_t})
    else:
        schema = schema.extend({cv.Required(CONF_VID): cv.hex_uint16_t})
    if pid:
        schema = schema.extend({cv.Optional(CONF_PID, default=pid): cv.hex_uint16_t})
    else:
        schema = schema.extend({cv.Required(CONF_PID): cv.hex_uint16_t})
```

esphome/components/usb_uart/usb_uart.h
```diff
@@ -0,0 +1,151 @@
...
+enum UARTStopBitsOptions {
+  UART_CONFIG_STOP_BITS_1 = 0,
+  UART_CONFIG_STOP_BITS_1_5,
+  UART_CONFIG_STOP_BITS_2,
+};
+
+static const char *const PARITY_NAMES[] = {"NONE", "ODD", "EVEN", "MARK", "SPACE"};
+static const char *const STOP_BITS_NAMES[] = {"1", "1.5", "2"};
+
+class RingBuffer {
```
jesserockz, line 48: Implementing another ringbuffer?

https://github.com/esphome/esphome/blob/e94e71ded80b5d6b5c943fce0f08d1f44a7ebe62/esphome/core/ring_buffer.h#L13

esphome/components/usb_host/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+async def register_usb_client(config):
+    var = cg.new_Pvariable(config[CONF_ID], config[CONF_VID], config[CONF_PID])
+    await cg.register_component(var, config)
+    return var
+
+
+async def to_code(config):
+    add_idf_sdkconfig_option("CONFIG_USB_HOST_CONTROL_TRANSFER_MAX_SIZE", 1024)
+    cg.add_global(usb_host_ns.using)
```
jesserockz: Is this necessary?
```suggestion
    cg.add_global(usb_host_ns.using)
```

esphome/components/usb_uart/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+                                UART_STOP_BITS_OPTIONS, upper=True
+                            ),
+                            cv.Optional(CONF_PARITY, default="NONE"): cv.enum(
+                                UART_PARITY_OPTIONS, upper=True
+                            ),
+                            cv.Optional(CONF_DATA_BITS, default=8): cv.int_range(
+                                min=5, max=8
+                            ),
+                            cv.Optional(CONF_DUMMY_RECEIVER, default=False): cv.boolean,
+                            cv.Optional(CONF_DEBUG, default=False): cv.boolean,
```
jesserockz, line 96: Couldn't this leverage the existing uart `debug` schema and classes?

esphome/components/usb_uart/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+                        }
+                    )
+                ),
+                max_length(channels),
```
jesserockz:
```suggestion
                cv.Length(max=channels),
```

jesserockz (changes requested):

esphome/components/usb_uart/__init__.py
clydebarrow: The requirements for uart are a bit different, since it receives a byte at a time it has stuff to determine when to produce debug output. This isn't necessary with the usb uart since it comes in as packets anyway so the additional complication seemed redundant.

esphome/components/usb_uart/usb_uart.h
clydebarrow: This is a much lighter weight implementation, since it has no need to be thread- or interrupt-safe, and has the `peek()` function needed for uart.

jesserockz (approved):

RoganDawes:
fwiw, it appears that the esp32 OTG peripheral does support hubs, according to the last post in this thread: https://esp32.com/viewtopic.php?t=34167&start=20

RoganDawes:
https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/kconfig.html#config-usb-host-hubs-supported

```
diff --git a/esphome/components/usb_host/__init__.py b/esphome/components/usb_host/__init__.py
index b6ca77970..772783c13 100644
--- a/esphome/components/usb_host/__init__.py
+++ b/esphome/components/usb_host/__init__.py
@@ -58,6 +58,7 @@ async def register_usb_client(config):

 async def to_code(config):
     add_idf_sdkconfig_option("CONFIG_USB_HOST_CONTROL_TRANSFER_MAX_SIZE", 1024)
+    add_idf_sdkconfig_option("CONFIG_USB_HOST_HUBS_SUPPORTED", True)
     for device in config.get(CONF_DEVICES) or ():
```

worked to enable identification of the device via a hub. I suppose this might want to be controlled by a config option, though?

RoganDawes:
Would also be great to support the same debugging configuration as per https://esphome.io/components/uart.html#debugging

I tried, and `debug` is only expecting a true/false value, not a dict.

clydebarrow:
> fwiw, it appears that the esp32 OTG peripheral does support hubs, according to the last post in this thread: https://esp32.com/viewtopic.php?t=34167&start=20

Must have been added in a more recent IDF, it wasn't available when I wrote the driver.

RoganDawes:
> > fwiw, it appears that the esp32 OTG peripheral does support hubs, according to the last post in this thread: https://esp32.com/viewtopic.php?t=34167&start=20
> 
> Must have been added in a more recent IDF, it wasn't available when I wrote the driver.

Entirely possible!

Do you want to add it with a config option, or just add it regardless? Or should I do a PR for whichever you prefer?

clydebarrow:
Oh, please raise a PR - always welcome and appreciated. Config option is probably best as there is bound to be some overhead incurred by adding that support. And don't forget a docs PR to add the option.

RoganDawes:
PR in #9154

PR #8344, opened 2025-02-28
[pmsx003] Refactor Imports, Extract Constants, Improve Data Handling & Logging

ximex:

- sort imports
- extract magic numbers as const
- improve data send and data parse function
- improve checks with sensors are valid for which model
- log firmware_version and error_code (couldn't test that as i don't have the correct model)
- fixes this log warnings:
```
src/esphome/components/pmsx003/pmsx003.cpp: In member function 'void esphome::pmsx003::PMSX003Component::parse_data_()':
src/esphome/components/pmsx003/pmsx003.cpp:208:5: warning: this statement may fall through [-Wimplicit-fallthrough=]
  208 |     }
      |     ^
src/esphome/components/pmsx003/pmsx003.cpp:209:5: note: here
  209 |     case PMSX003_TYPE_5003S: {
      |     ^~~~
src/esphome/components/pmsx003/pmsx003.cpp:217:5: warning: this statement may fall through [-Wimplicit-fallthrough=]
  217 |     }
      |     ^
src/esphome/components/pmsx003/pmsx003.cpp:218:5: note: here
  218 |     case PMSX003_TYPE_X003: {
      |     ^~~~
```

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


ximex:
Some documentation that i found:
- PMS5003: https://www.aqmd.gov/docs/default-source/aq-spec/resources-page/plantower-pms5003-manual_v2-3.pdf
- PMS5003ST: https://raw.githubusercontent.com/Arduinolibrary/DFRobot_SEN0233_Air_Quality_Monitor/master/PMS5003ST%20series%20data%20manua_English_V2.6%20.pdf

DjordjeMandic:
Can you make PR title a bit more detailed? Something like: `PMSX003: Refactor Imports, Extract Constants, Improve Data Handling & Logging`

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -39,13 +39,13 @@ void PMSX003Component::set_pm_particles_100um_sensor(sensor::Sensor *pm_particle
   pm_particles_100um_sensor_ = pm_particles_100um_sensor;
 }
 
+void PMSX003Component::set_formaldehyde_sensor(sensor::Sensor *formaldehyde_sensor) {
+  formaldehyde_sensor_ = formaldehyde_sensor;
```
DjordjeMandic:
```suggestion
  this->formaldehyde_sensor_ = formaldehyde_sensor;
```

esphome/components/pmsx003/pmsx003.h
```diff
@@ -7,13 +7,23 @@
...
+static const uint8_t START_CHARACTER_1 = 0x42;
+static const uint8_t START_CHARACTER_2 = 0x4D;
+
+static const uint16_t PMS_STABILISING_MS = 30000;  // time taken for the sensor to become stable after power on in ms
+
+static const uint16_t PMS_CMD_MEASUREMENT_MODE_PASSIVE =
+    0x0000;  // use `PMS_CMD_MANUAL_MEASUREMENT` to trigger a measurement
+static const uint16_t PMS_CMD_MEASUREMENT_MODE_ACTIVE = 0x0001;  // automatically perform measurements
+static const uint16_t PMS_CMD_SLEEP_MODE_SLEEP = 0x0000;         // go to sleep mode
+static const uint16_t PMS_CMD_SLEEP_MODE_WAKEUP = 0x0001;        // wake up from sleep mode
```
DjordjeMandic: This should be in source file

DjordjeMandic (changes requested):

esphome/components/pmsx003/pmsx003.h
ximex: when should constants be in header and when is source file?

esphome/components/pmsx003/pmsx003.cpp
ximex: the others should then also changed like that, correct?

esphome/components/pmsx003/pmsx003.h
DjordjeMandic: As I understand it, constants should be placed in the source file unless they are part of an enum, a class, or needed as function parameters in the header. A header file should only expose what is necessary for using the class or module—anything meant solely for internal implementation should remain in the source file. If a header is included elsewhere, it should provide access only to what is essential for interacting with the class or module, keeping implementation details hidden.

esphome/components/pmsx003/pmsx003.cpp
DjordjeMandic: Yes, the change should be made for consistency with the contributing guidelines. If you’d like, you can update the other occurrences as well—it’s a good practice to follow the guidelines throughout the component. https://esphome.io/guides/contributing.html

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -116,55 +129,44 @@ void PMSX003Component::loop() {
...
-        length_matches = payload_length == 36;
-        break;
-    }
-
-    if (!length_matches) {
+    if (!this->check_payload_length_(payload_length)) {
       ESP_LOGW(TAG, "PMSX003 length %u doesn't match. Are you using the correct PMSX003 type?", payload_length);
       return false;
     }
     return true;
```
DjordjeMandic: Could this if (index == *) be reduced to switch (index) ?

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -173,163 +175,143 @@ optional<bool> PMSX003Component::check_byte_() {
...
-    }
+    if (this->formaldehyde_sensor_ != nullptr)
+      this->formaldehyde_sensor_->publish_state(formaldehyde);
+  }
+
+  // Temperature and Humidity
+  if (this->type_ == PMSX003_TYPE_5003ST || this->type_ == PMSX003_TYPE_5003T) {
+    const uint8_t temperature_offset = (this->type_ == PMSX003_TYPE_5003T) ? 24 : 30;
+
+    const float temperature = (int16_t) this->get_16_bit_uint_(temperature_offset) / 10.0f;
```
DjordjeMandic:
```suggestion
    const float temperature = static_cast<int16_t>(this->get_16_bit_uint_(temperature_offset)) / 10.0f;
```
Preferred over C-style cast because it is more explicit and safer, also easier to find with Ctrl+F

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -173,163 +175,143 @@ optional<bool> PMSX003Component::check_byte_() {
...
-  this->data_[data_index_++] = cmd;
-  this->data_[data_index_++] = (data >> 8) & 0xFF;
-  this->data_[data_index_++] = (data >> 0) & 0xFF;
-  int sum = 0;
-  for (int i = 0; i < data_index_; i++) {
-    sum += this->data_[i];
+bool PMSX003Component::check_payload_length_(uint16_t payload_length) {
+  switch (this->type_) {
+    case PMSX003_TYPE_X003:
+      return payload_length == 28 || payload_length == 20;  // 2*13+2 // TODO: payload_length=20?
```
DjordjeMandic: Could you clarify the TODO comment? Did you forgot to remove it?

esphome/components/pmsx003/pmsx003.cpp
ximex: i think easily only for 0-2. as then other code gets executed before the next check. but then we have some check with the switch and some with if. i would prefer to keep the if's as then it is consistent. or do you describe more in detail what you want

esphome/components/pmsx003/pmsx003.cpp
ximex: the check for `payload_length == 20` was already there (https://github.com/esphome/esphome/pull/8344/files#diff-999cd0d338a825182fde2018b73bd2e625969160484c41e9bcdb012147a7a88fL138) but i couldn't find any documentation for that case. so i'm not sure if this was wrong in the past. Any suggestions for a better comment? like: "No documentation found for `payload_length == 20`. Keeping it to don't break anything"

DjordjeMandic:
```suggestion
      // The expected payload length is typically 28 bytes.
      // However, a 20-byte payload check was already present in the code.
      // No official documentation was found confirming this.
      // Retaining this check to avoid breaking existing behavior.
      return payload_length == 28 || payload_length == 20;
```

Maybe this is enough?

esphome/components/pmsx003/pmsx003.cpp
ximex: @DjordjeMandic anything that i should do here now? if yes pls a bit more details

DjordjeMandic: Maybe like this? You will also need to move `const uint16_t payload_length = this->get_16_bit_uint_(2);` before switch statement. For example:
```cpp
optional<bool> PMSX003Component::check_byte_() {
  const uint8_t index = this->data_index_;
  const uint8_t byte = this->data_[index];
  const uint16_t payload_length = this->get_16_bit_uint_(2);

  switch (index) {
    case 0:
      return byte == START_CHARACTER_1;
    case 1:
      return byte == START_CHARACTER_2;
    case 2:
      return true;
    case 3:
      if (!this->check_payload_length_(payload_length)) {
        ESP_LOGW(TAG, "PMSX003 length %u doesn't match. Are you using the correct PMSX003 type?", payload_length);
        return false;
      }
      return true;
  }
```

ximex: and the rest of the function code into the "default" of the switch or after the switch?
i personally like the if (only) variant more

DjordjeMandic: > and the rest of the function code into the "default" of the switch or after the switch?
> i personally like the if (only) variant more

Rest after the switch but if you like it how it is now then it's fine.

DjordjeMandic (approved):

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -6,45 +6,58 @@ namespace pmsx003 {
...
 void PMSX003Component::set_temperature_sensor(sensor::Sensor *temperature_sensor) {
-  temperature_sensor_ = temperature_sensor;
+  this->temperature_sensor_ = temperature_sensor;
 }
-void PMSX003Component::set_humidity_sensor(sensor::Sensor *humidity_sensor) { humidity_sensor_ = humidity_sensor; }
-void PMSX003Component::set_formaldehyde_sensor(sensor::Sensor *formaldehyde_sensor) {
-  formaldehyde_sensor_ = formaldehyde_sensor;
+void PMSX003Component::set_humidity_sensor(sensor::Sensor *humidity_sensor) {
+  this->humidity_sensor_ = humidity_sensor;
 }
```
ximex: should this be moved into the header file as most other compoments does it?
is there a rule when a function implementation should be done in the header file? normally only the definition is in the header file and the impl in the c++ file. why is it here different?

ximex (commented):
@DjordjeMandic @swoboda1337 pls check my comment in cpp file

esphome/components/pmsx003/pmsx003.cpp
DjordjeMandic: Normally, function implementations belong in the .cpp file, while headers contain only declarations. However, short setter/getter functions are often defined in the header, especially if they are simple one-liners. This avoids unnecessary function calls and aligns with common practices. Since your setters are only used during code generation, keeping them in the header makes sense. If they grow in complexity(but should not, complex part should be done in python), moving them to a .cpp file would be better.

I'd move those to header.

ximex:
@DjordjeMandic @swoboda1337 now i hope it's ready for merge. pls review

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -116,242 +110,205 @@ void PMSX003Component::loop() {
...
-  if (index == 1)
-    return byte == 0x4D;
+  if (index == 1) {
+    if (byte == START_CHARACTER_2) {
+      return true;
+    } else {
+      ESP_LOGW(TAG, "Start character 2 mismatch: 0x%02X != 0x%02X", byte, START_CHARACTER_2);
+      return false;
+    }
+  }
```
DjordjeMandic: Logging line can be called only once and returned early.

For example

```cpp
uint8_t start_char = index == 0 ? START_CHARACTER_1 : START_CHARACTER_2;
if (index <= 1) {
  if (byte == start_char) 
    return true;

  ESP_LOGW(TAG, "Start character %u mismatch: 0x%02X != 0x%02X", index + 1, byte, start_char);  
  return false;
}
```

That should work.

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -116,242 +110,205 @@ void PMSX003Component::loop() {
...
 
 void PMSX003Component::parse_data_() {
-  switch (this->type_) {
-    case PMSX003_TYPE_5003ST: {
-      float temperature = (int16_t) this->get_16_bit_uint_(30) / 10.0f;
-      float humidity = this->get_16_bit_uint_(32) / 10.0f;
+  // Particle Matter
+  const uint16_t pm_1_0_std_concentration = this->get_16_bit_uint_(4);
+  const uint16_t pm_2_5_std_concentration = this->get_16_bit_uint_(6);
+  const uint16_t pm_10_0_std_concentration = this->get_16_bit_uint_(8);
```
DjordjeMandic, line 209: As i can see, these are not used if sensor pointers are null. Could then these be moved inside nullptr check to only call get 16 bit uint only if sensor ptr is not null? For example execute this code right before publishing state only if needed.

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -116,242 +110,205 @@ void PMSX003Component::loop() {
...
+
+  const uint16_t pm_1_0_concentration = this->get_16_bit_uint_(10);
+  const uint16_t pm_2_5_concentration = this->get_16_bit_uint_(12);
+  const uint16_t pm_10_0_concentration = this->get_16_bit_uint_(14);
+
+  const uint16_t pm_particles_03um = this->get_16_bit_uint_(16);
+  const uint16_t pm_particles_05um = this->get_16_bit_uint_(18);
+  const uint16_t pm_particles_10um = this->get_16_bit_uint_(20);
+  const uint16_t pm_particles_25um = this->get_16_bit_uint_(22);
```
DjordjeMandic, line 218: Same here, not used if sensor ptr is null.

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -116,242 +110,205 @@ void PMSX003Component::loop() {
...
-      case PMSX003_TYPE_5003ST:
-        length_matches = payload_length == 36;
-        break;
-    }
-
-    if (!length_matches) {
+    if (this->check_payload_length_(payload_length)) {
+      return true;
+    } else {
       ESP_LOGW(TAG, "PMSX003 length %u doesn't match. Are you using the correct PMSX003 type?", payload_length);
```
DjordjeMandic:
```suggestion
      ESP_LOGW(TAG, "Payload length %u doesn't match. Are you using the correct PMSX003 type?", payload_length);
```

esphome/components/pmsx003/pmsx003.cpp
```diff
@@ -116,242 +110,205 @@ void PMSX003Component::loop() {
...
+    // Note the pm particles 50um & 100um are not returned,
+    // as PMS5003T uses those data values for temperature and humidity.
+    const uint16_t pm_particles_50um = this->get_16_bit_uint_(24);
+    const uint16_t pm_particles_100um = this->get_16_bit_uint_(26);
+
+    if (this->pm_particles_50um_sensor_ != nullptr)
+      this->pm_particles_50um_sensor_->publish_state(pm_particles_50um);
+    if (this->pm_particles_100um_sensor_ != nullptr)
+      this->pm_particles_100um_sensor_->publish_state(pm_particles_100um);
+  }
```
DjordjeMandic:
```suggestion
  const uint16_t pm_1_0_concentration = this->get_16_bit_uint_(10);
  const uint16_t pm_2_5_concentration = this->get_16_bit_uint_(12);
  const uint16_t pm_10_0_concentration = this->get_16_bit_uint_(14);

  if (this->pm_1_0_std_sensor_ != nullptr)
    this->pm_1_0_std_sensor_->publish_state(this->get_16_bit_uint_(4));
  if (this->pm_2_5_std_sensor_ != nullptr)
    this->pm_2_5_std_sensor_->publish_state(this->get_16_bit_uint_(6));
  if (this->pm_10_0_std_sensor_ != nullptr)
    this->pm_10_0_std_sensor_->publish_state(this->get_16_bit_uint_(8));

  if (this->pm_1_0_sensor_ != nullptr)
    this->pm_1_0_sensor_->publish_state(pm_1_0_concentration);
  if (this->pm_2_5_sensor_ != nullptr)
    this->pm_2_5_sensor_->publish_state(pm_2_5_concentration);
  if (this->pm_10_0_sensor_ != nullptr)
    this->pm_10_0_sensor_->publish_state(pm_10_0_concentration);

  if (this->pm_particles_03um_sensor_ != nullptr)
    this->pm_particles_03um_sensor_->publish_state(this->get_16_bit_uint_(16));
  if (this->pm_particles_05um_sensor_ != nullptr)
    this->pm_particles_05um_sensor_->publish_state(this->get_16_bit_uint_(18));
  if (this->pm_particles_10um_sensor_ != nullptr)
    this->pm_particles_10um_sensor_->publish_state(this->get_16_bit_uint_(20));
  if (this->pm_particles_25um_sensor_ != nullptr)
    this->pm_particles_25um_sensor_->publish_state(this->get_16_bit_uint_(22));

  if (this->type_ != PMSX003_TYPE_5003T) {
    // Note the pm particles 50um & 100um are not returned,
    // as PMS5003T uses those data values for temperature and humidity.
    if (this->pm_particles_50um_sensor_ != nullptr)
      this->pm_particles_50um_sensor_->publish_state(this->get_16_bit_uint_(24));
    if (this->pm_particles_100um_sensor_ != nullptr)
      this->pm_particles_100um_sensor_->publish_state(this->get_16_bit_uint_(26));
  }
```

esphome/components/pmsx003/pmsx003.cpp
ximex: i like the current codestyle more but i will change it

esphome/components/pmsx003/pmsx003.cpp
ximex: also logged it now

esphome/components/pmsx003/pmsx003.cpp
ximex: also logged it now

DjordjeMandic (approved):
Looks good to me

ximex:
@jesserockz @swoboda1337 pls review/merge

kbx81 (approved):
Thanks! 🍻

PR #8305, opened 2025-02-24
[ld2450] Fix for "unknown" sensor states

kbx81:
# What does this implement/fix?

Fix issue where sensor states are sometimes "unknown". See https://github.com/esphome/esphome/pull/5674#issuecomment-2669281465

Test with:

```yaml
external_components:
  - source: github://pr#8305
    components: [ ld2450 ]
    refresh: 0s
```

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- fixes <link to issue>

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#<esphome-docs PR number goes here>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml

```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [ ] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8348, opened 2025-03-01
[ld2450] fix null exception & zone target_count not published

mistic100:

This adds missing null-checks in the LD2450 when the default number components are not used.

The error occurs when one use for example template numbers, but none of the default ld2450 numbers.

I also checked other optional variable, everything else seems fine.

## Types of changes

- [ ] Other

https://github.com/esphome/esphome/pull/8305

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Minimal excerpt

```yaml
ld2450:
  id: ld2450_radar
  uart_id: uart_ld2450

sensor:
  - platform: ld2450
    ld2450_id: ld2450_radar
    target_count:
      name: "Presence Target Count"

number:
  - platform: template
    name: "My number"
    min_value: 0
    max_value: 100
    step: 1
    initial_value: 0
    optimistic: true
    restore_value: true
```

## Checklist:

### DIFF
--- a/esphome/components/ld2450/ld2450.cpp
+++ b/esphome/components/ld2450/ld2450.cpp
@@ -110,8 +110,10 @@ LD2450Component::LD2450Component() {}
 void LD2450Component::setup() {
   ESP_LOGCONFIG(TAG, "Setting up HLK-LD2450...");
 #ifdef USE_NUMBER
-  this->pref_ = global_preferences->make_preference<float>(this->presence_timeout_number_->get_object_id_hash());
-  this->set_presence_timeout();
+  if (this->presence_timeout_number_ != nullptr) {
+    this->pref_ = global_preferences->make_preference<float>(this->presence_timeout_number_->get_object_id_hash());
+    this->set_presence_timeout();
+  }
 #endif
   this->restart_and_read_all_info();
 }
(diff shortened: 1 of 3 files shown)


mistic100:
@kbx81 :-)

mistic100:
In the original PR I mentioned the device became unresponsive as soon as I tried to set the builtin zone coordinates. I think this commit fix it : I had only declared the Zone 1 number components, thus having null pointer errors when reaching zones 2 and 3 publish state.

mistic100:
I just added a small change to fix the following bug : for a specific zone, the `target_count` is wrong if `still_target_count` or `moving_target_count` is not declared

esphome/components/ld2450/ld2450.cpp
```diff
@@ -281,10 +277,13 @@ void LD2450Component::process_zone_(uint8_t *buffer) {
     this->zone_config_[index].x2 = ld2450::hex_to_signed_int(buffer, start + 4);
     this->zone_config_[index].y2 = ld2450::hex_to_signed_int(buffer, start + 6);
 #ifdef USE_NUMBER
-    this->zone_x1_numbers_[index]->publish_state(this->zone_config_[index].x1);
-    this->zone_y1_numbers_[index]->publish_state(this->zone_config_[index].y1);
-    this->zone_x2_numbers_[index]->publish_state(this->zone_config_[index].x2);
-    this->zone_y2_numbers_[index]->publish_state(this->zone_config_[index].y2);
+    // only one null check as all coordinates are required for a single zone
+    if (this->zone_numbers_[index].x1 != nullptr) {
```
kbx81, line 281: This was making me a bit nervous since we had individual setters for each of these pointers...so I did [`b58c3b2` (#8348)](https://github.com/esphome/esphome/pull/8348/commits/b58c3b2fb8e4970dac26eefa924267838b1b0a18) which essentially forces setting them all as a single unit. It's still not perfect but I think it's generally handled a bit better since you'd never have a case where all four wouldn't be set.

This change also reduces the size of the binary by over 300 bytes:

Before:
```
Linking .pioenvs/apollo-mtr-1/firmware.elf
RAM:   [=         ]  14.2% (used 46380 bytes from 327680 bytes)
Flash: [========= ]  91.8% (used 1684154 bytes from 1835008 bytes)
```
After:
```
Linking .pioenvs/apollo-mtr-1/firmware.elf
RAM:   [=         ]  14.2% (used 46380 bytes from 327680 bytes)
Flash: [========= ]  91.8% (used 1683818 bytes from 1835008 bytes)
```

kbx81: @mistic100 please confirm if this change works as expected for your configuration(s) -- if there are no new issues, I'll get this merged. 😄

mistic100: everything's fine

kbx81 (approved):
Thanks! 🍻

PR #8354, opened 2025-03-02
[online_image] Allocate pngle manually to potentially use psram

jesserockz:

Instead of calling `pngle_new` which only uses `calloc`, use ESPHomes `RAMAllocator` so we can try store the pngle object in psram.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/components/online_image/png_image.cpp
```diff
@@ -40,6 +40,24 @@ static void draw_callback(pngle_t *pngle, uint32_t x, uint32_t y, uint32_t w, ui
...
 }
 
+PngDecoder::PngDecoder(OnlineImage *image) : ImageDecoder(image) {
+  {
+    pngle_t *pngle = this->allocator_.allocate(1);
+    if (!pngle) {
+      ESP_LOGE(TAG, "Failed to allocate memory for PNGLE engine!");
+      return;
+    }
+    this->pngle_ = pngle;
```
guillempages, line 52: The original code is calling pngle_reset after the calloc, to initialize the fields...
Shouldn't the constructor do that as well?

jesserockz: Absolutely!

nliviu:
Using this PR, I sometimes get `assert failed: heap_caps_free heap_caps_base.c:63 (heap != NULL && "free() target pointer is outside heap areas")`
The stack trace
```
0x4037784a: panic_abort at /data/cache/platformio/packages/framework-espidf/components/esp_system/panic.c:466
0x4037dff5: esp_system_abort at /data/cache/platformio/packages/framework-espidf/components/esp_system/port/esp_system_chip.c:84
0x40384d0d: __assert_func at /data/cache/platformio/packages/framework-espidf/components/newlib/assert.c:81
0x40378439: heap_caps_free at /data/cache/platformio/packages/framework-espidf/components/heap/heap_caps_base.c:63
0x40384d4d: free at /data/cache/platformio/packages/framework-espidf/components/newlib/heap.c:39
0x42029adb: pngle_reset at /piolibs/pngle/src/pngle.c:162
0x42014a53: esphome::online_image::PngDecoder::PngDecoder(esphome::online_image::OnlineImage*) at /data/build/weather-test-s3/src/esphome/components/online_image/png_image.cpp:50
0x420147e7: std::__detail::_MakeUniq::__single_object std::make_unique(esphome::online_image::OnlineImage*&&) at /data/cache/platformio/packages/toolchain-xtensa-esp32s3/xtensa-esp32s3-elf/include/c++/12.2.0/bits/unique_ptr.h:1065
0x420147e7: esphome::online_image::OnlineImage::update() at /data/build/weather-test-s3/src/esphome/components/online_image/online_image.cpp:172
```

esphome/components/online_image/png_image.cpp
```diff
@@ -40,6 +40,25 @@ static void draw_callback(pngle_t *pngle, uint32_t x, uint32_t y, uint32_t w, ui
   decoder->draw(x, y, w, h, color);
 }
 
+PngDecoder::PngDecoder(OnlineImage *image) : ImageDecoder(image) {
+  {
+    pngle_t *pngle = this->allocator_.allocate(1, 43888U);
```
guillempages: Why the explicit manual size?
Isn't it safer to let the automatic `sizeof(T)` kick in? (which is what the original `pngle_new` call does anyway)

guillempages (commented):
Hm. I've just tested this, and it resets just after logging "Allocating PNG decoder"; if I comment out the `external_components` part that includes this PR (and nothing else), it works 🤔

esphome/components/online_image/png_image.cpp
guillempages: Maybe the call to `pngle_reset()` after allocating wasn't the best idea... It looks like this method is trying to free some memory that has not been allocated yet.
The original code in the libary uses `calloc`, which initializes the struct to 0, but the RAMAllocator probably just keeps the contents of the memory; which makes the whole set of `if (pngle->xyz) free(pngle->xyz);` calls from the reset method fail :-(

guillempages:
> Hm. I've just tested this, and it resets just after logging "Allocating PNG decoder"; if I comment out the `external_components` part that includes this PR (and nothing else), it works 🤔

Maybe the call to pngle_reset() after allocating wasn't the best idea... It looks like this method is trying to free some memory that has not been allocated yet.
The original code in the libary uses calloc, which initializes the struct to 0, but the RAMAllocator probably just keeps the contents of the memory; which makes the whole set of if (pngle->xyz) free(pngle->xyz); calls from the reset method fail :-(

nliviu:
I added
```
memset(pngle, 0, 43888U);
```
before `pngle_reset(pngle);` in the constructor and did not get any panic after ~2 hours.
Before that, the panic kicked in after 15-30 minutes.
`sizeof(T)` cannot be used because the structure defintion is not public. It is defined in [pngle.c](https://github.com/kikuchan/pngle/blob/8bf3862f84a4643e75a7d92e046d7c07f934cead/src/pngle.c#L72).

guillempages:
Sounds good. 

Maybe we should then define a constant for this size, since it is being used on allocation, setting to 0, and in the destructor...

Or maybe even better,  create a pr in pngle to make the struct visible; last time I made a pr there was reviewed pretty fast.

jesserockz:
https://github.com/kikuchan/pngle/issues/13

jesserockz:
Made a PR to pngle, hopefully it gets added https://github.com/kikuchan/pngle/pull/14

jesserockz:
I have set this PR to draft pending a release of pngle. I have optimistically bumped the versions to 1.0.3 :crossed_fingers:

DavidAntonin:
@jesserockz pngle 1.1.0 is released, can you please updater PR. Thanks

esphome/components/online_image/__init__.py
```diff
@@ -75,7 +75,7 @@ def __init__(self):
 
     def actions(self):
         cg.add_define("USE_ONLINE_IMAGE_PNG_SUPPORT")
-        cg.add_library("pngle", "1.0.2")
+        cg.add_library("pngle", "1.0.3")
```
DavidAntonin, line 79: cg.add_library("pngle", "1.1.0")

DavidAntonin (changes requested):
change version to the last release
cg.add_library("pngle", "1.1.0")

PR #8370, opened 2025-03-07
Add output component for MCP4151 a 8-bit digital potentiometer with SPI

codermarti:

This Pull Request introduces a new output component for the MCP4151 digital potentiometer, allowing users to control the wiper in a straightforward manner within ESPHome. It leverages the standard ESPHome SPI infrastructure, enabling a fully integrated setup. By default, the output expects values in the [0..1] range (like a dimmer), yet can be easily mapped from real-world resistance values (e.g., kΩ) in user YAML. The component thus offers a flexible way to simulate or control variable resistances in Home Assistant.

### Key Features:

- **Supports Multiple MCP4151 Variants**: The MCP4151 family is available with various maximum resistances (e.g. 10 kΩ, 50 kΩ, 100 kΩ). The user can account for these differences by scaling or offsetting the ratio/steps accordingly in YAML or code.
- **Simple 0..1 Mapping**: Aligns with ESPHome’s standard output mechanism (like a dimmer). You can optionally add a small lambda if you wish to interpret the value as ohms/kΩ.
- **Logging & Debug**: Provides logging macros for setup and runtime to assist with troubleshooting.
- **Extended Real-World Applications**: By adding an exponential fit or custom math in YAML, you can interpret external data (e.g. temperature) and convert it into a dynamic resistance, then apply it to the MCP4151 output.

### Datasheets for MCP4151:
[https://www.microchip.com/en-us/product/mcp4151](https://www.microchip.com/en-us/product/mcp4151)

## Types of changes

- [ ] Other

none

- esphome/esphome-docs#4738

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
spi:
  clk_pin: D5  # GPIO14 (SCK)
  mosi_pin: D7 # GPIO13 (SDI/MOSI)
  interface: hardware

output:
  - platform: mcp4151
    id: digital_pot
    cs_pin: D8  # GPIO15 as Chip Select (CS)

# number-input for the resistance as slider in home assistant
number:
  - platform: template
    name: "MCP4151 Potentiometer"
    id: mcp4151_number
    min_value: 0
    max_value: 100
    step: 1
    unit_of_measurement: "kΩ"
    icon: "mdi:resistor"
    set_action:
      then:
        - output.set_level:
            id: digital_pot
            level: !lambda "return x/100;"

```

## Checklist:


esphome/components/mcp4151/mcp4151.cpp
```diff
@@ -0,0 +1,35 @@
...
+  // scale 0..1 => 0..255
+  uint8_t value = static_cast<uint8_t>(state * 255.0f);
+  ESP_LOGD(TAG, "  set value: %d", static_cast<int>(value));
+  this->last_value_ = value;
+  this->send_command_(value);
+}
+
+void MCP4151::send_command_(uint8_t value) {
+  this->enable();
+  this->transfer_byte(0x00);  // command Byte (Pot #0)
```
clydebarrow: Should be `write_byte` as you are not using the read value.

codermarti: thanks for your input, i've changed it

p1ngb4ck:
Hey, my own component for the MCP4461 is quite similiar, sorry to bother or interrupt.
The Wiper of the MCP4151 supports 257 steps (not 256 as uint8_t) and therefore requires a 2 bytes of combined cmd, address & value with value having 10 bits.
As far as I see, this writes only cmd + 1 uint8_t byte ? you could therefore set it to max 256 of 257 steps..

PS: Also - I checked and realized that quite all of the MCPxxxx potentiometers share mostly same cmd-architecture -> it should actually be possible to implement support for all of them with common codebase, do you agree?

codermarti:
> Hey, my own component for the MCP4461 is quite similiar, sorry to bother or interrupt. The Wiper of the MCP4151 supports 257 steps (not 256 as uint8_t) and therefore requires a 2 bytes of combined cmd, address & value with value having 10 bits. As far as I see, this writes only cmd + 1 uint8_t byte ? you could therefore set it to max 256 of 257 steps..
> 

Hey, 
thanks for your input. The main difference here is that the MCP413X/415X/423X/425X series communicates through SPI instead of I2C as your MCP4461. 

As stated in the datasheet it supports two Modes:
- 7-bit: 128 Resistors (129 Steps)
- 8-bit: 256 Resistors (257 Steps)

i think the 1 uint8_t byte should be enough for the 256 Resistors. 

> PS: Also - I checked and realized that quite all of the MCPxxxx potentiometers share mostly same cmd-architecture -> it should actually be possible to implement support for all of them with common codebase, do you agree?

for the SPI based MCP413X/415X/423X/425X series you're right. They all should work with this component. I'm not sure if mixing of I2C and SPI based components in one codebase is a good idea.

p1ngb4ck:
Yes I know they are i2c & spi.
But - cmds for read/write/increase/decrease are similar, byte generation for cmd/address&value is exactly similar , as well as the way status & terminal registers work.
Only differences of the chips are:
bustype - obviously
address-type - 7 or 8bit address types
number of wipers
eeprom for nonvolatile wipers available yes/no
write-protect&wiperlock support also depend on exact type, where eeproms variants generally also support those

> i think the 1 uint8_t byte should be enough for the 256 Resistors.
> 
Not really, you would not ever be able to set the wiper to the max position actually that way, which could be actually important to some users. It is actually 0 resistors + 256 resistors = 257 different states = 257 steps. uint8_t only has 0-255.
As the MCP4151 has only one wiper and the address for the wiper is 0 also, you can do it just the way you do, except you would need to work with a uint16_t value passed, check min/max to be 0/256 and then cast the value to be uint8_t after doing a bitshift of 1 to the last bit of the first byte in case the value is 256.
The even "better" alternative, to also support dual-channel versions of the chip right away, would be to use output-channels and bitshift the adress to the first byte also.
And of course you would need to modify the multiplication of the output to be 256.0f then and define the max to 256 on the number in yaml. 
If you "wanted" you could also support the dual wiper versions & the eeprom versions just the way I do btw. there is exactly 0 difference except the different Bus-initialization, so you could copy all that right away from my extension to test / tinker, if you like.
Indeed it would be lovely to know, what exact value your chip responds for the status & terminal-registers by default - I`d love to know if those could be used to determine if the chip actually "has" that wiper (if its single or dual) automatically in setup .. I only own the quad channel i2c variant and cannot test this therefore, as mine do "have" all wipers supported..

PR #8376, opened 2025-03-09
[touchscreen] Axis swap bugfix

clydebarrow:

When using `transform` to swap the axes of a touchscreen, the swapping is done *after* the scaling via the calibration values. Since calibration values are set automatically from display dimensions this results in the range of values being calculated from the wrong axis unless calibration has been set manually. Instead the swapping is done at the raw level.

This is _potentially_ a breaking change since to work-around this problem some users may have set calibration values that are now wrong.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


kbx81 (approved):
Thanks!

PR #8385, opened 2025-03-10
[const] Create component-level const repository

clydebarrow:

There is a lint rule that requires any constant used in 3 or more places to be defined in a single place. Currently that place is the core file `esphome/const.py`. The problem with this is that when a constant used by a component gets moved from a component-level file to the core file, it won't be pulled in with an external component, which makes testing PRs that involve refactored constants hard to test without cloning the PR.

Following from a discussion with @jesserockz this PR creates a new component-level place to put common constants, starting with a single constant. Obviously this should only be used for constants that are not referenced in core code.

Now a PR affected by such changes can be tested by including `const` as a component in the `external_components` config.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


nielsnl68:
Good idea, i like it,

Just wondering could we not move all const into this setup?
And direct the core/const to this same-component?
Or will this break to much?

clydebarrow:
It would either require a massive refactor (not going to happen) or adding a dependency from core to a component (again, not going to happen.)

So use for new components, or refactor individual components as and when worked on.

What *might* be feasible is to automatically make all consts currently in core available in the component. Can't just do a wildcard import because the linter barfs on that.

jesserockz (approved):

clydebarrow:

```yaml
external_components:
  - source: github://pr#8385
    components: [const, lvgl, qspi_dbi]
    refresh: 1h
```

---
(Added by the PR bot)

PR #8393, opened 2025-03-12
Tuya Select - Add int_datapoint option

Cossid:
Allow configuration of backing data type for Tuya Select, expanding from ``enum`` to also allowing ``int``

While Tuya has an enum type, it sometimes uses int in the same way for multi-option states.  This allows sending the value as int instead of enum, if that is the backing type of the MCU.  Defaults to previous behavior of ``enum``

## Types of changes

- [ ] Other

- fixes N/A

- esphome/esphome-docs#4749

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [X] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
select:
  - platform: "tuya"
    name: "Indicator Configuration"
    data_type: int
    entity_category: config
    enum_datapoint: 101
    optimistic: true
    options:
      0: "Inverted"
      1: "Normal"
```

## Checklist:


esphome/components/tuya/select/tuya_select.h
```diff
@@ -26,6 +27,7 @@ class TuyaSelect : public select::Select, public Component {
   bool optimistic_ = false;
   uint8_t select_id_;
   std::vector<uint8_t> mappings_;
+  uint8_t data_type_;
```
bearpawmaxim: Using `uint8_t` as `bool` may lead to cunfusion... Maybe to use boolean and name a flag like `enum_data_type_` ?

Cossid: While it is used as boolean in this context for this patch, this could easily be expanded to include boolean and binary as other options for example, two other formats Tuya supports, but how useful they would be is debatable.

I'm open to either way.

Cossid: In retrospect, it's even possible there should be a breaking change of changing `enum_datapoint` to just `datapoint` or `select_datapoint` or it becomes confusing

bearpawmaxim: > In retrospect, it's even possible there should be a breaking change

We may support both? And check whether any of the two specified. (for backward compatibility)

bearpawmaxim (approved):

esphome/components/tuya/select/tuya_select.h
Cossid: Tests aren't passing yet, looking into what the reason is, I think it's how I'm assinging bool in to_code.  Moving to draft until resolved.

esphome/components/tuya/select/__init__.py
```diff
@@ -35,13 +33,14 @@ def ensure_option_map(value):
...
+            cv.Optional(CONF_ENUM_DATAPOINT): cv.uint8_t,
+            cv.Optional(CONF_INT_DATAPOINT): cv.uint8_t,
             cv.Required(CONF_OPTIONS): ensure_option_map,
             cv.Optional(CONF_OPTIMISTIC, default=False): cv.boolean,
-            cv.Optional(CONF_DATA_TYPE, default="enum"): cv.enum(DATA_TYPE, lower=True),
         }
     )
-    .extend(cv.COMPONENT_SCHEMA)
+    .extend(cv.COMPONENT_SCHEMA),
+    cv.has_at_least_one_key(CONF_ENUM_DATAPOINT, CONF_INT_DATAPOINT),
```
bearpawmaxim: Maybe `cv.has_exactly_one_key` ? (to exclude situations when two options are specified)

Cossid: Yep, already changed to that, issue still exists.

esphome/components/tuya/select/__init__.py
```diff
@@ -31,12 +33,14 @@ def ensure_option_map(value):
...
-            cv.Required(CONF_ENUM_DATAPOINT): cv.uint8_t,
+            cv.Optional(CONF_ENUM_DATAPOINT): cv.uint8_t,
+            cv.Optional(CONF_INT_DATAPOINT): cv.uint8_t,
             cv.Required(CONF_OPTIONS): ensure_option_map,
             cv.Optional(CONF_OPTIMISTIC, default=False): cv.boolean,
         }
     )
-    .extend(cv.COMPONENT_SCHEMA)
+    .extend(cv.COMPONENT_SCHEMA),
+    cv.has_exactly_one_key(CONF_ENUM_DATAPOINT, CONF_INT_DATAPOINT),
```
bearpawmaxim, line 46: Need to wrap an entire schema with cv.All:
```
    CONFIG_SCHEMA = cv.All(
       ....,
       cv.has_exactly_one_key(...
    )
```

Then, the validation will work

bearpawmaxim (changes requested):

esphome/components/tuya/select/__init__.py
Cossid: Thanks, looks like that was exactly it, I'm not great on the config building part yet.

bearpawmaxim (approved):
Looks good. Thanks)

esphome/components/tuya/select/__init__.py
```diff
@@ -8,6 +8,8 @@
 DEPENDENCIES = ["tuya"]
 CODEOWNERS = ["@bearpawmaxim"]
 
+CONF_INT_DATAPOINT = "int_datapoint"
```
bearpawmaxim: Let's move it to esphome/const.py, after `CONF_ENUM_DATAPOINT`. (so it will be declared in the same place as the existing is)

Cossid: Done.

esphome/components/tuya/select/__init__.py
```diff
@@ -47,5 +54,10 @@ async def to_code(config):
     cg.add(var.set_select_mappings(list(options_map.keys())))
     parent = await cg.get_variable(config[CONF_TUYA_ID])
     cg.add(var.set_tuya_parent(parent))
-    cg.add(var.set_select_id(config[CONF_ENUM_DATAPOINT]))
+    if CONF_ENUM_DATAPOINT in config:
+        cg.add(var.set_select_id(config[CONF_ENUM_DATAPOINT]))
```
jesserockz:
```suggestion
    if (enum_datapoint := config.get(CONF_ENUM_DATAPOINT, None) is not None:
        cg.add(var.set_select_id(enum_datapoint))
```

esphome/components/tuya/select/__init__.py
```diff
@@ -47,5 +54,10 @@ async def to_code(config):
...
+        cg.add(var.set_is_int(False))
+    if CONF_INT_DATAPOINT in config:
+        cg.add(var.set_select_id(config[CONF_INT_DATAPOINT]))
```
jesserockz:
```suggestion
    if (int_datapoint := config.get(CONF_INT_DATAPOINT, None) is not None:
        cg.add(var.set_select_id(int_datapoint))
```

esphome/components/tuya/select/tuya_select.h
```diff
@@ -18,6 +18,7 @@ class TuyaSelect : public select::Select, public Component {
   void set_optimistic(bool optimistic) { this->optimistic_ = optimistic; }
   void set_select_id(uint8_t select_id) { this->select_id_ = select_id; }
   void set_select_mappings(std::vector<uint8_t> mappings) { this->mappings_ = std::move(mappings); }
+  void set_is_int(bool is_int) { this->is_int_ = is_int; }
```
jesserockz: Can probably just make this a second arg on `set_select_id`

jesserockz (changes requested):

jesserockz (approved):

esphome/components/tuya/select/__init__.py
Cossid: I thought there was an extra ( at the beginning, but it turns out it was actually a missing one after the config.get().
https://github.com/esphome/esphome/pull/8871 to fix.

jesserockz: My bad

PR #8394, opened 2025-03-12
Tuya Fan - Support presets.

Cossid:

Add preset modes to Tuya Fan, matching how other fan types support it.

Review notes: This required calling publish_state() on control() which Tuya components usually don't do (or do more directly through their uart communication).  But it is needed to hook into callbacks for things like `on_preset_set`.  I wasn't sure if there was a better way to do this, as I'm not sure calling the callback directly is exposed.  Let me know if this needs adjustment.  It appears to work fine on my fan switch.  I set all other states in control() just to be safe to make sure things don't get out of sync.  Uart should still be triggering everything the same as it always has.

## Types of changes

- [ ] Other

- fixes N/A

- esphome/esphome-docs#4751

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [X] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
tuya:

fan:
  - platform: "tuya"
    id: tuya_fan
    name: None
    icon: "mdi:ceiling-fan"
    switch_datapoint: 1
    speed_datapoint: 3
    speed_count: 4
    restore_mode: RESTORE_DEFAULT_OFF
    preset_modes:
      - "Turtle"
      - "Low"
      - "Medium"
      - "High"
```

## Checklist:


esphome/components/tuya/fan/tuya_fan.cpp
```diff
@@ -29,7 +27,6 @@ void TuyaFan::setup() {
     this->parent_->register_listener(*this->switch_id_, [this](const TuyaDatapoint &datapoint) {
       ESP_LOGV(TAG, "MCU reported switch is: %s", ONOFF(datapoint.value_bool));
       this->state = datapoint.value_bool;
-      this->publish_state();
```
jesserockz, line 32: If you remove these, any update from the mcu side wont be published until something else calls `this->publish_state()`

esphome/components/tuya/fan/tuya_fan.cpp
```diff
@@ -74,33 +75,35 @@ void TuyaFan::dump_config() {
...
 }
 
-fan::FanTraits TuyaFan::get_traits() {
-  return fan::FanTraits(this->oscillation_id_.has_value(), this->speed_id_.has_value(), this->direction_id_.has_value(),
-                        this->speed_count_);
-}
-
 void TuyaFan::control(const fan::FanCall &call) {
   if (this->switch_id_.has_value() && call.get_state().has_value()) {
+    this->state = *call.get_state();
```
jesserockz, line 85: Doing this turns the fan into having an "optimistic" state that may not match the actual device state. I dont think we should do this and restore the publishes from above.

esphome/components/tuya/fan/tuya_fan.cpp
```diff
@@ -74,33 +75,35 @@ void TuyaFan::dump_config() {
...
+    this->speed = *call.get_speed();
     if (this->speed_type_ == TuyaDatapointType::ENUM) {
       this->parent_->set_enum_datapoint_value(*this->speed_id_, *call.get_speed() - 1);
     } else if (this->speed_type_ == TuyaDatapointType::INTEGER) {
       this->parent_->set_integer_datapoint_value(*this->speed_id_, *call.get_speed());
     }
   }
+  this->preset_mode = call.get_preset_mode();
+  // publish_state() is required for callbacks for preset events.
+  this->publish_state();
```
jesserockz, line 111: This is fine as preset is an internal thing anyway.

jesserockz (changes requested):

esphome/components/tuya/fan/tuya_fan.cpp
Cossid: Yeah, removing those publishes was a misunderstanding, I read those as if statements, not callbacks, and I was trying to optimize by just calling `publish_state` once later, but being callbacks, they still need to be there.  I have added them back.
However, I'm unclear what this comment means now in the case of publish_states being restored.  All other speed fans call `get_state` here.  Was adding back `publish_state` enough to resolve this, or is there something else that needs consideration here?

Cossid: I'm moving this back to ready for review, please clarify if this requires further changes, and what they would be, and if it would also apply to the oscillation/direction/speed as well, and I'll adjust accordingly.

Cossid:
Not stale, patiently waiting for re-review after adjustments from previous feedback.

jesserockz:
Just coming back to this now again sorry. 

I'm not sure if I didn't see this earlier, but I'm not sure I see the purpose of adding presets here as components in ESPHome are supposed to represent the state of the external hardware that is attached, not be a virtual value in memory only (that is what `template` platforms are for).

Cossid:
It would be for conveying to HA what the available presets are aside from percentages.  I'm not sure how a template would accomplish that, it doesn't play well with the Tuya system.  In HA, there is a part of the card that shows available presets to be selected, which is why the other speed fans also have the preset option.  It also allows voice control to control by a name instead of a % value.

It it doesn't belong here, why does it exist on `speed_fan` and supports_presets is a part of the base fan component?

jesserockz:
`supports_presets` is part of the base fan component because it is a part of the entity type. 

Ok, I see your point about `speed_fan`.

Maybe presets for fans should be virtual if the hardware doesn't natively support it, but I think it needs to be done slightly differently. Right now the user has to list the presets, then check in a lambda what preset was set, then make another call to the fan to alter the speed etc. 

If we want to make this better, we should change the config to be something like:
```
yaml
fan:
  - platform: ...
    preset_modes:
      - name: Turtle
        speed: XX
        oscillation: true
        ...
      - name: ...
        ...
```

and handle setting everything internally correctly.

Thoughts?

Cossid:
You are correct that this would still require handling `on_preset_set` with a lambda and call instructions.

I like your proposal better, but I'm not sure I have the knowledge to implement it, especially across the other fan components.  It would also give a good opportunity to address https://github.com/esphome/issues/issues/6861 which I never moved after the issue tracker moved again.

PR #8416, opened 2025-03-16
Add Support for the 7.3inch E Ink Spectra 6 (E6)

### LINKED ISSUE #3054: Support for the 7.3inch E Ink Spectra 6 (E6)

Support for the 7.3inch E Ink Spectra 6 (E6)
https://www.waveshare.com/7.3inch-e-paper-hat-e.htm

There is already 7.30in-f - 7.3in 7-color display (black, white, red, yellow, blue, green, and orange)
So it should be very simple to adapt.  I think inverted: true is also needed. So its just a naming i think

**Additional context**

lublak:

Add Support for the 7.3inch E Ink Spectra 6 (E6).
I'll be honest, I had tried to set up the development environment. But I've failed so far and therefore can't test it at the moment. I'll add the pull request here anyway (if anyone wants to take over). Otherwise I'll have to see how I can manage it.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/feature-requests/issues/3054

- esphome/esphome-docs#5264

## Test Environment

TODO

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

TODO

```yaml
# Example config.yaml

```

## Checklist:


elsbrock:
While we wait to get this added, is there a way to try it locally with upstream esphome? Or do we need to use your branch?

lublak:
@elsbrock dry for the late response.
Currently you need to use my branch.
It's currently a merge conflict which I have not checked yet.

macgyver2089:
Hey @lublak do you happen to have sample/snippet yaml (maybe I’m missing where configuration examples would be in the structure) I’m testing with your dev branch, but since I haven’t looked up much about the Hat (E) I’d like to test with the same pins and configuration you were using to remove as many variables as possible.

lublak:
@macgyver2089 this is currently hard, i switched to a custom based esp32 code for my project (not anymore on esphome).
I used the ESP32-DevKitC V4 and the wiring was:
grafik
But i lost the configuration, but that should not be hard (written now, with my old wiring):

```yaml
# Example configuration entry
font:
  - file: 'fonts/Comic Sans MS.ttf'
    id: font1
    size: 8

spi:
  clk_pin: D18
  mosi_pin: D32

display:
  - platform: waveshare_epaper
    cs_pin: D5
    dc_pin: D17
    busy_pin: D4
    reset_pin: D16
#    inverted: true <= currently i need to check again if its needed be careful! 
    model: 7.30in-e
    full_update_every: 30
    lambda: |-
      it.print(0, 0, id(font1), "Hello World!");
```

macgyver2089:
@lublak Thanks! I had already tested it before you had replied ….I have been running with the reset pin as inverted and everything seems to be fine and work …only oddity I have noticed is that my refresh/prints (code told me partial refresh is not supported) do take about 25 seconds sooo I did increase the timeout so the log would stop printing out the timeout issues. But your code does seem to works as intended. 

ESP32-S3-DevKitC

lublak:
@macgyver2089 yes partial refresh is nut supported by the panel. That's correct # (sadly) and it needs 16 seconds for a full refresh I think.
But still "fast" for 6 colors! I will try this week to merge the upstream to my branch. Should be not that hard.

lublak:
@macgyver2089 i now merge upstream

iclanton:
I'm looking forward to being able to use this. @clydebarrow - what needs to happen to get this in?

macgyver2089:
I will add "it works" but it's not using the colors effectively I'm not sure where to add to the override to stop flattening the colors.

lublak:
@macgyver2089 
I currently map different colors to different hex codes. 
You have to keep the colors flat here, as the display ‘only’ supports 6 colors.
You can use https://en.wikipedia.org/wiki/Color_quantization to have "more" colors.

https://github.com/esphome/esphome/pull/8416/files#diff-c38cbef403570140a68a1ca5765902bc982f3edc8ff8dac1c25d2842a341d8f1R204

```cpp
uint8_t WaveshareEPaper6C::color_to_hex(Color color) {
  uint8_t hex_code;
  if (color.red > 127) {
    if (color.green > 170) {
      if (color.blue > 127) {
        hex_code = 0x1;  // White
      } else {
        hex_code = 0x2;  // Yellow
      }
    } else {
      hex_code = 0x3;  // Red (or Magenta)
    }
  } else {
    if (color.green > 127) {
      if (color.blue > 127) {
        hex_code = 0x5;  // Cyan -> Blue
      } else {
        hex_code = 0x6;  // Green
      }
    } else {
      if (color.blue > 127) {
        hex_code = 0x5;  // Blue
      } else {
        hex_code = 0x0;  // Black
      }
    }
  }

  return hex_code;
}
```

limengdu:
I've added this new feature of PR to the Waveshare documentation.

raccettura:
Can this PR also add the waveshare 13.3 variant?  Seems like the same display just 1600×1200px.

jesserockz:
Thanks for the work @lublak.
I have implemented this display in a new `epaper_spi` display platform in #10462 which should pave the way for better maintainability with the epaper displays going forward in ESPHome.

Please let me know if that works for you. I have tested locally on one I have.

Jesse

lublak:
@jesserockz oh i will look into it! Give me one or two days i will get back to you! :smile:

lublak:
@jesserockz hi :) i looked into it. For me it would work, but: Currently its a generic epaper_spi adapter.

```
void EPaper7p3InE::power_on() {
  ESP_LOGI(TAG, "Power on the display");
  this->command(0x04);
  this->waiting_for_idle_ = true;
}
```

So all Epaper7p3InE send a 0x04 command. 
I'm not really familiar with this topic. But is the command standardised across all models? Waveshare, etc.?

```
static constexpr const uint16_t WIDTH = 800;
static constexpr const uint16_t HEIGHT = 480;
```

And another idea, if we're generalising here, we could also generalise the ‘sizes’.
7in; 13in etc.
As far as I know, everyone uses the same command set (at least at waveshare), just different ‘sizes’. So you could define this directly in yaml. And that would make it even more generic.
Only "Hat"-Types are need to be defined: H; E; B; F; etc.

raccettura:
I have a 13in sitting on my desk right now begging to be ran off esphome ;)

PR #8453, opened 2025-03-23
[esp32_can] Configurable enqueue timeout

### LINKED ISSUE #3317: CAN Bus triggers watchdog reboot when nothing responds - no ACK
### The problem

Background info:
2 Devices total on the CAN bus, ESP32 with TJA1050 and Inverter.
If the inverter is powered off, disconnected or not Acknowledging messages there are no other devices on the CAN bus to acknowledge the CAN messages.

The CAN Bus triggers a watchdog timer reboot loop, resulting in safe boot.
This occurs when nothing ACKs the CAN bus messages, eg if the receiver is busy or goes offline.
I think the CAN bus controller keeps retrying to send messages, I suspect when more messages are sent it takes too long and triggers the watchdog timer.

Is there an option to detect this and set the controller so it does not require an ACK, rather than a WDT reboot??
Or just not trigger a WDT reboot?

Further info on No Ack:
No Ack Mode: The No Acknowledgement mode is similar to normal mode, however acknowledgements are not required for a message transmission to be considered successful. This mode is useful when self testing the TWAI controller (loopback of transmissions).

```
enumerator TWAI_MODE_NO_ACK[](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html#_CPPv4N11twai_mode_t16TWAI_MODE_NO_ACKE)
Transmission does not require acknowledgment. Use this mode for self testing
```
 

2022.5.1

pip

2022.4.7

ESP32-IDF

### Board

esp32doit-devkit-v1

can-bus

### Example YAML snippet

```yaml
# Example to demonstrate CAN bus WDT trigger and reboot loop 
esphome:
  name: test-esp32-send-can

esp32:
  board: esp32doit-devkit-v1
  framework:
    type: esp-idf
    version: latest

# Enable logging
logger:

# Enable Home Assistant API
api:

ota:

wifi:
  ssid: !secret wifi_ssid

canbus:
  - platform: esp32_can
    tx_pin: GPIO23
    rx_pin: GPIO22
    can_id: 4
    bit_rate: 500kbps

interval:
  - interval: 900ms
    then:
      - canbus.send:
          can_id: 0x359
          data: [0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]
      - canbus.send:
          can_id: 0x351
          data: [0x28, 0x02, 0x86, 0x03, 0xE8, 0x03, 0x00, 0x00]
      - canbus.send:
          can_id: 0x355
          data: [0x1E, 0x00, 0x64, 0x00, 0x00, 0x00, 0x00, 0x00]
      - canbus.send:
          can_id: 0x356
          data: [0xEF, 0x14, 0x00, 0x00, 0x22, 0x01, 0x00, 0x00]
      - canbus.send:
          can_id: 0x354
          data: [0x04, 0xC0, 0x00, 0x1F, 0x03, 0x00, 0x00, 0x00]
```

```txt
[10:51:30][D][canbus:033]: send extended id=0x351 rtr=FALSE size=8
[10:51:31][D][canbus:033]: send extended id=0x355 rtr=FALSE size=8
[10:51:32][D][canbus:033]: send extended id=0x356 rtr=FALSE size=8
[10:51:33][D][canbus:033]: send extended id=0x354 rtr=FALSE size=8
[10:51:34]E (10809) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
[10:51:34]E (10809) task_wdt:  - loopTask (CPU 1)
[10:51:34]E (10809) task_wdt: Tasks currently running:
[10:51:34]E (10809) task_wdt: CPU 0: IDLE
[10:51:34]E (10809) task_wdt: CPU 1: IDLE
[10:51:34]E (10809) task_wdt: Aborting.
[10:51:34]
[10:51:34]abort() was called at PC 0x400e4b4c on core 0
[10:51:34]
[10:51:34]Backtrace:0x40081bb6:0x3ffb0800 0x40088f5d:0x3ffb0820 0x4008f006:0x3ffb0840 0x400e4b4c:0x3ffb08b0 0x4008287d:0x3ffb08d0 0x40148b53:0x3ffbbad0 0x400e4dea:0x3ffbbaf0 0x4008a300:0x3ffbbb10
←[33mWARNING Found stack trace! Trying to decode it←[0m
←[33mWARNING Decoded 0x40081bb6: panic_abort at C:\Users\Paul\.platformio\packages\framework-espidf\components\esp_system/panic.c:404←[0m
←[33mWARNING Decoded 0x40088f5d: esp_system_abort at C:\Users\Paul\.platformio\packages\framework-espidf\components\esp_system/system_api.c:112←[0m
←[33mWARNING Decoded 0x4008f006: abort at C:\Users\Paul\.platformio\packages\framework-espidf\components\newlib/abort.c:46←[0m
←[33mWARNING Decoded 0x400e4b4c: task_wdt_isr at C:\Users\Paul\.platformio\packages\framework-espidf\components\esp_common\src/task_wdt.c:182 (discriminator 1)←[0m
←[33mWARNING Decoded 0x4008287d: _xt_lowint1 at C:\Users\Paul\.platformio\packages\framework-espidf\components\freertos\port\xtensa/xtensa_vectors.S:1105←[0m
←[33mWARNING Decoded 0x40148b53: cpu_ll_waiti at C:\Users\Paul\.platformio\packages\framework-espidf\components\hal\esp32\include/hal/cpu_ll.h:183
 (inlined by) esp_pm_impl_waiti at C:\Users\Paul\.platformio\packages\framework-espidf\components\esp_pm/pm_impl.c:827←[0m
←[33mWARNING Decoded 0x400e4dea: esp_vApplicationIdleHook at C:\Users\Paul\.platformio\packages\framework-espidf\components\esp_common\src/freertos_hooks.c:63←[0m
←[33mWARNING Decoded 0x4008a300: prvIdleTask at C:\Users\Paul\.platformio\packages\framework-espidf\components\freertos/tasks.c:3846←[0m
[10:51:34]
[10:51:34]
[10:51:34]ELF file SHA256: a32cf67929d8eb53
[10:51:34]
[10:51:34]Rebooting...
```

### Additional information

_No response_

patagonaa:

This reduces the CAN TX queueing timeout from 1000ms to 10ms to avoid blocking very long when trying to send messages to an unresponsive device / disconnected bus (causing a "Component xxx took a long time for an operation (1000ms)" message and possibly triggering the loop watchdog).

This also adds a warning message when a CAN message couldn't be queued.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/3317
- 

- esphome/esphome-docs#4774

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
canbus:
  - platform: esp32_can
    id: can
    tx_pin: GPIO27
    rx_pin: GPIO26
    use_extended_id: true
    can_id: 0x100
    bit_rate: 125kbps

interval:
  - interval: 2s
    then: 
      - canbus.send:
          data: [ 0x11, 0x22, 0x33 ]
```

## Checklist:


danielschramm (approved):
Good idea. I vote for merging.

esphome/components/esp32_can/esp32_can.cpp
```diff
@@ -124,7 +124,7 @@ canbus::Error ESP32Can::send_message(struct canbus::CanFrame *frame) {
     memcpy(message.data, frame->data, frame->can_data_length_code);
   }
 
-  if (twai_transmit(&message, pdMS_TO_TICKS(1000)) == ESP_OK) {
+  if (twai_transmit(&message, pdMS_TO_TICKS(10)) == ESP_OK) {
```
DjordjeMandic: `ticks_to_wait` in milliseconds parameter could be configurable in yaml. (should be done in another PR after this gets merged).

DjordjeMandic (approved):
LGTM

esphome/components/esp32_can/esp32_can.cpp
clydebarrow: No, that should be done in this PR. Also the default should be speed-dependent, at the minimum speed 10ms is going to be way too short.

patagonaa: @clydebarrow 
Be aware that this is not a _transmit timeout_ but a _transmit queue timeout_, so unless you are publishing lots of messages at once (faster than can be sent) or the bus is dead, this will never be hit (default queue length is 5, so by default up to 5 messages can be queued at once without blocking at all).
Also for the slowest common CAN speed (125kbit/s AFAIK, which is ~1ms per message) this is fast enough for multiple transmit tries.

Yes, to do this 100% correct, the value should be configurable and speed-dependent by default, but for 99.9% of use cases this is good enough (right now, everything hangs and the whole MCU possibly crashes once the TX queue is full, so this didn't seem to be much of an issue in the past).

DjordjeMandic: > No, that should be done in this PR. Also the default should be speed-dependent, at the minimum speed 10ms is going to be way too short.

Good point, I wasn’t really thinking about the speed factor. Shorter timeouts work fine at higher speeds, but at super slow rates, they could definitely cause issues.

DjordjeMandic: > Also for the slowest common CAN speed (125kbit/s AFAIK, which is ~1ms per message) this is fast enough for multiple transmit tries.
> 
> Yes, to do this 100% correct, the value should be configurable and speed-dependent by default, but for 99.9% of use cases this is good enough (right now, everything hangs and the whole MCU possibly crashes once the TX queue is full, so this didn't seem to be much of an issue in the past).

I'm not sure how CAN is implemented here, but if the bus is busy or a higher-priority ID takes over, the message should stay in the queue. For example, if a high-priority device fully utilizes the bus (like during a large firmware update), some messages could be lost if the bus stays busy for more than 10ms.

After reconsidering my previous suggestion, I think it makes sense to at least add a basic config entry for the queue timeout. I'd set the default to 500ms—this way, with five queue items, we'd have a total of 2.5s, which should avoid triggering the watchdog timer.

The bigger issue here is a busy bus rather than the actual packet transmission speed. While 10ms is plenty for sending a packet, it’s not nearly enough time to wait for the bus to become available.

clydebarrow: > or the bus is dead

This is the point - that's typically when the timeout comes into play. Having nobody to talk to shouldn't block the loop for a whole second, but it does at present (had me scratching my head when I first encountered it.)

> slowest common CAN speed (125kbit/s AFAIK

The canbus component allows speeds as low as 5kbits/s, so that should be accommodated whether it's common or not - at that speed 1000ms _is_ probably a reasonable timeout.

> we'd have a total of 2.5s, which should avoid triggering the watchdog timer.

Triggering the watchdog timer shouldn't be the sole objective - blocking the loop for even 500ms is unacceptable. It might be unavoidable at the very lowest bit rates, but in most cases a few ms is all that would be needed.

So the bottom line is that it should be configurable, and the default should be speed dependent. It's not a difficult implementation.

patagonaa: > if the bus is busy or a higher-priority ID takes over, the message should stay in the queue

It does, there is no _transmit timeout_, but if the queue is full, _something_ has to be done with the message, and dropping the message (after a period of time where a queue space should be free again under normal circumstances) and notifying the caller is the right thing to do IMO. If you absolutely may not drop messages and have to consider the bus being busy for a long time, the tx queue length should be increased instead (this is already configurable).

In the MCP2515 component, CAN messages that cannot be queued are just dropped immediately, which is still better than hanging the entire loop (though tbh, it isn't great that the canbus commponent just silently ignores errors from the underlying can component).

> I'd set the default to 500ms—this way, with five queue items, we'd have a total of 2.5s, which should avoid triggering the watchdog timer

It doesn't trigger the watchdog, but it still hangs the entire loop and is unnecessary under normal circumstances.  
Consider a component that sends data every second: after 5 seconds (when the queue is full), each second, the loop() will take 500ms once, which will cause all other components to become unresponsive.  
I'd rather drop some messages than hang the entire application for 500ms every second.

patagonaa: > Having nobody to talk to shouldn't block the loop for a whole second, but it does at present



> The canbus component allows speeds as low as 5kbits/s, so that should be accommodated whether it's common or not



> Triggering the watchdog timer shouldn't be the sole objective - blocking the loop for even 500ms is unacceptable



> bottom line is that it should be configurable, and the default should be speed dependent

yes, yes, yes, and yes 👍 

I'll see when I can get around to implementing it, you can ping me again in a few days if I forget (though I'll probably do it later today)

PS: I think you sent your message while I still wrote mine - just saw yours now 😅

patagonaa: I'm currently thinking of calling this `tx_queue_add_timeout`, any other suggestions/proposals? I'd like to avoid calling it `tx_queue_timeout` because it isn't a _timeout of the queued elements_ but rather a timeout for _adding messages to the queue_

Edit: `tx_enqueue_timeout` maybe?

clydebarrow: I think tx_queue_timeout would be fine, since it is the act of queuing that it affects, and from the user's perspective the exact mechanism it controls is of little significance.

esphome/components/canbus/__init__.py
```diff
@@ -68,6 +68,32 @@ def validate_raw_data(value):
...
+    "80KBPS": 80_000,
+    "83K3BPS": 83_300,
+    "95KBPS": 95_000,
+    "100KBPS": 100_000,
+    "125KBPS": 125_000,
+    "200KBPS": 200_000,
+    "250KBPS": 250_000,
+    "500KBPS": 500_000,
+    "800KBPS": 800_000,
+    "1000KBPS": 1_000_000,
```
clydebarrow: I'm not thrilled by the repetition of the same list of string constants. What about a mapping function:
```
def get_rate(value):
    match = re.match(r"(\d+)(?:K(\d+)?)?BPS", value, re.IGNORECASE)
    if not match:
        raise ValueError(f"Invalid rate format: {value}")
    fraction = match.group(2) or "0"
    return int((float(match.group(1)) + float(f"0.{fraction}")) * 1000)
```
Or just build this into the `get_default_tx_enqueue_timeout` function.

This would also avoid modifying the parent canbus component, and would be simple enough to update if the format of rates ever changes.

patagonaa:
As discussed, I added a config option for the enqueue timeout.  
By default, the timeout is set to the length of 10 messages (based on the configured baud rate), which should be enough to free a spot in the queue, even on a pretty congested bus with a bunch of collisions.

Naming-wise, I went with `tx_enqueue_timeout` because it's a little clearer because from the user perspective, it does make a difference imo:  
For `tx_queue_timeout`, as a user I would expect the queued messages to disappear from the queue if the timeout elapses (which they don't) and wonder why old messages are sent once the bus is reattached.  
Also, I would wonder why it takes `tx_queue_length * tx_queue_timeout` messages until one times out.

esphome/components/esp32_can/canbus.py
```diff
@@ -23,6 +25,8 @@
 CODEOWNERS = ["@Sympatron"]
 DEPENDENCIES = ["esp32"]
 
+CONF_TX_ENQUEUE_TIMEOUT = 'tx_enqueue_timeout'
```
clydebarrow:
```suggestion
CONF_TX_ENQUEUE_TIMEOUT = "tx_enqueue_timeout"
```
Just keeping consistent style. Good choice for the name BTW.

clydebarrow (changes requested):

esphome/components/canbus/__init__.py
patagonaa: Hmm, I just thought mapping the values explicitly was nicer than doing string manipulation, but I'll go with your change if you feel like that's better.

Also, due to the added log message, I'm already touching the `canbus` component anyway (and I think it belongs there as that functionality might be useful in other non-esp-specific places).

clydebarrow: Generally speaking, it's a Bad Idea to repeat the same constant, that's why the `CONF_` stuff is all defined, because it's so easy to make mistakes. Your assert checks that, but that's itself a sign of an underlying risk. The function is slightly less readable, but still obvious in its purpose and more maintainable and extensible in the long term.

I would actually like to see the bit rate setting be more flexible than a fixed set of strings, but that's for another day.

clydebarrow (approved):

clydebarrow:
Ok, just need a docs update now.

patagonaa:
> Ok, just need a docs update now.

already on it 😄

patagonaa:
(done, btw)

clydebarrow (approved):

PR #8457, opened 2025-03-24
[ethernet] P4 changes and 5.3.0 deprecated warnings 

swoboda1337:

Currently clk_mode controls both the mode and the pin. On P4 etc the clock pin is not fixed to only a few options the pin can be anything.  The type emac_rmii_clock_gpio_t is defined as an int on platforms other than the original esp32. Separate clk_mode into two configs, clk_mode and clk_pin, this is a breaking change but it has to be done. 

Work around IDF compile issue on P4, fields in ETH_ESP32_EMAC_DEFAULT_CONFIG are in the wrong order, copy that define and fix the order, can be removed once this pr is merged and released: https://github.com/espressif/esp-idf/pull/15637 

Fix deprecated warnings in 5.3.0 and later as `smi_mdc_gpio_num` and `smi_mdio_gpio_num` have been deprecated.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4769

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
ethernet:
  type: IP101
  mdc_pin: GPIO31
  mdio_pin: GPIO52
  power_pin: GPIO51
  clk:
    mode: CLK_EXT_IN
    pin: GPIO50
  phy_addr: 1
```

## Checklist:


randybb:
Tested with LAN8720 (olimex ESP32-POE-ISO). But I think it needs the breaking-change label as syntax for the clock pin has been changed.

esphome/components/ethernet/__init__.py
```diff
@@ -181,7 +166,11 @@ def _validate(config):
         {
             cv.Required(CONF_MDC_PIN): pins.internal_gpio_output_pin_number,
             cv.Required(CONF_MDIO_PIN): pins.internal_gpio_output_pin_number,
-            cv.Optional(CONF_CLK_MODE, default="GPIO0_IN"): cv.enum(
+            cv.Required(CONF_CLK_PIN): cv.Any(
+                pins.internal_gpio_output_pin_number,
+                pins.internal_gpio_input_pin_number,
+            ),
```
jesserockz:
```suggestion
            cv.Required(CONF_CLK_PIN): pins.internal_gpio_pin_number,
```

esphome/components/ethernet/ethernet_component.cpp
```diff
@@ -129,19 +145,27 @@ void EthernetComponent::setup() {
...
-#if ESP_IDF_VERSION_MAJOR >= 5
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
+  eth_esp32_emac_config_t esp32_emac_config = ETH_ESP32_EMAC_DEFAULT_CONFIG();
+  esp32_emac_config.smi_gpio.mdc_num = this->mdc_pin_;
+  esp32_emac_config.smi_gpio.mdio_num = this->mdio_pin_;
+  esp32_emac_config.clock_config.rmii.clock_mode = this->clk_mode_;
+  esp32_emac_config.clock_config.rmii.clock_gpio = this->clk_pin_;
+
+  esp_eth_mac_t *mac = esp_eth_mac_new_esp32(&esp32_emac_config, &mac_config);
+#elif ESP_IDF_VERSION_MAJOR >= 5
```
jesserockz: This should be always

jesserockz (changes requested):
As mentioned in discord, lets leave `clk_mode`, add new `clk` map and auto convert it eg:
```yaml
clk_mode: GPIO0_IN
```
to
```yaml
clk:
  pin: GPIO0
  mode: CLK_EXT_IN
```

esphome/components/ethernet/__init__.py
swoboda1337: > cv.Required(CONF_CLK_PIN): pins.internal_gpio_pin_number,

You can't actually do that, I tried:
```
yaml.representer.RepresenterError: ('cannot represent an object', <function _internal_number_creator.<locals>.validator at 0x70a952175a80>)
```

swoboda1337: I was going to try and figure out why but I forgot to look into it `pins.internal_gpio_pin_number` isn't used anywhere probably because its broken.

swoboda1337: Looks like `pins.internal_gpio_pin_number` is only used in the i2c code but its used like this:
```
pin_with_input_and_output_support = pins.internal_gpio_pin_number(
    {CONF_OUTPUT: True, CONF_INPUT: True}
)
```

esphome/components/ethernet/__init__.py
```diff
@@ -181,8 +185,17 @@ def _validate(config):
...
+                CLK_MODES_DEPRECATED, upper=True, space="_"
+            ),
+            cv.Optional(CONF_CLK): cv.Schema(
+                {
+                    cv.Required(CONF_MODE): cv.enum(CLK_MODES, upper=True, space="_"),
+                    cv.Required(CONF_PIN): cv.Any(
+                        pins.internal_gpio_output_pin_number,
+                        pins.internal_gpio_input_pin_number,
+                    ),
+                }
```
swoboda1337: As mentioned in the outdated review I can't use `pins.internal_gpio_pin_number` here

esphome/components/ethernet/__init__.py
jesserockz: I'm gonna fix that as it is misleading then

jesserockz: https://github.com/esphome/esphome/pull/9270

esphome/components/ethernet/__init__.py
```diff
@@ -177,14 +184,21 @@ def _validate(config):
         cv.Optional(CONF_PAGE_ID): cv.hex_int,
     }
 )
+CLK_SCHEMA = cv.Schema(
+    {
+        cv.Required(CONF_MODE): cv.enum(CLK_MODES, upper=True, space="_"),
+        cv.Required(CONF_PIN): pins.internal_gpio_pin_number,
```
swoboda1337, line 190: @jesserockz does this now require the pin to be both an input and an output? I had cv.Any so the pin could be an input, output or both. The clock can be configured to be an input or output, No guarantee the pin will be both. I am not sure if the clock pin will ever be on an output or input only pin but who knows.

jesserockz: No, this simply allows it to be either. This is checked in the respective `esp32/gpio_esp32_{variant}.py` files to ensure the pin is allowed to be used for both input and output and is not a restricted pin.

jesserockz (approved):

PR #8487, opened 2025-03-31
[mk2pvrouter] Add Mk2PVRouter component with sensor support

FredM67:

Add support for Mk2PVRouter telemetry, inspired from TeleInfo.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#7165

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [X] ESP8266
- [X] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
mk2pvrouter:
  uart_id: uart_bus

sensor:
  - platform: mk2pvrouter
    name: "Puissance instantanée"
    tag: "P"
    id: p_grid
    device_class: power
    icon: "mdi:flash"
    state_class: measurement
    unit_of_measurement: W
    accuracy_decimals: 0
    web_server:
      sorting_group_id: sorting_group_sensors

  # V and V1 (and other voltage/temperature tags) are reported by the
  # device in centi-units, so a `multiply: 0.01` filter is needed to
  # get plain volts/°C below.
  - platform: mk2pvrouter
    tag: "V"
    name: "Tension"
    id: v_grid
    device_class: voltage
    icon: "mdi:sine-wave"
    state_class: measurement
    unit_of_measurement: V
    accuracy_decimals: 2
    filters:
      - multiply: 0.01

  - platform: mk2pvrouter
    tag: "V1"
    name: "Tension L1"
    id: v1_grid
    device_class: voltage
    icon: "mdi:sine-wave"
    state_class: measurement
    unit_of_measurement: V
    accuracy_decimals: 2
    filters:
      - multiply: 0.01
```

## Checklist:


FredM67:
Runs since a couple of weeks w/o any problem on an ESP32.

FredM67:
Still running without known issues....

FredM67:
@esphbot review

swoboda1337:
@esphbot review

esphbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - Buffer-full path never resets `buf_index_` — component latches up permanently
> - Empty group causes out-of-bounds read and an unsigned underflow loop in the CRC path
> - Default filters are always prepended, silently double-scaling user-supplied filters

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

esphbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - `init_listeners()` can be emitted *after* the listener registrations, silently dropping every entity

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bdraco:
@esphbot gh_request

Suggest a split so this is reviewable by humans. Usually we only accept one new platform per PR

FredM67:
Per the review suggestion above, this PR has been split into 4 stacked PRs for easier review:

1. **This PR** — hub component + bare `sensor` platform (no tag-based defaults)
2. `mk2pvrouter-sensor-defaults` (base: this branch) — tag-based sensor defaults + unit tests
3. `mk2pvrouter-binary-sensor` (base: this branch) — `binary_sensor` platform
4. `mk2pvrouter-text-sensor` (base: this branch) — `text_sensor` platform

PRs 2-4 will be opened shortly and linked here.

bdraco:
Please don't forget to trim the docs pr as well

esphome/components/mk2pvrouter/mk2pvrouter.cpp
```diff
@@ -0,0 +1,251 @@
...
+        this->publish_value_(this->tag_, this->val_);
+
+        /* Advance buf_finger past end of group: value + TAB + CRC + CR */
+        buf_finger += field_len + 1 + 1 + 1;
+        if (buf_finger >= buf_end)
+          break;
+      }
+      this->buf_index_ = 0;
+      this->state_ = State::WAITING_FOR_START;
+      break;
```
bdraco, line 151: might be able to simplify this logic a bit

esphome/components/mk2pvrouter/mk2pvrouter.h
```diff
@@ -0,0 +1,76 @@
...
+    START_FRAME_RECEIVED,
+    END_FRAME_RECEIVED,
+  };
+
+  FixedVector<Mk2PVRouterListener *> mk2pvrouter_listeners_{};
+  uint16_t buf_index_{0};
+  State state_{State::WAITING_FOR_START};
+  char tag_[MAX_TAG_SIZE];
+  char val_[MAX_VAL_SIZE];
+  char buf_[MAX_BUF_SIZE];  // Large buffer last to reduce padding
```
bdraco: align members largest to smallest

esphome/components/mk2pvrouter/mk2pvrouter.h
```diff
@@ -0,0 +1,76 @@
...
+
+ protected:
+  static constexpr size_t CRC_SUFFIX_LEN = 1;
+  static constexpr uint32_t BAUD_RATE = 9600;
+
+  enum class State : uint8_t {
+    WAITING_FOR_START,
+    START_FRAME_RECEIVED,
+    END_FRAME_RECEIVED,
+  };
```
bdraco, line 52: some padding waste here due to alignment?

esphome/components/mk2pvrouter/mk2pvrouter.h
```diff
@@ -0,0 +1,76 @@
...
+static constexpr uint16_t MAX_BUF_SIZE = 256;  // Full frame with all features enabled
+
+/**
+ * @class Mk2PVRouterListener
+ * @brief Listener interface for receiving updates from the Mk2PVRouter.
+ *
+ * This class allows other components to register as listeners to receive updates
+ * for specific tags published by the Mk2PVRouter.
+ */
+class Mk2PVRouterListener {
```
bdraco, line 24: Trim AI essay comments, this repo prefers terse comments

esphome/components/mk2pvrouter/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+    listener_count: int = 0
+
+
+def _get_data() -> Mk2PVRouterData:
+    if DOMAIN not in CORE.data:
+        CORE.data[DOMAIN] = Mk2PVRouterData()
+    return CORE.data[DOMAIN]
+
+
+def final_validate(config: ConfigType) -> ConfigType:
```
bdraco:
```suggestion
def final_validate(config: ConfigType) -> None:
```

esphome/components/mk2pvrouter/mk2pvrouter.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * @return uint8_t The calculated CRC value.
+ */
+uint8_t Mk2PVRouter::calculate_crc_(const char *grp, size_t grp_len) {
+  uint8_t crc_tmp{0};
+  const auto effective_len = grp_len - CRC_SUFFIX_LEN;
+  for (size_t i = 0; i < effective_len; i++) {
+    crc_tmp += grp[i];
+  }
+  crc_tmp &= 0x3F;
+  crc_tmp += 0x20;
```
bdraco: nit: magic numbers here could be constexpr

esphome/components/mk2pvrouter/mk2pvrouter.cpp
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/**
+ * @brief Dumps the Mk2PVRouter configuration to the log.
+ *
+ * @note Logs the UART settings and other configuration details.
+ */
+void Mk2PVRouter::dump_config() {
+  ESP_LOGCONFIG(TAG, "Mk2PVRouter:");
+  ESP_LOGCONFIG(TAG, "  Listeners: %zu", this->mk2pvrouter_listeners_.size());
```
bdraco: Please follow logging best practices
https://developers.esphome.io/architecture/logging/#the-solution-combine-related-log-messages

esphome/components/mk2pvrouter/mk2pvrouter.h
```diff
@@ -0,0 +1,76 @@
...
+  static constexpr size_t CRC_SUFFIX_LEN = 1;
+  static constexpr uint32_t BAUD_RATE = 9600;
+
+  enum class State : uint8_t {
+    WAITING_FOR_START,
+    START_FRAME_RECEIVED,
+    END_FRAME_RECEIVED,
+  };
+
+  FixedVector<Mk2PVRouterListener *> mk2pvrouter_listeners_{};
```
bdraco: If this component is single instance this could be StaticVector and calced at codegen time instead. If its not single instance than FixedVector is fine

esphome/components/mk2pvrouter/mk2pvrouter.cpp
esphbot: Concrete simplification available. Replace `buf_finger += field_len + 1 + 1 + 1;` and the `>= buf_end` guard with `buf_finger = grp_end;`.

Outer `memchr` for LINE_FEED already handles termination. Success path then matches all three `continue` paths. Also drops an unchecked assumption: `+1+1+1` hardcodes a one-char CRC without referencing `CRC_SUFFIX_LEN`.

esphome/components/mk2pvrouter/mk2pvrouter.h
esphbot: Current order already optimal. Measured on 32-bit: `FixedVector` 12 bytes (ptr + two `size_t`), `buf_index_` 2, `state_` 1, then 8 + 8 + 256 of align-1 `char`. Total 287, rounded to 288 — one byte of tail padding.

Largest-to-smallest gives the identical 287. Char arrays have alignment 1, so no ordering recovers that byte.

esphome/components/mk2pvrouter/mk2pvrouter.h
esphbot: No waste found. Sum of member sizes 287; `sizeof` 288. One byte tail padding, the minimum for align-4.

The `// Large buffer last to reduce padding` note at line 69 already reflects this. Nothing left to reclaim.

esphome/components/mk2pvrouter/mk2pvrouter.h
esphbot: Still outstanding on head. Three blocks in the header (lines 8-16, 21-27, 39-45) plus `@brief`/`@param`/`@return` above all seven functions in the .cpp — roughly 60 of 251 lines.

Two worth keeping: the buffer-size rationale at 10-15 justifies `MAX_TAG_SIZE`/`MAX_VAL_SIZE`, and the frame diagram at 164-172 documents the CRC span. Rest restates the signatures. `teleinfo.cpp` covers the same protocol with one-line `/* */` comments.

esphome/components/mk2pvrouter/__init__.py
esphbot: Correct. Return value discarded at `esphome/config.py:1208` — `self.comp.final_validate_schema(conf)`, result unused.

So `-> ConfigType` promises a contract that does not exist. Change to `-> None` and call `schema(config)` for the raise side effect only.

esphome/components/mk2pvrouter/mk2pvrouter.cpp
esphbot: Still present at lines 54-55. Suggest `CRC_MASK = 0x3F` and `CRC_ASCII_OFFSET = 0x20`.

Naming also documents the valid CRC range, 0x20-0x5F. That range makes the `char` vs `uint8_t` comparison in `check_crc_` provably safe on signed-char targets.

esphome/components/mk2pvrouter/mk2pvrouter.cpp
esphbot: Still two calls at lines 237-238. Same in `sensor/mk2pvrouter_sensor.cpp:19-22`.

Repo pattern at `modbus_controller.cpp:483` — one `ESP_LOGCONFIG` with embedded `\n`. Merging drops a call site and a format string, saving flash.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

FredM67:
Thanks for the thorough review! Pushed a commit addressing the feedback:

- `final_validate` return type fixed to `-> None`, matching the convention used by `dsmr`/`emontx`/`i2c`/`modbus` (the return value isn't used by esphome's final-validation mechanism; `schema(config)` is called for its validation side-effect only).
- Magic numbers in `calculate_crc_` replaced with named constants (`CRC_MASK`, `CRC_OFFSET`).
- `dump_config()`'s two `ESP_LOGCONFIG` calls combined into one per the [logging guidelines](https://developers.esphome.io/architecture/logging/#the-solution-combine-related-log-messages).
- Trimmed the Doxygen-style `/** @brief ... */` comment blocks throughout `mk2pvrouter.h`/`.cpp` down to short one-liners (or removed where the code was already self-explanatory).
- Tightened up the comments in the `END_FRAME_RECEIVED` group-parsing loop; kept the logic (including the `buf_finger >= buf_end` guard) unchanged, since that guard prevents a negative-length `memchr` call on the next iteration.

Two items didn't need a code change:
- Struct member ordering: `mk2pvrouter_listeners_` → `buf_index_` → `state_` → `tag_`/`val_` → `buf_[256]` is already ordered largest-alignment-first to minimize padding (see the existing `// Large buffer last to reduce padding` comment).
- `FixedVector` vs `StaticVector` for `mk2pvrouter_listeners_`: `StaticVector` needs a compile-time size, but this component doesn't set `MULTI_CONF = False`, so multiple instances are allowed and the listener count is only known at `final_validate` time — `FixedVector` (sized at runtime via `init()`) is the right fit here.

Config validation passes on all 4 targets (esp32-ard, esp32-idf, esp8266-ard, rp2040-ard).

FredM67:
Follow-up on this review:

- Simplified the buffer advance after a successful field parse: since `grp_end` comes from a `memchr` search bounded by `buf_end`, it's always `< buf_end`, making the old `field_len`-based arithmetic plus its bounds re-check redundant. Replaced with `buf_finger = grp_end;`, which also unifies the success path with the existing `continue`-driven error paths. Pushed in `9cf34928f3`.
- Corrected the PR description's example: removed the stale `binary_sensor:` block (that platform moved to a separate PR in the split) and fixed the incorrect claim that the component applies a built-in `multiply: 0.01` default — it doesn't; the example now shows the filter explicitly.

Config validation re-verified on all 4 targets (esp32-ard, esp32-idf, esp8266-ard, rp2040-ard) after the change.

tests/test_build_components/common/uart_9600_even_7bits/esp32-ard.yaml
```diff
@@ -0,0 +1,14 @@
+# Common UART configuration for ESP32 Arduino tests - 9600 baud, EVEN parity, 7 data bits
+
```
bdraco, line 2: Drop esp32-ard.yaml as its already covered by esp32-idf.yaml.

Ard builds as an idf component and there is no ARD code here

bdraco: Note: comment was meant for the mk2pvrouter test file.  OK to keep the uart 9600 one

tests/components/mk2pvrouter/test.esp32-ard.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  uart_9600_even_7bits: !include ../../test_build_components/common/uart_9600_even_7bits/esp32-ard.yaml
+  mk2pvrouter: !include common.yaml
```
bdraco: We can drop the ard one as there is no ard specific code here

https://github.com/esphome/esphome/pull/8487/changes#r3876199502

FredM67:
Thanks for the thorough review! Going through each point:

**Already addressed by earlier commits** (`7fdee59e26`, `9cf34928f3` — these landed before some of your comments, based on timestamps):
- Loop simplification (`mk2pvrouter.cpp`) — `buf_finger = grp_end` now unifies the success path with the error `continue` paths (commit `9cf34928f3`).
- Magic numbers — `CRC_MASK`/`CRC_OFFSET` are now named `constexpr`.
- Logging — `dump_config()` now uses a single combined `ESP_LOGCONFIG` call.
- `final_validate` return type — already annotated `-> None`.
- Doxygen essay comments — trimmed to one-liners (e.g. `mk2pvrouter.h:21`/`33`).

**Checked, no change needed:**
- Member ordering/padding (`mk2pvrouter.h:52-57`) — I double-checked the actual layout: `FixedVector` (24B, 8-byte align) → `uint16_t` (2B) → `State` (1B) → `char[8]` → `char[8]` → `char[256]` gives zero *internal* padding; the struct only pads 5 bytes at the very end to satisfy 8-byte alignment, which is unavoidable regardless of member order given the FixedVector's alignment requirement. So it's already ordered largest-alignment-first with minimal waste.
- `FixedVector` vs `StaticVector` (`mk2pvrouter.h:52`) — `mk2pvrouter` is genuinely multi-instance (no `MULTI_CONF`, docs mention "multiple hubs"), and listener count varies per instance based on configured entities, so it can't be sized at codegen time. `FixedVector` is correct here.

**New fix (commit `20d1ad3f7d`):**
- Dropped `tests/components/mk2pvrouter/test.esp32-ard.yaml` — no Arduino-specific code path here, and ESP32 Arduino builds as an ESP-IDF component under the hood, so it's redundant with `esp32-idf`. Now matches the `teleinfo` component's test matrix (idf + esp8266-ard + rp2040-ard only). Kept the shared `uart_9600_even_7bits/esp32-ard.yaml` package per your note since other components use it.

tests/test_build_components/common/uart_9600_even_7bits/esp32-ard.yaml
FredM67: Confirmed, that's exactly what commit 20d1ad3f7d does — removed `tests/components/mk2pvrouter/test.esp32-ard.yaml`, kept the shared `uart_9600_even_7bits/esp32-ard.yaml` package intact for other components to use.

esphome/components/mk2pvrouter/mk2pvrouter.h
esphbot: Component single-instance. `MULTI_CONF` unset means `False` — `loader.py:77`, `getattr(module, "MULTI_CONF", False)`. `config.py:945` only splits into a list when truthy. Compare `teleinfo/__init__.py:8`, which sets `MULTI_CONF = True` explicitly.

So StaticVector applies. Count known at codegen: `cg.add_define` plus `StaticVector<Mk2PVRouterListener *, N>` drops the heap allocation and `init_listeners()`.

Caveat: linked docs mention multiple hubs. If that's intended, `MULTI_CONF = True` is missing, and the global count at `__init__.py:62-68` needs keying off `mk2pvrouter_id` first.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

FredM67:
Thanks for catching the MULTI_CONF gap and the comment accuracy issue — both are real, fixed in `134205e9fc`:

- **`MULTI_CONF = True`** added. You're right that the premise in my earlier reply was backwards — the schema never actually declared multi-instance support, so it silently defaulted to single-instance while the docs already promised `mk2pvrouter_id` for multiple hubs. Since multi-hub is the documented/intended design, this was the correct side to fix (not switching to StaticVector).
- **Buffer-sizing comment** rewritten to drop the incorrect `teleinfo.h`/`S_MC` attribution and cite mk2pvrouter's actual tag/value shapes instead.
- Re-ran `test_build_components.py -e config -c mk2pvrouter` — still 3/3 passing after the `MULTI_CONF` change.

On suggestion #2 (parser/CRC unit test coverage via the `ufm01` harness pattern) — agreed it's valuable and correctly flagged as non-blocking. I'll leave that as a fast-follow rather than growing this PR further; happy to open it as a tracked follow-up if that's preferred.

FredM67:
Correction on the buffer-sizing comment part of my previous reply (`134205e9fc`): I was wrong to remove the `teleinfo.h` reference.

`teleinfo.h` isn't esphome's `teleinfo` component — it's a real file in the Mk2PVRouter firmware itself ([PVRouter-1-phase](https://github.com/FredM67/PVRouter-1-phase/blob/main/Mk2_fasterControl_Full/teleinfo.h) / [PVRouter-3-phase](https://github.com/FredM67/PVRouter-3-phase/blob/main/Mk2_3phase_RFdatalog_temp/teleinfo.h)), which implements the exact telemetry protocol this component parses. It genuinely defines `S_MC` as a real 4-char tag and confirms the 6-digit signed value bound (the `P` tag). So the original comment's numbers and rationale were accurate all along.

Restored the comment in `01702816b3`, with a link added to make the provenance unambiguous going forward. The `MULTI_CONF = True` fix from `134205e9fc` still stands — that part of the review was correct.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

tests/components/mk2pvrouter/test.esp32-ard.yaml
esphbot: Confirmed done. `tests/components/mk2pvrouter/test.esp32-ard.yaml` absent at head, removed in `20d1ad3f7d`. Remaining targets: esp32-idf, esp8266-ard, rp2040-ard. Shared package `uart_9600_even_7bits/esp32-ard.yaml` kept per your follow-up.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

esphome/components/mk2pvrouter/mk2pvrouter.cpp
```diff
@@ -0,0 +1,182 @@
...
+          ESP_LOGE(TAG, "No group found");
+          break;
+        }
+
+        if (!this->check_crc_(buf_finger, grp_end))
+          continue;
+
+        field_len = get_field(this->tag_, buf_finger, grp_end, MAX_TAG_SIZE);
+        if (!field_len || field_len >= MAX_TAG_SIZE) {
+          ESP_LOGE(TAG, "Invalid tag.");
```
bdraco:
```suggestion
          ESP_LOGE(TAG, "Invalid tag");
```

Please follow https://developers.esphome.io/architecture/logging/

esphbot: Applied at head. `mk2pvrouter.cpp:136` now reads `ESP_LOGE(TAG, "Invalid tag");`. Sibling value path at line 143 keeps `%s` for the tag, which identifies which group failed. Marking resolved.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bdraco:
worth deleting the dead setup

esphome/components/mk2pvrouter/mk2pvrouter.cpp
FredM67: Fixed with commit 9dd8c67fcf

esphbot: Resolved at head. `buf_finger = grp_end` replaces pointer arithmetic (`9cf34928f3`). Group handling extracted into `process_group_` (`9dd8c67fcf`). Success path now matches error paths.

esphbot: Confirmed. `mk2pvrouter.cpp:107-150` reads cleanly now. Loop body handles framing only.

esphbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

esphome/components/mk2pvrouter/mk2pvrouter.h
```diff
@@ -0,0 +1,69 @@
...
+  static constexpr uint32_t BAUD_RATE = 9600;
+
+  enum class State : uint8_t {
+    WAITING_FOR_START,
+    START_FRAME_RECEIVED,
+    END_FRAME_RECEIVED,
+  };
+
+#ifdef MK2PVROUTER_LISTENER_COUNT
+  StaticVector<Mk2PVRouterListener *, MK2PVROUTER_LISTENER_COUNT> mk2pvrouter_listeners_{};
```
bdraco:
```suggestion
  StaticVector<Mk2PVRouterListener *, MK2PVROUTER_LISTENER_COUNT> mk2pvrouter_listeners_;
```

bdraco (dismissed):

bdraco (approved):

PR #8511, opened 2025-04-03
[api] Allow setting custom MAC address for HA

clydebarrow:

ESPHome sends its MAC address to HA, not for any networking use, but evidently to act as a unique identifier for the device. This causes problems if a hardware device fails and needs to be replaced with something identical - except that it will have a different MAC address. This results in HA refusing to believe that the device with the same name is the same device, and the only way to fix that is to remove the original device and re-add the "new" one. 

This PR provides a way to circumvent that problem, by allowing the previous MAC address to be reassigned to the new piece of hardware. This does not affect networking since the actual MAC address will still be used by the network components.

It's also helpful if a device needs to be temporarily replaced while some work on it is being done.

Other uses include assigning a locally-administered MAC address rather than relying on the pre-programmed one.

 
## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4792
## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
api:
  mac_address: 84:F7:03:E8:04:14
```

## Checklist:

### DIFF
--- a/esphome/components/api/api_server.cpp
+++ b/esphome/components/api/api_server.cpp
@@ -25,6 +25,9 @@ static const char *const TAG = "api";
 void APIServer::setup() {
   ESP_LOGCONFIG(TAG, "Setting up Home Assistant API server...");
   this->setup_controller();
+  if (this->mac_address.empty()) {
+    this->mac_address = std::move(get_mac_address_pretty());
+  }
   socket_ = socket::socket_ip(SOCK_STREAM, 0);
   if (socket_ == nullptr) {
     ESP_LOGW(TAG, "Could not create socket.");
@@ -139,6 +142,7 @@ void APIServer::loop() {
 void APIServer::dump_config() {
   ESP_LOGCONFIG(TAG, "API Server:");
   ESP_LOGCONFIG(TAG, "  Address: %s:%u", network::get_use_address().c_str(), this->port_);
+  ESP_LOGCONFIG(TAG, "  MAC address: %s", this->mac_address.c_str());
 #ifdef USE_API_NOISE
   ESP_LOGCONFIG(TAG, "  Using noise encryption: YES");
 #else
@@ -376,6 +380,8 @@ const std::vector<APIServer::HomeAssistantStateSubscription> &APIServer::get_sta
   return this->state_subs_;
 }
 uint16_t APIServer::get_port() const { return this->port_; }
+std::string APIServer::get_mac_address() const { return this->mac_address; }
+void APIServer::set_mac_address(std::string mac_address) { this->mac_address = std::move(mac_address); }
 void APIServer::set_reboot_timeout(uint32_t reboot_timeout) { this->reboot_timeout_ = reboot_timeout; }
 #ifdef USE_HOMEASSISTANT_TIME
 void APIServer::request_time() {
--- a/esphome/components/api/api_server.cpp
+++ b/esphome/components/api/api_server.cpp
@@ -25,8 +25,8 @@ static const char *const TAG = "api";
 void APIServer::setup() {
   ESP_LOGCONFIG(TAG, "Setting up Home Assistant API server...");
   this->setup_controller();
-  if (this->mac_address.empty()) {
-    this->mac_address = std::move(get_mac_address_pretty());
+  if (this->mac_address_.empty()) {
+    this->mac_address_ = std::move(get_mac_address_pretty());
   }
   socket_ = socket::socket_ip(SOCK_STREAM, 0);
   if (socket_ == nullptr) {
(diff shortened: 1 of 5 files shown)


clydebarrow:

```yaml
external_components:
  - source: github://pr#8511
    components: [api]
    refresh: 1h
```

---
(Added by the PR bot)

esphome/components/api/api_server.cpp
```diff
@@ -25,6 +25,9 @@ static const char *const TAG = "api";
 void APIServer::setup() {
   ESP_LOGCONFIG(TAG, "Setting up Home Assistant API server...");
   this->setup_controller();
+  if (this->mac_address_.empty()) {
+    this->mac_address_ = std::move(get_mac_address_pretty());
```
clydebarrow:
```suggestion
    this->mac_address_ = get_mac_address_pretty();
```

nagyrobi:
Finally!

paravoid:
I completely understand the problem statement. However quite honestly, I find the solution a little hacky and feel like I would be quite confused if I had experienced this as a user.

"MAC address" has a very specific meaning. I would expect a setting to alter the MAC address to actually alter what address is used on the wire in the Ethernet frames themselves. This is a setting often present in devices and operating systems, for exactly the reasons you mentioned. e.g. on Linux I can do `ip link set dev eth0 address $newmac` or `ifconfig eth0 hw ether $newmac`. Relatedly, having a setting called "MAC address" under "api_server" helps on one end, but feels like a clear hierarchy violation to me.

I realize this would be much more complicated, but perhaps this should be addressed in combination with changes to Home Assistant, renaming the identifier to some kind of "unique ID"/UID name instead, and have an ESPHome setting for it, with a default to the device's MAC address?

clydebarrow:
> "MAC address" has a very specific meaning. I would expect a setting to alter the MAC address to actually alter what address is used on the wire in the Ethernet frames themselves.

That would be a reasonable expectation, if it were in the `ethernet` or `wifi` config. The confusion here is that the mac address is not being used as a mac address but I agree that compounding that confusion is not a Good Idea.

> but perhaps this should be addressed in combination with changes to Home Assistant, renaming the identifier to some kind of "unique ID"/UID name instead

Well, that's a good idea, though it begs the question of what HA actually does with the identifier, and why it needs that in addition to the device name. Any change to HA should examine that first.

I did determine however that HA doesn't care what the "mac_address" looks like, so it's quite happy with any string for that value (albeit converted to upper case):

Screenshot 2025-04-04 at 9 24 40 AM

So as a first step, I propose that instead of the `mac_address` option in `api`, something like:

```
api:
  ha_identifier: mac_address  # the default, same behaviour as now
```
or:
```
api:
  ha_identifier: device_name # Use the `name` configured in the esphome block
```
or:
```
api:
  ha_identifier:
	source: custom
	value: 34:85:18:77:41:18 # Or any other desired string
```

Thoughts? Ideas for the option name?

bdraco:
Home Assistant has some specific rules when it comes to setting the unique ID. The link below might be helpful for some additional context:

https://developers.home-assistant.io/docs/entity_registry_index/?_highlight=burned#unique-id-requirements

bdraco:
Just to share a bit of context—ESPHome used to use the name as the unique ID, but that was switched to the MAC address instead: https://github.com/home-assistant/core/pull/83741.

It could be really helpful if there were a way to let Home Assistant know that a new device is meant to replace an old one. That way, it could handle the migration automatically while still staying aligned with the design guidelines.

bdraco:
Thinking about this a bit more, I believe we can actually handle this case on the Home Assistant side, without needing any changes in ESPHome.

Right now, if we come across a 'wrong' device with the same name, the integration logs a warning and stops: https://github.com/home-assistant/core/blob/7c488f1e54508c49a555aba59b7a62841a0e7b1f/homeassistant/components/esphome/manager.py#L457. The idea behind that is to avoid accidentally talking to the wrong device—like when someone’s using DHCP and the IP we saw last time is now pointing to a different device. If we didn’t catch that, things could get messy pretty fast.

What we could do is trigger a repair flow in that situation, asking the user if they want to migrate everything over to the new device at that IP address. That would include updating the config entry and all related entities to use the new device's unique ID. It’s definitely not a small task, but it might be a solid path forward.

bdraco:
Hit me up on Discord if you wanna talk more about it!

bdraco:
Working proof of concept for what that looks like on the Home Assistant side https://github.com/home-assistant/core/pull/142507

bdraco:
https://github.com/home-assistant/core/pull/142507 has been merged, which should address the issue in a different way and make this PR no longer necessary. Thanks for the effort though!

PR #8516, opened 2025-04-04
[sx126x] Add sx126x component

### LINKED ISSUE #2634: LORA support in the ESPHome

swoboda1337:

The SX126X component allows you to configure/control the SX1261, SX1261 and SX1268 transceivers in ESPHome. Supported frequencies range from 150 MHz to 960 MHz. Supported modulations include LoRa, FSK, GFSK, MSK and GMSK. There are many devices in the market with an ESP32 already packaged with these radios.

Datasheet: https://www.semtech.com/products/wireless-rf/lora-connect/sx1262#documentation
Docs: https://deploy-preview-4794--esphome.netlify.app/components/sx126x
SX127X PR: https://github.com/esphome/esphome/pull/7490

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/feature-requests/issues/2634

- esphome/esphome-docs#4794

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
external_components:
  - source: github://pr#8516
    components: [ sx126x ]
    refresh: 5min
```

## Checklist:


clydebarrow:
Great - I will do some testing with this.

danielkoek:
I've been testing this out locally—absolutely fantastic work! One feature that could be really useful is the ability to have nodes echo received messages to help extend range. For example, a distant node with a high-gain antenna could pick up a message, echo it back to the home node, and from there it could be passed into Home Assistant.

swoboda1337:
> I've been testing this out locally—absolutely fantastic work! One feature that could be really useful is the ability to have nodes echo received messages to help extend range. For example, a distant node with a high-gain antenna could pick up a message, echo it back to the home node, and from there it could be passed into Home Assistant.

Does this really need to be a feature though? You can already do this in yaml to echo the packet:
```
  on_packet:
    - sx126x.send_packet:
        data: !lambda 'return x;'
```

swoboda1337:
@bdraco Many thanks for reviewing the sx127x PR! 

This PR is for the sx126x chips, basically the same functionality as the sx127x except no OOK/ASK support.  The esphome configs are also very similar. The actual driver is completely different though, the registers and spi commands have all changed. 

If you have some time do you mind looking at this one too as its very similar to the sx127x?

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,514 @@
...
+  uint8_t buf[1];
+  buf[0] = mode;
+  this->write_opcode_(RADIO_SET_STANDBY, buf, 1);
+}
+
+void SX126x::wait_busy_() {
+  uint32_t start = millis();
+  while (this->busy_pin_->digital_read()) {
+    if (millis() - start > 1000) {
+      ESP_LOGE(TAG, "Wait busy timout");
```
bdraco:
```suggestion
      ESP_LOGE(TAG, "Wait busy timeout");
```

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,514 @@
...
+  this->write_opcode_(RADIO_SET_SLEEP, buf, 1);
+}
+
+void SX126x::set_mode_standby(SX126xStandbyMode mode) {
+  uint8_t buf[1];
+  buf[0] = mode;
+  this->write_opcode_(RADIO_SET_STANDBY, buf, 1);
+}
+
+void SX126x::wait_busy_() {
```
bdraco, line 449: Given how much we call wait_busy_() it might make more sense to use a state machine and check the state in ::loop

bdraco: Maybe something like (not sure this would work though)

```cpp
enum class TxState {
  IDLE,
  TRANSMITTING,
  WAITING_COMPLETE,
  ERROR
};

class SX126x : public Component, public SPIDevice<spi::BIT_ORDER_MSB_FIRST, spi::CLOCK_POLARITY_LOW,
                                                  spi::CLOCK_PHASE_LEADING, spi::DATA_RATE_1MHZ> {
  // ...
  TxState tx_state_{TxState::IDLE};
  uint32_t tx_start_time_{0};
  std::vector<uint8_t> pending_packet_;
  
  void start_transmit(const std::vector<uint8_t> &packet);
  void process_transmit();
  bool is_transmitting() const { return tx_state_ != TxState::IDLE; }
};

// In sx126x.cpp
void SX126x::start_transmit(const std::vector<uint8_t> &packet) {
  if (this->tx_state_ != TxState::IDLE) {
    ESP_LOGW(TAG, "Transmission already in progress");
    return;
  }
  
  // Validation
  if (this->payload_length_ > 0 && this->payload_length_ != packet.size()) {
    ESP_LOGE(TAG, "Packet size does not match config");
    return;
  }
  if (packet.empty() || packet.size() > this->get_max_packet_size()) {
    ESP_LOGE(TAG, "Packet size out of range");
    return;
  }
  
  // Start transmission
  this->set_mode_standby(STDBY_XOSC);
  if (this->payload_length_ == 0) {
    this->set_packet_params_(packet.size());
  }
  this->write_fifo_(0x00, packet);
  this->set_mode_tx();
  
  this->tx_state_ = TxState::TRANSMITTING;
  this->tx_start_time_ = millis();
}

void SX126x::process_transmit() {
  if (this->tx_state_ != TxState::TRANSMITTING) {
    return;
  }
  
  if (this->dio1_pin_->digital_read()) {
    // Transmission complete
    uint8_t buf[2] = {0xFF, 0xFF};
    this->write_opcode_(RADIO_CLR_IRQSTATUS, buf, 2);
    
    if (this->rx_start_) {
      this->set_mode_rx();
    } else {
      this->set_mode_sleep();
    }
    
    this->tx_state_ = TxState::IDLE;
    ESP_LOGD(TAG, "Transmission complete");
  } else if (millis() - this->tx_start_time_ > 4000) {
    // Timeout
    ESP_LOGE(TAG, "Transmit packet timeout");
    this->tx_state_ = TxState::ERROR;
    // Could add recovery logic here
  }
}

void SX126x::loop() {
  this->process_transmit();  // Process ongoing transmission
  
  // Rest of loop logic...
}
```

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,514 @@
...
+      this->read_opcode_(RADIO_GET_PACKETSTATUS, buf, 3);
+      if (this->modulation_ == PACKET_TYPE_LORA) {
+        rssi = buf[0];
+        snr = buf[1];
+      } else {
+        rssi = buf[2];
+        snr = 0;
+      }
+      this->read_opcode_(RADIO_GET_RXBUFFERSTATUS, buf, 2);
+      std::vector<uint8_t> packet(buf[0]);
```
bdraco: How large can this be? Would it make more sense to make it stack allocated?

esphome/components/sx126x/sx126x.h
```diff
@@ -0,0 +1,138 @@
...
+  SX126X_BW_15630,
+  SX126X_BW_20830,
+  SX126X_BW_31250,
+  SX126X_BW_41670,
+  SX126X_BW_62500,
+  SX126X_BW_125000,
+  SX126X_BW_250000,
+  SX126X_BW_500000,
+};
+
```
bdraco, line 48: There are lots of magic numbers in this one as well. Some suggested cleanups (although I may have mis-understood what they are used for)

```
  static constexpr uint32_t RESET_DELAY_US = 5000;
  static constexpr uint32_t POST_RESET_DELAY_US = 2000;
  static constexpr uint32_t TRANSMIT_TIMEOUT_MS = 4000;
  static constexpr uint32_t BUSY_TIMEOUT_MS = 1000; <-- although we likely want to get rid of this if possible
  
  // OCP (Over Current Protection) values
  static constexpr uint8_t OCP_60MA = 0x18;   // 60mA max current
  static constexpr uint8_t OCP_140MA = 0x38;  // 140mA max current
  
  // TCXO voltage
  static constexpr uint8_t TCXO_CTRL_3_3V = 0x07;  // 3.3V TCXO
  
  // LoRa detection optimization
  static constexpr float LORA_SYMBOL_TIMEOUT_THRESHOLD = 16.38f;
```

esphome/components/sx126x/sx126x.cpp
bdraco: It might be overkill though since than you need to think about races even more. I don't have a good feel on how long it might block the event loop for but a full second is too long.

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,514 @@
...
+    buf[4] = 0x00;
+    buf[5] = 0x00;
+    buf[6] = payload_length;
+    buf[7] = this->crc_enable_ ? 0x06 : 0x01;
+    buf[8] = 0x00;
+    this->write_opcode_(RADIO_SET_PACKETPARAMS, buf, 9);
+  }
+}
+
+void SX126x::transmit_packet(const std::vector<uint8_t> &packet) {
```
bdraco: The error handling here is logging. Should this be returning some type of enum error state?

Maybe something like
```
enum class SX126xError {
  NONE = 0,
  SPI_ERROR,
  TIMEOUT,
  INVALID_PARAMS,
  BUSY_TIMEOUT,
  TX_TIMEOUT,
  CONFIG_ERROR
};
```

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,514 @@
...
+  } else if (this->frequency_ > 770000000) {
+    buf[0] = 0xC1;
+    buf[1] = 0xC5;
+  } else if (this->frequency_ > 460000000) {
+    buf[0] = 0x75;
+    buf[1] = 0x81;
+  } else if (this->frequency_ > 425000000) {
+    buf[0] = 0x6B;
+    buf[1] = 0x6F;
+  }
```
bdraco, line 383: See comment about magic numbers on the .h

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,514 @@
...
+    buf[2] = this->coding_rate_;
+    buf[3] = (duration > 16.38f) ? 0x01 : 0x00;
+    this->write_opcode_(RADIO_SET_MODULATIONPARAMS, buf, 4);
+
+    // set packet params and sync word
+    this->set_packet_params_(this->payload_length_);
+    if (this->sync_value_.size() == 2) {
+      for (uint32_t i = 0; i < this->sync_value_.size(); i++) {
+        uint8_t data = this->sync_value_[i];
+        this->write_register_(REG_LORA_SYNCWORD + i, &data, 1);
```
bdraco: Can this be done in a single write or does it need to be multiple?

bdraco:
Added some comments. Since I don't have the hardware you'll have to tell me if they make sense or not for the use case. If any of them apply to sx127x please address them first in a followup as we can merge that rather quickly as the review size will be much smaller now that the base is merged.

esphome/components/sx126x/sx126x.cpp
swoboda1337: This won't block for long, calibration is probably the longest under 10 msecs, most commands have no wait time. You have to wait after every write, would be a huge mess to unroll everything.

esphome/components/sx126x/sx126x.cpp
swoboda1337: It gets passed to automations and callbacks. Its passed by reference `const std::vector<uint8_t> &packet`

bdraco: As long as they don't use it for longer than the callback duration is should be fine

esphome/components/sx126x/sx126x.cpp
swoboda1337: These are magic numbers in the manual too. No explanation as to what they mean. I agree about the others in the .h  comment, will address those. For these I think I will just add a comment with a reference to the doc?

bdraco: That's sounds fine. A comment with a link to the manual would be ideal. Even if it goes stale it will make it easier to figure out later

swoboda1337:
The only change to sx127x would be adding a return to transmit_packet. I can create a PR for that if this looks good.

esphome/components/sx126x/sx126x.cpp
swoboda1337: Added a comment here to explain and reduced the timeout.

bdraco:
> The only change to sx127x would be adding a return to transmit_packet. I can create a PR for that if this looks good.

Error handling LGTM. Please do the PR to add it to sx127x

swoboda1337:
> Error handling LGTM. Please do the PR to add it to sx127x

Done in https://github.com/esphome/esphome/pull/9351

esphome/components/sx126x/sx126x.h
```diff
@@ -0,0 +1,140 @@
...
+  uint8_t wakeup_();
+  Trigger<std::vector<uint8_t>, float, float> *packet_trigger_{new Trigger<std::vector<uint8_t>, float, float>()};
+  std::vector<SX126xListener *> listeners_;
+  std::vector<uint8_t> sync_value_;
+  InternalGPIOPin *busy_pin_{nullptr};
+  InternalGPIOPin *dio1_pin_{nullptr};
+  InternalGPIOPin *rst_pin_{nullptr};
+  std::string hw_version_;
+  SX126xBw bandwidth_;
+  char version_[16];
```
bdraco, line 118: Bit safer if there is any risk of undefined behavior from unit values
```diff
diff --git a/esphome/components/sx126x/sx126x.h b/esphome/components/sx126x/sx126x.h
index 1234567..abcdefg 100644
--- a/esphome/components/sx126x/sx126x.h
+++ b/esphome/components/sx126x/sx126x.h
@@ -115,23 +115,23 @@ class SX126x : public Component, public SPIDevice<SPIBitOrder::MSB_FIRST, SPICl
   InternalGPIOPin *dio1_pin_{nullptr};
   InternalGPIOPin *rst_pin_{nullptr};
   std::string hw_version_;
-  SX126xBw bandwidth_;
-  char version_[16];
-  uint32_t bitrate_;
-  uint32_t deviation_;
-  uint32_t frequency_;
-  uint32_t payload_length_;
-  uint32_t tcxo_delay_;
-  uint16_t preamble_detect_;
-  uint16_t preamble_size_;
-  uint8_t tcxo_voltage_;
-  uint8_t coding_rate_;
-  uint8_t modulation_;
-  uint8_t pa_ramp_;
-  uint8_t shaping_;
-  uint8_t spreading_factor_;
-  int8_t pa_power_;
-  bool crc_enable_;
-  bool rx_start_;
-  bool rf_switch_;
+  SX126xBw bandwidth_{SX126X_BW_125000};
+  char version_[16]{0};
+  uint32_t bitrate_{0};
+  uint32_t deviation_{0};
+  uint32_t frequency_{0};
+  uint32_t payload_length_{0};
+  uint32_t tcxo_delay_{0};
+  uint16_t preamble_detect_{0};
+  uint16_t preamble_size_{0};
+  uint8_t tcxo_voltage_{0};
+  uint8_t coding_rate_{0};
+  uint8_t modulation_{PKT_TYPE_LORA};
+  uint8_t pa_ramp_{0};
+  uint8_t shaping_{0};
+  uint8_t spreading_factor_{0};
+  int8_t pa_power_{0};
+  bool crc_enable_{false};
+  bool rx_start_{false};
+  bool rf_switch_{false};
 };
```

esphome/components/sx126x/sx126x.cpp
bdraco: Still a bit concerned about the frequent vector allocations here.

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,534 @@
...
+}
+
+void SX126x::wait_busy_() {
+  // wait if the device is busy, the maximum delay is only be a few ms
+  // with most commands taking only a few us
+  uint32_t start = millis();
+  while (this->busy_pin_->digital_read()) {
+    if (millis() - start > BUSY_TIMEOUT_MS) {
+      ESP_LOGE(TAG, "Wait busy timeout");
+      this->mark_failed();
```
bdraco, line 456: Thought about this some more. We mark it failed but execution continues outside of this helper

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+// OCP (Over Current Protection) values
+static constexpr uint8_t OCP_80MA = 0x18;   // 80 mA max current
+static constexpr uint8_t OCP_140MA = 0x38;  // 140 mA max current
+
+// LoRa low data rate optimization threshold
+static constexpr float LOW_DATA_RATE_OPTIMIZE_THRESHOLD = 16.38f;  // 16.38 ms
+
+uint8_t SX126x::wakeup_() {
+  this->wait_busy_();
```
bdraco: Could make this a bit more DRY
```diff
diff --git a/esphome/components/sx126x/sx126x.h b/esphome/components/sx126x/sx126x.h
index 1234567..abcdefg 100644
--- a/esphome/components/sx126x/sx126x.h
+++ b/esphome/components/sx126x/sx126x.h
@@ -95,6 +95,24 @@ class SX126x : public Component, public SPIDevice<SPIBitOrder::MSB_FIRST, SPICl
   void dump_config() override;
 
  protected:
+  // RAII class for SPI transactions
+  class SPITransaction {
+   public:
+    explicit SPITransaction(SX126x *parent) : parent_(parent) {
+      parent_->wait_busy_();
+      parent_->enable();
+    }
+    ~SPITransaction() {
+      parent_->disable();
+      if (add_delay_) {
+        delayMicroseconds(SWITCHING_DELAY_US);
+      }
+    }
+    void set_delay() { add_delay_ = true; }
+   private:
+    SX126x *parent_;
+    bool add_delay_{false};
+  };
+
   uint8_t read_fifo_(uint8_t offset, std::vector<uint8_t> &packet);
   void write_fifo_(uint8_t offset, const std::vector<uint8_t> &packet);
   void transmit_packet_(const std::vector<uint8_t> &packet);
diff --git a/esphome/components/sx126x/sx126x.cpp b/esphome/components/sx126x/sx126x.cpp
index 1234567..abcdefg 100644
--- a/esphome/components/sx126x/sx126x.cpp
+++ b/esphome/components/sx126x/sx126x.cpp
@@ -43,69 +43,60 @@ uint8_t SX126x::write_command_(uint8_t cmd, uint8_t *data, uint8_t size) {
 }
 
 uint8_t SX126x::read_fifo_(uint8_t offset, std::vector<uint8_t> &packet) {
-  this->wait_busy_();
-  this->enable();
+  SPITransaction transaction(this);
   this->transfer_byte(RADIO_READ_BUFFER);
   this->transfer_byte(offset);
   uint8_t status = this->transfer_byte(0x00);
   for (uint8_t &byte : packet) {
     byte = this->transfer_byte(0x00);
   }
-  this->disable();
   return status;
 }
 
 void SX126x::write_fifo_(uint8_t offset, const std::vector<uint8_t> &packet) {
-  this->wait_busy_();
-  this->enable();
+  SPITransaction transaction(this);
+  transaction.set_delay();
   this->transfer_byte(RADIO_WRITE_BUFFER);
   this->transfer_byte(offset);
   for (const uint8_t &byte : packet) {
     this->transfer_byte(byte);
   }
-  this->disable();
-  delayMicroseconds(SWITCHING_DELAY_US);
 }
 
 uint8_t SX126x::read_opcode_(uint8_t opcode, uint8_t *data, uint8_t size) {
-  this->wait_busy_();
-  this->enable();
+  SPITransaction transaction(this);
   this->transfer_byte(opcode);
   uint8_t status = this->transfer_byte(0x00);
   for (int32_t i = 0; i < size; i++) {
     data[i] = this->transfer_byte(0x00);
   }
-  this->disable();
   return status;
 }
 
 void SX126x::write_opcode_(uint8_t opcode, uint8_t *data, uint8_t size) {
-  this->wait_busy_();
-  this->enable();
+  SPITransaction transaction(this);
+  transaction.set_delay();
   this->transfer_byte(opcode);
   for (int32_t i = 0; i < size; i++) {
     this->transfer_byte(data[i]);
   }
-  this->disable();
-  delayMicroseconds(SWITCHING_DELAY_US);
 }
 
 void SX126x::read_register_(uint16_t reg, uint8_t *data, uint8_t size) {
-  this->wait_busy_();
-  this->enable();
+  SPITransaction transaction(this);
   this->write_byte(RADIO_READ_REGISTER);
   this->write_byte((reg >> 8) & 0xFF);
   this->write_byte((reg >> 0) & 0xFF);
   this->write_byte(0x00);
   for (int32_t i = 0; i < size; i++) {
     data[i] = this->transfer_byte(0x00);
   }
-  this->disable();
 }
 
 void SX126x::write_register_(uint16_t reg, uint8_t *data, uint8_t size) {
-  this->wait_busy_();
-  this->enable();
+  SPITransaction transaction(this);
+  transaction.set_delay();
   this->write_byte(RADIO_WRITE_REGISTER);
   this->write_byte((reg >> 8) & 0xFF);
   this->write_byte((reg >> 0) & 0xFF);
@@ -113,8 +104,6 @@ void SX126x::write_register_(uint16_t reg, uint8_t *data, uint8_t size) {
   for (int32_t i = 0; i < size; i++) {
     this->write_byte(data[i]);
   }
-  this->disable();
-  delayMicroseconds(SWITCHING_DELAY_US);
 }
 
 void SX126x::send_packet(const std::vector<uint8_t> &packet) {
@@ -427,15 +416,12 @@ void SX126x::set_mode_tx() {
 }
 
 uint8_t SX126x::wakeup_() {
-  this->wait_busy_();
-  this->enable();
+  SPITransaction transaction(this);
   this->transfer_byte(RADIO_NOP);
   uint8_t status = this->transfer_byte(0x00);
-  this->disable();
-  delayMicroseconds(SWITCHING_DELAY_US);
-  // TODO wait busy
+  transaction.set_delay();
+  // Note: destructor will add delay before checking busy again
   return status;
 }
 
 void SX126x::wait_busy_() {
```

esphome/components/sx126x/sx126x.cpp
```diff
@@ -0,0 +1,534 @@
...
+      this->read_opcode_(RADIO_GET_PACKETSTATUS, buf, 3);
+      if (this->modulation_ == PACKET_TYPE_LORA) {
+        rssi = buf[0];
+        snr = buf[1];
+      } else {
+        rssi = buf[2];
+        snr = 0;
+      }
+      this->read_opcode_(RADIO_GET_RXBUFFERSTATUS, buf, 2);
+      std::vector<uint8_t> packet(buf[0]);
```
bdraco: Is there a risk that the device could return a corrupted size and we try to allocate a massive amount of memory here? Should we have some type of validation on the size?

swoboda1337: We read the size as a byte and the max size is 255, so cant be more than 255. I just noticed that get_max_packet_size() returns 256 though will fix that.

esphome/components/sx126x/sx126x.cpp
swoboda1337: Yeah but the following reads and writes will fail too. The loop will stop. It would get messy to put ifs and returns everywhere. Important thing is that error is flagged and the user will see it.

esphome/components/sx126x/sx126x.cpp
swoboda1337: This doesnt save many loc and isnt it harder to read?

swoboda1337: I realized I could just remove wakeup_ altogether which saved more lines.

esphome/components/sx126x/sx126x.cpp
swoboda1337: The packet transport stuff uses SX126xListener and takes a vector. It is const and pass by reference so only one allocation. Wouldn't save anything by creating it later. 

The other listeners like remote receiver pass vectors. It is const and passed by reference to all listeners. 

The trigger takes a vector already. 

Could pre allocate it as a member variable

esphome/components/sx126x/sx126x.cpp
bdraco: The thought here was that RAII makes it harder to make refactoring mistakes in the future.

esphome/components/sx126x/sx126x.cpp
bdraco: Pre-allocate a vector as a member variable and reuse it seems like a good solution

esphome/components/sx126x/sx126x.cpp
swoboda1337: I get the point of using the RAII pattern in more complicated situations. If there were early returns or longer functions etc it would be cleaner. These functions are linear and only a few lines though. I think this makes it harder to read.  

If you still think it makes sense I can change it.

bdraco: Totally up to you. I just like using RAII when there’s a chance someone might tweak the code later and accidentally miss a lock or unlock. I've just seen it happen a bunch over the years, so it's kind of a habit now, more of a safety net than anything.

swoboda1337:
Created https://github.com/esphome/esphome/pull/9357 to pre-allocate the packet in sx127x plus fix dump_config issue someone found.

esphome/components/sx126x/sx126x.cpp
swoboda1337: Lets keep it the way it is then, easier to read and understand. If things change and these functions get more complex can change it.

esphome/components/sx126x/__init__.py
```diff
@@ -0,0 +1,317 @@
...
+    if config[CONF_MODULATION] == "LORA":
+        if config[CONF_BANDWIDTH] not in lora_bws:
+            raise cv.Invalid(f"{config[CONF_BANDWIDTH]} is not available with LORA")
+        if config[CONF_PREAMBLE_SIZE] > 0 and config[CONF_PREAMBLE_SIZE] < 6:
+            raise cv.Invalid("Minimum preamble size is 6 with LORA")
+        if config[CONF_SPREADING_FACTOR] == 6 and config[CONF_PAYLOAD_LENGTH] == 0:
+            raise cv.Invalid("Payload length must be set when spreading factor is 6")
+    else:
+        if config[CONF_BANDWIDTH] in lora_bws:
+            raise cv.Invalid("{config[CONF_BANDWIDTH]} is not available with FSK")
```
bdraco:
```suggestion
            raise cv.Invalid(f"{config[CONF_BANDWIDTH]} is not available with FSK")
```

bdraco (approved):
Thanks @swoboda1337

swoboda1337:
> Thanks @swoboda1337

Thanks for taking the time to review these.

PR #8533, opened 2025-04-08
Make udp component able to send and listen to IPv6 multicast

HeMan:

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4891

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/config_validation.py
```diff
@@ -1185,6 +1183,16 @@ def ipaddress(value):
...
 
 
+def ipaddress_multi_broadcast(value):
+    try:
+        address = ip_address(value)
+    except ValueError as exc:
+        raise Invalid(f"{value} is not a valid IP address") from exc
+    if not (address.is_multicast or (address == IPv4Address("255.255.255.255"))):
+        raise Invalid(f"{value} is not a multicast address nor local broadcast address")
+    return address
```
clydebarrow:
```suggestion
    address = ip_address(value)
    if address.is_multicast or address == IPv4Address("255.255.255.255"):
        return address
    raise Invalid(f"{value} is neither a multicast address nor the local broadcast address")
```
`ip_address` already raises the appropriate exception.

esphome/config_validation.py
```diff
@@ -1171,9 +1171,7 @@ def ipv4address(value):
 def ipv4address_multi_broadcast(value):
     address = ipv4address(value)
```
clydebarrow:
```suggestion
    return ipaddress_multi_broadcast(ipv4address(value))
```
Other than the `ipv4address` call,the code is just repeated in the other func.

clydebarrow (changes requested):
Will do some testing shortly.

esphome/components/udp/__init__.py
```diff
@@ -38,6 +41,21 @@
 )
 
 
+def _final_validate(config):
+    enable_ipv6 = fv.full_config.get().get("network").get("enable_ipv6")
```
jesserockz: Please use consts here. `CONF_NETWORK` does not yet exist
```suggestion
    enable_ipv6 = fv.full_config.get().get(CONF_NETWORK).get(CONF_ENABLE_IPV6)
```

esphome/components/udp/__init__.py
```diff
@@ -38,6 +41,21 @@
...
+        cv.ipv4address_multi_broadcast(config[CONF_LISTEN_ADDRESS])
+    return config
+
+
+def require_internal_with_name(config):
+    if CONF_NAME in config and CONF_INTERNAL not in config:
+        raise cv.Invalid("Must provide internal: config when using name:")
+    return config
+
+
```
jesserockz, line 121: Has no purpose?
```suggestion
```

jesserockz (changes requested):

esphome/components/udp/udp_component.cpp
```diff
@@ -13,8 +13,17 @@ static const char *const TAG = "udp";
 void UDPComponent::setup() {
 #if defined(USE_SOCKET_IMPL_BSD_SOCKETS) || defined(USE_SOCKET_IMPL_LWIP_SOCKETS)
   for (const auto &address : this->addresses_) {
+#if USE_NETWORK_IPV6
```
asergunov, line 16: Can't we support both ip4 and ip6 addresses same time? Looks like disabling ipv4 just because IPv6 enabled is not expected

HeMan: The `sockaddr_in6` accepts both IPv4 and IPv6 traffic.

esphome/components/udp/__init__.py
```diff
@@ -10,9 +10,17 @@
     CONF_SENSORS,
 )
 import esphome.config_validation as cv
-from esphome.const import CONF_DATA, CONF_ID, CONF_PORT, CONF_TRIGGER_ID
+from esphome.const import (
+    CONF_DATA,
+    CONF_ENABLE_IPV6,
```
dwmw2, line 16: Why have an explicit `enable_ipv6` option? You can tell if the address given is IPv6 or Legacy IP. When fixing the external syslog_component I did it like this: https://github.com/TheStaticTurtle/esphome_syslog/pull/17/files

dwmw2:
Does this also add unicast support? Now that ESPHome includes a syslog component which is based on the udp component, it's a regression from the external one which I was previously using that supported IPv6.

esphome/components/udp/__init__.py
HeMan: I use it in `_final_validate` to only allow IPv4 addresses when IPv6 is not enabled.

dwmw2: But why have a separate configuration for it at all? Just let the user set an IPv6 or Legacy IP address as they desire.

We already have `#if USE_NETWORK_IPV6` and we should only allow IPv6 addresses if *that* is set, I suppose. Ideally we'd also have a corresponding `#if USE_NETWORK_IPV4` which gates Legacy IP. Perhaps we should start doing that now, even if it's *always* set because we haven't fully implemented IPv6-only yet.

(While we're thinking about this... for things like syslog we should also allow names, and do DNS lookups just like we do for MQTT)

esphome/components/udp/udp_component.cpp
dwmw2: Does this work even if we don't have an IPv6 address? What if IPv6 is *enabled* in the config but there's no RA/DHCPv6 and we only pick up a Legacy IP address from DHCP?

esphome/components/udp/__init__.py
HeMan: It's not a separate configuration. It will give a decent error message if you set an IPv6 address but haven't enabled IPv6. The user could set legacy IP address regardless, but IPv6 only if it is enabled in the config.

dwmw2: Ah, I see. That also explains why I couldn't see mention of the `enable_ipv6` option in the corresponding docs PR.

That doesn't seem like something we should have to do for ourselves in the UDP component, surely? There should be a generic validator for IP addresses which allows Legacy IP and/or IPv6 as appropriate for the config without everyone having to do this for themselves?

HeMan:
> Does this also add unicast support? Now that ESPHome includes a syslog component which is based on the udp component, it's a regression from the external one which I was previously using that supported IPv6.

Yes

DJJo14:
Does this udp ipv6 also work for [openthread](https://esphome.io/components/openthread/)? specially with [package transport](https://esphome.io/components/sensor/packet_transport/)?

dwmw2:
Any progress on this? I'd like to fix the syslog again; it regressed into the 1990s when it was merged. The external component used to support IPv6.

esphome/components/socket/socket.cpp
```diff
@@ -50,7 +50,7 @@ std::unique_ptr<Socket> socket_ip_loop_monitored(int type, int protocol) {
 }
 
 socklen_t set_sockaddr(struct sockaddr *addr, socklen_t addrlen, const std::string &ip_address, uint16_t port) {
-#if USE_NETWORK_IPV6
+#if USE_NETWORK_IPV6 && !USE_HOST
```
dwmw2, line 53: Hm, that makes it build but won't actually work. This should https://github.com/esphome/esphome/pull/11996

esphome/components/udp/__init__.py
dwmw2: This works without an extra option: https://github.com/esphome/esphome/pull/11889/commits/7a6054b89f2f0d358fcfdc0b5f29ed4b8df4b31d

HeMan:
@dwmw2 I bumped into a problem with libretiny-devices. I think the Arduino code needs to be upgraded.

dwmw2:
What's the LibreTiny issue? I already upgraded the LWIP there last year, I think.

PR #8536, opened 2025-04-09
Sort resolved IP addresses for dashboard

dwmw2:
In https://github.com/esphome/esphome/pull/7414 I fixed the name resolution in helpers.resolve_ip_address() to return a *list* of addresses sorted in an approximation of RFC6724 preference.

I missed the fact that the dashboard code calls directly into the underlying zeroconf resolve_host() functions. As those return a list, we get a TypeError("expected string or bytes-like object, got 'list'")>) raised when trying to use the result of the lookup.

In https://github.com/esphome/esphome/pull/7747 the traceback was fixed by just using the *first* element of that list, but the list still wasn't *sorted*, so we're essentially using a *randomly* selected address which happened to be the first one to respond to mDNS.

Fix it properly by providing a helpers.sort_ip_addresses() function which performs the same sort as we do for the OTA connection.

This involves calling gethostinfo() so we can see the address type and scope for each address, then applying the same ordering criteria, then getnameinfo() to turn it back into a string again for the output.

(In the original resolve_ip_address() we *wanted* the getaddrinfo() tuples, so we didn't perform the final getnameinfo() and it all seemed a little more natural).

## Types of changes

- [ ] Other

## Test Environment

Only unit tested locally with the following hack:
```
--- a/esphome/__main__.py
+++ b/esphome/__main__.py
@@ -42,7 +42,7 @@ from esphome.const import (
     SECRETS_FILES,
 )
 from esphome.core import CORE, EsphomeError, coroutine
-from esphome.helpers import get_bool_env, indent, is_ip_address
+from esphome.helpers import get_bool_env, indent, is_ip_address, sort_ip_addresses
 from esphome.log import Fore, color, setup_log
 from esphome.util import (
     get_serial_ports,
@@ -374,6 +374,12 @@ def upload_program(config, args, host):
     remote_port = int(ota_conf[CONF_PORT])
     password = ota_conf.get(CONF_PASSWORD, "")
 
+
+    addrlist = [ '10.0.0.1', '2001::6', 'fe80::2', 'fe80::1%lo' ]
+    slist = sort_ip_addresses(addrlist)
+    for l in slist:
+        _LOGGER.info(l);
+
     if (
         CONF_MQTT in config  # pylint: disable=too-many-boolean-expressions
         and (not args.device or args.device in ("MQTT", "OTA"))

```
```
INFO 2001::6
INFO fe80::1%lo
INFO 10.0.0.1
INFO fe80::2
```

### DIFF
--- a/esphome/helpers.py
+++ b/esphome/helpers.py
@@ -200,12 +200,20 @@ def resolve_ip_address(host, port):
     return res
 
 
-def sort_ip_addresses(address_list):
+def sort_ip_addresses(address_list: list[str]) -> list[str]:
+    """Takes a list of IP addresses in string form, e.g. from mDNS or MQTT,
+    and sorts them into the best order to actually try connecting to them.
+
+    This is roughly based on RFC6724 but a lot simpler: First we choose
+    IPv6 addresses, then Legacy IP addresses, and lowest priority is
+    link-local IPv6 addresses that don't have a link specified (which
+    are useless, but mDNS does provide them in that form.)
+    """
     import socket
 
     # First "resolve" all the IP addresses to get getaddrinfo() tuples
     # which indicate the type of address, scope, etc.
-    res = []
+    res: list[str] = []
     for addr in address_list:
         # This should always work as these are supposed to be IP addresses
         try:
@@ -220,7 +228,7 @@ def sort_ip_addresses(address_list):
     res.sort(key=addr_preference_)
 
     # Finally, turn the getaddrinfo() tuples back into plain hostnames.
-    sorted_address_list = []
+    sorted_address_list: list[str] = []
     for r in res:
         sorted_address_list.append(socket.getnameinfo(r[4], socket.NI_NUMERICHOST)[0])
 
--- a/esphome/helpers.py
+++ b/esphome/helpers.py
@@ -226,10 +226,9 @@ def sort_ip_addresses(address_list: list[str]) -> list[str]:
     for addr in address_list:
         # This should always work as these are supposed to be IP addresses
         try:
-            r = socket.getaddrinfo(
+            res += socket.getaddrinfo(
                 addr, 0, proto=socket.IPPROTO_TCP, flags=socket.AI_NUMERICHOST
             )
-            res += r
         except OSError:
             _LOGGER.info("Failed to parse IP address '%s'", addr)
 
--- a/esphome/helpers.py
+++ b/esphome/helpers.py
@@ -207,7 +207,8 @@ def sort_ip_addresses(address_list: list[str]) -> list[str]:
     This is roughly based on RFC6724 but a lot simpler: First we choose
     IPv6 addresses, then Legacy IP addresses, and lowest priority is
     link-local IPv6 addresses that don't have a link specified (which
-    are useless, but mDNS does provide them in that form.)
+    are useless, but mDNS does provide them in that form). Addresses
+    which cannot be parsed are silently dropped.
     """
     import socket
 
--- a/esphome/helpers.py
+++ b/esphome/helpers.py
@@ -200,6 +200,33 @@ def resolve_ip_address(host, port):
     return res
 
 
+def sort_ip_addresses(address_list):
+    import socket
+
+    # First "resolve" all the IP addresses to get getaddrinfo() tuples
+    # which indicate the type of address, scope, etc.
+    res = []
+    for addr in address_list:
+        # This should always work as these are supposed to be IP addresses
+        try:
+            r = socket.getaddrinfo(
+                addr, 0, proto=socket.IPPROTO_TCP, flags=socket.AI_NUMERICHOST
+            )
+            res = res + r
+        except OSError:
+            _LOGGER.info("Failed to parse IP address '%s'", addr)
+
+    # Now use that information to sort them.
+    res.sort(key=addr_preference_)
+
+    # Finally, turn the getaddrinfo() tuples back into plain hostnames.
+    sorted_address_list = []
+    for r in res:
+        sorted_address_list.append(socket.getnameinfo(r[4], socket.NI_NUMERICHOST)[0])
+
+    return sorted_address_list
+
+
 def get_bool_env(var, default=False):
     value = os.getenv(var, default)
     if isinstance(value, str):
--- a/esphome/helpers.py
+++ b/esphome/helpers.py
@@ -229,7 +229,7 @@ def sort_ip_addresses(address_list: list[str]) -> list[str]:
             r = socket.getaddrinfo(
                 addr, 0, proto=socket.IPPROTO_TCP, flags=socket.AI_NUMERICHOST
             )
-            res = res + r
+            res += r
         except OSError:
             _LOGGER.info("Failed to parse IP address '%s'", addr)
 
--- a/esphome/helpers.py
+++ b/esphome/helpers.py
@@ -228,11 +228,7 @@ def sort_ip_addresses(address_list: list[str]) -> list[str]:
     res.sort(key=addr_preference_)
 
     # Finally, turn the getaddrinfo() tuples back into plain hostnames.
-    sorted_address_list: list[str] = []
-    for r in res:
-        sorted_address_list.append(socket.getnameinfo(r[4], socket.NI_NUMERICHOST)[0])
-
-    return sorted_address_list
+    return [socket.getnameinfo(r[4], socket.NI_NUMERICHOST)[0] for r in res]
 
 
 def get_bool_env(var, default=False):
--- a/esphome/helpers.py
+++ b/esphome/helpers.py
@@ -211,9 +211,17 @@ def sort_ip_addresses(address_list: list[str]) -> list[str]:
     """
     import socket
 
-    # First "resolve" all the IP addresses to get getaddrinfo() tuples
-    # which indicate the type of address, scope, etc.
-    res: list[str] = []
+    # First "resolve" all the IP addresses to getaddrinfo() tuples of the form
+    # (family, type, proto, canonname, sockaddr)
+    res: list[
+        tuple[
+            int,
+            int,
+            int,
+            Union[str, None],
+            Union[tuple[str, int], tuple[str, int, int, int]],
+        ]
+    ] = []
     for addr in address_list:
         # This should always work as these are supposed to be IP addresses
         try:
--- a/tests/unit_tests/test_helpers.py
+++ b/tests/unit_tests/test_helpers.py
@@ -267,3 +267,13 @@ def test_sanitize(text, expected):
     actual = helpers.sanitize(text)
 
     assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "text, expected",
+    ((["127.0.0.1", "fe80::1", "2001::2"], ["2001::2", "127.0.0.1", "fe80::1"]),),
+)
+def test_sort_ip_addresses(text, expected):
+    actual = helpers.sort_ip_addresses(text)
+
+    assert actual == expected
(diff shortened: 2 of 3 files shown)


dwmw2:
Worked out how to run dashboard locally and test... with an added print of the sorted results I can see it's doing the right thing.
```
2025-04-09 14:58:00,420 INFO 101 GET /run (127.0.0.1) 0.52ms
['2001:8b0:10b:5:f2f5:bdff:fe4f:556c', '172.31.31.168', 'fe80::f2f5:bdff:fe4f:556c']
2025-04-09 14:58:00,836 INFO Running command 'esphome --dashboard run ~/esphome/landing-blind-1.yaml --device 2001:8b0:10b:5:f2f5:bdff:fe4f:556c'
```

esphome/helpers.py
```diff
@@ -200,6 +200,33 @@ def resolve_ip_address(host, port):
     return res
 
 
+def sort_ip_addresses(address_list):
```
bdraco: Please add typing and a test

esphome/helpers.py
```diff
@@ -200,6 +200,33 @@ def resolve_ip_address(host, port):
...
+
+    # Now use that information to sort them.
+    res.sort(key=addr_preference_)
+
+    # Finally, turn the getaddrinfo() tuples back into plain hostnames.
+    sorted_address_list = []
+    for r in res:
+        sorted_address_list.append(socket.getnameinfo(r[4], socket.NI_NUMERICHOST)[0])
+
+    return sorted_address_list
```
bdraco: This could return a list comp

esphome/helpers.py
```diff
@@ -200,6 +200,33 @@ def resolve_ip_address(host, port):
     return res
 
 
+def sort_ip_addresses(address_list):
+    import socket
+
+    # First "resolve" all the IP addresses to get getaddrinfo() tuples
+    # which indicate the type of address, scope, etc.
+    res = []
```
bdraco: Please type this list

bdraco:
Please explain how they are sorted and why they are sorted in that order in the opening text.

esphome/helpers.py
dwmw2: Unresolving that one again as I made it
```
 res: list[str] = []
```
.... but it *isn't* a list of strings. But the tests worked and everything; is this typing in python purely cosmetic?

It's actually the same form of list that [socket.getaddrinfo](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) returns. I'll see if I can work out how to express that...

bdraco: Typing is not enforced at runtime.  

You could steal the type from aiohappyeyeballs

dwmw2:
```
    # First "resolve" all the IP addresses to getaddrinfo() tuples of the form
    # (family, type, proto, canonname, sockaddr)
    res: list[
        tuple[
            int,
            int,
            int,
            Union[str, None],
            Union[tuple[str, int], tuple[str, int, int, int]],
        ]
    ] = []
```
Is there a way to actually *test* this?

bdraco: Run mypy manually

esphome/helpers.py
```diff
@@ -200,6 +200,46 @@ def resolve_ip_address(host, port):
...
 
 
+def sort_ip_addresses(address_list: list[str]) -> list[str]:
+    """Takes a list of IP addresses in string form, e.g. from mDNS or MQTT,
+    and sorts them into the best order to actually try connecting to them.
+
+    This is roughly based on RFC6724 but a lot simpler: First we choose
+    IPv6 addresses, then Legacy IP addresses, and lowest priority is
+    link-local IPv6 addresses that don't have a link specified (which
+    are useless, but mDNS does provide them in that form). Addresses
```
bdraco, line 210: We should be getting them in `ip%scope` from mDNS. Are we losing the scope somehow in conversions?

dwmw2: I've seen that happen, yes. I believe we also get scopeless LL addresses from MQTT.

Either way, if it does happen, those addresses go last in the list (perhaps we could even drop them, but a clear failure to connect to a LL address is a better error signal than apparently finding no addresses).

dwmw2: My recollection is that the zeroconf function which was *supposed* to return an address with scope , did not. Maybe it uses inet_ntop() instead of getnameinfo() as I do here? I don't believe I dug into that one but it's a separate problem.

bdraco: That sounds like a bug that we should fix in `zeroconf`.  I'd be curious to know more as I'm one of the `zeroconf` maintainers.

dwmw2: Here's an example, with a `print(addr_list)` at about line 177 of esphome/helpers.py:

```
INFO Resolving IP address of landing-blind-1.local in mDNS
['172.31.31.173', '2001:8b0:10b:5:f2f5:bdff:fe4f:556c', 'fe80::f2f5:bdff:fe4f:556c']
INFO Connecting to 2001:8b0:10b:5:f2f5:bdff:fe4f:556c port 3232...
```

I believe that list comes fairly much directly from the call to [zeroconf.parsed_scoped_addresses()](https://python-zeroconf.readthedocs.io/en/latest/api.html#zeroconf.ServiceInfo.parsed_scoped_addresses) at about line 173 of esphome/zeroconf.py, which was *supposed* to have returned a properly scoped address, but didn't.

I think that's about as far as I got last time before concluding it was "Someone Else's Problem" at least for the time being... Hello, "Someone Else" :)

Should we take this discussion to a bug report elsewhere, or Discord?

```
~/git/esphome$ pip3 install zeroconf
Requirement already satisfied: zeroconf in ./venv/lib/python3.12/site-packages (0.146.3)
Requirement already satisfied: ifaddr>=0.1.7 in ./venv/lib/python3.12/site-packages (from zeroconf) (0.2.0)
```

dwmw2:
```
~/git/esphome$ python3
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import zeroconf
>>> zc = zeroconf.Zeroconf()
>>> info = zeroconf.AddressResolver("landing-blind-1.local.")
>>> info.request(zc, 10000)
True
>>> print(info.parsed_scoped_addresses())
['172.31.31.173', '2001:8b0:10b:5:f2f5:bdff:fe4f:556c', 'fe80::f2f5:bdff:fe4f:556c']

```

bdraco: > Should we take this discussion to a bug report elsewhere, or Discord?

An issue report for zeroconf would be great.

I don't have an IPv6 stack ESPHome device to replicate this on. I'll have to set one up later this week.

dwmw2: I can give you an account on a machine on this network if it helps. Send me a SSH public key and preferred username.

bdraco: Looks like I can reproduce it with an Apple TV

```
>>> nfo = zeroconf.AddressResolver("Living-Room.local.")
>>> nfo.request(zc, 10000)
True
>>> nfo.ip_addresses_by_version(IPVersion.All)
[ZeroconfIPv4Address('192.168.106.157'), ZeroconfIPv6Address('fdde:ad00:beef:ca:2:fd49:7976:9df'), ZeroconfIPv6Address('fdde:ad00:beef:ca:1002:e71c:7ba8:f92d'), ZeroconfIPv6Address('fe80::460:6848:f128:356e')]
>>> 
```

dwmw2: https://github.com/python-zeroconf/python-zeroconf/issues/1567

bdraco: Looks like the response comes in over IPv4 first so we don't associate a scope to it.  Probably not an easy problem to solve.

tests/unit_tests/test_helpers.py
```diff
@@ -267,3 +267,13 @@ def test_sanitize(text, expected):
     actual = helpers.sanitize(text)
 
     assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "text, expected",
+    ((["127.0.0.1", "fe80::1", "2001::2"], ["2001::2", "127.0.0.1", "fe80::1"]),),
+)
+def test_sort_ip_addresses(text, expected):
```
bdraco:
```suggestion
def test_sort_ip_addresses(text: list[str], expected: list[str]) -> None:
```

esphome/helpers.py
dwmw2: It's also *always* going to happen with MQTT anyway, which is why I didn't lose much sleep over it when I first saw it.

tests/unit_tests/test_helpers.py
```diff
@@ -267,3 +267,13 @@ def test_sanitize(text, expected):
     actual = helpers.sanitize(text)
 
     assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "text, expected",
+    ((["127.0.0.1", "fe80::1", "2001::2"], ["2001::2", "127.0.0.1", "fe80::1"]),),
```
bdraco, line 274: Let's add a scoped address to this test so whenever zeroconf does start returning them, we know it works

dwmw2: I think that'll break on diverse test setups. Even scoped to %lo might not work in some environments.

esphome/helpers.py
```diff
@@ -200,6 +200,46 @@ def resolve_ip_address(host, port):
...
+            Union[tuple[str, int], tuple[str, int, int, int]],
+        ]
+    ] = []
+    for addr in address_list:
+        # This should always work as these are supposed to be IP addresses
+        try:
+            r = socket.getaddrinfo(
+                addr, 0, proto=socket.IPPROTO_TCP, flags=socket.AI_NUMERICHOST
+            )
+            res = res + r
```
bdraco, line 232:
```suggestion
            res += r
```

esphome/helpers.py
```diff
@@ -200,6 +200,46 @@ def resolve_ip_address(host, port):
...
+            Union[tuple[str, int], tuple[str, int, int, int]],
+        ]
+    ] = []
+    for addr in address_list:
+        # This should always work as these are supposed to be IP addresses
+        try:
+            r = socket.getaddrinfo(
+                addr, 0, proto=socket.IPPROTO_TCP, flags=socket.AI_NUMERICHOST
+            )
+            res += r
```
bdraco:
```suggestion
            res += socket.getaddrinfo(
                addr, 0, proto=socket.IPPROTO_TCP, flags=socket.AI_NUMERICHOST
            )
```

actually even simpler

tests/unit_tests/test_helpers.py
bdraco: Tested it manually

bdraco (approved):

tests/unit_tests/test_helpers.py
dwmw2:
```
        (["fe80::1", "fe80::2%lo"], ["fe80::2%lo", "fe80::1"]),
```
Does work here (I *had* tested that manually long ago when I first added the sorting that I'm just re-using in thie PR).

But if the test is run on a system with a `lo` interface (or perhaps even without a LL address on the `lo` interface? I can't test that as I don't live in 1995), it would drop the 'invalid' scoped address and return only `fe80::1`.

Perhaps we *could* design a test which is less rigid about the 'expected' result... I think it's overkill though?

bdraco:
Thanks @dwmw2

bdraco:
Side note: we should refactor this to pass the sorted list of addresses back over the command line at some point and support happyeyeballs (`aioesphomeapi` already does)

PR #8539, opened 2025-04-09
[wts01] Add WTS01 temperature sensor component

alepee:

Add support for WTS01 temperature sensor used in Sonoff TH Elite (THR316, THR320, THR316D, THR320D) devices. This component reads temperature data over UART from these sensors and makes it available in ESPHome.

## Types of changes

- [ ] Other

- https://github.com/esphome/issues/issues/4189
- https://community.home-assistant.io/t/support-for-sonoff-wts01temperature-sensor-aka-ds18b20/745652
- https://community.home-assistant.io/t/sonoff-th-origin-16a-switch-thr316-with-ds18b20/481859

- esphome/esphome-docs#4811

## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [x] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
uart:
  rx_pin: GPIO17
  baud_rate: 9600

sensor:
  - platform: wts01
     name: "WTS01 Temperature"

     # Other options from the sensor component
     filters:
       - throttle: 60s
     accuracy_decimals: 0 # default is 1
     on_value:
       then:
         logger.log: "WTS01 sensor value: {{ id(wts01_sensor).get_temperature() }}"
```

## Checklist:


esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,104 @@
+#include "wts01.h"
+#include "esphome/core/log.h"
+#include <cmath>  // Pour std::abs
+
+namespace esphome {
+namespace wts01 {
+
+void WTS01Sensor::setup() {
+  // Empty setup - initialization is handled in the loop function
+}
```
swoboda1337: Remove setup then and remove the override in the header.

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,104 @@
...
+    buffer_pos_ = 0;
+    return;
+  }
+
+  // Add byte to buffer
+  buffer_[buffer_pos_++] = c;
+
+  // Process complete packet
+  if (buffer_pos_ == PACKET_SIZE) {
+    process_packet_();
```
swoboda1337: Everything protected should have `this->` in front: https://esphome.io/guides/contributing.html#codebase-standards

esphome/components/wts01/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+import esphome.codegen as cg
+from esphome.components import uart
+import esphome.config_validation as cv
+from esphome.const import CONF_ID, CONF_UART_ID, CONF_UPDATE_INTERVAL
+
+DEPENDENCIES = ["uart"]
+CODEOWNERS = ["@alepee"]
+
+wts01_ns = cg.esphome_ns.namespace("wts01")
+WTS01Sensor = wts01_ns.class_("WTS01Sensor", cg.Component)
```
swoboda1337: You have component here but pollingcomponent in the header. Should be the same. This should also inherit from uart::UARTDevice, look at the gps component.

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <cmath>  // Pour std::abs
+
+namespace esphome {
+namespace wts01 {
+
+void WTS01Sensor::setup() {
+  // Empty setup - initialization is handled in the loop function
+}
+
+void WTS01Sensor::update() {
```
swoboda1337: This should not be a polling component, should change it to a regular component. You can publish the temperature in `process_packet_` right after you decode it, no reason to delay it.

esphome/components/wts01/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+async def to_code(config):
+    var = cg.new_Pvariable(config[CONF_ID])
+    await cg.register_component(var, config)
+
+    if CONF_UART_ID in config:
+        uart_component = await cg.get_variable(config[CONF_UART_ID])
+    else:
+        uart_component = await cg.get_variable(config[uart.CONF_UART_ID])
+
+    cg.add(var.construct(uart_component))
```
swoboda1337: You should be a uart::UARTDevice and shouldnt have to do this, either way this function should be called set_uart if you are going to do it this way.

esphome/components/wts01/sensor.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+
+async def to_code(config):
+    parent = await cg.get_variable(config[CONF_WTS01_ID])
+
+    if CONF_TEMPERATURE in config:
+        sens = await sensor.new_sensor(config[CONF_TEMPERATURE])
+        # Créer une lambda qui appelle get_temperature() sur le parent
+        cg.add(sens.set_state_lambda(f"return {parent.get_id()}->get_temperature();"))
```
swoboda1337: Take a look at this component as an example: https://github.com/esphome/esphome/tree/dev/esphome/components/a02yyuw

You can see it only has one sensor and uses the UART. Its `__init__.py` is empty and everything is in `sensor.py`. You will also notice it inherits from sensor as well:
```
public sensor::Sensor, public Component, public uart::UARTDevice {
```

alepee: Thanks for the review and guidance.
I got it and managed to change my implementation accordingly. The overall design is much simpler this way 😄

swoboda1337: > Thanks for the review and guidance. I got it and managed to change my implementation accordingly. The overall design is much simpler this way 😄

Yes looks much better. However there is one more thing that can be simplified. Since you now inherit from sensor you are the sensor. Note `temperature_sensor_ ` is actually being set to `this` so:
```
    // Publish to sensor component if available
    if (this->temperature_sensor_ != nullptr) {
      this->temperature_sensor_->publish_state(temperature);
    }
```
can become:
```
      this->publish_state(temperature);
```

You can also remove:
```
    cg.add(var.set_temperature_sensor(var))
```
```
  void set_temperature_sensor(sensor::Sensor *temperature_sensor) { this->temperature_sensor_ = temperature_sensor; }
```
and
```
  sensor::Sensor *temperature_sensor_{nullptr};
```

You also don't need the function `get_temperature` if someone wants to access the temp in a lambda they can just do this: https://esphome.io/components/sensor/index.html#lambda-calls

alepee:
@swoboda1337 I also added an `update_interval` option to let user to define the granularity they need in [this commit](https://github.com/alepee/esphome/commit/0a83dc207b98d88871170b806538c011ba36090d)

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  // Check if it's time to publish the state
+  if (this->update_interval_ > 0) {
+    const uint32_t now = millis();
+    if (now - this->last_update_ >= this->update_interval_) {
+      this->last_update_ = now;
+      this->publish_state_();
+    }
+  }
```
swoboda1337: Instead of doing this why not control that with filters: https://esphome.io/components/sensor/index.html#sensor-filters

A throttle filter, throttle_average filter and/or a delta filter can limit the updates if required. The user can add filters in their yaml. Its better to use existing filter code instead trying to rewrite that inside your component.

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void WTS01Sensor::dump_config() {
+  ESP_LOGCONFIG(TAG, "WTS01 Temperature Sensor:");
+  LOG_SENSOR("  ", "Temperature", this->temperature_sensor_);
+  ESP_LOGCONFIG(TAG, "  Update Interval: %u ms", this->update_interval_);
+  this->check_uart_settings(9600);
+}
+
+void WTS01Sensor::publish_state_() {
```
swoboda1337: Remove this function your class already has publish_state inherited from sensor, call that

esphome/components/wts01/wts01.h
```diff
@@ -0,0 +1,45 @@
...
+  static const uint8_t PACKET_SIZE = 9;
+  static const uint8_t HEADER_1 = 0x55;
+  static const uint8_t HEADER_2 = 0x01;
+  static const uint8_t HEADER_3 = 0x01;
+  static const uint8_t HEADER_4 = 0x04;
+
+  uint8_t buffer_[PACKET_SIZE];
+  uint8_t buffer_pos_{0};
+  float current_temperature_{NAN};
+  sensor::Sensor *temperature_sensor_{nullptr};
```
swoboda1337: Remove this variable it points to `this` anyways, dont need it as you inherit from sensor and your class is the sensor.

esphome/components/wts01/wts01.cpp
alepee: I found several components going this way, and mimic the behaviour. I'm ok with filters and agree to simplify my own code.
The only drawback to this would be the overhead for users. Maybe some examples in the docs to illustrate can remediate to this issue.

swoboda1337: Yeah an example in the docs sounds good

alepee:
@swoboda1337 thanks for your feedback, I made an update accordingly and think I'm ready to merge :)

The doc updated with example usage
https://deploy-preview-4811--esphome.netlify.app/components/sensor/wts01.html

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,94 @@
...
+    int sign = 1;
+
+    // Handle negative temperatures
+    if (temp > 127) {
+      temp -= 128;
+      sign = -1;
+    }
+
+    // Calculate temperature (temp + decimal/100)
+    float temperature = sign * (static_cast<float>(temp) + (static_cast<float>(this->buffer_[7]) / 100.0f));
```
swoboda1337, line 83: Seems like all of this:
```
    uint8_t temp = this->buffer_[6];
    int sign = 1;

    // Handle negative temperatures
    if (temp > 127) {
      temp -= 128;
      sign = -1;
    }

    // Calculate temperature (temp + decimal/100)
    float temperature = sign * (static_cast<float>(temp) + (static_cast<float>(this->buffer_[7]) / 100.0f));
```
 could be replaced by one line:
 ```
 float temperature = (int16_t) ((temp1 << 8) | temp2) / 100.0f;
 ```

swoboda1337: Yeah looking good except that one piece of code.

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,94 @@
...
+  // Based on Tasmota implementation
+  // Format: 55 01 01 04 01 11 16 12 95
+  // header            T  Td Ck  - T = Temperature, Td = Temperature decimal, Ck = Checksum
+
+  // Check if packet is valid (we could add checksum verification here if needed)
+  if (this->buffer_[0] == HEADER_1 && this->buffer_[1] == HEADER_2 && this->buffer_[2] == HEADER_3 &&
+      this->buffer_[3] == HEADER_4) {
+    // Extract temperature value
+    uint8_t temp = this->buffer_[6];
+    int sign = 1;
```
swoboda1337: Sorry misread the code the first time it looks fine. Should use float or int32_t instead of int though.

alepee: Nice catch! Would an `int8_t` be more appropriate here, since `sign` can only hold `1` or `‑1`?

alepee: Got an explanation from Claude about `32int_t` vs `8int_t` 😃 

> All three platforms (ESP32, ESP8266, and RP2040) are 32-bit microcontrollers, so using `int32_t` aligns with their native word size, which could provide better performance. 
> 
> On 32-bit processors, operations with 32-bit values are often more efficient than with 8-bit values due to:
> 1. Natural alignment with processor registers
> 2. Potential need for fewer instructions (no need for additional masking/sign extension)
> 
> While `int8_t` would work and save some memory, the performance benefit of using the processor's native word size (`int32_t`) might be why the reviewer suggested it. For a single variable like `sign`, the memory difference is negligible, but the potential performance gain might be worthwhile, especially in frequently executed code.
>

alepee:
Hey @swoboda1337 is everything ok for you ?

alepee:
@swoboda1337 would it be possible to validate this PR for it to have a chance to be released with 2025.5 ? 🙏

swoboda1337:
Way too late for 2025.5, will need Jesse or Keith to merge it.

swoboda1337:
FYI the beta is normally cut a week before the release: https://beta.esphome.io/changelog/2025.5.0

alepee:
Indeed, I received the 2025.5 update on my ESPHome this morning 😅  let's aim for next month release.

davet2001:
Tested the PR on my hardware here and all working well. :+1:

igorlistopad:
@swoboda1337 Will this PR be added in 2025.6?

swoboda1337:
> @swoboda1337 Will this PR be added in 2025.6?

Not up to me. There are 100s of PRs waiting though.

alepee:
Is everything ok for you @swoboda1337 regarding this PR, or is there any required changes? If it is ok for you, could you approve it, or is it on somebody else?

igorlistopad:
Hi @jesserockz, could you take a look at this component?
This component has been ready for more than 2 months and is awaiting review.
If you are satisfied with its implementation, please include it in the next beta version.

swoboda1337:
There are many components waiting for much longer. I have one that has been waiting for 9 months.

CodeMystic21:
Esphome really has to do something about their maintenance.
If they don't even manage to review prs.
The open list of prs is just getting longer and longer ...

If you're not someone "important" who knows @jesserockz, you will get ignored.

igo-r:
This is terrible, something that has long been implemented in other systems, and ESPHome has to wait forever for adoption.
The device will be removed from sale sooner than ESPHome will receive its support.
In general, such a policy is completely demotivating, and discourages any desire to contribute to ESPHome.
When your small PR can wait months or years.

CodeMystic21:
> This is terrible, something that has long been implemented in other systems, and ESPHome has to wait forever for adoption. The device will be removed from sale sooner than ESPHome will receive its support. In general, such a policy is completely demotivating, and discourages any desire to contribute to ESPHome. When your small PR can wait months or years.

Be careful with such messages...
If you write something they don't like, you will get censored pretty quickly.

davet2001:
I have found some reliability issues with this when running on a sonoff THR320.

The startup of the 1-wire sensor seems to be as follows:

WTS01 transmits at what looks like 9600bps, 8bits, none parity
1. Power on WTS01 v+
2. WTS01 device transmits a very fast burst on the DATA_OUT pin for about 300ms (not sure what this is, possibly identification data).
3. WTS01 transmits a series of 8 byte messages, approx every 100ms, this continues for about 5s.
4. WTS01 transmits one 8 byte message every ~1s and continues this indefinitely

What I observed: The timing of the power on of the WTS device seems to be significant. If I set the V+ power pin to ALWAYS_ON, then power off the device completely, and power it back on, V+ comes on at startup, the WT01 immediately sends the burst of data, and then ESP reports 'nan' (not a number) continuously. Strangely the WTS is transmitting valid data, but the ESP isn't recognising it.

image

If I then _reboot_ the device without powering off (warm boot), the WTS01 V+ remains powered, but the ESP wakes up and starts listening to it. In that scenario the ESP successfully reads data (and updates it on my lcd display in my test case).

A partial workaround seems to be:

If I insert a V+ power off for 0.5s to occur with a priority of -50 (after the end of startup followed by a 1s delay), this resets the WTS01 and it begins working normally
```
  on_boot:
    #after data/comms is initialised.
    - priority: -50
      then:
        # Power cycle the 1-wire bus power supply
        # to get comms into a known state
        - delay: 1s
        - switch.turn_off: sensor_power
        - delay: 0.5s
        - switch.turn_on: sensor_power 
```
When I do this, I get valid temperature values for cold boot, but but unfortunately for warm boot, I get only 1-2 valid readings then it stops updating.

image

Again, the WTS01 is still transmitting valid data, but the ESP device is ignoring it.

image

Some speculation:
Can the driver cope with the sudden burst of data? Can it cope with that at all stages of the boot? I wonder if a buffer is getting overflowed if that burst of data comes at the wrong time (e.g. while the ESP is busy doing something else such as wifi startup). Potentially we are very much relying on RTOS queues/buffers in that scenario, and if data comes in really quick we might begin to hit the 'buffer full' code and find out if it can recover by itself. It seems to me like it can't.

TLDR: I'm not finding this fully robust at the moment.

alepee:
Thank you so much for the deep dive and details @davet2001  🙏 

Following your suggestion, I've pushed a first commit that replaces the blocking `while (available())` loop with a non-blocking `if` condition. This change alone should prevent the UART buffer from overflowing during data bursts and make the component more cooperative with the ESPHome scheduler.

Additionally, your comment prompted me to investigate further. I realised that the component was not validating the packet checksum, which could lead to corrupt data (and thus `nan` values) being processed. I've added a second commit that implements a checksum verification. The component will now discard any invalid packets.

With these two changes combined, non-blocking processing and a data integrity check, I am much more confident that the stability issues you observed should be fully resolved.

If you have a moment, would you be able to re-test with these latest commits and let me know if everything works as expected on your end?

davet2001:
Hi @alepee I have retested this - unfortunately I don't see any improvement (in fact slightly worse).

- Without my workaround startup code, the sensor reports 'nan' on cold boot, and works fine on warm boot
- With my workaround startup code, it reads a fixed value once on cold boot (always 25.0) then freezes with that value, and on warm boot, it reads one valid value and freezes on that value.

In both cases, I'm certain the micro is still running because I have a hh:mm:ss time on the display which is still counting.

Workaround code:
```
  on_boot:
    #after data/comms is initialised.
    - priority: -50
      then:
        # Power cycle the 1-wire bus power supply
        # to get comms into a known state
        - delay: 1s
        - switch.turn_off: sensor_power
        - delay: 0.5s
        - switch.turn_on: sensor_power
```

davet2001:
After some more debugging, I think I've got to the bottom of this. It looks like the state machine logic was not quite right for synchronising to the header.

It could handle the straightforward case with normal data coming in the right order, but if something went wrong it would not recover. 

The use of 
```
  if (c == HEADER_1 && this->buffer_pos_ == 0) {
    this->buffer_[this->buffer_pos_++] = c;
    return;
  }
```
Does not handle the case of resetting the state machine if the first byte is invalid. Looking at the startup sequence on the scope and putting in some debug IO toggles, I could see that bytes were being received, but all of them were being rejected.

The problem starts towards the end of the rapid burst of data during startup of the WTS01.

I used this debug code:
```
  GPIO_00_HI;
  GPIO_13_LO;
  if (c == HEADER_1 && this->buffer_pos_ == 0) {
    this->buffer_[this->buffer_pos_++] = c;
    return;
  }

  // Check header sequence
  if (this->buffer_pos_ == 1 && c != HEADER_2) {
    this->buffer_pos_ = 0;
    GPIO_00_LO;
    return;
  }

  if (this->buffer_pos_ == 2 && c != HEADER_3) {
    this->buffer_pos_ = 0;
    GPIO_00_LO;
    return;
  }

  if (this->buffer_pos_ == 3 && c != HEADER_4) {
    this->buffer_pos_ = 0;
    GPIO_00_LO;
    return;
  }

  // If we reach here, so far so good. Add byte to buffer.
  this->buffer_[this->buffer_pos_++] = c;

  // Process complete packet
  if (this->buffer_pos_ >= PACKET_SIZE) {
    GPIO_13_HI;
    GPIO_00_LO;
    this->process_packet_();
    this->buffer_pos_ = 0;
  }
```
This means IO-0 is telling me about the start of each byte coming in, and stays high until either the packet is rejected or accepted.
IO13 goes high only when a complete packet is received, and resets on receipt of the next byte.

image

As you can see from this trace of a cold boot including a WTS power cycle after about 10s:
 - We stop receiving valid packets after a about 18s (1)
 - This seems to be at the end of the burst of incoming data (2)
 - Valid bytes are still being transmitted by the WTS-01 (3)
 - The state machine is still receiving the first byte (4)

But... we are never getting to the end of a packet.


I have a working mod that now gives me reliable comms, even if I remove the power cycle workaround. Will add to the PR as a suggestion.

I also spotted another minor improvement to make the sequence a bit more resilient. I will add both as suggestions to the PR.

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,100 @@
...
+  this->check_uart_settings(9600);
+}
+
+void WTS01Sensor::handle_char_(uint8_t c) {
+  // Reset buffer if we're at the start of a new packet
+  if (c == HEADER_1 && this->buffer_pos_ == 0) {
+    this->buffer_[this->buffer_pos_++] = c;
+    return;
+  }
+
```
davet2001: You need to reset the state machine/buffer counter if the first byte doesn't match, not just continue if it does match.

```suggestion
  // State machine for processing the header. Reset if something doesn't match.
  if (this->buffer_pos_ == 0 && c != HEADER_1) {
    this->buffer_pos_ = 0;
    return;
  }

```

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,100 @@
...
+  if (this->buffer_pos_ == 3 && c != HEADER_4) {
+    this->buffer_pos_ = 0;
+    return;
+  }
+
+  // Add byte to buffer
+  this->buffer_[this->buffer_pos_++] = c;
+
+  // Process complete packet
+  if (this->buffer_pos_ == PACKET_SIZE) {
```
davet2001: Better to have a greater-or-equal-check when dealing with a buffer position counter like this. This change will reduce the chance of accidentally overrunning the buffer in case of any logic errors.
```suggestion
  if (this->buffer_pos_ >= PACKET_SIZE) {
```

davet2001 (changes requested):
Suggested changes to fix the comms reliability problem and improve robustness.

alepee:
> Suggested changes to fix the comms reliability problem and improve robustness.

Thanks a lot for your review and suggested changes. Those totally make sense

davet2001 (approved):

igorlistopad:
@alepee I want to express my gratitude to you for the work done.
I am now developing the firmware for my device and conducted all the tests of this component with full logger.
Everything works just fine.

esphome/components/wts01/wts01.h
```diff
@@ -0,0 +1,37 @@
...
+
+  uint8_t buffer_[PACKET_SIZE];
+  uint8_t buffer_pos_{0};
+  float current_temperature_{NAN};
+
+  void handle_char_(uint8_t c);
+  void process_packet_();
+
+ private:
+  static constexpr const char *TAG = "wts01";
```
jesserockz:
```suggestion
```

esphome/components/wts01/wts01.h
```diff
@@ -0,0 +1,37 @@
...
+  float get_setup_priority() const override { return setup_priority::DATA; }
+
+  float get_temperature() const { return this->current_temperature_; }
+
+ protected:
+  static const uint8_t PACKET_SIZE = 9;
+  static const uint8_t HEADER_1 = 0x55;
+  static const uint8_t HEADER_2 = 0x01;
+  static const uint8_t HEADER_3 = 0x01;
+  static const uint8_t HEADER_4 = 0x04;
```
jesserockz:
```suggestion
```

esphome/components/wts01/sensor.py
```diff
@@ -0,0 +1,36 @@
...
+from esphome.components import sensor, uart, wts01
+import esphome.config_validation as cv
+from esphome.const import (
+    DEVICE_CLASS_TEMPERATURE,
+    STATE_CLASS_MEASUREMENT,
+    UNIT_CELSIUS,
+)
+
+CONF_WTS01_ID = "wts01_id"
+DEPENDENCIES = wts01.DEPENDENCIES
```
jesserockz:
```suggestion
CODEOWNERS = ["@alepee"]
DEPENDENCIES = ["uart"]
```

esphome/components/wts01/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""WTS01 temperature sensor component for Sonoff TH Elite (THR316, THR320, THR316D, THR320D)."""
+
+CODEOWNERS = ["@alepee"]
+DEPENDENCIES = ["uart"]
```
jesserockz:
```suggestion
```

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,99 @@
...
+static const uint8_t HEADER_1 = 0x55;
+static const uint8_t HEADER_2 = 0x01;
+static const uint8_t HEADER_3 = 0x01;
+static const uint8_t HEADER_4 = 0x04;
+
+void WTS01Sensor::loop() {
+  // Process one character at a time received from the sensor
+  if (available()) {
+    uint8_t c;
+    if (read_byte(&c)) {
```
jesserockz: > All uses of class members and member functions should be prefixed with this-> to distinguish them from global functions/variables.

https://developers.esphome.io/contributing/code/#c-code-style
```suggestion
  if (this->available()) {
    uint8_t c;
    if (this->read_byte(&c)) {
```

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,99 @@
...
+      this->handle_char_(c);
+    }
+  }
+}
+
+void WTS01Sensor::dump_config() {
+  ESP_LOGCONFIG(TAG, "WTS01 Temperature Sensor:");
+  LOG_SENSOR("  ", "Temperature", this);
```
jesserockz:
```suggestion
  LOG_SENSOR("", "WTS01 Temperature Sensor", this);
```

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,99 @@
...
+  this->check_uart_settings(9600);
```
jesserockz: If this requires 9600, it should be using `uart.final_validate_device_schema` in `sensor.py` to ensure the baud is set to 9600 at validation time.
```suggestion
```

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,99 @@
...
+  if (temp > 127) {
+    temp -= 128;
+    sign = -1;
+  }
+
+  // Calculate temperature (temp + decimal/100)
+  float temperature = sign * (static_cast<float>(temp) + (static_cast<float>(this->buffer_[7]) / 100.0f));
+
+  // Store temperature
+  this->current_temperature_ = temperature;
```
jesserockz:
```suggestion
```

esphome/components/wts01/wts01.h
```diff
@@ -0,0 +1,37 @@
...
+namespace esphome {
+namespace wts01 {
+
+class WTS01Sensor : public sensor::Sensor, public uart::UARTDevice, public Component {
+ public:
+  void loop() override;
+  void dump_config() override;
+  float get_setup_priority() const override { return setup_priority::DATA; }
+
+  float get_temperature() const { return this->current_temperature_; }
```
jesserockz: `.state` already exists on sensors
```suggestion
```

esphome/components/wts01/wts01.h
```diff
@@ -0,0 +1,37 @@
...
+ protected:
+  static const uint8_t PACKET_SIZE = 9;
+  static const uint8_t HEADER_1 = 0x55;
+  static const uint8_t HEADER_2 = 0x01;
+  static const uint8_t HEADER_3 = 0x01;
+  static const uint8_t HEADER_4 = 0x04;
+
+  uint8_t buffer_[PACKET_SIZE];
+  uint8_t buffer_pos_{0};
+  float current_temperature_{NAN};
```
jesserockz:
```suggestion
```

jesserockz (dismissed):

ssieb:
You need to read more than 1 byte per loop.  Here's my component:
https://github.com/ssieb/esphome_components/tree/master/components/wts01

alepee:
Thanks for the review @jesserockz 🙏

GrumpyMeow:
Came here to confirm that this PR works on my THR320D with temperature-probe. Thank you very much!

igorlistopad:
Hi, @alepee, are you okay?
Could you review and apply the necessary changes and mark the PR as ready for review to complete the implementation of this component?

igorlistopad:
All the changes proposed by me and @jesserockz have been tested and are working as expected.

I also conducted tests on suddenly disconnecting the sensor and reconnecting it while the device was running, and everything worked as expected after reconnecting the sensor.

davet2001:
@alepee can the PR be changed from 'Draft' to 'Ready'?

alepee:
@davet2001, I did not take the time to acknowledge @jesserockz review yet.
If you want to contribute on this one, I can grant you write access to my fork

davet2001:
@alepee The access invite has expired, please could you resend it?

alepee:
Done @davet2001 :)

igorlistopad:
@alepee Hi, if you don't have much time, could you send me an invitation so I can accept the edits and change the PR to Ready?
We've been standing still for a long time.
I've been using the device for over a month and have accepted all the changes locally.

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+namespace esphome {
+namespace wts01 {
+
+static const char *const TAG = "wts01";
+static const uint8_t PACKET_SIZE = 9;
+static const uint8_t HEADER_1 = 0x55;
+static const uint8_t HEADER_2 = 0x01;
+static const uint8_t HEADER_3 = 0x01;
+static const uint8_t HEADER_4 = 0x04;
```
swoboda1337, line 13:
```suggestion
```

esphome/components/wts01/wts01.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include "esphome/core/component.h"
+#include "esphome/components/sensor/sensor.h"
+#include "esphome/components/uart/uart.h"
+
+namespace esphome {
+namespace wts01 {
+
```
swoboda1337, line 9:
```suggestion

constexpr uint8_t PACKET_SIZE = 9;
constexpr uint8_t HEADER_1 = 0x55;
constexpr uint8_t HEADER_2 = 0x01;
constexpr uint8_t HEADER_3 = 0x01;
constexpr uint8_t HEADER_4 = 0x04;

```

swoboda1337 (changes requested):

swoboda1337:
I commited all the suggested changes, just one left to resolve. BTW anyone can create a PR on alepee:feat/add-support-for-wts01 if alepee merges it it will update this PR, you don't need permissions to do that.

igorlistopad:
> I commited all the suggested changes, just one left to resolve. BTW anyone can create a PR on alepee:feat/add-support-for-wts01 if alepee merges it it will update this PR, you don't need permissions to do that.

Thanks, all that's left is to remove `this->check_uart_settings(9600);` and add `FINAL_VALIDATE_SCHEMA`, as well as replace `if (available())` with `while (this->available())`, otherwise we'll be processing messages for too long.

alepee:
Thank you @swoboda1337 ❤️

esphome/components/wts01/wts01.cpp
alepee: I kept the baud rate check in the cpp file and added an additional check in sensor.py for validation time.

alepee:
I will take some time tomorrow to check the CI. I hope to get it done shortly

alepee:
All checks have passed! ✅ 🎉

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,88 @@
...
+  int32_t sign = 1;
+
+  // Handle negative temperatures
+  if (temp > 127) {
+    temp -= 128;
+    sign = -1;
+  }
+
+  // Calculate temperature (temp + decimal/100)
+  float temperature = sign * (static_cast<float>(temp) + (static_cast<float>(this->buffer_[7]) / 100.0f));
```
swoboda1337:
```suggestion
  int8_t temp = this->buffer_[6];
  int32_t sign = 1;

  // Handle negative temperatures
  if (temp < 0) {
    sign = -1;
  }

  // Calculate temperature (temp + decimal/100)
  float temperature = static_cast<float>(temp) + sign * static_cast<float>(this->buffer_[7]) / 100.0f);
```
Should cast to an int8_t so you don't have to manually check for 127 and subtract by 128

esphome/components/wts01/wts01.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+}
+
+void WTS01Sensor::dump_config() { LOG_SENSOR("", "WTS01 Temperature Sensor", this); }
+
+void WTS01Sensor::handle_char_(uint8_t c) {
+  // State machine for processing the header. Reset if something doesn't match.
+  if (this->buffer_pos_ == 0 && c != HEADER_1) {
+    this->buffer_pos_ = 0;
```
swoboda1337:
```suggestion
```
Don't need this line

swoboda1337 (changes requested):

swoboda1337:
Docs need to be updated after the switch from rst to md

swoboda1337 (dismissed):

swoboda1337 (approved):

PR #8542, opened 2025-04-10
[esp32, debug] Add ``cpu_frequency`` config option and debug sensor

clydebarrow:

* Add a sensor and log messages giving CPU frequency to the esp32 debug component
* Revise the existing `arch_get_cpu_freq_hz()` helper function to use only documented API calls.
* Add an option to set the cpu frequency
* Use generic documented ESP-IDF calls to get reset and sleep reasons.

NOTE: a side-effect of this PR is that the default CPU frequency for Arduino becomes 160MHz, rather than the 240MHz value baked into the Arduino core. ESP-IDF default was already 160.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4812
## Test Environment

- [x] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
esp32:
  board: um_feathers2
  cpu_frequency: 240MHz

debug:

sensor:
  - platform: debug
    cpu_frequency:
      name: CPU Frequency


text_sensor:
  - platform: debug
    device:
      name: "Device Info"

```

## Checklist:


clydebarrow:

```yaml
external_components:
  - source: github://pr#8542
    components: [debug, esp32]
    refresh: 1h
```

---
(Added by the PR bot)

esphome/components/esp32/__init__.py
```diff
@@ -693,6 +746,7 @@ async def to_code(config):
                 f"VERSION_CODE({framework_ver.major}, {framework_ver.minor}, {framework_ver.patch})"
             ),
         )
+        cg.add(RawExpression(f"setCpuFrequencyMhz({freq})"))
```
swoboda1337, line 752: Is this line needed? Seems to be some arduino function that doesnt support newer variants. Setting both of these should be all that is needed, don't need to call that:
```
      CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ_240: "y"
      CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ: "240"
```

esphome/components/esp32/__init__.py
```diff
@@ -628,6 +675,12 @@ async def to_code(config):
         add_idf_sdkconfig_option("CONFIG_ESP_TASK_WDT_CHECK_IDLE_TASK_CPU0", False)
         add_idf_sdkconfig_option("CONFIG_ESP_TASK_WDT_CHECK_IDLE_TASK_CPU1", False)
 
+        # Set default CPU frequency
+        add_idf_sdkconfig_option(f"CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ_{freq}", True)
+        add_idf_sdkconfig_option(
+            f"CONFIG_{config[CONF_VARIANT]}_DEFAULT_CPU_FREQ_{freq}", True
```
swoboda1337: This one isnt needed should only need these two:
```
      CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ_240: "y"
      CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ: "240"
```
You can see here that they rename the old style configs into the new ones: https://github.com/espressif/esp-idf/blob/1c468f68259065ef51afd114605d9122f13d9d72/components/esp_system/sdkconfig.rename.esp32s3

There are similar rename files for esp32, esp32s2, esp32s3 and esp32c3

Those two configs should work on all variants

clydebarrow: Well, in testing I found it necessary, perhaps because we are only up to 5.1.5 and the change you linked to is more recent.

esphome/components/esp32/__init__.py
clydebarrow: Can't set those for Arduino, the ESP-IDF libraries used in Arduino are pre-compiled.

swoboda1337: Ok its under an ARDUINO if my bad ignore this

esphome/components/esp32/__init__.py
swoboda1337: Hmm another reason to switch to 5.3 :)

swoboda1337: That does exist in 5.1 though https://github.com/espressif/esp-idf/blob/release/v5.1/components/esp_system/sdkconfig.rename.esp32s3

swoboda1337: BTW I just tested this using 5.1 on S3 and this works:
```
      CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ_240: "y"
      CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ: "240"
```
You only have one of these configs btw, maybe try adding `CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ`

clydebarrow: tested again without the device specific version and seems to work ok. `CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ` and similar are only used by `menuconfig` as far as I can tell, they do not affect the compilation.

esphome/components/debug/debug_esp32.cpp
```diff
@@ -1,11 +1,15 @@
...
 #include "esphome/core/log.h"
+#include "esphome/core/hal.h"
 
 #include <esp_heap_caps.h>
 #include <esp_system.h>
 #include <esp_chip_info.h>
 #include <esp_partition.h>
+#ifdef USE_ESP_IDF
+#include <esp_clk_tree.h>
+#endif
```
swoboda1337: Still need this include?

esphome/components/debug/debug_component.cpp
```diff
@@ -25,6 +25,9 @@ void DebugComponent::dump_config() {
 #ifdef USE_SENSOR
   LOG_SENSOR("  ", "Free space on heap", this->free_sensor_);
   LOG_SENSOR("  ", "Largest free heap block", this->block_sensor_);
+#ifdef USE_ESP32
+  LOG_SENSOR("  ", "CPU frequency", this->cpu_frequency_sensor_);
+#endif
```
swoboda1337: Since this now uses arch_get_cpu_freq_hz() any reason this has to be esp32 only as that function should exist on all platforms.

esphome/components/debug/debug_esp32.cpp
clydebarrow: No.

esphome/components/debug/debug_component.cpp
clydebarrow: It does exist on the other platforms, though those I looked at don't actually query the current frequency, they just report the configured frequency from a #define, so it's of less use/interest. But easy enough.

esphome/components/debug/debug_esp32.cpp
```diff
@@ -289,6 +290,8 @@ void DebugComponent::get_device_info_(std::string &device_info) {
   device_info += features;
   device_info += " Cores:" + to_string(info.cores);
   device_info += " Revision:" + to_string(info.revision);
+  device_info += str_sprintf("|CPU Frequency: %u MHz", arch_get_cpu_freq_hz() / 1000000);
+  ESP_LOGD(TAG, "CPU Frequency: %u MHz", arch_get_cpu_freq_hz() / 1000000);
```
swoboda1337: As discussed in discord there are some errors on p4 with the latest IDF
```
Compiling .pioenvs/esp32p4-test/src/esphome/components/debug/debug_libretiny.cpp.o
src/esphome/components/debug/debug_esp32.cpp: In member function 'std::string esphome::debug::DebugComponent::get_reset_reason_()':
src/esphome/components/debug/debug_esp32.cpp:51:10: error: enumeration value 'ESP_RST_UNKNOWN' not handled in switch [-Werror=switch]
   51 |   switch (esp_reset_reason()) {
      |          ^
src/esphome/components/debug/debug_esp32.cpp:51:10: error: enumeration value 'ESP_RST_USB' not handled in switch [-Werror=switch]
src/esphome/components/debug/debug_esp32.cpp:51:10: error: enumeration value 'ESP_RST_JTAG' not handled in switch [-Werror=switch]
src/esphome/components/debug/debug_esp32.cpp:51:10: error: enumeration value 'ESP_RST_EFUSE' not handled in switch [-Werror=switch]
src/esphome/components/debug/debug_esp32.cpp:51:10: error: enumeration value 'ESP_RST_PWR_GLITCH' not handled in switch [-Werror=switch]
src/esphome/components/debug/debug_esp32.cpp:51:10: error: enumeration value 'ESP_RST_CPU_LOCKUP' not handled in switch [-Werror=switch]
src/esphome/components/debug/debug_esp32.cpp: In member function 'void esphome::debug::DebugComponent::get_device_info_(std::string&)':
src/esphome/components/debug/debug_esp32.cpp:300:48: error: format '%u' expects argument of type 'unsigned int', but argument 2 has type 'long unsigned int' [-Werror=format=]
  300 |   device_info += str_sprintf("|CPU Frequency: %u MHz", arch_get_cpu_freq_hz() / 1000000);
      |                                               ~^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                |                              |
      |                                                unsigned int                   long unsigned int
      |                                               %lu
In file included from src/esphome/components/sensor/sensor.h:3,
                 from src/esphome/components/debug/debug_component.h:9,
                 from src/esphome/components/debug/debug_esp32.cpp:1:
src/esphome/components/debug/debug_esp32.cpp:301:17: error: format '%u' expects argument of type 'unsigned int', but argument 5 has type 'long unsigned int' [-Werror=format=]
  301 |   ESP_LOGD(TAG, "CPU Frequency: %u MHz", arch_get_cpu_freq_hz() / 1000000);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~
```

esphome/components/esp32/__init__.py
```diff
@@ -70,7 +78,40 @@
...
+    variant = config[CONF_VARIANT]
+    # if not specified in config, set to 160MHz if supported, the fastest otherwise
+    if cpu_frequency is None:
+        choices = CPU_FREQUENCIES[variant]
+        if "160MHZ" in choices:
+            cpu_frequency = "160MHZ"
+        else:
+            cpu_frequency = choices[-1]
+        config[CONF_CPU_FREQUENCY] = cpu_frequency
+    elif cpu_frequency not in CPU_FREQUENCIES[variant]:
```
jesserockz, line 112: might want to put a catch in here for `variant not in CPU_FREQUENCIES` otherwise you will get a not nice error instead of validation error for new variants

clydebarrow: If it's a variant that the user specified that is not in the implemented list, it will fail anyway:

```
  Unknown value 'ESP32C7', did you mean 'ESP32C6', 'ESP32C3', 'ESP32C2'?.
```

If a new variant is added, and this table is not updated, it will fail if a test for that variant has been added to the CI tests. But if that were missed, I've added an assertion.

esphome/components/debug/debug_esp32.cpp
```diff
@@ -29,6 +19,67 @@ namespace debug {
...
+  if (reason < sizeof(RESET_REASONS) / sizeof(RESET_REASONS[0])) {
+    reset_reason = RESET_REASONS[reason];
+  } else {
+    reset_reason = "unknown source";
+  }
+  ESP_LOGD(TAG, "Reset Reason: %s", reset_reason.c_str());
+  return "Reset by " + reset_reason;
+}
+
+const char *const WAKEUP_CAUSES[] = {
```
swoboda1337: Static?

kbx81 (approved):
Thanks! 🍻

PR #8544, opened 2025-04-10
[API] Sub devices and areas

### LINKED ISSUE #1335: Improve HA device registry and support multiple devices in one config

I recently got started with ESPHome and I like it very much! Most of my cases for ESPHome are in situations where I try to build a 'digital twin' for multiple physical devices. Think of a physical standing desk or Xiaomi MiFlora devices.

However, currently when you configure this, it will add all entities to a single espressif device in Home Assistant. It would be great if you could use the ESPHome YAML to create devices and decide which sensors are linked to which devices.

Use cases:
- ESP32 with Xaomi MiFlora devices
- ESP32 with multiple Switchbot Curtains
- ESP32 with multiple Oral B toothbrushes
(or even a combination of the ones above)

Eventually, this would allow us to show the `model`, `manufacturer` and `firmware` of the device we mimic, instead of our ESPHome device. I was looking to port https://github.com/devWaves/SwitchBot-MQTT-BLE-ESP32, however the features regarding device registry and information are not possible (yet) with ESPHome. 

**Screenshots**
Current situation
image

Proposed situation:
image

(and with multiple devices, there would be multiple entries)

Prefixing all entities with the digital twin device name, however this will be messy at some point and doesn't show up nice in the Home Assistant device registry.

**Additional context**

This would mainly be a cosmetic change, so I can understand that this is not something which has any priority over the other issues. However, perhaps there is someone that would like to tackle this, or work with me to see how we can add this. 
(unfortunately my C++ is really limited, however I can do Python :-)).

**Example YAML**
```yaml
# If not added to a device, it will be added to the main device
sensor:
  - platform: wifi_signal
    name: "WiFi Signal"
    update_interval: 60s

# If added to a device, it will create a new device in the Device Registry
device:
  desk:
    name: Desk Mick
    model: EK5
    manufacturer: Flexispot

  - platform: uart
    name: "Preset 1"
    id: switch_preset1
    icon: mdi:numeric-1-box
    data: [0x9b, 0x06, 0x02, 0x04, 0x00, 0xac, 0xa3, 0x9d]
    uart_id: desk_uart
```

### LINKED ISSUE #6953: Config validation does not ensure all names are unique per platform/domain
### The problem

Consider the following YAML
```
binary_sensor:
  - platform: template
    name: ""
    device_class: gas
    lambda: 'return true;'
  - platform: template
    name: ""
    device_class: lock
    lambda: 'return true;'
  - platform: template
    name: "xyz"
    device_class: garage_door
    lambda: 'return true;'
  - platform: template
    name: "xyz"
    device_class: lock
    lambda: 'return true;'    
```

The additional `binary_sensor` named `""` and `"xyz"` should be an error as its going to generate the same object_id, and this produce a unique id conflict
```
2025-04-20 09:19:05.870 ERROR (MainThread) [homeassistant.components.binary_sensor] Platform esphome does not generate unique IDs. ID 24:4C:AB:06:49:9C-binary_sensor-xyz is already used by binary_sensor.poolhouse81proxy_xyz - ignoring binary_sensor.poolhouse81proxy_xyz
2025-04-20 09:19:05.871 ERROR (MainThread) [homeassistant.components.binary_sensor] Platform esphome does not generate unique IDs. ID 24:4C:AB:06:49:9C-binary_sensor-81_pool_house_proxy is already used by binary_sensor.poolhouse81proxy - ignoring binary_sensor.poolhouse81proxy
```

dev

Home Assistant Add-on

_No response_

ESP32

### Board

_No response_

_No response_

### YAML Config

```yaml

```

```txt

```

### Additional information

_No response_

dala318:

This PR adds support for sub-devices and areas in ESPHome, enabling a single ESP to represent multiple logical devices in Home Assistant. This is a foundational feature for building smart home hubs and gateways.

Additionally, this PR adds validation to prevent duplicate entity names within the same platform and device, providing clearer configuration errors.

### 1. Hub/Gateway Functionality
When an ESP acts as a bridge for multiple devices connected through various protocols:
- **RF bridges**: Each 433MHz/868MHz device (door sensors, PIR sensors, etc.) appears as its own device in Home Assistant rather than entities under the bridge
- **Modbus gateways**: Individual Modbus devices (energy meters, HVAC controllers) get their own device representation
- **I2C/SPI sensor arrays**: Multiple sensors on the same bus can be grouped as separate logical devices
- **Serial/UART bridges**: Devices connected via serial protocols can each have their own identity

### 2. Multi-Zone Controllers
- One ESP controlling multiple rooms or areas
- Each zone appears as a separate device with its own area assignment
- Better organization without deploying multiple physical ESPs

### Sub-Devices and Areas
- Group entities into logical sub-devices that appear separately in Home Assistant
- Assign sub-devices to specific areas (rooms) for better organization
- Each sub-device can have its own area assignment independent of the ESP's physical location

### Flexible Configuration
- Define areas using either simple string format or structured format with `id` and `name`
- Create multiple sub-devices, each assignable to different areas
- Entities can be assigned to specific sub-devices using `device_id`

### Implementation Details
- Uses hash-based unique IDs for devices and areas
- Memory-efficient design using `const char*` for names
- Separate compile-time flags (`USE_AREAS` and `USE_DEVICES`) to minimize overhead
- Areas can be used without devices if needed
- Validates entity names to prevent duplicates within the same platform
- Duplicate detection based on (platform, sanitized_name) tuple
- Same entity names only work across different platforms (not across devices due to hash collisions)

### API Protocol
- New `AreaInfo` and `DeviceInfo` messages in API protocol
- Entities report their device association to Home Assistant
- Clean separation between physical ESP device and sub-devices

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/feature-requests/issues/1335
- fixes https://github.com/esphome/issues/issues/6953

- esphome/esphome-docs#4813

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

### Example 1: RF Bridge Gateway
```yaml
# ESP32 acting as RF bridge for multiple 433MHz devices
esphome:
  name: rf-bridge
  area:
    id: bridge_area
    name: "Utility Room"
  
  devices:
    - id: front_door_device
      name: "Front Door Sensor"
      area_id: entrance_area
    - id: kitchen_motion_device
      name: "Kitchen Motion"
      area_id: kitchen_area
    - id: garage_door_device
      name: "Garage Door"
      area_id: garage_area
  
  areas:
    - id: entrance_area
      name: "Entrance"
    - id: kitchen_area
      name: "Kitchen"
    - id: garage_area
      name: "Garage"

# RF receiver configuration
remote_receiver:
  pin: GPIO4
  dump: rc_switch

# Each RF device appears as a separate device in HA
binary_sensor:
  - platform: remote_receiver
    name: "Front Door"
    device_id: front_door_device
    rc_switch_raw:
      code: '101010110101'
      
  - platform: remote_receiver
    name: "Kitchen Motion"
    device_id: kitchen_motion_device
    rc_switch_raw:
      code: '110011001100'
      
  - platform: remote_receiver
    name: "Garage Door"
    device_id: garage_door_device
    rc_switch_raw:
      code: '111000111000'
```

### Example 2: Multi-Zone Controller
```yaml
# ESP32 controlling multiple rooms
esphome:
  name: multi-room-controller
  
  devices:
    - id: living_room_device
      name: "Living Room Controller"
      area_id: living_area
    - id: kitchen_device
      name: "Kitchen Controller"
      area_id: kitchen_area
  
  areas:
    - id: living_area
      name: "Living Room"
    - id: kitchen_area
      name: "Kitchen"

sensor:
  - platform: dht
    pin: GPIO5
    temperature:
      name: "Temperature"
      device_id: living_room_device
    humidity:
      name: "Humidity"
      device_id: living_room_device

switch:
  - platform: gpio
    pin: GPIO10
    name: "Light"
    device_id: living_room_device
```

### Example 3: Area Configuration
```yaml
# Area configuration for the ESP device
esphome:
  name: my-device
  area:
    id: living_room
    name: "Living Room"

# Additional areas can be defined for sub-devices
  areas:
    - id: kitchen
      name: "Kitchen"
    - id: bedroom
      name: "Bedroom"

# Sub-devices can reference these areas
  devices:
    - id: kitchen_controller
      name: "Kitchen Controller"
      area_id: kitchen
```

### Example 4: Entity Name Validation
```yaml
# Duplicate entity names on same device/platform will fail validation
binary_sensor:
  - platform: template
    name: "Motion"
    lambda: return false;
    
  - platform: template
    name: "Motion"  # ERROR: Duplicate binary_sensor entity with name 'Motion' found
    lambda: return true;

# Duplicates across different platforms are allowed
sensor:
  - platform: template
    name: "Motion"  # OK - different platform
    lambda: return 0.0;

# Duplicates on different devices don't work (discovered during testing)
# This causes hash collisions in Home Assistant
binary_sensor:
  - platform: template
    name: "Status"
    device_id: device_1
    lambda: return false;
    
  - platform: template
    name: "Status"  # ERROR: Hash collision - won't work in HA
    device_id: device_2
    lambda: return true;
```

## Area Configuration Formats

ESPHome supports two formats for configuring areas:

### Simple String Format
```yaml
# Simple string format (still supported)
esphome:
  name: my-device
  area: "Kitchen"
```

### Structured Format
```yaml
# Structured format (recommended when using sub-devices)
esphome:
  name: my-device
  area:
    id: kitchen
    name: "Kitchen"
```

**Note**: Both formats are supported. The simple string format remains valid for basic use cases where you just want to assign the ESP device to a room. The structured format with `id` and `name` fields is recommended when you plan to use sub-devices, as the `id` field allows you to reference areas when creating sub-devices (using `area_id` in device configuration).

### Duplicate Entity Names Now Fail Validation

Previously, ESPHome would silently accept duplicate entity names within the same platform. This caused problems in Home Assistant where duplicate entities would not appear correctly and generate warnings about duplicate unique IDs.

**What's changed:**
- Duplicate entity names within the same platform now fail validation with a clear error message, even if they're on different devices
- The error tells you exactly which entity is duplicated
- This prevents the confusing situation where entities work in ESPHome but fail to appear correctly in Home Assistant

**Important Note:** During testing, it was discovered that entities with the same name on different devices cause hash collisions in Home Assistant, making them indistinguishable. While we could potentially add device_id to the EntityState messages to support this in the future, such a change would require weeks of planning and execution to avoid breaking existing Home Assistant installations. For now, we enforce unique names across all devices to prevent these issues.

**What you need to do:**
If your configuration has duplicate entity names, you'll see an error like:
```
Duplicate binary_sensor entity with name 'Motion' found. Each entity must have a unique name within its platform across all devices.
```

To fix this, give each entity a unique name:
```yaml
# Before (would previously be accepted, now fails):
binary_sensor:
  - platform: gpio
    name: "Motion"
    pin: GPIO5
  - platform: gpio
    name: "Motion"  # ERROR: Duplicate name
    pin: GPIO6

# After (fixed):
binary_sensor:
  - platform: gpio
    name: "Front Motion"
    pin: GPIO5
  - platform: gpio
    name: "Back Motion"
    pin: GPIO6
```

**Note:** This applies to all entities within the same platform, regardless of device. You can still have:
- Same names across different platforms (e.g., a `sensor.temperature` and `binary_sensor.temperature`)

What we discovered during testing:
- Same names on different devices within the same platform don't work due to hash collisions in Home Assistant
- While this might have appeared to work in ESPHome, it would fail in Home Assistant with duplicate ID errors
- We now detect and prevent this at validation time to avoid confusion

## Checklist:


nielsnl68:
Interesting idea. I will follow your progress.

dala318:
Marking as ready for review even if for now only 2 entity types are implemented, to make change properly visible for core maintainers. Once the overall concept is agreed/accepted I can bring it back to draft and continue with the rest of entity types.

> EDIT: Made a local PR with all entities that I could think should have this addition added in
> - dala318/esphome#10

dala318:
> Marking as ready for review even if for now only 2 entity types are implemented, to make change properly visible for core maintainers. Once the overall concept is agreed/accepted I can bring it back to draft and continue with the rest of entity types.

Merged my local PR for completing change for all entity types. So now all entity types that I think should get this addition are part of the change

bdraco:
One thing to keep in mind is that this does add some complexity in HA, since we’ll need to manage multiple devices — including handling add/remove/update when changes happen. We’ll also need to figure out how to model the parent/child relationships. That might not be too bad though, especially if all the sub-devices can just be tied to the main device.

nielsnl68:
@bdraco would this also help with the ideas i opted last month with you?
I was thinking it is.

bdraco:
> @bdraco would this also help with the ideas i opted last month with you? I was thinking it is.

You'll have to refresh my memory as that was 5000 context switches ago :)

dougiteixeira:
I have some considerations, just to contribute to the discussion:

* Adding subdevices: created when the entity is created/updated, as happens with most integrations.

* Removing devices and subdevices:
a) The main device and subdevices should be removed when deleting the configuration entry.
b) If you want to delete a subdevice (perhaps because it was removed from the code/firmware), you can implement the manual removal on the device page (something already common in Home Assistant), I think it is acceptable due to the DIY nature of ESPHome.

* The parent/child model, I believe it should be very simple: the parent device is the physical device and the subdevices are linked to the parent (via_device).

Although it adds an extra layer of complexity to the ESPHome code in Home Assistant, the way it is being implemented is simple for the end user and will bring versatility and organization to projects with ESPHome.

dala318:
Good to see some discussion, any ideas or improvents are welcomed.

About how to make use of this in Home Assistant and how these sub-devices are added/removed I hope will be possible to solve within the home assistant integration (a very draft PR linked in description). And there I hope to get some help from experienced on the matter.

Also, as I see it, the config of main and sub devices should be done in ESP Home, and HA integration just reports accordingly. (Suggested area might be the exception)

So this PR would focus on making the necessary information configurable and available over the ESP Home API. But planning for the HA integration is a key to that, so keep on having opinions now so that the correct thing is implemented.

nielsnl68:
> One thing to keep in mind is that this does add some complexity in HA, since we’ll need to manage multiple devices — including handling add/remove/update when changes happen. We’ll also need to figure out how to model the parent/child relationships. That might not be too bad though, especially if all the sub-devices can just be tied to the main device.

I like to use this in my ESPNOW project as well. And one of my troughs are that it should be possible that there are multiple parents directly (Via USB) or Ethernet/wifi to HA but being able to connect the same child's natively and based the RSSI signal it should handle one or the other parent.

I'm very aware that this could add more complexity to this setup, but i hope we could figure out if we can plan it a way so in the near future we do not need to rework everything to get the above implemented.

dala318:
There might be a need for two different implementations for solving what I understand as two different needs.

If I understand the ESPNOW related comment above correctly, that aims to identify another **ESP-device** that connects to HA via the ESPHome device as a device of its own. That to me more sounds like how Bluetooth devices that are connected vi the Bluetooth Proxy are identified in HA as connected via an ESPHome device. (How those can "jump" between different proxies in HA I'm not aware of). Those **devices** have a full status and identifier.

This change aims to group a selected set of **entities** in the ESPHome config as a device of its own. Handy if say one ESP serves several purposes. Physically one device, but logically more than one usage, ex. Temp and pressure for water circuit + Relay for fence gate. Such **sub-devices** have only the states and function related to the selected entities. The **main device** will be the one that reports the actual device status and get OTA updates etc.

But if possible to fuse the two use-cases to one implementation I'm all for it.

nielsnl68:
@dala318, you did understand me correctly.
And i understand your approach as well and like you see this as a good starting point.
As long you can make as open as possible to hook other ideas into the same infrastructure i'm already happy.
At a later point we can think of a way to implement something like the bluetooth proxy. 

But for me your work is essential for what i have in mind.

dala318:
> As long you can make as open as possible to hook other ideas into the same infrastructure i'm already happy.
> At a later point we can think of a way to implement something like the bluetooth proxy. 

I'm all for open an expandable implementation. So if you have ideas what might be needed for making this more compatible to your needs please make suggestions.

.github/workflows/ci-api-proto.yml
jesserockz, line 1: New PR if you want this in CI please

esphome/components/api/api.proto
```diff
@@ -186,6 +186,12 @@ message DeviceInfoRequest {
   // Empty
 }
 
+message SubDeviceInfo {
+  string id = 1;
```
jesserockz: Do we need a string here?
Can we get away with a `fixed32 id` and do all the linking in the code?

esphome/components/devices/__init__.py
jesserockz, line 1: I feel like this should go into `core/config.py` to live under the `esphome` key in configuration and not be a component

esphome/core/entity_base.h
```diff
@@ -61,6 +65,7 @@ class EntityBase {
   bool internal_{false};
   bool disabled_by_default_{false};
   EntityCategory entity_category_{ENTITY_CATEGORY_NONE};
+  StringRef device_id_{""};
```
jesserockz: This really should just be a pointer to the SubDevice object

tests/dashboard/test_web_server.py
jesserockz, line 1: Separate PR please

jesserockz (changes requested):

nielsnl68:
> > As long you can make as open as possible to hook other ideas into the same infrastructure i'm already happy.
> > At a later point we can think of a way to implement something like the bluetooth proxy.
> 
> I'm all for open an expandable implementation. So if you have ideas what might be needed for making this more compatible to your needs please make suggestions.

For now, please make it bidirectional. And i think everything is fine.

.github/workflows/ci-api-proto.yml
dala318: Oh, yes agree.
I added it only to get the expected result of the codegen since I never got the environment properly set-up locally to generate the files. Planning to remove it once the rest is approved.
Maybe a more proper addition in another PR would be to archive the generated artifacts so that one could download and commit them, avoiding the hazzle to copy-paste git-diffs.

esphome/components/api/api.proto
dala318: Yes, seem like a more robust solution.

esphome/components/devices/__init__.py
dala318: To do this I think I would need some more guidance from the more experienced participants.
I added it as a separate component to make it fully optional and easier to scale and adapt in the future. (And not interfere to much with the core config.)
What are the benefits or rationale of fusing into the `esphome` config part?
That part holds the "main" device information, and yes, this is a partial clone to that and the sub-device data goes into the same API data as most of what's in that config-part.
I really don't have any objections to moving this to `core/config.py`, just want to make sure it's the agreed way forward.

tests/dashboard/test_web_server.py
dala318: I was hoping to revert this part, got some intermittent test crashes in CI.
Can try to remove, if it keep happening I will push it as a separate PR

esphome/components/api/api_connection.cpp
```diff
@@ -337,6 +338,7 @@ bool APIConnection::try_send_cover_info(APIConnection *api, void *v_cover) {
   msg.disabled_by_default = cover->is_disabled_by_default();
   msg.icon = cover->get_icon();
   msg.entity_category = static_cast<enums::EntityCategory>(cover->get_entity_category());
+  msg.device_id = cover->get_device_id();
   return api->send_list_entities_cover_response(msg);
 }
```
dala318: Seem to have forgotten a whole bunch of entity types, will fix.

esphome/components/api/api.proto
dala318: Changed into uint32

esphome/core/entity_base.h
dala318: Is now an uint32_t

.github/workflows/ci-api-proto.yml
dala318: Created https://github.com/esphome/esphome/pull/8698

tests/dashboard/test_web_server.py
dala318: Added PR https://github.com/esphome/esphome/pull/8699
Change reverted on this branch

esphome/config_validation.py
```diff
@@ -347,6 +348,13 @@ def icon(value):
     )
 
 
+def sub_device_id(value):
+    # Duplicate definition of SubDevice to avoid circular import
+    SubDevice = cg.esphome_ns.class_("SubDevice")
```
dala318: I don't like having to define the SubDevice class twice (here and in config.py) but can't find a nice way to avoid circular dependencies.
Could work if it was defined "publicly" in this file and imported to core.py, but kind of think it don't belong to `config_validation.py`.
An alternative could be to move declaration to `core/__init__.py` but that seem a bit drastic.

esphome/components/devices/__init__.py
dala318: I've made an attempt in removing the `components/device/` stuff and put it in `core/config.py` and a new `core/sub_devices.h`
Please review if it's how you expected.

esphome/config_validation.py
dala318: CoPilot proposed using lazy imports so will give that an attempt (and see if the clang-tidy etc. agrees to doing it so)

esphome/core/config.py
```diff
@@ -405,3 +416,12 @@ async def to_code(config):
 
     if config[CONF_PLATFORMIO_OPTIONS]:
         CORE.add_job(_add_platformio_options, config[CONF_PLATFORMIO_OPTIONS])
+
+    if config[CONF_SUB_DEVICES]:
+        for dev_conf in config[CONF_SUB_DEVICES]:
+            dev = cg.new_Pvariable(dev_conf[CONF_ID])
+            cg.add(dev.set_uid(hash(str(dev_conf[CONF_ID])) % 0xFFFFFFFF))
```
bdraco: The UID generation using hash(str(device)) % 0xFFFFFFFF could cause collisions

bdraco: How do we validate Sub-device IDs are unique within a configuration

esphome/core/config.py
```diff
@@ -405,3 +416,12 @@ async def to_code(config):
...
+            cg.add(dev.set_name(dev_conf[CONF_NAME]))
+            cg.add(dev.set_area(dev_conf[CONF_AREA]))
```
bdraco: It seems like devices could be in the same area, however they are only set by name

esphome/components/api/api.proto
```diff
@@ -188,6 +188,12 @@ message DeviceInfoRequest {
   // Empty
 }
 
+message SubDeviceInfo {
+  uint32 uid = 1;
+  string name = 2;
+  string suggested_area = 3;
```
bdraco: Area seems like it should have its own id since multiple devices can be in the same area

esphome/core/sub_device.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include "esphome/core/string_ref.h"
+
+namespace esphome {
+
+class SubDevice {
+ public:
+  void set_uid(uint32_t uid) { uid_ = uid; }
+  uint32_t get_uid() { return uid_; }
+  void set_name(std::string name) { name_ = std::move(name); }
```
bdraco: I wonder if StringRef makes sense here

esphome/core/config.py
dala318: Do you have any proposals?
I can concatenate name as well to add more variation.
Could generate just some random id, but thinking to use device id was to use something that generates the same uid every compilation (unless you change id of course)

dala318: Will not the schema validation solve that?

cv.GenerateID(CONF_ID): cv.declare_id(SubDevice),

esphome/core/config.py
dala318: As far as I have understood Home Assistant this thing will only be used at device creation the first time to propose an Area for the device. If the user selects another area it will be stored and used thereafter.

esphome/core/sub_device.h
dala318: I did use StringRef for a while in earlier versions but it only made me do more conversions.
If there are benefits of still using it I can change back.

One more reason I changed to std::string was also that when in a component referencing original StringRef definition in Python it created circular dependencies.

esphome/components/api/api.proto
dala318: Could be done but in my opinion that would only complicate things. Fairly easy to just duplicate strings, or use substitutions in yaml.

I suppose the sub-devices could inherit area from main config as default if given there.

esphome/core/config.py
jesserockz: Yes declare id will ensure uniqueness

esphome/components/api/api.proto
bdraco: I'm thinking down the road that the ESP might end up being a hub for devices in the future in different areas.  The area implementation in HA may be improved in the future to better deal with overlapping names so it would be good to have a stable id to be able to join multiple areas

nielsnl68: Yes, @bdraco i like to use this PR idea to extend the futures of my espnow component indeed.

esphome/core/config.py
bdraco: we use fnv1a_32 for cases like this as the collision risk is much lower

nielsnl68:
> > > As long you can make as open as possible to hook other ideas into the same infrastructure i'm already happy.
> > > At a later point we can think of a way to implement something like the bluetooth proxy.
> > 
> > 
> > I'm all for open an expandable implementation. So if you have ideas what might be needed for making this more compatible to your needs please make suggestions.
> 
> For now, please make it bidirectional. And i think everything is fine.

One more suggestion is to have the option to allow adding new sub-devices dynamically when a new ESPNOW device connect's to the receiver device.
For me this is just a future wish not something to have now. But when you can have this in mind while working on this i would be very happy. I have no clue about the possibility of my suggestion.

bdraco:
> > > > As long you can make as open as possible to hook other ideas into the same infrastructure i'm already happy.
> > > > At a later point we can think of a way to implement something like the bluetooth proxy.
> > > 
> > > 
> > > I'm all for open an expandable implementation. So if you have ideas what might be needed for making this more compatible to your needs please make suggestions.
> > 
> > 
> > For now, please make it bidirectional. And i think everything is fine.
> 
> One more suggestion is to have the option to allow adding new sub-devices dynamically when a new ESPNOW device connect's to the receiver device. For me this is just a future wish not something to have now. But when you can have this in mind while working on this i would be very happy. I have no clue about the possibility of my suggestion.

I think its important to make sure we can do that before moving this forward as it could be very hard to add later without breaking something with the current design.

esphome/core/config.py
bdraco: esphome/core/helpers.h:uint32_t fnv1_hash(const std::string &str); is likely good enough for this

bdraco: But this is happening in python ... hmm

bdraco: https://github.com/Bluetooth-Devices/fnv-hash-fast/blob/main/src/fnv_hash_fast/fnv_wrapper.h

bdraco: `hash()` also isn't stable across python versions

bdraco:
```diff
diff --git a/esphome/core/config.py b/esphome/core/config.py
index f3d8b7e71..f4a31c535 100644
--- a/esphome/core/config.py
+++ b/esphome/core/config.py
@@ -38,6 +38,15 @@ from esphome.helpers import copy_file_if_changed, get_str_env, walk_files
 
 _LOGGER = logging.getLogger(__name__)
 
+
+def fnv1a_32bit_hash(string: str) -> int:
+    """FNV-1a 32-bit hash function."""
+    hash_value = 2166136261
+    for char in string:
+        hash_value ^= ord(char)
+        hash_value = (hash_value * 16777619) & 0xFFFFFFFF
+    return hash_value
+
 StartupTrigger = cg.esphome_ns.class_(
     "StartupTrigger", cg.Component, automation.Trigger.template()
 )
@@ -420,7 +429,7 @@ async def to_code(config):
     if config[CONF_SUB_DEVICES]:
         for dev_conf in config[CONF_SUB_DEVICES]:
             dev = cg.new_Pvariable(dev_conf[CONF_ID])
-            cg.add(dev.set_uid(hash(str(dev_conf[CONF_ID])) % 0xFFFFFFFF))
+            cg.add(dev.set_uid(fnv1a_32bit_hash(str(dev_conf[CONF_ID]))))
             cg.add(dev.set_name(dev_conf[CONF_NAME]))
             cg.add(dev.set_area(dev_conf[CONF_AREA]))
             cg.add(cg.App.register_sub_device(dev))
```

dala318:
> > > > > As long you can make as open as possible to hook other ideas into the same infrastructure i'm already happy.
> > > > > At a later point we can think of a way to implement something like the bluetooth proxy.
> > > > 
> > > > 
> > > > I'm all for open an expandable implementation. So if you have ideas what might be needed for making this more compatible to your needs please make suggestions.
> > > 
> > > 
> > > For now, please make it bidirectional. And i think everything is fine.
> > 
> > One more suggestion is to have the option to allow adding new sub-devices dynamically when a new ESPNOW device connect's to the receiver device. For me this is just a future wish not something to have now. But when you can have this in mind while working on this i would be very happy. I have no clue about the possibility of my suggestion.
> 
> I think its important to make sure we can do that before moving this forward as it could be very hard to add later without breaking something with the current design.

So should the sub-devices (and others in the future) be kept in a dynamic array?

Next possible issue down the road if so, the sub-devices are currently sent over via the DeviceInfoResponse proto package. Is that sent cyclic?

If thinking reusable design it sound for me more that other connected devices when the ESPHome instance act as gateway should extend on how the BluetoothProxy solves the same.

If it all could be one solution I'm all OK with that but don't have any clear proposal on how to do it or any of the alternatives.

nielsnl68:
> So should the sub-devices (and others in the future) be kept in a dynamic array?

I think that could be the way to go. But check how the BluetoothProxy does it do.

> If thinking reusable design it sound for me more that other connected devices when the ESPHome instance act as gateway should extend on how the BluetoothProxy solves the same.

I am just a side-liner in this and have no idea of it is easy to implement it. That said i think that you should use the same technique indeed.

> If it all could be one solution I'm all OK with that.

I agree with this. 

> but don't have any clear proposal on how to do it or any of the alternatives.

Maybe have a check how the BluetoothProxy it could result in a proposal.

dala318:
> > So should the sub-devices (and others in the future) be kept in a dynamic array?
> 
> I think that could be the way to go. But check how the BluetoothProxy does it do.
> 
This is a bit outside my comfort zone to make statements about, but Bluetooth Proxy seem to have its own proto messages for communication with Home Assistant `BluetoothDeviceRequest` and `BluetoothDeviceConnectionResponse`.
But neither of these seem to have either `name` or `suggested_area` as payload.

So for this to work as some type of coordinating device registry between Virtual Sub Devices and Physical Connected Devices (Bluetooth, ESPNow etc.) some other more advanced solution is needed. And I would need some help from others to propose solutions.

esphome/components/api/api.proto
dala318: So should I add some core component `area` which both the main esphome config and sub_devices config reference? Or is it a sub-section of the esphome config?

esphome/core/config.py
dala318: Added function `fnv1a_32bit_hash` to helpers.py since it needs to be used in at least two places.

nielsnl68:
I have created a thread in the #development-projects (https://discord.com/channels/429907082951524364/1379102816886329375) Here can we easier brainstorm about the questions.

martijnbuts:
Interesting pr, is this also related to the ha roadmap that have some information about device context?

esphome/core/config.py
```diff
@@ -417,3 +433,12 @@ async def to_code(config):
...
     if config[CONF_PLATFORMIO_OPTIONS]:
         CORE.add_job(_add_platformio_options, config[CONF_PLATFORMIO_OPTIONS])
+
+    if config[CONF_SUB_DEVICES]:
+        for dev_conf in config[CONF_SUB_DEVICES]:
+            dev = cg.new_Pvariable(dev_conf[CONF_ID])
+            cg.add(dev.set_uid(fnv1a_32bit_hash(str(dev_conf[CONF_ID]))))
+            cg.add(dev.set_name(dev_conf[CONF_NAME]))
+            cg.add(dev.set_area(dev_conf[CONF_AREA]))
+            cg.add(cg.App.register_sub_device(dev))
```
bdraco: We could pre-reserve this vector for ones that are known at build time

bdraco:
I had some time because of a delayed flight to look at this and made some changes to try to address my review comments. I'm not sure I got the hashing right for the area id, but out of time before boarding so I pushed the changes to help get this PR moving. I'll take a look again when I have some more free cycles

esphome/core/sub_area.h
```diff
@@ -0,0 +1,20 @@
...
+#include <string>
+#include <cstdint>
+
+namespace esphome {
+
+class SubArea {
+ public:
+  void set_area_id(uint32_t area_id) { area_id_ = area_id; }
+  uint32_t get_area_id() { return area_id_; }
+  void set_name(std::string name) { name_ = std::move(name); }
```
bdraco: We might be able to get way with const char * here if we always define these at build time and save a few bytes of RAM

esphome/core/sub_device.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include "esphome/core/string_ref.h"
+
+namespace esphome {
+
+class SubDevice {
+ public:
+  void set_device_id(uint32_t device_id) { device_id_ = device_id; }
+  uint32_t get_device_id() { return device_id_; }
+  void set_name(std::string name) { name_ = std::move(name); }
```
bdraco: We might be able to get way with const char * here if we always define these at build time and save a few bytes of RAM

bdraco:
@nielsnl68 Did you expect to be able to register ESPNOW sub-devices late or before HA connects? 

Ideally we know at build time so we can reduce the memory footprint.

bdraco:
@nielsnl68 

I guess we have to register them before HA connects or we have no way to send them to HA unless we completely redesign this (which would be OK, but we need to understand the requirement now before merging this PR)

bdraco:
Also I was able to significantly reduce the changes in `api_connection.*`

tests/components/esphome/common.yaml
```diff
@@ -17,10 +17,13 @@ esphome:
     version: "1.1"
     on_update:
       logger.log: on_update
+  sub_areas:
+    - id: another_area
+      name: Another area
```
dala318: If going this path, should then not also the normal `area` setting above use the same logic?

dala318: I do understand and agree to this approach way of keeping data structured and avoid sending possible identical strings several times, but config wise it not as intuitive or requires several lines of config to just get a string passed to backend.
At least propose to rename `sub_areas` to `areas` or `sub_device_areas`, sub-area implies it a part of some other area.

bdraco:
added missing device_id to ListEntitiesSirenResponse which should fix the error

tests/components/esphome/common.yaml
bdraco: If we change the existing suggested area we end up with a breaking change...

bdraco: I guess we could rename `SubAreaInfo` have a top level `AreaInfo`  and migrate over time so we get rid of `suggested_area` eventually but leave if for backwards compat.

bdraco: I think that works. Let me implement that

bdraco:
We are going to need some integration tests to handle cases

- same area as top level device
- different area as top level device
- no area

But we will need to do an aioesphomeapi release first with the protobuf changes

tests/components/esphome/common.yaml
bdraco: Implemented that. Made it DeviceInfo to be consistent

esphome/core/config.py
bdraco: done

esphome/core/sub_area.h
bdraco: done

esphome/core/sub_device.h
bdraco: done

bdraco:
@nielsnl68 I think you should be able to handle this in your ESPNOW PR by defining a `device_id` for each peer which maps to the configured device here.

We will probably need to add something in the python code to be able to register them dynamically from other components, but thats something that can be added later.

bdraco:
I'm pretty happy with this now. I'd like to get a 👍 from another reviewer on the protobuf structure before we update aioesphomeapi.  Once we have that, I can take care of getting a release done for that and than we can write some integration tests to make sure the API works how we want it.

bdraco:
### File tests/dummy_main.cpp will need some fixes

bdraco:
Will try to look again tomorrow as it's getting late here and have another flight tomorrow

nielsnl68:
> @nielsnl68 Did you expect to be able to register ESPNOW sub-devices late or before HA connects?
> 
> Ideally we know at build time so we can reduce the memory footprint.

In a perfect world i would love to see a late Ha connection, at the moment that a new espnow device connects to the host device. That said, to begin with i would not be mad when it is a before Ha connection for now.

bdraco:
> > @nielsnl68 Did you expect to be able to register ESPNOW sub-devices late or before HA connects?
> 
> > 
> 
> > Ideally we know at build time so we can reduce the memory footprint.
> 
> 
> 
> In a perfect world i would love to see a late Ha connection, at the moment that a new espnow device connects to the host device. That said, to begin with i would not be mad when it is a before Ha connection for now. 

If we need to send something later than upon connection, we need to redesign nearly the entire entity system and this PR as it's currently being sent in DeviceInfo which only happens when HA first connections. Right now it's the same for entity info and initial state, they only get sent on HA first connection, and there is no way to add new entities after the connection is established to HA.  

In theory we could send DeviceInfo and redo the entity info send for all the entities and that would require minimal changes on the HA side, however the webserver and mqtt would need to have a similar redesign to be able to do it after the api connection has already been set up

nielsnl68:
> If we need to send something later than upon connection, we need to redesign nearly the entire entity system and this PR as it's currently being sent in DeviceInfo which only happens when HA first connections. Right now it's the same for entity info and initial state, they only get sent on HA first connection, and there is no way to add new entities after the connection is established to HA.

I have no issue with this. I'm already happy that this is added for now and that you both considered to make it possible to use this together with my espnow component.

nielsnl68:
... and we know now what we need to do to make the changes to allow late connects. 
That is everything i was hopping for.

esphome/core/config.py
```diff
@@ -417,3 +455,70 @@ async def to_code(config):
...
+            _LOGGER.warning(
+                "Using 'area' as a string is deprecated. Please use the new format:\n"
+                "area:\n"
+                "  id: %s\n"
+                '  name: "%s"',
+                area_slug,
+                area_conf,
+            )
+            # Create a synthetic area for backwards compatibility
+            area_var = cg.Pvariable(f"area_{area_slug}", Area)
```
bdraco: We can just use the slug here

esphome/core/config.py
```diff
@@ -417,3 +455,70 @@ async def to_code(config):
...
+            area_id = fnv1a_32bit_hash(area_conf)
```
bdraco: We need to raise on hash collision. It should be extremely rare but will be very confusing if it does happen

esphome/core/config.py
```diff
@@ -417,3 +455,70 @@ async def to_code(config):
...
+        # Reserve space for devices
+        cg.add(cg.RawStatement(f"App.reserve_device({len(devices)});"))
+        cg.add_define("USE_DEVICES")
+
+        # Process additional areas
+        areas: list[dict[str, str]]
+        if areas := config[CONF_AREAS]:
+            for area_conf in areas:
+                area = cg.new_Pvariable(area_conf[CONF_ID])
+                area_id = fnv1a_32bit_hash(str(area_conf[CONF_ID]))
```
bdraco: We need to raise on hash collision. It should be extremely rare but will be very confusing if it does happen

esphome/core/config.py
```diff
@@ -417,3 +455,70 @@ async def to_code(config):
...
+                cg.add(area.set_area_id(area_id))
+                cg.add(area.set_name(area_conf[CONF_NAME]))
+                cg.add(cg.App.register_area(area))
+
+        # Process devices
+        for dev_conf in devices:
+            dev = cg.new_Pvariable(dev_conf[CONF_ID])
+            cg.add(dev.set_device_id(fnv1a_32bit_hash(str(dev_conf[CONF_ID]))))
```
bdraco: We need to raise on hash collision. It should be extremely rare but will be very confusing if it does happen

bdraco:
bump https://github.com/esphome/esphome/pull/9173

bdraco:
We are going to need end to end integration tests for this change as well once we have aioesphomeapi updated

bdraco (dismissed):

bdraco:
Might need a little more cleanup but I think the concept is good at this point and well covered with the integration tests

bdraco (dismissed):

bdraco:
I added all the tests for this as well.  It should be ready for another reviewer at this point.

bdraco (dismissed):

bdraco:
I updated the opening text of the PR to better describe the use cases here

bdraco (dismissed):

jesserockz:
Can be merged after HA core code is ready and fully tested

bdraco:
HA core PR is ready and tested https://github.com/home-assistant/core/pull/147343

nielsnl68:
> HA core PR is ready and tested https://github.com/home-assistant/core/pull/147343

Is there an easy way to test it? I have one device (shelly 2plus) that is being used for 2 rooms which can benefit with this set of PR's.

bdraco:
cd /config
curl -o- -sSL https://gist.githubusercontent.com/bdraco/43f8043cb04b9838383fd71353e99b18/raw/core_integration_pr | bash /dev/stdin -d esphome -p 147343


You can install HA PR as a custom component and restart. It might work on stable HA if there aren't too many changes. You'll also need to copy translations/XXX.json to /config/custom_components/esphome/translations from /usr/src/homeassistant/components/esphome/translations/

esphome/core/entity_base.cpp
```diff
@@ -29,16 +36,21 @@ void EntityBase::set_icon(const char *icon) { this->icon_c_str_ = icon; }
 
 // Entity Object ID
 std::string EntityBase::get_object_id() const {
+  std::string suffix = "";
+#ifdef USE_DEVICES
+  if (this->device_ != nullptr) {
+    suffix = "@" + str_sanitize(str_snake_case(this->device_->get_name()));
```
bdraco: I think we need to use device_id here so we known if its still valid or not and we have a reference point

bdraco: I'm not sure this is going to work.  It gets super messy to move the entities between devices
```diff
diff --git a/homeassistant/components/esphome/entity.py b/homeassistant/components/esphome/entity.py
index fcf1260de39..af43d5dae4f 100644
--- a/homeassistant/components/esphome/entity.py
+++ b/homeassistant/components/esphome/entity.py
@@ -68,17 +68,62 @@ def async_static_info_updated(
 
         # Create new entity if it doesn't exist
         if not (old_info := current_infos.pop(info.key, None)):
-            entity = entity_type(entry_data, platform.domain, info, state_type)
-            add_entities.append(entity)
+            # Check if this entity might have moved from another device (object_id changed)
+            # Try to find an entity with a different object_id but same base properties
+            new_unique_id = build_unique_id(device_info.mac_address, info)
+            
+            # Look for entities with different object_id patterns (e.g., temperature@old_device)
+            # that might have moved to this device
+            found_migration = False
+            if "@" in info.object_id:
+                # Extract base object_id without device suffix
+                base_object_id = info.object_id.split("@")[0]
+                
+                # Search through all entities of this platform to find potential matches
+                for entity_entry in ent_reg.entities.get_entries_for_domain(platform.domain):
+                    if (entity_entry.platform == DOMAIN and 
+                        entity_entry.unique_id and 
+                        entity_entry.unique_id != new_unique_id):
+                        # Check if this could be the same entity that moved devices
+                        # This is a heuristic - we look for entities with similar patterns
+                        try:
+                            # Parse the unique_id to check if it's the same base entity
+                            # that just moved to a different device
+                            old_entity_id = entity_entry.entity_id
+                            if base_object_id in old_entity_id:
+                                # Found a potential match - migrate it
+                                ent_reg.async_update_entity(
+                                    entity_entry.entity_id,
+                                    new_unique_id=new_unique_id
+                                )
+                                found_migration = True
+                                break
+                        except Exception:
+                            continue
+            
+            if not found_migration:
+                # No migration found, create new entity
+                entity = entity_type(entry_data, platform.domain, info, state_type)
+                add_entities.append(entity)
             continue
 
-        # Entity exists - check if device_id has changed
+        # Entity exists - check if device_id or object_id has changed
+        old_unique_id = build_unique_id(device_info.mac_address, old_info)
+        new_unique_id = build_unique_id(device_info.mac_address, info)
+        
+        if old_unique_id != new_unique_id:
+            # Unique ID changed (likely due to object_id change when moving devices)
+            entity_id = ent_reg.async_get_entity_id(platform.domain, DOMAIN, old_unique_id)
+            if entity_id:
+                # Update the entity's unique_id to the new one
+                ent_reg.async_update_entity(entity_id, new_unique_id=new_unique_id)
+        
+        # Check if device assignment changed
         if old_info.device_id == info.device_id:
             continue
 
         # Entity has switched devices, update its device assignment
-        unique_id = build_unique_id(device_info.mac_address, info)
-        entity_id = ent_reg.async_get_entity_id(platform.domain, DOMAIN, unique_id)
+        entity_id = ent_reg.async_get_entity_id(platform.domain, DOMAIN, new_unique_id)
         # entity_id should never be None here because old_info is not None,
         # which means the entity was previously created and is in the registry
         assert entity_id is not None
```

bdraco: So I think we don't need to add it to the object id because we already have the device_id

esphome/core/entity_base.cpp
```diff
@@ -33,8 +40,7 @@ std::string EntityBase::get_object_id() const {
   if (!this->flags_.has_own_name && App.is_name_add_mac_suffix_enabled()) {
     // `App.get_friendly_name()` is dynamic.
     return str_sanitize(str_snake_case(App.get_friendly_name()));
-  } else {
-    // `App.get_friendly_name()` is constant.
+  } else {  // `App.get_friendly_name()` is constant.
```
bdraco:
```suggestion
  } else {
    // `App.get_friendly_name()` is constant.
```

esphome/core/application.h
```diff
@@ -102,11 +109,18 @@ class Application {
       this->name_ = name;
       this->friendly_name_ = friendly_name;
     }
-    this->area_ = area;
+    // area is now handled through the areas system
```
bdraco:
```suggestion
```

bdraco:
I couldn't get away with not fixing https://github.com/esphome/issues/issues/6953 now that we could have multiple unnamed entities with different device_ids so I took care of it. My first thought was to disallow it but it's still a valid use case especially if you aren't using native API so I used the same strategy we used for collisions in core for entity_id, append _2,_3 etc. It's not so great in that if change the order they could get swapped, but that's still better than collisions or forcing a breaking change.

bdraco:
Jesse and I discussed on discord and we decided to make it fail validation if there are multiple (device_id, platform, name) since it would be confusing to allow the same name on a device.  This replaces the appending _2,_3,etc.  This will be a breaking change though since we currently allow it

bdraco:
now validates duplicates
```
esphome compile ol.yaml 
INFO ESPHome 2025.7.0-dev
INFO Reading configuration ol.yaml...
INFO Detected timezone 'Europe/Madrid'
Failed config

sensor.template: [source ol.yaml:719]
  
  Duplicate sensor entity with name 'Lambda Sensor 199' found. Each entity on a device must have a unique name within its platform.
  platform: template
  name: Lambda Sensor 199
  lambda: return 199.0;
```

bdraco:
## Breaking Changes

### Duplicate Entity Names Now Fail Validation

Previously, ESPHome would silently accept duplicate entity names within the same platform and device. This caused problems in Home Assistant where the second entity would not appear and generate warnings in the logs about duplicate unique IDs.

**What's changed:**
- Duplicate entity names on the same device and platform now fail validation with a clear error message
- The error tells you exactly which entity is duplicated and where
- This prevents the confusing situation where entities work in ESPHome but fail to appear in Home Assistant

**What you need to do:**
If your configuration has duplicate entity names, you'll see an error like:
```
Duplicate binary_sensor entity with name 'Motion' found. Each entity on a device must have a unique name within its platform.
```

To fix this, give each entity a unique name:
```yaml
# Before (would previously be accepted, now fails):
binary_sensor:
  - platform: gpio
    name: "Motion"
    pin: GPIO5
  - platform: gpio
    name: "Motion"  # ERROR: Duplicate name
    pin: GPIO6

# After (fixed):
binary_sensor:
  - platform: gpio
    name: "Front Motion"
    pin: GPIO5
  - platform: gpio
    name: "Back Motion"
    pin: GPIO6
```

**Note:** This only applies to entities on the same device and platform. You can still have:
- Same names across different platforms (e.g., a `sensor.temperature` and `binary_sensor.temperature`)
- Same names on different devices (using the new `device_id` feature)

bdraco:
1. alarm_control_panel - Fixed duplicate "Alarm Panel" names by adding "2" suffix
  2. ade7880 - Fixed duplicate sensor names by adding channel prefixes (Channel A, B, C)
  3. binary_sensor_map - Fixed duplicate "Binary Sensor Map" names by adding type suffixes (Group, Sum, Bayesian)
  4. dallas_temp - Fixed duplicate "Dallas Temperature" names by adding numbers (1, 2)
  5. demo - Fixed duplicate "Demo Temperature Sensor" names by adding numbers (1, 2)
  6. heatpumpir - Fixed duplicate "HeatpumpIR Climate" names by adding protocol names (Mitsubishi, Daikin, Panasonic)
  7. light - Fixed duplicate "RGB Light" names by adding "Initial State" suffix
  8. ltr390 - Fixed duplicate sensor names by adding numbers (1, 2) to all sensors
  9. lvgl - Fixed duplicate names between sensor and number platforms by adding platform suffixes (Sensor, Number)
  10. opentherm - Fixed duplicate "Boiler Central Heating 2 active" by changing one to "status"
  11. remote_transmitter - Fixed duplicate "RC5" button names by adding "Raw" suffix
  12. packages - Updated test to use absolute_humidity instead of binary_sensor_map from GitHub

bdraco (dismissed):

bdraco (approved):

jesserockz (approved):

Drafteed:
@bdraco @dala318 Thank you, you did a great job! 🎉

I've been waiting for this feature in ESPHome for a long time. Now I can finally separate my 5 virtual devices with a bunch of different entities on one ESP into separate devices. Thank you 👍. I'll be happy to test this feature during beta testing.

martijnbuts:
Hi, i have a ESP32-S3 that controls: monobloc heat pump that is outside and a thermostat control that is inside.  
i added the following code: 
  devices:
    - id: warmtepomp
      name: "Warmtepomp"
      area_id: buiten
    - id: thermostaat
      name: "Thermostaat"
      area_id: woonkamer

  areas:
    - id: buiten
      name: "Buiten"
    - id: woonkamer
      name: "woonkamer"

climate:
  - platform: thermostat
    name: "Thermostat Climate Controller"
    web_server:
      sorting_group_id: Algemeen
    sensor: Inside_temperature
    device_id: thermostaat
left rest of code out 

  - platform: modbus_spy
    name: "Heat exchanger water inlet temperature"
    id: heat_exchanger_water_inlet_temperature
    web_server:
      sorting_group_id: WP_monoblok
    device_class: "temperature"
    accuracy_decimals: 1
    state_class: "measurement"
    unit_of_measurement: "°C"
    device_id: warmtepomp
left rest of code out 

 

It compiles good but how did you validate that it works as 2 devices?
i did expect to see the thermostat in the area livingroom or in Dutch woonkamer

Is it correct that the esphome webserver does not support the device id and i should use the sorting group?

Drafteed:
@bdraco After testing i got some things:

**First thing:**

```yaml
esphome:
  name: esphome-hub
  devices:
    - id: ringbell_device
      name: "Ringbell"
    - id: intercom_device
      name: "Intercom"
    - id: le2_device
      name: "Energy Meter"
    - id: tion_device
      name: "TION Breezer"
    - id: xgimi_device
      name: "XGIMI Projector (BLE)"

climate:
  # Expect that device name will be used
  - platform: tion
    id: tion_climate
    device_id: tion_device

fan:
  # Expect that device name will be used
  - platform: tion
    id: tion_fan
    device_id: tion_device
```

After adding the devices to the HA, these two entities will be missing. Adding `name: ""` to each solves this.

**Second thing:**

```yaml
binary_sensor:
  # Ringbell device
  - platform: gpio
    id: ringbell_call_binary
    pin:
      number: GPIO4
      mode:
        input: true
        pulldown: true
    name: "Call"
    device_id: ringbell_device
    device_class: sound

  # Intercom device
  - platform: gpio
    id: intercom_call_binary
    pin:
      number: GPIO5
      inverted: true
      mode:
        input: true
        pullup: true
    filters:
      - delayed_on: 40ms
      - delayed_off: 4000ms
    name: "Call"
    device_id: intercom_device
    device_class: sound
```

Two different virtual devices are using the same name. After adding them to HA, all devices added successfully. `binary_sensor.intercom_call` shows as `unknown`.

Logs:
```
Error doing job: Task exception was never retrieved (None)
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 1470, in async_init
    subscription("added", flow.flow_id)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/config/config_entries.py", line 398, in async_on_flow_init_remove
    "flow": hass.config_entries.flow.async_get(flow_id),
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/data_entry_flow.py", line 234, in async_get
    raise UnknownFlow
homeassistant.data_entry_flow.UnknownFlow

Error doing job: Task exception was never retrieved (None)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1535, in _async_process_registry_update_or_remove
    self.async_write_ha_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1018, in async_write_ha_state
    self._async_write_ha_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1140, in _async_write_ha_state
    self.__async_calculate_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1080, in __async_calculate_state
    state = self._stringify_state(available)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1024, in _stringify_state
    if (state := self.state) is None:
                 ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 546, in state
    value = self.native_value
            ^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/esphome/sensor.py", line 102, in native_value
    if not self._has_state or (state := self._state).missing_state:
           ^^^^^^^^^^^^^^^
AttributeError: 'EsphomeSensor' object has no attribute '_has_state'. Did you mean: '__attr_state'?
```

After rebooting HA, all sensors (except buttons) become unavailable. No errors in the logs.

When I reboot the ESP device, all sensors are restored, except `binary_sensor.intercom_call`:
```
Platform esphome does not generate unique IDs. ID 9C:9C:1F:C6:49:50-binary_sensor-call@1130295909 is already used by binary_sensor.intercom_call - ignoring binary_sensor.Intercom_Call
```

If I reboot HA again, all sensors will become unavailable again.

To fix this I had to rename the entities in esp and manually cleanup `deleted_entities` from `core.entity_registry` due to https://github.com/home-assistant/core/pull/145278 and https://github.com/home-assistant/core/pull/145191

However, the error when adding a device still exists (not sure is related).

Drafteed:
> It compiles good but how did you validate that it works as 2 devices?
> i did expect to see the thermostat in the area livingroom or in Dutch woonkamer

@martijnbuts This only works in beta version of HA 2025.7.0.

dala318:
@Drafteed 
> After adding the devices to the HA, these two entities will be missing. Adding name: "" to each solves this.

Without a 'name' to the entity they will be configured as 'internal: true'. That's why you don't see them.

martijnbuts:
> > It compiles good but how did you validate that it works as 2 devices?
> > i did expect to see the thermostat in the area livingroom or in Dutch woonkamer
> 
> @martijnbuts This only works in beta version of HA 2025.7.0.

i do use the dev build from this morning. I also think that with older versions i did get a compile error. 
i am sure i use the dev build from this morning because i found a duplicate entity i had to remove to be able to create the esp32-32 firmware.

martijnbuts:
I found the problem, for some unknown reason my dev version did compile but for some reason use a old config. I did find it when i was updating web server sorting group and it did not change on the website.

PR #8561, opened 2025-04-14
New I2C sensor XDB401

dala318:

Adds support for I2C pressure sensor XDB 401
https://www.aliexpress.com/item/1005007148779919.html

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4821
## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [x] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
sensor:
  - platform: xdb401
    i2c_id: i2c_bus
    update_interval: 20s
    pressure:
      name: "Pressure"
    temperature:
      name: Temperature
```

## Checklist:


esphome/components/xdb401/xdb401.cpp
```diff
@@ -0,0 +1,179 @@
...
+static const float CONVERT_PRESSURE = (float) 0x800000;
+static const float SCALE_PRESSURE = 1.0e6;
+static const float SCALE_TEMPERATURE = 100.0;
+
+static const int CHECK_DELAY = 5u;
+static const int CHECK_ATTEMPTS = 6u;
+static const int MARK_FAIL_AFTER = 5;
+
+void XDB401Component::setup() {
+  ESP_LOGCONFIG(TAG, "Setting up XDB401...");
```
jesserockz: https://developers.esphome.io/architecture/logging/#minimizing-impact
```suggestion
  ESP_LOGCONFIG(TAG, "Running setup");
```

esphome/components/xdb401/xdb401.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  float temperature;
+  float pressure;
+  i2c::ErrorCode err_code = this->read_(temperature, pressure);
+  if (err_code != i2c::ERROR_OK) {
+    // this->mark_failed("I2C Communication Failed...");  // Need latest dev
+    this->mark_failed();
+    ESP_LOGE(TAG, "I2C Communication Failed...");
```
jesserockz: https://github.com/esphome/esphome/blob/3470305d9d41837a8412eee0b4f6fdf247e6beed/esphome/core/log_const_en.h#L3
```suggestion
    this->mark_failed(ESP_LOG_MSG_COMM_FAIL);
```

esphome/components/xdb401/xdb401.h
```diff
@@ -0,0 +1,32 @@
...
+namespace esphome {
+namespace xdb401 {
+
+/// This class implements support for the xdb401 pressure and temperature i2c sensors.
+class XDB401Component : public PollingComponent, public i2c::I2CDevice {
+ public:
+  void set_temperature_sensor(sensor::Sensor *temperature_sensor) { this->temperature_sensor_ = temperature_sensor; }
+  void set_raw_pressure_sensor(sensor::Sensor *pressure_sensor) { this->pressure_sensor_ = pressure_sensor; }
+
+  float get_setup_priority() const override { return setup_priority::DATA; }
```
jesserockz: This is the default and doesn't need to be overriden
```suggestion
```

esphome/components/xdb401/xdb401.cpp
```diff
@@ -0,0 +1,179 @@
...
+  if (err_code != i2c::ERROR_OK) {
+    ESP_LOGE(TAG, "Error writing config to device, code: %u", err_code);
+    return err_code;
+  }
+
+  // Wait for measurement ready
+  uint8_t meas_resp[1] = {};
+  bool meas_mode = false;
+  for (int i = 1; i <= CHECK_ATTEMPTS; i++) {
+    delay(CHECK_DELAY);
```
jesserockz, line 81: This can lead to warnings in the logs about this component taking too long.

Please alter this to use a state machine.

```
Simple state machine:
update: 
  set state to make_measurement
  enable_loop

loop: Do only one action per loop call
if state is make_measurement:
	send command and set state to wait_for_measurement
else if state is wait_for_measurement:
  read measure response and set state to read_data if ready
	set state back to idle if attempts reached
    disable_loop
else if state is read_data
  read pressure
  read humidity
  publish state
  set state to idle
  disable_loop
```

jesserockz (changes requested):

esphome/components/xdb401/xdb401.cpp
dala318: Agree with this approach and will make a change to avoid blocking cyclic execution.
Specially since pressure measurements in the very most cases are not as time critical that you need them in the same cycle as requested.

(In more general terms I'm missing a more dynamic scheduler with different frequencies. Now it seem to be set to one default frequency for all cyclic actions, and no real event-driven interrupts. For the future it could be nice to select between some alternatives ranging from "time consuming main-loop but less often" and "quick main-loop executed more often")

PR #8585, opened 2025-04-17
support(CONF_ID): allow custom object_id via id: in setup_entity

Akari787:
### Summary

This PR enhances the `setup_entity()` function in `cpp_helpers.py` to provide a more robust and flexible `object_id` generation logic. The update addresses edge cases and allows better control through `id:` in YAML configuration.

### 🔧 Changes

- `CONF_ID` is now considered for `object_id` generation. If provided, it's used directly (snake-cased).
- If `CONF_NAME` is missing, `CORE.friendly_name` is used as a fallback.
- If `CONF_NAME` results in a string of only underscores (e.g. `"__"`), a unique `object_id` will be generated using an 8-character MD5 hash.
- Otherwise, the sanitized `CONF_NAME` is used as usual.

### 💡 Motivation

When using non-English names (e.g. Chinese characters), the `sanitize(snake_case(...))` logic often transforms them into strings containing only underscores (`"___"`), which are not usable as valid `object_id`s. 

This becomes problematic in scenarios where multiple entities (e.g. sensors) are configured with non-English `name:` values, leading to **duplicate `unique_id` errors** due to identical `object_id`s like `"___"`.

By:
- Prioritizing `id:` when provided,
- Falling back to friendly name when name is missing,
- And using an MD5 digest when `name:` produces only underscores,

we ensure consistent and unique `object_id` values across different languages and setups.

### 🧪 Test Coverage

Tested with:
- Configs using `name:` with Chinese and other non-ASCII characters
- Configs with only `id:`
- Edge cases like `name: "___"` and `name: ""`

### 2025-05-12
- **Enhanced `setup_entity` error handling**  
  - Removed automatic MD5 fallback when `name:` sanitizes to empty or all underscores.  
  - If sanitized `name:` is invalid and no `id:` is provided, now logs an error and raises a `cv.Invalid` exception with a clear message asking the user to set an explicit `id:`.  
  - Preserves original behavior for all valid ASCII `name:` values.


jesserockz:
This would break all entities in HA that users have where there is a name and id in config.

Akari787:
> This would break all entities in HA that users have where there is a name and id in config.

Thank you for catching that—I completely overlooked the impact on users who already rely on `name:` to generate their `object_id`. I’ve now revised the implementation to preserve the existing priority:

Always generate `object_id` from `name:` first.

Only if the sanitized `name:` is invalid (empty or all underscores) do we:

Fall back to `id:` if it’s provided, or

Otherwise use an 8‑character MD5 hash of the name.

With this change, every existing entity whose `name:` already maps to a valid ID will continue to use that same ID, and only the broken edge cases (non‑ASCII names collapsing to underscores) will receive the fallback logic.

esphome/cpp_helpers.py
```diff
@@ -100,9 +101,17 @@ async def setup_entity(var, config):
...
     else:
-        add(var.set_object_id(sanitize(snake_case(config[CONF_NAME]))))
+        object_id = sanitize(snake_case(config[CONF_NAME]))
+        #    If the result is empty or only underscores (e.g. from non-ASCII characters),fallback to a more reliable identifier
+        if not object_id or set(object_id) == {'_'}:
+            if CONF_ID in config:
+                object_id = sanitize(snake_case(str(config[CONF_ID])))
+            else:
+                from hashlib import md5
+                object_id = md5(config[CONF_NAME].encode("utf-8")).hexdigest()[:8]
```
glmnet: wouldn't it be better to raise an error here?
ask the user to enter a valid id for this case

Akari787: Thank you for the suggestion! You're absolutely right that raising an error would make the issue more visible and explicit.

However, doing so would unfortunately cause existing configurations from non-English users (who haven't set an `id:`) to fail entirely — even if their setups were previously working despite generating invalid `object_id`s.

To avoid breaking those configurations, a more graceful approach might be to **keep the MD5 fallback** for backward compatibility, but also **emit a warning** in the logs. This way, users are notified of the problem and encouraged to provide an explicit `id:`, without immediately breaking their devices.

glmnet: What do you mean by their configurations already working?

I mean it is ok that the program compiles and runs, but with this change you are going to change an id which is ___ into a hash (and probably already break something ha side) so I would vote to fail here and give the user control of what name he would like to use. 

The fix is really simple from the user side, you can add in the message "Cannot determine an id from the given name. Please explicitly set a meaningful id: for this entity"

Akari787: Do you mean changing it to something like this?

```python
        object_id = sanitize(snake_case(str(config[CONF_ID])))
    else:
        _LOGGER.error(f"The name '{config[CONF_NAME]}' is invalid after sanitization.")
        raise cv.Invalid(
            f"Cannot determine an object_id from the given name '{config[CONF_NAME]}'. "
            f"Please explicitly set a meaningful id: for this entity."
        )
```

As you mentioned, falling back to an MD5 hash might have already broken something on the HA side — that's something I hadn't considered before. Thanks for the clear and helpful explanation!

glmnet: Yes I meant that. 
May be the error log is not necessary 

I'm not sure if this change will affect the desire to make the entity names translatable. But for me it is good enough

esphome/cpp_helpers.py
```diff
@@ -100,9 +102,19 @@ async def setup_entity(var, config):
     """Set up generic properties of an Entity"""
     add(var.set_name(config[CONF_NAME]))
     if not config[CONF_NAME]:
-        add(var.set_object_id(sanitize(snake_case(CORE.friendly_name))))
+        object_id = sanitize(snake_case(CORE.friendly_name))
     else:
-        add(var.set_object_id(sanitize(snake_case(config[CONF_NAME]))))
+        object_id = sanitize(snake_case(config[CONF_NAME]))
+        #    If the result is empty or only underscores (e.g. from non-ASCII characters),fallback to a more reliable identifier
+        if not object_id or set(object_id) == {'_'}:
```
glmnet: the CI pipeline is failing because it is expecting you to use `"_"` (double quotes) here instead of `'_'`

Akari787: Thanks for pointing that out!

PR #8599, opened 2025-04-21
[I2S] Minor speedups: 

calumapplepie:

- Don't create a new vector on every loop, retain it between iterations.  Allocate the memory on start(), and release it on stop().
- Fix a bug where only half the buffer would be read (dividing the number of elements by their size, instead of multiplying)

## Types of changes

- [ ] Other

N/A (probably)

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
+++ b/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
@@ -171,7 +171,7 @@ size_t I2SAudioMicrophone::read(int16_t *buf, size_t len) {
     return 0;
   }
   if (bytes_read == 0) {
-    this->status_set_warning();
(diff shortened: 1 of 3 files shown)


calumapplepie:
Hey, while I was working on my ADC microphone, I noticed that the data callbacks aren't invoked if the I2S is being read via the read() method directly; eg, if you setup microphone callbacks, they won't get the data that goes to the voice assistant.  This is actually pretty bad, since it means that if callbacks are present, about half the data gets processed by the callbacks, and half by the wakeword; which is obviously pretty bad for both.  I'll probably retitle this PR soon and fix this issue.

PR #8612, opened 2025-04-24
[online_image] Support 24 bit bmp images

jesserockz:

Adds support for 24-bit bmp images

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4907

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/online_image/bmp_image.cpp
+++ b/esphome/components/online_image/bmp_image.cpp
@@ -86,36 +86,44 @@ int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
     index = this->data_offset_;
   }
   while (index < size) {
-    if (this->bits_per_pixel_ == 1) {
-      uint8_t current_byte = buffer[index];
-      for (uint8_t i = 0; i < 8; i++) {
-        size_t x = (this->paint_index_ % this->width_) + i;
+    switch (this->bits_per_pixel_) {
+      case 1: {
+        uint8_t current_byte = buffer[index];
+        for (uint8_t i = 0; i < 8; i++) {
+          size_t x = (this->paint_index_ % this->width_) + i;
+          size_t y = (this->height_ - 1) - (this->paint_index_ / this->width_);
+          Color c = (current_byte & (1 << (7 - i))) ? display::COLOR_ON : display::COLOR_OFF;
+          this->draw(x, y, 1, 1, c);
+        }
+        this->paint_index_ += 8;
+        this->current_index_++;
+        index++;
+        break;
+      }
+      case 24: {
+        if (index + 2 >= size) {
+          this->decoded_bytes_ += index;
+          return index;
+        }
+        uint8_t b = buffer[index];
+        uint8_t g = buffer[index + 1];
+        uint8_t r = buffer[index + 2];
+        size_t x = this->paint_index_ % this->width_;
         size_t y = (this->height_ - 1) - (this->paint_index_ / this->width_);
-        Color c = (current_byte & (1 << (7 - i))) ? display::COLOR_ON : display::COLOR_OFF;
+        Color c = Color(r, g, b);
         this->draw(x, y, 1, 1, c);
+        this->paint_index_++;
+        this->current_index_ += 3;
+        index += 3;
+        if (x == this->width_ - 1 && this->padding_bytes_ > 0) {
+          index += this->padding_bytes_;
+          this->current_index_ += this->padding_bytes_;
+        }
+        break;
       }
-      this->paint_index_ += 8;
-      this->current_index_++;
-      index++;
-    } else if (this->bits_per_pixel_ == 24) {
-      if (index + 2 >= size) {
-        this->decoded_bytes_ += index;
-        return index;
-      }
-      uint8_t b = buffer[index];
-      uint8_t g = buffer[index + 1];
-      uint8_t r = buffer[index + 2];
-      size_t x = this->paint_index_ % this->width_;
-      size_t y = (this->height_ - 1) - (this->paint_index_ / this->width_);
-      Color c = Color(r, g, b);
-      this->draw(x, y, 1, 1, c);
-      this->paint_index_++;
-      this->current_index_ += 3;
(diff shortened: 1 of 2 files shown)


esphome/components/online_image/bmp_image.cpp
```diff
@@ -79,17 +86,37 @@ int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
...
   }
   while (index < size) {
-    size_t paint_index = this->current_index_ - this->data_offset_;
-
-    uint8_t current_byte = buffer[index];
-    for (uint8_t i = 0; i < 8; i++) {
-      size_t x = (paint_index * 8) % this->width_ + i;
-      size_t y = (this->height_ - 1) - (paint_index / this->width_bytes_);
-      Color c = (current_byte & (1 << (7 - i))) ? display::COLOR_ON : display::COLOR_OFF;
+    if (this->bits_per_pixel_ == 1) {
```
guillempages: should this be a switch instead of an "if-else" chain? To both keep consitency with the change above (where the BPP are checked) and to be "future proof" if additional BPPs are implemented?

esphome/components/online_image/bmp_image.cpp
```diff
@@ -79,17 +86,37 @@ int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
...
+      uint8_t current_byte = buffer[index];
+      for (uint8_t i = 0; i < 8; i++) {
+        size_t x = (this->paint_index_ % this->width_) + i;
```
guillempages: Don't you need to initialize `paint_index` as was being done previously? It is being "class-initialized" to 0, but I'm not sure if the class is being reused when re-downloading. (And if it is not, it might be a possible optimization for the future, to save some extra (de-)allocation).

esphome/components/online_image/bmp_image.cpp
```diff
@@ -79,17 +86,37 @@ int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
...
+        size_t y = (this->height_ - 1) - (this->paint_index_ / this->width_);
```
guillempages, line 114: previously `width_bytes_` was being used; was this previously a mistake, or is this a copy&paste error here?

esphome/components/online_image/bmp_image.cpp
```diff
@@ -79,17 +86,37 @@ int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
...
+      }
+      this->paint_index_ += 8;
+      this->current_index_++;
+      index++;
+    } else if (this->bits_per_pixel_ == 24) {
+      if (index + 2 >= size) {
+        this->decoded_bytes_ += index;
+        return index;
+      }
+      uint8_t b = buffer[index];
```
guillempages: Don't know enough about the format, so just asking: Is it always stored as "BGR"? Or should somehow support for "RGB" also be added?

esphome/components/online_image/bmp_image.cpp
```diff
@@ -79,17 +86,37 @@ int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
...
+      uint8_t g = buffer[index + 1];
+      uint8_t r = buffer[index + 2];
+      size_t x = this->paint_index_ % this->width_;
+      size_t y = (this->height_ - 1) - (this->paint_index_ / this->width_);
```
guillempages: same here; is `this->width_` right, or should it be `this->width_bytes_`?

esphome/components/online_image/bmp_image.cpp
jesserockz: This is intentional to align with the paint index and make the different bpp consistent in their implementations as much as possible.

esphome/components/online_image/bmp_image.cpp
jesserockz: BMP is little endian and generally stores as BGR

esphome/components/online_image/bmp_image.cpp
jesserockz: As above

esphome/components/online_image/bmp_image.cpp
jesserockz: The way the code is architected at the moment, the decoder is used once then deallocated. If we were to change that, we can make those changes at the time to match.

esphome/components/online_image/bmp_image.cpp
jesserockz: Done

esphome/components/online_image/bmp_image.cpp
```diff
@@ -86,36 +86,44 @@ int HOT BmpDecoder::decode(uint8_t *buffer, size_t size) {
...
-      index += 3;
-      if (x == this->width_ - 1 && this->padding_bytes_ > 0) {
-        index += this->padding_bytes_;
-        this->current_index_ += this->padding_bytes_;
-      }
+      default:
+        ESP_LOGE(TAG, "Unsupported bits per pixel: %d", this->bits_per_pixel_);
+        return DECODE_ERROR_UNSUPPORTED_FORMAT;
     }
   }
```
guillempages: Just a suggestion, but IIRC, I did some tests on performance a long time ago, and found out that, doing the while within the switch (instead of the switch within the while) was faster...

So you might want to do e.g. 
```
switch( ...) {
  case 1: {
    while (index < size) {
        ...
    }
   break;
  }
  case 24: {
     while (index < size) {
      ...
     }
    break;
default:
  ESP_LOGE(...)
  return
}
```
(sorry, github doesn't let me do  a proper suggestion here ☹️  )

jesserockz: Makes sense as it only needs to switch once and its not like it is gonna change

guillempages (approved):

PR #8383, opened 2025-03-10
[mipi_spi] New display driver for MIPI DBI devices

clydebarrow:
# What does this implement/fix?

This driver brings together 3 kinds of MIPI-DBI (Data Bus Interface) display panels. All these panels implement the MIPI DCS (Display Command Set) and have on-board LCD refresh ram.

The interfaces supported are:

* 1 bit SPI - as per the existing `ili9xxx` driver
* Quad SPI - as per the existing `qspi_dbi` driver
* I80 bus, implemented as octal SPI. New to ESPHome

Features of this driver are:

* The driver includes all the models supported by the existing single and quad spi drivers, plus new ones, plus some I80 bus displays.
* Configuration is all handled in Python - no embedded init sequences. Adding a new model only requires a few extra lines of Python.
* Pin assignments and data rate are defaulted by the model where it relates to an board with integrated ESP and display.. This means for those boards the display config may be just two lines.
* Multiple enable pins can be assigned where boards have more than one pin to set for the display to operate.
* Hardware rotation is handled automatically by the `rotation` option if possible, falling back to software rotation only if needed.
* Custom init sequences including delays can be specified for the CUSTOM model.
* 16 and 18 bit pixel format displays are supported
* The display buffer supports 16 and 8 bit color depths, orthogonal to the controller format.
* 16 bit spi data mode for certain Waveshare models is supported.
* Color inversion supported
* AMOLED brightness setting supported.
* MADCTL alternate x/y mirror bits supported
* SLPOUT delays are handled automatically and kept to a minimum
* 18 bit output has been optimised from the ili9xxx driver.

List of supported driver chips:

* RM690B0
* ILI9341
* ILI9481
* ILI9486
* ILI9488
* ILI9488_A
* ST7796
* ST7789V
* GC9A01A
* GC9D01N
* AXS15231
* ST7735
* CUSTOM

List of so far supported integrated boards:

* M5STACK
* M5CORE
* S3BOX
* S3BOXLITE
* WAVESHARE-4-TFT
* Pico-ResTouch-LCD-3.5
* WT32-SC01-PLUS
* ESP32-2432S028
* JC3248W535
* JC3636W518
* LANBON_L8
* T4-S3
* T-EMBED
* T-DISPLAY
* T-DISPLAY-S3
* T-DISPLAY-S3-PRO
* T-DISPLAY-S3-AMOLED
* T-DISPLAY-S3-AMOLED-PLUS

## Types of changes

[...]

PR #8613, opened 2025-04-24
[st7789v] Fractional display buffer

lhartmann:

Adds fractional framebuffer support for the st7789v display driver, allowing usage on low-RAM devices (e.g. ESP8266 and Geekmagic SmallTV Ultra).

Changes:
* `fragmentation` is added to the display configuration. Optional int >=1, default 1.
* The display buffer is reduced to 1/fragments in size.
* `st7789v::update()` function calls `display::do_update_()` once per fragment.
* Internal counters were added to `st7789v`, in order to keep track of the active fragment.
* `st7789v::fill(Color)` was specialized for performance.

Side-effects:
* Multiple calls to `display::do_update_()` will trigger the display lambda multiple times.
  * Randomized things will be different per-fragment (not frame).
  * Increments/decrements will happen once per fragment (not frame).
* Slower than full-framebuffer, but still way faster than no-framebuffer.
* Testcard does not render properly, only shows the first fragment.

Mitigation: `fragmentation` defaults to 1.
* 1 fragment means a full-framebuffer, which is the old behavior.
* 1 fragment means lambda is only called once per frame, which is the old behavior.
* `fragmentation>=1` only changes lambda behavior for users that enable it

## Types of changes

- [ ] Other

-  fixes (https://github.com/esphome/feature-requests/issues/3131), but locally on the st7789v display driver.

- esphome/esphome-docs#4840

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [X] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Tested with a GeekMagic SmallTV Ultra, ESP8266-based, 240x240 pixels:
* Does not work at all without fragmentation.
* Using `eightbitcolor: True`
  * `fragmentation: 2` was unstable. Resets with out-of-memory exception after a few seconds.
  * `fragmentation: 3` was stable.
* Using `eightbitcolor: False` (i.e., 16-bit RGB565)
  * `fragmentation: 6`, was stable.
* Higher `fragmentation` values were stable, but only made the update slower.

```yaml
display:
  - id: my_display
    platform: st7789v
    model: custom
    spi_id: spihwd
    height: 240
    width: 240
    offset_height: 0
    offset_width: 0
    fragmentation: 3
    dc_pin: GPIO00
    reset_pin: GPIO02
    eightbitcolor: True
    update_interval: 1s
    spi_mode: mode3
    data_rate: 40000000
    lambda: |-
      // Calculate the center positions
      int center_x = it.get_width() / 2;
      int center_y = it.get_height() / 2;

      // Print "Hello World" centered
      it.printf(center_x, 240-20, id(roboto), id(color_green), TextAlign::BOTTOM_CENTER, "Hello World");

      // This is a bad-example, to demonstrate side-effect multiplication.
      // With even fragmentation this is cut in half, with k on top and k+1 below.
      static int framecounter = 0;
      it.printf(center_x, 240, id(roboto), TextAlign::BOTTOM_CENTER, "%d", framecounter++);

```

## Checklist:


kbx81:
@lhartmann are you aware that this component is deprecated? See the warning on the docs page: https://esphome.io/components/display/st7789v

Performance is _significantly_ improved with the [ILI9xxx component](https://esphome.io/components/display/ili9xxx).

lhartmann:
Yup. I just could not get it working with ili9xxx (yet). The device I am testing with has CS tied to GND. Not sure if this is the reason it doesn't work. 

Also, ili9xxx driver has an inconsistency on draw pixel versus pixels, where the second bypasses the buffer entirely. Still got some more reading/understanding to do before I feel can touch something there.

kbx81:
I would very, very highly recommend focusing your efforts on the ILI9xxx  component; the performance of this one is quite poor by comparison. If you're unsure of how to proceed with and/or have questions about that one, hop on Discord and/or start a GH discussion. 🙂

clydebarrow:
I don't believe the ili9xxx driver works on ESP8266, due to the 8266's inablity to fetch data from flash except as words (PROGMEM and all that.)

More generally though, this is as far as I can tell just to support one device. While I understand the value of repurposing old hardware, it seems unlikely to be of use for anything else much (correct me if I'm wrong here.) Should it just stay as an external component that doesn't have to be kept up to date with other display changes?

clydebarrow:
Also see https://github.com/esphome/esphome/pull/8383 - this is going to supersede the ili9xxx driver. It currently doesn't work on ESP8266, but can be made to do so fairly easily, so if this feature is going to go in, that might be a better base. It's not merged yet though, still waiting for review.

lhartmann:
> I don't believe the ili9xxx driver works on ESP8266, due to the 8266's inablity to fetch data from flash except as words (PROGMEM and all that.)

Every time I added ili9xxx to the config it would crash on boot, and there were no messages I could undersand.

> More generally though, this is as far as I can tell just to support one device.

While the concept of fractional framebuffer should be usefull for RAM savings on any device, especially ESP8266 and ESP32 without PSRAM, I only implemented for st7789v, and tested on a GeekMagic SmallTV Ultra. I don't think there is any negative impact for any other device/user, but if keeping this as an external repo is the option then just reject the pull, and I will fix the mess I made by commiting to my own dev branch (oops). 

https://github.com/esphome/esphome/pull/8383 looks great. Will take a deeper look later.

<details>  <summary>Rambling</summary>
I still think the drivers should break interface - protocol - buffer - drawing on separate classes, and template them together based on the configuration. Color depth could be used as a template parameter both for display and images, so that on a `draw(display, image, ...)` call the compiler could deduce the right types without runtime tests, much like OpenCV does on it's classes. Also, instead of running multiple checks on `draw_pixel(...)` one could ask for an iterator-style handle, all checks would be performed once, and an iterator should only be returned if tests are ok. Make the iterator mostly inline, and performance should be good. All of this could be done with minimal user-facing impact, but would still be a much larger effort, and discussion for another time.
</details>

kbx81:
I've got no major objections to these changes; I just want to be sure that effort is being focused where it can have the greatest impact. 🙂
My biggest hesitation is that this is introducing config options to a deprecated platform which (and please do correct me if I'm wrong here) adds complexity in an effort to implement workarounds for low-memory devices which we [no longer recommend](https://esphome.io/guides/faq#not-recommended). 🙈

clydebarrow:
I would suggest waiting until the new mipi_spi display driver is merged and revisit it then. Adding to ili9xxx is not a viable option, and it would be best if the st7789v driver was allowed to die quietly.

lhartmann:
> I would suggest waiting until the new mipi_spi display driver is merged and revisit it then. Adding to ili9xxx is not a viable option, and it would be best if the st7789v driver was allowed to die quietly.

Ok, st7789v may rest in peace. Will cancel the pull.

However, the appropriate place for fractional framebuffer implementation is most likelly the `DisplayBuffer` class, which is inherited by all buffered display drivers. I believe I can implement fragmentation by overriding `do_update_(...)`, I 'm just not sure what to do with the `get_***_internal(...)` functions. Ideally they should return the fragment size, but I worry this change may break some drivers.

clydebarrow:
> the appropriate place for fractional framebuffer implementation is most likelly the `DisplayBuffer` class

That would make sense.

PR #8640, opened 2025-04-28
[i2c] Allow buffers in PSRAM

clydebarrow:

The i2c peripheral was passing the `ESP_INTR_FLAG_IRAM` flag to the i2c setup, so if a write or read call with a buffer in PSRAM was passed, it failed.

Test for PSRAM being used and disable that flag in that case.

## Types of changes

- [ ] Other

- fixes https://discord.com/channels/429907082951524364/1362186392930619624/1366322628599746560

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


clydebarrow:

```yaml
external_components:
  - source: github://pr#8640
    components: [i2c]
    refresh: 1h
```

---
(Added by the PR bot)

clydebarrow:
There is the question of whether that flag is needed even without PSRAM. Its main purpose is to allow i2c operations with the cache disabled, e.g. when writing flash. That only happens in ESPHome during OTA, at which time the i2c will be paused anyway.

esphome/components/i2c/i2c_bus_esp_idf.cpp
```diff
@@ -67,7 +67,11 @@ void IDFI2CBus::setup() {
       ESP_LOGV(TAG, "i2c_timeout set to %" PRIu32 " ticks (%" PRIu32 " us)", timeout_ * 80, timeout_);
     }
   }
+#ifdef USE_PSRAM
```
jesserockz: This flag alone is probably not enough to determine if psram is actually being used as you can enable the psram component on any device which sets this flag even if the device doesnt have psram. 
If it works either way then that is fine too.

clydebarrow: If the flag's not set it guarantees that PSRAM isn't in use - unless the user has added sdkconfig options.

But after reviewing the ESP-IDF docs, I'm confident the flag is unnecessary in ESPHome.

jesserockz (approved):

PR #8653, opened 2025-04-29
feat: ``wifi.configure`` now emits error after reconnecting to old AP

Rapsssito:

Before this PR, `wifi.configure` would just call the error callback after the connection timeout, before it had reconnected to the old AP. Now, it waits until the old connection has been reestablished to call the error callback. This facilitates the interaction flow when configuring a new Wi-Fi network.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:

### DIFF
--- a/esphome/components/wifi/wifi_component.h
+++ b/esphome/components/wifi/wifi_component.h
@@ -484,13 +484,15 @@ template<typename... Ts> class WiFiConfigureAction : public Action<Ts...>, publi
     global_wifi_component->enable();
     // Set timeout for the connection
     this->set_timeout("wifi-connect-timeout", this->connection_timeout_.value(x...), [this]() {
-      this->connecting_ = false;
       // If the timeout is reached, stop connecting and revert to the old AP
       global_wifi_component->disable();
       global_wifi_component->save_wifi_sta(old_sta_.get_ssid(), old_sta_.get_password());
       global_wifi_component->enable();
-      // Callback to notify the user that the connection failed
-      this->error_trigger_->trigger();
+      // Start a timeout for the fallback if the connection to the old AP fails
+      this->set_timeout("wifi-fallback-timeout", 30000, [this]() {
+        this->connecting_ = false;
+        this->error_trigger_->trigger();
+      });
     });
   }
 
@@ -503,6 +505,7 @@ template<typename... Ts> class WiFiConfigureAction : public Action<Ts...>, publi
     if (global_wifi_component->is_connected()) {
       // The WiFi is connected, stop the timeout and reset the connecting flag
       this->cancel_timeout("wifi-connect-timeout");
+      this->cancel_timeout("wifi-fallback-timeout");
       this->connecting_ = false;
       if (global_wifi_component->wifi_ssid() == this->new_sta_.get_ssid()) {
         // Callback to notify the user that the connection was successful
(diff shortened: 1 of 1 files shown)


esphome/components/wifi/wifi_component.h
```diff
@@ -484,13 +484,15 @@ template<typename... Ts> class WiFiConfigureAction : public Action<Ts...>, publi
...
     this->set_timeout("wifi-connect-timeout", this->connection_timeout_.value(x...), [this]() {
-      this->connecting_ = false;
       // If the timeout is reached, stop connecting and revert to the old AP
       global_wifi_component->disable();
       global_wifi_component->save_wifi_sta(old_sta_.get_ssid(), old_sta_.get_password());
       global_wifi_component->enable();
-      // Callback to notify the user that the connection failed
-      this->error_trigger_->trigger();
+      // Start a timeout for the fallback if the connection to the old AP fails
+      this->set_timeout("wifi-fallback-timeout", 30000, [this]() {
```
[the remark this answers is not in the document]
jesserockz: I would reuse `this->connection_timeout_.value(x...)` here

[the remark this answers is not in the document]
Rapsssito: Done!

Rapsssito:
@jesserockz, just pinging you as a friendly reminder.

esphome/components/wifi/wifi_component.h
```diff
@@ -483,14 +483,16 @@ template<typename... Ts> class WiFiConfigureAction : public Action<Ts...>, publi
...
-      this->connecting_ = false;
+    this->set_timeout("wifi-connect-timeout", this->connection_timeout_.value(x...), [this, x...]() {
       // If the timeout is reached, stop connecting and revert to the old AP
       global_wifi_component->disable();
       global_wifi_component->save_wifi_sta(old_sta_.get_ssid(), old_sta_.get_password());
       global_wifi_component->enable();
-      // Callback to notify the user that the connection failed
-      this->error_trigger_->trigger();
+      // Start a timeout for the fallback if the connection to the old AP fails
+      this->set_timeout("wifi-fallback-timeout", this->connection_timeout_.value(x...), [this]() {
```
jesserockz, line 492: This timeout will always trigger with the connection timeout, no matter how long it takes to go back to the original ap.
The timeout will never be cancelled as `connecting_` stays true until this timeout meaning `loop` just returns.

Rapsssito: I am sorry, but I don't understand what you mean. `connecting_` stays true so the `loop()` can check the WiFi status.

The fallback timeout does not trigger if the fallback WiFi is connected (thanks to the connection condition check in the `loop()`). I have tested it and it works.

jesserockz: Ah my bad, the bits were all flipping in my head.

But this also means the error trigger will not fire if the old ap was connected back to?
I feel like it should still fire otherwise the user will not know that this action "failed"

Rapsssito: The error will still trigger, take a look at the now lines 512-515:
```
      if (global_wifi_component->wifi_ssid() == this->new_sta_.get_ssid()) {
        // Callback to notify the user that the connection was successful
        this->connect_trigger_->trigger();
      } else {
        // Callback to notify the user that the connection failed
        this->error_trigger_->trigger();
      }
```

jesserockz: Ah, existing code that was hidden in the PR diff.

jesserockz (approved):

PR #8655, opened 2025-04-30
[micro_wake_word] add new VPE features

### LINKED ISSUE #2828: Allow micro_wake_word model to be set from a template select value
Trying to create a template select in Home Assistant that can set/change the wake word used by micro_wake_word. However when using what I believe is the proper lambda syntax to load the value, an error is generated when trying to install the new package.

I should add that when all current models are added to the micro_wake_word configuration, the device will respond to all of the wake words, not just the desired one. That would work if there was a way to enable/disable each wake word independently in some way that could be controlled from Home Assistant.

I'm using BigBobbas ESP32-S3-Box-3 configuration for a HA voice assistant that can also be a media receiver. Since the audience of users might have different preferences on wake word, I'm trying to assist with finding a way to make it selectable without having to recompile ESPHome to change and experiment with what someone likes best.

**Additional context**
Here is my YAML for micro_wake_word:
```
select:
  - platform: template
    entity_category: config
    name: Wake word model
    id: wake_word_model
    optimistic: true
    restore_value: true
    options:
      - alexa
      - hey_jarvis
      - hey_mycroft
      - okay_nabu
    initial_option: hey_jarvis
micro_wake_word:
  vad:
  models: 
    - model: !lambda 'return id(wake_word_model).state;'
```

Here are the logs when trying to install the configuration to the S3:
```
INFO ESPHome 2024.7.2
INFO Reading configuration /config/living-room-satellite.yaml...
INFO Updating https://github.com/gnumpi/esphome_audio@dev-next
INFO Detected timezone 'America/New_York'
Failed config

micro_wake_word: [source /config/living-room-satellite.yaml:650]
  vad: 
  models: 
    - 
      Not a valid model name, local path, http(s) url, or github shorthand.
      model: !lambda |-
        return id(wake_word_model).state;
```

### LINKED ISSUE #3039: micro_wake_word: Support "id" and "internal" for mww

Looking at the improvements of mww in [home-assistant-voice-pe](https://github.com/esphome/home-assistant-voice-pe/tree/dev/esphome/components/micro_wake_word) I noticed that you implemented support for `id` and `internal` for mww. Example:

```yml
micro_wake_word:
  id: mww
  models:
    - model: hey_mycroft
      id: hey_mycroft
    - model: https://github.com/kahrendt/microWakeWord/releases/download/stop/stop.json
      id: stop
      internal: true
```

Since I want to experiment with the stop word feature as implemented in `home-assistant-voice-pe` I would like to ask, if you could please implement it in the upstream mww.

Using the implementation in home-assistant-voice-pe as external component does not work due to the further dependencies.

**Additional context**

kahrendt:

- Runs mWW inferences in a task to avoid slowing down the main loop.
- Individual models can be enabled or disabled with new yaml actions
  - A future PR will add support to the voice_assistant component so they can be enabled/disabled via HA (models marked internal will not be exposed to HA)
  - Model's enabled state is restored from flash
- Adds new option for ``stop_after_detection``. Defaults to true to mimic current behavior. Can be disabled if the device has duplex audio support (currently requires separate I2S busses) and the user wants to leave mWW always running
- Bumps the [esp-tflite-micro library](https://github.com/espressif/esp-tflite-micro) to the latest version

This is a breaking change in that only the first model listed will be enabled on the first boot (currently all models would be enabled).

## Types of changes

- [ ] Other

- closes esphome/feature-requests#3039
- closes esphome/feature-requests#2828 Models can now be enabled/disabled via yaml actions, so this is now possible

- esphome/esphome-docs#4858

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

Silly example, you probably wouldn't want to enable/disable models in the wake_word_detected action.

```yaml
# Example config.yaml

micro_wake_word:
  microphone:
    gain_factor: 4
  models:
    - model: hey_jarvis
      id: hey_jarvis_model
    - model: okay_nabu
      id: okay_nabu_model
  on_wake_word_detected:
    - if:
        condition:
          - micro_wake_word.model_is_enabled: hey_jarvis_model
        then:
          - micro_wake_word.disable_model: hey_jarvis_model
        else:
          - micro_wake_word.enable_model: hey_jarvis_model
```

## Checklist:


esphome/core/defines.h
```diff
@@ -79,6 +79,7 @@
 #define USE_LVGL_TEXTAREA
 #define USE_LVGL_TILEVIEW
 #define USE_LVGL_TOUCHSCREEN
+#define USE_MICRO_WAKE_WORD
```
jesserockz, line 82: Last time I checked I came after E :wink:
But also, this should go inside the `#ifdef USE_ESP32` block below

jesserockz (approved):
One minor thing below I will fix in a new PR to not hold this one up

PR #8680, opened 2025-05-03
[gps] Add hdop sensor

### LINKED ISSUE #3140: [gps] Add HDOP sensor
TinyGPSPlus in currently used version 1.0.2 supports reading HDOP (Horizontal Dilution Of Precision).
However the current `gps` component does not define a sensor for it.
Additionally:
1. I think it is helpful to know that defining `update_interval` within `gps` allows to have more frequent updates than current default of 20s. An update of the `gps` documentation should mention this as it is already supported without any change of `gps` component source code.
2. Current defaults for all sensors is defined as -1. But -1 is a valid latitude and longitude. I think a default of `nan` would be better.
3. Debug logs of sensors use 2 lines per value. This could easily be reduced to 1 line per value.

In order to quantify how good/bad a location (lat/long) report is, it is useful to know the HDOP of the location.

**Additional context**
I am about to contribute 2 changes:
1. esphome/esphome/components/gps: Add `hdop` sensor
2. esphome-docs/components/gps.rst: document `hdop` and `update_interval`

realzoulou:
TinyGPSPlus in currently used version 1.0.2 supports reading HDOP (Horizontal Dilution Of Precision).
However the current `gps` component does not define a sensor for it.
This add `hdop` as an additional sensor for `gps`.
Additional changes
- Change debug logs to 1 line per value instead of 2
- Change defaults from -1 to NAN, satellites defaults to 0

## Types of changes

- [X] Code quality improvements to existing code or addition of tests
- [ ] Other

- fixes https://github.com/esphome/feature-requests/issues/3140

- esphome/esphome-docs#4882

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
  hdop:
    name: "HDOP"
```

## Checklist:

  - [X] Documentation added/updated in https://github.com/esphome/esphome-docs/pull/4879

## Tested with
1. Heltec HTIT Tracker (https://heltec.org/project/wireless-tracker) and its onboard UC6580 GNSS chipset
2. Same Heltec, but with external u-blox NEO-F10N (https://www.sparkfun.com/sparkfun-gnss-l1-l5-breakout-neo-f10n-sma.html)

### DIFF
--- a/esphome/components/gps/__init__.py
+++ b/esphome/components/gps/__init__.py
@@ -25,6 +25,7 @@ GPS = gps_ns.class_("GPS", cg.Component, uart.UARTDevice)
 GPSListener = gps_ns.class_("GPSListener")
 
 CONF_GPS_ID = "gps_id"
+CONF_HDOP = "hdop"
 MULTI_CONF = True
 CONFIG_SCHEMA = cv.All(
     cv.Schema(
@@ -40,7 +41,7 @@ CONFIG_SCHEMA = cv.All(
             ),
             cv.Optional(CONF_SPEED): sensor.sensor_schema(
                 unit_of_measurement=UNIT_KILOMETER_PER_HOUR,
-                accuracy_decimals=6,
+                accuracy_decimals=3,
             ),
             cv.Optional(CONF_COURSE): sensor.sensor_schema(
                 unit_of_measurement=UNIT_DEGREES,
@@ -48,10 +49,14 @@ CONFIG_SCHEMA = cv.All(
             ),
             cv.Optional(CONF_ALTITUDE): sensor.sensor_schema(
                 unit_of_measurement=UNIT_METER,
-                accuracy_decimals=1,
+                accuracy_decimals=2,
             ),
             cv.Optional(CONF_SATELLITES): sensor.sensor_schema(
                 accuracy_decimals=0,
+                state_class=STATE_CLASS_MEASUREMENT,
+            ),
+            cv.Optional(CONF_HDOP): sensor.sensor_schema(
+                accuracy_decimals=3,
                 state_class=STATE_CLASS_MEASUREMENT,
             ),
         }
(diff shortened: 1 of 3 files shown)


coogle:
Sure I can take a peek.

esphome/components/gps/__init__.py
```diff
@@ -40,20 +41,24 @@
             ),
             cv.Optional(CONF_SPEED): sensor.sensor_schema(
                 unit_of_measurement=UNIT_KILOMETER_PER_HOUR,
-                accuracy_decimals=6,
```
coogle, line 43: Why have you changed the accuracy here?

esphome/components/gps/gps.cpp
```diff
@@ -37,30 +40,29 @@ void GPS::loop() {
         this->latitude_ = tiny_gps_.location.lat();
         this->longitude_ = tiny_gps_.location.lng();
 
-        ESP_LOGD(TAG, "Location:");
-        ESP_LOGD(TAG, "  Lat: %f", this->latitude_);
-        ESP_LOGD(TAG, "  Lon: %f", this->longitude_);
```
coogle: No need to change the spacing on these, please leave as-is

coogle (changes requested):
Overall looks fine, but I think you've made some minor changes that were unnecessary and I'd like to see those undone unless there is a reason for them.

esphome/components/gps/__init__.py
realzoulou: Speed accuracy in GNSS is far below 6 decimals. Even 2 decimals are too much accuracy. If you disagree then I will revert back to original 6 decimals.

coogle: Nah, if there is a good reason for it I'm fine with it. However, according to [this](https://www.gps.gov/systems/gps/performance/accuracy)

> The government provides the GPS signal in space with a global average user range rate error (URRE) of ≤0.006 m/sec over any 3-second interval, with 95% probability.

So we should probably use that as the precision?

realzoulou: I updated to 3 decimals

esphome/components/gps/__init__.py
```diff
@@ -92,5 +97,9 @@ async def to_code(config):
         sens = await sensor.new_sensor(config[CONF_SATELLITES])
         cg.add(var.set_satellites_sensor(sens))
 
+    if CONF_HDOP in config:
+        sens = await sensor.new_sensor(config[CONF_HDOP])
```
jesserockz:
```suggestion
    if hdop_config := config.get(CONF_HDOP):
        sens = await sensor.new_sensor(hdop_config)
```

jesserockz (changes requested):

jesserockz (approved):

coogle (approved):

jesserockz:
Thanks for looking over @coogle

PR #8690, opened 2025-05-04
Pulse Counter Sensor: Pulse counter dir pin

antohei:

Added counting direction input pin to pulse counter sensor component for changing the direction of counting.

Includes also the fix in esphome/esphome#8689: A negative total number of pulses was not possible in the pulse counter sensor. This fixes the issue by changing uint32_t to int32_t.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4885

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example configuration entry
sensor:
  - platform: pulse_counter
    pin: GPIOXX
    dir_pin: GPIOXX
    name: "Pulse Counter"

```

## Checklist:


nagyrobi:
Direction should be something templatable, not limited to a pin only. There could be other useful state sources to change direction, not only a physical pin, eg. a binary sensor provided by any other platform.

antohei:
> Direction should be something templatable, not limited to a pin only. There could be other useful state sources to change direction, not only a physical pin, eg. a binary sensor provided by any other platform.

Yes, the direction input and also the pulse input should be templatable. However, when using the pulse counter peripheral (use_pcnt: true in yaml), this is not possible, neither for the pulse input nor for the direction input. The pulse_gpio_num and ctrl_gpio_num of the pcnt_config_t structure require GPIO numbers.

jesserockz:
I would say allow a gpio pin, a binary sensor or a switch in config.

Config validation can be written to ensure you are using a gpio pin when pcnt is true (if any direction control is desired)

antohei:
> I would say allow a gpio pin, a binary sensor or a switch in config.
> 
> Config validation can be written to ensure you are using a gpio pin when pcnt is true (if any direction control is desired)

Sorry, this is beyond my skills. Maybe someone could do this?

esphome/components/pulse_counter/pulse_counter_sensor.h
```diff
@@ -66,12 +67,13 @@ class PulseCounterSensor : public sensor::Sensor, public PollingComponent {
   explicit PulseCounterSensor(bool hw_pcnt = false) : storage_(*get_storage(hw_pcnt)) {}
 
   void set_pin(InternalGPIOPin *pin) { pin_ = pin; }
+  void set_dir_pin(InternalGPIOPin *pin) { dir_pin_ = pin; }
```
DjordjeMandic:
```suggestion
  void set_dir_pin(InternalGPIOPin *pin) { this->dir_pin_ = pin; }
```

esphome/components/pulse_counter/pulse_counter_sensor.cpp
```diff
@@ -29,17 +29,31 @@ void IRAM_ATTR BasicPulseCounterStorage::gpio_intr(BasicPulseCounterStorage *arg
...
     case PULSE_COUNTER_INCREMENT: {
-      auto x = arg->counter + 1;
-      arg->counter = x;
+      if (arg->dir_pin != nullptr && arg->dir_pin->digital_read()) {
+        auto x = arg->counter - 1;
+        arg->counter = x;
+      } else {
+        auto x = arg->counter + 1;
+        arg->counter = x;
+      }
```
DjordjeMandic: I'd prefer something like this:
```suggestion
      auto x = arg->counter;
      if (arg->dir_pin != nullptr && arg->dir_pin->digital_read()) {
        --x;
      } else {
        ++x;
      }
      arg->counter = x;
```

Could it then be shortened to this? How does overflow behave in this scenario? 
```suggestion
      if (arg->dir_pin != nullptr && arg->dir_pin->digital_read()) {
        --arg->counter;
      } else {
        ++arg->counter;
      }
```

This is shorter but ugly:
```cpp
      arg->counter += (arg->dir_pin != nullptr && arg->dir_pin->digital_read()) ? -1 : 1;
```

esphome/components/pulse_counter/pulse_counter_sensor.cpp
```diff
@@ -29,17 +29,31 @@ void IRAM_ATTR BasicPulseCounterStorage::gpio_intr(BasicPulseCounterStorage *arg
...
     case PULSE_COUNTER_DECREMENT: {
-      auto x = arg->counter - 1;
-      arg->counter = x;
+      if (arg->dir_pin != nullptr && arg->dir_pin->digital_read()) {
+        auto x = arg->counter + 1;
+        arg->counter = x;
+      } else {
+        auto x = arg->counter - 1;
+        arg->counter = x;
+      }
```
DjordjeMandic: Same here, could be bit shorter by moving x before the if statement, and assignment operation after.

DjordjeMandic (changes requested):
Nitpicking mostly.

DjordjeMandic:
I agreed with others, binary sensor, switch or gpio pin or "positive/negative" could be used if pcnt is not available.

Sadly I also do not know how to validate one config parameter ('dir') to check for those multiple types.

antohei:
C++20 deprecated increment/decrement operations on volatile types, "arg->counter" is volatile.

This is simpler than using local non-volatile variable:

```c++
case PULSE_COUNTER_INCREMENT: {
  if (arg->dir_pin != nullptr && arg->dir_pin->digital_read()) {
    arg->counter = arg->counter - 1;
  } else {
    arg->counter = arg->counter + 1;
  }
} break;
```

PR #8694, opened 2025-05-05
[Packet transport] Ping timeout sensor

dala318:

Adds a binary status sensor per remote provider that tells if the connection is up (checks for ping-pong timeout)

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome-docs#4955

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

packet_transport:
  - id: my_packet_transport
    providers:
      - name: Provider 1
        status_sensor:
          name: Provider 1 connected
          # All other from binary_sensor

```

## Checklist:


esphome/components/packet_transport/__init__.py
```diff
@@ -76,9 +83,19 @@ def provider_name_validate(value):
     )
 }
 
+STATUS_SENSOR_SCHEMA = binary_sensor_schema(
+    device_class=DEVICE_CLASS_CONNECTIVITY,
+    entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
+).extend(
+    {
+        cv.GenerateID(CONF_TRANSPORT_ID): cv.use_id(PacketTransport),
```
clydebarrow: Not needed (and not used.)

esphome/components/packet_transport/__init__.py
```diff
@@ -173,10 +190,18 @@ async def register_packet_transport(var, config):
...
+    define_binary_sensor = False
     for provider in config[CONF_PROVIDERS]:
         name = provider[CONF_NAME]
         if encryption := provider.get(CONF_ENCRYPTION):
             cg.add(var.set_provider_encryption(name, hash_encryption_key(encryption)))
+        if status_sensor := provider.get(CONF_STATUS_SENSOR):
+            define_binary_sensor = True
+            sens = await new_binary_sensor(status_sensor)
+            cg.add(var.set_provider_status_sensor(name, sens))
+    if define_binary_sensor:
```
clydebarrow: this is ok, but a more elegant way would be:
```suggestion
    if any(CONF_STATUS_SENSOR in p for p in config[CONF_PROVIDERS]):
```

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -314,11 +314,36 @@ void PacketTransport::send_data_(bool all) {
...
+  for (const auto &provider : this->providers_) {
+    if (provider.second.status_sensor != nullptr) {
+      uint32_t key_response_age = now - provider.second.last_key_response_time;
+      if (key_response_age > (this->ping_pong_recyle_time_ * 2u)) {
+        if (provider.second.status_sensor->state) {
+          ESP_LOGW(TAG, "Ping status for %s timeout at %u with age %u", provider.first.c_str(), now, key_response_age);
+          provider.second.status_sensor->publish_state(false);
+        }
+        for (auto &sensor : this->remote_sensors_[provider.first]) {
+          sensor.second->publish_state(0);
```
clydebarrow:
```suggestion
          sensor.second->publish_state(NAN);
```
You will also need to guard this with `#ifdef USE_SENSOR`

jesserockz:
The test will fail as the `binary_sensor` headers have not been included as there is no top level `binary_sensor` in the config and it is not auto loaded (which we discourage these days)

It would be more config, but better to make it so it is:

```yaml
binary_sensor:
  - platform: packet_transport
    type: ping
    provider_id: ...
    ...
```

the binary sensor platform for packet_transport would need to be altered to use a cv.typed_schema and can have a default type of `sensor` or similar

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -314,11 +314,36 @@ void PacketTransport::send_data_(bool all) {
...
+        }
+        for (auto &sensor : this->remote_binary_sensors_[provider.first]) {
+          sensor.second->publish_state(false);
```
clydebarrow: I'm not sure this is a good idea - unlike `sensor` there is no way of flagging a binary sensor as unavailable, and defaulting to false is making an assumption that is guaranteed to be wrong some of the time.

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -437,7 +462,8 @@ void PacketTransport::process_(const std::vector<uint8_t> &data) {
     if (decoder.decode(PING_KEY, key) == DECODE_OK) {
       if (key == this->ping_key_) {
         ping_key_seen = true;
-        ESP_LOGV(TAG, "Found good ping key %X", (unsigned) key);
+        provider.last_key_response_time = millis() / 1000;
+        ESP_LOGV(TAG, "Found good ping key %X at timestamp %u", (unsigned) key, provider.last_key_response_time);
```
clydebarrow:
```suggestion
        ESP_LOGV(TAG, "Found good ping key %X at timestamp %" PRIu32, (unsigned) key, provider.last_key_response_time);
```

esphome/components/packet_transport/packet_transport.h
```diff
@@ -5,7 +5,7 @@
 #ifdef USE_SENSOR
 #include "esphome/components/sensor/sensor.h"
 #endif
-#ifdef USE_BINARY_SENSOR
+#if defined(USE_BINARY_SENSOR) || defined(USE_STATUS_SENSOR)
```
clydebarrow:
```suggestion
#ifdef USE_STATUS_SENSOR
```
USE_BINARY_SENSOR must be defined if there is a binary sensor configured.

Also add USE_STATUS_SENSOR to core/defines.h

esphome/components/packet_transport/packet_transport.h
```diff
@@ -80,6 +84,7 @@ class PacketTransport : public PollingComponent {
       provider.last_code[0] = 0;
       provider.last_code[1] = 0;
       provider.name = hostname;
+      provider.last_key_response_time = 0;
```
clydebarrow: Change the definition of `Provider provider` to `Provider provider{}` which will zero-init all fields. Can then remove the explicit initialisers, and this ensures that the sensor pointer is null.

clydebarrow (changes requested):
You need to do more validation - if a status sensor is configured, then encryption and ping-pong need to be set.

clydebarrow:
> The test will fail as the `binary_sensor` headers have not been included

That's why I added it :-)

esphome/components/packet_transport/packet_transport.h
dala318: I'm not sure I agree.
Originally the line was `#ifdef USE_BINARY_SENSOR`
For that `esphome/components/binary_sensor/binary_sensor.h` was included to be used further down. That should still be the case.
For the status sensor the same base-definitions are used, and hence the include is needed.

clydebarrow: Sorry, should be the other way around - `#ifdef USE_BINARY_SENSOR`, i.e. what was there originally, is all that is required here (my comment was correct, just not the edit.) If USE_BINARY_SENSOR is not defined, you can't include the file.

dala318:
> You need to do more validation - if a status sensor is configured, then encryption and ping-pong need to be set.

Added dependency to ping-pong, which turn depends on encryption.

esphome/components/packet_transport/__init__.py
dala318: Removed

esphome/components/packet_transport/__init__.py
dala318: Changed according to proposal

esphome/components/packet_transport/packet_transport.cpp
dala318: Fixed according to comment

esphome/components/packet_transport/packet_transport.cpp
dala318: Added guard with USE_BINARY_SENSOR, but disables setting to False.

esphome/components/packet_transport/packet_transport.cpp
dala318: Fixed

esphome/components/packet_transport/packet_transport.h
dala318: Ok, reverted it to as originally, and depend on that adding a binary status sensor will automatically add the define USE_BINARY_SENSOR

esphome/components/packet_transport/packet_transport.h
dala318: Done

clydebarrow:
Still need to address Jesse's comment above. This will require adding a `binary_sensor.py` to allow configuration of the status sensor under the `binary_sensor:` key.

dala318:
> The test will fail as the `binary_sensor` headers have not been included as there is no top level `binary_sensor` in the config and it is not auto loaded (which we discourage these days)
> 
> It would be more config, but better to make it so it is:
> 
> ```yaml
> binary_sensor:
>   - platform: packet_transport
>     type: ping
>     provider_id: ...
>     ...
> ```
> 
> the binary sensor platform for packet_transport would need to be altered to use a cv.typed_schema and can have a default type of `sensor` or similar

The tricky part there that I'm struggling to resolve while implementing the proposed design above is that while verifying the new binary_sensor (type: ping) I don't have access to the general platform config which holds the keys `ping_pong_enable` and `providers>encryption`. (which both are preconditions to having a status-sensor)

It could be evaluated in runtime but then the user will not get any resonable warning at compiling, which I believe is much better.

Alternatively it could be solved in a "hacky" way to store away the platform settings in a global variable while validating that and validate the preconditions in the binary sensor code-generation.

Or do you have any additional alternative to solve it?

clydebarrow:
> I don't have access to the general platform config

You would do this in a FINAL_VALIDATE function, using `final_validate.full_config.get().get_path_for_id(<parent_id>)` to get the config for the parent component. Have a browse of the codebase for examples, e.g. `haier/switch/__init__.py` - see `_final_validate()`

dala318:
> > I don't have access to the general platform config
> 
> You would do this in a FINAL_VALIDATE function, using `final_validate.full_config.get().get_path_for_id(<parent_id>)` to get the config for the parent component. Have a browse of the codebase for examples, e.g. `haier/switch/__init__.py` - see `_final_validate()`

Ok, did not know about that one, will have a look.
Meanwhile I'm pushing these changes to a local PR /dala318/esphome/pull/17

dala318:
Have now merged the local PR where the status sensor is configured as a binary_sensor for platform with `type: status` and also updated the related PR for documentation to reflect this change.
Please have a look if you agree to change.

esphome/components/packet_transport/binary_sensor.py
```diff
@@ -1,19 +1,93 @@
...
+            raise cv.Invalid(
+                f"Provider {config[CONF_PROVIDER]} not found in {config[CONF_TRANSPORT_ID]}"
+            )
+    else:
+        # In current config logic this will never be reached since by enforcing ping-pong-enable
+        # at lest one provider needs to be configured, but keep it for future-proofing.
+        raise cv.Invalid(
+            f"No providers configured for {config[CONF_TRANSPORT_ID]}, "
+            f"but status sensor {config[CONF_ID]} requires one"
+        )
```
clydebarrow: I'd be inclined to reduce this to:

* check that there are providers configured
* check that the provider of interest is listed
* check that it has encryption configured

Then if _any_ of this fails, just issue the error that the provider requires encryption. Having several different error messages may lead to the user going through several edit steps, each getting a different message requiring another change. Identifying the ultimate goal immediately will help the user to know exactly what has to be done in a single step.

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -314,11 +314,41 @@ void PacketTransport::send_data_(bool all) {
...
+        for (auto &sensor : this->remote_sensors_[provider.first]) {
+          sensor.second->publish_state(NAN);
+        }
+#endif
+#ifdef USE_BINARY_SENSOR
+        /* Not possible to set a binary sensor unavailable. */
+        // for (auto &binary_sensor : this->remote_binary_sensors_[provider.first]) {
+        //   binary_sensor.second->publish_state(false);
+        // }
+#endif
```
clydebarrow: If USE_STATUS_SENSOR is defined, then USE_BINARY_SENSOR must be, so the guard here is redundant. Also, when it does become possible to set a binary sensor as unavailable, that should be done even if a status sensor is not configured which will require a different structure. So just remove this block for now, it's only confusing things.

clydebarrow (changes requested):

esphome/components/packet_transport/packet_transport.cpp
dala318: Ok, but should I then also remove or move the USE_STATUS_SENSOR so that sensors always become unavailable when ping-timeout is reached on provider?

clydebarrow: Ah, yes - the loop should be done if there are any sensors or a status sensor. Good pickup.

esphome/components/packet_transport/binary_sensor.py
dala318: Changed to one complete error message for all errors

esphome/components/packet_transport/packet_transport.cpp
dala318: Have a look and see if you agree with the changes

dala318: The proper implementation should be made available with esphome/esphome/pull/7016

esphome/components/packet_transport/binary_sensor.py
```diff
@@ -1,19 +1,83 @@
...
+        if p.get(CONF_NAME) == config[CONF_PROVIDER]
+    ):
+        if CONF_ENCRYPTION not in provider:
+            provider_encryption_enabled = False
+    else:
+        provider_encryption_enabled = False
+    if not (ping_pong_enabled and provider_encryption_enabled):
+        raise cv.Invalid(
+            f"Status sensor {config[CONF_ID]} requires provider {config[CONF_PROVIDER]} to have enctryption defined and ping-pong enabled for {config[CONF_TRANSPORT_ID]}"
+        )
```
clydebarrow: `next(` requires a default argument or it throws an exception if not found, which is never acceptable when validating the yaml. When testing, you need to test the failure paths as well as the success paths. That _can_ be done in ESPHome CI unit testing, but at a minimum do it locally before pushing the changes.

Printing the ids in the error message isn't useful because if the user hasn't provided specific ids they don't mean much.

The logic is a bit convoluted and can be simplified:

```suggestion
    if transport_config[CONF_PING_PONG_ENABLE] and any(
        CONF_ENCRYPTION in p
        for p in transport_config[CONF_PROVIDERS]
        if p[CONF_NAME] == config[CONF_PROVIDER]
    ):
        return config
    raise cv.Invalid(
        "Status sensor requires ping-pong to be enabled and the nominated provider to use encryption."
    )
```

esphome/components/packet_transport/binary_sensor.py
```diff
@@ -1,19 +1,83 @@
...
     var = await binary_sensor.new_binary_sensor(config)
     comp = await cg.get_variable(config[CONF_TRANSPORT_ID])
-    remote_id = str(config.get(CONF_REMOTE_ID) or config.get(CONF_ID))
-    cg.add(comp.add_remote_binary_sensor(config[CONF_PROVIDER], remote_id, var))
+    if config[CONF_TYPE] == CONF_DATA:
+        remote_id = str(config.get(CONF_REMOTE_ID) or config.get(CONF_ID))
+        cg.add(comp.add_remote_binary_sensor(config[CONF_PROVIDER], remote_id, var))
+    elif config[CONF_TYPE] == CONF_STATUS:
+        cg.add(comp.set_provider_status_sensor(config[CONF_PROVIDER], var))
+        cg.add_define("USE_STATUS_SENSOR")
```
clydebarrow: Since DATA is the default, test for STATUS first and output that, then output the DATA case in an `else` clause. There are no other possibilities.

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -314,11 +314,37 @@ void PacketTransport::send_data_(bool all) {
...
+      for (auto &sensor : this->remote_sensors_[provider.first]) {
+        sensor.second->publish_state(NAN);
+      }
+#endif
+      // Not possible to set a binary sensor unavailable, hence no equivalent loop
+      // as for the sensors above.
+    } else {
+#ifdef USE_STATUS_SENSOR
+      if (provider.second.status_sensor != nullptr && !provider.second.status_sensor->state) {
+        ESP_LOGI(TAG, "Ping status for %s restored at %u with age %u", provider.first.c_str(), now, key_response_age);
```
clydebarrow, line 345: Should be the same log level as the timeout log above. I'm not sure that WARN is the right level - it's not a failure, at least not in the device itself, so INFO would be better I think.

esphome/components/packet_transport/packet_transport.h
```diff
@@ -75,11 +79,7 @@ class PacketTransport : public PollingComponent {
 
   void add_provider(const char *hostname) {
     if (this->providers_.count(hostname) == 0) {
-      Provider provider;
-      provider.encryption_key = std::vector<uint8_t>{};
-      provider.last_code[0] = 0;
-      provider.last_code[1] = 0;
-      provider.name = hostname;
```
clydebarrow, line 82: You still need to set the provider name after creating the struct.

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -314,11 +314,37 @@ void PacketTransport::send_data_(bool all) {
...
+        provider.second.status_sensor->publish_state(false);
+      }
+#endif
+#ifdef USE_SENSOR
+      for (auto &sensor : this->remote_sensors_[provider.first]) {
+        sensor.second->publish_state(NAN);
+      }
+#endif
+      // Not possible to set a binary sensor unavailable, hence no equivalent loop
+      // as for the sensors above.
```
clydebarrow: It is possible now.

```
        sensor.second->invalidate_state();
```

esphome/components/packet_transport/packet_transport.cpp
```diff
@@ -474,12 +501,10 @@ void PacketTransport::process_(const std::vector<uint8_t> &data) {
...
-  ESP_LOGCONFIG(TAG,
-                "Packet Transport:\n"
-                "  Platform: %s\n"
-                "  Encrypted: %s\n"
-                "  Ping-pong: %s",
-                this->platform_name_, YESNO(this->is_encrypted_()), YESNO(this->ping_pong_enable_));
+  ESP_LOGCONFIG(TAG, "Packet Transport:");
+  ESP_LOGCONFIG(TAG, "  Platform: %s", this->platform_name_);
+  ESP_LOGCONFIG(TAG, "  Encrypted: %s", YESNO(this->is_encrypted_()));
+  ESP_LOGCONFIG(TAG, "  Ping-pong: %s", YESNO(this->ping_pong_enable_));
```
clydebarrow: Don't change this - what is in `dev` is correct.

clydebarrow (changes requested):

clydebarrow:

```yaml
external_components:
  - source: github://pr#8694
    components: [packet_transport]
    refresh: 1h
```

---
(Added by the PR bot)

esphome/components/packet_transport/binary_sensor.py
dala318: Fixed

esphome/components/packet_transport/packet_transport.cpp
dala318: Ok, reverting

esphome/components/packet_transport/packet_transport.cpp
dala318: Great news, will add

esphome/components/packet_transport/packet_transport.h
dala318: Yes, fixed

esphome/components/packet_transport/packet_transport.cpp
dala318: Changed to INFO for both

esphome/components/packet_transport/binary_sensor.py
```diff
@@ -1,19 +1,77 @@
...
+    if transport_config[CONF_PING_PONG_ENABLE] and any(
+        CONF_ENCRYPTION in p
+        for p in transport_config[CONF_PROVIDERS]
+        if p[CONF_NAME] == config[CONF_PROVIDER]
+    ):
+        return config
+    raise cv.Invalid(
+        "Status sensor requires ping-pong to be enabled and the nominated provider to use encryption."
+    )
+    return config
```
dala318: Remove this line, unreachable
(or swap if-statement above to catch the error and raise the exception)

clydebarrow:
```suggestion
```

clydebarrow: Not sure why you commented but didn't make the change?

clydebarrow (approved):

PR #8697, opened 2025-05-05
[modbus_controller] fix #6998

### LINKED ISSUE #6998: [Modbus Controller] multiple server_regiser addresses can be identical /overlap
### Identical Addresses
  
It is possible to define more than one `server_register` option containing identical address.

*This is unexpected behavior and will likely surprise users*

It appears that the first `server_register` that is configured with a particular address "wins" based on how the underlying C++ code handles the incoming modbus command requests

### Overlapping Addresses

It is possible to define `server_register` options with overlapping address ranges.

*This is unexpected behavior, but it is possible that some installations are taking advantage of this quirk*

2025.5.0-dev

Docker

NA  this can be seen without any HomeAssistant connections

ESP32

### Board

Lolin32-Lite

modbus_controller

### YAML Config

```yaml
### Duplicate Addresses


uart:
  - id: uart_modbus_server
    tx_pin: GPIO25
    rx_pin: GPIO35
    baud_rate: 57600

modbus:
  - uart_id: uart_modbus_server
    id: modbus_server
    role: server

modbus_controller:
  - modbus_id: modbus_server
    address: 0x1
    server_registers:
      - address: 0x0000
        value_type: U_WORD
        read_lambda: |-
          return(22);
      - address: 0x0000 # note the duplicate address
        value_type: U_WORD
        read_lambda: |-
          return(42);


### Overlapping Address Ranges



modbus_controller:
  - modbus_id: modbus_server
    address: 0x1
    server_registers:
      - address: 0x0000
        value_type: U_FP32
        read_lambda: |-
          return(22.0);
      - address: 0x0001 # note the overlapping address with the high word of the FP32 
        value_type: U_WORD
        read_lambda: |-
          return(42);
```

```txt

```

### Potential Fix for Identical Address Issue

In order to fix the identical addresses issue, it seems like a check can be added to the `modbus_controller/__init__.py` verification schema.

One possible solution is to create an empty set of server_register addresses and check on each server_register list item if the address exists in the set, if it does, fail the validation, if it is not in the set, add the register to the set.

### Potential Fix for Overlapping Addresses Issue

The overlapping register ranges quirk potentially allows for hidden registers  If individual modbus command 0x3 or 0x4 read commands are sent with the correct start address and with the correct number of registers, the current `modbus_controller` server component will allow these overlapping registers.

Fixing the overlapping ranges can also use a set of addresses, but instead of checking for an individual address in a set, create a check_set which is the range of addresses from `address .. (address + register_count(value_type) -1)`.  This check set can be inspected for intersection with the set of addresses.  if there are any intersections, fail the validation.  If the check set and set of addresses are disjoint then add the check set to the set of addresses.

> [!NOTE]
> It is possible that some installations are taking advantage of the quirk of register overlapping.

I am not sure if the overlapping check should be a configurable option or just forced on the users.

### Implications for future `Write*` Command Support

If server role support for 0x5, 0x6 or 0x10 modbus write (coil / register / registers) commands are added it would seem that this overlapping quirk will need to be eliminated.

#### Example

A block of 4 registers for a U_QWORD are reserved at address 0x0000 and then a U_WORD is reserved at address 0x001
```yaml
modbus_controller:
  - modbus_id: modbus_server
    address: 0x1
    server_write_registers:
      - address: 0x0000
        value_type: U_QWORD
        write_lambda: |-
          ESP_LOGD("write_lambda", "Updating address %u of QWORD\n", last_address);
      - address: 0x0001 # note the overlapping address with the U_QWORD
        value_type: U_WORD
        write_lambda: |-
          ESP_LOGD("write_lambda", "Updatint address %u of U_WORD\n", last_address);
```

What should happen when a write register command `0x6` is specified for register address  0x0001?

Should the second highest word in the U_QWORD be updated and its write_lambda be called, or should the U_WORD at address 0x0001 be updated and its write_lambda be called?

gotnone:
`server_register` address ranges may not overlap

This fixes https://github.com/esphome/issues/issues/6998

Additional validator `validate_address` added to `modbus_controller/__init__.py` 
- Check for overlapping address ranges in the the list of `server_registers`
- Calculate `last_address` based on `address` and `value_type` fields
- Check if `last_address` is `cv.uint16_t`
- Modify `address` verification to use `cv.uint16_t` instead of `cv.positive_int`

>[!NOTE]
> This PR may break user configurations that rely on the current quirk where the address ranges are permitted to overlap.

## Types of changes

- [X] Potentially Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

- fixes https://github.com/esphome/issues/issues/6998

- esphome/esphome-docs#4889

## Test Environment

- [X] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
modbus_controller:
  - modbus_id: modbus_server
    address: 0x1
    server_registers:
      - address: 0x0000
        value_type: U_QWORD
        read_lambda: |-
          return(22);
      - address: 0x0001
        value_type: U_WORD
        read_lambda: |-
          return(42);
```
## Output for correctly detected overlapping addresses

```
INFO ESPHome 2025.5.0-dev
INFO Reading configuration breakmodbus.yaml...
Failed config

modbus_controller: [source breakmodbus.yaml:50]
  - modbus_id: modbus_server
    address: 1
    server_registers:
      - address: 0
        value_type: U_QWORD
        return(22);

      Address range overlap for 1..1.
      - address: 1
        value_type: U_WORD
        return(42);
```

## Checklist:

 I don't know how to test to prove it prevents overlapping addresses, besides examining the results of a good configuration vs a configuration file with overlapping addresses

Added note specifying that `server_register` address ranges may not overlap


esphome/components/modbus_controller/__init__.py
```diff
@@ -126,13 +126,43 @@
...
+        if not register_set.isdisjoint(check_set):
+            raise cv.Invalid(f"Address range overlap for {address}..{last_address}")
+        register_set.update(check_set)
+
+        return register
+
+    return validator
+
+
+ModbusServerRegisterSet = set()
```
jesserockz: Please use `CORE.data[DOMAIN]["some key"]` for storing things like this.
Will need to add `DOMAIN = "modbus_controller` to the top of the file too

esphome/components/modbus_controller/__init__.py
```diff
@@ -126,13 +126,43 @@
 
 _LOGGER = logging.getLogger(__name__)
 
+
+def validate_address(register_set: set):
+    def validator(register: dict):
+        address = register[CONF_ADDRESS]
+        register_count = register.get(
+            CONF_REGISTER_COUNT, TYPE_REGISTER_MAP[register[CONF_VALUE_TYPE]]
+        )
```
jesserockz: `CONF_REGISTER_COUNT` is not a valid option in the `ModbusServerRegisterSchema` so this will always default to the second arg

jesserockz (changes requested):

esphome/components/modbus_controller/__init__.py
gotnone: OK.  I think I have this correct, now.

esphome/components/modbus_controller/__init__.py
gotnone: I extracted this code from my server_write_registers branch.

I am anticipating needing this check in the upcoming `server_write_registers` PR.  I can remove it, but then I'll just have to add it back in that PR.

I also have an upcoming PR in the works to fix issue https://github.com/esphome/issues/issues/7000.  It will also utilize this.

gotnone: OK removed.

esphome/components/modbus_controller/__init__.py
gotnone: I think I have this fixed like in the ble_server example

esphome/components/modbus_controller/__init__.py
```diff
@@ -126,13 +130,42 @@
 
 _LOGGER = logging.getLogger(__name__)
 
+
+def validate_address(register_set_key: str):
+    register_set = CORE.data.setdefault(DOMAIN, {}).setdefault(register_set_key, set())
+
+    def validator(register: dict):
```
jesserockz: Looking over this again, I think this needs to be inside the actual validator
```suggestion
    def validator(register: dict):
        register_set = CORE.data.setdefault(DOMAIN, {}).setdefault(register_set_key, set())```

gotnone: I had to modify this proposed fix to keep the linter happy, but now it is inside the validator

gotnone:
Modbus controller changes included in the 2025.7.0b beta still exhibit the address overlap issue.

Updated PR to compile with the latest modbus_controller code.

PR #8699, opened 2025-05-06
Improve stability for a test that crashes intermittently in CI

dala318:

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/tests/dashboard/test_web_server.py
+++ b/tests/dashboard/test_web_server.py
@@ -75,6 +75,9 @@ async def test_devices_page(dashboard: DashboardTestHelper) -> None:
     assert response.headers["content-type"] == "application/json"
     json_data = json.loads(response.body.decode())
     configured_devices = json_data["configured"]
-    first_device = configured_devices[0]
-    assert first_device["name"] == "pico"
-    assert first_device["configuration"] == "pico.yaml"
+    if len(configured_devices) == 0:
+        assert len(configured_devices) != 0
+    else:
+        first_device = configured_devices[0]
+        assert first_device["name"] == "pico"
+        assert first_device["configuration"] == "pico.yaml"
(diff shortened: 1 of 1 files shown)


tests/dashboard/test_web_server.py
```diff
@@ -75,6 +75,9 @@ async def test_devices_page(dashboard: DashboardTestHelper) -> None:
...
     configured_devices = json_data["configured"]
-    first_device = configured_devices[0]
-    assert first_device["name"] == "pico"
-    assert first_device["configuration"] == "pico.yaml"
+    if len(configured_devices) == 0:
+        assert len(configured_devices) != 0
+    else:
+        first_device = configured_devices[0]
+        assert first_device["name"] == "pico"
+        assert first_device["configuration"] == "pico.yaml"
```
nielsnl68: Why adding a if ... else: when you are planing to do an assert when the len(configured_devices) == 0 ?
From my understanding just calling ``assert len(configured_devices) != 0`` would be enough.
```suggestion
    assert len(configured_devices) != 0
    first_device = configured_devices[0]
    assert first_device["name"] == "pico"
    assert first_device["configuration"] == "pico.yaml"
```

dala318: The test crash happened on `configured_devices[0]` (which for some unknown reason was empty). Your suggestion does not avoid this.
The crash was more severe than just an test-failure, and blocked the test from finishing.

Have not seen the error in a while (nor the "assert fail") so may just skip this one.
(But was mighty annoying at the time it failed more than half of the runs)

nielsnl68: I get that.
Please be aware that you are testing two times the same thing.
First in your *if* statement and second within the *assert* call.

Remember the *assert* is basically an *if* statement which trigger an exception when the condition is false.

dala318: Got your point now, and revealing my limited pytest knowledge.
Did not realize test-case execution will halt on first assert error.
Then your proposal makes sense!

jesserockz (approved):
While this will stop the crash, it wont fix the random failure of this test. Still an improvement

PR #8701, opened 2025-05-06
[yaml] Added `!rel` directive to allow paths relative to yaml file

purrplingcat:

This small features solves loading of local `.h` files in `esphome.includes` (or local components) which are placed in (external) package via add a directive which translate relative path as relative to source yaml file instead of to main config yaml. By prefixing relative path with `!rel` tell esphome that file is relative to the current loaded yaml file. This relative path is transfered to an absolute path when requested yaml is loaded.

This could be usefull for other usecases which uses paths and you need that path to be relative to current yaml file instead of project main yaml file aka config directory.

```
.
├── common/
│   ├── base.yaml
│   └── includes/
│       └── colors.h
├── myproject.yaml
└── utils.h
```

```yaml
# common/base.yaml
esphome:
  includes:
    - utils.h # relative to project yaml file
    - !rel includes/colors.h # relative to current loaded yaml file
```

```yaml
# myproject.yaml
packages:
  base: !include common/base.yaml # Loads the package adding some C++ includes
```

**Result:** the `includes/colors.h` which is placed in `common/includes/colors.h` is sucessfully loaded and added to build instead of getting an error.

If the `!rel` is not the best name for this kind of directive, I am open to your suggestions.

## Types of changes

- [ ] Other

- esphome/esphome-docs#5322

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Checklist:


purrplingcat:
TODO for me:

- [x] Add tests
- [x] Write docs

fetzerch:
I think there's another interesting use case for this: Being able to specify relative paths to `external_components` within esphome project repositories that can be included via `package`.

I think there's even a use for this in the Voice Assistant PE repo in https://github.com/esphome/home-assistant-voice-pe/blob/739e47f4bfe671241cf766fd86584109493d161c/home-assistant-voice.yaml#L1562
Instead of being able to locally reference the directory, the component directory is referenced via an external git URL to the same repo.

I stumbled over this because I have a similar project setup ~~- just in a currently private repo where referencing via the external GitHub URL doesn't seem to work due to the credentials needed~~. It would be really nice if one could just use a relative local path like this:

```
external_components:
  - source:
    path: !rel esphome_components
    type: local
```

And I can actually confirm that your patch makes this work!

purrplingcat:
Will add `!rel` to docs, but open for review the code now.

purrplingcat:
Added docs for this feature, see esphome/esphome-docs#5322

jesserockz:
This may also close https://github.com/esphome/backlog/issues/8

purrplingcat:
I can't update it to a dev nranch due to merge conflicts. They are unclear for me to keep the dev working. I am asking for help with that.

jpeletier:
I don't want to spoil the PR, but rather than a new YAML directive, I'd rather have a "virtual" substitution variable containing the path (`Path`) of the file. That way it could be used anywhere, not just include files.

jpeletier:
Here is the alternative I am proposing: https://github.com/esphome/esphome/pull/12273
In essence, you get a built-in var, `this` and can use it like this for example:

`${this.dir}/mylib.h`

Full documentation [here](https://deploy-preview-5719--esphome.netlify.app/components/substitutions/#built-in-functions-and-vars).

Also, the variable can be used for many other things like logging:

```yaml
script:
  - id: myscript
    then:
      - logger.log: "My script in ${this.file.name} launched!"
      - logger.log: "Full path: ${this.file}"
```

PR #8719, opened 2025-05-08
Add missing ifdefs to api_pb2 files

bdraco:

TODO
- [ ] While it compiles a bit faster it doesn't seem to affect the flash size so it might already be optimized out making this PR pointless... more investigation needed

We had ifdefs in api_service but not in api_pb2 so we would compile in protobuf messages typed that the
device did not use

before
```
-rw-r--r--   1 bdraco  staff  3761924 May  8 03:18 api_pb2.cpp.o
```

after
```
-rw-r--r--   1 bdraco  staff  1982120 May  8 04:18 api_pb2.cpp.o
```

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


bdraco:
seems to make no difference

bdraco:
verified the compiler can figure out what is unused so lets not add any more complexity here

PR #8724, opened 2025-05-09
Optimize logger hot paths

bdraco:

  1. Use find() instead of count() + operator[] in level_for to reduce redundant map lookups.
  2. Replace character-by-character copying with memcpy in write_to_buffer_ for more efficient bulk copying.
  3. Pre-calculate the length of the reset color string in write_footer_() to avoid calling strlen() on every log message.

To test

```
external_components:
  - source: github://pr#8724
    components: [ logger ]
    refresh: 0s
```

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/logger/logger.cpp
+++ b/esphome/components/logger/logger.cpp
@@ -79,14 +79,22 @@ void Logger::log_vprintf_(int level, const char *tag, int line, const __FlashStr
   recursion_guard_ = true;
   this->reset_buffer_();
   // copy format string
+
+  // Optimization: Pre-calculate buffer capacity
   auto *format_pgm_p = reinterpret_cast<const uint8_t *>(format);
-  size_t len = 0;
+  const int max_format_len = this->buffer_remaining_capacity_() - 1;  // Reserve 1 for null terminator
+  int i = 0;
   char ch = '.';
-  while (!this->is_buffer_full_() && ch != '\0') {
-    this->tx_buffer_[this->tx_buffer_at_++] = ch = (char) progmem_read_byte(format_pgm_p++);
+
+  // Optimization: Perform a single boundary check outside the loop
+  // and combine the null-terminator check with the increment operation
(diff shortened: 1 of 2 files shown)


esphome/components/logger/logger.cpp
```diff
@@ -79,14 +79,22 @@ void Logger::log_vprintf_(int level, const char *tag, int line, const __FlashStr
   recursion_guard_ = true;
   this->reset_buffer_();
   // copy format string
+
+  // Optimization: Pre-calculate buffer capacity
```
bdraco: need to double check this one

esphome/components/logger/logger.cpp
```diff
@@ -102,16 +102,21 @@ void Logger::log_vprintf_(int level, const char *tag, int line, const __FlashStr
 #endif
 
 int HOT Logger::level_for(const char *tag) {
-  if (this->log_levels_.count(tag) != 0)
-    return this->log_levels_[tag];
```
bdraco, line 106: Was it intentional to call the operator here.. doesn't that mean every tag gets inserted if its not in the levels?

bdraco: That seems like a bug though since set_log_level can change at run time as there is a select entity, and if we don't have a level defined for the tag we end up mutating the map the first time its missing and than we would always end up with the default

bdraco: nevermind the original had a count check so it wouldn't put in a default and worked right.

bdraco: yeah new code has ONE map transversal, original code had TWO if it exists

bdraco:
this mostly gets refactored away after https://github.com/esphome/esphome/pull/8736 so closing

PR #8765, opened 2025-05-13
Logger Recursion Guard per Task on ESP32

### LINKED ISSUE #5102: Logger is not thread safe
### The problem

If a task is created using `xTaskCreate()` and the task creates log entries with `ESP_LOG`, not all will be visible in the log.

Reason:
The logger omits the entry, if another task is creating an entry at the same time.
`recursion_guard_`
https://github.com/esphome/esphome/blob/45276cc24452b7ea1a7a0c6f88ed11ca38a4a1e6/esphome/components/logger/logger.cpp#L67

Possible fix:
Replace `recursion_guard_` with a mutex created by `xSemaphoreCreateMutex` and claim/release it with `xSemaphoreTake` and `xSemaphoreGive`.

Working PoC:
```
static SemaphoreHandle_t log_guard = xSemaphoreCreateMutex();

void HOT Logger::log_vprintf_(int level, const char *tag, int line, const char *format, va_list args) {  // NOLINT
  if (level > this->level_for(tag))
    return;
  xSemaphoreTake(log_guard, portMAX_DELAY);
  this->reset_buffer_();
  this->write_header_(level, tag, line);
  this->vprintf_to_buffer_(format, args);
  this->write_footer_();
  this->log_message_(level, tag);
  xSemaphoreGive(log_guard);
}
```

2023.10.6

pip

_No response_

ESP32

### Board

esp32dev

logger

### Example YAML snippet

_No response_

_No response_

### Additional information

_No response_

bdraco:

The previous implementation used a global recursion guard, which caused log messages to be blocked in multi-task environments. To address this, we now leverage pthread thread-local storage (TLS) to track recursion state per task.  

## Key Changes  
- **Main Task:** Uses a dedicated boolean flag for minimal overhead.  
- **Other Tasks:** Utilize pthread TLS with a dynamically created key via `pthread_key_create`.  

## Design Highlights  
- Completely thread-safe--no locks, mutexes, or maps required.  
- Automatic state cleanup on task termination, handled by pthread TLS.  
- Improved reliability with significantly fewer dropped logs.  
- Backward compatible across all supported platforms.  
- Tested successfully with both IDF (4.4.x and 5.1.x) and Arduino.  

## Implementation Notes
- **Dynamic TLS Allocation:** The implementation uses `pthread_key_create` to dynamically allocate a TLS key.
- **No Fixed Key Limitations:** ESP-IDF's pthread implementation uses a linked list approach for TLS, allowing an unlimited number of keys (limited only by heap memory).
- **Efficient Per-Task State:** This approach enables per-task state tracking without the need for locks, mutexes, or maps.
- **FreeRTOS TLS Slots Not Used:** This implementation does not rely on the fixed FreeRTOS task local storage slots (CONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS).

## Limitations & Considerations  
- **libretiny Compatibility:** While libretiny shares similar recursion challenges, its limited multi-task capabilities (no Bluetooth) and constrained API make the impact less critical. Addressing this would require a different approach tailored to its specific limitations. 

## Testing

```
external_components:
  - source: github://pr#8765
    components: [ logger ]
    refresh: 0s
```

## Type of change

- [ ] Other

- fixes https://github.com/esphome/issues/issues/5102

## Test Environment

- [x] ESP32
- [x] ESP32 IDF 5.1 and 4.4
- [x] ESP8266
- [ ] RP2040
- [x] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/logger/__init__.py
+++ b/esphome/components/logger/__init__.py
@@ -254,6 +254,7 @@ async def to_code(config):
         config[CONF_TX_BUFFER_SIZE],
     )
     if CORE.is_esp32:
+        cg.add(log.create_pthread_key())
         task_log_buffer_size = config[CONF_TASK_LOG_BUFFER_SIZE]
         if task_log_buffer_size > 0:
             cg.add_define("USE_ESPHOME_TASK_LOG_BUFFER")
--- a/esphome/components/logger/__init__.py
+++ b/esphome/components/logger/__init__.py
@@ -258,6 +258,7 @@ async def to_code(config):
         if task_log_buffer_size > 0:
             cg.add_define("USE_ESPHOME_TASK_LOG_BUFFER")
             cg.add(log.init_log_buffer(task_log_buffer_size))
+
     cg.add(log.set_log_level(initial_level))
     if CONF_HARDWARE_UART in config:
         cg.add(
--- a/esphome/components/logger/__init__.py
+++ b/esphome/components/logger/__init__.py
@@ -261,10 +261,6 @@ async def to_code(config):
         if task_log_buffer_size > 0:
             cg.add_define("USE_ESPHOME_TASK_LOG_BUFFER")
             cg.add(log.init_log_buffer(task_log_buffer_size))
-    elif CORE.is_libretiny:
-        # For LibreTiny platforms, ensure we have enough thread local storage pointers
-        # Similar to ESP32, we need to ensure index 1 is available
-        cg.add_build_flag("-DCONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS=2")
     cg.add(log.set_log_level(initial_level))
     if CONF_HARDWARE_UART in config:
         cg.add(
--- a/esphome/components/logger/__init__.py
+++ b/esphome/components/logger/__init__.py
@@ -256,7 +256,8 @@ async def to_code(config):
     if CORE.is_esp32:
         # For ESP32 platforms, ensure we have at least 2 thread local storage pointers
         # Index 0 is used by pthread, we need index 1 for our logger recursion guard
-        add_idf_sdkconfig_option("CONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS", 2)
+        if CORE.using_esp_idf:
+            add_idf_sdkconfig_option("CONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS", 2)
         task_log_buffer_size = config[CONF_TASK_LOG_BUFFER_SIZE]
         if task_log_buffer_size > 0:
             cg.add_define("USE_ESPHOME_TASK_LOG_BUFFER")
--- a/esphome/components/logger/__init__.py
+++ b/esphome/components/logger/__init__.py
@@ -259,10 +259,6 @@ async def to_code(config):
             cg.add_define("USE_ESPHOME_TASK_LOG_BUFFER")
             cg.add(log.init_log_buffer(task_log_buffer_size))
 
-        # Ensure we have enough thread local storage pointers for the logger's per-task recursion guard
-        if CORE.using_esp_idf:
-            add_idf_sdkconfig_option("CONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS", 2)
-
     cg.add(log.set_log_level(initial_level))
     if CONF_HARDWARE_UART in config:
         cg.add(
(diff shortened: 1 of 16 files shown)


bdraco:
CONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS is 1 on arduino 

will need to come up with another way

esphome/components/logger/logger.cpp
```diff
@@ -69,7 +81,7 @@ void HOT Logger::log_vprintf_(int level, const char *tag, int line, const char *
 
   recursion_guard_ = false;
 }
-#endif  // !USE_ESP32
+#endif  // USE_ESP32
```
bdraco:
```suggestion
#endif  // !USE_ESP32
```

esphome/components/logger/logger.h
```diff
@@ -222,18 +222,18 @@ class Logger : public Component {
   std::map<std::string, int> log_levels_{};
   CallbackManager<void(int, const char *, const char *)> log_callback_{};
   int current_level_{ESPHOME_LOG_LEVEL_VERY_VERBOSE};
-#ifdef USE_ESP32
-  std::atomic<bool> recursion_guard_{false};
 #ifdef USE_ESPHOME_TASK_LOG_BUFFER
   std::unique_ptr<logger::TaskLogBuffer> log_buffer_;  // Will be initialized with init_log_buffer
 #endif
+#if defined(USE_ESP32)
```
bdraco:
```suggestion
#ifdef USE_ESP32
```

esphome/components/logger/logger.cpp
```diff
@@ -51,23 +58,21 @@ void HOT Logger::log_vprintf_(int level, const char *tag, int line, const char *
...
-#endif  // USE_ESP32
-
-#ifndef USE_ESP32
-// Implementation for platforms that do not support atomic operations
-// or have to consider logging in other tasks
+#else
+// Implementation for all other platforms
 void HOT Logger::log_vprintf_(int level, const char *tag, int line, const char *format, va_list args) {  // NOLINT
-  if (level > this->level_for(tag) || recursion_guard_)
+  if (level > this->level_for(tag) || global_recursion_guard_)
```
bdraco, line 67: It wasn't obvious that this is a global recursion guard which is what lead to the linked issue in the first place so I renamed it

esphome/components/logger/__init__.py
```diff
@@ -259,6 +259,10 @@ async def to_code(config):
             cg.add_define("USE_ESPHOME_TASK_LOG_BUFFER")
             cg.add(log.init_log_buffer(task_log_buffer_size))
 
+        # Ensure we have enough thread local storage pointers for the logger's per-task recursion guard
+        if CORE.using_esp_idf:
+            add_idf_sdkconfig_option("CONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS", 2)
```
bdraco: We already get 2 on Arduino (and we can't change it)

esphome/components/logger/logger.cpp
```diff
@@ -51,23 +58,21 @@ void HOT Logger::log_vprintf_(int level, const char *tag, int line, const char *
   }
 #endif  // USE_ESPHOME_TASK_LOG_BUFFER
 
-  recursion_guard_.store(false, std::memory_order_release);
+  // Reset the recursion guard for this task
+  this->reset_task_log_recursion_(is_main_task);
```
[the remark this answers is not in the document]
bdraco: Not a bad idea but not for this PR. Also would need to consider the performance impact, the additional stack usage, and cpu cycles for each logger call.  That could be impactful in VERBOSE or VERY_VERBOSE.

esphome/components/logger/logger_esp32.cpp
```diff
@@ -209,6 +209,8 @@ const char *const UART_SELECTIONS[] = {
 
 const char *Logger::get_uart_selection_() { return UART_SELECTIONS[this->uart_]; }
 
+void Logger::create_pthread_key() { pthread_key_create(&LOG_RECURSION_KEY, nullptr); }
+
```
bdraco:
```suggestion
```

esphome/components/logger/logger.h
```diff
@@ -119,6 +135,9 @@ class Logger : public Component {
   /// Set up this component.
   void pre_setup();
   void dump_config() override;
+#ifdef USE_ESP32
+  void create_pthread_key();
+#endif
```
bdraco:
```suggestion
```

jesserockz (approved):

bdraco:
Thanks for the review, and thanks for catching the TLS issue

PR #8769, opened 2025-05-13
[sen5x] Fix validation for values read from hardware

### LINKED ISSUE #59: when air-1 is power cycled, it reports impossible out-of-range NOX values
I've noticed when I power-cycle my air-1, it sends multiple impossible and invalid values for NOX to home assistant.

Air-1 factory firmware: 24.10.11.1 (ESPHome 2024.12.2)

NOX is an index and the only valid values are integers between 1-500 inclusively. https://sensirion.com/media/documents/9F289B95/6294DFFC/Info_Note_NOx_Index.pdf

Image

These impossible out-of-range NOX values are then saved into the statistics and history which leads to borked 🙈 data, charts, and reports.

This is likely due to the low-level ESPHome module not correctly handling power-on. Official Sensirion
https://sensirion.com/media/documents/2B6FC1F3/649C3D0E/PS_AN_Read_RHT_VOC_and_NOx_RAW_signals_v2_D1.pdf 

> RAW NOx:  During the first 10..11 seconds after power-on or device reset, this value will be 0xFFFF as well.

I suspect the ESP code needs to filter the data. When it gets values outside the range of 1-500 inclusively, to not report data or to report a silent error like `unavailable`.

Note that I have seen values other than FFFF reported to home assistant. Looking at the SQL tables directly for the NOX data, I see a value `3277`
in the states table...
Image

in the short term statistics table

Image

and statistics table

Image

### LINKED ISSUE #7017: when sen5x (air-1) is power cycled, it reports impossible out-of-range NOX values
### The problem

upstreaming issue https://github.com/ApolloAutomation/AIR-1/issues/59

air-1 sensor that uses the send5x it sends multiple impossible and invalid values for NOX to home assistant.

Air-1 factory firmware: 24.10.11.1 (ESPHome 2024.12.2)

NOX is an index and the only valid values are integers between 1-500 inclusively. https://sensirion.com/media/documents/9F289B95/6294DFFC/Info_Note_NOx_Index.pdf

Image

These impossible out-of-range NOX values are then saved into the statistics and history which leads to borked 🙈 data, charts, and reports.

This is likely due to the low-level ESPHome module not correctly handling power-on. Official Sensirion https://sensirion.com/media/documents/2B6FC1F3/649C3D0E/PS_AN_Read_RHT_VOC_and_NOx_RAW_signals_v2_D1.pdf

> RAW NOx: During the first 10..11 seconds after power-on or device reset, this value will be 0xFFFF as well.

I suspect the ESP code needs to filter the data. When it gets values outside the range of 1-500 inclusively, to not report data or to report a silent error like unavailable.

Note that I have seen values other than FFFF reported to home assistant. Looking at the SQL tables directly for the NOX data, I see a value 3277
in the states table...

Image

in the short term statistics table

Image

and statistics table

Image

You'l notice the error data is 3277. That is likely 32767 which is being scaled by .1 to 3276.7 and then rounded to 3277.
and of course, 32767 is 0x7FFF

Their bug is likely here https://github.com/esphome/esphome/blob/bec9d914199035fe765fc8fdd746ced566eadca2/esphome/components/sen5x/sen5x.cpp#L361
where they are doing exactly as I guessed. Scaling by 10 and checking for 0xFFFF.
I'll bet the sen5x hardware is sending 0x7FFF and therefore the xFFFF check fails.

The fix is to more safely filter the values. Any value outside 1-500 is invalid and therefore it should be filtered to be within those limits.

2024.12.2

Home Assistant Add-on

2025.5.1

ESP32

### Board

apollo air-1

sen5x

### YAML Config

```yaml
https://github.com/ApolloAutomation/AIR-1/tree/main/Integrations/ESPHome
```

```txt

```

### Additional information

_No response_

kbx81:

Humidity, temperature, VOC and NOx values are signed and were not being validated correctly, resulting in bad values being reported in some situations. See Sensiron's example code [here](https://github.com/Sensirion/embedded-i2c-sen5x/blob/4584621fc5a4efbcbe04b1712a759ff657a522fb/sen5x_i2c_example_usage.c#L164-L185).

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/issues/issues/7017
- fixes https://github.com/ApolloAutomation/AIR-1/issues/59

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


kbx81:
Test with:
```yaml
external_components:
  - source: github://pr#8769
    components: [sen5x]
    refresh: 1h
```

daradib:
It looks like the code still treats humidity, temperature, VOC, and NOx values as unsigned (`uint16_t measurements[8];`). Meaning that if temperature is -1 Celsius (admittedly an unusual case), it will not be reported correctly.

FYI this behavior is now explicitly documented in the [SEN6x datasheet](https://sensirion.com/media/documents/FAFC548D/6731FFFA/Sensirion_Datasheet_SEN6x.pdf) but I guess they had left it out of the [SEN5x datasheet](https://sensirion.com/media/documents/6791EFA0/62A1F68F/Sensirion_Datasheet_Environmental_Node_SEN5x.pdf):

> If no data is available at all (e.g. measurement not running for at least one second), all values will be at their upper limit (0xFFFF for uint16, 0x7FFF for int16)...During the first 10..11 seconds after power-on or device reset, this value [NOx Index] will be 0x7FFF as well.

esphome/components/sen5x/sen5x.cpp
```diff
@@ -337,47 +338,71 @@ void SEN5XComponent::update() {
       ESP_LOGD(TAG, "read data error (%d)", this->last_error_);
       return;
     }
+    ESP_LOGVV(TAG, "pm_1_0 raw = 0x%.4x", measurements[0]);
```
diablodale: I caution using the term "raw" in these logging statements. Because sensirion raw measurements are different than sensirion measurements. The measurements you have in the measurements[] are not the raw ones. They are the post-processed measurements.

esphome/components/sen5x/sen5x.cpp
```diff
@@ -337,47 +338,71 @@ void SEN5XComponent::update() {
...
       pm_4_0 = NAN;
+    }
+    ESP_LOGVV(TAG, "pm_10_0 raw = 0x%.4x", measurements[3]);
     float pm_10_0 = measurements[3] / 10.0;
-    if (measurements[3] == 0xFFFF)
+    if (measurements[3] == UINT16_MAX) {
       pm_10_0 = NAN;
+    }
+    ESP_LOGVV(TAG, "humidity raw = 0x%.4x", measurements[4]);
     float humidity = measurements[4] / 100.0;
```
diablodale: I recommend casting humidity, temp, voc, and nox to `int16` for consistency and to self-document. All four of these post-processed measurements are `int16`, not just temp.
Since this is cpp code, I also recommend a reinterpret_cast instead of legacy c-cast. That is, unless ESPHome has a style rule that writes to use c-casts.

diablodale (changes requested):
In general, this aligns with the research results I've documented in the related issue https://github.com/esphome/issues/issues/7017#issuecomment-2876381804.
I recommend two tweaks to the code, as in the review details

diablodale:
> Test with:
> 
> ```yaml
> external_components:
>   - source: github://pr#8769
>     components: [sen5x]
>     refresh: 1h
> ```

Unable to test as when I use this config...

```yaml
substitutions:
  name: apollo-air-1-0e4b5c
  friendly_name: Apollo AIR-1 0e4b5c
packages:
  ApolloAutomation.AIR-1: github://ApolloAutomation/AIR-1/Integrations/ESPHome/AIR-1.yaml@24.10.11.1
esphome:
  name: ${name}
  name_add_mac_suffix: false

external_components:
  - source: github://pr#8769
    components: [sen5x]
    refresh: 1h

logger:
  level: VERY_VERBOSE

api:
  encryption:
    key: <redacted>
wifi:
  ssid: !secret wifi_ssid
```

It fails to install with:

```
INFO ESPHome 2025.4.2
INFO Reading configuration /config/esphome/apollo-air-1-0e4b5c.yaml...
ERROR Unable to import component sen5x.sensor:
  File "/esphome/esphome/loader.py", line 181, in _lookup_module
  File "/usr/lib/python3.11/importlib/__init__.py", line 126, in import_module
  File "/data/external_components/2f5fcacd/esphome/components/sen5x/sensor.py", line 6, in <module>
    from esphome.const import (
ImportError: cannot import name 'CONF_GAIN_FACTOR' from 'esphome.const' (/esphome/esphome/const.py)
Failed config

sensor.sen5x: [source /data/packages/b3b7f414/Integrations/ESPHome/AIR-1.yaml:178]
  
  Platform not found: 'sensor.sen5x'.
  platform: sen5x
  id: sen55
  pm_1_0: 
    name: PM <1µm Weight concentration
    id: pm_1_0
    accuracy_decimals: 1
  pm_2_5: 
    name: PM <2.5µm Weight concentration
    id: pm_2_5
    accuracy_decimals: 1
  pm_4_0: 
    name: PM <4µm Weight concentration
```

esphome/components/sen5x/sen5x.cpp
kbx81: So I might clarify what I was after with these log messages a bit.

I'm a bit old-school. When I'm troubleshooting hardware like this, I want to see the "raw" bits coming off of the bus. I do logging this way--and using hexadecimal--so I can see that. I don't care about pretty formatting, signs, etc. I just want to see, bit-by-bit, what came off the wire.

So, about casting these -- they're some bytes, [read from the device/bus byte-by-byte](https://github.com/esphome/esphome/blob/032949bc779880241421b8ee5c56af16adc81ddf/esphome/components/sensirion_common/i2c_sensirion.cpp#L17-L31) and then [just "dropped" into this `uint16_t` buffer](https://github.com/esphome/esphome/blob/032949bc779880241421b8ee5c56af16adc81ddf/esphome/core/helpers.h#L191-L193) we have and I'm printing it here as such. It's rough-around-the-edges and not "proper", but it gave me exactly what I wanted to see. That said, this won't be useful for everybody as we all have our own approach to troubleshooting, so I'm of course happy to change it. 🙂

esphome/components/sen5x/sen5x.cpp
kbx81: Ah! I had noticed that in the docs but didn't really connect that here. Easy fix. 🙂

kbx81:
> It fails to install with:
> ```
> ImportError: cannot import name 'CONF_GAIN_FACTOR' from 'esphome.const' (/esphome/esphome/const.py)
> Failed config
> ```

Cool, apparently we have a constant moved in `dev`... 🙈

esphome/components/sen5x/sen5x.cpp
diablodale: - logging good, 100% with you to see 2-bytes in hex with very-verbose. And you wrote removing the word "raw" was ok too so "senserion raw measurement" and the "senserios measurement" isn't misunderstood again
- my feedback on casting is NOT in your logging statement. Instead, I request you cast the math for measurements[4,5,6,7] all like this...

`float humidity = reinterpret_cast<int16>(measurements[4]) / 100.0;`

or if you really dig c-casts...
`float humidity = (int16_t)measurements[4] / 100.0;`

This bug was due to the implementer not understanding the signed-ness and distiction with the "senserion raw measurement" and the "senserios measurement". Casting here on measurements 4,5,6,7 make the code self-documenting that these ARE signed and should be treated that way.

diablodale:
> Cool, apparently we have a constant moved in `dev`... 🙈

I see a related commit 2 weeks ago https://github.com/kbx81/esphome/commit/a31a5e74bdfa3ece19c502a91f81b0d93414bd67

Anything I need to do?

esphome/components/sen5x/sen5x.cpp
kbx81: Got it! Will update and push a change 🙂

kbx81:
> Anything I need to do?

If you want, you can try building & running with `dev` -- it's pretty simple and instructions are [here](https://esphome.io/guides/faq#how-do-i-use-the-latest-bleeding-edge-version).

diablodale:
> > Anything I need to do?
> 
> If you want, you can try building & running with `dev` -- it's pretty simple and instructions are [here](https://esphome.io/guides/faq#how-do-i-use-the-latest-bleeding-edge-version).

There is nothing on the table of contents at that link with the word "dev". This feels outside the bounds of the sen5x issue I reported

kbx81:
> > > Anything I need to do?
> > 
> > 
> > If you want, you can try building & running with `dev` -- it's pretty simple and instructions are [here](https://esphome.io/guides/faq#how-do-i-use-the-latest-bleeding-edge-version).
> 
> There is nothing on the table of contents at that link with the word "dev". This feels outside the bounds of the sen5x issue I reported

That's strange, the link I provided should've taken you directly to the relevant section on the page (it seems to work fine for me here).

In any event, the "How do I use the latest bleeding-edge version?" section explains how you can use ESPHome `dev`, the aim here being that you can then complete your build, allowing you to test this change on your hardware. Normally this extra effort wouldn't be required, but we don't have an alternative mechanism right now (though one is in the works) to cope with that constant having been moved.

esphome/components/sen5x/sen5x.cpp
kbx81: Ok, `reinterpret_cast` won't work here because:
```
src/esphome/components/sen5x/sen5x.cpp: In lambda function:
src/esphome/components/sen5x/sen5x.cpp:362:22: error: invalid cast from type 'uint16_t' {aka 'short unsigned int'} to type 'int16_t' {aka 'short int'}
  362 |     float humidity = reinterpret_cast<int16_t>(measurements[4]) / 100.0;
```
I think `static_cast` is the next best thing, so I've used that, instead.

esphome/components/sen5x/sen5x.cpp
kbx81: [`0a0d135` (#8769)](https://github.com/esphome/esphome/pull/8769/commits/0a0d135daa7664cfed597dad2cab1d01eef91192)

diablodale:
> That's strange, the link I provided should've taken you directly to the relevant section on the page (it seems to work fine for me here).

I agree...I blame it on android chrome. Or me. Anyhoo, I see your instructions now and esphome dev installed, worked around ota password prob, and now I have the config I pasted above (+ ota password) installed on my air-1. I have not powercycled the air-1, instead it was flashed in esphome dev ota and immediately showing me the logs which I set to VV.

This is the first block of sen5x data. We can see the `0x03c4` i2c command for post-processed measurements and the vv logs you added.
```
[18:51:34][VV][i2c.arduino:176]: 0x69 TX 03C4
[18:51:34][VV][i2c.arduino:148]: 0x69 RX 0008380009090009090009090DA5C814770302589F000A5A
[18:51:34][VV][sen5x:341]: pm_1_0 = 0x0008
[18:51:34][VV][sen5x:346]: pm_2_5 = 0x0009
[18:51:34][VV][sen5x:351]: pm_4_0 = 0x0009
[18:51:34][VV][sen5x:356]: pm_10_0 = 0x0009
[18:51:34][VV][sen5x:361]: humidity = 0x0da5
[18:51:34][VV][sen5x:366]: temperature = 0x1477
[18:51:34][VV][sen5x:371]: voc = 0x0258
[18:51:34][VV][sen5x:376]: nox = 0x000a
```

I have 49 of these blocks as I'm writing this. All of them have `nox = 0x000a` which is decimal 10 which is scaled by 10 so 10/10 = 1.0 which shows up a few lines down in the logs as `'SEN55 NOX': Sending state 1.00000`.

I see reasonable data in homeassistant itself -- curves and trends of entity charts are aligned. HA continues to believe this is the same device and it receiving the many entity data streams.

The esphome air-1 VV logs seem reasonable since it was the ESP that was flashed and rebooted. The sen5x is independent on the i2c bus and it has for days been beyond its startup behavior. I will powercycle the air-1 after this post. That should trigger the startup behavior of the sen5x. I do not know if I can connect to the air-1 fast enough to see the invalid 0x7fff values in the log.

Attached is the log showing the compile and the blocks I write about above.
[logs_apollo-air-1-0e4b5c_run.zip](https://github.com/user-attachments/files/20232080/logs_apollo-air-1-0e4b5c_run.zip)

diablodale:
FYI these tests today were via esphome 2025.6.0-dev20250515, and this PR with the 3 commits at 8c001aab56ea0af191fc5c1604aacf1f0fee6005

I could not connect to the air-1 fast enough to see the first log entries for the sen5x.
[logs_apollo-air-1-0e4b5c_logs after powercycle.zip](https://github.com/user-attachments/files/20232413/logs_apollo-air-1-0e4b5c_logs.after.powercycle.zip)

On line 66 of above log file you can see `'PM 4 To 10 µm': Received new state 0.000000` appearing out of sequence. This means one or more sen5x readings have already occurred and are not captured in this log file. The readings I can see in the file have `nox = 0x0000` and then later `nox = 0x000a`.

- 👍Home assistant *does not* show an impossible `3277` nox value.
- 🫤Home assistant shows the impossible `0` nox value. Validating all values as doc'd https://github.com/esphome/issues/issues/7017#issuecomment-2876381804 will fix this and other invalid values in all 8 of the measurements.

I powercycled a 2nd time, and again I didn't see `3277` but I did see `0`.
[logs_apollo-air-1-0e4b5c_logs after powercycle 2.zip](https://github.com/user-attachments/files/20232662/logs_apollo-air-1-0e4b5c_logs.after.powercycle.2.zip)

image

kbx81:
@diablodale I've extended the validation for VOC and NOX values based on your findings and Trevor's approval [here](https://github.com/ApolloAutomation/AIR-1/issues/59#issuecomment-2885110967). If you agree with these changes, feel free to mark the PR as "approved" and I'll merge so we can get it into the coming release. 🙂

esphome/components/sen5x/sen5x.cpp
```diff
@@ -25,6 +25,9 @@ static const uint16_t SEN5X_CMD_TEMPERATURE_COMPENSATION = 0x60B2;
 static const uint16_t SEN5X_CMD_VOC_ALGORITHM_STATE = 0x6181;
 static const uint16_t SEN5X_CMD_VOC_ALGORITHM_TUNING = 0x60D0;
 
+static const uint16_t SEN5X_MAX_INDEX_VALUE = 500;
```
diablodale: Any constant that will later be used in a comparison with VOC and NOX should be signed. It is almost always unsafe to compare signed with unsigned, or unsigned with signed.

esphome/components/sen5x/sen5x.cpp
```diff
@@ -370,12 +373,12 @@ void SEN5XComponent::update() {
     }
     ESP_LOGVV(TAG, "voc = 0x%.4x", measurements[6]);
     float voc = static_cast<int16_t>(measurements[6]) / 10.0;
-    if (measurements[6] == INT16_MAX) {
+    if (voc < SEN5X_MIN_INDEX_VALUE || voc > SEN5X_MAX_INDEX_VALUE) {
```
diablodale: Comparing floats like this is not safe. float representations and calculations often have very very small variations which makes simple comparisons unreliable. Compares where one value is `1.0` might work more often as `1` has a special representation in IEEE floats, but `500.0` does not and can vary often.

Both the logic here for VOC and NOX below should do their comparisons with the signed integer value in measurements[]...not its float transformation. For example...

```c++
// change to signed and scale by 10 for later comparison
static const int16_t SEN5X_MAX_INDEX_VALUE = 500 * 10;  // maximum index value, scaled up by 10
static const int8_t SEN5X_MIN_INDEX_VALUE = 1 * 10;     // minimum index value, scaled up by 10

// ...existing code...

// safer compare, assign voc only once
ESP_LOGVV(TAG, "voc = 0x%.4x", measurements[6]);
int16_t voc_idx = static_cast<int16_t>(measurements[6]);
float voc = (voc_idx < SEN5X_MIN_INDEX_VALUE || voc_idx > SEN5X_MAX_INDEX_VALUE) ? NAN : static_cast<float>(voc_idx) / 10.0f;

ESP_LOGVV(TAG, "nox = 0x%.4x", measurements[7]);
int16_t nox_idx = static_cast<int16_t>(measurements[7]);
float nox = (nox_idx < SEN5X_MIN_INDEX_VALUE || nox_idx > SEN5X_MAX_INDEX_VALUE) ? NAN : static_cast<float>(nox_idx) / 10.0f;
```

diablodale (commented):
the float compare approach is unsafe. Below I detail it and offer fixes

esphome/components/sen5x/sen5x.cpp
kbx81: Very good. Changes made in [`af33709` (#8769)](https://github.com/esphome/esphome/pull/8769/commits/af3370972d654282512db7f4392bd0581bb4340c).

I also switched the preceding lines to use the ternary operator for consistency...and it's a bit cleaner, anyway. 🙂

esphome/components/sen5x/sen5x.cpp
kbx81: Updated in [`af33709` (#8769)](https://github.com/esphome/esphome/pull/8769/commits/af3370972d654282512db7f4392bd0581bb4340c)

diablodale:
Good news, and some dev2dev feedback.

1. This passes my testing and I no longer see invalid values for NOX or VOC
2. The code can be slightly faster and smaller if you implement a tiny code change in the feedback below

## Good after flash no powercycle

nox during flash

logfile [after flash no powercycle.zip](https://github.com/user-attachments/files/20265455/after.flash.no.powercycle.zip)

NOX did not show any invalid values after flashing with the safer compares. Other measurements also returned reasonable values.

## Good powercycle values, shows NAN working

nox after powercycle

logfile [after powercycle.zip](https://github.com/user-attachments/files/20265462/after.powercycle.zip)

## Dev2dev feedback

I saw earlier in the week you using an `uint8_t` for `1`, which suggested to me you are intending to write code with memory and perf in mind. Given that, the code I proposed intentionally used `10.0f` and casting to float. You used that for NOX and VOC but didn't propogate it to the additional changes you made. I recommend you add `f` and cast to avoid `double`. Why?

- Most modern compilers like gcc in the ESPHome platform, treat float-ish constants in C++ code as `double`. Only when `f` is appended does the compiler use a `float`. This saves 4 bytes to represent `10.0` as a 32-bit float instead of a 64-bit float.
- More importantly, when a `double` is used, the compiler promotes all values into `double`. This results in multiple promotions and 64-bit division and then converting the results down to 32-bit. In our sen5x scenario, there is no need for 64-bit division by ten.
- And on the ESP32, all this 64-bit float work is done in *software*, not on the FPU.
- https://godbolt.org/z/MT48Wao8q shows the same two calculations and the only C++ difference is the `f` on the ten. But the compiled assembly is drastically different. The `f` float approach is less code and keeps everything in 32-bit and on the FPU. Details I write in the code comments at the godbolt

kbx81:
> Good news, and some dev2dev feedback.
> 
> 1. This passes my testing and I no longer see invalid values for NOX or VOC

Awesome! Glad it's fixed.

> 2. The code can be slightly faster and smaller if you implement a tiny code change in the feedback below
> ...

Haha, keeping the `(u)int` size appropriate for the value is largely just a personal nitpick of mine; sometimes there are reasons to use wider-than-needed sizes but, more often than not, they're just copypasta (which I'm absolutely guilty of myself now and then).

In terms of the missing `f` -- simple oversight on my part! Thank you for the related detail -- _some_ of this I knew, but some I didn't. Regardless, this is something I'll pay more attention to in the future! I've added this change in here, as well.

diablodale:
Looks good to me. I have no further feedback.

I checked this PR at 3e44c1994af8bacf93502044823defa44fbd1f05 which includes the float constants.

- compile, flashed air-1: data after flash looks good -- similar to the data and graphs earlier today
- powercycled air-1: data after powerup looks good -- also similar to earlier results

The invalid 0x0000 values are being caught and changed to NAN which appears as "unavailable" in home assistant.
I haven't been able to see the 0x7fff, I believe this is because the code is working and I can't connect to log fast enough.

diablodale (approved):
lgtm

jesserockz (approved):

PR #8782, opened 2025-05-13
[online_image] Last-Modified-Date and ETag response caching

candrews:

Use the Last-Modified-Date and ETag response headers (if present) when making subsequent requests to implement conditional response caching. If the data on the server is not modified, then a 304 Not Modified response results and the online_image component will reuse the previously received data instead of downloading and processing it again.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching

The `on_download_finished` callback now has a boolean argument provided that is `true` when a new image is downloaded (cache miss); it's `false` when there is a cache hit.

## Types of changes

- [ ] Other

- https://github.com/esphome/esphome-docs/pull/4906

## Test Environment

- [X] ESP32
- [X] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:

### DIFF
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -3,10 +3,10 @@
 #include "esphome/core/log.h"
 
 static const char *const TAG = "online_image";
-static const std::string ETAG_HEADER_NAME = "etag";
-static const std::string IF_NONE_MATCH_HEADER_NAME = "if-none-match";
-static const std::string LAST_MODIFIED_HEADER_NAME = "last-modified";
-static const std::string IF_MODIFIED_SINCE_HEADER_NAME = "if-modified-since";
+static const char *const ETAG_HEADER_NAME = "etag";
+static const char *const IF_NONE_MATCH_HEADER_NAME = "if-none-match";
+static const char *const LAST_MODIFIED_HEADER_NAME = "last-modified";
+static const char *const IF_MODIFIED_SINCE_HEADER_NAME = "if-modified-since";
 
 #include "image_decoder.h"
 
--- a/esphome/components/online_image/online_image.cpp
+++ b/esphome/components/online_image/online_image.cpp
@@ -217,10 +217,10 @@ void OnlineImage::loop() {
     ESP_LOGD(TAG, "Image fully downloaded, read %zu bytes, width/height = %d/%d", this->downloader_->get_bytes_read(),
              this->width_, this->height_);
     ESP_LOGD(TAG, "Total time: %lds", ::time(nullptr) - this->start_time_);
-    this->end_connection_();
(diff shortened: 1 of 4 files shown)


esphome/components/online_image/online_image.h
```diff
@@ -173,6 +173,14 @@ class OnlineImage : public PollingComponent,
...
    */
   int buffer_height_;
+  /**
+   * The value of the ETag HTTP header provided in the last response.
+   */
+  std::string etag_ = "";
+  /**
+   * The value of the Last-Modified HTTP header provided in the last response.
+   */
+  std::string last_modified_ = "";
```
jesserockz, line 185: While it wont make a huge difference, these should to be reset if the url is changed `set_url(...)` to allow the next request to be smaller and not send irrelevant headers

esphome/components/http_request/http_request.cpp
jesserockz, line 1: Added for debugging during development?

jesserockz (changes requested):

esphome/components/http_request/http_request.cpp
candrews: I shouldn't have pushed that commit - I removed it.

esphome/components/online_image/online_image.h
candrews: Great point! I've done that: 5bf871dfd82bedc154be2416b90413f520993d36

candrews:
In my testing, this PR works great. I'm eager to get it merged, please let know what I can do. Thank you in advance for your review! :)

esphome/components/online_image/online_image.cpp
```diff
@@ -3,6 +3,10 @@
 #include "esphome/core/log.h"
 
 static const char *const TAG = "online_image";
+static const std::string ETAG_HEADER_NAME = "etag";
+static const std::string IF_NONE_MATCH_HEADER_NAME = "if-none-match";
+static const std::string LAST_MODIFIED_HEADER_NAME = "last-modified";
+static const std::string IF_MODIFIED_SINCE_HEADER_NAME = "if-modified-since";
```
clydebarrow: Declaring this way results in a copy of the string in flash, which is then used at startup to construct a `std::string` object in RAM, which means a copy of the data plus the class overhead permanently occupying RAM.

These should instead just be `const char * const` - when passed to a function requiring a `std::string` it will still be copied, but this will be a transient allocation instead of a permanent chunk of RAM being used up.

candrews: I didn't know that, thank you explaining. I've made that change.

jesserockz (approved):

PR #8793, opened 2025-05-14
[core] Ability of dynamically creating of components, sensors etc in setup() method of component

zkoalexey:

I’m interested in the possibility of dynamically creating components during the initial setup at startup. I’ve discovered that dynamic component creation is possible, but for correct configuration it requires re-sorting the remaining components: in this process, the API component (sort weight 200) gains the ability to detect the newly added components (average sort weight more than 200).

As an example, I’ve implemented the Dallas Automatic Temperature Searcher component (PR #8794), which should automatically locate an arbitrary number of sensors on the OneWire bus and add them to the App as components and sensors for use with Home Assistant. 

I believe this could significantly extend esphome capabilities.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [x] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


jesserockz:
ESPHome always works under the fact that "you get only what you configure". 
I don't believe we should add dynamic entities which will break this assumption and open the potential for more issues down the line.

zkoalexey:
Esphome is a great product that can cross that line, I think. The PR changes cost (almost) nothing now, but this will give an opportunity to enthusiasts develop such components and experiment with this feature. Although they will not be accepted into the main repository at this time, they can be used using `external components`. The community will be able to accumulate experience about it, find corner-cases. And then decide whether it's worth it or not. But without this changes it is impossible.

Although if external component will be able to replace the folder core?..

jesserockz:
Recent changes in the ESPHome core code has made the compoenent and entity vectors a static size and it cannot increase at runtime anymore making this PR ususable.

- #10018
- #10020

PR #8804, opened 2025-05-15
Reduce number of calls to fetch time in the main loop

bdraco:

This optimization reduces the number of calls to `millis()` in the main loop by caching time values for each component's loop execution. Each component now uses `App.get_loop_component_start_time()` instead of directly calling `millis()`.

Note: I was conservative in what I switched over to use the new helper as components relying on precise timing might need careful verification

## Performance Impact

- Each `millis()` call takes approximately 1.25μs
- Before: ~29 calls to `millis()` per loop
- After: ~9 calls to `millis()` per loop
- Savings: ~25μs per main loop iteration (approximately 3% efficiency improvement)

## Implementation Details

1. Added a cached time value (`loop_component_start_time_`) in the `Application` class
2. Added an efficient getter method that returns this cached value
3. Updated all relevant components to use the cached time value
4. Improved the watchdog timer to use the cached time when available
5. Enhanced the `WarnIfComponentBlockingGuard` to work with the cached time values

## Benchmarking

The optimization was verified with the following benchmark code:

```yaml
api:
  services:
    - service: run_timer_benchmark
      then:
        - lambda: |-
            ESP_LOGI("benchmark", "Running timer benchmark...");

            const int iterations = 100000;
            volatile uint32_t dummy = 0;

            struct Timer {
              uint32_t cached_value;
              uint32_t get() const { return cached_value; }
            } timer;

            // Warm-up
            timer.cached_value = millis();
            dummy = timer.get();
            dummy = millis();

            // millis() timing
            int64_t start = micros();
            for (int i = 0; i < iterations; ++i) {
              dummy = millis();
            }
            int64_t end = micros();
            double millis_us = (end - start) / static_cast<double>(iterations);

            // Cached member access timing
            timer.cached_value = 123456789;
            start = micros();
            for (int i = 0; i < iterations; ++i) {
              dummy = timer.get();
            }
            end = micros();
            double cached_us = (end - start) / static_cast<double>(iterations);

            ESP_LOGI("benchmark", "millis():                   %.3f µs per call", millis_us);
            ESP_LOGI("benchmark", "Cached class member access: %.6f µs per call", cached_us);
```

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


esphome/core/component.cpp
```diff
@@ -240,10 +240,14 @@ void PollingComponent::stop_poller() {
...
 
-WarnIfComponentBlockingGuard::WarnIfComponentBlockingGuard(Component *component)
-    : started_(millis()), component_(component) {}
-WarnIfComponentBlockingGuard::~WarnIfComponentBlockingGuard() {
-  uint32_t blocking_time = millis() - this->started_;
+WarnIfComponentBlockingGuard::WarnIfComponentBlockingGuard(Component *component, uint32_t start_time)
+    : started_(start_time), component_(component) {}
+uint32_t WarnIfComponentBlockingGuard::finish() {
+  // Use millis() to get the most accurate blocking time
+  // This is intentionally using millis() directly rather than the cached time
```
bdraco:
```suggestion
```

esphome/core/component.h
```diff
@@ -339,11 +339,16 @@ class PollingComponent : public Component {
...
+  WarnIfComponentBlockingGuard(Component *component, uint32_t start_time);
+
+  // Finish the timing operation and return the current time
+  uint32_t finish();
+
   ~WarnIfComponentBlockingGuard();
 
  protected:
   uint32_t started_;
+  uint32_t end_time_{0};
```
bdraco:
```suggestion
```

esphome/core/application.cpp
```diff
@@ -94,7 +104,7 @@ void Application::loop() {
     delay_time = std::min(next_schedule, delay_time);
     delay(delay_time);
   }
-  this->last_loop_ = now;
+  this->last_loop_ = millis();
```
bdraco:
```suggestion
  this->last_loop_ = last_op_end_time;
```

jesserockz (approved):

bdraco:
Thanks

PR #8812, opened 2025-05-15
Remove init proto_write_buffer_ reserve

bdraco:

We always call reserve with the correct size now before writing a protobuf messages so this is unneeded and likely wastes RAM. create_buffer() already reserves the exact size needed.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


bdraco:
Seems to be slower in some cases. While it's working fine, reserve strategy could still use improvements.

I really want to rework all this to use a table as I think we can reduce the generated protobuf code by 40%. Hopefully able to make it perform just as well or better

PR #8819, opened 2025-05-16
Reduce protobuf message size calculation overhead

bdraco:

About 45% of these can be pre-calculated to the maximum size. We might go a few bytes over but its a bit faster than having to calculate it out and in all cases its lower than the pre-reserved 64 bytes

## Types of changes

- [ ] Other

- fixes <link to issue>

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

```

## Checklist:


bdraco:
I want to get rid of the dynamic allocation in

size_t padding = 0;
  size_t msg_len = 4 + payload_len + padding;
  size_t frame_len = 3 + msg_len + noise_cipherstate_get_mac_length(send_cipher_);
  auto tmpbuf = std::unique_ptr<uint8_t[]>{new (std::nothrow) uint8_t[frame_len]};
  if (tmpbuf == nullptr) {
    HELPER_LOG("Could not allocate for writing packet");
    return APIError::OUT_OF_MEMORY;
  }

  tmpbuf[0] = 0x01;  // indicator
  // tmpbuf[1], tmpbuf[2] to be set later
  const uint8_t msg_offset = 3;
  const uint8_t payload_offset = msg_offset + 4;
  tmpbuf[msg_offset + 0] = (uint8_t) (type >> 8);  // type
  tmpbuf[msg_offset + 1] = (uint8_t) type;
  tmpbuf[msg_offset + 2] = (uint8_t) (payload_len >> 8);  // data_len
  tmpbuf[msg_offset + 3] = (uint8_t) payload_len;
  // copy data
  std::copy(payload, payload + payload_len, &tmpbuf[payload_offset]);
  // fill padding with zeros
  std::fill(&tmpbuf[payload_offset + payload_len], &tmpbuf[frame_len], 0);

  NoiseBuffer mbuf;
  noise_buffer

We should be able to add a padding to the protobuf buffer and than transform it to noise in place

bdraco:
I think it would be better to keep exact sizes because that we can optimize away the inserts

PR #8868, opened 2025-05-21
[rp2040] Allow changing watchdog timeout

kuba2k2:
# What does this implement/fix?

This is a simple change, which allows to change the watchdog timeout on RP2040. It is, by default, set to the maximum value (about 8.3 seconds), but with this PR it's changeable using `watchdog_timeout` under `rp2040:`.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):** N/A
**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** 

- esphome/esphome-docs#4928

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

## Example entry for `config.yaml`:

```yaml
# Example config.yaml
rp2040:
  board: rpipico
  watchdog_timeout: 0s
```

## Checklist:
  - [x] The code change is tested and works locally.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

If user exposed functionality or configuration variables are added/changed:
  - [x] Documentation added/updated in [esphome-docs](https://github.com/esphome/esphome-docs).

PR #8869, opened 2025-05-21
[rp2040_esphost] Support ESPHost library for Wi-Fi on Pi Pico

kuba2k2:

This component supports the `ESPHost` library, present in the Arduino-Pico framework. It allows an RP2040 (Pi Pico) without the Wi-Fi module to connect to the Internet using an external, SPI-connected, ESP32 module.

More info in the documentation PR, and on [esp-hosted](https://github.com/espressif/esp-hosted) main page.

**Note:** this PR requires #8868 to function correctly. It has a `watchdog_timeout` property support.

## Types of changes

- [ ] Other

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):** esphome/esphome-docs#4929

## Test Environment

- [ ] ESP32
- [ ] ESP32 IDF
- [ ] ESP8266
- [x] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml
rp2040:
  board: rpipico
  watchdog_timeout: 0s

spi:
  clk_pin: GPIO10
  mosi_pin: GPIO11
  miso_pin: GPIO12
  interface: hardware

rp2040_esphost:
  cs_pin: GPIO13
  handshake_pin: GPIO14
  data_ready_pin: GPIO15
  reset_pin: GPIO16
  data_rate: 10MHz
```

## Checklist:

### DIFF
--- a/esphome/components/wifi/wifi_component_pico_w.cpp
+++ b/esphome/components/wifi/wifi_component_pico_w.cpp
@@ -20,7 +20,12 @@ namespace wifi {
 
 static const char *const TAG = "wifi_pico_w";
 
+#ifdef USE_RP2040_ESPHOST
+static bool s_sta_connecting = false;  // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
+#endif
+
 bool WiFiComponent::wifi_mode_(optional<bool> sta, optional<bool> ap) {
+#ifndef USE_RP2040_ESPHOST
   if (sta.has_value()) {
     if (sta.value()) {
       cyw43_wifi_set_up(&cyw43_state, CYW43_ITF_STA, true, CYW43_COUNTRY_WORLDWIDE);
@@ -32,9 +37,27 @@ bool WiFiComponent::wifi_mode_(optional<bool> sta, optional<bool> ap) {
     }
   }
   return true;
+#else
+  uint8_t current_mode = WiFi.getMode();
+  bool current_sta = current_mode & WIFI_STA;
+  bool current_ap = current_mode & WIFI_AP;
+  bool enable_sta = sta.value_or(current_sta);
+  bool enable_ap = ap.value_or(current_ap);
+  if (current_sta == enable_sta && current_ap == enable_ap)
+    return true;
+
+  uint8_t mode = 0;
+  if (enable_sta)
+    mode |= WIFI_STA;
+  if (enable_ap)
+    mode |= WIFI_AP;
+  WiFi.mode(static_cast<WiFiMode_t>(mode));
+  return true;
+#endif
 }
 
 bool WiFiComponent::wifi_apply_power_save_() {
+#ifndef USE_RP2040_ESPHOST
   uint32_t pm;
   switch (this->power_save_) {
     case WIFI_POWER_SAVE_NONE:
@@ -49,6 +72,9 @@ bool WiFiComponent::wifi_apply_power_save_() {
   }
   int ret = cyw43_wifi_pm(&cyw43_state, pm);
   return ret == 0;
+#else
+  return false;
+#endif
 }
 
 // TODO: The driver doesnt seem to have an API for this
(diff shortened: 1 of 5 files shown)


esphome/components/wifi/wifi_component_pico_w.cpp
jesserockz, line 1: As this file is named for the pico_w board specifically, we either need to rename this file, or split your new changes into a new cpp file

esphome/components/wifi/wifi_component_pico_w.cpp
```diff
@@ -207,13 +278,36 @@ network::IPAddress WiFiComponent::wifi_dns_ip_(int num) {
...
 
-void WiFiComponent::wifi_pre_setup_() {}
+void WiFiComponent::wifi_pre_setup_() {
+#ifdef USE_RP2040_ESPHOST
+  // check if initHW() returns false by verifying the esp-hosted MAC address
+  uint8_t mac_addr[6], mac_invalid[6];
+  memset(mac_addr, 0xAA, 6);
+  memset(mac_invalid, 0xAA, 6);
+  // disable the watchdog temporarily (the connection check takes time)
+  watchdog_disable();
```
jesserockz, line 296: How long on average? We should not disable the watchdog, but use the `WatchdogManager` to temporarily alter the timeout

jesserockz (changes requested):

esphome/components/wifi/wifi_component_pico_w.cpp
kuba2k2: It takes about a second when it succeeds (ESP is reset first, so it needs to boot and send an "init" message).
However, if it doesn't succeed (e.g. incorrect wiring, unsupported ESP firmware, etc.) it will block for about 10-15 seconds, even with the shortest timeout.

This is a limitation of Arduino-Pico - it doesn't use FreeRTOS and threading by default. Everything runs on a single thread (incl. the connection check and ESPHome's main loop, which also feeds the watchdog).

The maximum watchdog timeout is 8.3 seconds (ESPHome already sets it to the maximum by default), and it is not enough to perform the connection check.

Again, this is only the case when the connection *fails* - the device is inaccessible [over network] by then anyway.
I added the check mainly so that it produces an error log message, instead of just freezing for 8.3 seconds and rebooting.

esphome/components/wifi/wifi_component_pico_w.cpp
kuba2k2: I think `wifi_component_rp2040.cpp` would be a fitting name, since it would roughly match how other platforms' implementations are named.

swoboda1337:
I will create an `esp32_hosted` component... does it make sense to call this `rp2040_hosted`?

kuba2k2:
There is really no such thing as `esp32-hosted` - esp-hosted is the name of the library/protocol developed by Espressif. "Hosted" seems to mean "controlled by a host" in this case - but the RP2040 *is* the host.

For RP2040, I first developed it from scratch using the esp-hosted library (with an lwIP driver, threading, etc). It worked, but was very unstable due to lwIP not using threading in Arduino-Pico.

This component is called `rp2040_esphost` - because it uses the `ESPHost` library, not because it uses the `esp-hosted` protocol/firmware.

swoboda1337:
> `ESPHost`

Yeah I see. It doesnt use the hosted library, I thought it was. Ok

For esp32 I think esp32_hosted does make sense.

jesserockz:
> This component is called `rp2040_esphost` - because it uses the `ESPHost` library, not because it uses the `esp-hosted` protocol/firmware.

@kuba2k2 What firmware is flashed onto the esp32 though? Is it not the [slave firmware](https://github.com/espressif/esp-hosted-mcu/tree/main/slave) from esp-hosted?

Generally, ESPHome components are named after the hardware they are reading from/controlling, not the library name.

I feel like, if you can use the same esp32/firmware on the secondary/slave processor, then this should even fall under [esp32_hosted](https://github.com/esphome/esphome/tree/dev/esphome/components/esp32_hosted) and the python side can handle if the host is an esp32 or a rp2040 and do the different paths.

Both components do the same basic thing, set sdkconfig/build flags.

PR #8879, opened 2025-05-22
[i2s-audio] ensure mic task isn't pinned to a core

kahrendt:

On an ESP32, whenever a FreeRTOS task uses a floating point operation, then the task is pinned to its current core for its lifetime. Most of the audio related tasks avoid floating point operations whenever feasible, but apparently the I2S driver initialization code in esp-idf using a floating point operation to ensure a setting is valid. Since the microphone task loads the driver, this always pinned the microphone task to its current core.

This PR moves the I2S driver starting and stopping into the main loop, ensuring that the microphone task can use whatever core is free when it has work to do. I verified with ``xPortGetCoreID()`` that this fixes the situation and the task can run wherever. This should help speed things up slightly, since the high priority mic task won't interrupt other tasks as frequently since it can run on either core.

This also refactors the code so that the component's error status is only set by the main loop, allows the ``stop_driver_ `` function to unload a partially started driver, and simplifies the logic around tracking if the driver is started. 

I will be making a similar PR for the ``i2s_audio`` speaker code, but that's a bit more of a refactor.

## Types of changes

- [ ] Other

## Test Environment

- [ ] ESP32
- [x] ESP32 IDF
- [ ] ESP8266
- [ ] RP2040
- [ ] BK72xx
- [ ] RTL87xx

```yaml
# Example config.yaml

microphone:
  - platform: i2s_audio
    id: i2s_mics
    i2s_din_pin: GPIO15
    adc_type: external
    pdm: false
    sample_rate: 16000
    bits_per_sample: 32bit
    i2s_mode: secondary
    i2s_audio_id: i2s_input
    channel: stereo

```

## Checklist:

### DIFF
--- a/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
+++ b/esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
@@ -31,9 +31,9 @@ static const char *const TAG = "i2s_audio.microphone";
 
 enum MicrophoneEventGroupBits : uint32_t {
   COMMAND_STOP = (1 << 0),  // stops the microphone task
+  STATE_DRIVER_STARTED = (1 << 5),
   TASK_STARTING = (1 << 10),
   TASK_RUNNING = (1 << 11),
-  TASK_STOPPING = (1 << 12),
   TASK_STOPPED = (1 << 13),
 
   ALL_BITS = 0x00FFFFFF,  // All valid FreeRTOS event group bits
@@ -292,6 +292,7 @@ bool I2SAudioMicrophone::start_driver_() {
 
   this->status_clear_error();
   this->configure_stream_settings_();  // redetermine the settings in case some settings were changed after compilation
+  xEventGroupSetBits(this->event_group_, MicrophoneEventGroupBits::STATE_DRIVER_STARTED);
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
```diff
@@ -345,29 +346,21 @@ void I2SAudioMicrophone::stop_driver_() {
 #endif
   this->parent_->unlock();
   this->status_clear_error();
+  xEventGroupClearBits(this->event_group_, MicrophoneEventGroupBits::STATE_DRIVER_STARTED);
 }
 
 void I2SAudioMicrophone::mic_task(void *params) {
   I2SAudioMicrophone *this_microphone = (I2SAudioMicrophone *) params;
```
bdraco, line 339: Is there any chance this can be called before the driver is ready?

esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
```diff
@@ -487,14 +473,27 @@ void I2SAudioMicrophone::loop() {
...
   switch (this->state_) {
     case microphone::STATE_STARTING:
-      if ((this->task_handle_ == nullptr) && !this->status_has_error()) {
+      if (this->status_has_error()) {
+        break;
+      }
+
+      if (!(xEventGroupGetBits(this->event_group_) & MicrophoneEventGroupBits::STATE_DRIVER_STARTED)) {
+        if (!this->start_driver_()) {
+          this->status_momentary_error("I2S driver failed to start, attempting again in 1 second", 1000);
```
bdraco: so we need something like `xEventGroupClearBits(this->event_group_, MicrophoneEventGroupBits::STATE_DRIVER_STARTED);` if it fails?

esphome/components/i2s_audio/microphone/i2s_audio_microphone.cpp
kahrendt: I don't believe so. I could be missing something, but here's how I think the sequence should go.

It can only happen when ``stop_driver_`` is called, but this call happens only in two spots, line 462 and 493. Line 462 can only happen if the task actually starts and finishes. The task can only ever start if the driver is loaded due the startup sequence: 

Enter the STATE_STARTING case (line 475)
1) If the driver isn't loaded according to the cleared bit, try to start the driver (line 480, 481)
   - If the driver successfully starts, the driver started bit is set in line 295
   - If it fails, the ``start_driver_`` function will return false before reaching line 295, so the bit won't be set. It sets the momentary error and breaks out of the case (lines 482, 483)
3) Check if the task handle is valid and try to create it (lines 487-489); this can only happen if the driver is loaded due to the break on line 483.
4) The task doesn't get created successfully (line 491), so it stops the driver. It shouldn't ever reach here unless the driver is started.

I do check if the driver started bit is set in the mic task, but it is pointless if this sequence is followed. I should probably take it out in that case.

kahrendt: Ah, there is an edge case. ``stop_driver_`` can return early if something fails, so it would never clear the bit. But this means that the next time the mic starts, it will skip over installing the driver, but it may have been partially stopped.

kahrendt: I refactored this and no longer track if the driver is started. Whenever the mic starts, it tries to load the driver. If unsuccessful, it unloads whatever it started; if the task fails to starts, it unloads it completely. When the microphone stops, it also unloads the driver while deleting the task.

bdraco (approved):

jesserockz (approved):

